PR #11087, opened 2019-10-24
Added metrics for cfg

olaaun:
These were added to the metrics proxy response, but not all nodes are running metrics proxy...

Test(s) will come


freva (commented):
Does this mean we can remove these checks from metrics-proxy then?

freva:
I see `metrics-proxy` depends on `container-core`, so it should be possible for it to reuse these?

freva (approved):

PR #11091, opened 2019-10-24
GC no longer used ports.

baldersheim:
@bratseth and @arnej27959 PR
@arnej27959 I assume that we might remove the 3 unused placeholders now too ?


bratseth (approved):
👍

PR #11092, opened 2019-10-24
Revert "search-nodes config is long gone. Now it is taken over by dispatch co…"

hmusum:
Reverts vespa-engine/vespa#11090

Broke centos-build, probably some reference in CMakeLists.txt somewhere


hmusum:
Yeah, in  vespa-engine/vespa/container-search/CMakeLists.txt

PR #11093, opened 2019-10-24
Revert "Revert "search-nodes config is long gone. Now it is taken over by dispatch co…""

baldersheim:
Reverts vespa-engine/vespa#11092


baldersheim:
@hmusum or @bratseth PR

hmusum (approved):

PR #11105, opened 2019-10-25
Create one Dispatcher component per search cluster

bratseth:
This avoids creating an excessive number of connections
to search clusters when the application (incorrectly) creates
many local provider chains to the same search cluster.


baldersheim (approved):
This should be fine. Monitoring is inside dispatcher.

bratseth:
Yes, I forgot Olli already fixed that. It's all going to be fantastic then.

PR #11110, opened 2019-10-25
Extract DeploymentContext from InternalDeploymentTester

mpolden:
`MetricsReporterTest` should give you an idea of the API.

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
@@ -1,0 +1,490 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.controller.deployment;
+
+import com.yahoo.component.Version;
+import com.yahoo.config.provision.ApplicationId;
+import com.yahoo.config.provision.AthenzDomain;
+import com.yahoo.config.provision.AthenzService;
+import com.yahoo.config.provision.zone.ZoneId;
+import com.yahoo.security.KeyAlgorithm;
+import com.yahoo.security.KeyUtils;
+import com.yahoo.security.SignatureAlgorithm;
+import com.yahoo.security.X509CertificateBuilder;
+import com.yahoo.vespa.hosted.controller.Application;
+import com.yahoo.vespa.hosted.controller.ControllerTester;
+import com.yahoo.vespa.hosted.controller.Instance;
+import com.yahoo.vespa.hosted.controller.api.identifiers.DeploymentId;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.ApplicationVersion;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.JobId;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.JobType;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.RunId;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.SourceRevision;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.TesterCloud;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.TesterId;
+import com.yahoo.vespa.hosted.controller.api.integration.routing.RoutingEndpoint;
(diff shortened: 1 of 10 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
```diff
@@ -0,0 +1,496 @@
...
+        this.applicationId = TenantAndApplicationId.from(instanceId);
+        this.instanceId = instanceId;
+        this.testerId = TesterId.of(instanceId);
+        this.jobs = tester.controller().jobController();
+        this.routing = tester.serviceRegistry().routingGeneratorMock();
+        this.cloud = tester.serviceRegistry().testerCloud();
+        this.runner = runner;
+        this.tester = tester;
+        this.readyJobsTrigger = new ReadyJobsTrigger(tester.controller(), maintenanceInterval, jobControl);
+        this.nameServiceDispatcher = new NameServiceDispatcher(tester.controller(), maintenanceInterval, jobControl,
```
mpolden: Things that can be safely shared across contexts should probably continue living in `InternalDeploymentTester`, so these maintainers can be moved back or perhaps to `ControllerTester`.

jonmv: Yes.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /** Returns the last submitted application version */
+    public Optional<ApplicationVersion> lastSubmission() {
+        return Optional.ofNullable(lastSubmission);
+    }
+
+    /** Runs and returns all remaining jobs for the application, at most once, and asserts the current change is rolled out. */
+    public DeploymentContext completeRollout() {
+        readyJobsTrigger.run();
```
jonmv:
```suggestion
        triggerJobs();
```

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/InternalDeploymentTester.java
```diff
@@ -137,26 +105,41 @@ public InternalDeploymentTester() {
...
+        return nameServiceDispatcher;
+    }
+
+    /** Returns the default deployment context owned by this */
+    public DeploymentContext deploymentContext() {
+        return defaultContext;
+    }
+
+    /** Create a new deployment context for given application */
+    public DeploymentContext newDeploymentContext(String tenantName, String applicationName, String instanceName) {
```
jonmv: It's possible we should leave the instance out here, and support it in the context somehow. Let's see how it turns out.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/InternalDeploymentTester.java
```diff
@@ -137,26 +105,41 @@ public InternalDeploymentTester() {
...
-                                                           instanceName,
-                                                           nextProjectId.getAndIncrement());
+        return newDeploymentContext(tenantName, applicationName, instanceName).application();
     }
 
     /** Submits a new application, and returns the version of the new submission. */
-    public ApplicationVersion newSubmission(TenantAndApplicationId id, ApplicationPackage applicationPackage,
-                                            SourceRevision revision, String authorEmail, long projectId) {
-        return jobs.submit(id, revision, authorEmail, projectId, applicationPackage, new byte[0]);
+    public ApplicationVersion newSubmission(TenantAndApplicationId id, ApplicationPackage applicationPackage, SourceRevision sourceRevision) {
```
jonmv: We should probably inline methods like this? (Work for later.)

jonmv (approved):
Nice!

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/InternalDeploymentTester.java
mpolden: I think these methods should be removed altogether, and rather have tests always use a context.

jonmv: Exactly.

PR #11114, opened 2019-10-26
Jvenstad/use production for prod job names

jonmv:
@hmusum or @bratseth or @mpolden please review and merge.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -1581,6 +1581,7 @@ public class ApplicationApiHandler extends LoggingRequestHandler {
         }
     }
 
+    /** Returns test config for indicated job, with production deployments of the default instance. */
     private HttpResponse testConfig(ApplicationId id, JobType type) {
         // TODO jonmv: Support non-default instances as well; requires API change in clients.
         ApplicationId defaultInstanceId = TenantAndApplicationId.from(id).defaultInstance();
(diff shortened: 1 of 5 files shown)


freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1591,12 +1592,11 @@ private HttpResponse testConfig(ApplicationId id, JobType type) {
...
         var testedZone = type.zone(controller.system());
 
-        // This should not happen, unless one asks for test-config for a production job, which one shouldn't have use for.
-        if (deployments.stream().anyMatch(deploymentId ->      deploymentId.zoneId().equals(testedZone)
-                                                          && ! deploymentId.applicationId().equals(id)))
-            throw new IllegalStateException("Conflict: " + testedZone + " contains both default and " + id.instance().value() + " instances");
+        // If a production job is specified, the production deployment of the _default instance_ is the relevant one,
+        // as user instances should not exist in prod. TODO jonmv: Remove this when multiple instances are supported (above).
+        if ( ! type.isProduction())
+            deployments.add(new DeploymentId(id, testedZone));
```
hakonhall: Collectors.toSet() makes no guarantee about the mutability of the returned Set: "There are no guarantees on the type, mutability, ...". Because of this I have never tried nor seen a `.add` so I guess it works... But consider using the toSet(Supplier) instead.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1591,12 +1592,11 @@ private HttpResponse testConfig(ApplicationId id, JobType type) {
                                     .collect(Collectors.toSet());
         var testedZone = type.zone(controller.system());
 
-        // This should not happen, unless one asks for test-config for a production job, which one shouldn't have use for.
-        if (deployments.stream().anyMatch(deploymentId ->      deploymentId.zoneId().equals(testedZone)
-                                                          && ! deploymentId.applicationId().equals(id)))
-            throw new IllegalStateException("Conflict: " + testedZone + " contains both default and " + id.instance().value() + " instances");
+        // If a production job is specified, the production deployment of the _default instance_ is the relevant one,
```
hakonhall: So tests are only run on the default instance because non-default instances are assumed to be tested implicitly by testing the default?

But someone set up only non-default instances?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
jonmv: The assumption is that only the "default" instance is defined in deployment.xml for now, and config for a production job is therefore aimed at one of its deployments — not the user's `dev` instance, and not the user's non-existent production deployments :P Will need to expose instances in the test config API, but will do later.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
jonmv: Will do.

PR #11121, opened 2019-10-26
Deconstruct all components in one task.

gjoranv:
Please review. Will merge after #11120 has passed system tests.

The executor has only one thread, so this should not make a
significant difference.


hmusum (approved):

baldersheim (commented):
:+

PR #11125, opened 2019-10-27
Increase client timeout for config requests

hmusum:
We have seen that loaded config servers sometimes have short periods where requests time out. Increase client timeout.


bratseth (approved):

gjoranv:
👍

PR #11131, opened 2019-10-28
Add "safe mode" flag support to vespa-set-node-state tool

vekterli:
@geirst please review this tiny bit of happy Perl code 🙈 System testing of the flag will come shortly.

Enables the tool to trigger the same state transition checks that are
used by orchestration components. Caveat: due to the REST API currently
returning HTTP 200 OK even for denied transitions (but with a JSON payload
indicating failure instead), the tool will return with status code 0
even if the state change failed.


geirst (approved):
👍

PR #11134, opened 2019-10-28
Revert "Reapply "Create versioned zookeeper bundles""

hmusum:
Reverts vespa-engine/vespa#11133

In case something breaks...


hakonhall (approved):

hmusum:
Not needed

PR #11137, opened 2019-10-28
Store commit details of latest version

mpolden:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/versions/VersionStatus.java
```diff
@@ -263,8 +260,8 @@ private static VespaVersion createVersion(DeploymentStatistics statistics,
         }
 
         // Preserve existing commit details if we've previously computed status for this version
-        var commitSha = controllerVersion.commitSha();
-        var commitDate = controllerVersion.commitDate();
+        var commitSha = latestVersion.commitSha();
+        var commitDate = latestVersion.commitDate();
```
jonmv: Shouldn't this be the details of the version `statistics.version()`?

mpolden: Ideally yes, but the details only exist in a `ControllerVersion` and the first time a new `statistics.version()` appears it will be due to a controller upgrade...

jonmv:
Allright!

PR #11141, opened 2019-10-28
Fix application handler bindings to work with instance

freva:
Currently the old logs path is matched before the new one (because `logs` is lexicographically before `environment`), likely there is a bug in `GlobPattern::compareTo`, but I don't wanna risk changing that cause that just might break everything.

At least this removes duplication between `ApplicationHandler.java` and `services.xml`.


hmusum (approved):
Great!

PR #11142, opened 2019-10-28
Gjoranv/allow duplicate bundles

gjoranv:
Take a look, but do not merge. I have yet to get proof of concept in a proper test.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
+++ b/container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
@@ -174,6 +174,7 @@ public class BundleLoader {
     public synchronized Set<Bundle> use(List<FileReference> bundles) {
         Set<Bundle> bundlesToUninstall = retainOnly(bundles);
         osgi.allowDuplicateBundles(bundlesToUninstall);
+        log.info(() -> bundlesToUninstall.isEmpty() ? "Adding bundles to allowed duplicates: " + bundlesToUninstall : "");
         install(bundles);
         startBundles();
 
--- a/container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
+++ b/container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
@@ -10,7 +10,6 @@ import org.osgi.framework.Bundle;
 import org.osgi.framework.wiring.BundleRevision;
 
 import java.io.File;
-import java.util.Arrays;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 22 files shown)


container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
```diff
@@ -47,16 +53,19 @@ public BundleLoader(Osgi osgi) {
     }
 
     /** Returns the number of bundles installed by this call. */
```
baldersheim: Comment is outdated

baldersheim (commented):
This is unfamiliar territory. I think it is better that we go through this one together.

container-core/src/main/java/com/yahoo/container/core/config/BundleLoader.java
gjoranv: Thanks!

baldersheim (approved):
Merge at will

gjoranv:
Awaiting merge until a system test run has successfully started.

jdisc_core/src/main/java/com/yahoo/jdisc/application/OsgiFramework.java
```diff
@@ -60,6 +60,8 @@
     /**
      * Synchronously refresh all bundles currently loaded. Once this method returns, the
      * class loaders of all bundles will reflect on the current set of loaded bundles.
+     *
+     * NOTE: This method is no longer used by the Jdisc container framework, but kept for completeness.
```
bjorncs: Mark as deprecated?

container-core/src/main/java/com/yahoo/osgi/OsgiImpl.java
```diff
@@ -11,12 +11,14 @@
...
 import java.util.Collection;
 import java.util.List;
+import java.util.logging.Logger;
 
 /**
  * @author Tony Vaagenes
  * @author bratseth
  */
 public class OsgiImpl implements Osgi {
+    private static Logger log = Logger.getLogger(OsgiImpl.class.getName());
```
bjorncs: Missing a `final`

bjorncs (approved):

jdisc_core/src/main/java/com/yahoo/jdisc/application/OsgiFramework.java
gjoranv: I originally did, but opted against it. It's really a natural part of the framework, and hence just a comment that it's not used by the jdisc container. It is extensively used in unit/integration tests.

PR #11144, opened 2019-10-29
Use DiskSpeed.any for testers

bratseth:


freva (changes requested):
This wont work, these variables are just passed to `servicesXml()` which generates the `services.xml` for the tester app, but it does not write `disk-speed` in `<resources>`. Need to fix: https://github.com/vespa-engine/vespa/blob/9a1efcb9aca844b1e723f740dff8f908161b505c/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java#L707

bratseth:
Thanks, added.

freva (approved):

jonmv (approved):

PR #11151, opened 2019-10-29
Store commit details of latest version

mpolden:
Since the controller version defaults to VTAG, we have to use a version that's
always higher.


jonmv (approved):
Delightful :)

PR #11154, opened 2019-10-30
Update UpgraderTest

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/UpgraderTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/UpgraderTest.java
@@ -3,30 +3,24 @@ package com.yahoo.vespa.hosted.controller.maintenance;
 
 import com.yahoo.component.Version;
 import com.yahoo.config.provision.Environment;
-import com.yahoo.config.provision.RegionName;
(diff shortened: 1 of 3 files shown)


jonmv:
I removed some tests that were well covered by others.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/UpgraderTest.java
```diff
@@ -37,570 +31,431 @@
...
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  * @author bratseth
  */
 public class UpgraderTest {
 
+    InternalDeploymentTester tester = new InternalDeploymentTester().atHourOfDay(5);
```
mpolden: Should be `private final`.

mpolden (commented):
Great!

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/UpgraderTest.java
jonmv: Yeah.

PR #11160, opened 2019-10-30
Redesign maintenance grace period edge event emitting

vekterli:
@geirst please review.

This has two separate fixes for timer-induced event/log spamming for certain edge cases in the cluster controller. One is to use the working state rather than the published state for timer decisions, the other is more fundamental and moves event emitting for the observed edge case out of the timer code entirely.


geirst (approved):
👍

PR #11162, opened 2019-10-30
Check capacity by requested, not assigned resources

bratseth:


kkraune (approved):
I supposed the requested resources are all populated at the time the maintainer runs first time with this?

bratseth:
Yes, but could be populated with the assigned resources, as that is the fallback until a redeployment fills in the right value.

PR #11164, opened 2019-10-30
Allow fetching metrics for all consumers

olaaun:


metrics-proxy/src/main/java/ai/vespa/metricsproxy/http/ValuesFetcher.java
```diff
@@ -49,6 +49,10 @@ public ValuesFetcher(MetricsManager metricsManager,
                 .collect(Collectors.toList());
     }
 
+    public List<MetricsPacket> fetchAllMetrics() throws JsonRenderingException {
```
gjoranv: Please update `fetch(String)` to use this method instead of using metricsManager directly, to avoid duplication.

gjoranv (changes requested):

gjoranv (approved):

PR #11167, opened 2019-10-31
It is better to just lookup a function by name from the hierarchical …

baldersheim:
…maps in shadoworder than creating a new joined map for every lookup.

@bratseth or @lesters PR
This improves rankprofile deploy from 30s to 20s for a large application.


bratseth (approved):

lesters:
Nice. Have you coined a new term with "shadoworder"?

PR #11170, opened 2019-10-31
Jvenstad/convert more deployment tests

jonmv:
@mpolden please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -250,8 +250,7 @@ public InternalStepRunner(Controller controller) {
             return Optional.of(running);
         }
         catch (ConfigServerException e) {
-            if (   e.getErrorCode() == OUT_OF_CAPACITY && type.isTest()
```
mpolden: 💯

mpolden (approved):
Great!

PR #11172, opened 2019-10-31
Simplify ConfigServerRestExecutorImpl

freva:
This shouldn't have an functional changes, just some cleanup refactoring to make it easier to reuse this for future `/configserver/v1` proxy API.

The discovery API removed in first commit is not used as it is not reachable today since it will be rejected here: https://github.com/vespa-engine/vespa/blob/c154ec58f64daa83c1f3631fa5695144ae46cdaa/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/zone/v2/ZoneApiHandler.java#L79

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/proxy/ConfigServerRestExecutorImpl.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/proxy/ConfigServerRestExecutorImpl.java
@@ -17,7 +17,6 @@ import org.apache.http.client.methods.HttpPatch;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpRequestBase;
-import org.apache.http.conn.ssl.X509HostnameVerifier;
 import org.apache.http.entity.InputStreamEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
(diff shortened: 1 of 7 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/proxy/ConfigServerRestExecutorImpl.java
```diff
@@ -86,25 +83,17 @@ public ProxyResponse handle(ProxyRequest proxyRequest) throws ProxyException {
         // TODO: Use config server VIP for all zones that have one
```
bjorncs: Do you know why we do not use the configserver VIPs for non-aws zones?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/proxy/ConfigServerRestExecutorImpl.java
```diff
@@ -238,56 +231,19 @@ private boolean queueFirstServerIfDown(List<URI> allServers, ProxyRequest proxyR
         return true;
     }
 
-    @SuppressWarnings("deprecation")
+    private HostnameVerifier createHostnameVerifier(ZoneId zoneId) {
+        return new AthenzIdentityVerifier(Set.of(zoneRegistry.getConfigServerHttpsIdentity(zoneId)));
```
bjorncs: Note: this hostname verifier will not accept the controller certificate.

bjorncs (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/proxy/ConfigServerRestExecutorImpl.java
freva: Primarily its because we use /zone/v2 in dashboard to access application config which is only available on the configserver to which the application was deployed (See VESPA-11845).

PR #11175, opened 2019-10-31
Extend the immutable interface and cache allFields in rankprofile

baldersheim:
@bjorncs PR
@bratseth FYI
Brings perfomance further down from 15s to 13s

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
@@ -1,9 +1,13 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.searchdefinition;
 
+import com.yahoo.config.application.api.ApplicationPackage;
 import com.yahoo.searchdefinition.document.ImmutableSDField;
+import com.yahoo.searchdefinition.document.SDField;
 import com.yahoo.vespa.documentmodel.SummaryField;
 
+import java.io.Reader;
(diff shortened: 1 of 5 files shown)


bratseth (approved):

config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
```diff
@@ -15,11 +19,22 @@
  */
 public interface ImmutableSearch {
 
+    String getName();
+    Index getIndex(String name);
+    SDField getConcreteField(String name);
```
bjorncs: Consider changing the return type to `ImmutableSDField`

config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
```diff
@@ -15,11 +19,22 @@
...
+    List<SDField> allConcreteFields();
```
bjorncs: Consider changing the return type to `List<ImmutableSDField>`

bjorncs (approved):

config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
baldersheim: Will do in a follow up PR

config-model/src/main/java/com/yahoo/searchdefinition/ImmutableSearch.java
baldersheim: Will do in a follow up PR

PR #11176, opened 2019-10-31
Jvenstad/convert more deployment tests

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
@@ -214,7 +214,7 @@ public class MetricsReporterTest {
 
         context.submit(applicationPackage).deploy();
         reporter.maintain();
-        assertEquals("Deployment queues name services requests", 6, metrics.getMetric(MetricsReporter.NAME_SERVICE_REQUESTS_QUEUED).intValue());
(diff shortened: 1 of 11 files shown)


jonmv:
Note second commit.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
```diff
@@ -214,7 +214,7 @@ public void name_service_queue_size_metric() {
 
         context.submit(applicationPackage).deploy();
         reporter.maintain();
-        assertEquals("Deployment queues name services requests", 6, metrics.getMetric(MetricsReporter.NAME_SERVICE_REQUESTS_QUEUED).intValue());
+        assertEquals("Deployment queues name services requests", 15, metrics.getMetric(MetricsReporter.NAME_SERVICE_REQUESTS_QUEUED).intValue());
```
mpolden: Why does this increase?

jonmv: The application gets a rotation earlier (the first time it rolls out), so deployments in test (x1) and staging (x2) also cause 3 requests. This would previously happen on every subsequent rollout of the application, as well.

jonmv: That is, it would be 6 only the first time an application fully deployed, and 15 on every following rollout.

mpolden (approved):

PR #11178, opened 2019-10-31
Precompute the fixed set of feature types once.

baldersheim:
@bratseth PR This brings it down from 13s to 10s -> DONE
In total 3x for deploy and 20x for getConfig() for a large application.


bratseth (approved):
Great

PR #11183, opened 2019-10-31
Create /configserver/v1

freva:
`/configserver/v1` is a new proxy API on controller to configservers. Differences from `/zone/v2`:
* `/configserver/v1` proxies only select APIs, currently `/flags/v1`, `/nodes/v2/`, `/orchestrator/v1`.
* `/configserver/v1` is limited to operators only.
* `/configserver/v1` supports the controller "zone".
* `/configserver/v1` only proxies through VIP (`/zone/v2` uses VIP for AWS zones, iterates through config servers on-prem due to VESPA-11845)

This API will be used for the console operator dashboard, but we cloud also start using it in `hv`, this will simplify the code that deals with the controller zone for `node` and `flag` commands, while the `orchestrator` command does not even support it yet.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
@@ -20,7 +20,8 @@ import java.util.Set;
 enum PathGroup {
 
     /** Paths exclusive to operators (including read), used for system management. */
-    classifiedOperator("/configserver/v1/{*}"),
+    classifiedOperator(Optional.of("/api"),
+                       "/configserver/v1/{*}"),
 
     /** Paths used for system management by operators. */
     operator("/controller/v1/{*}",
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
@@ -20,7 +20,8 @@ import java.util.Set;
 enum PathGroup {
 
     /** Paths used for system management by operators. */
-    operator("/controller/v1/{*}",
+    operator("/configserver/v1/{*}",
+             "/controller/v1/{*}",
              "/flags/v1/{*}",
              "/nodes/v2/{*}",
              "/orchestrator/v1/{*}",
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
@@ -19,9 +19,11 @@ import java.util.Set;
  */
 enum PathGroup {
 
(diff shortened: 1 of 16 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    private HttpResponse notFound(Path path) {
+        return ErrorResponse.notFoundError("Nothing at " + path);
+    }
+
+    private URI getEndpoint(ZoneId zoneId) {
+        return CONTROLLER_ZONE.equals(zoneId) ? zoneRegistry.apiUrl() : zoneRegistry.getConfigServerVipUri(zoneId);
+    }
+
+    private HostnameVerifier createHostnameVerifier(ZoneId zoneId) {
+        if (CONTROLLER_ZONE.equals(zoneId)) return null;
```
freva: I assume this will make it use the default hostname verifier?

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
```diff
@@ -20,7 +20,8 @@
 enum PathGroup {
 
     /** Paths used for system management by operators. */
-    operator("/controller/v1/{*}",
+    operator("/configserver/v1/{*}",
```
freva: I still don't understand this... Now `/configserver/v1` is on the same list as `/zone/v2`, but `/zone/v2` read is available for everyone, yet if I don't have `/configserver/v1` in here, all tests in `ConfigServerApiHandlerTest` fail with 403 (including GETs).

If this is enough to lock down read & write to operators, why does `/zone/v2` read work?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
bjorncs: Most likely, I would opt for `new DefaultHostnameVerifier()` to be sure.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
freva: Figured it out

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
freva: Fixed

freva:
@bjorncs PTAL.

bjorncs (approved):
👍 🥇

MJPIPER99:
We Wish we had a (Call complete) while speaking & text & Internet service
combination package. Add On FireFox/Microsoft.

PR #11186, opened 2019-11-01
Print error message if deployment fails, not getting compile version

jonmv:
@hmusum please review.

This would make it impossible to deploy under user tenants to a new application. I don't really want people to do this, as we'd like them to stop using the user tenants, but we said it should continue to work.


jonmv:
Long term we just need to expose more info through the API, so we can display correct error messages to the end user.

hmusum (approved):

hmusum:
Not relevant anymore when people are using the ```sandbox```tenant when deploying to dev

PR #11193, opened 2019-11-01
Define the ONCE node report type

hakonhall:
Also defines a snippet generator to limit the amount of text written to node
repo.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/process/ChildProcessException.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/process/ChildProcessException.java
@@ -2,6 +2,8 @@
 
 package com.yahoo.vespa.hosted.node.admin.task.util.process;
 
+import com.yahoo.vespa.hosted.node.admin.task.util.text.SnippetGenerator;
+
 /**
  * Base class for child process related exceptions, with a util to build an error message
  * that includes a large part of the output.
@@ -10,10 +12,7 @@ package com.yahoo.vespa.hosted.node.admin.task.util.process;
  */
 @SuppressWarnings("serial")
 public abstract class ChildProcessException extends RuntimeException {
-    private static final int MAX_OUTPUT_PREFIX = 200;
-    private static final int MAX_OUTPUT_SUFFIX = 200;
-    // Omitting a number of chars less than 10 or less than 10% would be ridiculous.
-    private static final int MAX_OUTPUT_SLACK = Math.max(10, (10 * (MAX_OUTPUT_PREFIX + MAX_OUTPUT_SUFFIX)) / 100);
+    private static final SnippetGenerator snippetGenerator = new SnippetGenerator();
 
     /**
      * An exception with a message of the following format:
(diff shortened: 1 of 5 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public class SnippetGenerator {
+
+    private static final int ASSUMED_OMIT_TEXT_LENGTH = omitText(1234).length();
+
+    private static String omitText(int omittedLength) { return "[..." + omittedLength + " chars omitted]"; }
+
+    /** Returns a snippet of approximate size. */
+    public String makeSnippet(String text, int sizeHint) {
+        if (text.length() < sizeHint) return text;
```
freva: This should be `<=`

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
```diff
@@ -0,0 +1,30 @@
...
+        int maxSuffixLength = Math.max(0, (sizeHint - ASSUMED_OMIT_TEXT_LENGTH) / 2);
+        int maxPrefixLength = Math.max(0, sizeHint - ASSUMED_OMIT_TEXT_LENGTH - maxSuffixLength);
+        String snippet =
+                text.substring(0, maxPrefixLength) +
+                omitText(text.length() - maxPrefixLength - maxSuffixLength) +
+                text.substring(text.length() - maxSuffixLength);
+
+        // It would be silly to return a snippet when the full text is barely longer.
+        // Note: Say ASSUMED_OMIT_TEXT_LENGTH=23: text will be returned whenever sizeHint<23 and text.length()<28.
+        if (text.length() <= 1.05 * snippet.length() + 5) return text;
```
freva: This could be done without generating the snippet first?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.node.admin.task.util.text;
+
+/**
+ * @author hakon
+ */
+public class SnippetGenerator {
```
freva: I spent way too much time trying to optimize the original version on Friday, but couldn't get rid of the `while` loop while keeping the result perfect. But you can get it within 1 character error by doing `int assumedDigits = (int) (Math.log10(text.length() - sizeHint) + 1);` and then remove the `while` loop.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
freva: Also you're better off returning `text` if it smaller than omit text length.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/process/ChildProcessException.java
```diff
@@ -36,44 +35,13 @@ protected ChildProcessException(RuntimeException cause,
...
-                    .append("... [")
-                    .append(possiblyHugeOutput.length() - maxOutputPrefix - maxOutputSuffix)
-                    .append(" chars omitted] ...")
-                    .append(possiblyHugeOutput.substring(possiblyHugeOutput.length() - maxOutputSuffix));
-        }
-
-        stringBuilder.append("'");
-
-        return stringBuilder.toString();
+    private static String makeSnippet(String problem, String commandLine, String possiblyHugeOutput) {
```
freva: :+1:

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
hakonhall: Done

By "omit text length" I take it you mean ASSUMED_OMIT_TEXT_LENGTH and not omitText(text.length).length(). Done.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/text/SnippetGenerator.java
hakonhall: Yes but the implementation will complicate the code. I'll do it.

hakonhall: Code ended up being just fine.

PR #11216, opened 2019-11-05
Jvenstad/remove more deprecated deployment spec usage

jonmv:
@mpolden please review.


jonmv:
@bratseth last commit.

bratseth:
👍 re the last commit. So with this we support running multiple instances in prod?

jonmv:
Yes.

jonmv:
@bratseth feel free to review the now last commit: removing `singleInstance` things from `DeploymentSpec`.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
```diff
@@ -661,17 +661,21 @@ public void testApplicationApi() {
...
-        // Sixth attempt has a multi-instance deployment spec, and fails.
+        // Sixth attempt has a multi-instance deployment spec, and is accepted.
         ApplicationPackage multiInstanceSpec = new ApplicationPackageBuilder()
                 .instances("instance1,instance2")
+                .systemTest()
+                .stagingTest()
                 .environment(Environment.prod)
                 .region("us-central-1")
                 .parallel("us-west-1", "us-east-3")
+                .endpoint("default", "foo", "us-central-1", "us-west-1", "us-east-3")
```
mpolden: Is there a test that verifies endpoint configuration combined with multiple instances?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentSteps.java
```diff
@@ -39,9 +40,10 @@ public DeploymentSteps(DeploymentInstanceSpec spec, Supplier<SystemName> system)
 
     /** Returns jobs for this, in the order they are declared */
```
bratseth: Not: This javadoc should now say "in the order they should be executed" now.

bratseth (approved):
Great. Assuming you know it isn't too early to remove hasSingleInstance

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentSteps.java
jonmv: Hehe, yes, I keep changing that doc back and forth :p Will replace the whole class next, to orchestrate instances, anyway.

jonmv:
There are no usages of the deprecated stuff left, and I believe the only code that cares about test/staging being inside/outside of instances is the deployment orchestration code.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
jonmv: What calls are made against the rotation repository?

jonmv: We haven't really supported multiple instances in the deployment spec before, so no. I'll add one. Can you think of other places such tests are needed?

jonmv: Ah, endpoints.

jonmv:
Adding a test now, but there is a hard-coded "ignore instance in global endpoint URLs" that's blocking me. Why is that there?

jonmv:
@mpolden PTAL, note last to second commit.

jonmv:
@mpolden PTAL :)

mpolden (approved):

PR #11220, opened 2019-11-05
add metrics for application allocations

andreer:

@smorgrav FYI - didn't want to gate this on you but I think you're the one who knows which of the app, applicationId dimensions should be and are actually used?


smorgrav:
I've lost track of which dimensions that we necessary and used at various places. I imagine that it is important for alerting (ie alert rules) and for metric dashboard (e.g list metrics on the application dashboard). @ldalves or @yngveaasheim are better candidates for this knowledge

olaaun (approved):
LGTM. applicationId is used for alerting rules, while app is used in various dashboards.

Next step is emitting this metric without vespaVersion.

ldalves:
@andreer please take a look here: https://git.ouroath.com/vespa/vespa-yahoo/blob/28fdd9be1c10cb5ef8ead89a9bae316cee99753a/hosted/host-admin/src/main/java/com/yahoo/vespa/hosted/hostadmin/task/docker/HostedNodeAgent.java#L575

PR #11223, opened 2019-11-05
Make it possible specify a cpu speedup factor for a flavor

bratseth:
This allows us to model the speed difference between different
cpu generations.


freva (approved):
LGTM, though I think it would be better to do the rollout factor calculation in the code that generates `node-flavors.xml`

bratseth:
Thanks. I like to have the true information reach this side. Anyway, it is short lived.

PR #11224, opened 2019-11-06
Gjoranv/reapply allow duplicate bundles

gjoranv:
Only the second last commit is significant, except for the two re-apply commits.
* Avoid making the pre-installed bundles (X-JDisc-Preinstall-Bundle) invisible for component construction. 
* Avoid uninstalling above bundles when they have been added back by another bundle.

FYI: @kkraune


baldersheim (approved):

kkraune:
Thanks! new files should have copyright header

PR #11229, opened 2019-11-06
Do not follow links when deleting recursively

mpolden:
...like the Javadoc already says. Discovered the hard way :sweat_smile:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/UnixPath.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/UnixPath.java
@@ -178,8 +178,17 @@ public class UnixPath {
         return this;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


hakonhall (approved):

hakonhall (commented):
Wait, isDirectory() should still be true if it points to a directory. It should be transparent.

hakonhall:
I think isDirectory() should be documented to be either a directory or a symlink pointing to a directory. If explicit testing is needed - as in recursiveDelete(), there should be a call to a new UnixPath.isSymbolicLink.

mpolden:
> I think isDirectory() should be documented to be either a directory or a symlink pointing to a directory.

Sure, will make that change before merging.

mpolden:
@hakonhall PTAL.

tokle (approved):

hakonhall (commented):
👍

PR #11231, opened 2019-11-06
Add '-bundle.jar' to file suffixes to copy to the application.

gjoranv:
This is the default classifier name when using the 'AttachBundle' config setting.


baldersheim (approved):

bjorncs (commented):
👍

PR #11243, opened 2019-11-07
Pick all slobroks by index

bratseth:
@arnej27959 I think moving some slobroks should be extremely safe now?


arnej27959 (approved):
yes, we think this should be OK now.

PR #11244, opened 2019-11-07
Let Z-curve attribute be authoritative for position field existence

### LINKED ISSUE #11208: Partial update on position field fails despite HTTP 200
Hello! I have a position field called `latLong` and when trying to do a partial update (via HTTP PUT at via HTTP PUT `/document/v1/cluster/sd/docid/1` like so
```
{   
 "fields": {
        "latLong": {
            "assign": "N37.401;W121.996"
        }
    },
    "create": true
}
```
I get back an HTTP 200. However, issuing a subsequent HTTP GET at the path `/document/v1/cluster/sd/docid/1` does not show a `latLong` field at all. If I do a PUT (via HTTP POST `/document/v1/cluster/sd/docid/1`) like so:
```
{
    "fields": {
        "latLong": "N37.401;W121.996"
    }
}
```
doing a subsequent HTTP GET then shows that the `latLong` field is correctly inserted. 
```
"latLong": { "x": -121996000, "y": 37401000 },
```

It looks like an incremental update of `latLong` will now work, but not when `latLong` never previoiusly existed previously on the document. This seems like a bug since I get the expected behavior for other field types (i.e. that an incremental update creates it if it doesn't exist)?

vekterli:

Avoids inconsistencies caused by concrete position struct fields not
being created or cleared by partial updates, only the Z-curve attribute
fields.

This fixes #11208

### DIFF
--- a/searchcore/src/tests/proton/server/documentretriever_test.cpp
+++ b/searchcore/src/tests/proton/server/documentretriever_test.cpp
@@ -134,10 +134,12 @@ const TensorDataType tensorDataType(ValueType::from_spec(tensor_spec));
 
 struct MyDocumentStore : proton::test::DummyDocumentStore {
     mutable std::unique_ptr<Document> _testDoc;
+    bool _set_position_struct_field;
 
     MyDocumentStore()
         : proton::test::DummyDocumentStore(),
-          _testDoc()
+          _testDoc(),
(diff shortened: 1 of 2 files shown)


geirst (approved):
👍

toregge (approved):

PR #11246, opened 2019-11-08
Smorgrav/user api add roles

smorgrav:
Expanding the user management API to better reflect normal usage (you add/remove multiple roles to a user vs adding multiple users to a role).

Turns out this was supported and tested in the rest api - so this change only enables the implementations to be more efficient in dealing with these requests.


jonmv (commented):
`∠( ᐛ 」∠)＿`

PR #11251, opened 2019-11-08
Log parent hosts that are not ready

hmusum:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
```diff
@@ -28,6 +30,8 @@
  */
 class Activator {
 
+    private static final Logger logger = Logger.getLogger(InfraDeployerImpl.class.getName());
```
freva: Wrong class name?

hmusum: Fixed

freva (approved):

PR #11267, opened 2019-11-11
Always try to build config model for old versions

hmusum:
Try to build old config models, but do not fail deployment if skip-old-config-models validation override is true and building an old model fails. In manually deployed zones we previously did not build old config models, since skip-old-config-models was always true. With this change we will try to build those old models, but not fail if building one or more old models fails.


bratseth (approved):
Nice

PR #11268, opened 2019-11-11
Count open files on Darwin.

toregge:
@baldersheim : please review


baldersheim (commented):
This is the wrong place for this functionality. This should be down in fastos.

PR #11276, opened 2019-11-12
Allow removing half of the cluster controllers

bratseth:


vekterli (approved):
:+1: (with the assumption of ZK quorum reconfiguration happening prior to the nodes actually disappearing)

PR #11281, opened 2019-11-12
Balder/compress result space

baldersheim:
@vekterli PR

### DIFF
--- a/document/src/vespa/document/select/resultlist.cpp
+++ b/document/src/vespa/document/select/resultlist.cpp
@@ -1,7 +1,7 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "resultlist.h"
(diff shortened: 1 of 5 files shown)


document/src/vespa/document/select/resultlist.cpp
```diff
@@ -96,39 +97,56 @@ ResultList::combineVariables(
 ResultList
 ResultList::operator&&(const ResultList& other) const
 {
-    ResultList result;
+    ResultList results;
 
     // TODO: optimize
+    vespalib::hash_set<uint32_t> resultForNoVariables;
```
vekterli: Since we're dealing with a fixed set of possible values, this could also be a bit set to avoid added overhead with allocation per operator invocation. Same goes for the other places we do this.

vekterli (approved):
:+1:

document/src/vespa/document/select/resultlist.cpp
baldersheim: Fixed

PR #11284, opened 2019-11-12
Add flag to control reboot interval

hakonhall:
- Adds reboot-interval-in-days flag that are read on every 25m maintain().
 - Changes the reboot algorithm to evenly distribute reboots after 1x-2x reboot intervals after the last reboot. Reboots are now guaranteed to be scheduled before 2x reboot interval.
 - Removes the possibility of overriding durations from environment, which was almost never used.


flags/src/main/java/com/yahoo/vespa/flags/custom/NodeMaintainerDurations.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * {@code NodeRepositoryMaintenance} depends on various durations, e.g. the reboot interval. This class
+ * defines a serialization format for specifying such durations as a JSON object with keys matching
+ * the name (reboot_interval) and with the value being the duration in second as an integer.
```
freva: Let the names be the same as class names? That's what we use to enable/disable maintainers.

flags/src/main/java/com/yahoo/vespa/flags/custom/NodeMaintainerDurations.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author hakonhall
+ */
+public class NodeMaintainerDurations {
```
freva: Add `hashCode()` and `equals()`

flags/src/main/java/com/yahoo/vespa/flags/custom/NodeMaintainerDurations.java
hakonhall: The current names are fairly descriptive, while attaching a duration to a Maintainer's subclass name may not be. Also, a Maintainer's interval may be different from the duration controlled here - for instance making smaller adjustments to fail_grace in main would typically not affect the NodeFailer's Maintainer interval.

freva (approved):
LGTM, but I think it would be better to get the flag value in `NodeRebooter`, thereby avoiding having to restart for this to take effect. The interval is in practice overwritten to 25 minutes anyway.

flags/src/main/java/com/yahoo/vespa/flags/custom/NodeMaintainerDurations.java
hakonhall: NA

flags/src/main/java/com/yahoo/vespa/flags/custom/NodeMaintainerDurations.java
hakonhall: NA

hakonhall:
Done

freva (approved):
:1st_place_medal:

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRebooter.java
```diff
@@ -49,13 +52,32 @@ protected void maintain() {
...
+        if (overdue.get().isNegative())
             return false;
-        else // schedule with a probability such that reboots of nodes are spread roughly over the reboot interval
-            return random.nextDouble() < (double) interval().getSeconds() / (double)rebootInterval.getSeconds();
+
+        // Use a probability such that each maintain() schedules the same number of reboots,
+        // as long as 0 <= overdue <= rebootInterval, with the last maintain() in that interval
+        // naturally scheduling the remaining with probability 1.
+
+        int configServers = 3;
```
mpolden: Should be determined automatically by using `CuratorDatabaseClient#cluster`.

hakonhall: Fixed in #11313

PR #11293, opened 2019-11-14
Enable free lists for multi-value mapping used in all multi-value att…

geirst:
…ributes.

This should reduce the need for compaction which is CPU intensive for some applications.

@toregge please review
@baldersheim FYI

### DIFF
--- a/searchlib/src/tests/attribute/compaction/attribute_compaction_test.cpp
+++ b/searchlib/src/tests/attribute/compaction/attribute_compaction_test.cpp
@@ -174,9 +174,9 @@ TEST_F("Test that no compaction of int8 array attribute increases address space 
     f.populate(range1, 1000);
     f.hammer(range2, 101);
     AddressSpace afterSpace = f.getMultiValueAddressSpaceUsage("after");
-    // 100 * 1000 dead arrays due to new values for docids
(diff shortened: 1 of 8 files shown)


searchlib/src/tests/attribute/compaction/attribute_compaction_test.cpp
```diff
@@ -174,9 +174,9 @@ TEST_F("Test that no compaction of int8 array attribute increases address space
     f.populate(range1, 1000);
     f.hammer(range2, 101);
     AddressSpace afterSpace = f.getMultiValueAddressSpaceUsage("after");
-    // 100 * 1000 dead arrays due to new values for docids
+    // Only 1000 dead arrays (due to new values for docids) as free lists are used.
```
toregge: The main body of this test and the next test should be factored out to a helper function that optionally holds an attribute guard during populate/hammer, and if guard was held, releases it and calls commit(true) and commit() before getting multi value address space usage and returning afterSpace.dead(). When guard is not held, free lists will be used, and no compaction will be performed. When guard has been held, compaction will be performed if it has been enabled.

toregge (approved):

toregge (approved):

PR #11302, opened 2019-11-14
Write TLS config to zookeeper config file based on Vespa config

hmusum:
tlsForQuorumCommunication will always be off for now, next step is to wire in use of feature flag to control setting in Vespa config zookeeper-server config

### DIFF
--- a/zookeeper-server/zookeeper-server-3.5/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperServerImpl.java
+++ b/zookeeper-server/zookeeper-server-3.5/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperServerImpl.java
@@ -5,6 +5,8 @@ import com.google.inject.Inject;
 import com.yahoo.cloud.config.ZookeeperServerConfig;
 import com.yahoo.component.AbstractComponent;
 import com.yahoo.log.LogLevel;
+import com.yahoo.security.tls.TlsContext;
+
 import static com.yahoo.vespa.defaults.Defaults.getDefaults;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


zookeeper-server/zookeeper-server-3.5/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperServerImpl.java
```diff
@@ -76,6 +79,41 @@ private String transformConfigToString(ZookeeperServerConfig config) {
...
+
+    private String createTlsQuorumConfig(ZookeeperServerConfig config) {
+        StringBuilder sb = new StringBuilder();
+
+        // Common config
+        sb.append("ssl.quorum.hostnameVerification=false\n");
+        sb.append("ssl.quorum.clientAuth=NEED\n");
+        sb.append("ssl.quorum.ciphersuites=").append(String.join(",", new TreeSet<>(TlsContext.ALLOWED_CIPHER_SUITES))).append("\n");
+        sb.append("ssl.quorum.enabledProtocols=").append(String.join(",", new TreeSet<>(TlsContext.ALLOWED_PROTOCOLS))).append("\n");
+        sb.append("ssl.quorum.protocol=TLSv1.2\n");
```
bjorncs: This should set to 'TLS'. It's effectively setting max TLS version, disabling TLSv1.3 even when it's listed in `enableProtocols`. See table in https://bugs.openjdk.java.net/browse/JDK-8202625.

bjorncs (approved):

zookeeper-server/zookeeper-server-3.5/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperServerImpl.java
hmusum: ZooKeeper doc just says that 'TLSv1.2' is default value, but not anything about other possible values. But this makes sense, thanks, will fix

bjorncs (approved):

PR #11303, opened 2019-11-14
Reduce timeouts for non-suspend ConfigServer REST API calls from node admin

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/ConfigServerApiImpl.java
```diff
@@ -202,24 +209,34 @@ private static CloseableHttpClient createClient(SSLConnectionSocketFactory socke
...
                 .build();
     }
 
+    private static void setRequestConfigOverride(Params paramsOrNull, HttpRequestBase request) {
+        if (paramsOrNull == null) return;
+
+        RequestConfig.Builder builder = RequestConfig.copy(request.getConfig());
+
+        paramsOrNull.getConnectionTimeout().ifPresent(connectionTimeout -> {
+            builder.setConnectTimeout((int) connectionTimeout.getSeconds());
```
freva: Shouldn't this be millis?

hakonhall:
Decided against increasing the timeout for suspend, which might hide the problem observed on controller-cd-2 on November 12. Instead, this PR allows tuning the timeouts of suspend independent of timeout used for other ConfigServer REST calls. And reduces current non-suspend from 15s to 10s (and getting connection from the 200-connection pool to 1s).

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/ConfigServerApi.java
```diff
@@ -9,13 +10,26 @@
...
+        public Optional<Duration> getConnectionTimeout() { return connectionTimeout; }
+    }
 
     <T> T get(String path, Class<T> wantedReturnType);
 
     <T> T post(String path, Object bodyJsonPojo, Class<T> wantedReturnType);
 
     <T> T put(String path, Optional<Object> bodyJsonPojo, Class<T> wantedReturnType);
 
+    <T> T put(Params params, String path, Optional<Object> bodyJsonPojo, Class<T> wantedReturnType);
```
freva: Consider putting the optional argument at the end

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/ConfigServerApiImpl.java
hakonhall: Great catch - I just verified this but forgot to update the code...

freva (approved):
:100:

PR #11305, opened 2019-11-14
Stop services for nodes in parked state

hmusum:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
@@ -249,6 +249,11 @@ public class NodeAgentImpl implements NodeAgent {
         }
     }
 
+    private void stopServicesIfNeeded(NodeAgentContext context) {
+        if (hasStartedServices)
+            stopServices(context);
+    }
+
     private void stopServices(NodeAgentContext context) {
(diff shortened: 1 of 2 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -402,12 +402,15 @@ void doConverge(NodeAgentContext context) {
...
             case reserved:
-            case parked:
             case failed:
             case inactive:
                 removeContainerIfNeededUpdateContainerState(context, container);
                 updateNodeRepoWithCurrentAttributes(context);
                 break;
+            case parked:
+                updateNodeRepoWithCurrentAttributes(context);
+                stopServices(context);
```
hakonhall: This will log "Stopping services" on every tick.

hmusum: Yeah, good point, will add a stopServicesIfNeeded() and use that instead

hakonhall (approved):

hakonhall:
We should probably do the same with inactive and failed?

freva:
I think the criteria should be if the node no longer has an `owner`. This would also allow operator to manually park/fail a node (which will not clear the `owner`) without host-admin doing anything to the container.

hmusum:
Sounds good to me, I'll create a PR implementing this

PR #11316, opened 2019-11-15
Handle single value string buckets correctly

bratseth:
The new tensor unit test at the bottom is unrelated.


bratseth:
Ping @baldersheim

baldersheim (approved):

PR #11320, opened 2019-11-15
add skeleton for NearestNeighborItem

arnej27959:

this is a preview with unfinished code.  some questions I have are:
 - if targetNumHits should have no default, it should probably be a 3rd argument to the YQL function, instead of an annotation.  For now it has a default to avoid making the annotation mandatory.
- how should buildNearestNeighbor get the arguments from the AST? I used the "fetchFieldRead" method but I'm not sure why that works, and if we want an integer 3rd argument I have no idea how to do that
- when serializing to binary format, should we use IntegerCompressor for targetNumHits instead of just buffer.putInt()?
- missing: validating that the "query property" string actually refers to a property containing a tensor of the correct type
- missing: how to actually pass the query property to the backend?

@bratseth please review
@geirst FYI
@baldersheim feel free to take a look

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/query/NearestNeighborItem.java
+++ b/container-search/src/main/java/com/yahoo/prelude/query/NearestNeighborItem.java
@@ -1,0 +1,56 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+package com.yahoo.prelude.query;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
+ * @author arnej
+ */
+public class NearestNeighborItem extends SimpleTaggableItem {
+
+    private int targetNumber = 1000;
+    private String field;
+    private String property;
+
+    public int getTargetNumHits() { return targetNumber; }
+    public void setTargetNumHits(int target) { this.targetNumber = target; }
+
+    public String getQueryProperty() { return property; }
+
+    public NearestNeighborItem(String fieldName, String queryProperty) {
+        this.field = fieldName;
+        this.property = queryProperty;
+    }
+
+    public String getIndexName() { return field; }
+
+    @Override
+    public void setIndexName(String index) { this.field = index; }
+
+    @Override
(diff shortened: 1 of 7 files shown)


bratseth:
- targetNumHits: We want to make it optional later so it should probably remain an annotation. It can have a default (say, -1) and then fail with an IllegalArgumentException at some point if not set.

- fetchFieldread is correct, but I don't understand how that stuff works either. It don't get the feeling we're supposed to be able to find out how it works, and I want to rewrite the whole thing in JavaCC tbh.

- IntegerCompressor: I don't think it matters much either way ...

container-search/src/main/java/com/yahoo/prelude/query/NearestNeighborItem.java
```diff
@@ -0,0 +1,56 @@
...
+
+package com.yahoo.prelude.query;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
+ * @author arnej
+ */
+public class NearestNeighborItem extends SimpleTaggableItem {
```
bratseth: @Beta

container-search/src/main/java/com/yahoo/prelude/query/NearestNeighborItem.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private int targetNumber = 1000;
+    private String field;
+    private String property;
+
+    public int getTargetNumHits() { return targetNumber; }
```
bratseth: This one needs javadoc

container-search/src/main/java/com/yahoo/prelude/query/NearestNeighborItem.java
```diff
@@ -0,0 +1,56 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+package com.yahoo.prelude.query;
+
+import java.nio.ByteBuffer;
+
+/**
+ *
```
bratseth: Add doc

bratseth (approved):
LGTM!

baldersheim (approved):

PR #11329, opened 2019-11-18
Retry deleting deployments when jobs end

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
@@ -35,8 +35,6 @@ import com.yahoo.vespa.hosted.controller.api.integration.deployment.TesterId;
 import com.yahoo.vespa.hosted.controller.api.integration.organization.DeploymentFailureMails;
 import com.yahoo.vespa.hosted.controller.application.ApplicationPackage;
 import com.yahoo.vespa.hosted.controller.application.Deployment;
-import com.yahoo.vespa.hosted.controller.application.DeploymentJobs;
-import com.yahoo.vespa.hosted.controller.application.DeploymentJobs.JobReport;
 import com.yahoo.vespa.hosted.controller.application.TenantAndApplicationId;
 import com.yahoo.yolean.Exceptions;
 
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -534,9 +532,11 @@ private boolean servicesConverged(ApplicationId id, JobType type, Version platfo
 
     private Optional<RunStatus> deactivateReal(RunId id, DualLogger logger) {
         try {
-            logger.log("Deactivating deployment of " + id.application() + " in " + id.type().zone(controller.system()) + " ...");
-            controller.applications().deactivate(id.application(), id.type().zone(controller.system()));
-            return Optional.of(running);
+            return retrying(10, () -> {
```
mpolden: Isn't retrying handled by the `ConfigServer` implementation?

jonmv: Not for, e.g., `APPLICATION_LOCK_TIMEOUT`, which frequently happens for certain applications.

mpolden: Ok, then add that case? `ConfigServerClient` is the place we have put other retrying logic.

jonmv: The deploy and deactivate implementations in the config server client are only retries over the three servers, as they need to hit the correct one. There is no other retrying logic there, for those paths.

mpolden (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
jonmv: I want to let the whole step (deleteTester, deleteApplication) fail and be retried, but that requires some additional data stored for the steps in the job machinery. This is preferable because it allows the job to retry on a different controller, during upgrade or failure scenarios. I will get to it later, when I implement the delay for production tests, I think.

jonmv: (They initially retried until they succeeded, but sometimes the config server was just wrong, and refused to give an OK response, so the job would run for ever, and it was changed to be a single attempt instead.)

PR #11330, opened 2019-11-18
Bratseth/smarter rebalancing

bratseth:
Please look at each commit and read the commit messages.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
@@ -1,9 +1,17 @@
 // Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.maintenance;
 
+import com.yahoo.config.provision.ApplicationId;
+import com.yahoo.config.provision.ApplicationLockException;
+import com.yahoo.config.provision.Deployer;
+import com.yahoo.config.provision.Deployment;
+import com.yahoo.config.provision.HostSpec;
 import com.yahoo.config.provision.NodeResources;
 import com.yahoo.config.provision.NodeType;
+import com.yahoo.config.provision.TransientException;
 import com.yahoo.jdisc.Metric;
+import com.yahoo.jdisc.core.Main;
+import com.yahoo.log.LogLevel;
 import com.yahoo.transaction.Mutex;
 import com.yahoo.vespa.hosted.provision.Node;
 import com.yahoo.vespa.hosted.provision.NodeList;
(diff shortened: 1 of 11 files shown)


freva (approved):
LGTM, but we should keep the lock for the entire duration of mark wantToRetire, prepare, activate(, unmark wantToRetire)

bratseth:
It's a bit more ugly but ok, done. PTAL

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -90,20 +107,50 @@ private Move findBestMove(NodeList allNodes) {
...
-    private void markWantToRetire(Node node) {
+    private boolean markWantToRetire(Node node, boolean wantToRetire) {
         try (Mutex lock = nodeRepository().lock(node)) {
             Optional<Node> nodeToMove = nodeRepository().getNode(node.hostname());
-            if (nodeToMove.isEmpty()) return;
-            if (nodeToMove.get().state() != Node.State.active) return;
+            if (nodeToMove.isEmpty()) return false;
+            if (nodeToMove.get().state() != Node.State.active) return false;
+
+            if (node.status().wantToRetire() != wantToRetire)
```
freva: If they are equal, we should return `false` or at least make sure not to "undo" this in `finally` if we fail to achieve the wanted result.

bratseth: If it's already set to want to retire it's fine that we do it.
However, as you point out it is not fine to undo it if somebody else sets it and we failed (regardless of when it happens).

PTAL - I'm going to run out of "try" clauses soon.

bratseth: PTAL

freva (approved):
:100:

bratseth:
Hello Merge Stop Enforcer

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -90,20 +107,52 @@ private Move findBestMove(NodeList allNodes) {
...
+        try (MaintenanceDeployment deployment = new MaintenanceDeployment(application, deployer, nodeRepository())) {
+            if ( ! deployment.isValid()) return false;
+
+            boolean couldMarkRetiredNow = markWantToRetire(move.node, true);
+            if ( ! couldMarkRetiredNow) return false;
+
+            try {
+                if ( ! deployment.prepare()) return false;
+                if (nodeRepository().getNodes(application, Node.State.reserved).stream().noneMatch(node -> node.hasParent(move.toHost.hostname())))
+                    return false; // Deployment is not moving the from node to the target we identified for some reason
```
freva: This leaves the reserved node with the application for 20-27 mins (until it is expired by `ReservationExpirer`), this is especially unfortunate in CD when we want to fail integration tests when some node are stuck in a non-active state.

We should prepare the application again with the node unmarked as wantToRetire or move it directly to dirty if that is safe.

FYI: @hmusum

bratseth: The latter should be fine here. After lunch ...

bratseth: https://github.com/vespa-engine/vespa/pull/11355

PR #11333, opened 2019-11-18
More descriptive error message, and use empty rather than EPOCH for time

jonmv:
@mpolden please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackage.java
```diff
@@ -106,6 +108,16 @@ public ApplicationPackage withTrustedCertificate(X509Certificate certificate) {
         return trustedCertificates;
     }
 
+    private static <Type> Optional<Type> parse(Inspector buildMetaObject, String fieldName, Function<Inspector, Type> mapper) {
```
mpolden: Nit: Capital letters is an established convention for generic types. `Type` reads as a concrete class name, especially when coupled with a concrete class like the `mapper` type here. Prefer `T` or `TYPE`.

mpolden (approved):

PR #11345, opened 2019-11-19
Throw for invalid state

mpolden:


jonmv (commented):
Hmm... Loadbalancers can be from a previous instance, I suppose, so this will crash. Perhaps this can be fixed by swapping lines 81 and 82?

PR #11347, opened 2019-11-19
Skip tester owned nodes when considering best moves

jonmv:
@bratseth or @hmusum please review.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
@@ -99,6 +99,7 @@ public class Rebalancer extends Maintainer {
         Move bestMove = Move.none;
         for (Node node : allNodes.nodeType(NodeType.tenant).state(Node.State.active)) {
             if (node.parentHostname().isEmpty()) continue;
+            if (node.allocation().map(allocation -> allocation.owner().instance().isTester()).orElse(false)) continue;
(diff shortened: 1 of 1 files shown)


hmusum (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -99,6 +99,7 @@ private Move findBestMove(NodeList allNodes) {
         Move bestMove = Move.none;
         for (Node node : allNodes.nodeType(NodeType.tenant).state(Node.State.active)) {
             if (node.parentHostname().isEmpty()) continue;
+            if (node.allocation().map(allocation -> allocation.owner().instance().isTester()).orElse(false)) continue;
```
bratseth: Since this is just active nodes you can just say
if (node.allocation().get().owner().instance().isTester())) continue;

bratseth (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
jonmv: Thanks.

jonmv:
```suggestion
            if (node.allocation().get().owner().instance().isTester())) continue;
```

PR #11351, opened 2019-11-19
use chrono instead of fastos time

havardpe:

@baldersheim please review

### DIFF
--- a/fnet/src/vespa/fnet/packetqueue.cpp
+++ b/fnet/src/vespa/fnet/packetqueue.cpp
@@ -2,7 +2,6 @@
 
 #include "packetqueue.h"
 #include "packet.h"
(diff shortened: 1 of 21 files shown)


fnet/src/vespa/fnet/packetqueue.cpp
```diff
@@ -226,19 +225,20 @@ FNET_PacketQueue::DequeuePacket(FNET_Context *context)
 FNET_Packet*
 FNET_PacketQueue::DequeuePacket(uint32_t maxwait, FNET_Context *context)
```
baldersheim: Future improvement by using milliseconds as param type.

baldersheim (approved):

PR #11355, opened 2019-11-19
Fast-track cleanup after failed rebalancing

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -139,20 +140,29 @@ private boolean deployTo(Move move) {
...
+                                        .findAny();
+                if (expectedNewNode.isEmpty()) return false;
                 if ( ! deployment.activate()) return false;
 
                 log.info("Rebalancer redeployed " + application + " to " + move);
                 return true;
             }
             finally {
-                markWantToRetire(move.node, false); // Necessary if this failed, no-op otherwise
+                markWantToRetire(nodeRepository().getNode(move.node.hostname()), false); // Necessary if this failed, no-op otherwise
```
freva:
```suggestion
                markWantToRetire(move.node, false); // Necessary if this failed, no-op otherwise
```
`markWantToRetire()` will get the node after taking the lock anyway...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -139,20 +140,29 @@ private boolean deployTo(Move move) {
...
+                if (expectedNewNode.isPresent()) { // Immediately clean up if we reserved the node but could not activate
+                     Optional<Node> reservedNewNode = nodeRepository().getNode(expectedNewNode.get().hostname(), Node.State.reserved);
```
freva: Why do we need to get it again? We are still inside the lock

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
bratseth: No, we're still inside the outer try block.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
bratseth: Yes, but the deployment action can have changed it.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
freva: Yes, isn't that the one we care about, since it has `MaintenanceDeployment` which has the lock?

bratseth: Ah, I read "after taking the lock a**w**ay.

bratseth:
Anything else @freva ?

freva (approved):

PR #11356, opened 2019-11-19
Jvenstad/handle dev deployments without spec in deployment view

jonmv:
@mpolden please review this patchwork. Will be replaced pretty soon.


jonmv:
Or @oyving if you're impatient.

mpolden (approved):

PR #11362, opened 2019-11-20
Add sudo support to SystemCtl

mpolden:


mpolden:
or @tokle

tokle (approved):

hakonhall:
Most operations host-admin do requires root privileges. Why cannot host-admin just run as root?

mpolden:
It does. In this case we're using `SystemCtl` outside host-admin to restart host-admin itself, without having to run as root.

PR #11366, opened 2019-11-20
Deprecate useLocalNode

bratseth:
The container will figure this out on its own,
and turning this on will make it impossible to query this
cluster from another node.


baldersheim (commented):
We need to discuss this one.

baldersheim (approved):

PR #11372, opened 2019-11-20
Remove cipher suite not supported by Java

hmusum:
TLS_CHACHA20_POLY1305_SHA256 is not supported by Java, see https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/sun/security/ssl/CipherSuite.java#L483

@bjorncs FYI


security-utils/src/main/java/com/yahoo/security/tls/TlsContext.java
```diff
@@ -13,6 +13,7 @@
  */
 public interface TlsContext extends AutoCloseable {
 
+    // TODO: Where does this set come from?
```
vekterli: See https://github.com/vespa-engine/vespa/blob/master/vespalib/src/vespa/vespalib/net/tls/impl/iana_cipher_map.cpp#L16-L20

freva (approved):
So do need to update `iana_cipher_map.cpp` too then?

vekterli:
I don't think thats necessary since ciphers will be negotiated for each connection between any C++ and Java components

security-utils/src/main/java/com/yahoo/security/tls/TlsContext.java
hmusum: Thanks, will resolve TODO in  another PR

PR #11374, opened 2019-11-21
Balder/milliseconds in config rebased 1

baldersheim:
@havardpe PR


havardpe (approved):
Going forward we should strive to make duration calculations less dependent on time resolution.

PR #11375, opened 2019-11-21
Remove more cipher suites not supported by Java 11 from set configure…

hmusum:
…d for use by ZooKeeper

@bjorncs @vekterli FYI regarding the TODO I added in TlsContext, I am not sure if what I wrote there makes sense


tokle (approved):

zookeeper-server/zookeeper-server-3.5/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperServerImpl.java
```diff
@@ -165,8 +165,11 @@ private String createTlsQuorumConfig(ZookeeperServerConfig config, Optional<Tran
 
     private TreeSet<String> getCipherSuites() {
         Set<String> cipherSuites = new HashSet<>(TlsContext.ALLOWED_CIPHER_SUITES);
-        // Remove cipher suite not supported by Java
+        // Remove cipher suites not supported by Java 11
```
bjorncs: Another approach is to query Java's `SSLContext` for the ciphers supported by the current VM. You may then get rid of the hardcoded blacklist. See https://github.com/vespa-engine/vespa/blob/master/security-utils/src/main/java/com/yahoo/security/tls/DefaultTlsContext.java#L57 for inspiration.

PR #11378, opened 2019-11-21
Reduce the number of different ways to get the time.

baldersheim:
@havardpe @vekterli @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/util/statebuf.cpp
+++ b/searchlib/src/vespa/searchlib/util/statebuf.cpp
@@ -3,6 +3,7 @@
 #include "statebuf.h"
 
 #include <vespa/log/log.h>
+
 LOG_SETUP(".searchlib.util.statebuf");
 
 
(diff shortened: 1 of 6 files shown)


searchlib/src/vespa/searchlib/util/statebuf.cpp
```diff
@@ -168,35 +145,28 @@ StateBuf::appendHex(unsigned long val) noexcept
...
-    /*
-     * clock_gettime() is supposed to be async signal safe.
-     * gettimeofday() is not documented to be async signal safe.
-     */
-    int gtres = clock_gettime(CLOCK_REALTIME, &ts);
-    if (gtres != 0) {
-        LOG_ABORT("should not be reached");
-    }
-    appendTimestamp(ts);
+    appendTimestamp(std::chrono::system_clock::now().time_since_epoch());
```
toregge: Is this method async signal safe ?

baldersheim: It uses clock_gettime(CLOCK_REALTIME) under the hood if available.

havardpe (commented):
👍

PR #11381, opened 2019-11-21
Defer GC bucket info merge until all responses have been received

### LINKED ISSUE #11373: Background document GC does not update bucket DB atomically, causing false positives for merge pending metrics
Background document GC is implemented in terms of `RemoveSelection` ops sent in parallel to bucket replicas. These are low priority and may be stuck in persistence queues for a while, depending on the load on the replica nodes.

Today, when a reply is received on the distributor, the GC maintenance operation will immediately update the bucket DB with the information returned as part of the reply. If a periodic DB metric updating scan happens to see this bucket in a state where not all replies have been received, it will likely be counted as a bucket requiring a merge. This causes confusing merge pending metric noise in otherwise quiescent systems.

GC ops (and other ops, for that matter) should preferably update the bucket DB as one atomic operation when all replies have been received. This needs some extra care to make sure doing so does not overwrite any more recent bucket DB updates received as part of concurrent feed load.

vekterli:
@geirst please review

Avoids causing false positives in the merge pending metrics due to
partial (and mismatching) bucket info being merged into the DB one
by one as responses are received. Instead, wait until all responses
have been received and merge as one atomic operation.

This fixes #11373

### DIFF
--- a/storage/src/vespa/storage/distributor/operations/idealstate/garbagecollectionoperation.h
+++ b/storage/src/vespa/storage/distributor/operations/idealstate/garbagecollectionoperation.h
@@ -2,7 +2,9 @@
 #pragma once
 
 #include "idealstateoperation.h"
+#include <vespa/storage/bucketdb/bucketcopy.h>
 #include <vespa/storage/distributor/messagetracker.h>
+#include <vector>
 
 namespace storage::distributor {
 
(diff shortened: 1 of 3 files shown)


storage/src/vespa/storage/distributor/operations/idealstate/garbagecollectionoperation.h
```diff
@@ -22,6 +24,10 @@ class GarbageCollectionOperation : public IdealStateOperation
 
 protected:
     MessageTracker _tracker;
+private:
+    std::vector<BucketCopy> _replica_info;
```
geirst: Instantiate after _tracker in constructor?

geirst (approved):
👍  A nitpick to consider

PR #11384, opened 2019-11-22
don't deploy to us-east-1 when user specifies aws-us-west-2a

andreer:


jonmv (approved):
Tsk, tsk, such picky users.

PR #11395, opened 2019-11-25
Base64 support for fbench

ohad7:

Added a -d option to fbench to allow base64 decode before sending the request body. This is useful for cases where the POST body is binary e.g protobuf and we want to avoid the base64 decode on the client

### DIFF
--- a/fbench/src/fbench/client.cpp
+++ b/fbench/src/fbench/client.cpp
@@ -226,7 +226,7 @@ Client::run()
             int cLen = _args->_usePostMode ? urlSource.nextContent() : 0;
             auto content = urlSource.content();
             if (_args->_usePostMode && _args->_base64Decode) {
-                auto base64_decoded = Base64::decode(std::string(urlSource.content(), cLen));
+                auto base64_decoded = Base64::decode(std::string(content, cLen));
                 content = base64_decoded.c_str();
                 cLen = base64_decoded.size();
             }
--- a/fbench/src/fbench/client.cpp
+++ b/fbench/src/fbench/client.cpp
@@ -273,7 +273,7 @@ Client::run()
                 auto base64_content = std::string(urlSource.content(), cLen);
                 auto decoded = base64_decode(base64_content);
                 content = decoded.c_str();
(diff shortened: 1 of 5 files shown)


fbench/src/fbench/client.cpp
```diff
@@ -220,8 +224,15 @@ Client::run()
                 strcat(_linebuf, _args->_queryStringToAppend.c_str());
             }
             int cLen = _args->_usePostMode ? urlSource.nextContent() : 0;
+            auto content = urlSource.content();
```
baldersheim: Specify type explicit as it is not obvious what it is. You are also reassigning it further down.

fbench/src/fbench/client.cpp
```diff
@@ -220,8 +224,15 @@ Client::run()
...
+            if (_args->_usePostMode && _args->_base64Decode) {
+                auto base64_decoded = Base64::decode(std::string(content, cLen));
+                content = base64_decoded.c_str();
```
baldersheim: Here you have a lifetime issue. **base64_decoded** must live at least as long as **content**.

fbench/src/fbench/client.cpp
```diff
@@ -220,8 +224,15 @@ Client::run()
                 strcat(_linebuf, _args->_queryStringToAppend.c_str());
             }
             int cLen = _args->_usePostMode ? urlSource.nextContent() : 0;
+            auto content = urlSource.content();
+            if (_args->_usePostMode && _args->_base64Decode) {
+                auto base64_decoded = Base64::decode(std::string(content, cLen));
```
baldersheim: You should not wrap the input as an std::string. That will make a copy of the content. Just use the the decode(const char *, int length) method.

baldersheim (changes requested):

fbench/src/fbench/client.cpp
ohad7: Java developer negligence... fixed your other comments as comments as well.

baldersheim (approved):

PR #11396, opened 2019-11-25
Allow tenant containers to run on dev hosts

tokle:


config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
```diff
@@ -71,4 +71,11 @@ public NodeType childNodeType() {
         return childNodeTypes;
     }
 
+    /**
+     * @param type Child {@link NodeType}
+     * @return true if the {@link NodeType} can run on this host, false otherwise.
+     */
+    public boolean supportsChild(NodeType type) {
```
mpolden: How about naming this `canRun`? `host.type().canRun(NodeType.tenant)` reads better I think.

config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
```diff
@@ -71,4 +71,11 @@ public NodeType childNodeType() {
...
+        return isDockerHost() && childNodeTypes.contains(type);
```
mpolden: Can be simplified to just `childNodeTypes.contains(type)` as `isDockerHost` only checks the emptiness of `childNodeTypes`.

mpolden:
ABI spec must be updated, but otherwise LGTM.

mpolden (approved):

PR #11397, opened 2019-11-25
Catch exceptions setting up config source, and exit

jonmv:
@baldersheim please review and merge.

### DIFF
--- a/configutil/src/apps/modelinspect/main.cpp
+++ b/configutil/src/apps/modelinspect/main.cpp
@@ -83,7 +83,7 @@ Application::getConfigUri()
         return config::ConfigUri::createFromSpec(_cfgId, config::ServerSpec(_specString));
     }
     catch (std::exception &e) {
-        LOG(fatal, "Failed to set up model configuration source, will exit: %s", e.what());
+        std::cerr << "FATAL ERROR: failed to set up model configuration: " << e.what() << "\n";
         exit(1);
     }
 }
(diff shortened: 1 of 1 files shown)


configutil/src/apps/modelinspect/main.cpp
```diff
@@ -76,8 +76,20 @@ Application::getSources(void)
...
 }
 
+config::ConfigUri
+Application::getConfigUri()
+{
+    try {
+        return config::ConfigUri::createFromSpec(_cfgId, config::ServerSpec(_specString));
+    }
+    catch (std::exception &e) {
+        std::cerr << "FATAL ERROR: failed to set up model configuration source.\n";
```
jonmv: Argh, forgot `e.what()` here.

jonmv: There.

configutil/src/apps/modelinspect/main.cpp
```diff
@@ -76,8 +76,20 @@ Application::getSources(void)
...
 }
 
+config::ConfigUri
+Application::getConfigUri()
+{
+    try {
+        return config::ConfigUri::createFromSpec(_cfgId, config::ServerSpec(_specString));
+    }
+    catch (std::exception &e) {
+        std::cerr << "FATAL ERROR: failed to set up model configuration: " << e.what() << "\n";
```
baldersheim: use LOG(FATAL,....

baldersheim (changes requested):
Fix logging so we also get it in the vespalog in startard format if run with redirection

configutil/src/apps/modelinspect/main.cpp
jonmv:
```suggestion
        LOG(fatal, "Failed to set up model configuration source, will exit: %s", e.what());
```

baldersheim (approved):

PR #11400, opened 2019-11-25
Balder/sanitize the clock 2

baldersheim:
@havardpe PR

### DIFF
--- a/staging_vespalib/src/tests/clock/clock_benchmark.cpp
+++ b/staging_vespalib/src/tests/clock/clock_benchmark.cpp
@@ -3,6 +3,7 @@
 #include <vespa/vespalib/util/clock.h>
 #include <cassert>
 #include <vector>
+#include <atomic>
 
 using vespalib::Clock;
 using fastos::TimeStamp;
@@ -17,9 +18,13 @@ struct NSValue : public UpdateClock {
     int64_t _value;
 };
 
-struct NSVolatileValue : public UpdateClock {
+struct NSVolatile : public UpdateClock {
     void update() override { _value = std::chrono::steady_clock::now().time_since_epoch().count(); }
     volatile int64_t _value;
+};
+struct NSAtomic : public UpdateClock {
+    void update() override { _value.store(std::chrono::steady_clock::now().time_since_epoch().count()); }
+    std::atomic<int64_t> _value;
 };
 
 class TestClock : public FastOS_Runnable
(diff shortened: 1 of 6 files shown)


havardpe (commented):
1 consider using vespalib::BenchmarkTimer
2 consider testing a cheaper store than sequential consistency
3 the test seems a bit on the complex side...

staging_vespalib/src/tests/clock/clock_benchmark.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+struct NSValue : public UpdateClock {
+    void update() override { _value = std::chrono::steady_clock::now().time_since_epoch().count(); }
+    int64_t _value;
+};
+
+struct NSVolatile : public UpdateClock {
+    void update() override { _value = std::chrono::steady_clock::now().time_since_epoch().count(); }
+    volatile int64_t _value;
+};
```
vekterli: We should never want to use `volatile` over `atomic`, so I'm not sure if we need this? It's not well defined for concurrent access, and it's theoretically highly compiler dependent what we actually get out.

staging_vespalib/src/tests/clock/clock_benchmark.cpp
```diff
@@ -0,0 +1,171 @@
...
+struct NSAtomic : public UpdateClock {
+    void update() override { _value.store(std::chrono::steady_clock::now().time_since_epoch().count()); }
```
vekterli: As @havardpe mentioned, this will incur an artificially high performance impact since a plain `atomic::store` incurs a full memory store/load fence in all cases. Should probably make memory order for the store operation parameterized to match that of the load in the test body (i.e. `seq_cst` -> `seq_cst`, `release` -> `acquire` etc)

staging_vespalib/src/tests/clock/clock_benchmark.cpp
baldersheim: I just wanted it here top compare performance and generated code.

staging_vespalib/src/tests/clock/clock_benchmark.cpp
baldersheim: I agree, but that is not a dimension that is interesting to test here. So I just wanted too keep it simple.
I am only interested in read performance, not write. Igf you are updating with frequnce of 1Mhz or higher it would matter, but not when using 100hz-100khz.

PR #11401, opened 2019-11-25
Need to use a volatile to ensure compiler will always generate correc…

baldersheim:
…t code independent of inlining and optimizations.

@havardpe PR

### DIFF
--- a/staging_vespalib/src/vespa/vespalib/util/clock.h
+++ b/staging_vespalib/src/vespa/vespalib/util/clock.h
@@ -20,12 +20,12 @@ private:
     Clock(const Clock &);
     Clock & operator = (const Clock &);
 
-    mutable volatile int64_t  _timeNS;
-    int                       _timePeriodMS;
-    std::mutex                _lock;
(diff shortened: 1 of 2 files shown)


staging_vespalib/src/vespa/vespalib/util/clock.h
```diff
@@ -20,7 +20,7 @@ class Clock : public FastOS_Runnable
     Clock(const Clock &);
     Clock & operator = (const Clock &);
 
-    mutable fastos::SteadyTimeStamp _timeNS;
+    mutable volatile int64_t  _timeNS;
```
havardpe: actually, to ensure it is visible to other threads, it needs to be atomic

havardpe (commented):
I think volatile is the wrong solution here

staging_vespalib/src/vespa/vespalib/util/clock.h
baldersheim: Fixed, memory_order_relaxed.

PR #11406, opened 2019-11-25
Use resources in node response

freva:
* Updates controller's and node-admin's node-repository clients to use `resources` object.
* Removes usage of `canonicalFlavor` (is equal to `flavor`) and `description` (removed a while ago) from controller's node-repository client.
* Adds `diskSpeed` and `storageType` to `/application/v4/` nodes response.

Must be merged with the corresponding PR in internal repo.


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
```diff
@@ -210,4 +202,178 @@ public int hashCode() {
         unknown
     }
 
+    public static class Builder {
```
mpolden: Great!

mpolden (approved):

PR #11407, opened 2019-11-25
Jvenstad/move user tenants to /login

jonmv:
@freva please review.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/athenz/ZmsClientMock.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/athenz/ZmsClientMock.java
@@ -74,6 +74,13 @@ public class ZmsClientMock implements ZmsClient {
     }
 
     @Override
+    public void addRoleMember(AthenzIdentity member, AthenzRole role) {
+        if ( ! role.roleName().equals("tenancy.vespa.hosting.admin"))
+            throw new IllegalArgumentException("Mock only supports adding tenant admins, not " + role.roleName());
+        getDomainOrThrow(role.domain(), true).tenantAdmin(member);
+    }
+
+    @Override
     public boolean getMembership(AthenzRole role, AthenzIdentity identity) {
         if (role.roleName().equals("admin")) {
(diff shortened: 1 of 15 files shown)


jonmv:
I'll move it to /login. Need a separate handler, though.

freva (approved):

freva:
```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.123 s <<< FAILURE! - in com.yahoo.vespa.hosted.controller.api.role.PathGroupTest
[ERROR] uniqueMatches(com.yahoo.vespa.hosted.controller.api.role.PathGroupTest)  Time elapsed: 0.073 s  <<< FAILURE!
java.lang.AssertionError: Paths '/athenz/v1/user' and '/athenz/v1/{*}' overlap.
	at com.yahoo.vespa.hosted.controller.api.role.PathGroupTest.uniqueMatches(PathGroupTest.java:48)
```

jonmv:
Hehe, _that_ test?! Allright :p

jonmv:
Or ... the console has to have some separate parts for on-prem and public, anyway, and I think this stuff really belongs in the "Athenz API handler". How easy is it to let the console use different paths for these resources? The signup won't even exist in public, nor will the Athenz domain list and the property list which is already part of this handler.

freva:
> How easy is it to let the console use different paths for these resources?

Assuming it's only the path that is the issue (we keep the JSON format unchanged), it shouldn't be too hard. The code that gets it is here:
https://git.ouroath.com/vespa/vaas-console/blob/075971715f788c644e9b72b9709e4c93a2487df6/src/App/shared/auth/authStorage.js#L19

Additionally, it should return some non-200 code if the user is not logged in, at which point this method will be run:
https://git.ouroath.com/vespa/vaas-console/blob/075971715f788c644e9b72b9709e4c93a2487df6/src/App/shared/auth/AuthProvider.js#L194

jonmv:
Ok, good. If the user does not provide valid credentials, they will receive a 403 from the filter before ever reaching the handler, so that should be OK.

freva (approved):

freva:
Done in #11465 instead

PR #11421, opened 2019-11-27
Fail when building a keystore with expired certificates

tokle:


bjorncs (approved):

tokle:
there are cases where a keystore / truststore may contain expired certs.

PR #11424, opened 2019-11-27
Include application id in flag definition for use-old-metrics-checks

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -164,7 +164,7 @@
             "use-old-metrics-checks", true,
             "Whether to use old metrics checks",
             "Takes effect on restart of Docker container",
```
olaaun:
```suggestion
             "Takes effect on next host admin tick",
```

olaaun (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hmusum: Yeah, looks like we want this on every tick

PR #11425, opened 2019-11-27
Reduce code visibility and minor cleanup.

baldersheim:
@havardpe or @toregge PR
Pure code health. No intended semantic changes.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/matchview.h
+++ b/searchcore/src/vespa/searchcore/proton/server/matchview.h
@@ -7,7 +7,7 @@
 #include <vespa/searchcore/proton/common/docid_limit.h>
 #include <vespa/searchcore/proton/documentmetastore/documentmetastorecontext.h>
 #include <vespa/searchcore/proton/matching/match_context.h>
-#include <vespa/searchcore/proton/matching/matcher.h>
+#include <vespa/searchcore/proton/summaryengine/isearchhandler.h>
 #include <vespa/searchcorespi/index/indexsearchable.h>
 #include <vespa/searchlib/attribute/attributevector.h>
 
@@ -15,7 +15,9 @@ namespace proton {
 
 namespace matching {
     class SessionManager;
+    class Matcher;
 }
+
 class MatchView {
     using SessionManagerSP = std::shared_ptr<matching::SessionManager>;
     Matchers::SP                         _matchers;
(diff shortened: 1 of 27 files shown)


searchcore/src/vespa/searchcore/proton/server/matchview.h
```diff
@@ -60,7 +62,7 @@ class MatchView {
     matching::MatchContext::UP createContext() const;
 
     std::unique_ptr<search::engine::SearchReply>
-    match(const ISearchHandler::SP &searchHandler,
+    match(const std::shared_ptr<ISearchHandler> &searchHandler,
```
havardpe: prefer passing shared pointers by value

toregge (approved):

searchcore/src/vespa/searchcore/proton/server/searchabledocsubdb.cpp
```diff
@@ -91,19 +93,19 @@ createIndexManagerInitializer(const DocumentDBConfig &configSnapshot, SerialNum
                               const IndexConfig &indexCfg,
                               std::shared_ptr<searchcorespi::IIndexManager::SP> indexManager) const
 {
-    Schema::SP schema(configSnapshot.getSchemaSP());
+    const Schema::SP & schema(configSnapshot.getSchemaSP());
```
havardpe: please decide if you want a shared pointer (use value) or if you just want to inspect the inner state of an object (use const reference to actual object directly).

Also, the shared pointer itself is not an inherently shareable object.

searchcore/src/vespa/searchcore/proton/server/matchview.h
baldersheim: I normally do, but the usage pattern prevented the a simple change.
handler->match(handler, *searchRequest, *threadBundle)

searchcore/src/vespa/searchcore/proton/server/searchabledocsubdb.cpp
baldersheim: Only reference was needed. Fixed.

searchcore/src/vespa/searchcore/proton/server/matchview.h
baldersheim: enable-shared-from-this is the better option. Will do later on.

PR #11434, opened 2019-11-27
Always include diskSpeed and storageType in resources object

freva:
I've been migrating all our clients to use the `resources` object instead of the old fields, in java code I could use the `DiskSpeed::getDefault` or `StorageType::getDefault` methods if the value was not set in the node response.

This does is not possible and is quite ugly when dealing with some of our other clients, f.ex. the javascript client in console & dashboard. I think we should just always set these, and as soon as the moratorium is over, I already have a PR ready to remove all the old fields which will simplify the response quite a bit.


bratseth (approved):
👍

bratseth (approved):

PR #11435, opened 2019-11-27
Bratseth/value function

bratseth:
Support bound functions ("closures") in tensor generate such that we can refer to other tensors, and index-generating expressions in tensor value. Used together this provides flexible tensor reshaping expressions.

More type parameter propagation and pulling Name up in the last PR is respectively arguably and strictly API breaking, but I don't think it will actually hurt anybody.


lesters (approved):
Phew. Great work! LGTM!

PR #11436, opened 2019-11-28
Remove old node fields

freva:


kkraune (commented):
Great! much better names :-)

bratseth:
Great! Note that cpu core count and vcpu are both needed in the flavor config (but not here afaik, so just mentioning it).

hakonhall (approved):

PR #11449, opened 2019-11-28
Fix ever-growing message tracker for concurrent Get operations

vekterli:
@geirst please review.

When Get requests initiated outside the main distributor thread were sent
via the `MessageSender` that is implemented by the main `Distributor` instance,
they would be implicitly registered with the pending message tracker.
Not only was this thread unsafe, the registrations would never be cleared
away since the reply pipeline bypassed it entirely. This would cause a silent
memory leak that would build up many small allocations over time.

We now dispatch Get requests directly through the storage link chain, bypassing
the message tracking component. This both fixes the leak and avoids extra
overhead for the Get requests.

Note: the concurrent Get feature is _not_ enabled by default.

Also fixes an issue where concurrent Get operations weren't correctly
gracefully aborted when the node shuts down.


geirst (approved):
👍

PR #11450, opened 2019-11-29
template cell types also

arnej27959:
* also, move common parameters to abstract superclass

@havardpe please review
@geirst FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_iterator.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_iterator.cpp
@@ -2,29 +2,13 @@
 
 #include "nearest_neighbor_iterator.h"
 
+using search::tensor::DenseTensorAttribute;
 using vespalib::ConstArrayRef;
+using vespalib::tensor::DenseTensorView;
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_iterator.cpp
```diff
@@ -34,32 +18,24 @@ namespace search::queryeval {
...
-          _distanceHeap(distanceHeap),
+
+    NearestNeighborImpl(Params params_in)
+        : NearestNeighborIterator(params_in),
+          _lhs(params().queryTensor.cellsRef().typify<LCT>()),
+          _fieldTensor(params().tensorAttribute.getTensorType()),
           _lastScore(0.0)
     {
-        assert(_fieldTensor.fast_type() == _queryTensor.fast_type());
+        assert(params().queryTensor.fast_type() == params().queryTensor.fast_type());
```
havardpe: I think this will always be true

havardpe (approved):
looks good, minor typo

I just realized we might be able to use select to choose the appropriate create_impl factory function to avoid having to enumerate all possible cell types... (possible future improvement)

PR #11451, opened 2019-11-29
Preserve DNS records for inactive load balancers

mpolden:
This change ensures that we preserve DNS records, as long as the underlying load
balancer exists, even if it's inactive.

Inactive routing policies are excluded in the application API so that endpoints
for deleted clusters are not shown in console.


hmusum (approved):
Great!

PR #11467, opened 2019-12-02
tensor peek

havardpe:
initial implementation of tensor peek (very slow). Supports peeking tensor sub-spaces in addition to peeking individual cell values. Also supports auto-conversion from numeric expression result to string label for mapped dimensions. Access outside tensors detected compile-time will result in type resolution failure. Run-time out-of-bounds access results in the value 0.0

@arnej27959 please review
@bratseth FYI


arnej27959 (approved):
👍

PR #11470, opened 2019-12-02
Jvenstad/remove old job status

jonmv:
@mpolden please review. 

The three first commits are part of an existing PR. The next three commits are also intended as a separate PR, before merging the last commit here.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
```diff
@@ -124,11 +119,6 @@
 
     // JobStatus field
     private static final String jobTypeField = "jobType";
-    private static final String errorField = "jobError";
```
mpolden: Does the previous version handle removal of these fields?

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
```diff
@@ -336,6 +336,12 @@ public void testApplicationApi() {
 
         deploymentTester.triggerJobs();
 
+        // POST a triggering to force a production job to start without successful tests
+        tester.assertResponse(request("/application/v4/tenant/tenant2/application/application2/instance/instance1/job/production-us-west-1", POST)
+                                      .data("{\"skipTests\":true}")
```
mpolden: 😆

mpolden (commented):
Nice cleanup!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
jonmv: Yes, all fields were already optional, except the job type.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
jonmv: 🙈

jonmv:
Merging it all together. Loss of the old job status data shouldn't be a problem, as the new one is already used for everything important.

PR #11473, opened 2019-12-02
Expose Cost API in Controller API classes

oyving:
Classes to parse Cost API results using ObjectMapper in hv tool

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/application/v4/CostResource.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/application/v4/CostResource.java
@@ -9,8 +9,10 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
-import java.util.List;
 
+/**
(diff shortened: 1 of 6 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/application/v4/CostResource.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import java.util.List;
+
+@Path("")  //Ensures that the produces annotation is inherited
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface CostResource {
```
mpolden: Author should be added, and javadoc. Also goes for the other classes.

oyving: Added the author tag, but skipped javadoc. 0 classes in that module have it, and they are just use to parse JSON data.

mpolden (approved):

PR #11474, opened 2019-12-02
Only run credentials maintainer for tenant containers

mpolden:


freva (commented):
This logic should be in host-admin, we can change `NodeAgentFactory` to also take in the initial context which is already generated in `NodeAdminImpl` (see `NodeAgentWithSchedulerFactory`).

mpolden:
Agree, that would be preferable. Good call!

Sent from a dive bar, somewhere in SF.

PR #11475, opened 2019-12-02
Preserve AssignValueUpdates to struct fields in indexing docproc

vekterli:
@geirst please review
@baldersheim FYI

Would previously be rewritten as `MapValueUpdates` for unknown reasons.
This wouldn't actually work, as an exception would be thrown during
serialization when sanity checking code figured out that the MapValue
update was attempted used for a non-array/wset field data type.


baldersheim (approved):

geirst (approved):
👍

baldersheim (approved):

PR #11476, opened 2019-12-02
Make parent-host-not-ready warning more user-friendly

hakonhall:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
```diff
@@ -119,10 +119,26 @@ private static void validateParentHosts(ApplicationId application, NodeList node
...
+            } else {
+                messageBuilder.append(nonActiveHosts.stream()
+                        .sorted()
+                        .limit(3)
+                        .collect(Collectors.joining(", ")))
+                        .append(", and others");
+            }
+            var message = messageBuilder.toString();
+
+            logger.log(LogLevel.INFO, application + ": " + message);
```
freva: Now that the log is pretty much the same as the exception, this is not really needed, all exceptions that lead to non 200 response from deploy are logged in https://github.com/vespa-engine/vespa/blob/259fde75fd9852cf02c275447e1f4b38114d030a/configserver/src/main/java/com/yahoo/vespa/config/server/http/HttpErrorResponse.java#L36

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
hakonhall: Application would then not be part of the log message, so I have made that part of the message (which unfortunately is redundant for the user).

freva (approved):

PR #11479, opened 2019-12-03
Wrap soft and hard doom in an CombinedDoom.

baldersheim:
If softdoom factor has been overridden, we shall not adjust the automatic one.

@havardpe PR


havardpe (commented):
I would prefer there to be only one class named Doom (no nesting inside CombinedDoom)
I would prefer doom.soft_doom() to doom.soft().doom(),
and doom.hard_doom() to doom.hard().doom()
This would make it harder to confuse the different dooms

baldersheim:
@havardpe Now there is only one doom. Now it is clear at the point of use which one you are using.

PR #11483, opened 2019-12-03
Bjorncs/system flags improvements

bjorncs:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
```diff
@@ -48,10 +48,8 @@ public static SystemFlagsDataArchive fromZip(InputStream rawIn) {
                 String name = entry.getName();
                 if (!entry.isDirectory() && name.startsWith("flags/") && name.endsWith(".json")) {
                     Path filePath = Paths.get(name);
-                    String filename = filePath.getFileName().toString();
-                    FlagData flagData = FlagData.deserializeUtf8Json(zipIn.readAllBytes());
-                    verifyFlagDataMatchesDirectoryName(filePath, flagData);
-                    builder.addFile(filename, flagData);
+                    String rawData = new String(zipIn.readAllBytes());
```
hakonhall: Would be nice to use an explicit Charset - otherwise this is system-specific and dependent on both client and server.

hakonhall (approved):

hakonhall (approved):

PR #11486, opened 2019-12-03
fix bug in hasAnyPositionField

arnej27959:
* clearly returning "true" in all cases is not useful
* but ordering of checks in hasAnyPostionDataType was wrong also:
  PositionDataType.INSTANCE is actually instanceof StructuredDataType
  so it would always return false.

@baldersheim please check if this was more or less what you wanted
@hmusum please review and merge


hmusum (approved):

baldersheim (commented):
:+

PR #11489, opened 2019-12-03
Only include node states that are meterable in metering

oyving:
Maintain a list of node states that are meterable, and make sure we include only these in our metering snapshots. Unfortunately, we maintain this list inside the Metering maintainer and not on the node state class itself. The reason for this is that we have _many_ different `Node.State` and `NodeState` classes that do exactly the same.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ResourceMeterMaintainer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ResourceMeterMaintainer.java
@@ -100,7 +100,7 @@ public class ResourceMeterMaintainer extends Maintainer {
     private static final Set<Node.State> METERABLE_NODE_STATES = Set.of(
             Node.State.reserved,   // an application will soon use this node
             Node.State.active,     // an application is currently using this node
-            Node.State.inactive    // an application owner might set a node inactive
+            Node.State.inactive    // an application is not using it, but it is reserved for being re-introduced or decommissioned
     );
 
     private boolean isNodeStateMeterable(Node node) {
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ResourceMeterMaintainer.java
```diff
@@ -85,9 +87,23 @@ protected void maintain() {
...
+    }
+
+    /**
+     * Checks if the node is in some state where it is in active use by the tenant,
+     * and not transitioning out of use, in a failed state, etc.
+     */
+    private static final Set<Node.State> METERABLE_NODE_STATES = Set.of(
+            Node.State.reserved,   // an application will soon use this node
+            Node.State.active,     // an application is currently using this node
+            Node.State.inactive    // an application owner might set a node inactive
```
olaaun: Nitpick, but I believe application owner has no control over deactivation of nodes?

olaaun (approved):
LGTM👍

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ResourceMeterMaintainer.java
freva: Correct, moved to inactive by configserver when the node is no longer needed (retired manually, requested flavor changed, number of nodes changed).

oyving: Aha! But they are assigned to a user, and just waiting for clean-up of the node?

freva: Yes, they are assigned to the user and just waiting to be expired, the purpose of that state is if the user deploy unintended change and starts losing nodes, they can undo it without losing data (the inactive nodes will be assigned back to them)

oyving:
@olaaun and @freva - I updated the comment based on your feedback. Don't want any misconceptions lingering in the code. Looks good now?

freva (approved):

olaaun (approved):

PR #11501, opened 2019-12-04
Preserve array updates with element index matching in indexing docproc

vekterli:
@geirst please review. More fun from the indexing docproc land discovered as I'm working to extend system test coverage of this stuff.
@baldersheim FYI

The resulting `MapValueUpdate` would previously be constructed with the
wrong type for the index, causing an exception to be thrown and for the
update to fail entirely.


baldersheim (approved):
👍 More magic fun

geirst (commented):
👍

PR #11504, opened 2019-12-04
Hide implementation to reduce FastOS_ visibility.

baldersheim:
@vekterli PR

### DIFF
--- a/staging_vespalib/src/vespa/vespalib/util/clock.cpp
+++ b/staging_vespalib/src/vespa/vespalib/util/clock.cpp
@@ -41,9 +41,8 @@ Updater::Updater(Clock & clock, double timePeriod)
 Updater::~Updater() = default;
 
 void
-Updater::Run(FastOS_ThreadInterface *thread, void *arguments)
+Updater::Run(FastOS_ThreadInterface *thread, void *)
 {
-    (void) arguments;
     _clock._running = true;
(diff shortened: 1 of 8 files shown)


staging_vespalib/src/vespa/vespalib/util/clock.cpp
```diff
@@ -1,24 +1,79 @@
...
-     _stop(false),
+     _updater(std::make_unique<clock::internal::Updater>(*this, timePeriod)),
      _running(false)
 {
     setTime();
 }
 
 Clock::~Clock()
 {
+    _updater.reset();
```
vekterli: Is it possible for this to race with the clock updater thread?
1. `Updater::stop()` is called while clock thread is waiting on the condition variable and is not holding the mutex
2. `Updater::stop()` returns before the clock thread has woken up to exit its loop
3. `_updater` is deleted, causing `Updater::Run()` to poke at freed memory

vekterli (commented):
One question on thread safety, otherwise LGTM

staging_vespalib/src/vespa/vespalib/util/clock.cpp
baldersheim: That is correct. I have added an explicit join if necessary.
The only use had protection on the outside due a threadPool.Close.

PR #11510, opened 2019-12-05
Lesters/add onnx unit tests

lesters:
@bratseth Please review. This was necessary to find the supported ONNX opsets for each operation (and correctly support them). This is all in `OnnxOperationTestCase.java` which is a good start point for this PR.

Documentation is incoming.


bratseth (approved):
Great!

PR #11514, opened 2019-12-05
Allow nearest neighbor operator where attribute tensor and query tens…

geirst:
…or have different cell types (float vs double).

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #11517, opened 2019-12-05
Allow minus sign at the start of a label.

bratseth:


lesters:
@bratseth @havardpe Are there any implications for the backend here?

havardpe:
yes, but I am working on supporting more relaxed labels in the back-end as well

vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
```diff
@@ -163,7 +163,7 @@ public String toString() {
     /** Supports building of a tensor address */
     public static class Builder {
 
-        private Pattern identifierPattern = Pattern.compile("[A-Za-z0-9_]+");
+        private Pattern identifierPattern = Pattern.compile("[-,A-Za-z0-9_@]([A-Z,a-z0-9_@$])*");
```
havardpe: I would suggest renaming this one now, sice we do not want it to be confused with identifiers

havardpe:
this will allow labels to contain anything in the back-end (including '-'):
https://github.com/vespa-engine/vespa/pull/11524

lesters (approved):

vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
bratseth: Good point; done

baldersheim (approved):

PR #11523, opened 2019-12-06
Only remove unused CNAMEs when removing obsolete policies MERGEOK

hmusum:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/RoutingPolicies.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/RoutingPolicies.java
@@ -153,10 +153,17 @@ public class RoutingPolicies {
             LOGGER.log(LogLevel.WARNING, "Removing " + removalCandidates + ". Active load balancers " +
                                          activeLoadBalancers);
         }
+        removalCandidates.forEach(allPolicies::remove);
+
+        // Find CNAMEs in use
+        var cnamesUsed = allPolicies.stream()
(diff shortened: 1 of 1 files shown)


tokle (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/RoutingPolicies.java
```diff
@@ -153,10 +153,17 @@ private void removeObsoletePolicies(AllocatedLoadBalancers loadBalancers, Set<Ro
...
+        // Find CNAMEs in use
+        var cnamesUsed = allPolicies.stream()
+                .map(policy -> policy.endpointIn(controller.system()).dnsName())
+                .collect(Collectors.toSet());
+
         for (var policy : removalCandidates) {
             var dnsName = policy.endpointIn(controller.system()).dnsName();
-            controller.nameServiceForwarder().removeRecords(Record.Type.CNAME, RecordName.from(dnsName), Priority.normal);
-            allPolicies.remove(policy);
+            if ( ! cnamesUsed.contains(dnsName))
```
mpolden: When does this happen? `dnsName` should be unique per policy the policy itself is for a single cluster.

hmusum: Internal ticket updated with logs showing why this PR was made, it looked like removing an old policy for the same CNAME ended up removing the CNAME as well, even though a newer policy existed.

PR #11524, opened 2019-12-06
improve label parsing in ranking expressions.

havardpe:
unquoted labels can now contain almost anything
labels can now also be quoted (in tensor create/peek)
NB: indexes in verbose tensor create cannot be quoted
expressions in tensor peek must have () around them
expression results are rounded before conversion
the sign is kept when expression results are converted to labels
trivial number expressions in tensor peek are converted to labels
... by the parser, to better inline bindings created by tensor lambdas
verbatim tensor peek for indexed dimensions will fail type resolution
... when the label cannot be converted to a valid index

@arnej27959 please review
@bratseth @lesters FYI


arnej27959 (approved):
👍

PR #11534, opened 2019-12-09
Revert "Revert "Allow config of ssl cipher suites and protocol version" MERGEOK"

jonmv:
Reverts vespa-engine/vespa#11503

@freva @bjorncs @tokle


freva (approved):

bjorncs:
This PR is incomplete. Replaced with https://github.com/vespa-engine/vespa/pull/12021 that contains additional commits that were reverted but not included here.

PR #11520, opened 2019-12-05
Revert "Revert "Use shell script wrapper for vespa-rpc-invoke""

toregge:
Reverts vespa-engine/vespa#11519

@geirst: please review

PR #11538, opened 2019-12-10
Revert "Revert "Revert "Use shell script wrapper for vespa-rpc-invoke"""

jonmv:
Reverts vespa-engine/vespa#11520

System tests fail with `command not found: vespa-rpc-invoke`.


jonmv:
@toregge @baldersheim @freva

freva (approved):

PR #11545, opened 2019-12-11
Jvenstad/remove old job status

jonmv:
@mpolden please review.

This is WIP, and there is quite some cleanup to do. Nevertheless, I think this all works now, i.e., the new deployment spec syntax, with nested parallel and serial blocks of steps, and with declared production tests. 

Some APIs need to remain until replacements have been available for a while, and some cleanup will have to wait for this as well. 

@bratseth FYI.

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
@@ -99,6 +99,13 @@ public class ApplicationPackageBuilder {
         return this;
     }
 
+    public ApplicationPackageBuilder test(String regionName) {
+        environmentBody.append("      <test>");
+        environmentBody.append(regionName);
+        environmentBody.append("</test>\n");
+        return this;
+    }
+
     public ApplicationPackageBuilder parallel(String... regionName) {
         environmentBody.append("    <parallel>\n");
         Arrays.stream(regionName).forEach(this::region);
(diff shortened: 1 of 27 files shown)


bratseth:
Great! If/once we have enough testing to know this works we need to put out a post letting people know about these new capabilities. There are many hacks out there to compensate for not having multiple instances, and more people should have prod tests.

jonmv:
I'm pretty confident it works, with the unit scenario tests we have, but documentation is not done. I would also like to have this verified in the CD system, of course. 
In any case, we should ask anyone with production tests to declare these with the new syntax, so we may stop running them as part of the deployment job.

jonmv:
Ok, so there are some kinks to straighten out ... Expect some more commits :)

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
```diff
@@ -99,6 +99,13 @@ public ApplicationPackageBuilder region(String regionName) {
         return this;
     }
 
+    public ApplicationPackageBuilder test(String regionName) {
+        environmentBody.append("      <test>");
```
mpolden: Have we decided on this tag name? `<test>` already means something else.

mpolden (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
jonmv: Yes, it was through a review.

PR #11548, opened 2019-12-12
Bratseth/mixed tensor parse  

bratseth:


bratseth:
@lesters please review

- Support mixed format: {key1:[1,2,3],key2:[4,5,6]}
- Validate tensor inner structure if used (but allow flat arrays in string form, not ranking expressions)
- Interpret tensor values according to the order in which the tensor dimensions are written.

@havardpe fyi

lesters (approved):
Great!

PR #11550, opened 2019-12-12
Update config change validator to require restart (and not refeed) wh…

toregge:
…en adding attribute aspect to a struct field.

@geirst : please review


geirst (approved):
👍

PR #11551, opened 2019-12-12
async concurrent llvm compilation

havardpe:
* Allow llvm compilation to be performed concurrently (remove global mutex and assert that llvm is compiled with thread support when initializing, also verify with unit tests).
* Functions are now always owned by shared pointers and inherit from enable_shared_from_this to enable extending their life-time during async operations.
* The compile cache will compile all expressions in the calling thread by default, but now supports attaching an Executor to it that will handle compilation asynchronously.

The compile cache did not get its own pool of threads since proton wants to control thread usage. In order to enable concurrent async compilation in proton; attach an appropriate executor to the compile cache before setting up ranking profiles and sync with the executor before serving queries (to avoid slow initial queries that wait for compilation completion).


havardpe:
@arnej27959 please review
@baldersheim @geirst FYI

arnej27959 (approved):
looks good to me

PR #11552, opened 2019-12-12
Revert "Revert "Allow minus sign at the start of a label.""

baldersheim:
Reverts vespa-engine/vespa#11547
@bratseth or @geirst PR
This was indeed the culprit. We need to improved soon.


bratseth:
Yes, I have a plan for it. Just need to finish something else first ...

PR #11558, opened 2019-12-13
dsl library for vespa search queries

yehzu:
Hi, this is a java implementation for composing Vespa queries from the client-side.
Please kindly review. :) 

@bratseth @jobergum


client/src/main/java/com/yahoo/vespa/client/dsl/DotProduct.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "dotProduct(" + fieldName + ", " + Q.gson.toJson(weightedSet) + ")";
+    }
+
+    @Override
+    boolean hasPositiveSearchField(String fieldName) {
+        // TODO: implementation
```
bratseth: Better throw for now then.

client/src/main/java/com/yahoo/vespa/client/dsl/DotProduct.java
```diff
@@ -0,0 +1,50 @@
...
+        return false;
+    }
+
+    @Override
+    boolean hasPositiveSearchField(String fieldName, Object value) {
+        // TODO: implementation
+        return false;
```
bratseth: Better throw for now then.

client/src/main/java/com/yahoo/vespa/client/dsl/DotProduct.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    boolean hasNegativeSearchField(String fieldName) {
+        // TODO: implementation
+        return false;
```
bratseth: Better throw for now then.

client/src/main/java/com/yahoo/vespa/client/dsl/DotProduct.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    boolean hasNegativeSearchField(String fieldName, Object value) {
+        // TODO: implementation
+        return false;
```
bratseth: Better throw for now then.

bratseth (approved):
This is a great contribution.
- Could you add some javadoc to the class headings so it's possible to know what they are for without cross-referencing the test?
- Ideally we should use "ai.vespa" as package root instead of "com.yahoo" now, but it might create problems for you?

yehzu:
- Yes, I will add some javadocs to clarify the purpose of these classes. 
- No, It has no concern for us to move the root package to `ai.vespa`. I will submit another commit for it
Thanks for your review :)

bratseth (approved):
Great!

bratseth:
You need to replace "&" by "&amp;" in the Javadoc:

An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /source/client/src/main/java/ai/vespa/client/dsl/FixedQuery.java:371: error: bad HTML entity
[ERROR]      * build the vespa query string join by '&'

PR #11559, opened 2019-12-13
Add a faster, but simpler pattern matcher. Builds tensoraddresses 5 t…

baldersheim:
…imes+ faster than the regex one.

@bratseth or @geirst PR


bratseth (approved):
Nice!

PR #11560, opened 2019-12-13
Jvenstad/deplorch cleanup 1

jonmv:
@freva please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/versions/VespaVersion.java
```diff
@@ -157,8 +157,10 @@ public boolean canChangeTo(Confidence target, Instant instant) {
             if (!target.equalOrHigherThan(normal)) return true; // Confidence can always change to < normal
 
             var hourOfDay = instant.atZone(ZoneOffset.UTC).getHour();
-            // Confidence can only be raised between 05:00:00 and 11:59:59 UTC
-            return hourOfDay >= 5 && hourOfDay <= 11;
+            var dayOfWeek = instant.atZone(ZoneOffset.UTC).getDayOfWeek();
+            // Confidence can only be raised between 05:00:00 and 11:59:59 UTC, and not during weekends.
```
freva:
```suggestion
            // Confidence can only be raised between 05:00:00 and 11:59:59 UTC, and not during weekends or Friday.
```

freva (approved):

PR #11561, opened 2019-12-13
Avoid fast past update restart race with concurrently created replica

vekterli:
@geirst please review. This fix was hammered out in a cramped seat at 35,000 feet (occasionally interrupted by a high fidelity surround sound setup of screaming infants) so I will likely polish these changes a bit later...!

After the recent change to allow safe path updates to be restarted
as fast path updates iff all observed document timestamps are equal,
a race condition regression was introduced. If the bucket that the
update operation was scheduled towards got a new replica concurrently
created _between_ the time that safe path Gets were sent and received,
it was possible for updates to be sent to inconsistent replicas. This
is because the Get and Update operations use the current database
state at _their_ start time, not a stable snapshot state from the start
time of the two phase update operation itself.

Add an explicit check that the replica state between sending Gets and
Updates is unchanged. If it has changed, a fast path restart is _not_
permitted.


geirst (approved):
👍

PR #11565, opened 2019-12-13
avoid race in unit test

havardpe:
fix race in unit test
(also reduce cost of test overall)

@arnej27959 please review


arnej27959 (approved):
👍

PR #11569, opened 2019-12-16
Add convenience method

bratseth:


container-search/src/main/java/com/yahoo/search/searchchain/ExecutionFactory.java
```diff
@@ -66,6 +66,15 @@ public Execution newExecution(Chain<? extends Searcher> searchChain) {
                              new Execution.Context(searchChainRegistry, indexFacts, specialTokens, rendererRegistry, linguistics));
     }
 
+    /**
+     * Creates a new execution starting at a search chain.
+     * An execution instance should be used once to execute a (tree of) search chains.
+     */
+    public Execution newExecution(String searchChainId) {
+        return new Execution(searchChainRegistry().getChain(searchChainId),
```
arnej27959: consider chaining to newExecution(searchChainRegistry().getChain(searchChainId))
to avoid duplicating context creation

arnej27959 (approved):
looks good to me

container-search/src/main/java/com/yahoo/search/searchchain/ExecutionFactory.java
bratseth: I'll do it the next time I'm there - thanks

PR #11570, opened 2019-12-16
Limit number of simultaneous OS upgrades 

mpolden:
We can consider being smarter about picking which nodes to upgrade (e.g. pick
nodes having disjunct sets of applications) if this proves too slow.


mpolden:
or @bratseth (@freva is out this week).

bratseth (approved):
Nice work!

PR #11578, opened 2019-12-17
Jvenstad/old instance data cleanup

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
@@ -146,8 +146,16 @@ public class ApplicationController {
         Once.after(Duration.ofMinutes(1), () -> {
             Instant start = clock.instant();
             int count = 0;
-            for (Application application : curator.readApplications()) {
-                lockApplicationIfPresent(application.id(), this::store);
+            for (TenantAndApplicationId id: curator.readApplicationIds()) {
+                lockApplicationIfPresent(id, application -> {
+                    if (id.tenant().value().startsWith("by-"))
+                        application = application.with(DeploymentSpec.empty);
+                    else
+                        for (InstanceName instance : application.get().deploymentSpec().instanceNames())
+                            if ( ! application.get().instances().keySet().contains(instance))
+                                application = application.withNewInstance(instance);
+                    store(application);
+                });
                 count++;
             }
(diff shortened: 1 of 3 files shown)


jonmv:
Makes it clunky to completely delete applications with more than one instance, but there aren't any of those yet. Will fix this this in a more elegant way shortly.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -146,8 +146,16 @@
         Once.after(Duration.ofMinutes(1), () -> {
             Instant start = clock.instant();
             int count = 0;
-            for (Application application : curator.readApplications()) {
-                lockApplicationIfPresent(application.id(), this::store);
+            for (TenantAndApplicationId id: curator.readApplicationIds()) {
+                lockApplicationIfPresent(id, application -> {
+                    if (id.tenant().value().startsWith("by-"))
```
mpolden: In the future we should disallow creation of by tenants altogether and remove this.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -718,6 +718,7 @@ public void deleteApplication(TenantAndApplicationId id, Optional<Credentials> c
         if (instances.size() > 1)
             throw new IllegalArgumentException("Could not delete application; more than one instance present: " + instances);
 
+        lockApplicationOrThrow(id, application -> store(application.with(DeploymentSpec.empty)));
```
mpolden: Hopefully not a long-lived solution. 🙂

mpolden (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
jonmv: Hehe.

jonmv: No, this all needs to be solved properly. Deletion at application level, with a confirmatino dialogue with a list of all instances to be deleted, or something.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
jonmv: Absolutely.

PR #11582, opened 2019-12-17
Bratseth/quoted tensor labels 

bratseth:
@lesters please review

This allows any string as a (mapped) tensor label, where non-identifier strings must be quoted by single or double quotes.

Strings containing double quotes will be output with single quotes and vice versa, so the backend must also parse both kinds. Let me know if this isn't ok @havardpe .


lesters (approved):
LGTM.

havardpe:
why can't we use the same format as the quoted string we already support?

PR #11589, opened 2019-12-18
Stop requiring at least 1 Gb heap when using heaps size as percentage…

hmusum:
… of physical memory

Make it easier to get this working for e.g. Docker containers with less
memory.


baldersheim (approved):

hmusum:
@mpolden FYI

PR #11593, opened 2019-12-18
Jvenstad/deplorch cleanup 1

jonmv:
@mpolden please review.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/xml/DeploymentSpecXmlReader.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/xml/DeploymentSpecXmlReader.java
@@ -92,10 +92,9 @@ public class DeploymentSpecXmlReader {
 
     /** Reads a deployment spec from XML */
     public DeploymentSpec read(String xmlForm) {
-        if (DeploymentSpec.empty.xmlForm().equals(xmlForm))
-            return DeploymentSpec.empty;
-
         Element root = XML.getDocument(xmlForm).getDocumentElement();
+        if (isEmptySpec(root))
+            return DeploymentSpec.empty;
 
         List<Step> steps = new ArrayList<>();
         if ( ! containsTag(instanceTag, root)) { // deployment spec skipping explicit instance -> "default" instance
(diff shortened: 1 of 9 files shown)


jonmv:
Huh, what? Thought I already merged the 4 first commits here.

jonmv:
There we go ...

config-model-api/src/main/java/com/yahoo/config/application/api/xml/DeploymentSpecXmlReader.java
```diff
@@ -420,6 +419,13 @@ private boolean readActive(Element regionTag) {
                                            "to control whether the region should receive production traffic");
     }
 
+    private static boolean isEmptySpec(Element root) {
+        if ( ! XML.getChildren(root).isEmpty()) return false;
+        return    root.getAttributes().getLength() == 0
+               || root.getAttributes().getLength() == 1 && root.hasAttribute("version");
```
mpolden: Consider adding parentheses to make the evaluation order explicit.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackage.java
```diff
@@ -49,22 +49,40 @@
...
      * it must not be further changed by the caller.
      */
     public ApplicationPackage(byte[] zippedContent) {
+        this(zippedContent, false);
+    }
+
+    /**
+     * Creates an application package from its zipped content.
+     * This <b>assigns ownership</b> of the given byte array to this class;
+     * it must not be further changed by the caller.
```
mpolden: Consider copying the array, like we do in other classes.

mpolden (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackage.java
jonmv: I guess this isn't done because the array can be pretty large?

bratseth: Yes.

jonmv:
@mpolden PTAL at the last commit as well.

mpolden (approved):

PR #11597, opened 2019-12-18
This is the correct way of shutting down the the Dispatcher.

baldersheim:
Avoid using Thread.interrupt for shutdown, handle it gracefully.
1 - Shutdown the SearchCluster
  1 - Set it to closed.
  2 - Synchronize clearing of monitored nodes.
  3 - Wake up monitor thread if sleeping.
  4 - Join with monitorthread.
2 - Release the invokerPool.

@bratseth PR


bratseth (approved):
👍

PR #11605, opened 2019-12-20
If it is full, it is full. No point in moving data to new buffers as …

baldersheim:
…nothing will be added anyway.

@arnej27959 or @geirst or @havardpe PR


havardpe (commented):
👍

PR #11606, opened 2019-12-20
Application metrics api

gjoranv:
This should be ok to review commit-by-commit.

~~Note that the components are not set up in the model yet, so this is just for adding the code.~~


olaaun (approved):

gjoranv:
@olaaun Please review the last 8 commits after merge. The only significant changes are a bug fix and adding the components to metrics-proxy via the config model.

olaaun:
LGTM.
Perhaps we should cache metrics differently? The cache doesn't consider if two requests are for different consumers, so we could end up returning the wrong metric set

gjoranv:
Good catch. I forgot the cache when I added consumer handling. Will create a new PR to fix.

PR #11610, opened 2019-12-20
Jvenstad/deplorch cleanup 1

jonmv:
@hakonhall please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
@@ -3,6 +3,7 @@ package com.yahoo.vespa.hosted.controller.persistence;
 
 import com.google.common.util.concurrent.UncheckedTimeoutException;
 import com.google.inject.Inject;
+import com.yahoo.collections.Pair;
 import com.yahoo.component.Version;
 import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.HostName;
@@ -29,6 +30,7 @@ import com.yahoo.vespa.hosted.controller.versions.OsVersion;
 import com.yahoo.vespa.hosted.controller.versions.OsVersionStatus;
 import com.yahoo.vespa.hosted.controller.versions.VersionStatus;
 import com.yahoo.vespa.hosted.controller.versions.VespaVersion;
+import org.apache.zookeeper.data.Stat;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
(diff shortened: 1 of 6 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -376,15 +380,23 @@ public void writeLastRun(Run run) {
     }
 
     public void writeHistoricRuns(ApplicationId id, JobType type, Iterable<Run> runs) {
-        curator.set(runsPath(id, type), asJson(runSerializer.toSlime(runs)));
+        Path path = runsPath(id, type);
+        cachedHistoricRuns.remove(path);
```
hakonhall: Normally the `remove` would be placed after `set`, but in this case even the reads are done while holding the same exclusive lock it seems, and in case the ordering doesn't matter.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -97,6 +99,8 @@
     private final Curator curator;
     private final Duration tryLockTimeout;
 
+    private final Map<Path, Pair<Integer, NavigableMap<RunId, Run>>> cachedHistoricRuns = new ConcurrentHashMap<>();
```
hakonhall: Would be nice to explain the Integer

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
jonmv: Assume the write fails. Since the cache has been modified with a write that is not reflected in ZK, I thought it better to discard the cache entry here. Writes to the historic runs are only done when a finished run is moved there, and an old entry is (optionally) evicted. This write is retried until it succeeds. 

So, actually, since the runs are indexed by run ID, the cache modification (addition of the new run) would be overwritten again by the next attempt to write the (same) run (with the same run ID) to the cache, so then ordering doesn't matter at all.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
jonmv: You're right.

jonmv:
```suggestion
    // For each job id (path), store the ZK node version and its deserialised data — update when version changes. 
    private final Map<Path, Pair<Integer, NavigableMap<RunId, Run>>> cachedHistoricRuns = new ConcurrentHashMap<>();
```

PR #11611, opened 2019-12-20
Add more details to the trace.

baldersheim:
@havardpe PR The last one before Christmas


baldersheim:
@havardpe Now it should be closer to your preferences.

havardpe (approved):

PR #11614, opened 2019-12-20
Count incomplete upgrades in VersionStatus

mpolden:


mpolden:
Should probably only count consecutive failed runs. Will fix.

mpolden:
PTAL

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/versions/VersionStatus.java
```diff
@@ -202,13 +201,24 @@ private static NodeVersions findSystemApplicationVersions(Controller controller)
...
+                for (var run : runsOfFailingJobs) {
+                    if (run.status() == RunStatus.success) {
+                        consecutiveFailingRuns.clear();
+                    } else {
+                        consecutiveFailingRuns.add(run);
+                    }
+                }
+                consecutiveFailingRuns.stream().map(run -> run.versions().targetPlatform())
+                                      .forEach(version -> versionMap.put(version, versionMap.getOrDefault(version, DeploymentStatistics.empty(version))
+                                                                                            .withFailing(id)));
```
jonmv: How about
```java
                jobs.failing()
                    .not().failingApplicationChange()
                    .not().withStatus(RunStatus.outOfCapacity)
                    .mapToList(JobStatus::runs)
                    .forEach(runs -> runs.descendingMap().values().stream()
                                         .dropWhile(run -> ! run.hasEnded())
                                         .takeWhile(run -> run.hasFailed())
                                         .map(run -> run.versions().targetPlatform())
                                         .forEach(version -> versionMap.put(version,
                                                                            versionMap.getOrDefault(version, DeploymentStatistics.empty(version))
                                                                                      .withFailing(id))));
```

the jobs contains runs for all job types, each of which may be relevant — not just the consecutive failures for the last job type. 

Also, jobs which are running (not `success`) aren't failures.

mpolden: Thanks, I got lost in the *List jungle when I tried a different approach. 😅

jonmv: Haha, I know. I get lost myself >_<

jonmv (approved):

mpolden:
Java build passed.

PR #11617, opened 2019-12-26
Vespa doc op feature add

leisureshadow:
### There are three features added in this PR:
* create-if-non-existent
* treat bag as map // bag-as-map-fields
* treat tuple as map // simple-object-fields

### create-if-non-existent
To create a new doc when the operation is update and the doc is non-existent

### treat bag as map
We cannot just use Map when the value is a Struct. e.g. field matchTerm type map<long, Term> where Term is a Struct.
The schema of Term cannot be preserved if we store the data to disk.
To preserve the schema, use Bag instead of Map.
The first element of Tuple is key. The second element of Tuple is value.

### treat tuple as map
When the type of values are different in a Map, the schema cannot be correctly stores as well.
To preserve it, use Tuple instead of Map.
The alias of Tuple field is key. The corresponding element in Tuple is value.


lesters (approved):
LGTM!

PR #11619, opened 2020-01-01
Balder/add executeinfo

baldersheim:
@havardpe and @geirst PR
The intention is to provide more information to fetchPostings making it easier to make smarter decisions.

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
@@ -317,6 +317,13 @@ IntermediateBlueprint::calculateState() const
     return state;
 }
 
+ExecuteInfo
+IntermediateBlueprint::computeNext(const Blueprint & child, const ExecuteInfo & execInfo) const
+{
+    (void) child;
+    return execInfo;
+}
+
 bool
 IntermediateBlueprint::should_do_termwise_eval(const UnpackInfo &unpack, double match_limit) const
 {
@@ -414,9 +421,11 @@ IntermediateBlueprint::visitMembers(vespalib::ObjectVisitor &visitor) const
 void
 IntermediateBlueprint::fetchPostings(const ExecuteInfo &execInfo)
 {
+    ExecuteInfo childInfo = ExecuteInfo(execInfo.isStrict() && inheritStrict(0), execInfo.hitRate());
     for (size_t i = 0; i < _children.size(); ++i) {
-        ExecuteInfo childInfo = ExecuteInfo(execInfo.isStrict() && inheritStrict(i), execInfo.hitRate());
-        _children[i]->fetchPostings(childInfo);
+        Blueprint & child = *_children[i];
+        child.fetchPostings(childInfo);
+        childInfo = computeNext(child, ExecuteInfo(execInfo.isStrict() && inheritStrict(i+1), childInfo.hitRate()));
     }
 }
 
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
@@ -3,7 +3,7 @@
 #include "blueprint.h"
 #include "leaf_blueprints.h"
 #include "intermediate_blueprints.h"
-#include "equiv_blueprint.h"
+#include <vespa/searchlib/fef/termfieldmatchdataarray.h>
 #include <vespa/vespalib/objects/visit.hpp>
 #include <vespa/vespalib/objects/objectdumper.h>
 #include <vespa/vespalib/objects/object2slime.h>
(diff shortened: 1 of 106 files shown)


searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
```diff
@@ -308,6 +317,13 @@ IntermediateBlueprint::calculateState() const
     return state;
 }
 
+ExecuteInfo
+IntermediateBlueprint::computeNext(const Blueprint & child, const ExecuteInfo & execInfo) const
```
havardpe: since strictness is handled externally, maybe this function should only be about hit ratios?

searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
```diff
@@ -403,11 +419,13 @@ IntermediateBlueprint::visitMembers(vespalib::ObjectVisitor &visitor) const
...
-IntermediateBlueprint::fetchPostings(bool strict)
+IntermediateBlueprint::fetchPostings(const ExecuteInfo &execInfo)
 {
+    ExecuteInfo childInfo = ExecuteInfo(execInfo.isStrict() && inheritStrict(0), execInfo.hitRate());
     for (size_t i = 0; i < _children.size(); ++i) {
-        bool strictChild = (strict && inheritStrict(i));
-        _children[i]->fetchPostings(strictChild);
+        Blueprint & child = *_children[i];
+        child.fetchPostings(childInfo);
+        childInfo = computeNext(child, ExecuteInfo(execInfo.isStrict() && inheritStrict(i+1), childInfo.hitRate()));
```
havardpe: minor consistency issue: asking if a non-existing child should inherit strictness. (not an actual issue in foreseeable future)

searchlib/src/vespa/searchlib/queryeval/executeinfo.cpp
```diff
@@ -0,0 +1,20 @@
+// Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "executeinfo.h"
+
+namespace search::queryeval {
+
+const ExecuteInfo ExecuteInfo::TRUE(true, 1.0);
+const ExecuteInfo ExecuteInfo::FALSE(true, 1.0);
```
havardpe: Maybe the FALSE one should actually be false.

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
```diff
@@ -0,0 +1,29 @@
...
+
+namespace search::queryeval {
+
+    /**
+     * Holds information about how query will be executed and how large part of corpus will pass through.
+     * @author baldersheim
+     */
+class ExecuteInfo {
+public:
+    ExecuteInfo() : ExecuteInfo(false, 1.0) { }
```
havardpe: consider using 0.0 as default (is this only needed by tests?)

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
```diff
@@ -0,0 +1,29 @@
...
+    ExecuteInfo(bool strict, double hitRate_in)
```
havardpe: consider hiding this if the create function is preferred.

searchlib/src/vespa/searchlib/queryeval/intermediate_blueprints.cpp
```diff
@@ -243,6 +243,11 @@ AndBlueprint::createIntermediateSearch(const MultiSearch::Children &subSearches,
     return SearchIterator::UP(search);
 }
 
+ExecuteInfo
+AndBlueprint::computeNext(const Blueprint & child, const ExecuteInfo &execInfo) const {
+    return ExecuteInfo(execInfo.isStrict(), (execInfo.hitRate()*child.getState().estimate().estHits)/get_docid_limit());
```
havardpe: the blueprint already has a function called hit_ratio that does this conversion (estimate to ratio)

havardpe (commented):
looks good overall (except FALSE being true)
please consider my comments

searchlib/src/vespa/searchlib/queryeval/executeinfo.cpp
baldersheim: Fixed

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
baldersheim: Hidden

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
baldersheim: Why 0.0 ? 1.0 sounds like a better(more conservative) value.
Only used in tests.

searchlib/src/vespa/searchlib/queryeval/intermediate_blueprints.cpp
baldersheim: Done, using that.

searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
baldersheim: Fixed

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
havardpe: reason for comment: false is the default for bool and 0.0 is the default for double.
I think it would be better to just remove it if possible.

PR #11628, opened 2020-01-02
Normalized vcpu capping

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContext.java
```diff
@@ -42,6 +42,14 @@ default NodeType nodeType() {
 
     String vespaUserOnHost();
 
+    /**
+     * The vcpu value in NodeSpec is multiplied by the speedup factor per cpu core compared to a historical baseline
```
hakonhall: The "normalizedVcpu" and "vcpu" methods should swap names.

hakonhall (approved):

PR #11618, opened 2019-12-27
Jvenstad/deplorch cleanup 1

jonmv:
@smorgrav take a look if you wish.

PR #11629, opened 2020-01-02
Jvenstad/instances with separate change

jonmv:
@mpolden please review. This fixes a bug introduced in #11618


jonmv:
I guess the bug fix could be a separate PR — hang on.

jonmv:
Ah, well ... then I get conflicts. Perhaps we can keep it together.

jonmv:
I'll do it separately.

PR #11631, opened 2020-01-02
Adds steps to the run REST API

hakonhall:
Adds a "steps" field to the
/application/v4/tenant/{tenant}/application/{application}/instance/{instance}/job/{jobtype}/run/{number}
REST API with the following structure:

  "steps": {
    "deployTester": {
      "status": "succeeded",
      "startMillis": 1577972067092
    },
    "deployReal": {
      "status": "succeeded",
      "startMillis": 1577972067092
    },
    ...
  }

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
@@ -12,7 +12,6 @@ import com.yahoo.vespa.hosted.controller.deployment.StepInfo;
 import com.yahoo.vespa.hosted.controller.deployment.StepRunner;
 import org.jetbrains.annotations.TestOnly;
 
-import java.time.Clock;
 import java.time.Duration;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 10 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
```diff
@@ -34,18 +34,21 @@
...
     @TestOnly
-    public JobRunner(Controller controller, Duration duration, JobControl jobControl, ExecutorService executors, StepRunner runner) {
+    public JobRunner(Controller controller, Duration duration, JobControl jobControl, ExecutorService executors,
+                     StepRunner runner, Clock clock) {
         super(controller, duration, jobControl);
         this.jobs = controller.jobController();
         this.jobs.setRunner(this::advance);
         this.executors = executors;
         this.runner = runner;
+        this.clock = clock;
```
jonmv: All `Maintainer`s have access to the controller's clock through `controller().clock()` already.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTester.java
```diff
@@ -90,7 +87,7 @@ public DeploymentTester(ControllerTester controllerTester) {
         cloud = (MockTesterCloud) tester.controller().jobController().cloud();
         var jobControl = new JobControl(tester.controller().curator());
         runner = new JobRunner(tester.controller(), Duration.ofDays(1), jobControl,
-                               JobRunnerTest.inThreadExecutor(), new InternalStepRunner(tester.controller()));
+                               JobRunnerTest.inThreadExecutor(), new InternalStepRunner(tester.controller()), tester.clock());
```
jonmv: And the tester's clock is the same as the controller's, in this case.

jonmv (approved):
I'd like to see the extra clock reference go, but otherwise looks good :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
hakonhall: Fixed

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTester.java
hakonhall: Fixed

jonmv (approved):
Very good! :)

PR #11661, opened 2020-01-06
VESPA-16236: Adding endpoint as a dimension to handled.requests and handled.latency

bjormel:

@yngveaasheim Not sure if it really makes any sense to include endpoint as a dimension to handled.latency 

@arnej27959


arnej27959 (approved):
looks good to me

yngveaasheim:
@bjormel Yes, we should be the same set of tags for handled.requests and handled.latency to be able to correlate changes in either.

PR #11667, opened 2020-01-07
Modify result immediately

bratseth:


container-search/src/main/java/com/yahoo/search/federation/FederationResult.java
```diff
@@ -94,6 +94,7 @@ private TargetResult(FederationSearcher.Target target, FutureResult futureResult
         public Optional<Result> getIfAvailable(long timeout) {
             if (availableResult.isPresent()) return availableResult;
             availableResult = futureResult.getIfAvailable(timeout, TimeUnit.MILLISECONDS);
+            availableResult.ifPresent(result -> target.modifyTargetResult(result));
```
arnej27959: it looks like this will cause "modifyTargetResult" to run a non-deterministic number of times, which could be a problem

bratseth: I don't think so? 
The first line returns the result if it's already assigned.
And the class documents that only a single thread accesses this so there should be no race.

arnej27959 (approved):
ok, i think i understand it now

PR #11668, opened 2020-01-07
Migrate AWS target group health checks

tokle:


mpolden (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -183,6 +184,12 @@
             "Whether to disable CM3.", "Takes effect on next host admin tick",
             HOSTNAME);
 
+    public static final UnboundBooleanFlag MIGRATE_AWS_TARGET_GROUP_HEALTH_CHECKS = defineFeatureFlag(
+            "migrate-aws-health-checks", false,
+            "Migrate target group health checks to https",
+            "Takes effect on config server restart",
```
mpolden: Doesn't it take effect on redeploy?

PR #11671, opened 2020-01-07
Setup default build settings based on detected vespa build platform using make.

toregge:
@geirst : please review
@aressem : FYI


geirst:
I think it is better for @aressem to review this

aressem (approved):

PR #10510, opened 2019-09-04
VESPA-16236: endpoint as a dimension to search requests

bjormel:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #11679, opened 2020-01-07
Revert "VESPA-16236: endpoint as a dimension to search requests"

bjormel:
Reverts vespa-engine/vespa#10510

This in now done in [ThreadedRequestHandler](https://github.com/vespa-engine/vespa/blob/master/container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java) handled for all handlers, not only "/search" #11661


baldersheim:
Close or complete ?

bjormel:
This is just a cleanup, as we changed this to be able to track both search and feed access.
@yngveaasheim can you please review

yngveaasheim (approved):
👍

Issue #8788, opened 2019-03-14
Your project vespa-engine vespa is using buggy third-party libraries [WARNING]


FDUSELAB2:
Hi, there!

We are a research team working on third-party library analysis. We have found that some widely-used third-party libraries in your project have major/critical bugs, which will degrade the quality of your project. We highly recommend you to update those libraries to new versions.

We have attached the buggy third-party libraries and corresponding jira issue links below for you to have more detailed information. We have analyzed the api call related to the following libraries and found one library that is using the API call that might invoke buggy methods in the library of the history.
1. commons-codec commons-codec
version: 1.4
API call in your project:org.apache.commons.codec.binary.Base64.setInitialBuffer(byte[],int,int)

Jira issues:
[Base64InputStream#read(byte[]) incorrectly returns 0 at end of any stream which is multiple of 3 bytes long](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-101?filter=allopenissues)
version:1.4
[ArrayIndexOutOfBoundsException when doing multiple reads() on encoding Base64InputStream](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-105?filter=allopenissues)
version:1.4
[Base64 encoding issue for larger avi files](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-108?filter=allopenissues)
version:1.4
[org.apache.commons.codec.net.URLCodec.ESCAPE_CHAR isn't final but should be](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-111?filter=allopenissues)
version:1.2;1.3;1.4
[org.apache.commons.codec.language.RefinedSoundex.US_ENGLISH_MAPPING should be package protected MALICIOUS_CODE](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-113?filter=allopenissues)
version:1.4
[org.apache.commons.codec.language.Soundex.US_ENGLISH_MAPPING should be package protected MALICIOUS_CODE](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-114?filter=allopenissues)
version:1.4
[Caverphone encodes names starting and ending with "mb" incorrectly.](https://issues.apache.org/jira/projects/CODEC/issues/CODEC-117?filter=allopenissues)
version:1.4
[...]

PR #11681, opened 2020-01-07
Bjorncs/apache commons libraries cleanup

bjorncs:
Part of #8788

### DIFF
--- a/container-test/pom.xml
+++ b/container-test/pom.xml
@@ -39,10 +39,6 @@
       <artifactId>commons-collections</artifactId>
     </dependency>
     <dependency>
-      <groupId>commons-lang</groupId>
-      <artifactId>commons-lang</artifactId>
-    </dependency>
-    <dependency>
       <groupId>io.airlift</groupId>
       <artifactId>airline</artifactId>
       <exclusions>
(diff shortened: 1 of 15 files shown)


container/pom.xml
```diff
@@ -45,10 +45,6 @@
           <groupId>io.airlift</groupId>
           <artifactId>airline</artifactId>
         </exclusion>
-         <exclusion>
```
gjoranv: Please run 'mvn dependency:tree' before and after to check for any changes.

container-test/pom.xml
```diff
@@ -38,10 +38,6 @@
       <groupId>commons-collections</groupId>
       <artifactId>commons-collections</artifactId>
     </dependency>
-    <dependency>
```
gjoranv: Please verify with 'mvn dependency:tree'

gjoranv:
Please build the whole vespa repo before these changes and run dependency:tree for:
* container
* container-test
* config-model-fat

Keep the files and repeat the procedure with these changes. Then, compare before and after dependency trees.

container/pom.xml
bjorncs: Nothing there.

container-test/pom.xml
bjorncs: It's not there 👍

gjoranv (approved):

PR #11683, opened 2020-01-07
Deprovision tenant hosts affected by cloud event

olaaun:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
@@ -49,28 +49,31 @@ public class CloudEventReporter extends Maintainer {
         for (var awsRegion : zonesByCloudNativeRegion.keySet()) {
             List<CloudEvent> events = eventFetcher.getEvents(awsRegion);
             for (var event : events) {
-                deprovisionHosts(awsRegion, event);
-                submitIssue(event);
+                List<String> deprovisionedHosts = deprovisionHosts(awsRegion, event);
+                submitIssue(event, deprovisionedHosts);
             }
         }
     }
 
-    private void deprovisionHosts(String awsRegion, CloudEvent event) {
-        for (var zone : zonesByCloudNativeRegion.get(awsRegion)) {
-            nodeRepository.list(zone.getId())
-                    .stream()
-                    .filter(shouldDeprovisionHost(event))
-                    .map(node -> node.hostname().value())
-                    .forEach(hostname -> {
-                        log.info(String.format("Setting host %s to wantToRetire and wantToDeprovision", hostname));
-                        nodeRepository.retireAndDeprovision(zone.getId(), hostname);
-                        event.affectedInstances.removeIf(hostname::contains);
-                    });
-        }
+    private List<String> deprovisionHosts(String awsRegion, CloudEvent event) {
+        return zonesByCloudNativeRegion.get(awsRegion)
+                .stream()
+                .flatMap(zone ->
+                    nodeRepository.list(zone.getId())
+                            .stream()
+                            .filter(shouldDeprovisionHost(event))
+                            .map(node -> {
+                                if (!node.wantToDeprovision() || !node.wantToRetire())
+                                    log.info(String.format("Setting host %s to wantToRetire and wantToDeprovision", node.hostname().value()));
+                                    nodeRepository.retireAndDeprovision(zone.getId(), node.hostname().value());
+                                return node.hostname().value();
+                            })
+                )
+                .collect(Collectors.toList());
     }
 
-    private void submitIssue(CloudEvent event) {
-        if (event.affectedInstances.isEmpty())
+    private void submitIssue(CloudEvent event, List<String> deprovisionedHosts) {
+        if (event.affectedInstances.size() == deprovisionedHosts.size())
             return;
         Issue issue = eventFetcher.createIssue(event);
         if (!issueHandler.issueExists(issue)) {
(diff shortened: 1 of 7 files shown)


andreer (approved):
LGTM, but I suggest getting an additional reviewer on this to be on the safe side 👍

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
```diff
@@ -27,33 +32,67 @@
...
+                    .forEach(hostname -> {
+                        log.info(String.format("Setting host %s to wantToRetire and wantToDeprovision", hostname));
+                        nodeRepository.retireAndDeprovision(zone.getId(), hostname);
+                        event.affectedInstances.removeIf(hostname::contains);
+                    });
+        }
+    }
+
+    private void submitIssue(CloudEvent event) {
+        if (event.affectedInstances.isEmpty())
```
freva: If this is possible, this needs to be moved to the parent method so we can skip `deprovisionHosts()` as well.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
```diff
@@ -27,33 +32,67 @@
...
 
+    private void deprovisionHosts(String awsRegion, CloudEvent event) {
+        for (var zone : zonesByCloudNativeRegion.get(awsRegion)) {
+            nodeRepository.list(zone.getId())
+                    .stream()
+                    .filter(shouldDeprovisionHost(event))
+                    .map(node -> node.hostname().value())
+                    .forEach(hostname -> {
+                        log.info(String.format("Setting host %s to wantToRetire and wantToDeprovision", hostname));
+                        nodeRepository.retireAndDeprovision(zone.getId(), hostname);
```
freva: Since you are iterating over multiple zones, I assume this will throw if you try to do this on the wrong zone?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
```diff
@@ -27,33 +32,67 @@
...
-                    log.log(Level.INFO, String.format("Filed an issue with the title '%s'", issue.summary()));
-                }
+                deprovisionHosts(awsRegion, event);
+                submitIssue(event);
             }
         }
     }
 
+    private void deprovisionHosts(String awsRegion, CloudEvent event) {
+        for (var zone : zonesByCloudNativeRegion.get(awsRegion)) {
```
freva: Instance name contains zone name in it, so this is not really needed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
```diff
@@ -27,33 +32,67 @@
...
+        if (event.affectedInstances.isEmpty())
+            return;
+        Issue issue = eventFetcher.createIssue(event);
+        if (!issueHandler.issueExists(issue)) {
+            issueHandler.file(issue);
+            log.log(Level.INFO, String.format("Filed an issue with the title '%s'", issue.summary()));
+        }
+    }
+
+    private Predicate<Node> shouldDeprovisionHost(CloudEvent event) {
```
freva: Also check if the node already has `wantToDeprovision` and `wantToRetire` set already.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporterTest.java
```diff
@@ -0,0 +1,161 @@
...
+        equalListsIgnoreOrder(List.of("host4.com", "confighost.com"), getHostnames(awsZone2.getId()));
+
+        Map<IssueId, MockIssueHandler.MockIssue> createdIssues = tester.serviceRegistry().issueHandler().issues();
+        assertEquals(1, createdIssues.size());
+        String description = createdIssues.get(IssueId.from("1")).issue().description();
+        assertEquals("[confighost]", description);
+
+    }
+
+    private void equalListsIgnoreOrder(List<?> l1, List<?> l2) {
```
freva: Consider returning `Set` from `getHostnames()` and use `assertEquals(Set.of(..), getHostnames(...));`

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
olaaun: The idea was to remove instances from affectedInstances if they were deprovisioned. I can make it more clear, by letting `deprovisionHosts` return the list of deprovisioned hosts, and skip ticket submission if `event.affectedInstances.size() == deprovisionedHosts.size()`

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
olaaun: If this is done in the wrong zone, all hosts should've been filtered away by `shouldDeprovisionHost`

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudEventReporter.java
freva: Sounds good. Event should be immutable.

freva (approved):

PR #11688, opened 2020-01-07
Fix format strings.

toregge:
@baldersheim : please review


baldersheim (approved):

baldersheim:
BTW, where is this an issue. I thought %z was standard now.
@vekterli FYI

vekterli:
`%z` is for `size_t`, which should be == `uint64_t` on all platforms we'd ever care about, but I suppose _technically_ it's not guaranteed to match. So a compiler may choose to complain at us.

Luckily, soon the functionality found in [fmtlib](https://github.com/fmtlib/fmt) will be part of the standard library so we can ditch all of these funny characters altogether! 😎

PR #11696, opened 2020-01-08
Minor cleanup/simplification, no functional changes

hmusum:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/StorageGroup.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/StorageGroup.java
@@ -412,13 +412,16 @@ public class StorageGroup {
         }
 
         private Optional<String> childAsString(Optional<ModelElement> element, String childTagName) {
-            return element.map(modelElement -> modelElement.childAsString(childTagName));
+            if (element.isEmpty()) return Optional.empty();
+            return Optional.ofNullable(element.get().childAsString(childTagName));
         }
         private Optional<Long> childAsLong(Optional<ModelElement> element, String childTagName) {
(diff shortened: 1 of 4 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/StorageGroup.java
```diff
@@ -412,21 +412,17 @@ else if (! nodesElement.isPresent() && subGroups.isEmpty() && context.getDeployS
         }
 
         private Optional<String> childAsString(Optional<ModelElement> element, String childTagName) {
-            if ( ! element.isPresent()) return Optional.empty();
-            return Optional.ofNullable(element.get().childAsString(childTagName));
+            return element.map(modelElement -> modelElement.childAsString(childTagName));
```
jonmv: `element.get().childAsString(childTagName)` is never `null`?

jonmv (approved):
One question :)

config-model/src/main/java/com/yahoo/vespa/model/content/StorageGroup.java
hmusum: Hm, nice catch. Shouldn't always trust the IDE's suggested changes it seems. Will fix.

jonmv: Hmm, `.map(...)` handles the result being null. I don't think it did before?

jonmv: Bah, seems it always did :p

jonmv: Well, learned something new today, then :)

PR #11699, opened 2020-01-08
Add option to use default signature for onnx models

lesters:
@jonmv Please review. Avoids class loading issue in https://github.com/vespa-engine/vespa/pull/11594


jonmv (approved):
OK :)

PR #11706, opened 2020-01-08
Avoid checking every node when setting upgrade target

mpolden:


jonmv (approved):
Very nice!

jonmv:
This broke convergence in main, as application package for routing in certain zones does not exist. Previously, this was avoided by the `convergedOn` returning `true` when the application owned no nodes in the zone, but this check was removed here, causing `upgrade` to be called, and fail, for routing in these zones.

mpolden:
> this was avoided by the convergedOn returning true when the application owned no nodes in the zone

Sounds like there should be a test for this then. 🙂

PR #11718, opened 2020-01-09
Warn on missing query feature type declarations

bratseth:
Warn on	missing	query feature type declarations in rank profiles
which use tensors. This may help users to remember to declare them
to avoid running into confuson as tensors are interpreted as
string which are hashed to scalars.


lesters (approved):
LGTM!

PR #11722, opened 2020-01-09
Avoid re-reading version status for every application

mpolden:
Calling `systemVersion()` implies reading complete version status from
ZooKeeper, and doing this for every single app is naturally expensive.


jonmv (approved):
Thanks!

PR #11723, opened 2020-01-09
Treat rank features as tensors (or doubles)

bratseth:
The purpose of this is to prepare to disallow rank feature
strings (which can't be parsed to doubles or tensors)
on Vespa 8. Currently, such strings will be converted to
numbers by hashing them during evaluation, but this is not
useful (as the hash can be computed before setting if desired),
and leads to confusion when a feature is intended set from
a tensor string but ends up as a hash due to missing type
information in the configuration.

Changes:
- Parse numeric rank features into doubles as soon as possible
- Deprecate the methods accessing rank features as strings
- Allow double features to be accessible as tensors


lesters (approved):
LGTM!

Issue #6516, opened 2018-08-07
Error constructing FilterBindingsProvider after FilterChainRepository removes empty filter chain.


paolodedios:
After deploying an application with an access control filter, the FilterChainRepository removes an empty filter chain called 'access-control-chain' but then fails to instantiate FilterBindingsProvider due to missing filter chain with id 'access-control-chain'

Below is a snippet of the error message.

```
[2018-08-07 21:06:50.276] INFO    : container        Container.com.yahoo.container.core.config.HandlersConfigurerDi     Installing bundles from the latest application
[2018-08-07 21:06:50.277] INFO    : container        Container.com.yahoo.container.core.config.BundleLoader     0 bundles were removed, and 0 bundles were installed.
[...]

PR #11725, opened 2020-01-09
Fail if using 'access-control' in non-hosted context

bjorncs:
Fixes error reported in https://github.com/vespa-engine/vespa/issues/6516.

### DIFF
--- a/config-model/src/test/java/com/yahoo/vespa/model/container/xml/AccessControlTest.java
+++ b/config-model/src/test/java/com/yahoo/vespa/model/container/xml/AccessControlTest.java
@@ -45,6 +45,7 @@ public class AccessControlTest extends ContainerModelBuilderTestBase {
             ContainerCluster.ROOT_HANDLER_PATH);
 
     public AccessControlTest() {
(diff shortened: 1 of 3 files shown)


config-model/src/test/java/com/yahoo/vespa/model/container/xml/AccessControlTest.java
```diff
@@ -44,6 +44,10 @@
             StateHandler.STATE_API_ROOT,
             ContainerCluster.ROOT_HANDLER_PATH);
 
+    public AccessControlTest() {
```
gjoranv: Please add a comment explaining why Before is not used, e.g. "`isHosted`must be initialized before the Before method of the super class".

gjoranv (approved):

config-model/src/test/java/com/yahoo/vespa/model/container/xml/AccessControlTest.java
bjorncs: Fixed

PR #11737, opened 2020-01-10
Produce JDK 8 compatible object files for Vespa clients.

aressem:


baldersheim (commented):
Keep in mind that we only should not build java 8 for public.

PR #11740, opened 2020-01-10
Add resolving of input types for stateless model evaluation

lesters:
@bratseth Please take a look.

With this, you can see for instance with the `ModelEvaluationTest.testMl_serving` test that the `Placeholder` model input is not resolved properly in `MapEvaluationTypeContext.resolveType`.


bratseth:
Should be ok now @lesters

bratseth (approved):

lesters:
@bratseth Thanks!

PR #11744, opened 2020-01-10
Make sure to close file handler used for zookeeper log messages

hmusum:
This avoid .lck files being left behind after stopping container


hakonhall (approved):
Although it would mean that if the JVM crashes, a lck file would be left and not removed?

hmusum:
Yes, good point, I have added back removal of leftover .lck files

hakonhall (approved):

PR #11748, opened 2020-01-10
Optimize handling of "matched-elements-only" when the input field as …

geirst:
…retrieved from the document instance.

In this case we now filter on matching elements while converting to slime,
instead of converting to slime and then filter as done before.

@baldersheim please review
@toregge FYI


baldersheim (approved):
Looks very nice to me.

PR #11755, opened 2020-01-13
Support modify of mixed tensors

bratseth:
It should be possible to modify a tensor field with a mixed type using the "blocks" short form.

### DIFF
--- a/document/src/main/java/com/yahoo/document/update/TensorModifyUpdate.java
+++ b/document/src/main/java/com/yahoo/document/update/TensorModifyUpdate.java
@@ -15,7 +15,7 @@ import java.util.function.DoubleBinaryOperator;
 /*
  *  An update for the subset of the cells in a tensor.
  *
- *  The cells to update are contained in a sparse tensor (has only mapped dimensions).
+ *  The cells to update are contained as entries of the sparse dimension(s) of the tensor.
  */
 public class TensorModifyUpdate extends ValueUpdate<TensorFieldValue> {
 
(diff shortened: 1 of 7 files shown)


document/src/main/java/com/yahoo/document/update/TensorModifyUpdate.java
```diff
@@ -29,8 +30,8 @@ public TensorModifyUpdate(Operation operation, TensorFieldValue tensor) {
     }
 
     private void verifyCompatibleType(TensorType type) {
-        if (type.dimensions().stream().anyMatch(dim -> dim.isIndexed()) ) {
-            throw new IllegalArgumentException("Tensor type '" + type + "' is not compatible as it contains some indexed dimensions");
+        if (type.rank() > 0 && type.dimensions().stream().noneMatch(dim -> dim.isMapped()) ) {
```
geirst: Class comment should be updated now that operation is supported for mixed tensors.

geirst (approved):
👍

document/src/main/java/com/yahoo/document/update/TensorModifyUpdate.java
geirst: This will break how we handle such updates in backend. We treat the cells to update as a sparse tensor and assumes this has been verified before serialized. To support the blocked syntax, the resulting tensor must still be a sparse one. Also see tensor_modify_update_treats_the_input_tensor_as_sparse().

bratseth: Okay, let's convert in this layer for now https://github.com/vespa-engine/vespa/pull/11766

PR #11757, opened 2020-01-13
Avoid checking every node when setting upgrade target

mpolden:
Same as #11706, with one additional commit.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/SystemUpgrader.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/SystemUpgrader.java
@@ -30,12 +30,8 @@ public class SystemUpgrader extends InfrastructureUpgrader {
 
     @Override
     protected void upgrade(Version target, SystemApplication application, ZoneApi zone) {
-        // TODO(mpolden): Simplify this by comparing with version from NodeRepository#targetVersionsOf instead
-        if (minVersion(zone, application, Node::wantedVersion).map(target::isAfter)
-                                                              .orElse(true)) {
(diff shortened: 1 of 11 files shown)


mpolden:
Test failure (`SearchClusterTest.requireThatZeroDocsAreFine:177`) is unrelated. Unstable test?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/SystemUpgrader.java
```diff
@@ -63,8 +63,14 @@ protected boolean shouldUpgrade(Version target, SystemApplication application, Z
...
-                    .map(target::isAfter)                                     // Upgrade if target is after any wanted version
-                    .orElse(true);                                            // Upgrade if there are no nodes allocated
+                    // Upgrade if target is after any wanted version
+                    .map(target::isAfter)
+                    // Skip upgrade if there are no nodes allocated. This is overloaded to mean that the zone is not
+                    // expected to have a deployment of this application.
+                    // TODO(mpolden): Once all zones are either directly routed or not: Change this to
+                    //                always deploy proxy app and wait for convergence in zones that are not directly
+                    //                routed.
+                    .orElse(false);
```
jonmv: I guess this would make it impossible to set up a new zone with a routing app, unless this was manually deployed first to give the application some nodes ... ? But that probably doesn't matter.

jonmv (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/SystemUpgrader.java
mpolden: Would probably need a manual deployment first yes, but it's the same behaviour as before.

PR #11760, opened 2020-01-13
Test validation of the blog TF model

bratseth:


lesters (commented):
The blog post model has been added to `config-model` here, but the `BlogEvaluationTestCase` here is in `model-integration`, which already has this model under `src/test/models`. Intentional?

bratseth:
Yes, loading and validation of the model is tested just by adding it.

lesters (approved):
:+1:

Issue #11686, opened 2020-01-07
Regression introduced in Vespa 7.141 may cause data loss or inconsistencies when using 'create: true' updates


vekterli:
### High-level summary

There exists a regression introduced in Vespa 7.141 where updates marked as `create: true` (i.e. create if missing) may cause **data loss or undetected inconsistencies** in certain edge cases. This regression was introduced as part of an optimization effort to greatly reduce the common-case overhead of updates when replicas are out of sync (see #3703, #11319).

The known edge cases triggering this bug are:
- An update is sent towards a document that is entirely missing on one more more replicas. This may happen e.g. during data migration caused by node failures or cluster topology changes.
- An update is sent towards a document whose replica set changes between the time the update is started on a distributor and the time update operations are sent.

It is _not_ believed that updates _without_ `create: true` are impacted in practice by this bug.

### Impact
When this bug is triggered, the most common outcome is that the document version resulting from the update is as if the document never existed in the first place. I.e. **existing document data will be discarded**. If the distributor process is terminated before the update completes it is also possible (though far less likely) that the documents are left in an undetectable state of inconsistency.

A warning of the following form should be emitted in the Vespa logs on the content nodes if the bug is triggered:
```
Update operation for 'id:foo::bar' in bucket Bucket(BucketSpace(0x0000000000000001), BucketId(0x4000000000f00baa)) updated documents with different timestamps. This should not happen and may indicate undetected replica divergence. Found ts=0 on node 5, ts=1576000000000001 on node 6
```
Where `id:foo::bar` is the ID of the affected document. Note that this log message is emitted _after_ the update has been applied on the replicas, so it's possible for such messages to be missed if the distributor process is terminated before it can be printed.

Such log warnings may be observed even when not using `create: true`, but since such updates cannot implicitly create new documents from scratch on replicas, it is not believed that this actually triggers an inconsistency or data loss.

### Status
All known issues are fixed on HEAD (#11561, #11613), and the optimizations have additionally been temporarily entirely disabled (#11612) until we have further increased test coverage. Will update this issue once we have an Open-Source RPM release containing the fixes.

Until then, anyone using updates with `create: true` may add the following config override to their `services.xml` file and redeploy the application:
```xml
<config name="vespa.config.content.core.stor-distributormanager">
<restart_with_fast_update_path_if_all_get_timestamps_are_consistent>false</restart_with_fast_update_path_if_all_get_timestamps_are_consistent>
</config>
```
[...]

PR #11761, opened 2020-01-13
Avoid inconsistent auto-created document versions taking precedence 

vekterli:
@geirst please review

Create-if-missing updates have a rather finicky behavior in the backend,
wherein they'll set the timestamp of the previous document to that of the
_new_ document timestamp if the update ended up creating a document from
scratch. This particular behavior confuses the "after the fact" timestamp
consistency checks, since it will seem like the document that was created
from scratch is a better candidate to force convergence towards rather
than the ones that actually updated an existing document.

With this change we therefore detect this case specially and treat the
received timestamps as if the document updated had a timestamp of zero.
This matches the behavior of regular (non auto-create) updates.

Note that another venue for solving this would be to alter the returned
timestamp in the backend to be zero instead, but this would cause issues
during rolling upgrades since some of the content nodes would be returning
zero timestamps while others would be returning non-zero. This would in
turn trigger false positives for the inconsistency sanity checks.

Also note that this is a fallback path that should not be hit unless
the a-priori inconsistency checks in the two-phase update operation
somehow fails to recognize that the document versions may be out of
sync.

This relates to issue #11686

### DIFF
--- a/storage/src/tests/distributor/updateoperationtest.cpp
+++ b/storage/src/tests/distributor/updateoperationtest.cpp
@@ -14,8 +14,6 @@
 #include <vespa/vespalib/gtest/gtest.h>
 
 using namespace document;
-using namespace storage;
-using namespace storage::distributor;
 using namespace storage::api;
 using namespace std;
 using namespace storage::lib;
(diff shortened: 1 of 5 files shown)


vekterli:
Unrelated CI failure:
```
[ERROR]   ModelEvaluationTest.testMl_serving:66->assertHasMlModels:110 expected:<...lt.y).type: tensor(d[]1[10])
> but was:<...lt.y).type: tensor(d[0[],d]1[10])
```

geirst (approved):
👍

PR #11783, opened 2020-01-14
Add vds.datastored.bucket_space.buckets_total as a fleetcontroller consumer for storage cluster to expose metric in host info.

hakonhall:


vekterli (approved):
:+1:

PR #11784, opened 2020-01-14
Add dummy objects for new deployments to list in instance response.

jonmv:
@freva please review. 
@smorgrav @oyving FYI.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -954,6 +954,8 @@ public class ApplicationApiHandler extends LoggingRequestHandler {
                 if (!instance.rotations().isEmpty()) {
                     toSlime(instance.rotations(), instance.rotationStatus(), deployment, deploymentObject);
                 }
+
+
             }
 
             if (recurseOverDeployments(request)) // List full deployment information when recursive.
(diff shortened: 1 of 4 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -968,6 +970,17 @@ private void toSlime(Cursor object, Instance instance, DeploymentStatus status,
...
         }
+        // Add dummy values for not-yet-existent prod deployments.
+        status.jobSteps().keySet().stream()
+              .filter(job -> job.application().instance().equals(instance.name()))
+              .filter(job -> job.type().isProduction() && job.type().isDeployment())
+              .map(job -> job.type().zone(controller.system()))
+              .filter(zone -> ! instance.deployments().containsKey(zone))
+              .forEach(zone -> {
+                    Cursor deploymentObject = instancesArray.addObject();
+                    deploymentObject.setString("environment", zone.environment().value());
```
freva: Both console and dashboard expect that this object also contains `tenant`, `application` and `instance`.
Dashboard additionally expects that `url` is set.

jonmv: Certainly won't set URL, as that would just point to a 404. So I can set the application ID, but this is surely pointless, as all three values are set on the grandparent object (the instance object container the array of deployments). I also intended for the missing `url` to serve as the indication that this is a dummy object.

jonmv: Also, It seems to be the application id (tenant, application, instance) isn't set, unless it is a "recursive request"?

jonmv (commented):
Urghj.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
freva: Ah, yes, you're right. Let me fix the clients to make sure they can handle this change first (so they don't make the request that'll lead to 404).

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
jonmv: 🙇

PR #11790, opened 2020-01-15
Update cluster spec for reserved nodes

freva:
Previously we would re-reserve `reserved` nodes from node-repo which would only update the reserved history time to extend the reservation period. This however didn't include potential updates done to these nodes in `NodeAllocation`, such as updating the `ClusterSpec` if it has changed: https://github.com/vespa-engine/vespa/blob/e42294b01c863ca4ba0986f3e52f568dd746f797/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java#L250


freva:
FYI: @hmusum

bratseth (approved):
👍

PR #11796, opened 2020-01-15
Athenz domain cannot be set on instance level, just on the root

jonmv:
@bratseth or @tokle please review and merge.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
@@ -155,7 +155,7 @@ public class DeploymentInstanceSpec extends DeploymentSpec.Steps {
     }
 
     /** Returns the athenz domain if configured */
-    // TODO jonmv: Remove when the commit with this comment is deployed on all config servers.
+    // TODO jonmv: Remove when 7.162 is older than the oldest deployed version.
     public Optional<AthenzDomain> athenzDomain() { return Optional.empty(); }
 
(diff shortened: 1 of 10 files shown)


config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
```diff
@@ -182,9 +155,10 @@ public boolean canChangeRevisionAt(Instant instant) {
     }
 
     /** Returns the athenz domain if configured */
-    public Optional<AthenzDomain> athenzDomain() { return athenzDomain; }
+    // TODO jonmv: Remove when the commit with this comment is deployed on all config servers.
```
bratseth: That's not sufficient because it is accessed from the config model. Better to give a date where you can be sure all config models accessing it are gone, without requiring the random person seeing this in the future to do detective work, e.g "After February 2020".

bratseth (changes requested):
Otherwise LGTM!

tokle (approved):

config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
jonmv: Of course. I have no idea what I was thinking :)

jonmv:
```suggestion
    // TODO jonmv: Remove when 7.162 is older than the oldest deployed version.
```

PR #11807, opened 2020-01-16
Jvenstad/cap log sizes during deployment jobs

jonmv:


freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/BufferedLogStore.java
```diff
@@ -24,13 +25,22 @@
  */
 public class BufferedLogStore {
 
-    static final int chunkSize = 1 << 13;
+    private static final int defaultChunkSize = 1 << 13; // 8kb per node stored in ZK.
+    private static final int defaultMaxLogSize = 1 << 16; // 64Mb max per run.
```
jonmv: Hehe, This is 64kb >_<

PR #11812, opened 2020-01-16
Hmusum/cleanup log fetching

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
@@ -1,0 +1,36 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.config.server.http;
+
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.http.Header;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Optional;
+
+/**
+ * Proxies response back to client, keeps Content-Type header if it is present
+ *
+ * @author Ola Aunrønning
```
olaaun: Can't really remember, but freva might've been the author of `ProxyResponse`

olaaun (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
hmusum: Seems like you both have done work there, I'll add freva as author in next PR

PR #11814, opened 2020-01-16
Refuse to start container if resolved IP address does not match node-repo IP

freva:


node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integrationTests/DockerTester.java
```diff
@@ -75,8 +74,7 @@
         ipAddresses.addAddress(HOST_HOSTNAME.value(), "f000::");
         for (int i = 1; i < 4; i++) ipAddresses.addAddress("host" + i + ".test.yahoo.com", "f000::" + i);
 
-        ProcessExecuter processExecuter = mock(ProcessExecuter.class);
-        uncheck(() -> when(processExecuter.exec(any(String[].class))).thenReturn(new Pair<>(0, "")));
+        TerminalImpl terminal = new TerminalImpl(command -> new TestChildProcess2(0, ""));
```
hakonhall: Why not use TestTerminal?

freva: This is basically "integration test" for node-admin: it runs the entire node-admin, starts multiple containers which run regular ticks. There is no way to list all the expected commands here...

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/DockerOperationsImpl.java
```diff
@@ -193,31 +212,17 @@ public ProcessResult executeCommandInContainerAsRoot(NodeAgentContext context, S
...
-            }
-            return new ProcessResult(0, result.getSecond(), "");
-        } catch (IOException e) {
-            throw new RuntimeException(String.format("IOException while executing %s in network namespace for %s (PID = %d)",
-                    Arrays.toString(wrappedCommand), context.containerName().asString(), containerPid), e);
-        }
+        return terminal.newCommandLine(context)
+                .add("nsenter", String.format("--net=/proc/%d/ns/net", containerPid), "--")
+                .add(command)
+                .executeSilently();
```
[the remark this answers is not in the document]
freva: Not really practical at this level - this method is executed every tick to for example list the current rules to see if a modification is needed.

node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integrationTests/DockerTester.java
hakonhall: Yes, TestTerminal needs a method that returns a given exit code & response forever. If it's non-trivial it's OK to leave as-is.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/DockerOperationsImpl.java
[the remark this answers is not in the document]
hakonhall: got it

hakonhall (approved):

PR #11815, opened 2020-01-16
Hold a JVM-wide reentrant lock to grab mutex — helps ZK stale reads?

jonmv:

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/Lock.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/Lock.java
@@ -28,18 +28,18 @@ public class Lock implements Mutex {
 
     /** Take the lock with the given timeout. This may be called multiple times from the same thread - each matched by a close */
     public void acquire(Duration timeout) throws UncheckedTimeoutException {
-        boolean acquired;
+        boolean acquired = false;
         try {
-            lock.tryLock(timeout.toMillis(), TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/Lock.java
```diff
@@ -15,31 +16,46 @@
...
             acquired = mutex.acquire(timeout.toMillis(), TimeUnit.MILLISECONDS);
+            lock.tryLock(); // Should be available to only this thread, while holding the above mutex.
         }
         catch (Exception e) {
+            if (acquired) release();
             throw new RuntimeException("Exception acquiring lock '" + lockPath + "'", e);
         }
 
-        if (! acquired)
+        if ( ! lock.isHeldByCurrentThread()) {
```
hakonhall: I found the previous version of this code easer to read, probably because of the implicit non-obvious assumption that tryLock() is assumed to always return true, and then relying on that here.

hakonhall (approved):

zkfacade/src/main/java/com/yahoo/vespa/curator/Lock.java
jonmv: Right. Fixed. PTAL.

hakonhall (approved):

jonmv:
May hold off merging this until I've investigated a little further.

PR #11818, opened 2020-01-16
Reboot on memory change

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -205,15 +209,18 @@ private void startContainer(NodeAgentContext context) {
...
                 return Optional.empty();
             }
 
             shouldRestartServices(context, existingContainer.get()).ifPresent(restartReason -> {
                 context.log(logger, "Will restart services: " + restartReason);
-                restartServices(context, existingContainer.get());
+                orchestratorSuspendNode(context);
+
+                dockerOperations.restartVespa(context);
+                context.log(logger, "Restarted services");
```
hakonhall: there is already a log() above so this can be removed.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -286,31 +277,40 @@ public void suspend(NodeAgentContext context) {
...
+        if (containerState == STARTING)
+            reasons.add("Container failed to start");
+
+        return reasons;
     }
 
-    private void removeContainer(NodeAgentContext context, Container existingContainer, String reason, boolean alreadySuspended) {
-        context.log(logger, "Will remove container: " + reason);
+    private void removeContainer(NodeAgentContext context, Container existingContainer, List<String> reasons, boolean alreadySuspended) {
+        context.log(logger, "Will remove container: " + (reasons.size() == 1 ? reasons.get(0) : "\n" + String.join("\n", reasons)));
```
hakonhall: I'd vote for comma-separated rather than newline, as the latter is not so good with grep and is visually extravagant.

hakonhall (approved):

PR #11824, opened 2020-01-16
Only consider non-retired node when determining number of groups/cluster size

freva:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
@@ -60,7 +60,8 @@ public class NodePrioritizer {
         this.spareHosts = findSpareHosts(allNodes, capacity, spares);
 
         NodeList nodesInCluster = allNodes.owner(appId).type(clusterSpec.type()).cluster(clusterSpec.id());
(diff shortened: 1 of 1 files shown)


yngveaasheim:
@smorgrav  I have to defer the approval of this PR to you. The rationale here is to allow in-place node resource allocation changes even in the case where there are retiring nodes in the cluster, without the deployment handling see this as a reduction in the node count for the cluster. I guess this boils down to the potentially changes semantics of 'currentClusterSize' being fine.

freva:
@smorgrav These values are only used to determine if it is safe to perform inplace resize as per these rules: https://github.com/vespa-engine/vespa/blob/ca80f4634d471c078047337e1d10a7f73c163900/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/InPlaceResizeProvisionTest.java#L32:L45

The specific case that we try to remediate here is
1. User performs a resource increase that is only successful on `x` of `n` nodes where `x < n`. This leads to the application having `n + n - x` nodes, where `n - x` are retired.
2. The user reverts the change, logically this should be possible to do in-place as all of the `n + n - x` nodes have enough resources to satisfy the new request, but because the cluster size is seemingly reduced from `n + n - x` to `n`, no in-place resize can be done and the user instead receives `n` new nodes, ending up with a total of `3n -x `.

smorgrav:
LGTM

smorgrav (approved):

PR #11830, opened 2020-01-17
Vekterli/support weak internal read consistency for client gets

vekterli:
@geirst please review
@baldersheim FYI

If configured, Get operations initiated by the client are flagged
with weak internal consistency. This allows the backend to bypass
certain internal synchronization mechanisms, which minimizes latency
at the cost of possibly not observing a consistent view of the
document. This config should only be used in a very restricted set
of cases where the document set is effectively read-only, or cross-
field consistency or freshness does not matter.

To enable the weak consistency, use an explicit config override:

```
<config name="vespa.config.content.core.stor-distributormanager">
  <use_weak_internal_read_consistency_for_client_gets>
    true
  </use_weak_internal_read_consistency_for_client_gets>
</config>
```

### DIFF
--- a/storage/src/tests/distributor/distributortest.cpp
+++ b/storage/src/tests/distributor/distributortest.cpp
@@ -187,6 +187,12 @@ struct DistributorTest : Test, DistributorTestUtil {
         configureDistributor(builder);
     }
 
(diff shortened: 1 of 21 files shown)


geirst (approved):
👍

PR #11831, opened 2020-01-17
Jvenstad/staging testup

jonmv:
@hakonhall please review.

Large diff on two test files is mostly a change to let the test files have the same ordering of elements as the rendered JSON.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
```diff
@@ -258,6 +262,8 @@ static Step stepOf(String step) {
             case "installTester"      : return installTester;
             case "deactivateTester"   : return deactivateTester;
             case "copyVespaLogs"      : return copyVespaLogs;
+            case "startStagingSetup"  : return startStagingSetup;
```
hakonhall: These changes must roll out before any changes that writes such steps to ZK?

jonmv: Well, a non-upgraded controller will be unable to read (to modify or display) the run status for any staging test that has been started by an upgraded controller, until it upgrades itself. I don't think that's a major consequence.

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
jonmv: Of course you’re right it’s the right thing to do, but I’m impatient and I don’t think it’ll be an actual problem here :P

PR #11835, opened 2020-01-17
Use bucket_space metric in retirement

hakonhall:
This makes the Cluster Controller use the
vds.datastored.bucket_space.buckets_total, dimension bucketSpace=default, to
determine whether a content node manages zero buckets, and if so, will allow
the node to go permanently down. This is used when a node is retiring, and it
is to be removed from the application.

The change is guarded by the use-bucket-space-metric, default true. If the new
metric doesn't work as expected, we can revert to using the current/old metric
by flipping the flag. The flag can be controlled per application.

### DIFF
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
@@ -59,9 +59,14 @@ public class HostInfoTest {
         assertThat(metrics.get(3).getName(), equalTo("vds.datastored.bucket_space.buckets_total"));
         assertThat(hostInfo.getClusterStateVersionOrNull(), is(123));
 
-        Optional<Metrics.Value> value = hostInfo.getMetrics()
-                .getValueAt("vds.datastored.bucket_space.buckets_total", Map.of("bucketSpace", "default"));
-        assertThat(value.map(Metrics.Value::getLast), equalTo(Optional.of(129L)));
+        assertThat(hostInfo.getMetrics()
+                        .getValueAt("vds.datastored.bucket_space.buckets_total", Map.of("bucketSpace", "default"))
+                        .map(Metrics.Value::getLast),
+                equalTo(Optional.of(129L)));
+        assertThat(hostInfo.getMetrics()
+                        .getValueAt("vds.datastored.bucket_space.buckets_total", Map.of("bucketSpace", "global"))
+                        .map(Metrics.Value::getLast),
+                equalTo(Optional.of(0L)));
     }
 
     @Test
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
@@ -10,12 +10,15 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.TreeMap;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.nullValue;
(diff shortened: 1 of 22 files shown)


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
```diff
@@ -56,7 +56,7 @@ protected void setUp(boolean dontInitializeNode2) throws Exception {
             Set<ConfiguredNode> nodesInSlobrok = FleetControllerTest.toNodes(1, 3, 5, 7);
 
             ContentCluster cluster = new ContentCluster(
-                    "music", nodes, distribution, 4 /* minStorageNodesUp*/, 0.0 /* minRatioOfStorageNodesUp */);
+                    "music", nodes, distribution, 4 /* minStorageNodesUp*/, 0.0, /* minRatioOfStorageNodesUp */true);
```
vekterli: Consider moving the comment (or comma) here and above to make the inline comment point at the right parameter

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -83,6 +83,13 @@
             "Takes effect on next node agent tick. Change is orchestrated, but does NOT require container restart",
             HOSTNAME, APPLICATION_ID);
 
+    public static final UnboundBooleanFlag USE_BUCKET_SPACE_METRIC = defineFeatureFlag(
+            "use-bucket-space-metric", true,
+            "Whether to use vds.datastored.bucket_space.buckets_total (true) instead of " +
+            "vds.datastored.alldisks.buckets (false, legacy).",
+            "Takes efefct on the next deployment of the application",
```
vekterli: Typo: s/efefct/effect/

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
```diff
@@ -49,11 +52,16 @@ public void testFullSet() throws IOException {
...
+        assertThat(metrics.size(), is(4));
+        assertThat(metrics.get(0).getValue().getLast(), is(5095L));
         assertThat(metrics.get(0).getName(), equalTo("vds.datastored.alldisks.buckets"));
+        assertThat(metrics.get(3).getValue().getLast(), is(129L));
+        assertThat(metrics.get(3).getName(), equalTo("vds.datastored.bucket_space.buckets_total"));
         assertThat(hostInfo.getClusterStateVersionOrNull(), is(123));
+
+        Optional<Metrics.Value> value = hostInfo.getMetrics()
+                .getValueAt("vds.datastored.bucket_space.buckets_total", Map.of("bucketSpace", "default"));
+        assertThat(value.map(Metrics.Value::getLast), equalTo(Optional.of(129L)));
```
vekterli: For completeness, consider also adding a check that the `global` `bucketSpace` dimension has zero buckets.

vekterli (approved):
:+1: Someone else might want to review the feature flag bits, since I'm not very knowledgeable about how they're implemented

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
hakonhall: Done

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hakonhall: Done

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/hostinfo/HostInfoTest.java
hakonhall: Done

hakonhall:
Adding @hmusum as reviewer for the flags/config model changes

hmusum (approved):
Flags and config model changes look good to me

PR #11836, opened 2020-01-17
Add bindings for requests that should be sent to tester containers

hmusum:
Add bindings to be able to send requests to /tester/v1 API on tester
containers (not fully implemented yet).


configserver/src/main/java/com/yahoo/vespa/config/server/http/TesterClient.java
```diff
@@ -0,0 +1,75 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.logging.Logger;
+
+/**
+ * @author musum
+ */
+public class TesterClient {
+
+    private final HttpClient httpClient = VespaHttpClientBuilder.create().build();
```
tokle: When using https you need to provide a tlscontext for mutual TLS

tokle (approved):

PR #11837, opened 2020-01-17
Bratseth/dont allow wrong serviceport take 2

bratseth:


bratseth:
Yes, the tester app wouldn't be deployable in the upgrade window. PTAL

bjorncs (approved):

PR #11840, opened 2020-01-17
Gjoranv/node info

gjoranv:
This makes the format from `/applicationmetrics/v1/values` equal to the future `/metrics/v2/values` api.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsNodesConfigGenerator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsNodesConfigGenerator.java
@@ -22,7 +22,7 @@ public class MetricsNodesConfigGenerator {
 
     private static MetricsNodesConfig.Node.Builder toNodeBuilder(MetricsProxyContainer container) {
         var builder = new MetricsNodesConfig.Node.Builder()
-                .nodeId(container.getHost().getConfigId())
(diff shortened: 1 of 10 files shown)


bratseth:
I don't think you should rename a field as that will break any clients of this?

gjoranv:
Probably true, because we still don't bundle def-files with the fat model. I will add it as a new field with a TODO to remove the old one when the oldest hosted version is past using it.

hmusum:
You should not rename a field like this. See https://docs.vespa.ai/documentation/cloudconfig/configapi-dev.html#def-compatibility. Essentially, you can't remove a field without a default value. 

I guess you could add a default value for the field you want to remove, wait until all users have the config with default value and then remove it.

Another caveat: You cannot remove a field if there are config overrides referring to the field you want to remove.

bratseth:
I commented on changing the web service API:
"Expose hostname, and rename nodeId -> role."

But yes, you also cannot rename (that is, remove) a config value.

gjoranv:
I thought more about this, and it is ok because there are no user configs to this config def. The config model always produces the value for the correct version. I have done this before without issues. But I will play it safe and add a temporary default before renaming.

Regarding the web API, this is only an internal API, not public.

bratseth:
It's not an internal API: https://docs.vespa.ai/documentation/reference/metrics.html

gjoranv:
That's the old API, where this field does not exist.

olaaun (approved):

PR #11842, opened 2020-01-17
accept and store json endpoint cert metadata on deploy

andreer:
also refactor from tlsSecretKeys -> several "endpoint certificate" classes


configserver/src/main/java/com/yahoo/vespa/config/server/tenant/EndpointCertificateRetriever.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public EndpointCertificateRetriever(SecretStore secretStore) {
+        this.secretStore = secretStore;
+    }
+
+    public Optional<EndpointCertificateSecrets> readEndpointCertificateSecrets(EndpointCertificateMetadata metadata) {
+        return Optional.of(readFromSecretStore(metadata));
+    }
+
+    private EndpointCertificateSecrets readFromSecretStore(EndpointCertificateMetadata endpointCertificateMetadata) {
```
tokle: We've had cases where the version of private key and cert were incompatible. Would it make sense to validate that the cert is indeed issued using the private key?

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/EndpointCertificateMetadataSerializer.java
```diff
@@ -0,0 +1,55 @@
...
+    //          such that what is serialized on version N+1 can be read by version N:
+    //          - ADDING FIELDS: Always ok
+    //          - REMOVING FIELDS: Stop reading the field first. Stop writing it on a later version.
+    //          - CHANGING THE FORMAT OF A FIELD: Don't do it bro.
+
+    private final static String keyNameField = "keyName";
+    private final static String certNameField = "certName";
+    private final static String versionField = "version";
+
+    public static void toSlime(EndpointCertificateMetadata metadata, Cursor object) {
```
tokle: Won't this violate the principle of "serialized on version N+1 can be read by version N"?

tokle (changes requested):

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/EndpointCertificateRetriever.java
andreer: Yes, I think I have a ticket on this too. Might as well add it now.

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/EndpointCertificateMetadataSerializer.java
andreer: The configserver can already read this format: https://github.com/vespa-engine/vespa/blob/master/configserver/src/test/java/com/yahoo/vespa/config/server/tenant/TlsSecretsKeysTest.java#L49-L58

We only start writing it now.

tokle (approved):

bjorncs (approved):
LGTM changes to security-utils

PR #11847, opened 2020-01-19
Optimize filter table diff

freva:
First 3 commits are non-functional changes

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/FilterTableLineEditor.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/FilterTableLineEditor.java
@@ -47,7 +47,7 @@ class FilterTableLineEditor implements LineEditor {
 
     @Override
     public List<String> onComplete() {
-        return position == wantedRules.size() ? List.of() : wantedRules.subList(position, wantedRules.size());
+        return wantedRules.subList(position, wantedRules.size());
     }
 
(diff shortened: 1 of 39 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/FilterTableLineEditor.java
```diff
@@ -29,15 +29,33 @@ static FilterTableLineEditor from(Acl acl, IPVersion ipVersion) {
...
+        // Insert the rules between position and index before index
+        List<String> toInsert = wantedRules.subList(position, index);
+        position = ++index;
+        return LineEdit.insertBefore(toInsert);
     }
 
     @Override
     public List<String> onComplete() {
-        return this.wantedRules;
+        return position == wantedRules.size() ? List.of() : wantedRules.subList(position, wantedRules.size());
```
hakonhall: You can always return `wantedRules.subList(position, wantedRules.size())`, even if `position == wantedRules.size()`.

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/FilterTableLineEditor.java
freva:
```suggestion
        return wantedRules.subList(position, wantedRules.size());
```

hakonhall (approved):

PR #11857, opened 2020-01-20
Require scheme and host as well

jonmv:
@freva


jonmv:
Urgh, no.

freva (approved):
:1st_place_medal:

PR #11860, opened 2020-01-20
Log curator state changes SUSPENDED, RECONNECTED and LOST

jonmv:
@hakonhall what do you think?

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
@@ -118,7 +118,7 @@ public class Curator implements AutoCloseable {
         if (this.curatorFramework != null) {
             validateConnectionSpec(connectionSpec);
             validateConnectionSpec(zooKeeperEnsembleConnectionSpec);
-            addFakeListener();
+            addLoggingListener();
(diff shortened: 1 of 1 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
```diff
@@ -201,7 +204,11 @@ public InterProcessLock createMutex(String lockPath) {
     // To avoid getting warning in log, see ticket 6389740
```
hakonhall: Remove/edit this? Also, rename method to not include "Fake" since it is no longer fake?

hakonhall (approved):
I'm guessing this caused log spam at the time ZK is brought down (for e.g. upgrade) or up (at start). It might be difficult to avoid such log. But it's only 1 line each time, so it seems perfectly fine AFAIK, and quite useful.

zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
jonmv: Hehe, good pointds.

jonmv:
Let's see how it goes in CD :)

PR #11865, opened 2020-01-21
Remove fail-starting-node-on-ip-mismatch flag

freva:
Enabled everywhere since yesterday, 0 issues.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupPreparer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupPreparer.java
@@ -65,9 +65,10 @@ public class GroupPreparer {
         boolean dynamicProvisioningEnabled = hostProvisioner.isPresent() && dynamicProvisioningEnabledFlag
                 .with(FetchVector.Dimension.APPLICATION_ID, application.serializedForm())
                 .value();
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupPreparer.java
```diff
@@ -65,9 +65,10 @@ public GroupPreparer(NodeRepository nodeRepository, Optional<HostProvisioner> ho
         boolean dynamicProvisioningEnabled = hostProvisioner.isPresent() && dynamicProvisioningEnabledFlag
                 .with(FetchVector.Dimension.APPLICATION_ID, application.serializedForm())
                 .value();
+        // Do not in-place resize in dynamically provisioned zones
         boolean inPlaceResizeEnabled = enableInPlaceResize
                 .with(FetchVector.Dimension.APPLICATION_ID, application.serializedForm())
-                .value();
+                .value() && !dynamicProvisioningEnabled;
```
hakonhall: Unrelated change?

freva: Yes, but I too lazy to make a separate PR for that

hakonhall (approved):

PR #11872, opened 2020-01-21
Handle expections when finding file references to delete

hmusum:
If we get an exception when getting file references to delete catch
exception and continue with other applications, as we want to make
sure this is done even in case of some (transient) failure.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -427,18 +427,17 @@ public class ApplicationRepository implements com.yahoo.config.provision.Deploye
         Set<String> fileReferencesInUse = new HashSet<>();
         // Intentionally skip applications that we for some reason do not find
         // or that we fail to get file references for (they will be retried on the next run)
-        listApplications().stream()
-                .map(this::getOptionalApplication)
-                .map(Optional::get)
-                .forEach(application -> {
-                    try {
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -426,12 +426,19 @@ public HttpResponse filedistributionStatus(ApplicationId applicationId, Duration
...
         Set<String> fileReferencesInUse = new HashSet<>();
         // Intentionally skip applications that we for some reason do not find
+        // or that we fail to get file references for (they will be retried on the next run)
         listApplications().stream()
                 .map(this::getOptionalApplication)
                 .map(Optional::get)
-                .forEach(application -> fileReferencesInUse.addAll(application.getModel().fileReferences().stream()
-                                                                           .map(FileReference::value)
-                                                                           .collect(Collectors.toSet())));
+                .forEach(application -> {
```
mpolden: I think this will still fail for the case we looked at earlier today (where `getOptionalApplication` is the method that throws, but it isn't called until `forEach` evaluates the stream calls).

How about rewriting it to a regular loop and wrap everything except `listApplications` in a try-catch?
```
for (var application : listApplications()) {
    try {
      ...
    } catch (Exception e) {
      ...
    }
}
```

hmusum: Yes, thanks, done.

mpolden (approved):
👍

PR #11878, opened 2020-01-21
Stay within your buffer.

baldersheim:
@vekterli PR


vekterli (commented):
:+1: `memcmp` would also do nicely

PR #11882, opened 2020-01-22
match both 64.00 and 64,00

kkraune:
the regexp is strictly a bit too permissive, but found it more readable to not escape the other dots ...

this makes tests run on my non-US laptop


freva (commented):
I suggest adding `Locale.ENGLISH` to https://github.com/vespa-engine/vespa/blob/8e4ea2399ecfe9939378211b81c402ad315505eb/config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/ResourcesReductionValidator.java#L63 instead

kkraune:
makes sense

PR #11888, opened 2020-01-22
Include build time of latest submission in ownership tickets

jonmv:
@mpolden please review.


mpolden (approved):
Does this need a change in internal repo?

jonmv:
You found it :)

PR #11897, opened 2020-01-22
Revert "andreer/deploy with endpoint certificate metadata"

jonmv:
Reverts vespa-engine/vespa#11893

Broke integration tests. @andreer


andreer:
There is a matching pr in other repo that must also be reverted

PR #11903, opened 2020-01-23
Implement /routing/v1/ API

mpolden:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private HttpResponse get(Path path) {
+        if (path.matches("/routing/v1/status/tenant/{tenant}/application/{application}/instance/{instance}/environment/{environment}/region/{region}")) return deploymentStatus(path);
+        if (path.matches("/routing/v1/status/environment/{environment}/region/{region}")) return zoneStatus(path);
+        return ErrorResponse.notFoundError("Nothing at " + path);
+    }
+
+    private HttpResponse setZoneStatus(Path path, boolean in) {
+        var zone = zoneFrom(path);
+        var status = in ? GlobalRouting.Status.in : GlobalRouting.Status.out;
```
jonmv: Nit: wouldn't it be more succinct if the parameter to this method was the enum constant? You could just print it(s name) in the output, as well.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        var root = slime.setObject();
+        var zonePolicy = controller.applications().routingPolicies().get(zone);
+        zoneStatusToSlime(root, zonePolicy);
+        return new SlimeJsonResponse(slime);
+    }
+
+    private HttpResponse setDeploymentStatus(Path path, boolean in) {
+        var deployment = deploymentFrom(path);
+        routingPoliciesOf(deployment);
+        var status = in ? GlobalRouting.Status.in : GlobalRouting.Status.out;
```
jonmv: Same.

jonmv (approved):
Nice!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
mpolden: Maybe, will consider it in next PR. Realised I have to resolve the TODO I added here now anyway. 😞

PR #11905, opened 2020-01-23
Propagate reservedTo

bratseth:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/AddNode.java
```diff
@@ -19,26 +20,30 @@
...
     public final Optional<FlavorOverrides> flavorOverrides;
     public final Optional<NodeResources> nodeResources;
+    public final Optional<TenantName> reservedTo;
     public final NodeType nodeType;
     public final Set<String> ipAddresses;
     public final Set<String> additionalIpAddresses;
 
-    public static AddNode forHost(String hostname, String nodeFlavor, Optional<FlavorOverrides> flavorOverrides, NodeType nodeType, Set<String> ipAddresses, Set<String> additionalIpAddresses) {
-        return new AddNode(hostname, Optional.empty(), Optional.of(nodeFlavor), flavorOverrides, Optional.empty(), nodeType, ipAddresses, additionalIpAddresses);
+    public static AddNode forHost(String hostname, String nodeFlavor, Optional<FlavorOverrides> flavorOverrides, Optional<TenantName> reservedTo, NodeType nodeType, Set<String> ipAddresses, Set<String> additionalIpAddresses) {
```
freva: This is exclusively for adding config-server/controllers...

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -600,6 +600,7 @@ private HttpResponse nodes(String tenantName, String applicationName, String ins
             Cursor nodeObject = nodesArray.addObject();
             nodeObject.setString("hostname", node.hostname().value());
             nodeObject.setString("state", valueOf(node.state()));
+            node.reservedTo().ifPresent(tenant -> nodeObject.setString("reservedTo", tenant.value()));
```
freva: This is will only ever return allocated _tenant_ nodes...

freva (approved):
LGTM, a few changes seem unnecessary...

bratseth:
Yes, I could be more accurate in supporting this only precisely in the right cases, but it would be better if we could reduce the number and length of these code paths - right now it is more work to do something like this on the client side than the server side - and that should be simpler if we keep things the same.

PR #11909, opened 2020-01-23
Bump version used when manually building Vespa

vekterli:
@toregge please review
@aressem FYI

Prior 7.0.0 version would trigger the negotiation of pre-protobuf
protocols in the content layer backend, as these were introduced
in 7.41. Bump to the latest open source RPM release at the time
of writing.


VERSION
```diff
@@ -1 +1 @@
-7.0.0
+7.164.8
```
toregge: Consider 7.164.0, to support the detection of snapshot versions (cf https://github.com/vespa-engine/system-test/pull/474)

toregge (approved):

VERSION
vekterli: Fixed

toregge (approved):

PR #11910, opened 2020-01-23
Since-timestamp on Orc suspended status

hakonhall:
- Introduce a new type of suspension status: PERMANENTLY_DOWN, which
   is set when a node is scheduled to be removed from the application.
   A normal resume call will NOT clear PERMANENTLY_DOWN.
 - Store a JSON for each suspended host: Contains status, and a since timestamp
   for the time when the node was suspended. The suspension timestamp
   is preserved when switching statuses between suspension statuses.
 - The JSON is stored in a new path, void of any zone. This means that
   we eventually can get rid of the zone-part of the application instance
   reference.

### DIFF
--- a/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
+++ b/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
@@ -2,11 +2,20 @@
 package com.yahoo.vespa.orchestrator.status;
 
 /**
- * Enumeration of the different status' a host can have.
+ * Enumeration of the different statuses a host can have.
  *
  * @author oyving
  */
 public enum HostStatus {
+    /** The services on the host is supposed to be up. */
     NO_REMARKS,
-    ALLOWED_TO_BE_DOWN;
+
+    /** The services on the host is allowed to be down. */
+    ALLOWED_TO_BE_DOWN,
+
+    /**
+     * Same as ALLOWED_TO_BE_DOWN, but in addition, it is expected
+     * the host may be removed from its application at any moment.
+     */
+    PERMANENTLY_DOWN;
 }
--- a/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
+++ b/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
@@ -8,14 +8,19 @@ package com.yahoo.vespa.orchestrator.status;
  */
 public enum HostStatus {
     /** The services on the host is supposed to be up. */
-    NO_REMARKS,
+    NO_REMARKS(false),
 
     /** The services on the host is allowed to be down. */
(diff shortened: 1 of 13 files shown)


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
```diff
@@ -247,6 +285,55 @@ public HostStatus getHostStatus(ApplicationInstanceReference applicationInstance
...
+                    })));
+        }
+
+        // For backwards compatibility we'll add HostInfos from the old hosts-allowed-down ZK path,
+        // using the creation time as the since path. The new hosts ZK path should contain a subset of
+        // the hostnames under hosts-allowed-down ZK path. Eventually these sets should be identical.
+        // Once that's true we can stop writing to hosts-allowed-down, remove this code, and all
+        // data in hosts-allowed-down can be removed.
+        Set<HostName> legacyHostsDown = hostsDownFor(application);
+        Map<HostName, HostInfo> legacyHostInfos = legacyHostsDown.stream().collect(Collectors.toMap(
```
freva: Filter for what's not already in `hostInfos`?

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
```diff
@@ -2,11 +2,20 @@
...
  *
  * @author oyving
  */
 public enum HostStatus {
+    /** The services on the host is supposed to be up. */
     NO_REMARKS,
-    ALLOWED_TO_BE_DOWN;
+
+    /** The services on the host is allowed to be down. */
+    ALLOWED_TO_BE_DOWN,
```
freva: Consider adding a method that returns true when the status is one of the down statuses to avoid doing `!= NO_REMARKS`

freva (approved):

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
```diff
@@ -45,7 +48,7 @@
     final static String HOST_STATUS_CACHE_COUNTER_PATH = "/vespa/host-status-service-cache-counter";
```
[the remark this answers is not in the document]
hakonhall: Done

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
hakonhall: Done

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostStatus.java
hakonhall: Done

freva (approved):
:100:

PR #11919, opened 2020-01-24
Unify routing control

mpolden:
Support global routing control of deployments using rotations as well. The
existing implementation of this is in internal repo and will be
removed/simplified.

FYI @oyving


jonmv (approved):
LGTM :)

PR #11925, opened 2020-01-24
avoid float-double conversion for join/mul and reduce/add

havardpe:
A shot in the dark to see if it affects the under-performing float version of un-optimized matrix-matrix multiplication.

@arnej27959 please review


arnej27959 (approved):
LGTM

PR #11932, opened 2020-01-24
Propagate all source and provider properties by default

bratseth:
This changes default behavor from only propagating native properties.
I believe that will probably be safe, and it just seems more correct.

If somebody have problems with it they can recover the old behavior
by using the strict-contracts config. That config is not mentioned
in the documentation and not touched by the model. We should remove
it on Vespa 8.


bratseth:
Test failure is unrelated.

arnej27959 (approved):
very well...

PR #11933, opened 2020-01-24
More compact convergence output

jonmv:
@hakonhall please review. There is more to come: convergence summary for install steps, and the actual adaptive timeout.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/ConvergenceSummary.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/ConvergenceSummary.java
@@ -1,0 +1,120 @@
+// Copyright 2020 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.controller.deployment;
+
+import java.util.Objects;
+
+/**
+ * Summary of node and service status during a deployment job.
+ *
+ * @author jonmv
+ */
+public class ConvergenceSummary {
+
+    private final long nodes;
+    private final long down;
+    private final long upgradingOs;
+    private final long needPlatformUpgrade;
+    private final long upgradingPlatform;
+    private final long needReboot;
+    private final long rebooting;
+    private final long needRestart;
+    private final long restarting;
+    private final long services;
+    private final long upgradingApplication;
+
+    public ConvergenceSummary(long nodes, long down, long upgradingOs, long needPlatformUpgrade, long upgradingPlatform,
+                              long needReboot, long rebooting, long needRestart, long restarting, long services, long upgradingApplication) {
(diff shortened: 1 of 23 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelperTest.java
```diff
@@ -182,6 +182,7 @@ public void testResponsesWithDirectDeployment() {
     private void compare(HttpResponse response, String expected) throws JSONException, IOException {
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         response.render(baos);
+        System.err.println(baos.toString());
```
jonmv: Oops. 🙈

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/ConvergenceSummary.java
```diff
@@ -0,0 +1,130 @@
...
+    public long upgradingOs() {
+        return upgradingOs;
+    }
+
+    /** Number of nodes in need of a platform upgrade. */
+    public long needPlatformUpgrade() {
+        return needPlatformUpgrade;
+    }
+
+    /** Number of nodes down for platform upgrade. */
```
hakonhall: "Number of suspended nodes with pending platform upgrade"? E.g. we actually don't know whether a node is suspended because it plans to upgrade.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -320,24 +346,49 @@ else if (timedOut(id, deployment.get(), endpointTimeout)) {
...
 
         if (timedOut(id, deployment.get(), installationTimeout)) {
+            logger.log(nodeList.asList().stream()
+                               .flatMap(node -> nodeDetails(node, true))
+                               .collect(toList()));
             logger.log(INFO, "Installation failed to complete within " + installationTimeout.toMinutes() + " minutes!");
             return Optional.of(installationFailed);
         }
 
-        logger.log("Installation not yet complete.");
```
hakonhall: (I liked this message in the original, as it made it crystal clear I need to wait longer. But it could be this is superfluous with the other changes...)

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
jonmv: Should be covered by the High-level status, indeed.

PR #11935, opened 2020-01-24
Count bits faster when hardware supports it.

baldersheim:
@toregge PR


bratseth:
Nit: "accel**e**rated"

toregge (approved):

PR #11939, opened 2020-01-24
Add flag for when to generate routing config for tester applications

hmusum:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -214,6 +214,13 @@ public class Flags {
             "Takes effect immediately",
             ZONE_ID, HOSTNAME);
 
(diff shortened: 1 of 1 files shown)


jonmv (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -214,6 +214,13 @@
             "Takes effect immediately",
             ZONE_ID, HOSTNAME);
 
+    public static final UnboundBooleanFlag GENERATE_ROUTING_CONFIG_FOR_TESTER_APPLICATIONS = defineFeatureFlag(
+            "generate-routing-config-for-tester-applications", true,
+            "Whether config server should generate routing config (lb-services) for tester applications",
+            "Takes effect immediately",
```
hakonhall: Normally config changes guarded by a flag would only take effect on the next (re)deploy?

hmusum: Yes, you are right, thanks. Fixed in https://github.com/vespa-engine/vespa/pull/11945

PR #11940, opened 2020-01-24
Jvenstad/adaptive deployment job timeout

jonmv:
@hakonhall please review.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
@@ -56,6 +56,10 @@ public class NodeRepositoryNode {
     private String currentOsVersion;
     @JsonProperty("wantedOsVersion")
     private String wantedOsVersion;
+    @JsonProperty("currentFirmwareCheck")
+    private Long currentFirmwareCheck;
+    @JsonProperty("wantedFirmwareCheck")
+    private Long wantedFirmwareCheck;
     @JsonProperty("failCount")
     private Integer failCount;
     @JsonProperty("environment")
@@ -335,6 +339,22 @@ public class NodeRepositoryNode {
 
     public void setWantedOsVersion(String wantedOsVersion) {
         this.wantedOsVersion = wantedOsVersion;
+    }
+
+    public Long getCurrentFirmwareCheck() {
+        return currentFirmwareCheck;
+    }
+
+    public void setCurrentFirmwareCheck(Long currentFirmwareCheck) {
+        this.currentFirmwareCheck = currentFirmwareCheck;
+    }
+
+    public Long getWantedFirmwareCheck() {
+        return wantedFirmwareCheck;
+    }
+
+    public void setWantedFirmwareCheck(Long wantedFirmwareCheck) {
+        this.wantedFirmwareCheck = wantedFirmwareCheck;
     }
 
     public Map<String, JsonNode> getReports() {
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
@@ -6,8 +6,10 @@ import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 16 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
```diff
@@ -76,6 +82,8 @@
     private NodeHistory[] history;
     @JsonProperty("allowedToBeDown")
     private Boolean allowedToBeDown;
+    @JsonProperty("suspendedSince")
```
hakonhall: Now that I see it - this needs to be called suspendedSinceMillis.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -346,20 +347,48 @@ else if (timedOut(id, deployment.get(), endpointTimeout)) {
...
             }
         }
-        controller.jobController().locked(id, lockedRun -> lockedRun.withSummary(summary));
 
-        if (timedOut(id, deployment.get(), installationTimeout)) {
+        boolean failed = false;
+
+        NodeList suspendedTooLong = nodeList.suspendedSince(controller.clock().instant().minus(installationTimeout));
+        if ( ! suspendedTooLong.isEmpty()) {
+            logger.log(INFO, "Some nodes have been suspended for more than " + installationTimeout.toMinutes() + " minutes.");
```
hakonhall: Instead of "Some nodes" - write the list or a sample node? (btw, I think you should write a util to convert a list of hosts to String, and write the full list if the number of elements is below some reasonable threshold, while otherwise write a prefix of the list and then something like ... (and X more hosts). You could use this util here.)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -346,20 +347,48 @@ else if (timedOut(id, deployment.get(), endpointTimeout)) {
...
+            if (summary.needPlatformUpgrade() > 0 || summary.needReboot() > 0 || summary.needRestart() > 0)
+                logger.log(INFO, "No nodes allowed to suspend to progress installation for " + installationTimeout.toMinutes() + " minutes.");
+            else
+                logger.log(INFO, "Nodes not able to start with new application package.");
+            failed = true;
+        }
+
+        Duration timeout = JobRunner.jobTimeout.minusHours(1); // Time out before job dies.
+        if (timedOut(id, deployment.get(), timeout)) {
+            logger.log(INFO, "Installation failed to complete within " + timeout.toHours() + "hours!");
```
hakonhall: Should probably add some advise on what the customer should do next when the deployment fails?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
jonmv: Would the number of allowed down hosts ever exceed this threshold? For the start and end (with failure), I think all ghosts should be printed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
jonmv: Yes.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
jonmv: Fair enough.

jonmv:
Will make amends later :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
hakonhall: Possibly - I'm just thinking such a util might be used to dump out a lot of hosts - like all hosts or all but a few - but if that's not possible there is no point is making it complicated.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
```diff
@@ -180,8 +180,8 @@ private ApplicationVersion applicationVersionFrom(Inspector versionObject) {
         if ( ! summaryArray.valid())
             return Optional.empty();
 
-        if (summaryArray.entries() != 11)
-            throw new IllegalArgumentException("Convergence summary must have 11 entries");
+        if (summaryArray.entries() != 12)
+            throw new IllegalArgumentException("Convergence summary must have 12 entries");
```
jonmv: Argh, forgot the previous, with 11 entries, was merged already >_<

PR #11941, opened 2020-01-25
Gjoranv/metrics v2 in container 2

gjoranv:
Reapplies #11915 with one additional commit to exclude `httpclient` in container-dev.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -119,8 +119,8 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <scope>test</scope>
     </dependency>
+
     <dependency>
       <groupId>com.yahoo.vespa</groupId>
       <artifactId>testutil</artifactId>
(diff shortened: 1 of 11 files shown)


hmusum (approved):

hmusum (changes requested):
Build fails

gjoranv:
Yes, and there is unfortunately no easy fix. I forgot to mark as do-not-merge. Done now.

gjoranv:
@bjorncs, the best way forward is to have a simple VespaHttpClient that is not based on apache. This application does not have very strict requirements to flexibility or performance. If that is not feasible, you must make the httpclient version in vespa and hosted:tenant-base the same. Now they are different, and they shouldn't be in any case.

gjoranv:
FYI: @oyving

gjoranv:
We decided to upgrade `httpclient` and `httpcore` to the same versions as those used in Vespa. It is currently not feasible to create a VespaHttpClient based on the http client in the Java standard lib.

@olaaun Please review. Build should pass after the modified last commit that adds httpclient on the test classpath of application.

olaaun (approved):

gjoranv:
@olaaun PTAL. Added a commit to rename 'serviceid' to 'serviceId'

olaaun (approved):

gjoranv:
Awaiting merge until 7.167 release is tagged, which includes upgrade of apache httpcomponents in hosted Vespa's tenant:base.

PR #11943, opened 2020-01-25
Remove use-bucket-space-metric feature flag

hakonhall:
The flag controlled config read by the Cluster Controller. Therefore, I have
left the ModelContextImpl.Properties method and implementation (now always
returning true), but the model has stopped using that method internally, and
the config is no longer used in the CC.

The field in the fleetcontroller.def is left unchanged and documented as
deprecated.


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
```diff
@@ -42,8 +42,8 @@ protected void setUp(boolean dontInitializeNode2) throws Exception {
         {
             Set<ConfiguredNode> nodes = FleetControllerTest.toNodes(0, 1, 2, 3);
             ContentCluster cluster = new ContentCluster(
-                    "books", nodes, distribution, 6 /* minStorageNodesUp*/, 0.9 /* minRatioOfStorageNodesUp */,
-                    true /* determineBucketsFromBucketSpaceMetric */);
+                    "books", nodes, distribution, 6 /* minStorageNodesUp*/, 0.9 /* minRatioOfStorageNodesUp */
+                    /* determineBucketsFromBucketSpaceMetric */);
```
hmusum: Remove comment for the removed argument as well

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
```diff
@@ -57,8 +57,8 @@ protected void setUp(boolean dontInitializeNode2) throws Exception {
             Set<ConfiguredNode> nodesInSlobrok = FleetControllerTest.toNodes(1, 3, 5, 7);
 
             ContentCluster cluster = new ContentCluster(
-                    "music", nodes, distribution, 4 /* minStorageNodesUp*/, 0.0 /* minRatioOfStorageNodesUp */,
-                    true /* determineBucketsFromBucketSpaceMetric */);
+                    "music", nodes, distribution, 4 /* minStorageNodesUp*/, 0.0 /* minRatioOfStorageNodesUp */
+                    /* determineBucketsFromBucketSpaceMetric */);
```
hmusum: Remove comment for the removed argument as well

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -217,8 +214,9 @@ public String athenzDnsSuffix() {
         @Override
         public double defaultTermwiseLimit() { return defaultTermwiseLimit; }
 
+        // TODO: Remove
         @Override
-        public boolean useBucketSpaceMetric() { return useBucketSpaceMetric; }
+        public boolean useBucketSpaceMetric() { return true; }
```
hmusum: Consider making a default implementation of the method in the interface, then you could remove this now

hmusum (approved):

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
hakonhall: Fixed

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/restapiv2/StateRestApiTest.java
hakonhall: Fixed

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
hakonhall: I don't think I can do that since the interface is in config-model-api: Old deployments would only see the old interface and therefore no default implementation.

hmusum: Maybe I don't get this, but old config models implement the method, so that should not be an issue. Anyway, this is nitpicking, so I wouldn't bother.

hmusum (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
hakonhall: The implementation is in ModelContextImpl, in the configserver bundle, and is always at the latest config server Vespa version, and independent on the particular version of the model running?

hmusum: You are completely right, I wrongly assumed this was in the model. Thanks.

hakonhall:
Can be merged once all applications have upgraded to 7.166.

hakonhall:
FYI @vekterli

hakonhall:
This can now be merged, but need to resolve merge conflicts

PR #11952, opened 2020-01-26
Use testerReady call via config server for tester containers if flag …

hmusum:
…is set

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
@@ -417,7 +417,7 @@ public class InternalStepRunner implements StepRunner {
                            .collect(toList()));
         if (nodeList.summary().converged()) {
             if (endpointsAvailable(testerId, zone, logger)) {
-                if (containersAreUp(testerId, zone, logger)) {
+                if (testerContainersAreUp(testerId, zone, logger)) {
                     logger.log("Tester container successfully installed!");
                     return Optional.of(running);
                 }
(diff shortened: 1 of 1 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -455,6 +456,22 @@ private boolean containersAreUp(ApplicationId id, ZoneId zoneId, DualLogger logg
         return true;
     }
 
+    /** Returns true iff all containers in the tester deployment give 100 consecutive 200 OK responses on /status.html. */
+    private boolean testerContainersAreUp(ApplicationId id, ZoneId zoneId, DualLogger logger) {
+        if (useConfigServerForTesterAPI(zoneId)) {
+            DeploymentId deploymentId = new DeploymentId(id, zoneId);
+            if (controller.jobController().cloud().testerReady(deploymentId)) {
```
jonmv: Should this use the config server way as well?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -525,38 +542,58 @@ private String humanize(Node.ServiceState state) {
...
+                                                        controller.applications().contentClustersByZone(deployments));
+        if (useConfigServerForTesterAPI(zoneId)) {
+            controller.serviceRegistry().configServer().startTests(getTesterDeploymentId(id), suite, config);
+        } else {
+            controller.jobController().cloud().startTests(testerEndpoint.get(), suite, config);
+        }
         return Optional.of(running);
     }
 
+    private boolean testerReady(RunId id, URI testerEndpoint) {
```
jonmv: Was this used? (Reviewing on phone...)

jonmv (approved):
Please look at comments:)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
hmusum: It does (if argument is deployment id the config server will be used)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
hmusum: Oh, forgot to remove that one (ended up not being able to use it after all), will remove in next PR

PR #11969, opened 2020-01-27
Bjorncs/response based retry handler

bjorncs:


http-utils/src/main/java/ai/vespa/util/http/retry/DelaySupplier.java
```diff
@@ -0,0 +1,44 @@
...
+        Exponential(Duration startDelay, Duration maxDelay) {
+            this.startDelay = startDelay;
+            this.maxDelay = maxDelay;
+        }
+
+        @Override
+        public Duration getDelay(int executionCount) {
+            Duration nextDelay = startDelay;
+            for (int i = 1; i < executionCount; ++i) {
+                nextDelay = nextDelay.multipliedBy(2);
```
hakonhall: `multipliedBy(1 << (executionCount-1))` or similar?

hakonhall (approved):

http-utils/src/main/java/ai/vespa/util/http/retry/DelaySupplier.java
bjorncs: That was a bit more elegant yes 👍

PR #11980, opened 2020-01-28
Balder/avoid crossing the creek for water.

baldersheim:
@vekterli PR

### DIFF
--- a/document/src/vespa/document/fieldvalue/serializablearray.cpp
+++ b/document/src/vespa/document/fieldvalue/serializablearray.cpp
@@ -36,6 +36,7 @@ SerializableArray::set(EntryMap entries, ByteBuffer buffer,
         _unlikely->_compSerData = std::move(buffer);
         _unlikely->_serializedCompression = comp_type;
         _unlikely->_uncompressedLength = uncompressed_length;
+        _uncompSerData = ByteBuffer();
     } else {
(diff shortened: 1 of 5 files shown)


document/src/vespa/document/fieldvalue/serializablearray.cpp
```diff
@@ -26,20 +26,19 @@ class BufferMap : public BufferMapT {
...
+SerializableArray::set(EntryMap entries, ByteBuffer buffer,
+                       CompressionConfig::Type comp_type, uint32_t uncompressed_length)
 {
-
+    _entries = std::move(entries);
     if (CompressionConfig::isCompressed(comp_type)) {
         _unlikely = std::make_unique<RarelyUsedBuffers>();
         _unlikely->_compSerData = std::move(buffer);
         _unlikely->_serializedCompression = comp_type;
         _unlikely->_uncompressedLength = uncompressed_length;
```
vekterli: Should `_uncompSerData` also be reset in this branch?

vekterli (approved):
:+1: one question regarding member resetting

document/src/vespa/document/fieldvalue/serializablearray.cpp
baldersheim: Yes, it should, even if the place this method is used, this is under control.

PR #11988, opened 2020-01-28
Change fieldset query-command to add, not override commands

bratseth:
The logic in FieldSet stated that any explicit query-command
will deactivate all the standard derived commands. That is
inconsistent with query-command on fields, not mentioned in
the doc, and will generally break things.

This changes to the normal behavior where query-commands are
additive.

@baldersheim or @geirst can you remember if there was some reason for this? It's before OS so blame doesn't tell me anything. Seems just wrong to me.


baldersheim (approved):
The changes looks fine to me. However the change in semantics that you mention I can not shed any light on. No bells are ringing. I think that @geirst or @arnej27959 must chime in on this one.

bratseth:
@geirst did offline. Merging.

PR #11990, opened 2020-01-28
Use /usr/bin/sudo instead of searching $PATH to avoid broken

toregge:
sudo wrapper script.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #11993, opened 2020-01-29
New metric: seconds node has been suspended

hakonhall:


freva (approved):
All the maintainer should just take the clock from `NodeRepository`, but doesn't have to fix that in this PR.

hakonhall:
Thinking about it a bit more, I'm not too enthused about `nodeRepository.clock()` - it seems unnatural for the maintainer to depend on the node repository having and exposing a clock.

PR #11998, opened 2020-01-29
Add ReadableAttributeVector accessor to IAttributeManager

vekterli:

Provides a unified interface for fetching both regular as well as
imported attributes. Exposing `ReadableAttributeVector` instead of
raw `AttributeVector` instances enforces that all access is done via
appropriate acquired read guards.

Refactor document selection processing code to use the new interface
in order to prepare for imported field support in selections.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/common/selectcontext.cpp
+++ b/searchcore/src/vespa/searchcore/proton/common/selectcontext.cpp
@@ -3,20 +3,22 @@
 #include "selectcontext.h"
 #include "cachedselect.h"
 #include <vespa/document/select/value.h>
-#include <vespa/searchlib/attribute/attributeguard.h>
+#include <vespa/searchlib/attribute/attribute_read_guard.h>
+#include <vespa/searchlib/attribute/readable_attribute_vector.h>
+#include <cassert>
 
 namespace proton {
 
 using document::select::Value;
 using document::select::Context;
-using search::AttributeGuard;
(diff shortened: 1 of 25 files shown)


searchcore/src/vespa/searchcore/proton/common/selectcontext.cpp
```diff
@@ -26,23 +28,30 @@ SelectContext::SelectContext(const CachedSelect &cachedSelect)
...
 
-
 void
 SelectContext::dropAttributeGuards()
 {
     _guards->clear();
 }
 
+const search::attribute::AttributeReadGuard&
+SelectContext::read_guard_at_index(uint32_t index) const noexcept
```
geirst: Do we need to expose AttributeReadGuard here or could we just return the IAttributeVector interface instead (that you get from the guard)? If so, we should probably rename the function to something like attribute_at_index() or guarded_attribute_at_index().

geirst (approved):
👍 One comment to consider.

searchcore/src/vespa/searchcore/proton/common/selectcontext.cpp
vekterli: Agreed, it's better to just expose the `IAttributeVector`. Will do this as a follow-up since I'm already working on this code base as part of the "imported fields in selections" task

toregge (approved):

PR #12002, opened 2020-01-29
Support phrase-segmenting false in more cases

bratseth:
Support and segmenting with simple queries also
when the terms are bound by a control character
in the simple language such that it is tokenized by us.


arnej27959 (approved):
looks good to me

PR #12006, opened 2020-01-30
Prepare for setting PERMANENTLY_DOWN

hakonhall:


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
```diff
@@ -99,12 +99,13 @@ public boolean setHostStatus(ApplicationInstanceReference application, HostName
 
     /**
      * Cache is checked for freshness when this mapping is created, and may be invalidated again later
-     * by other users of the cache. Since this function is backed by the cache, any such invalidations
+     * by other users of the cache. Since this function is backed by the cache, any such invalidation
      * will be reflected in the returned mapping; all users of the cache collaborate in repopulating it.
      */
     @Override
-    public Function<ApplicationInstanceReference, Set<HostName>> getSuspendedHostsByApplication() {
-        return application -> hostInfosCache.getHostInfos(application).suspendedHostsnames();
```
hakonhall: This line is fairly new. The older code read the cache counter only once, and cleared the cache only once (or didn't clear it at all). While the current code will do cache invalidation on every invocation of the returned Function. The old behavior is restored with this PR.

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
```diff
@@ -99,12 +99,13 @@ public boolean setHostStatus(ApplicationInstanceReference application, HostName
...
+    public Function<ApplicationInstanceReference, HostInfos> getHostInfosByApplicationResolver() {
+        hostInfosCache.refreshCache();
+        return hostInfosCache::getCachedHostInfos;
```
freva: Why not just 
```suggestion
        return hostInfosCache::getHostInfos;
```
?

freva (approved):

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/ZookeeperStatusService.java
hakonhall: I tried to explain that to the left?

freva: :man_facepalming:

PR #12007, opened 2020-01-30
Header/body does not exist anymore. Deprecate and remove own internal…

baldersheim:
… usage.

@arnej27959 PR


baldersheim:
@arnej27959 Ping

document/src/tests/documenttestcase.cpp
```diff
@@ -585,7 +585,7 @@ TEST(DocumentTest, testReadSerializedFile)
 
     EXPECT_TRUE(buf2.empty());
     buf2.rp(0);
-    EXPECT_EQ(len - 13, buf2.size()); // Size is smaller as we are merging to one chunk.
+    EXPECT_EQ(len, buf2.size());
```
vekterli: Do we have any files checked in testing that we can read from legacy multi-chunk serializations, or is this sufficiently ancient by now that we don't consider it necessary anymore?

vekterli (approved):
:+1:

document/src/tests/documenttestcase.cpp
baldersheim: It is fairly ancient, but I still think we have some.
I will check.

PR #12009, opened 2020-01-30
Bjorncs/support access token in athenz filter

bjorncs:
This PR changes the config definition of the filter. We'll have to wait with merging until the hosted tester-app no longer uses this filter.

FYI @hmusum

### DIFF
--- a/vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/AthenzRole.java
+++ b/vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/AthenzRole.java
@@ -7,6 +7,8 @@ import java.util.Objects;
  * @author tokle
  */
 public class AthenzRole {
+    private static final String DOMAIN_ROLE_NAME_DELIMITER = ":role.";
+
     private final AthenzDomain domain;
     private final String roleName;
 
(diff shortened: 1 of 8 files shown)


vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/AthenzRole.java
```diff
@@ -20,6 +22,16 @@ public AthenzRole(String domain, String roleName) {
         this.roleName = roleName;
     }
 
+    public static AthenzRole fromString(String string) {
```
tokle: consider naming the method to reflect that it expects the format <domain>:role.<rolename>

tokle (approved):

vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/AthenzRole.java
bjorncs: Addressed in new commit.

bjorncs:
@tokle PTAL

tokle (approved):

tokle (approved):

PR #12010, opened 2020-01-30
Do not hold the lock when visiting documents inflight.

baldersheim:
@vekterli PR

### DIFF
--- a/searchlib/src/vespa/searchlib/docstore/writeablefilechunk.cpp
+++ b/searchlib/src/vespa/searchlib/docstore/writeablefilechunk.cpp
@@ -203,6 +203,13 @@ seek_past(LidInfoWithLidV::const_iterator begin, LidInfoWithLidV::const_iterator
    return begin;
 }
 
+struct LidAndBuffer {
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/docstore/writeablefilechunk.cpp
```diff
@@ -225,12 +233,18 @@ WriteableFileChunk::read(LidInfoWithLidV::const_iterator begin, size_t count, IB
                         assert(chunk == _active->getId());
                         buffer = _active->getLid(li.getLid());
                     }
-                    visitor.visit(li.getLid(), buffer);
+                    auto copy = std::make_unique<char[]>(buffer.size());
```
vekterli: Should we consider using `Alloc` here in case of some crazy big documents?

vekterli (approved):
:+1:

searchlib/src/vespa/searchlib/docstore/writeablefilechunk.cpp
baldersheim: Good point, fixed.

PR #12012, opened 2020-01-30
Fix unstable test

mpolden:


bratseth (approved):
Nice. Let's see if it helps. @baldersheim fyi

PR #12016, opened 2020-01-31
All fields except build number are optional now

jonmv:
@mpolden please review. This commit could be used to patch controllers failing to deserialise applications. 

@smorgrav FYI.


tokle (approved):
@mpolden says 👍

tokle:
Java build succesful

PR #12024, opened 2020-01-31
Handle empty source revision

mpolden:
Would be good to have this in Monday's release, but @jonmv should probably have
a look first.

`ApplicationVersion` needs a serious makeover...

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
@@ -258,15 +258,13 @@ public class ApplicationSerializer {
     }
 
     private void toSlime(ApplicationVersion applicationVersion, Cursor object) {
-        if (applicationVersion.buildNumber().isPresent() && applicationVersion.source().isPresent()) {
-            object.setLong(applicationBuildNumberField, applicationVersion.buildNumber().getAsLong());
-            toSlime(applicationVersion.source().get(), object.setObject(sourceRevisionField));
-            applicationVersion.authorEmail().ifPresent(email -> object.setString(authorEmailField, email));
-            applicationVersion.compileVersion().ifPresent(version -> object.setString(compileVersionField, version.toString()));
-            applicationVersion.buildTime().ifPresent(time -> object.setLong(buildTimeField, time.toEpochMilli()));
-            applicationVersion.sourceUrl().ifPresent(url -> object.setString(sourceUrlField, url));
-            applicationVersion.commit().ifPresent(commit -> object.setString(commitField, commit));
-        }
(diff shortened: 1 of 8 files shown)


bratseth (approved):
LGTM!

jonmv:
I would guess the problem was that an empty application version was written as the `latestVersion` — fixed here. We should also discard any empty `latestVersion`s — see #12028. What are the other changes here?

Yes, `ApplicationVersion` is in the process of a make-over :p But there are many involved parties. Summary: it is an OK version as long as it has a build number. Everything else is optional. Some versions have only a `commit`, and some have the full source revision. Moving to `commit` long-term, together with the `sourceUrl`, as this is what we always used it for, and is more general as well. The fields are user-defined.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
```diff
@@ -258,15 +258,13 @@ private void zoneIdToSlime(ZoneId zone, Cursor object) {
     }
 
     private void toSlime(ApplicationVersion applicationVersion, Cursor object) {
-        if (applicationVersion.buildNumber().isPresent() && applicationVersion.source().isPresent()) {
```
mpolden: @jonmv This was the bug. When source revision is empty, no fields are written to the `latestVersion` object, but the object itself is created, thus becoming an unknown version. Subsequent reads fail because the `Application` constructor throws if `latestVersion` is unknown so fromSlime/toSlime is asymmetric.

> What are the other changes here?

Cleanup of `SourceRevision`. `SourceRevision` where all fields are empty is now treated as an empty source revision.

jonmv: Yes, this was the bug. We still have serialised `unknown` `latestVersion`s, though? Or did you clean them out by hand?

Are the fields all empty? I'd imagine the `commit` field wasn't. The plan was to remove `SourceRevision`, since they're only used for the `commit` and the generated source URLs.

jonmv:
Aha, I see you fixed the problem by hand. Good.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
mpolden: >  Or did you clean them out by hand?

Yes.

> Are the fields all empty?

In the production data, `sourceRevision` is either a) `null`, b) an object with all non-empty or fields or c) an object with all empty fields.

mpolden: > The plan was to remove SourceRevision.

Please do! 🔥

PR #12027, opened 2020-01-31
Make recall work also if the query does not contain 'foo'

bratseth:
@simonthoresen wtf??


bratseth:
As, sorry. I get it now - this is "fine".

PR #12028, opened 2020-01-31
Clear any existing empty latestVersions

jonmv:


jonmv:
First part of commit message was already covered by someone else, presumably you :)

jonmv:
Or ... do the changes in #12024 work?

mpolden:
#12024 doesn't do this, no. In the affected system we fixed this manually in ZooKeeper (only one application).

jonmv:
Then this is not needed.

PR #12033, opened 2020-02-01
Don't write empty latestVersions, and clear any existing empty

jonmv:
@hmusum or @mpolden or @bratseth  please review and merge. 

Original problem was threefold: 

1. empty source revisions lead to empty `latestVersions` being read (even though they weren't empty)
1. empty source revisions lead to empty `latestVersions` being written
1. empty `latestVersions` were now serialised due to 2 

#12016 fixed 1., while #12024 fixed 2. I opened #12028 to fix 3., but this was already done by hand. 

Now, #12024 introduced a different problem, as @hmusum found. The other changes in #12024, to `SourceRevision`, aren't important, as the plan is to remove that class anyway. However, we _need_ the fix to problem 2. 

This PR fixes both 2. and 3.


bratseth (approved):

mpolden:
Thanks! What a mess... :-)

PR #12034, opened 2020-02-02
Add support for importing LightGBM models

lesters:
@bratseth Please review. This should also work with model-evaluation, however there was an assumption there that arguments could be converted to tensors. Since LightGBM supports categorical features we need to have string arguments as well. See if you agree with the changes in `ModelsEvaluationHandler` and `FunctionEvaluator` or if there is some more elegant way of handling this currently.

### DIFF
--- a/application/src/main/java/com/yahoo/application/Application.java
+++ b/application/src/main/java/com/yahoo/application/Application.java
@@ -2,6 +2,7 @@
 package com.yahoo.application;
 
 import ai.vespa.rankingexpression.importer.configmodelview.MlModelImporter;
+import ai.vespa.rankingexpression.importer.lightgbm.LightGBMImporter;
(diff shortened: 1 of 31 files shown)


bratseth (approved):
Great! On the categorical features, those are just sets, and a mapped 1d tensor with unit weights is a set, set intersection is tensor join and set union is tensor merge. In this way we could deprecate the special (and limited) support for set operations.

Similar as to tensors in general we'd introduce the concept of names to dimensions. It's good really because it gives the set a notion of a type where we can validate that only sets of the same type (i.e having the same kind of data) is combined, but I suppose it is missing fro these tools so we'd have to add it in the usual way.

This way would give us more expressive power with sets with a smaller language and interface - what do you think?

lesters:
Love the idea. Merging this for now and we can work on this going forward.

PR #12045, opened 2020-02-03
Bratseth/optimization on sddocname and recall

bratseth:


bratseth:
Test failure is unrelated.

container-search/src/main/java/com/yahoo/prelude/querytransform/QueryRewrite.java
```diff
@@ -200,6 +204,21 @@ private static void removeOtherNonrankedChildren(CompositeItem parent, int index
                 parent.removeItem(i);
         }
     }
+
+    private static boolean isRanked(Item item) {
+        if (item instanceof CompositeItem) {
+            boolean isRanked = false;
+            for (Item child : ((CompositeItem)item).items())
+                isRanked |= isRanked(child);
```
baldersheim: if (isRanked(child)) return true;
No need to iterate all children if one is unhappy.

container-search/src/main/java/com/yahoo/prelude/querytransform/QueryRewrite.java
```diff
@@ -200,6 +204,21 @@ private static void removeOtherNonrankedChildren(CompositeItem parent, int index
...
+            return isRanked;
```
baldersheim: return false;

baldersheim (approved):

container-search/src/main/java/com/yahoo/prelude/querytransform/QueryRewrite.java
bratseth: ok, done

PR #12050, opened 2020-02-03
Allow resources in dedicated nodes

bratseth:


bratseth:
https://github.com/vespa-engine/vespa/pull/12050

freva (approved):

PR #12030, opened 2020-01-31
Send ping every second truly async to all nodes who does not have any…

baldersheim:
… pending pings.

@havardpe and @bratseth PR

PR #12051, opened 2020-02-03
Require all up when going up, but require all down to go down.

baldersheim:
@bratseth PR
@hmusum @yngveaasheim FYI

### DIFF
--- a/container-core/abi-spec.json
+++ b/container-core/abi-spec.json
@@ -11,6 +11,23 @@
     ],
     "fields": []
   },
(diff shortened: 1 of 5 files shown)


baldersheim:
Let us hold this one until #12030 and #12031 has been rolled out.

bratseth (approved):
👍 this is still very easy to understand

baldersheim:
Do I sense some irony ? :)

bratseth:
No :-) I think this part really is very explicit and nice to understand, and it's an area where it really should be.

PR #12054, opened 2020-02-03
Start container with uncapped CPU and start capping 30s after first s…

freva:
…uccessful health check


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -333,16 +351,25 @@ private void removeContainer(NodeAgentContext context, Container existingContain
     }
 
 
-    private void updateContainerIfNeeded(NodeAgentContext context, Container existingContainer) {
+    /** Returns true iff container is correctly sized */
```
hakonhall: If doc

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
freva:
```suggestion
```

PR #12055, opened 2020-02-03
Handle default-deploy argument to bootstrap.sh, deploys artifacts

toregge:
for the default (minimal) bootstrap build.

@geirst : please review
@bjorncs , @aressem , @baldersheim : FYI


toregge:
Build job uploads these artifacts in a later build step.

PR #12058, opened 2020-02-03
Balder/do not say you are done until you are

baldersheim:
@bratseth or @mpolden PR


bratseth (commented):
👍

mpolden:
You beat me to it, I found the other Executor just before I left work today. 😅

Was thinking we could just skip starting the built-in ping thread in the test as there seems to be a flag for it: https://github.com/vespa-engine/vespa/blob/master/container-search/src/main/java/com/yahoo/search/cluster/ClusterMonitor.java#L45, but I suppose this works too.

baldersheim:
Good point. However starting a thread in teh constructor is not that good pattern anyway. I will move the starting of the thread to a separate start() method for full external control and the same for SearchCluster. That will these more test friendly.

PR #12059, opened 2020-02-04
Do not start cluster monitor thread in test as it will race with expl…

baldersheim:
…icit ping in test.

@mpolden or @bratseth PR

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -1954,6 +1954,7 @@
     "methods": [
       "public void <init>(com.yahoo.search.cluster.NodeManager)",
       "public void <init>(com.yahoo.search.cluster.NodeManager, boolean)",
+      "public void start()",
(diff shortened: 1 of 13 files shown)


bratseth (commented):
There's a lot of possible states with all these fields that may be initialized or not. I think we may simplify a bit by not creating the ClusterMonitor before the startClusterMonitoring method is called (it's initialized there anyway so not usable before).

Then we can also move the pingfactory into the clustermonitor and pass it in the constructor. Perhaps also the nodes?

Then we could perhaps also let the clustermonitor be passed in instead of the pingfactory and the boolean, such that it would be up to the caller to start it or not etc. (Not quite sure about this one).

baldersheim:
@bratseth Now you can have another look.

bratseth (approved):
Great

mpolden:
👍

PR #12063, opened 2020-02-04
Implement skeleton of a HNSW index on top of data stores.

geirst:
Currently only supports adding documents to layer 0 with simple strategy for selecting neighbors.

@arnej27959 please review
@toregge FYI


arnej27959 (approved):
looks good to me

PR #12065, opened 2020-02-04
track the total number of connection objects

havardpe:

@vekterli please review


fnet/src/vespa/fnet/connection.cpp
```diff
@@ -489,6 +491,7 @@ FNET_Connection::FNET_Connection(FNET_TransportThread *owner,
       _cleanup(nullptr)
 {
     assert(_socket && (_socket->get_fd() >= 0));
+    ++_num_connections;
```
vekterli: These could be `fetch_add(1, std::memory_order_relaxed)`/`fetch_sub(1, std::memory_order_relaxed)` instead to avoid fences, but it won't matter much in the grand scheme of things.

vekterli (approved):
:+1:

PR #12069, opened 2020-02-05
Increase memory sampling frequency to every 20s instead of every 60s …

baldersheim:
…in order to detect changes faster.

Also log anomalies.

@geirst PR


geirst (approved):
👍 
@toregge FYI

PR #12070, opened 2020-02-05
Never register non-production deployments in global endpoint

mpolden:
or @oyving


oyving (approved):
LGTM. Is there any serialisation for ZK issues or clean-up in AWS to consider when rolling this out?

mpolden:
I did the cleanup after deploying the patch.

PR #12077, opened 2020-02-05
Use a cached threadpool for execution of connect. To avoid failing co…

baldersheim:
…nnect blocking other connects.

@havardpe PR


havardpe:
We should consider having a max limit on connector threads to avoid self-dosing.

baldersheim:
I added a max limit of 8.

PR #12084, opened 2020-02-06
Geirst/hnsw index more functionality

geirst:
@arnej27959 please review

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hnsw_index_base.h
+++ b/searchlib/src/vespa/searchlib/tensor/hnsw_index_base.h
@@ -84,7 +84,7 @@ protected:
     NodeStore _nodes;
     LinkStore _links;
     uint32_t _entry_docid;
-    uint32_t _entry_level;
+    int _entry_level;
 
     static search::datastore::ArrayStoreConfig make_default_node_store_config();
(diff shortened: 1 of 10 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_index_base.h
```diff
@@ -72,29 +78,47 @@ class HnswIndexBase : public NearestNeighborIndex {
     using LinkArray = vespalib::Array<uint32_t>;
 
     const DocVectorAccess& _vectors;
+    RandomLevelGenerator& _level_generator;
     Config _cfg;
     NodeRefVector _node_refs;
     NodeStore _nodes;
     LinkStore _links;
     uint32_t _entry_docid;
+    uint32_t _entry_level;
```
arnej27959: i found it easier to have level as "int" always.  that also allows _entry_level = -1 to signal no valid level.

arnej27959 (approved):
looks good to me

searchlib/src/vespa/searchlib/tensor/hnsw_index_base.h
geirst: Fixed

PR #12086, opened 2020-02-06
Change deployment metric dimensions

jonmv:
@ldalves please review and merge.


ldalves (approved):
Great.. I think this will cover most of the use cases... if we miss `applicationName` and `instanceName` we'll find it out soon. :)

PR #12089, opened 2020-02-06
Include all endpoints in endpoints field

mpolden:
With this change, the console only needs to render the `endpoints` field. The
field will now include all know endpoints, regardless of the underlying routing
method used. No more `serviceUrls` or `globalRotation`.


freva (approved):
Great! :1st_place_medal: :100:

PR #12090, opened 2020-02-06
Add debug logging when getting config response and polling internal q…

hmusum:
…ueue

### DIFF
--- a/config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
+++ b/config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
@@ -234,6 +234,7 @@ public class JRTConfigRequester implements RequestWaiter {
         noApplicationWarningLogged = Instant.MIN;
         connection.setSuccess();
         sub.setLastCallBackOKTS(Instant.now());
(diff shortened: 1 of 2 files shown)


config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
```diff
@@ -234,6 +234,7 @@ private void handleOKRequest(JRTClientConfigRequest jrtReq,
         noApplicationWarningLogged = Instant.MIN;
         connection.setSuccess();
         sub.setLastCallBackOKTS(Instant.now());
+        log.log(LogLevel.DEBUG, "OK response received in handleOkRequest: " + jrtReq);
```
bjorncs: Consider using a lambda instead to ensure toString is only invoked if debug-logging is enabled.
```log.log(LogLevel.DEBUG, () -> "OK response received in handleOkRequest: " + jrtReq);```

bjorncs (approved):

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
hmusum: Yes, thanks, forgot that this time, fixed

bjorncs (approved):

Issue #12068, opened 2020-02-05
Segmentation fault when visiting


griseau:
As mentioned on Gitter, I'm facing a segmentation fault when trying to use vespa-visiting. 

The command I'm executing is : `vespa-visit -v --priority LOW_1 -s "post.text_clean_with_stopwords=\"*skin*\"" --fieldset "post:permalink,text_raw,channel,uid" > /workspace/skin_documents.json`

In vespa.log only thing I can see is : 
`1580892627.908253	vespa-container	1072/6732	config-sentinel	config-sentinel.service	event	crash/1 name="searchnode" pid=3982 signal="Segmentation fault"`

Vespa version: 7.165.5 with only one node at the moment

vekterli:
I can't see anything about the command itself that should indicate a potential problem.

Does this happen consistently if you run the command? A core dump backtrace of the offending thread would be very helpful to pinpoint the cause. If your system is not configured to collect core dumps, please consider temporarily enabling these (see the [Vespa documentation](https://docs.vespa.ai/documentation/reference/files-processes-and-ports.html#core-dumps) for an example of how to do so).

griseau:
Yes tried it a few time, always having the same problem.

Not sure how to share it. Is there any way to retrieve the core dumps from a specified timestamp ?

vekterli:
If you have configured `/proc/sys/kernel/core_pattern` the cores should be in the directory configured through it. Since this is very likely the same bug being hit every time, any core dump file should do.

If you have a file, say `proton-bi.core.1234`, something like this should give you the dump of the offending thread:
```
gdb --batch -ex bt $VESPA_HOME/sbin/vespa-proton-bin proton-bi.core.1234
```
Please let me know if this doesn't work

griseau:
Looks ok, you can find the [dump.txt](https://github.com/vespa-engine/vespa/files/4159070/dump.txt). 

Let me know if you have any question

vekterli:
Very interesting! Glob expressions are internally rewritten into regular expressions and it seems like something here has triggered the regex implementation into annihilating itself with a recursion so deep it runs out of stack space (touching a guard page and segfaulting the process). If I were to hazard a guess, this is likely due to the combination of document content and resulting regular expression triggering catastrophic backtracking during regex evaluation. Needless to say, this isn't optimal behavior.

I'll look more closely into this. For now, you could try working around the issue by using the explicit regex operator instead:
```
post.text_clean_with_stopwords=~"skin"
[...]

PR #12096, opened 2020-02-06
Simplify regexes generated from document selection glob patterns

vekterli:
@geirst @baldersheim please review

Attempts to simplify the resulting regex as much as possible to help
minimize the number of possible catastrophic backtracking cases that
can be triggered by wildcard regexes.

The following simplifications are currently performed:
* `'' -> /^$/`   (empty string match)
* `'*' -> //`     (any string match)
* `'*foo*' -> /foo/`  (substring match)
* `'*foo' -> /foo$/` (suffix match)
* `'foo*' -> /^foo/` (prefix match)
* collapsing runs of consecutive `*` wildcards into a single
  wildcard. `***` is identical to `**` which is identical to `*` etc,
  as all these match 0-n characters each. This also works with
  simplification, i.e. `***foo***` -> `/foo/` and `***` -> `//`

Could be further simplified still by moving '?' matches as far to the left
as possible around wildcards, effectively turning `*?*?*?*` into `???`,
but there are limits to how much we can do to avoid self foot vs. gun
interaction.

This relates to issue #12068

### DIFF
--- a/document/src/tests/documentselectparsertest.cpp
+++ b/document/src/tests/documentselectparsertest.cpp
@@ -632,7 +632,7 @@ TEST_F(DocumentSelectParserTest, operators_2)
     PARSEI("id.group == \"xyzzy\"", *_doc[10], True);
 }
 
-TEST_F(DocumentSelectParserTest, operators3)
+TEST_F(DocumentSelectParserTest, operators_3)
 {
     createDocs();
     {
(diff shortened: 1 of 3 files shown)


baldersheim (approved):

document/src/tests/documentselectparsertest.cpp
```diff
@@ -637,7 +632,7 @@ void DocumentSelectParserTest::testOperators2()
     PARSEI("id.group == \"xyzzy\"", *_doc[10], True);
 }
 
-void DocumentSelectParserTest::testOperators3()
+TEST_F(DocumentSelectParserTest, operators3)
```
geirst: operators3 -> operators_3 for consistency?

geirst (approved):
👍

document/src/tests/documentselectparsertest.cpp
vekterli: Done

PR #12097, opened 2020-02-06
Implement wrapper for std::atomic of type EntryRef and use it in hnsw…

geirst:
… index to get snapshotting of level arrays.

@vekterli please review
@arnej27959 @toregge FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hnsw_index_base.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/hnsw_index_base.cpp
@@ -3,6 +3,7 @@
 #include "hnsw_index_base.h"
 #include "random_level_generator.h"
 #include <vespa/vespalib/datastore/array_store.hpp>
+#include <vespa/vespalib/util/rcuvector.hpp>
 
 namespace search::tensor {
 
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_index_base.cpp
```diff
@@ -65,18 +64,16 @@ HnswIndexBase::get_link_array(uint32_t docid, uint32_t level) const
...
+    return _links.get(levels[level].load_acquire());
 }
 
 void
 HnswIndexBase::set_link_array(uint32_t docid, uint32_t level, const LinkArrayRef& links)
 {
     auto links_ref = _links.add(links);
-    // TODO: Add memory barrier?
-    auto node_ref = _node_refs[docid];
+    auto node_ref = _node_refs[docid].load_acquire();
```
vekterli: If this is always invoked from the writer thread, it's technically sufficient to use a `load_relaxed()` here since the changes to these structures shall always be visible to that thread anyway. But acquire loads are almost free on x86-64 anyway, so it's hardly something to prematurely optimize for.

vekterli (approved):
:+1:

PR #12094, opened 2020-02-06
Add latest application and normal conf. Vespa, and blockers, to dep.g…

jonmv:
…raph.view API

@smorgrav please review and merge.

PR #12106, opened 2020-02-07
Revert "Add latest application and normal conf. Vespa, and blockers, to dep.g…"

aressem:
Reverts vespa-engine/vespa#12094

Broke unit tests


hmusum (approved):

jonmv:
I don't think this broke the tests.

PR #12120, opened 2020-02-07
Support setting PERMANENTLY_DOWN at end of retirement

hakonhall:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -122,6 +122,13 @@
             "Takes immediate effect for new batch suspensions.",
             HOSTNAME);
 
+    public static final UnboundBooleanFlag RETIRE_WITH_PERMANENTLY_DOWN = defineFeatureFlag(
+            "retire-with-permanently-down", true,
+            "If enabled, retirement will end with setting the host status to PERMANENTLY_DOWN, " +
+            "instead of ALLOWED_TO_BE_DOWN (old behavior).",
+            "Takes effect on the first run of RetiredExpirer.",
```
freva: First or next?

freva (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hakonhall: Changed to "next"

freva (approved):

PR #12127, opened 2020-02-09
Config cleanup 4

hmusum:

### DIFF
--- a/config-model/src/main/java/com/yahoo/config/model/producer/AbstractConfigProducer.java
+++ b/config-model/src/main/java/com/yahoo/config/model/producer/AbstractConfigProducer.java
@@ -7,7 +7,6 @@ import com.yahoo.config.model.ApplicationConfigProducerRoot;
 import com.yahoo.config.model.deploy.DeployState;
 import com.yahoo.config.subscription.ConfigInstanceUtil;
 import com.yahoo.log.LogLevel;
-import com.yahoo.text.Utf8;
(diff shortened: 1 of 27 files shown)


baldersheim (approved):
I see that there are also writeFiles in some other places that seems like they can go.
-bash-4.2$ git grep writeFiles
config-model/src/main/java/com/yahoo/config/model/producer/AbstractConfigProducer.java:    public void writeFiles(File directory) throws java.io.IOException {
config-model/src/main/java/com/yahoo/vespa/model/ConfigProducer.java:    void writeFiles(File directory) throws IOException;
config-model/src/main/java/com/yahoo/vespa/model/Host.java:    public void writeFiles(File directory) {
config-model/src/main/java/com/yahoo/vespa/model/VespaModel.java:    public void writeFiles(File baseDirectory) throws IOException {
config-model/src/main/java/com/yahoo/vespa/model/VespaModel.java:        super.writeFiles(baseDirectory);
config-model/src/main/java/com/yahoo/vespa/model/VespaModel.java:                cp.writeFiles(destination);
config-model/src/main/java/com/yahoo/vespa/model/search/SearchCluster.java:    public void writeFiles(File directory) throws java.io.IOException {
config-model/src/main/java/com/yahoo/vespa/model/search/SearchCluster.java:        super.writeFiles(directory);

hmusum:
Yes, thanks, didn't see those. Removed.

baldersheim:
Merge at will :)

PR #12128, opened 2020-02-09
Disable retire-with-permanently-down by default

hakonhall:


hakonhall:
Since PERMANENTLY_DOWN would otherwise be left for each such retired host due to missing cleanup a.t.m.

freva (approved):

PR #12129, opened 2020-02-10
Config cleanup 5

hmusum:


config-proxy/src/main/java/com/yahoo/vespa/config/proxy/RpcConfigSourceClient.java
```diff
@@ -183,15 +171,14 @@ public void shutdownSourceConnections() {
             activeSubscribers.clear();
         }
         exec.shutdown();
-        for (JRTConfigRequester requester : requesterPool.values()) {
+        if (requester != null)
```
freva: Its never `null`?

hmusum: Yes, thanks, fixed

hmusum:
@freva could you please take a look again?

freva (approved):

PR #12135, opened 2020-02-10
Hnsw simplifications with distance function interface

geirst:
@arnej27959 please review


searchlib/src/vespa/searchlib/tensor/distance_functions.h
```diff
@@ -0,0 +1,34 @@
...
+// TODO: Make distance functions hardware optimized.
+
+/**
+ * Calculates the square of the standard Euclidean distance.
+ */
+template <typename FloatType>
+class SquaredEuclideanDistance : public DistanceFunction {
+public:
+    double calc(const vespalib::tensor::TypedCells& lhs, const vespalib::tensor::TypedCells& rhs) const override {
+        auto lhs_vector = lhs.template typify<FloatType>();
```
arnej27959: i don't see that you should need "template" here...

arnej27959 (approved):
looks good to me, one very minor doubt only

searchlib/src/vespa/searchlib/tensor/distance_functions.h
geirst: You are right. It was (somehow) needed when this was done inside the templated HnswIndex class, but not here. Fixed.

Issue #11588, opened 2019-12-17
Partial update of array<position> field doesn't work


wardsng:
Search definition snippet:
```
        field latLong type position {
            indexing: attribute
        }
        field allLatLong type array<position> {
            indexing: attribute
        }
```

Doing a document PUT like so works; subsequent GET shows `allLatLong` is there
```
{
    "fields": {
        "allLatLong":  ["N41o40'51;W72o56'19"]
        "latLong": "N41o40'51;W72o56'19"
    }
}
```

Doing a partial document UPDATE like so does not work:
```
{
    "fields": {
        "allLatLong":{
            "assign": ["N41o40'51;W72o56'19"]
        },
        "latLong": {
            "assign":"N41o40'51;W72o56'19"
        }
    },
    "create": true
}
```
The update to `allLatLong` does not take effect

vekterli:
My tingling spider sense tells me this is very likely to be an issue with the indexing docproc, as it needs to transform position fields into a zcurve representation before passing them on to the backend.

I'll add a system test for this, which also verifies if it works with `mode="store-only"` (quite likely).

vekterli:
After some debugging of the new system test it looks like this is probably a related issue to that fixed some time ago in https://github.com/vespa-engine/vespa/pull/11244

The backend _does_ receive a zcurve attribute update, but this is not reflected by the Get operations (which pulls the data from the doc store instead). This means that the update _does_ actually take effect for searches, it's just not visible when retrieving the document with Gets (which will be an issue when merging)

Debug log excerpt from the searchnode process:

Initial feed of 2-element array (non-zcurve position field data is stored in doc store separately):
```
handleValue(position_array_zcurve, 2): Array(size: 2,
          6146690698477297747,
          6148167025757726373
        )
```
Partial update assigning a single element array:
```
handleFieldUpdate(position_array_zcurve, 2): FieldUpdate(Field(position_array_zcurve, id 880245166, ArrayDataType(
            NumericDataType(Long, id 4), id 58874399), header)
            AssignValueUpdate(Array(size: 1,
            6146710697028297953
          ))
        )

handleValueUpdate(position_array_zcurve, 2): AssignValueUpdate(Array(size: 1,
          6146710697028297953
        ))
```

vekterli:
This should be fixed in versions >= **7.176**, which should presumably be available as an open-source release in a few days

PR #12137, opened 2020-02-10
Patch retrieved array of position fields from zcurve array attribute

vekterli:
@geirst please review

Ensures that updates made to the authoritative underlying zcurve array
attribute are visible in documents retrieved via Get/Visit et al.
Would otherwise return any outdated values that were present in the
docstore that were part of the document version originally fed.

This relates to issue #11588


config/src/tests/file_subscription/my.cfg
```diff
@@ -1 +1 @@
-myField "foobar"
+myField "foo"
```
vekterli: Can't remember ever touching this file or adding it to the commit, so not sure where this diff came from 🤔 IDE must have changed it somehow and snuck it under the radar. Will remove.

baldersheim (approved):

geirst (approved):
👍

PR #12151, opened 2020-02-11
If debugging of SearchHandler is turned on, the 1k first request will…

baldersheim:
… have trace and timing information.

The 1k number is configurable in the container-http config.

@bratseth PR
@yngveaasheim FYI

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
+++ b/container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
@@ -69,14 +69,14 @@ public class HttpSearchResponse extends ExtendedResponse {
         }
     }
 
-    public ListenableFuture<Boolean> waitableRender(OutputStream stream) throws IOException {
+    public ListenableFuture<Boolean> waitableRender(OutputStream stream) {
         return waitableRender(result, query, rendererCopy, stream);
     }
 
     public static ListenableFuture<Boolean> waitableRender(Result result,
                                                            Query query,
                                                            Renderer<Result> renderer,
-                                                           OutputStream stream) throws IOException {
+                                                           OutputStream stream) {
         SearchResponse.trimHits(result);
         SearchResponse.removeEmptySummaryFeatureFields(result);
         return renderer.render(stream, result, query.getModel().getExecution(), query);
--- a/container-core/src/main/java/com/yahoo/container/jdisc/ThreadedHttpRequestHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/jdisc/ThreadedHttpRequestHandler.java
@@ -93,7 +93,7 @@ public abstract class ThreadedHttpRequestHandler extends ThreadedRequestHandler 
 
     /** Render and return whether the channel was closed */
     private void render(HttpRequest request, HttpResponse httpResponse,
-                        LazyContentChannel channel, long startTime) throws IOException {
+                        LazyContentChannel channel, long startTime) {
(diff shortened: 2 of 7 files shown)


container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
```diff
@@ -69,14 +69,14 @@ private static void populateHeaders(HeaderFields outputHeaders,
         }
     }
 
-    public ListenableFuture<Boolean> waitableRender(OutputStream stream) throws IOException {
```
bratseth: I don't think we can remove this throw because it breaks the API

container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
```diff
@@ -69,14 +69,14 @@ private static void populateHeaders(HeaderFields outputHeaders,
...
+    public ListenableFuture<Boolean> waitableRender(OutputStream stream) {
         return waitableRender(result, query, rendererCopy, stream);
     }
 
     public static ListenableFuture<Boolean> waitableRender(Result result,
                                                            Query query,
                                                            Renderer<Result> renderer,
-                                                           OutputStream stream) throws IOException {
```
bratseth: Same: I don't think we can remove this throw because it breaks the API

bratseth (changes requested):
Otherwise LGTM!

container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
baldersheim: Hmm, It did not fail the api checks......

container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
baldersheim: Hmm, It did not fail the api checks......

container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
baldersheim: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.21
13.4.21. Method and Constructor Throws
Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

container-search/src/main/java/com/yahoo/search/handler/HttpSearchResponse.java
baldersheim: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.21
13.4.21. Method and Constructor Throws
Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

bratseth:
But we also need compile-time compatibility.

baldersheim:
I added #12153 to ensure we are compatible. But why is this not detected in the api check ?
And removing throws should not give any compile errors. Adding will do if not caught or declared throws at usage, but removing should not.

bratseth:
If you remove a throws declaration from a method then usages which catches the exception will get a compile error, and subclasses cannot throw it.

Yes you are right the ABI checker should have caught this, because this is what we care about, but I guess since it is specifically an *ABI* checker it doesn't.

PR #12167, opened 2020-02-12
Only use 2 cores for config proxy

baldersheim:
@hmusum PR


hmusum (approved):
We have had issues earlier in config proxy (with file distribution) with using too few threads in thread pools. I'd like to review usage of `Runtime.getRuntime().availableProcessors()` in the code before merging this

PR #12171, opened 2020-02-13
Take load balancer lock per application

mpolden:
After this we can remove the global lock, which should fix the lock timeout
issues we occasionally see in us-west-1 during release.


hmusum (approved):
Great!

PR #12180, opened 2020-02-13
Do not depend on on ClusterInfo config as it changes to often and cau…

baldersheim:
…ses a instant clusterwide hickup on any container cluster changes like node retirement.

The corner case it was used for is not worth the cost.

@bratseth PR


baldersheim:
I will make a component out of the RpcResourcePool first

bratseth (approved):
Fine, but if something hurts you should generally try to do it more to make it stop.
We might consider warning about this kind of configuration on the config model side, but i agree it's not important.

baldersheim:
Yes, the root cause need to be fixed independently.

bjormel:
That was quick, thanks!

baldersheim:
I always have a few branch lying around, just in case...... :)

PR #12182, opened 2020-02-13
Splunk: Setting default application Id values in _meta

bjormel:

### DIFF
--- a/logforwarder/src/apps/vespa-logforwarder-start/cf-handler.cpp
+++ b/logforwarder/src/apps/vespa-logforwarder-start/cf-handler.cpp
@@ -77,17 +77,22 @@ CfHandler::doConfigure()
     fclose(fp);
     rename(tmpPath.c_str(), path.c_str());
 
-    path = cfFilePath(config.splunkHome, "inputs.conf");
-    tmpPath = path + ".new";
-    fp = fopen(tmpPath.c_str(), "w");
-    if (fp == NULL) return;
-
-    fprintf(fp, "[default]\n");
-    fprintf(fp, "host = %s\n", getenv("HOSTNAME"));
-    fprintf(fp, "_meta = vespa_tenant::%s vespa_application::%s vespa_instance::%s\n", getenv("VESPA_TENANT"), getenv("VESPA_APPLICATION"), getenv("VESPA_INSTANCE"));
-    fclose(fp);
-    rename(tmpPath.c_str(), path.c_str());
-
+    if (getenv("VESPA_HOSTNAME") != NULL &&
+        getenv("VESPA_TENANT") != NULL &&
+        getenv("VESPA_APPLICATION")!= NULL &&
+        getenv("VESPA_INSTANCE") != NULL )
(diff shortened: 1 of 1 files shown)


bjormel:
@arnej27959 would something like this work? 

The goal is to allow tenants to avoid having to maintain application specific information in their Splunk log forwarding apps.
With _meta specified in /opt/splunkforwarder/etc/system/local/inputs.conf this will be applied unless the customer has overridden -meta in their app.

bjormel:
https://git.ouroath.com/vespa/vespa-yahoo/pull/3618 to set the application id info as env variables in docker.

Current env variables:
`US-WEST-1 [bjormel@le04555:~]$ sudo docker exec le04555-v6-11  env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/y/bin:/home/y/sbin
HOSTNAME=le04555-v6-11.ostk.bm2.prod.gq1.yahoo.com
TERM=xterm
LANG=en_US.UTF-8
LANGUAGE=en_US.UTF-8
LC_ALL=en_US.UTF-8
HOME=/root`

bjormel:
The environment values are now available in CD:
`CD-US-CENTRAL-1 [root@17491290-v6-2:/]$ xargs --null --max-args=1 < /proc/$(pgrep vespa-runserver | head -1)/environ | grep -E "VESPA_(TENANT|APPLICATION|INSTANCE)"`
`VESPA_INSTANCE=default`
`VESPA_TENANT=vespa`
`VESPA_APPLICATION=canary-pipeline`

logforwarder/src/apps/vespa-logforwarder-start/cf-handler.cpp
```diff
@@ -76,6 +77,17 @@ CfHandler::doConfigure()
...
     rename(tmpPath.c_str(), path.c_str());
 
+    path = cfFilePath(config.splunkHome, "inputs.conf");
+    tmpPath = path + ".new";
+    *fp = fopen(tmpPath.c_str(), "w");
+    if (fp == NULL) return;
+
+    fprintf(fp, "[default]\n");
+    fprintf(fp, "host = %s\n", getenv("HOSTNAME"));
+    fprintf(fp, "_meta = vespa_tenant::%s vespa_application::%s vespa_instance::%s\n", getenv("VESPA_TENANT"), getenv("VESPA_APPLICTAION"), getenv("VESPA_INSTANCE"));
```
arnej27959: must be VESPA_APPLICATION here.
also, we should probably detect if any of these are unset (getenv returns null pointer), i guess...

arnej27959 (changes requested):
will push a commit fixing so it compiles...

bjormel:
@arnej27959 sorry, did not notice that you had fixed the typos, and ended up doing the change without pulling first, causing a mess.

Should I create a new branch/PR?

arnej27959 (approved):
looks better now :-)

PR #12183, opened 2020-02-13
extend crypto engine api

havardpe:
send spec for client connections to enable SNI as well as server name
verification

@vekterli please review
@bjorncs FYI


vekterli (approved):
:+1:

PR #12184, opened 2020-02-13
Create a resourcepool so that we do not need to reconnect to content …

baldersheim:
…cluster on changes to container cluster.

@bratseth PR
Let us do this one before #12180

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
@@ -59,6 +59,7 @@ public class ContainerSearch extends ContainerSubsystem<SearchChains>
         for (AbstractSearchCluster searchCluster : searchClusters) {
             if ( ! ( searchCluster instanceof IndexedSearchCluster)) continue;
             owningCluster.addComponent(new DispatcherComponent((IndexedSearchCluster)searchCluster));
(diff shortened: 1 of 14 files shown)


bratseth (approved):
💯

config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
```diff
@@ -59,6 +59,7 @@ private void initializeDispatchers(Collection<AbstractSearchCluster> searchClust
         for (AbstractSearchCluster searchCluster : searchClusters) {
             if ( ! ( searchCluster instanceof IndexedSearchCluster)) continue;
             owningCluster.addComponent(new DispatcherComponent((IndexedSearchCluster)searchCluster));
+            owningCluster.addComponent(new RpcResourcePoolComponent((IndexedSearchCluster)searchCluster));
```
bratseth: /source/config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java:[62,44] cannot find symbol
  symbol:   class RpcResourcePoolComponent
  location: class com.yahoo.vespa.model.container.search.ContainerSearch
[INFO] 1 error

I guess you mean RpcResourcePool

baldersheim: No the class name is correct, but I forgot to do 'git add'.

PR #12186, opened 2020-02-14
Revert "Only use 2 cores for config proxy"

hmusum:
Reverts vespa-engine/vespa#12167

System tests failing, one of the errors is:
`Improperly specified VM option 'ThreadStackSize=256-XX:MaxJavaStackTraceDepth=1000'`

, but some other failures as well


baldersheim:
👍

PR #12187, opened 2020-02-14
Reapply "Only use 2 cores for config proxy"

hmusum:
First commit is the same as https://github.com/vespa-engine/vespa/pull/12167, second commit fixes a missing space in logserver start script


aressem (approved):

baldersheim (commented):
👍

PR #12190, opened 2020-02-14
Jonmv/differentiate between prod and non prod instances in rest api

jonmv:


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/jobs.json
```diff
@@ -335,7 +335,7 @@
       "runs": [
         {
           "id": 1,
-          "url": "http://localhost:8080/application/v4/tenant/tenant1/application/application1/instance/instance1/job/run/run 1 of dev-us-east-1 for tenant1.application1.instance1",
+          "url": "http://localhost:8080/application/v4/tenant/tenant1/application/application1/instance/instance1/job?/run/run 1 of dev-us-east-1 for tenant1.application1.instance1",
```
freva: ?

freva (commented):
Is there anything that prevents a dev deployment to a production instance?

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/jobs.json
jonmv: Oops.

jonmv:
Yes, there is a check that manual deployments are not made to an environment that is not manually deployed.

freva:
Yeah, the environment part is fine, but I can have a production deployment on instance `default` and then deploy directly to `dev` also for instance `default` right? In that case, this API will return `default` as a production instance?

jonmv:
Yes. Removed that part, as it wasn't too useful.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/jobs.json
jonmv: The `?` is not the worst part — read the rest of the line!

freva (approved):

PR #12191, opened 2020-02-14
Twice as many stripes.

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestormanager.h
+++ b/storage/src/vespa/storage/persistence/filestorage/filestormanager.h
@@ -45,65 +45,43 @@ class FileStorManager : public StorageLinkQueued,
                         public framework::HtmlStatusReporter,
                         public StateListener,
                         private config::IFetcherCallback<vespa::config::content::StorFilestorConfig>,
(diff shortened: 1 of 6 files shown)


storage/src/vespa/storage/persistence/filestorage/filestormanager.h
```diff
@@ -45,65 +45,43 @@ class FileStorManager : public StorageLinkQueued,
                         public framework::HtmlStatusReporter,
                         public StateListener,
                         private config::IFetcherCallback<vespa::config::content::StorFilestorConfig>,
-                        private MessageSender
+                        public MessageSender
```
vekterli: Just curious why this was changed from `private` to `public`

vekterli (approved):
:+1:

storage/src/vespa/storage/persistence/filestorage/filestormanager.h
baldersheim: If not it not possible to construct it with std::make_unique<>, as the MessageSender interface provided by *this would be private.

PR #12196, opened 2020-02-14
Allow count attribute on persistence-threads

baldersheim:
@vekterli PR


vekterli (approved):
:+1: seems like this was already supported in the config model then?

baldersheim:
Yes, but not tested and not allowed by schema....

PR #12164, opened 2020-02-12
Add config class for the future Telegraf component

gjoranv:

PR #12199, opened 2020-02-14
Added Telegraf component

olaaun:
Based on the config file defined in https://github.com/vespa-engine/vespa/pull/12164


metrics-proxy/pom.xml
```diff
@@ -132,6 +132,11 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.velocity</groupId>
+      <artifactId>velocity</artifactId>
+      <scope>compile</scope>
```
gjoranv: Please skip explicit compile scope, it is the default.

metrics-proxy/src/main/java/ai/vespa/metricsproxy/telegraf/TelegrafRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.List;
+
+/**
+ * @author olaa
+ */
+public class TelegrafRegistry {
+
+    private static final List<Telegraf> telegrafInstances = Collections.synchronizedList(new ArrayList<>());
+
+    public void addInstance(Telegraf telegraf) {
```
gjoranv: All methods must be synchronized, as they are accessed from multiple threads.

metrics-proxy/src/main/resources/templates/cloudwatch_plugin.vm
```diff
@@ -0,0 +1,30 @@
...
+
+# Configuration for AWS CloudWatch output.
+[[outputs.cloudwatch]]
+  region = "$cloudwatchRegion"
+  namespace = "$cloudwatchNamespace"
+#if( $isHosted )
+  access_key = "$cloudwatchSecretKey"
+  secret_key = "$cloudwatchAccessKey"
+#else
+  profile = "$cloudwatchProfile"
```
gjoranv: This should only be output if the profile is non-empty. Is that implicitly being taken care of?

metrics-proxy/src/main/resources/templates/cloudwatch_plugin.vm
```diff
@@ -0,0 +1,30 @@
...
+#end
+
+# Configuration for Vespa input plugin
+[[inputs.vespa]]
+  url = "http://localhost:19092/metrics/v2/values?consumer=$vespaConsumer"
+#if( $isHosted )
+  tls_cert = "${VESPA_CERTIFICATE_PATH}"
+  tls_key = "${VESPA_KEY_PATH}"
+  insecure_skip_verify = true
+#end
```
gjoranv: Please add trailing newline.

metrics-proxy/src/main/java/ai/vespa/metricsproxy/telegraf/Telegraf.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static com.yahoo.yolean.Exceptions.uncheck;
+
+/**
+ * @author olaa
+ */
+public class Telegraf extends AbstractComponent {
+
+    private static final String TELEGRAF_CONFIG_PATH = "/etc/telegraf/telegraf.conf";
+    private static final String TELEGRAF_CONFIG_TEMPLATE_PATH = "src/main/resources/templates/cloudwatch_plugin.vm";
```
gjoranv: We probably need to consider these file paths, especially where to read the template path from.

metrics-proxy/pom.xml
```diff
@@ -132,6 +132,11 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.velocity</groupId>
+      <artifactId>velocity</artifactId>
```
gjoranv: Please add an exclude for this artifact in config-model/pom.xml's dependency on metrics-proxy. It adds an extra 0.5 MB.

gjoranv (approved):

PR #12201, opened 2020-02-14
Default number of threads should be a multiple of number of stripes, …

baldersheim:
…which is now 4

@vekterli PR


vekterli (approved):
:+1: doc should probably also be updated, as I believe it says 6 threads are the default

PR #12208, opened 2020-02-15
Disallow configuring the wrong port when hosted

bratseth:


bratseth:
This is the second stage of https://github.com/vespa-engine/vespa/pull/11837

bjorncs (approved):

PR #12209, opened 2020-02-15
Since we are accessing the handler while under read/writer lock, ther…

baldersheim:
…e is no reason to use a shared_ptr.

@toregge PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/common/handlermap.hpp
+++ b/searchcore/src/vespa/searchcore/proton/common/handlermap.hpp
@@ -115,7 +115,7 @@ public:
 
     /**
      * Returns the handler for the given document type. If no handler was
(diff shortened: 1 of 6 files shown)


toregge (approved):

searchcore/src/vespa/searchcore/proton/common/handlermap.hpp
```diff
@@ -115,6 +113,20 @@ class HandlerMap {
         return HandlerSP();
     }
 
+    /**
+     * Returns the handler for the given document type. If no handler was
+     * registered, this method returns an empty shared pointer.
```
toregge: nullptr instead of empty shared pointer ?

baldersheim: Fixed

PR #12210, opened 2020-02-15
Rename field name -> jobName to follow /deployment api

freva:
Rename job name field used for dev deployments to be the same as used in /deployment API:
https://github.com/vespa-engine/vespa/blob/20f84687ed1362a7f76404e2f20a6619fca05119/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelper.java#L630


jonmv (approved):

jonmv:
Thanks.

PR #12217, opened 2020-02-17
Update InternalStepRunner.java

kkraune:
The current output is a bit misleading:
11:17:06 Restarting services on host h478a...
11:17:06 Restarting services on host h479a...
11:17:06 Deployment successful.

it looks like the restart are completed


jonmv (approved):

jonmv:
Mm, good point :)

PR #12218, opened 2020-02-17
Add tenant access rules for /routing/v1/

mpolden:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
```diff
@@ -53,7 +55,8 @@
                Optional.of("/api"),
                "/application/v4/tenant/{tenant}/application/",
                "/application/v4/tenant/{tenant}/cost",
-               "/application/v4/tenant/{tenant}/cost/{date}"),
+               "/application/v4/tenant/{tenant}/cost/{date}",
+               "/routing/v1/status/tenant/{tenant}"),
```
jonmv: Seems to me everything under `/status/` is `GET`, and could lie here, with `.../{tenant}/{*}`.

jonmv:
Please see comment about path groups. These are another thing we should clean up, ref. application api package.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
mpolden: Thanks!

mpolden:
PTAL

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
```diff
@@ -153,7 +156,8 @@
                          "/application/v4/tenant/{tenant}/application/{application}/environment/test/region/{region}/instance/{ignored}",
                          "/application/v4/tenant/{tenant}/application/{application}/environment/test/region/{region}/instance/{ignored}/deploy",
                          "/application/v4/tenant/{tenant}/application/{application}/environment/staging/region/{region}/instance/{ignored}",
-                         "/application/v4/tenant/{tenant}/application/{application}/environment/staging/region/{region}/instance/{ignored}/deploy"),
+                         "/application/v4/tenant/{tenant}/application/{application}/environment/staging/region/{region}/instance/{ignored}/deploy",
+                         "/routing/v1/inactive/tenant/{tenant}/application/{application}/instance/{ignored}/environment/prod/region/{region}"),
```
jonmv: Ah, there was one `/inactive/` as well? So I guess this is to allow application owners to manipulate endpoint status? The `athenzTenantAdmin` does not cover this group — it was split out specifically for the legacy pipeline. Putting this under the `applicationInfo` should work for both sets of roles (Athenz and otherwise).

mpolden: Done. There's too many groups!

jonmv: Yes, should be able to remove some already now.

jonmv (approved):
Great!

PR #12221, opened 2020-02-17
Production tests must start after a successful deployment

jonmv:
@mpolden please review and merge. This fixes an error where a production test would be marked as complete as long as it was on the correct version, and _any_ production deployment had previously completed

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobList.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobList.java
@@ -141,24 +141,9 @@ public class JobList extends AbstractFilteringList<JobStatus, JobList> {
             return mapToList(Function.identity());
         }
 
-        /** Returns the subset of jobs where the run of the given type occurred before the given instant */
-        public JobList endedBefore(Instant threshold) {
-            return matching(run -> run.end().orElse(Instant.MAX).isBefore(threshold));
-        }
-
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobList.java
```diff
@@ -141,24 +141,9 @@ public JobList present() {
...
-        /** Returns the subset of jobs where the run of the given type occurred before the given instant */
-        public JobList startedBefore(Instant threshold) {
-            return matching(run -> run.start().isBefore(threshold));
-        }
-
-        /** Returns the subset of jobs where the run of the given type occurred after the given instant */
-        public JobList startedAfter(Instant threshold) {
-            return matching(run -> run.start().isAfter(threshold));
+        /** Returns the subset of jobs where the run of the given type occurred at or before the given instant */
+        public JobList endedNoLaterThan(Instant threshold) {
```
mpolden: Consider naming `endedAtOrBefore` to match doc.

Nit: Javadoc says "of the given type", but there is no type given here that I can tell.

mpolden (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobList.java
jonmv: Makes sense :)

PR #12225, opened 2020-02-17
Use enable_shared_from_this to simplify interface and keep the detail…

baldersheim:
…s local.

@havardpe PR


searchcore/src/vespa/searchcore/proton/server/searchview.cpp
```diff
@@ -161,8 +161,9 @@ SearchView::getDocsumsInternal(const DocsumRequest & req)
 }
 
 std::unique_ptr<SearchReply>
-SearchView::match(const ISearchHandler::SP &self, const SearchRequest &req, ThreadBundle &threadBundle) const {
-    return _matchView->match(self, req, threadBundle);
+SearchView::match(const SearchRequest &req, ThreadBundle &threadBundle) const {
+    ISearchHandler::SP self = const_cast<SearchView *>(this)->shared_from_this();
```
havardpe: I do not like this const-cast. Maybe this function should not be const (since it is allowed to change the object state).

searchcore/src/vespa/searchcore/proton/server/searchview.h
```diff
@@ -8,7 +8,7 @@
 
 namespace proton {
 
-class SearchView : public ISearchHandler
+class SearchView : public ISearchHandler, public std::enable_shared_from_this<ISearchHandler>
```
havardpe: A class inheriting from enable_shared_from_this should pass itself as template parameter.

Also consider making the constructors private and replace them with static create functions returning shared pointers to ensure that objects are actually owned by shared pointers.

havardpe (commented):
do we really need to send a shared pointer to a non-const object? if so, consider making the function itself non-const as well since calling it might lead to the object being modified.

searchcore/src/vespa/searchcore/proton/server/searchview.h
baldersheim: Fixed.

searchcore/src/vespa/searchcore/proton/server/searchview.cpp
baldersheim: Const cast was not necessary. Just passing a const ISearchHandler was sufficient.

PR #12226, opened 2020-02-17
Define enable-fleet-sshd-config flag

hakonhall:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -55,6 +55,12 @@
             "Whether to enable Nessus.", "Takes effect on next host admin tick",
             HOSTNAME);
 
+    public static final UnboundBooleanFlag ENABLE_FLEET_SSHD_CONFIG = defineFeatureFlag(
+            "enable-fleet-sshd-config", false,
+            "Whether fleet should manage the /etc/ssh/sshd_config file.",
+            "Takes effect on next host admin tick.",
+            HOSTNAME);
+
```
smorgrav: LGTM - but we already are opt in here - will this revert that when applied by default?

smorgrav (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hakonhall: Yes. So I'm thinking we should remove the opt-in, and later opt-in through our known rollout mechanisms (new Vespa version or flag flip).

PR #12228, opened 2020-02-17
[VESPA-15448] Add metrics for client side failures

yngveaasheim:

@vekterli Please review and merge.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
@@ -615,18 +615,24 @@ public class VespaMetricSet {
         metrics.add(new Metric("vds.distributor.puts.sum.latency.average")); // TODO: Remove in Vespa 8
         metrics.add(new Metric("vds.distributor.puts.sum.ok.rate"));
         metrics.add(new Metric("vds.distributor.puts.sum.failures.total.rate"));
+        metrics.add(new Metric("vds.distributor.puts.sum.failures.notfound.rate"));
+        metrics.add(new Metric("vds.distributor.puts.sum.failures.test_and_set_failed"));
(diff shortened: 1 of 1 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
```diff
@@ -636,12 +642,14 @@
...
         metrics.add(new Metric("vds.distributor.gets.sum.ok.rate"));
         metrics.add(new Metric("vds.distributor.gets.sum.failures.total.rate"));
+        metrics.add(new Metric("vds.distributor.gets.sum.failures.notfound.rate"));
         metrics.add(new Metric("vds.distributor.visitor.sum.latency.max"));
         metrics.add(new Metric("vds.distributor.visitor.sum.latency.sum"));
         metrics.add(new Metric("vds.distributor.visitor.sum.latency.count"));
         metrics.add(new Metric("vds.distributor.visitor.sum.latency.average")); // TODO: Remove in Vespa 8
         metrics.add(new Metric("vds.distributor.visitor.sum.ok.rate"));
         metrics.add(new Metric("vds.distributor.visitor.sum.failures.total.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.notfound.rate"));
```
vekterli: Visitors operate on bucket granularities and `notfound` is a document-level metric, so this should never be nonzero. I.e. this can be removed.

vekterli (approved):
:+1: see comment for consideration

config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
yngveaasheim: Thanks! Removed.

vekterli (approved):

PR #12229, opened 2020-02-17
Add imported fields to generated concrete document type

vekterli:
@baldersheim please review. Tested locally, but a system test should be added. Unit testing of this is currently, uh, challenging.

Otherwise, code that needs to have knowledge of imported fields to
operate properly (such as `DocumentRouteSelector`) would fail
to observe these. This would in turn cause feeding of concrete
document types to fail when document selections mention imported
fields.


baldersheim (approved):
Test can be added to documentgen-test module.

PR #12230, opened 2020-02-17
Bjorncs/tls hostname validation jrt wiring 

bjorncs:
FYI @vekterli

### DIFF
--- a/security-utils/src/main/java/com/yahoo/security/tls/HostnameVerification.java
+++ b/security-utils/src/main/java/com/yahoo/security/tls/HostnameVerification.java
@@ -1,0 +1,7 @@
+// Copyright 2020 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security.tls;
+
+/**
+ * @author bjorncs
+ */
+public enum HostnameVerification { ENABLED, DISABLED }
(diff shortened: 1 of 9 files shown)


security-utils/src/main/java/com/yahoo/security/tls/HostnameVerification.java
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security.tls;
+
+/**
+ * @author bjorncs
+ */
+public enum HostnameVerification { ENABLED, DISABLED }
```
vekterli: Should this be `HostnameValidation` to be in line with the config?

vekterli (approved):
:+1:

security-utils/src/main/java/com/yahoo/security/tls/HostnameVerification.java
bjorncs: The concept is called hostname verification in Java hence the name. I agree this could be renamed to be more in line with config name.

vekterli: It makes sense to use the concept that most closely maps to what's used on the underlying platform, so let's stick with verification for now.

PR #12237, opened 2020-02-18
Let only successful run expire due to age

jonmv:
@mpolden please review.

This fixes a problem where a NoSuchElementException was thrown when the
only run was a success, this expired, and a new run failed.
The solution is simply to throw away the success in this case, as we
have no logs for it anyway

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunnerTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunnerTest.java
@@ -338,6 +338,25 @@ public class JobRunnerTest {
     }
 
     @Test
+    public void onlySuccessfulRunExpiresThenAnotherFails() {
+        DeploymentTester tester = new DeploymentTester();
+        JobController jobs = tester.controller().jobController();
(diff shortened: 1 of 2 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunnerTest.java
```diff
@@ -337,6 +337,25 @@ public void historyPruning() {
         assertFalse(jobs.firstFailing(jobId).isPresent());
     }
 
+    @Test
+    public void onlySuccessfulRunExpiresThenAnotherFails() {
+        DeploymentTester tester = new DeploymentTester();
+        JobController jobs = tester.controller().jobController();
```
mpolden: To var or not, that is the question.

mpolden (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunnerTest.java
jonmv: Yeah, I'm not settled on that one yet :)

PR #12248, opened 2020-02-18
fail faster and record when missing endpoint cert

andreer:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -273,9 +275,14 @@ public InternalStepRunner(Controller controller) {
             Optional<RunStatus> result = startTime.isBefore(controller.clock().instant().minus(Duration.ofHours(1)))
                                          ? Optional.of(deploymentFailed) : Optional.empty();
             switch (e.getErrorCode()) {
+                case CERTIFICATE_NOT_READY:
+                    if (startTime.plus(endpointCertificateTimeout).isBefore(controller.clock().instant())) {
+                        logger.log("Deployment failed to find provisioned endpoint certificate after " + endpointCertificateTimeout);
+                        return Optional.of(RunStatus.endpointCertificateTimeout);
```
jonmv:
```suggestion
                        return Optional.of(endpointCertificateTimeout);
```

jonmv: You statically imported it already :)

jonmv (approved):
Yes, very good! :)

PR #12259, opened 2020-02-19
Revert "Revert "Revert "Revert "Create a resourcepool so that we do not need to reconnect to content .…""""

baldersheim:
Reverts vespa-engine/vespa#12258


baldersheim:
@gjoranv PR

gjoranv (approved):

Issue #12256, opened 2020-02-19
vespa-http-client-jar-with-dependencies.jar null pointer exception


dodysw2:
When reading from stdin, with `--verbose` param, the client throws exception:
```
Exception in thread "main" java.lang.NullPointerException
        at java.base/java.io.File.<init>(File.java:278)
        at com.yahoo.vespa.http.client.runner.Runner.main(Runner.java:89)
```
It seems, there should be null check on `commandLineArgs.getFile()`

https://github.com/vespa-engine/vespa/blob/2686f79b2df3e52e2d43521db4c0b09597c07e73/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/runner/Runner.java#L89

jobergum:
Thank you for the detailed report, we will fix this.

olaaun:
Thanks for pointing this out @dodysw2. A fix has been merged and will be available in a release with version >= 7.181.5

PR #12260, opened 2020-02-19
Read input size from FormatInputStream instead of file 

olaaun:
Addresses #12256


freva (commented):
Have you tested this? Looking at the [documentation](
https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#available()), I'd assume that this would return a number <= file size since at some point, for large files, it would have to do some blocking IO to fetch more data?

olaaun:
> Have you tested this? Looking at the [documentation](https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#available()), I'd assume that this would return a number <= file size since at some point, for large files, it would have to do some blocking IO to fetch more data?

Good point. I did test it, but only on smaller files. I guess we can keep getting input size from file, and skip printing those stats when using stdin

freva (approved):

PR #12271, opened 2020-02-19
Use a static connectionpool repo to reduce number of threads.

baldersheim:
@hmusum PR

### DIFF
--- a/config/src/main/java/com/yahoo/config/subscription/impl/JRTManagedConnectionPools.java
+++ b/config/src/main/java/com/yahoo/config/subscription/impl/JRTManagedConnectionPools.java
@@ -1,0 +1,56 @@
+package com.yahoo.config.subscription.impl;
+
+import com.yahoo.config.subscription.ConfigSourceSet;
+import com.yahoo.vespa.config.JRTConnectionPool;
+import com.yahoo.vespa.config.TimingValues;
+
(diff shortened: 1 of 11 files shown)


config/src/main/java/com/yahoo/config/subscription/impl/JRTManagedConnectionPools.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+    private Map<ConfigSourceSet, CountedPool> pools = new HashMap<>();
+
+    public synchronized JRTConfigRequester acquire(ConfigSourceSet sourceSet, TimingValues timingValues) {
+        CountedPool countedPool = pools.get(sourceSet);
+        if (countedPool == null) {
+            countedPool = new CountedPool(new JRTConnectionPool(sourceSet));
+            pools.put(sourceSet, countedPool);
+        }
+        countedPool.count++;
```
hmusum: Won't count be 2 then for a newly created CountedPool? So ContedPool should either start count on 0 or only increment in this place when countedPool is not null?

baldersheim: Good catch. Test added and fixed.

hmusum (approved):

PR #12283, opened 2020-02-20
Control -XX:CompressedClassSpaceSize jvm option.

baldersheim:
@hmusum PR


config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
```diff
@@ -114,7 +114,6 @@
     public static final String STATE_HANDLER_CLASS = "com.yahoo.container.jdisc.state.StateHandler";
     public static final String LOG_HANDLER_CLASS = com.yahoo.container.handler.LogHandler.class.getName();
     public static final String DEFAULT_LINGUISTICS_PROVIDER = "com.yahoo.language.provider.DefaultLinguisticsProvider";
-    public static final String CMS = "-XX:+UseConcMarkSweepGC -XX:MaxTenuringThreshold=15 -XX:NewRatio=1";
```
hmusum: `CMS` is used in YahooContainerModelAmender

baldersheim: hrmf, fixed

hmusum (approved):

PR #12286, opened 2020-02-20
Guard against out of bounds stale matching element.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (commented):
Overall looks good, but please add out-of-bounds unit tests in matched_elements_filter_test.cpp. One case for map and one for array. Should be pretty easy as setup is already there.

geirst (approved):
👍

PR #12292, opened 2020-02-20
Move crypto utility code out into vespalib and use for test credentials

vekterli:
@havardpe please review
@bjorncs FYI

Currently offers only the following functionality:
* Generate P-256 EC private keys and export to PEM
* Generate X509v3 certificates and export to PEM

Instead of using hardcoded private key/certs for unit tests, use crypto
utility code to generate new credentials once per test process. Since
these certs now use a SAN of `localhost` it also means we no longer need
to disable hostname validation for networked unit tests.


vekterli:
Travis build failure is due to an unrelated change

havardpe (approved):

havardpe (approved):

PR #12293, opened 2020-02-20
Arnej/use nns index from blueprint

arnej27959:

@geirst please review
@havardpe @baldersheim FYI


geirst (approved):
👍

havardpe (commented):
The hit estimate is an issue here, since indicating that we match all documents will make us non-strict in most cases. However, we cannot claim that we match only k documents unless we check that an index is available up front.

PR #12296, opened 2020-02-20
Consider reserved nodes when finding the highest cluster index

gjoranv:
FYI: @freva

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
@@ -121,9 +121,7 @@ class Preparer {
      */
     private int findHighestIndex(ApplicationId application, ClusterSpec cluster) {
         int highestIndex = -1;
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -122,7 +122,8 @@ private void moveToActiveGroup(List<Node> surplusNodes, int wantedGroups, Option
     private int findHighestIndex(ApplicationId application, ClusterSpec cluster) {
         int highestIndex = -1;
         for (Node node : nodeRepository.getNodes(application,
-                                                 Node.State.active, Node.State.inactive, Node.State.parked, Node.State.failed)) {
+                                                 Node.State.reserved, Node.State.active, Node.State.inactive,
```
mpolden: This should just include all states for which `Node.State#isAllocated` is `true`.

freva (approved):

bratseth (approved):

PR #12297, opened 2020-02-20
Add one entrypoint as a POC 

oyving:
@bjorncs - please sanity check my approach. Do not merge yet. If we agree on the approach I'll add entrypoints to more classes.

I am not entirely happy with how we are doing this, so I am _very_ open to suggestions. Basically I am trying to:

1. Take the request URI (which should always be against `/`).
2. Try to guess a path from the binding and add it to the URI.
3. Take the entrypoint from the handler and add it to the URI.
4. Test if the URI matches the binding.
5. If it does, add it to the list of `links` we output.


container-disc/src/main/java/com/yahoo/container/usability/BindingsOverviewHandler.java
```diff
@@ -99,6 +104,76 @@ static JSONArray renderRequestHandlers(JdiscBindingsConfig bindingsConfig,
         return ret;
     }
 
+    static JSONArray renderLinks(URI requestUri,
+                                 JdiscBindingsConfig bindingsConfig,
+                                 Map<ComponentId, ? extends RequestHandler> handlersById)
+    {
+        var entryPointSet = new LinkedHashSet<URI>();
```
bjorncs: What do you think about having these URIs sorted lexically?

container-core/src/main/java/com/yahoo/container/jdisc/HandlerEntryPoint.java
```diff
@@ -0,0 +1,19 @@
+package com.yahoo.container.jdisc;
+
+import java.net.URI;
+
+public class HandlerEntryPoint {
```
bjorncs: Consider adding optional description and/or category to be listed with the handler URIs. This could also be used to aggregate the links somehow.

bjorncs (approved):
I like the overall solution 👍  See comments for minor adjustments/enhancements.

container-core/src/main/java/com/yahoo/container/jdisc/NavigableRequestHandler.java
```diff
@@ -0,0 +1,7 @@
+package com.yahoo.container.jdisc;
+
+import java.util.List;
+
+public interface NavigableRequestHandler {
+    List<HandlerEntryPoint> entryPoints();
```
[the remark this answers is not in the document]
oyving: That was the idea, at least.

container-core/src/main/java/com/yahoo/container/jdisc/HandlerEntryPoint.java
oyving: We want more information here as we go, but for now I just want to get out the URLs.

container-disc/src/main/java/com/yahoo/container/usability/BindingsOverviewHandler.java
oyving: I don't know! :)

bjorncs: Ok, let's to the current ordering for the initial approach.

bjorncs (approved):

bjorncs:
Btw, you need to update the ABI specification for container-core. Run `mvn package -Dabicheck.writeSpec -pl container-core` and add the result as another commit.

baldersheim:
Still relevant ? or close ?

PR #12305, opened 2020-02-21
Bjorncs/stabilize log handler test

bjorncs:


container-accesslogging/pom.xml
```diff
@@ -65,6 +65,11 @@
       <artifactId>jackson-databind</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
```
baldersheim: Another assert framework ? @bratseth and @hmusum 
Won't junit suffice. I thought we are getting rid of the hamcrest stuff. This seems like a replacement.

bjorncs: assertj is a modern incarnation of fest-assert. It's api is way better than hamcrest in my opinion. The simple assert methods from junit does not provide good error messages on failure.

bratseth: You're adding a new dependency to be able to write assertThat(31).isEqualTo(first) instead of assertEquals(31, first).

I suppose there's nothing I can say which adds helpfully to what immediately should become apparent to somebody reading that sentence, and this isn't important so feel free to do whatever you want, but I'll keep removing these when I encounter them in stuff I'm working on.

bjorncs: I agree that assertj is overkill for assert-a-equals-b kind of assertions. 

Assertj is handy for writing more complex assertions in a type safe manner. Examples from this PR:
`assertThat(content).hasLineCount(logEntries);`
`assertThat(uncompressed).exists();`
`assertThat(compressed).doesNotExist();`
`assertThat(f1).startsWith("./testcompression/logfilehandlertest.");`

bratseth: All those can be done with a single line of code.

bjorncs: Yes, they were all one-liners in the original code. The new assertj statements are more readable in my opinion + provide better error messages on failure.

bjorncs: Good error messages are a great aid when hunting bugs that are hard to reproduce locally. For instance, the generated error message from `startsWith` will include the content of the actual string and the expected prefix. The original `assertTrue` expression would only output something like "expected true but got false".

bratseth: Yes, some times it is worth writing a few lines of code to make things more understandable, perhaps a custom assert method in the test. In a subset of those cases it is possible to find a library that already implements the assert method you have in mind. But I bet you wouldn't pull it in to avoid writing a few lines of code in any other circumstamnces.

hmusum (approved):

PR #12314, opened 2020-02-23
Define completeness of SuperModel and DuperModel

hakonhall:
In order for Orchestrator to remove application data from ZooKeeper, it must
know which applications do NOT exist. Since the duper model starts with 0
applications, always, the only way of knowing what applications do not exist is
for the bootstrap code to notify the super model/duper model when bootstrap is
complete. There are 2 sources of applications that must signal completeness:

 - The super model, once all applications have been redeployed in
   ConfigServerBootstrap.
 - The infrastructure application, in the InfrastructureProvisioner the first
   time it runs.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfrastructureProvisioner.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfrastructureProvisioner.java
@@ -2,6 +2,7 @@
 package com.yahoo.vespa.hosted.provision.maintenance;
 
 import com.yahoo.config.provision.InfraDeployer;
+import com.yahoo.log.LogLevel;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
 
 import java.time.Duration;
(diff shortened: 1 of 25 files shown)


hakonhall:
FYI @freva

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/InfraDeployerImpl.java
```diff
@@ -51,9 +51,19 @@ public InfraDeployerImpl(NodeRepository nodeRepository, Provisioner provisioner,
...
+            var deployment = new InfraDeployment(api);
+            try {
+                deployment.activate();
+            } catch (RuntimeException e) {
+                logger.log(LogLevel.INFO, "Failed to activate " + application, e);
+                // loop around to activate the next application
+            }
+        });
+
+        duperModel.infraApplicationsIsNowComplete();
```
freva: What if `activate()` throws?

hakonhall: I'm not sure. On one hand we do not want to mark the duper model as complete. But on the other hand, it would be wrong to start the config server without the required infra apps. But failing construction might also be a bit dangerous if the failure is triggered by zk data it is hard to inspect as the zk server is brought down if the config server goes down.

So we should trigger deploy in InfrastructureProvisioner constructor and let RuntimeException propagate there (but not in maintain()) after a delay of say 1m, to fail config server startup?

hmusum (approved):

hakonhall:
I'll wait for feedback from @freva

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfrastructureProvisioner.java
```diff
@@ -23,17 +23,22 @@
     InfrastructureProvisioner(NodeRepository nodeRepository, InfraDeployer infraDeployer, Duration interval) {
         super(nodeRepository, interval);
         this.infraDeployer = infraDeployer;
+
+        // If this fails, we fail the component graph construction and bootstrap of config server,
+        // which is what we want.
+        try {
+            infraDeployer.activateAllSupportedInfraApplications(true);
```
freva: Consider moving this to `NodeRepositoryMaintenance` or remove https://github.com/vespa-engine/vespa/blob/6c122d41cb1d890c285ffd1786eb7cc16f8eba55/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java#L90

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfrastructureProvisioner.java
hakonhall: Nice catch. I renamed and kept the call from NodeRepositoryMaintenance().

freva (approved):

PR #12317, opened 2020-02-24
Remove use of fest-assert

bjorncs:
Motivation: remove the number of 'fluent test assertion' libraries in use.


bratseth (approved):
👍 👍 👍

PR #12318, opened 2020-02-24
reprovision endpoint cert when necessary

andreer:
WIP - needs testing

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
@@ -41,6 +41,7 @@ import com.yahoo.vespa.hosted.controller.api.integration.organization.Mail;
 import com.yahoo.vespa.hosted.controller.application.ApplicationPackage;
 import com.yahoo.vespa.hosted.controller.application.Deployment;
 import com.yahoo.vespa.hosted.controller.application.TenantAndApplicationId;
+import com.yahoo.vespa.hosted.controller.certificate.EndpointCertificateException;
 import com.yahoo.vespa.hosted.controller.maintenance.JobRunner;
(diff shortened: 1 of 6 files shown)


andreer:
@jonmv Please review change to InternalStepRunner

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -299,6 +300,20 @@ public InternalStepRunner(Controller controller) {
...
+                    Optional<RunStatus> result = startTime.isBefore(controller.clock().instant().minus(Duration.ofHours(1)))
+                            ? Optional.of(deploymentFailed) : Optional.empty();
+                    if (startTime.plus(endpointCertificateTimeout).isBefore(controller.clock().instant())) {
+                        logger.log("Deployment failed to find provisioned endpoint certificate after " + endpointCertificateTimeout);
+                        return Optional.of(RunStatus.endpointCertificateTimeout);
+                    }
+                    return result;
+                default:
+                    throw e; // Should be surfaced / fail deployment
+            }
```
jonmv: Since the cert timeout is lower than the general 1h timeout, this should be simplified :)

jonmv: (Drop computing `result` and just return `Optional.empty()` in its place.

jonmv (changes requested):

andreer:
@jonmv: Do we need to make any matching change to the maven plugin?

jonmv:
No, that's just a dumb client tailing the logs.

jonmv:
It's possible you may need to update the integration test client, but I'm working on killing that one right now ... so rather wait :)

jonmv (approved):
Very good :)

tokle (approved):

Issue #12139, opened 2020-02-11
Improved insights into cluster GC operations


neuhausler:
Provide statistics and/or log entries about GC operations.

vekterli:
A new count metric has been added to the distributor processes which tracks the number of documents removed by background GC operations:
```
vds.idealstate.garbage_collection.documents_removed
```

neuhausler:
Perfect, thanks!  .. btw, like the term "idealstate".

PR #12320, opened 2020-02-24
Add count metric for number of documents garbage collected

vekterli:
@geirst please review. Moving StorageAPI to protobuf keeps paying off with dividends 🦄✨ 

New distributor metric available as:
```
vds.idealstate.garbage_collection.documents_removed
```
Add documents removed statistics to `RemoveLocation` responses,
which is what GC is currently built around. Could technically have
been implemented as a diff of before/after BucketInfo, but GC is
very low priority so many other mutating ops may have changed the
bucket document set in the time span between sending the GC ops
and receiving the replies.

This relates to issue #12139


geirst (approved):
👍

PR #12321, opened 2020-02-24
Arnej/rework ann filter bm

arnej27959:

@geirst please review

### DIFF
--- a/eval/src/tests/ann/gist_benchmark.cpp
+++ b/eval/src/tests/ann/gist_benchmark.cpp
@@ -18,167 +18,10 @@
 #include "nns.h"
 #include "for-sift-hit.h"
 #include "for-sift-top-k.h"
-
-std::vector<TopK> bruteforceResults;
-
-struct PointVector {
-    float v[NUM_DIMS];
-    using ConstArr = vespalib::ConstArrayRef<float>;
-    operator ConstArr() const { return ConstArr(v, NUM_DIMS); }
-};
-
-static PointVector *aligned_alloc(size_t num) {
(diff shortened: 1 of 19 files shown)


eval/src/tests/ann/gist_benchmark.cpp
```diff
@@ -0,0 +1,295 @@
...
+#define NUM_Q 1000
+
+#include "doc_vector_access.h"
+#include "nns.h"
+#include "for-sift-hit.h"
+#include "for-sift-top-k.h"
+
+std::vector<TopK> bruteforceResults;
+
+struct PointVector {
```
geirst: I don't think we should rely on this amount of copy-paste-mutate. A lot of this code is already in sift_benchmark.cpp, the only difference being the number of dimensions. Please move the following to a common place: PointVector, DocVectorAdapter, computeDistance, read_queries, read_docs, read_data, BfHitComparator, BfHitHeap. I am afraid this code will be hard to maintain otherwise.

geirst (commented):
Overall looks good, but I think we should remove some of the code duplication first. Otherwise I am afraid this code will be hard to maintain.

arnej27959:
@geirst PTAL

geirst:
Overall looks very good, but you forgot to add "hnsw-like.h":
`/source/eval/src/tests/ann/xp-hnswlike-nns.cpp:3:10: fatal error: hnsw-like.h: No such file or directory
 #include "hnsw-like.h"
          ^~~~~~~~~~~~~
compilation terminated.`

PR #12322, opened 2020-02-24
Optimize for bool attributes. 

baldersheim:
@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/features/attributefeature.cpp
+++ b/searchlib/src/vespa/searchlib/features/attributefeature.cpp
@@ -351,7 +351,7 @@ private:
 };
 
 fef::FeatureExecutor &
-createAttributeExecutor(const IAttributeVector *attribute, const vespalib::string &attrName, const vespalib::string &extraParam, vespalib::Stash &stash)
+createAttributeExecutor(uint32_t numOutputs, const IAttributeVector *attribute, const vespalib::string &attrName, const vespalib::string &extraParam, vespalib::Stash &stash)
 {
     if (attribute == nullptr) {
         LOG(warning, "The attribute vector '%s' was not found in the attribute manager, returning default values.",
@@ -381,7 +381,7 @@ createAttributeExecutor(const IAttributeVector *attribute, const vespalib::strin
             { SingleValueExecutorCreator<IntegerAttributeTemplate<int64_t>> creator;      if (creator.handle(attribute)) return creator.create(stash); }
             {
                 auto boolAttribute = dynamic_cast<const SingleBoolAttribute *>(attribute);
-                if (boolAttribute) {
+                if (boolAttribute && (numOutputs == 1)) {
                     return stash.create<BoolAttributeExecutor>(*boolAttribute);
                 }
             }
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/features/attributefeature.cpp
```diff
@@ -468,7 +488,10 @@ AttributeBlueprint::setup(const fef::IIndexEnvironment & env,
                    "the value at the given index of an array attribute, "
                    "the given key of a weighted set attribute, or"
                    "the tensor of a tensor attribute", output_type);
-    if (!_tensorType.is_tensor()) {
+    const fef::FieldInfo * fInfo = env.getFieldByName(_attrName);
+    if (!_tensorType.is_tensor() &&
+        !((fInfo->collection() == schema::CollectionType::SINGLE) && (fInfo->get_data_type() == schema::DataType::BOOL)))
```
havardpe: please simplify this condition to make it clearer what we are checking

searchlib/src/vespa/searchlib/features/attributefeature.cpp
```diff
@@ -365,6 +379,12 @@ createAttributeExecutor(const IAttributeVector *attribute, const vespalib::strin
             { SingleValueExecutorCreator<IntegerAttributeTemplate<int8_t>> creator;       if (creator.handle(attribute)) return creator.create(stash); }
             { SingleValueExecutorCreator<IntegerAttributeTemplate<int32_t>> creator;      if (creator.handle(attribute)) return creator.create(stash); }
             { SingleValueExecutorCreator<IntegerAttributeTemplate<int64_t>> creator;      if (creator.handle(attribute)) return creator.create(stash); }
+            {
+                auto boolAttribute = dynamic_cast<const SingleBoolAttribute *>(attribute);
+                if (boolAttribute) {
+                    return stash.create<BoolAttributeExecutor>(*boolAttribute);
```
havardpe: This should be possible if and only if we also chose to drop additional outputs.
It feels a bit fragile to check meta-data in one end and actual attribute implementation class in the other end.

havardpe (commented):
We need to separate executors with different number of outputs better. Right now the fallback for not being able to find the flag attribute is to create an executor producing 4 zero values (which would overwrite outside our 1 allocated output)

searchlib/src/vespa/searchlib/features/attributefeature.cpp
baldersheim: Done

searchlib/src/vespa/searchlib/features/attributefeature.cpp
baldersheim: Fixed and added some more sanity checks.

baldersheim:
@havardpe Now you can have a look again.

havardpe (commented):
still need fever zero fallback values for bool attributes

searchlib/src/vespa/searchlib/features/attributefeature.cpp
```diff
@@ -337,15 +351,17 @@ struct MultiValueExecutorCreator {
 };
 
 fef::FeatureExecutor &
-createAttributeExecutor(const IAttributeVector *attribute, const vespalib::string &attrName, const vespalib::string &extraParam, vespalib::Stash &stash)
+createAttributeExecutor(uint32_t numOutputs, const IAttributeVector *attribute, const vespalib::string &attrName, const vespalib::string &extraParam, vespalib::Stash &stash)
 {
     if (attribute == nullptr) {
         LOG(warning, "The attribute vector '%s' was not found in the attribute manager, returning default values.",
                 attrName.c_str());
         std::vector<feature_t> values(4, 0.0f);
```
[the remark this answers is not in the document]
baldersheim: Correct, done

baldersheim:
Superceeded by rebased variant #12464

PR #12335, opened 2020-02-25
Keep Athenz user domains lookup for 1min, per identity

jonmv:
@mpolden please review. @bjorncs @tokle FYI. 

There is already some delay on Athenz' end, so adding another minute on our side shouldn't matter much? It would, however, make access control _much_ faster, for all requests using Athenz for authorisation, and also reduce the number of calls we do against Athenz by a great deal.


jonmv:
Ah, I see there are other lookups we do, as well. Let me add more of this.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/filter/AthenzRoleFilter.java
```diff
@@ -46,11 +50,13 @@
 
     private final AthenzFacade athenz;
     private final TenantController tenants;
+    private final ExecutorService executor;
```
mpolden: Should probably be shut down on deconstruction? (Not 100% on whether this applies to filters)

jonmv:
I don't think it matters much. The controller JDisc isn't redeployed and reconfigured, like regular applications. Would have to make its superclass extend `AbstractComponent` to achieve. @bjorncs, any thoughts?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/filter/AthenzRoleFilter.java
bjorncs: That is correct. The filter should inherit `AbstractComponent` and do executor shutdown in `deconstruct()`.

mpolden (approved):

PR #12338, opened 2020-02-25
Trunkcate expressions in messages

bratseth:


vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -174,4 +174,16 @@ public static String stripInvalidCharacters(String string) {
         return stripped != null ? stripped.toString() : string;
     }
 
+    /**
+     * Returns a string which is never larger than the given number of characters.
+     * If the string is longer than the given length it will be truncated.
+     * If length is 4 or less the string will be truncated to length.
+     * If length is longer than 4, * it will be truncated at length-4 with " ..." added at the end.
```
lesters: "longer than 4, * it will"

lesters (approved):
LGTM.

PR #12340, opened 2020-02-25
Factor out generic thread pool to vespajlib.

baldersheim:
@havardpe PR
Refactor and fix shutdown bug.


havardpe:
looks good on this end, but missing the actual code...

PR #12345, opened 2020-02-26
Increase max number of connections per route for ZMS and ZTS client

bjorncs:
This should improve performance in situations where a client instance is shared among several threads.

FYI @tokle


jonmv (approved):

jonmv:
Thanks :)

PR #12347, opened 2020-02-26
Most files and directories in rpms should be owned by root.

toregge:
@hakonhall : please review
@aressem, @hmusum  : FYI


toregge:
Other changes already broke centos build.

PR #12348, opened 2020-02-26
Arnej/shrink if needed

arnej27959:

@geirst please review

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
@@ -47,7 +47,7 @@ HnswIndex::make_default_link_store_config()
 uint32_t
 HnswIndex::max_links_for_level(uint32_t level) const
 {
-    return (level == 0) ? _cfg.max_links_at_level_0() : _cfg.max_links_at_hierarchic_levels();
+    return (level == 0) ? _cfg.max_links_at_level_0() : _cfg.max_links_on_inserts();
 }
 
 uint32_t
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
```diff
@@ -287,8 +325,8 @@ HnswIndex::add_document(uint32_t docid)
     while (search_level >= 0) {
         // TODO: Rename to search_level?
         search_layer(input, _cfg.neighbors_to_explore_at_construction(), best_neighbors, search_level);
-        auto neighbors = select_neighbors(best_neighbors.peek(), max_links_for_level(search_level));
-        connect_new_node(docid, neighbors, search_level);
+        auto neighbors = select_neighbors(best_neighbors.peek(), _cfg.max_links_at_hierarchic_levels());
```
geirst: Shouldn't we still differentiate between max links for level 0 and hierarchical levels?

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
```diff
@@ -287,8 +325,8 @@ HnswIndex::add_document(uint32_t docid)
...
+        connect_new_node(docid, neighbors.used, search_level);
```
geirst: I guess we can remove the TODO on the next line?

geirst (approved):
👍  Please have a look at the question and nitpick

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
arnej27959: yes and no; we're supposed to select (up to ) M neighbors on all layers when inserting, but the "Mmax" used when shrinking is differentiatied (Mmax = M while Mmax0 = 2 * M)

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
arnej27959: right!

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
geirst: Ok, I see

PR #12361, opened 2020-02-27
Only get list of nodes when actually needed

hmusum:
Use a supplier and get list of nodes only when needed (when wanting to retire
a node). This should improve performance for all other patch requests.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodePatcher.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodePatcher.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static com.yahoo.config.provision.NodeResources.DiskSpeed.fast;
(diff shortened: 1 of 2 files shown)


hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodePatcher.java
```diff
@@ -133,9 +130,9 @@ private Node applyField(Node node, String name, Inspector value) {
             case "parentHostname" :
                 return node.withParentHostname(asString(value));
             case "ipAddresses" :
-                return IP.Config.verify(node.with(node.ipConfig().with(asStringSet(value))), nodes);
+                return IP.Config.verify(node.with(node.ipConfig().with(asStringSet(value))), nodes.get());
             case "additionalIpAddresses" :
-                return IP.Config.verify(node.with(node.ipConfig().with(IP.Pool.of(asStringSet(value)))), nodes);
+                return IP.Config.verify(node.with(node.ipConfig().with(IP.Pool.of(asStringSet(value)))), nodes.get());
```
hakonhall: If more than one of WANT_TO_RETIRE, ipAddresses, additionalIpAddresses are specified in the patch request, then this PR will invoke `nodeRepository.list(lock)` that many times. This happens e.g. on provisioning, when the last two are specified. At the risk of making this much more convoluted, consider listing at most once.

hakonhall: The overall impact of this PR should be a great win, so merging now...

PR #12366, opened 2020-02-27
Determine endpoints from routing policies for all routing methods

mpolden:
* Routing policies are now always fetched and stored on deploy. All zones
  already provide useful data in `/loadbalancers/v1/` independent of routing
  method used.
* All consumers now retrieve endpoints through `RoutingController`.
* There's a fallback that retrieves endpoints from `RoutingGenerator` which must
  exist until all applications have deployed at least once.
* Tests no longer use `RoutingGenerator` (except in the test for the previous bullet point)

FYI @tokle

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/RoutingController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/RoutingController.java
@@ -90,20 +90,23 @@ public class RoutingController {
 
     /** Returns global-scoped endpoints for given instance */
     public EndpointList endpointsOf(ApplicationId instance) {
+        return endpointsOf(controller.applications().requireInstance(instance));
+    }
+
+    public EndpointList endpointsOf(Instance instance) {
         var endpoints = new LinkedHashSet<Endpoint>();
         // Add global endpoints provided by rotations
-        for (var rotation : controller.applications().requireInstance(instance).rotations()) {
-            EndpointList.global(RoutingId.of(instance, rotation.endpointId()),
+        for (var rotation : instance.rotations()) {
+            EndpointList.global(RoutingId.of(instance.id(), rotation.endpointId()),
                                 controller.system(), systemRoutingMethods())
                         .requiresRotation()
-                        .primary()
-                        .ifPresent(endpoints::add);
+                        .forEach(endpoints::add);
         }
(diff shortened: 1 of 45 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/RoutingController.java
```diff
@@ -71,53 +68,57 @@ public RotationRepository rotations() {
...
+            }
         }
-        return legacyZoneEndpointsOf(deployment);
+        return EndpointList.copyOf(endpoints);
     }
 
-    /** Returns all non-global endpoint URLs for given deployments, grouped by their cluster ID and zone */
-    public Map<ZoneId, Map<ClusterSpec.Id, URI>> zoneEndpointsOf(Collection<DeploymentId> deployments) {
-        var endpoints = new TreeMap<ZoneId, Map<ClusterSpec.Id, URI>>(Comparator.comparing(ZoneId::value));
+    /** Returns all non-global endpoints for given deployments, grouped by their cluster ID and zone */
```
jonmv:
```suggestion
    /** Returns all non-global endpoints and corresponding cluster IDs for given deployments, grouped by their zone */
```

mpolden: Thanks!

jonmv (approved):
Looks great!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
```diff
@@ -169,6 +198,30 @@ private static String dnsSuffix(SystemName system, boolean legacy) {
...
+                     .filter(Predicate.not(String::isEmpty))
+                     .map(Endpoint::sanitizeUpstream)
+                     .collect(Collectors.joining("."));
+    }
+
+    /** Remove any invalid characters from a upstream part */
+    private static String sanitizeUpstream(String part) {
+        return truncate(part.toLowerCase()
+                            .replace('_', '-')
+                            .replaceAll("[^a-z0-9-]*", ""));
```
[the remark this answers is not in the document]
mpolden: This must match the implementation in the routing layer, unfortunately.

PR #12373, opened 2020-02-27
Revert "Only use json access log for config server"

aressem:
Reverts vespa-engine/vespa#12350

Seems to have broken:
https://factory-web.vespa.corp.yahoo.com/versions/1/builds/53870/products/1/tests/275607/testruns/38797158


hmusum (approved):

hmusum:
Thanks!

PR #12378, opened 2020-02-28
Prefer endpoints from routing generator

mpolden:
Integration test run revealed some missing pieces. Keep preferring endpoints
from routing generator until those are resolved.

@tokle or @jonmv

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
@@ -115,12 +115,13 @@ public class Endpoint {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         Endpoint endpoint = (Endpoint) o;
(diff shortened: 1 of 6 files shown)


tokle (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
```diff
@@ -115,12 +115,13 @@ public boolean equals(Object o) {
...
         if (o == null || getClass() != o.getClass()) return false;
         Endpoint endpoint = (Endpoint) o;
-        return url.equals(endpoint.url);
+        return url.equals(endpoint.url) &&
+               routingMethod == endpoint.routingMethod;
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(url);
```
jonmv: Will this be fixed? Seems a bit odd?

mpolden: Yes.

PR #12380, opened 2020-02-28
Export distributor GC documents_removed metric

vekterli:
@geirst please review

Exports both absolute document count within snapshot and rate of
change, as they can both be useful pieces of information when looking
at how many documents have been garbage collected.

This relates to issue #12139


geirst (approved):
👍

PR #12381, opened 2020-02-28
Balder/use single threaded executor

baldersheim:
@havardpe and @vekterli PR


staging_vespalib/src/vespa/vespalib/util/singleexecutor.h
```diff
@@ -0,0 +1,55 @@
...
+    void run() override;
+    void drain_tasks();
+    void run_tasks_till(uint64_t available);
+    void wait_for_room(MonitorGuard & guard);
+    uint64_t index(uint64_t counter) const {
+        return counter & (_taskLimit.load(std::memory_order_relaxed) - 1);
+    }
+
+    uint64_t numTasks() const {
+        return _wp.load(std::memory_order_relaxed) - _rp.load(std::memory_order_relaxed);
```
havardpe: It seems the write thread might observe tasks as run before they are actually run

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -0,0 +1,131 @@
...
+        producerGuard.wait(10ms);
+        _producerNeedWakeup.store(false, std::memory_order_relaxed);
+    }
+}
+
+ThreadExecutor::Stats
+SingleExecutor::getStats() {
+    uint64_t accepted = _wp.load(std::memory_order_relaxed);
+    Stats stats(_maxPending, (accepted - _lastAccepted), 0);
+    _lastAccepted = accepted;
```
havardpe: not thread safe. minor consequences

havardpe (commented):
relaxed memory order is ok for things you update yourself, but becomes dangerous when looking at things updated in other threads (prefer release/acquire in that case, unless the exactness of the result is not important)

staging_vespalib/src/vespa/vespalib/util/singleexecutor.h
baldersheim: Agreed, fixed.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Correct, but usage is always thread safe. Will reconsider after field testing and some maturing.

vekterli (commented):
:+1:

PR #12383, opened 2020-02-28
Fix order of upstream ID elements

mpolden:


hmusum (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
```diff
@@ -202,8 +202,8 @@ private static String dnsSuffix(SystemName system, boolean legacy) {
     private static String upstreamIdOf(String name, ApplicationId application, ZoneId zone) {
         return Stream.of(namePart(name, ""),
                          instancePart(application, ""),
-                         application.tenant().value(),
                          application.application().value(),
+                         application.tenant().value(),
```
jonmv: Sorry I didn't catch this >_<

PR #12390, opened 2020-03-02
Extract low-level ZK HostInfo ops to HostInfosServiceImpl

hakonhall:


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/status/HostInfosServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static String applicationPath(ApplicationId application) {
+        return "/vespa/host-status/" + application.serializedForm();
+    }
+
+    private static String hostsPath(ApplicationId applicationInstanceReference) {
+        return applicationPath(applicationInstanceReference) + "/hosts";
```
freva:
```suggestion
    private static String hostsPath(ApplicationId application) {
        return applicationPath(application) + "/hosts";
```

freva (approved):

PR #12393, opened 2020-03-02
Ensure that we get the offending query when a stackoverflow occur.

baldersheim:
@bratseth PR


baldersheim:
@bratseth Or should we just catch throwable ? I prefer being restrictive on catch to not hide stuff we do not know how to handle.

bratseth:
> I prefer being restrictive on catch to not hide stuff we do not know how to handle.

Agree

PR #12398, opened 2020-03-02
Always use container cluster ID as cluster ID for load balancers 

mpolden:
@hmusum and @jonmv

FYI @bratseth


config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
```diff
@@ -54,7 +59,8 @@ protected String toStringValue() {
                (cluster.group().isPresent() ? "/" + cluster.group().get().index() : "") +
                "/" + index +
                ( cluster.isExclusive() ? "/exclusive" : "") +
-               ( retired ? "/retired" : "");
+               ( retired ? "/retired" : "") +
+               ( cluster.combinedId().isPresent() ? "/" + cluster.combinedId().get().value() : "");
```
hmusum: During upgrade this will write a value with combined id that other instances that are not upgraded yet will not be able to read. Shouldn't we handle strings with combined id first and in a later version start writing it?

mpolden: Of course, thanks!

mpolden: Actually, the existing code appears to handle this. It ignores trailing components of the serialized string that it does not explicitly handle.

hmusum: Yes, it might not be an issue here. As long as you don't need the combined id I guess that will work, but if you later rely upon this (e.g. in a later PR that goes in before this is released) something might break.

hmusum (approved):

jonmv (approved):
LGTM.

jonmv:
Ping merge monster.

PR #12399, opened 2020-03-02
Bratseth/phrase segmenting feature flag

bratseth:
Since the flag is default true this should have no effect.


arnej27959 (approved):
looks good to me

PR #12404, opened 2020-03-02
Remove InstanceLookupService

hakonhall:
The lower-level methods on ServiceMonitor has removed the need for InstanceLookupService.


freva (approved):
:clap:

PR #12406, opened 2020-03-03
Fail early on submit with missing security/clients.pem

oyving:
These submissions would fail in the configuration server regardless, so we might as well fail them in the controller. This way users will get an error message immediately instead of having to see it in the deployment job view in the Console.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1956,6 +1956,12 @@ private HttpResponse submit(String tenant, String application, HttpRequest reque
 
         ApplicationPackage applicationPackage = new ApplicationPackage(dataParts.get(EnvironmentResource.APPLICATION_ZIP), true);
 
+        // if we are in public and 'security/clients.pem' is not present (i.e. we don't have any trusted
+        // certificates), we should fail early instead of waiting for the configserver to fail.
+        if (controller.system().isPublic() && applicationPackage.trustedCertificates().isEmpty()) {
```
mpolden: This belongs in `ApplicationPackageValidator`.

oyving: Should be fixed now. Take another look. I have no idea if the test works since the build got broken by a later `git pull` :(

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackageValidator.java
```diff
@@ -46,6 +46,13 @@ public void validate(Application application, ApplicationPackage applicationPack
         validateSteps(applicationPackage.deploymentSpec());
         validateEndpointRegions(applicationPackage.deploymentSpec());
         validateEndpointChange(application, applicationPackage, instant);
+        validateSecurityClientsPem(applicationPackage);
+    }
+
+    /** Verify that we have the security/clients.pem file for public systems */
+    public void validateSecurityClientsPem(ApplicationPackage applicationPackage) {
```
mpolden: Should be `private`.

mpolden (approved):

mpolden (commented):
Looks good, but `RoutingPoliciesTest.non_production_deployment_is_not_registered_in_global_endpoint` fails. Probably need to add a certificate to the package used in this test.

PR #12409, opened 2020-03-03
Revert "Bratseth/phrase segmenting feature flag" MERGEOK

bratseth:
Reverts vespa-engine/vespa#12399


aressem (approved):

hmusum:
This broke the config model API, since there are existing config models (in CD) that call `ModelContext.Properties.usePhraseSegmenting()`. I'll make a PR to add that back (with a default implementation returning `true`)

PR #12410, opened 2020-03-03
Write rotation regions to ZK

mpolden:
The existing approach is incorrect as the current deployment spec is not
guaranteed to be in sync with the assigned rotation.

E.g. if an empty deployment spec is submitted, an instance may still have a
rotation assigned, but the regions of that rotation disappear because they are
deduced from deployment spec which is now empty.

This change writes the rotation regions explicitly. Will start reading the field
in a future PR, after this is released.

This is necessary because to be able to generate global endpoints in systems
where zones can support a mix of `shared`, `exclusive` and `sharedLayer4` the
system must always know about the region targets of a rotation, even when
removing an application and/or instance.

FYI @tokle


oyving (approved):

mpolden:
merge defrost

PR #12411, opened 2020-03-03
Add back usePhraseSegmenting() to revert breaking API change MERGEOK

hmusum:


andreer (approved):

bratseth:
👍 but we shouldn't really include config models from earlier non-released builds ...

hmusum:
In CD we have to, since there are applications running on those versions

bratseth:
Bu then we can't do clean reverts :disappointed:

aressem:
Restoring CD to the most recent released version in prod before running would solve this (amongst several other things).

bratseth:
Yep. It's just a +1 on that

PR #12414, opened 2020-03-03
Require vespa-telegraf on CentOS 

gjoranv:
FYI: @olaaun


dist/vespa.spec
```diff
@@ -120,6 +120,7 @@ BuildRequires: flex >= 2.5.0
 BuildRequires: bison >= 3.0.0
 %if 0%{?centos}
 Requires: epel-release
+Requires: vespa-telegraf
```
aressem: There is no vespa-telegraf for centos 8 so this should be put into the el7 config further down in the file.

aressem (changes requested):

gjoranv:
@aressem PTAL

aressem (approved):

PR #12416, opened 2020-03-03
Ensure we retrieve the actual vector (used by nearest neighbor index)...

geirst:
… for a document being added instead of the origo vector.

Before the fix, getTensor() would return the default origo tensor as docid would not be less than getCommittedDocIdLimit().
This happened as the docid limit is updated at the end of handling a document being added, which is too late.
The fix is to ignore the docid limit in is this particular case.

Also update the unit test for tensor attributes to do commit() (which updates the docid limit) later.
Before the fix, this triggered the bug in the test "setTensor() updates nearest neighbor index".

@arnej27959 please review
@toregge @jobergum FYI


arnej27959 (approved):
👍

PR #12418, opened 2020-03-03
Bratseth/autoscaling debug logging

bratseth:
I can't find any autoscaling suggestions in the log so I need some debug messages.


bratseth:
Done from the wrong branch, I'll resubmit

PR #12425, opened 2020-03-04
Cleanup

mpolden:
@jonmv


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -572,7 +572,8 @@ public void locked(RunId id, UnaryOperator<Run> modifications) {
     public void locked(ApplicationId id, JobType type, Step step, Consumer<LockedStep> action) throws TimeoutException {
         try (Lock lock = curator.lock(id, type, step)) {
             for (Step prerequisite : step.prerequisites()) // Check that no prerequisite is still running.
-                try (Lock __ = curator.lock(id, type, prerequisite)) { ; }
```
jonmv: Can't say I agree with IntelliJ on this one, but NVM :)

jonmv (approved):
:)

PR #12426, opened 2020-03-04
Use Google RE2 as underlying regex engine

vekterli:
@geirst please review
@toregge please take a look at my CMake changes to see if they make sense (and the rest too if you want!)

This introduces guaranteed upper bounds for memory usage and
CPU time during regex evaluation. Most importantly, it removes
the danger of catastrophic backtracking that is currrently
present in GCC's `std::regex` implementation.

With this commit, RE2 will be used instead of `std::regex` for:
* Document selection regex/glob operators
* Attribute regex search
* Evaluation of mTLS authorization rules

This relates to issue #12068

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,11 @@
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 include(functions.cmake)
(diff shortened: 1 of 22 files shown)


toregge (commented):
LGTM

vekterli:
Merging needs to wait until some docker images have been rebuilt to include the dependencies.

geirst (approved):
👍

PR #12435, opened 2020-03-04
Bratseth/handle remote disk

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -142,6 +143,8 @@ private boolean similar(double r1, double r2, double threshold) {
             Optional<Flavor> bestFlavor = Optional.empty();
             for (Flavor flavor : nodeRepository.getAvailableFlavors().getFlavors()) {
                 if ( ! flavor.resources().satisfies(resources.nodeResources())) continue;
```
freva: `flavor.resources()` returns the real resources, need to run this through `hostResourcesCalculator` so the `bestFlavor` ends up as an advertised flavor.

freva (commented):
You should also do 
```
NodeResources nodeResources = nodeResourceLimits.enlargeToLegal(resources.nodeResources(), cluster.type());
```
for non-host-shared clouds.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -126,32 +126,31 @@ private boolean similar(double r1, double r2, double threshold) {
...
-                    bestFlavor = Optional.of(flavor);
-                    bestCost = costOf(flavor);
-                }
+
+                if (flavor.resources().storageType() == NodeResources.StorageType.remote)
+                    flavor = flavor.with(FlavorOverrides.ofDisk(resources.nodeResources().diskGb()));
+                var candidate = new AllocatableClusterResources(resources.with(flavor.resources()),
+                                                                hostResourcesCalculator.availableCapacityOf(flavor.name(), flavor.resources()));
+
+                if (best.isEmpty() || best.get().cost() > costOf(flavor.resources()))
```
freva:
```suggestion
                if (best.isEmpty() || best.get().cost() > candidate.cost())
```
Previously this would compare cost of advertised resources for the best match against cost of real resources for the candidate.

freva (approved):
LGTM, one minor bug.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
bratseth: > You should also do nodeResourceLimits.enlargeToLegal [...]

Only if we add host flavors that cannot be allocated, but ok.

bratseth:
I think this is done now, PTAL

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
freva: We already have (a few `t3-` flavors)

freva (approved):

PR #12452, opened 2020-03-05
Remove unused roles

mpolden:
@jonmv

FYI @tokle @smorgrav


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/user/Roles.java
```diff
@@ -57,10 +54,6 @@ public static Role toRole(TenantName tenant, String roleName) {
     /** Returns the {@link Role} the given tenant and role names correspond to. */
     public static Role toRole(TenantName tenant, ApplicationName application, String roleName) {
         switch (roleName) {
-            case "applicationAdmin": return Role.applicationAdmin(tenant, application);
-            case "applicationOperator": return Role.applicationOperator(tenant, application);
-            case "applicationDeveloper": return Role.applicationDeveloper(tenant, application);
-            case "applicationReader": return Role.applicationReader(tenant, application);
```
jonmv: Will the potential exception be caught by the caller?

mpolden: Yes, see internal PR.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Role.java
```diff
@@ -43,31 +43,16 @@ public static TenantRole athenzTenantAdmin(TenantName tenant) {
         return new TenantRole(RoleDefinition.athenzTenantAdmin, tenant);
     }
 
-    /** Returns a {@link RoleDefinition#tenantPipeline} for the current system and given tenant and application. */
-    public static ApplicationRole tenantPipeline(TenantName tenant, ApplicationName application) {
-        return new ApplicationRole(RoleDefinition.tenantPipeline, tenant, application);
+    /** Returns a {@link RoleDefinition#integrationTest} for the current system and given tenant and application. */
+    public static ApplicationRole integrationTest(TenantName tenant, ApplicationName application) {
+        return new ApplicationRole(RoleDefinition.integrationTest, tenant, application);
```
jonmv: On second thought, the integration test is a `hostedOperator`, so this isn't really needed.

jonmv (approved):
Great!

jonmv:
The whole `InstanceRole` is also unused, and can be deleted.

mpolden:
PTAL

jonmv (approved):

PR #12454, opened 2020-03-05
404 on listing applications for tenant which does not exist

jonmv:
@freva


freva (approved):
:+1: But there's also the same bug on `/tenant/:t/application/:a/instance`

jonmv:
Same code path.

PR #12457, opened 2020-03-05
Add projectId to deployment API

freva:
We can change & rename this field to something with URL later on so public system is supported, but adding it here for now to be able to display simple link back to SD (VESPA-17713).


jonmv (approved):

jonmv:
Thanks :)

PR #12461, opened 2020-03-05
Close lock on exception in MaintenanceDeployment

hakonhall:


mpolden (commented):
Instead of introducing a variable to avoid always closing, wouldn't the following accomplish the same thing?
```
try {
    lock = tryLock(application, nodeRepository);
} catch (Exception e) {
    close();
    throw e;
}
```

hakonhall:
You mean across the tryDeployment()...

Yes, I'm fine with either, although there is a slight benefit to the current variant that ends up suppressing any exception from the close() instead of hiding the original and usually more interesting.

bratseth:
👍

PR #12467, opened 2020-03-05
lz4 seems to be unused here.

baldersheim:
@hmusum and @freva PR
I am preparing to upgrade lz4 from 1.3.0 from 2014 to current version.
But trying to clean up first.

### DIFF
--- a/configserver/pom.xml
+++ b/configserver/pom.xml
@@ -215,11 +215,6 @@
       <artifactId>jersey-proxy-client</artifactId>
     </dependency>
     <dependency>
-      <groupId>net.jpountz.lz4</groupId>
-      <artifactId>lz4</artifactId>
-      <scope>compile</scope>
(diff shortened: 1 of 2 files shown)


freva (approved):

hmusum:
I seem to remember there was some issues when trying to do a similar upgrade some time ago. See https://github.com/vespa-engine/vespa/pull/9212. I can't remember the details and unfortunately there were not comments in PRs about what went wrong.

freva:
I'm don't know about configserver, but I thought I remembered similar issue with docker-api. However, I patched a random node in CD without it and it worked :man_shrugging:

baldersheim:
I suspect there are some ExportPackages somewhere.
I will try to remove seemingly unused stuff first, and hide the implementation purely in vespajlib.

PR #12475, opened 2020-03-06
Filter out tester apps

bratseth:


freva (approved):
This method doesn't really belong in `Maintainer`...

bratseth:
Not a problem right now imho. If we have a number of these or Maintainer gets too big we should move it.

PR #12480, opened 2020-03-06
Upgrade from net.jpountz.lz4 1.3.0 to org.lz4 1.7.1

baldersheim:
@hmusum @gjoranv PR
This should replace lz4.


hmusum (changes requested):
The package-info.java files should be in src/main/java/org/lz4, not in src/main/java/net/jpountz/lz4,

hmusum (approved):
Package name is unchanged, all OK

gjoranv (approved):
LGTM. Remember to update the hosted-dependencies-enforcer in the other repo.

PR #12484, opened 2020-03-06
Add JIT warmup of lz4.

baldersheim:
@bratseth PR

### DIFF
--- a/configdefinitions/src/vespa/dispatch.def
+++ b/configdefinitions/src/vespa/dispatch.def
@@ -48,6 +48,9 @@ numJrtTransportThreads int default=8
 # Number of JRT connections per backend node
 numJrtConnectionsPerNode int default=8
 
+# Number of seconds to spend warming up code to prevent JIT cold start issues.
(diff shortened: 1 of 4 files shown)


bratseth (approved):
Fine for testing it but I think we should move it into compressor so that we can say compressor.warmup(time).

configdefinitions/src/vespa/dispatch.def
```diff
@@ -48,6 +48,9 @@ numJrtTransportThreads int default=8
 # Number of JRT connections per backend node
 numJrtConnectionsPerNode int default=8
 
+# Number of seconds to spend warming up code to prevent JIT cold start issues.
+warmuptime double default=0.1
```
bratseth: Any particular reason for the double level of defaults? Just set 5 here?

baldersheim: Intention to avoid long unit test runtime.

baldersheim:
Moved into compressor.

PR #12486, opened 2020-03-06
Bratseth/nn query type resolving

bratseth:


config-model/src/test/java/com/yahoo/searchdefinition/derived/NearestNeighborTestCase.java
```diff
@@ -0,0 +1,37 @@
+// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
```
geirst: Please use updated copyright.

geirst (approved):
👍

config-model/src/test/java/com/yahoo/searchdefinition/derived/NearestNeighborTestCase.java
bratseth: We don't need to update copyright when we change a file.

geirst: Yes, but this is a new file.

bratseth: Ah, you're right

PR #12493, opened 2020-03-06
Use an interface that is more flexible.

baldersheim:
@hmusum PR


hmusum (approved):

hmusum (changes requested):
Test fails:
`[ERROR] /source/config/src/test/java/com/yahoo/vespa/config/LZ4PayloadCompressorTest.java:[27,43] incompatible types: byte[] cannot be converted to int`

PR #12499, opened 2020-03-08
Also remove hostnames from hostnamesById when removing hosts via application activate

hakonhall:
I have seen a few "Host not found" messages in CD the last 2 days, and found
this bug in how the duper model indices are maintained on application
activation.


hakonhall:
This should make tomorrow's release.

bratseth (approved):

hakonhall:
thanks

hakonhall:
Factory failed with a NPE in application's container_and_referenced_content test, but I don't see how that could be related...

hakonhall:
And I haven't been able to reproduce it, whether with mvn or IntelliJ.

hakonhall:
application passed on next build

PR #12503, opened 2020-03-08
When input is already utf8, you should take advantage of it to avoid …

baldersheim:
…string split and string decoding.

@hmusum or @vekterli PR


hmusum (approved):

vekterli (commented):
:+1:

PR #12507, opened 2020-03-09
add feature flag for net.ipv4.tcp_abort_on_overflow

andreer:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -169,6 +169,12 @@ public class Flags {
             "Takes effect on next tick or on host-admin restart (may vary where used).",
             HOSTNAME);
 
+    public static final UnboundBooleanFlag TCP_ABORT_ON_OVERFLOW = defineFeatureFlag(
+            "tcp-abort-on-overflow", false,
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -169,6 +169,12 @@
             "Takes effect on next tick or on host-admin restart (may vary where used).",
             HOSTNAME);
 
+    public static final UnboundBooleanFlag TCP_ABORT_ON_OVERFLOW = defineFeatureFlag(
+            "tcp-abort-on-overflow", false,
+            "Whether to set /proc/sys/net/ipv4/tcp_abort_on_overflow to 0 (false) or 1 (true)",
+            "Takes effect on next tick or on host-admin restart (may vary where used).",
```
hakonhall: Typically this would be host-admin and a feature flag task, meaning it takes effect when the flag is flipped.

hakonhall: ah, the SysctlTask does not use that. And I seem to remember vaguely why that was: Something changes the sysctl settings without us really finding the cause of it, possibly cm3, so we set these at every tick.

In any case, the flag is read on every tick.

hakonhall (approved):

PR #12510, opened 2020-03-09
Jonmv/longer convergence timeout and better debug info

jonmv:
@mpolden

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
@@ -121,7 +121,8 @@ public class InternalStepRunner implements StepRunner {
     static final Duration endpointTimeout = Duration.ofMinutes(15);
     static final Duration endpointCertificateTimeout = Duration.ofMinutes(15);
     static final Duration testerTimeout = Duration.ofMinutes(30);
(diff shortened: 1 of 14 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -380,9 +381,20 @@ else if (timedOut(id, deployment.get(), endpointTimeout)) {
...
             logger.log(nodeList.asList().stream()
                                .flatMap(node -> nodeDetails(node, true))
                                .collect(toList()));
+            logger.log("######## Details for nodes with pending changes ########");
+            logger.log(nodeList.not().in(nodeList.not().needsNewConfig()
+                                                 .not().needsPlatformUpgrade()
+                                                 .not().needsReboot()
+                                                 .not().needsRestart()
+                                                 .not().needsFirmwareUpgrade()
+                                                 .not().needsOsUpgrade())
```
mpolden: 😆

mpolden (approved):
Great!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
jonmv: Nothing quite like set arithmetics :p

PR #12512, opened 2020-03-09
Require 'file' and make 'profile' optional for shared-credentials

gjoranv:
- The default profile is 'default'


olaaun:
`TelegrafTest` fails

olaaun (approved):

PR #12515, opened 2020-03-09
void creating strings when not needed.

baldersheim:
Do not depend on toString(), use toUtf8 instead.

@vekterli and @bratseth PR

### DIFF
--- a/component/src/main/java/com/yahoo/component/Version.java
+++ b/component/src/main/java/com/yahoo/component/Version.java
@@ -279,7 +279,7 @@ public final class Version implements Comparable<Version> {
     public String toString() { return toStringValue(); }
 
     public Utf8Array toUtf8() {
-        return utf8 != null ? utf8 : new Utf8String(toString());
+        return utf8;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


baldersheim:
@hmusum PR too

bratseth (approved):

component/src/main/java/com/yahoo/component/Version.java
```diff
@@ -270,10 +276,14 @@ public String toFullString() {
      * The string representation of a Version specified here is a part of the API and will never change.
      */
     @Override
-    public String toString() { return stringValue; }
+    public String toString() { return toStringValue(); }
+
+    public Utf8Array toUtf8() {
+        return utf8 != null ? utf8 : new Utf8String(toString());
```
vekterli: Can `utf8` ever be `null` here?

vekterli (approved):
:+1: one question about possible simplification

hmusum (approved):

component/src/main/java/com/yahoo/component/Version.java
baldersheim: Good point, not anymore.

PR #12519, opened 2020-03-09
Bjorncs/remove pre binding of server socket

bjorncs:
This should fix the problem with the configserver server socket being opened too early.

FYI @andreer @gjoranv @bratseth @yngveaasheim


bratseth (approved):
Heh ... 👍

PR #12521, opened 2020-03-09
Bratseth/deprovisioned node state

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -608,8 +608,11 @@ public Node markNodeAvailableForNewAllocation(String hostname, Agent agent, Stri
                 List<Node> children = list().childrenOf(node).asList();
                 children.forEach(child -> requireRemovable(child, true, force));
                 db.removeNodes(children);
-                move(node, State.deprovisioned, Agent.system, Optional.empty());
                 List<Node> removed = new ArrayList<>(children);
+                if (zone.cloud().value().equals("aws"))
```
mpolden: Looks like we have this in several places now. We should find a way to model it properly soon.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/NodeState.java
```diff
@@ -9,5 +9,5 @@
  * @author freva
  */
 public enum NodeState {
-    provisioned, ready, reserved, active, inactive, dirty, failed, parked
+    provisioned, ready, reserved, active, inactive, dirty, failed, parked, deprovisioned
```
mpolden: Should also be added to `com.yahoo.vespa.hosted.controller.api.integration.noderepository.NodeState`.

config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
```diff
@@ -45,7 +45,7 @@
         this.description = description;
     }
 
-    public boolean isHost() {
+    public boolean isDockerHost() {
```
mpolden: Why not?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
bratseth: Agree. What I want to do is create a "cloud config" which defines the properties of a cloud and then only test for those properties.

config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
bratseth: Breaks the config server/config model API - in theory, did not look at usages.
I still want to do it but in a separate PR.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/NodeState.java
bratseth: Done - thanks

mpolden (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Agent.java
```diff
@@ -17,5 +17,6 @@
     FailedExpirer,
     InactiveExpirer,
     ProvisionedExpirer,
-    ReservationExpirer
+    ReservationExpirer,
+    DynamicProvisioningMaintainer
```
freva: Add to https://github.com/vespa-engine/vespa/blob/ddd5141d112d44f881da17e83d26ac18ad35c3c1/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeHistory.java#L35

bratseth: Thanks 

https://github.com/vespa-engine/vespa/pull/12523

PR #12525, opened 2020-03-10
Validate that endpoint points to an existing container cluster

mpolden:
@hmusum

FYI @andreer


andreer (approved):

hmusum:
Could this lead to a deployment failing for an already existing application?

mpolden:
@hmusum That would mean the application has declared an endpoint that always returns 404. I can try verifying prod applications tomorrow.

mpolden:
Verified that no currently deployed applications violate this validation.

PR #12526, opened 2020-03-10
Revert "Keep temp files if force is true (for debugging)"

hmusum:
Reverts vespa-engine/vespa#12433

Seeing issues in production where /tmp fills up with deploy_ranksetup directories


hakonhall (approved):

hmusum:
@olaaun FYI

PR #12533, opened 2020-03-11
Validate before prepare to avoid depending on the properties API

bratseth:
The problem we saw earlier with the sift nn has not resurfaced. I don't understand why but there was a log of changes to the test as well at that time.

Regardless, I think it is better to implement this by accessing rank features, the translation to rank properties is an (unfortunate) implementation detail.


geirst (approved):
👍

PR #12534, opened 2020-03-11
Bjorncs/jdisc metric unhandled exception

bjorncs:


container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
```diff
@@ -97,9 +97,12 @@ protected ThreadedRequestHandler(Executor executor, Metric metric, boolean allow
         dimensions.put("port", Integer.toString(uri.getPort()));
         String handlerClassName = getClass().getName();
         dimensions.put("handler-name", handlerClassName);
+        dimensions.putAll(extraDimensions);
         return this.metric.createContext(dimensions);
     }
 
+    private Metric.Context contextFor(Request request) { return contextFor(request, Map.of()); }
```
gjoranv: What is the difference between Map.of() and Collections.emptyMap()?

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedHttpRequestHandler.java
```diff
@@ -31,6 +32,7 @@
 
     public static final String CONTENT_TYPE = "Content-Type";
     private static final String RENDERING_ERRORS = "rendering_errors";
+    private static final String UNHANDLED_EXCEPTION_METRIC = "jdisc.http.handler.unhandled_exception";
```
gjoranv: Shouldn't it be `..unhandled_exceptions` with the plural 's'? If so, also rename the constant.

gjoranv (commented):
Other counter metrics use a plural 's' so I suggest to do that, and rename everywhere, constant, test method etc.

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
bjorncs: `Map.of()` was introduced in Java 9. The returned `Map` instance from both have the same semantics (immutable etc). I prefer the new syntax as it is shorter.

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedHttpRequestHandler.java
bjorncs: Agree - plural sounds better. I will update the PR with a rename.

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
gjoranv: When you static import the `emptyMap()` function, I think that is more describing and almost as short. But this is just nit of course. :)

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedHttpRequestHandler.java
gjoranv: 👍

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
bjorncs: Agree - if static import is used `emptyMap()` is more readable 👍

bjorncs:
@gjoranv PTAL

gjoranv (approved):
Optional: I suggest to rename the test method and also the metric constants to reflect that the metric names are 'exceptions' and not 'exception'.

gjoranv (approved):

PR #12540, opened 2020-03-11
In order to drain Q faster on sync, and also detect that we are in sy… 

baldersheim:
…nc faster,

we wake the consumer unconditionally on sync, and also unconditionally wake the producer when consumer is idle.

@havardpe and @vekterli PR

### DIFF
--- a/staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
+++ b/staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
@@ -10,15 +10,14 @@ SingleExecutor::SingleExecutor(uint32_t taskLimit)
       _wantedTaskLimit(_taskLimit.load()),
       _rp(0),
       _tasks(std::make_unique<Task::UP[]>(_taskLimit)),
-      _consumerMutex(),
+      _mutex(),
       _consumerCondition(),
-      _producerMutex(),
       _producerCondition(),
       _thread(*this),
       _lastAccepted(0),
       _maxPending(0),
       _wakeupConsumerAt(0),
-      _producerNeedWakeup(false),
+      _producerNeedWakeupAt(0),
       _wp(0)
 {
     _thread.start();
@@ -35,7 +34,7 @@ SingleExecutor::getNumThreads() const {
 
 uint64_t
 SingleExecutor::addTask(Task::UP task) {
-    Lock guard(_producerMutex);
+    Lock guard(_mutex);
     wait_for_room(guard);
     uint64_t wp = _wp.load(std::memory_order_relaxed);
     _tasks[index(wp)] = std::move(task);
@@ -44,10 +43,10 @@ SingleExecutor::addTask(Task::UP task) {
 }
 
 void
-SingleExecutor::sleepProducer(Lock & lock, duration maxWaitTime) {
-    _producerNeedWakeup.store(true, std::memory_order_relaxed);
+SingleExecutor::sleepProducer(Lock & lock, duration maxWaitTime, uint64_t wakeupAt) {
+    _producerNeedWakeupAt.store(wakeupAt, std::memory_order_relaxed);
     _producerCondition.wait_for(lock, maxWaitTime);
-    _producerNeedWakeup.store(false, std::memory_order_relaxed);
+    _producerNeedWakeupAt.store(0, std::memory_order_relaxed);
 }
 
 Executor::Task::UP
@@ -64,13 +63,27 @@ SingleExecutor::setTaskLimit(uint32_t taskLimit) {
     _wantedTaskLimit = vespalib::roundUp2inN(taskLimit);
 }
 
+void
+SingleExecutor::drain(Lock & lock) {
+    uint64_t wp = _wp.load(std::memory_order_relaxed);
+    while (numTasks() > 0) {
+        _consumerCondition.notify_one();
+        sleepProducer(lock, 100us, wp);
+    }
+}
+
+void
+SingleExecutor::startSync() {
+    _consumerCondition.notify_one();
+}
+
 SingleExecutor &
 SingleExecutor::sync() {
+    Lock lock(_mutex);
     uint64_t wp = _wp.load(std::memory_order_relaxed);
     while (wp > _rp.load(std::memory_order_acquire)) {
         _consumerCondition.notify_one();
-        Lock lock(_producerMutex);
-        sleepProducer(lock, 100us);
+        sleepProducer(lock, 100us, wp);
     }
     return *this;
 }
@@ -81,8 +94,10 @@ SingleExecutor::run() {
         drain_tasks();
         _producerCondition.notify_all();
         _wakeupConsumerAt.store(_wp.load(std::memory_order_relaxed) + (_taskLimit.load(std::memory_order_relaxed) / 4), std::memory_order_relaxed);
-        Lock lock(_consumerMutex);
-        _consumerCondition.wait_for(lock, 10ms);
+        Lock lock(_mutex);
+        if (numTasks() <= 0) {
+            _consumerCondition.wait_for(lock, 10ms);
+        }
         _wakeupConsumerAt.store(0, std::memory_order_relaxed);
     }
 }
@@ -101,9 +116,7 @@ SingleExecutor::run_tasks_till(uint64_t available) {
     if (_maxPending.load(std::memory_order_relaxed) < left) {
         _maxPending.store(left, std::memory_order_relaxed);
     }
-    uint64_t wakeupLimit = _producerNeedWakeup.load(std::memory_order_relaxed)
-            ? (available - (left / 4))
-            : 0;
+    uint64_t wakeupLimit = _producerNeedWakeupAt.load(std::memory_order_relaxed);
     while (consumed  < available) {
         Task::UP task = std::move(_tasks[index(consumed)]);
         task->run();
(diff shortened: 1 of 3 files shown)


baldersheim:
@jobergum FYI

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -41,12 +41,38 @@ SingleExecutor::addTask(Task::UP task) {
     return wp;
 }
 
+void
+SingleExecutor::wakeupConsumer() {
+    MonitorGuard guard(_consumerMonitor);
```
havardpe: You can skip taking the lock here; consider using std::mutex and std::condition_variable instead.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -41,12 +41,38 @@ SingleExecutor::addTask(Task::UP task) {
...
+SingleExecutor::sleepConsumer() {
+    _wakeupConsumerAt.store(_wp.load(std::memory_order_relaxed) + (_taskLimit.load(std::memory_order_relaxed) >> 2), std::memory_order_relaxed);
+    MonitorGuard guard(_consumerMonitor);
+    guard.wait(10ms);
+    _wakeupConsumerAt.store(0, std::memory_order_relaxed);
+}
+
+void
+SingleExecutor::wakeupProducer() {
+    MonitorGuard guard(_producerMonitor);
```
havardpe: You can skip taking the lock here; consider using std::mutex and std::condition_variable instead.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -60,7 +86,9 @@ SingleExecutor &
 SingleExecutor::sync() {
     uint64_t wp = _wp.load(std::memory_order_relaxed);
     while (wp > _rp.load(std::memory_order_acquire)) {
-        std::this_thread::sleep_for(1ms);
+        wakeupConsumer();
```
havardpe: note that this signal may be lost since no state is being passed with the signal (the other thread might call wait after the signal has been discarded since no-one was waiting).

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -41,12 +41,38 @@ SingleExecutor::addTask(Task::UP task) {
...
+    _producerNeedWakeup.store(false, std::memory_order_relaxed);
+}
+
 Executor::Task::UP
 SingleExecutor::execute(Task::UP task) {
     uint64_t wp = addTask(std::move(task));
     if (wp == _wakeupConsumerAt.load(std::memory_order_relaxed)) {
-        MonitorGuard guard(_consumerMonitor);
-        guard.signal();
+        wakeupConsumer();
```
havardpe: note that this signal may be lost since no state is being passed with the signal (the other thread might call wait after the signal has been discarded since no-one was waiting).

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -69,10 +97,8 @@ void
 SingleExecutor::run() {
     while (!_thread.stopped()) {
         drain_tasks();
-        _wakeupConsumerAt.store(_wp.load(std::memory_order_relaxed) + (_taskLimit.load(std::memory_order_relaxed) >> 2), std::memory_order_relaxed);
-        MonitorGuard guard(_consumerMonitor);
-        guard.wait(10ms);
-        _wakeupConsumerAt.store(0, std::memory_order_relaxed);
+        wakeupProducer();
```
havardpe: note that this signal may be lost since no state is being passed with the signal (the other thread might call wait after the signal has been discarded since no-one was waiting).

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -98,8 +124,7 @@ SingleExecutor::run_tasks_till(uint64_t available) {
         task->run();
         _rp.store(++consumed, std::memory_order_release);
         if (wakeupLimit == consumed) {
-            MonitorGuard guard(_producerMonitor);
-            guard.broadcast();
+            wakeupProducer();
```
havardpe: note that this signal may be lost since no state is being passed with the signal (the other thread might call wait after the signal has been discarded since no-one was waiting).

havardpe (commented):
some speculative signaling might improve overall behavior, but you will still have the typical race condition where one thread indicates that it needs to be woken and the other thread tries to wake it before the first thread was able to call wait. This will result in the signal being lost.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Yes, correct. The purpose is to wake up one that is sleeping waiting for low watermark to be reached.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Yes, correct. The purpose is to wake up one that is sleeping waiting for low watermark to be reached.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Yes, correct. The purpose is to wake up one that is sleeping waiting for low watermark to be reached.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Yes, correct. The purpose is to wake up one that is sleeping waiting for low watermark to be reached.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: OK, will do, have not used them directly before.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
havardpe: The purpose is irrelevant if the signal is lost. The thread you are trying to wake up will keep sleeping. This might not happen too often, but it could be a bit surprising when it does.

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
vekterli: [cppreference](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one) doc on `std::condition_variable` is pretty handy on this subject:

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock. However, some implementations (in particular many implementations of pthreads) recognize this situation and avoid this "hurry up and wait" scenario by transferring the waiting thread from the condition variable's queue directly to the queue of the mutex within the notify call, without waking it up.

baldersheim: Done

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Done

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: Now it will be attempted woken up again after every 100us. I think that should be good enough.

baldersheim:
@havardpe Now I think it should be good.

havardpe (commented):
you only need one mutex, unless you expect contention. Also, you could consider using that lock to protect the 'I want to be woken up' state in a way that implements the normal use of a condition variable. Waiting 100ns before waking the other thread in a loop will make it more likely to miss a wakeup yourself, but that might not be an issue since you only sleep for 100ns anyways...

baldersheim:
@havardpe PTAL

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
```diff
@@ -56,11 +63,27 @@ SingleExecutor::setTaskLimit(uint32_t taskLimit) {
     _wantedTaskLimit = vespalib::roundUp2inN(taskLimit);
 }
 
+void
+SingleExecutor::drain(Lock & lock) {
+    uint64_t wp = _wp.load(std::memory_order_relaxed);
+    while (numTasks() > 0) {
+        _consumerCondition.notify_one();
+        sleepProducer(lock, 100us, wp);
```
havardpe: letting go of the lock here will allow multiple threads to call drain at the same time. this might be ok, but could lead to creating more new buffers than you really need.

havardpe (approved):

staging_vespalib/src/vespa/vespalib/util/singleexecutor.cpp
baldersheim: I think that is not a normal usecase.

PR #12542, opened 2020-03-12
Remove compileVersion from application and instance responses 

jonmv:
@mpolden please review.


jonmv:
The only client that expects this field is the compileVersion mojo, in the vespa-maven-plugin. This should be at `RELEASE` version, so should be OK to remove this field now.

mpolden (approved):

mpolden:
merge defrost

PR #12545, opened 2020-03-12
Restrict directly routed endpoints to qualified deployments

mpolden:
@tokle


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/RoutingController.java
```diff
@@ -249,4 +264,29 @@ public void removeEndpointsInDns(Instance instance) {
         return Collections.unmodifiableList(routingMethods);
     }
 
+    /** Returns whether traffic can be directly routed to given deployment */
+    private boolean canRouteDirectlyTo(DeploymentId deploymentId, Application application) {
+        if (controller.system().isPublic()) return true; // Public always supports direct routing
+
+        // Check Athenz service presence
```
tokle: Consider adding a comment why this is required. 

something like: "test framework currently use this identity when executing requests against containers"

tokle (approved):

tokle (approved):

PR #12548, opened 2020-03-12
Hmusum/do not use default temp dir for rank setup validation

hmusum:


baldersheim (commented):
I think it is better to control what is the tempdir by starting the container with an environment variable set.

java -Djava.io.tmpdir=/path/to/tmpdir
https://docs.oracle.com/javase/8/docs/api/java/io/File.html#createTempFile-java.lang.String-java.lang.String-java.io.File-

hmusum:
Good point, I will look into that

hmusum:
Superseded by https://github.com/vespa-engine/vespa/pull/12551

PR #12549, opened 2020-03-12
Bratseth/system info

bratseth:
@hmusum AND @gjoranv please review


hmusum (approved):
👍

hmusum (changes requested):
Fails abi check:
[INFO] --- abi-check-plugin:7-SNAPSHOT:abicheck (default) @ container-core ---
[ERROR] Extra class: ai.vespa.cloud.Environment
[ERROR] Extra class: ai.vespa.cloud.SystemInfo
[ERROR] Extra class: ai.vespa.cloud.Zone

container-core/src/main/java/ai/vespa/cloud/SystemInfo.java
```diff
@@ -1,5 +1,32 @@
...
+ * This is available and can be injected when running in a cloud environment.
+ *
+ * @author bratseth
+ */
 public class SystemInfo {
+
+    private final Zone zone;
+
+    /** Do not use */
+    @SuppressWarnings("unused")
```
gjoranv: Suppressing "unused" should be unnecessary when you have the Inject annotation.

container-core/src/main/java/ai/vespa/cloud/SystemInfo.java
```diff
@@ -1,5 +1,32 @@
...
+ * Provides information about the system in which this container is running.
+ * This is available and can be injected when running in a cloud environment.
+ *
+ * @author bratseth
+ */
 public class SystemInfo {
+
+    private final Zone zone;
+
+    /** Do not use */
```
gjoranv: I suggest to skip this comment. What does it mean?

container-core/src/main/java/ai/vespa/cloud/Zone.java
```diff
@@ -0,0 +1,61 @@
...
+        this.environment = environment;
+        this.region = region;
+    }
+
+    public Environment environment() { return environment; }
+    public String region() { return region; }
+
+    /** Returns the string environment.region */
+    @Override
+    public String toString() { return environment + "." + region; }
```
gjoranv: Perhaps this should be e.g. ´stringValue()´, and return a String that is less tempting to parse from toString.

gjoranv (approved):

container-core/src/main/java/ai/vespa/cloud/SystemInfo.java
bratseth: IntelliJ don't think so, but I guess it's clear enough with the annotation anyway.

container-core/src/main/java/ai/vespa/cloud/SystemInfo.java
bratseth: It means to not use it ... I probably want to switch to another config later which will break any who use this method.

container-core/src/main/java/ai/vespa/cloud/Zone.java
bratseth: I would agree if this was further in but I think this is fine for an "edge API" class.

PR #12551, opened 2020-03-12
Set Java system property java.io.tmpdir to $VESPA_HOME/tmp

hmusum:
There could be more places that this should be set ...


baldersheim (approved):
If not all, it is a step in the right direction.

freva:
We need some cleanup policy merged with/before this.

hmusum:
For ylinux 6 images this is done by the nodectl script when starting vespa (via yinst, see https://git.ouroath.com/vespa/vespa-yahoo/blob/master/vespascripts/src/nodectl-instance.sh#L172-L175). For ylinux 6 and centos images we need to do something, agreed.

hmusum:
Cleanup now done, see https://git.ouroath.com/vespa/vespa-yahoo/pull/3893 and https://git.ouroath.com/vespa/centos-build/pull/96

PR #12552, opened 2020-03-12
Arnej/polymorphic distance feature

arnej27959:

@geirst please review
@havardpe FYI, please review if you have some spare time

NOTE: I am very unsure whether this strategy for recognizing when to trigger old versus new behavior is sufficient, maybe we need to be even more paranoid, especially about triggering the new distance-from-labeled-item variant.


geirst (approved):
👍

PR #12556, opened 2020-03-13
Cleanup rank setup validator

hmusum:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/RankSetupValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/RankSetupValidator.java
@@ -158,7 +158,8 @@ public class RankSetupValidator extends Validator {
     }
 
     private void validateFail(String output, SearchCluster sc, String sdName, DeployLogger deployLogger) {
-        StringBuilder errMsg = new StringBuilder("For search cluster '" + sc.getClusterName() + "', search definition '" + sdName + "': error in rank setup. Details:\n");
+        StringBuilder errMsg = new StringBuilder("For search cluster '").append(sc.getClusterName()).append("', ")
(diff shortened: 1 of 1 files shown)


config-model/src/main/java/com/yahoo/vespa/model/application/validation/RankSetupValidator.java
```diff
@@ -147,27 +152,27 @@ private boolean execValidate(String configId, SearchCluster sc, String sdName, D
...
     private void validateWarn(Exception e, DeployLogger deployLogger) {
-        String msg = "Unable to execute 'vespa-verify-ranksetup', validation of rank expressions will only take place when you start Vespa: " +
+        String msg = "Unable to execute '"+ binaryName + "', validation of rank expressions will only take place when you start Vespa: " +
                 Exceptions.toMessageString(e);
         deployLogger.log(LogLevel.WARNING, msg);
     }
 
     private void validateFail(String output, SearchCluster sc, String sdName, DeployLogger deployLogger) {
-        String errMsg = "For search cluster '" + sc.getClusterName() + "', search definition '" + sdName + "': error in rank setup. Details:\n";
+        StringBuilder errMsg = new StringBuilder("For search cluster '" + sc.getClusterName() + "', search definition '" + sdName + "': error in rank setup. Details:\n");
```
baldersheim: For consistency you should use append to avoid the '+' entirely here to.

baldersheim (approved):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/RankSetupValidator.java
hmusum: Yes, apparently IntelliJ didn't fix all this when I accepted the suggested change from String to StringBuilder. Will fix

baldersheim (approved):
Merge at will

PR #12564, opened 2020-03-13
When you have a container with both feed and query, you risk that dur…

baldersheim:
…ing warmup the throughput optimized

feed pipeline will overwhelm the search pipeline during warmup.
Here we add an option to enforce a soft start of the feed pipeline, by gradually increasing max number of feed threads.
It might not be the perfect solution as mesages will pile up in the dynamic throttle policy, but that should have a builtin
softstart already. So I think this might a good fit.

@bratseth or @hmusum or @bjorncs PR
@jobergum and @yngveaasheim FYI

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
@@ -43,6 +43,9 @@ public class FeedHandlerV3 extends LoggingRequestHandler {
     protected final ReplyHandler feedReplyHandler;
     private final Metric metric;
     private final Object monitor = new Object();
+    private int remainingThreadsForFeedingAllowance;
+    private final long msBetweenBumpingMaxThreads;
+    private long nextTimeToAllocateAnotherThread;
     private final AtomicInteger threadsAvailableForFeeding;
     private static final Logger log = Logger.getLogger(FeedHandlerV3.class.getName());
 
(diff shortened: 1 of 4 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
```diff
@@ -43,6 +43,9 @@
     protected final ReplyHandler feedReplyHandler;
     private final Metric metric;
     private final Object monitor = new Object();
+    private int remainingThreadsForFeedingAllowance;
+    private final long msBetweenBumpingMaxThreads;
+    private long nextTimeToAllocateAnotherThread;
```
hmusum: Minor: I would have used Duration and Instant for these two

hmusum (approved):
Code looks good to me.

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
baldersheim: Thanks, now I will use that for time in java from now on.

hmusum (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
```diff
@@ -43,6 +43,9 @@
     protected final ReplyHandler feedReplyHandler;
     private final Metric metric;
     private final Object monitor = new Object();
+    private int remainingThreadsForFeedingAllowance;
```
bjorncs: Why are these member variables not protected by a monitor?

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
```diff
@@ -66,17 +68,17 @@ public FeedHandlerV3(LoggingRequestHandler.Context parentCtx,
             remainingThreadsForFeedingAllowance = Math.max((int) (0.4 * threadpoolConfig.maxthreads()), 1);
             if (threadpoolConfig.softStartSeconds() > 0.0) {
                 threadsAvailableForFeeding = new AtomicInteger(0);
-                msBetweenBumpingMaxThreads = (long)threadpoolConfig.softStartSeconds() * 1000 / remainingThreadsForFeedingAllowance;
+                timeBetweenBumpingMaxThreads = Duration.ofMillis((long)(threadpoolConfig.softStartSeconds() * 1000) / remainingThreadsForFeedingAllowance);
             } else {
                 threadsAvailableForFeeding = new AtomicInteger(remainingThreadsForFeedingAllowance);
                 remainingThreadsForFeedingAllowance = 0;
-                msBetweenBumpingMaxThreads = 0;
+                timeBetweenBumpingMaxThreads = null;
```
bjorncs: Consider using `Duration.ZERO` as sentinel value instead.

bjorncs (commented):
See comments

PR #12565, opened 2020-03-13
Revert "Bratseth/search to schema"

bratseth:
Reverts vespa-engine/vespa#12563

Some test failures I think.


hmusum:
👍

hmusum:
Looks like the change started writing schemas/searchdefinitions only to `schemas`. This led to downgrade of config servers not working without manual intervention.

bratseth:
Guess I'll better do this in 2 stages then

PR #12566, opened 2020-03-13
Gjoranv/allow stealing builtin handler bindings  

gjoranv:
Should status.html also be protected from stealing?

I didn't add an `isHosted` check before throwing when a reserved binding is used. If you try to steal a binding today, the built-in handler wins, but deployment does not fail.


bratseth (approved):
Great! @andreer do we still need require status.html on containers?

andreer:
Absolutely yes. It's the check we use from load balancers to see if the container is up and healthy (and transitively, if the content cluster is healthy enough that we should serve from that zone). It's important enough that I'm wondering if I don't understand the context here :-)

bratseth:
I think you do, I just can never remember who uses this vs. state/v1.

PR #12572, opened 2020-03-16
Add count for number of times a message has been inhibited. 

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -311,6 +311,12 @@ FileStorHandlerImpl::updateMetrics(const MetricLockGuard &)
                 entry.second->addTotalValueWithCount(m.getTotal(), m.getCount());
             }
         }
+        for (auto & entry : disk.metrics->inhibited.getMetricMap()) {
+            metrics::LoadType loadType(entry.first, "ignored");
+            for (const auto & stripe : disk.metrics->stripes) {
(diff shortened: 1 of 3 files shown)


storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -945,6 +951,7 @@ FileStorHandlerImpl::Stripe::getNextMessage(uint32_t timeout, Disk & disk)
         PriorityIdx::iterator iter(idx.begin()), end(idx.end());
 
         while (iter != end && operationIsInhibited(guard, iter->_bucket, *iter->_command)) {
+            _metrics->inhibited[iter->_command->getLoadType()].inc();
```
vekterli: As discussed offline (well, technically online due to the current circumstances...!) the retry mechanics of `getNextMessage` means that this metric risks over-counting when attempting to dequeue a message.

vekterli (commented):
We definitely want more insight into queue contention and the underlying reasons, though I'm wondering if we shouldn't instead add the following metrics:
* Duration of time (rather than count) spent waiting for lock to be released. We can make this lazy so that it doesn't incur time tracking costs in the common case where there is no lock contention.
* Separate tracking of queue size and wait time for client operations vs maintenance operations. Merge and GC ops (the latter in particular) tend to skew these metrics wildly due their relatively lower priorities than client ops. Possibly further subdivided into read/write categories or even having a dimension per operation type.

To maintain backwards compatibility we'd likely need these to be separate from todays metrics.

baldersheim:
Will do something else.

PR #12576, opened 2020-03-16
Use docker image repo in ClusterSpec

hmusum:


config-provisioning/src/main/java/com/yahoo/config/provision/HostSpec.java
```diff
@@ -77,6 +88,7 @@ public HostSpec(String hostname, List<String> aliases, Optional<Flavor> flavor,
         this.version = Objects.requireNonNull(version, "Version cannot be null but can be empty");;
         this.networkPorts = Objects.requireNonNull(networkPorts, "Network ports cannot be null but can be empty");;
         this.requestedResources = Objects.requireNonNull(requestedResources, "RequestedResources cannot be null");
+        this.dockerImageRepo = Objects.requireNonNull(dockerImageRepo, "Version cannot be null but can be empty");;
```
freva:
```suggestion
        this.dockerImageRepo = Objects.requireNonNull(dockerImageRepo, "Docker image repo cannot be null but can be empty");
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/InfraDeployerImpl.java
```diff
@@ -88,7 +88,7 @@ public void prepare() {
                 Version targetVersion = infrastructureVersions.getTargetVersionFor(nodeType);
                 hostSpecs = provisioner.prepare(
                         application.getApplicationId(),
-                        application.getClusterSpecWithVersion(targetVersion),
+                        application.getClusterSpecWithVersion(targetVersion, Optional.empty()), // TODO: Use dockerImageRepo from zk
```
freva: Does this apply here? We are potentially allocating new nodes here, which do not have any docker image associated with them. Unless you mean the docker image from https://github.com/vespa-engine/vespa/blob/master/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/DockerImages.java
?

config-provisioning/src/test/java/com/yahoo/config/provision/serialization/AllocatedHostsSerializerTest.java
```diff
@@ -72,6 +73,7 @@ private void assertAllocatedHosts(AllocatedHosts expectedHosts, NodeFlavors conf
             assertEquals(expectedHost.networkPorts(), deserializedHost.networkPorts());
             assertEquals(expectedHost.aliases(), deserializedHost.aliases());
             assertEquals(expectedHost.requestedResources(), deserializedHost.requestedResources());
+            assertEquals(expectedHost.dockerImageRepo().orElse(""), deserializedHost.dockerImageRepo().orElse(""));
```
freva: I think you can just compare the Optionals directly
```suggestion
            assertEquals(expectedHost.dockerImageRepo(), deserializedHost.dockerImageRepo());
```

freva (approved):
LGTM.

I think the docker image repo should (at some point) become https://github.com/vespa-engine/vespa/blob/master/config-provisioning/src/main/java/com/yahoo/config/provision/DockerImage.java instead of string.

`ClusterSpec` should also just get a builder so we don't need to update 20 tests every time we add a new field. Plus the builder methods could take in things as VespaVersion as both `String` or `Version`, so 
```
ClusterSpec.request(ClusterSpec.Type.content, ClusterSpec.Id.from("test"),
Version.fromString("6.42"), false, Optional.empty(), Optional.empty())
```
Would be something like
```
ClusterSpec.builder(ClusterSpec.Type.content, "test").version("6.42").build()
```

hmusum:
Yeah, I agree that we should have a builder for ClusterSpec, will look into it (in another PR).

Regarding DockerImage: Yes, I thought about using that now, but decided to wait. I'll look at that in a forthcoming PR.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/InfraDeployerImpl.java
hmusum: No, probably not. I'll just remove it for now.

freva (approved):

PR #12578, opened 2020-03-16
Consume metrics timestamps in seconds not ms

bratseth:


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTest.java
```diff
@@ -141,4 +145,33 @@ public void testAutoscalingAws() {
                                tester.autoscale(application1, cluster1));
     }
 
+    @Test
```
freva: Remove

bratseth: tx

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/NodeMetricsDbTest.java
```diff
@@ -18,15 +18,15 @@ public void testNodeMetricsDb() {
...
         List<NodeMetrics.MetricValue> values = new ArrayList<>();
         for (int i = 0; i < 40; i++) {
-            values.add(new NodeMetrics.MetricValue("host0", "cpu.util", clock.instant().toEpochMilli(), 0.9f));
+            values.add(new NodeMetrics.MetricValue("host0", "cpu.util", clock.instant().getEpochSecond(), 0.9f));
             clock.advance(Duration.ofHours(1));
         }
         db.add(values);
 
-        assertEquals(30, db.getWindow(clock.instant().minus(Duration.ofHours(30)), Resource.cpu,    List.of("host0")).measurementCount());
+        assertEquals(29, db.getWindow(clock.instant().minus(Duration.ofHours(30)), Resource.cpu,    List.of("host0")).measurementCount());
```
freva: Why did this happen?

freva (approved):

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/NodeMetricsDbTest.java
bratseth: We're now going through an API which uses seconds which round the current time to the nearest thousand, which causes the window to end up right at the boundary of the 30th value.

freva: Oh, so there was 1/1000 chance this test would fail before? If it ran exactly on the second?

bratseth: Yes, I think so. "Problem" here is I don't really care precisely how many ends up in the window ..

PR #12581, opened 2020-03-16
Add feature flag for disabling RoutingGenerator

mpolden:
Because just like life, `config-model` is like a box of chocolate, you never
know what you're gonna get.

@jonmv


jonmv (approved):
Gift that keeps on giving

jonmv:
The gift that keeps on giving :)

PR #12589, opened 2020-03-17
Add builder for ClusterSpec 

hmusum:


bratseth (approved):
LGTM, and already an improvement so feel free to merge. The ClusterSpec is used either to request a desired cluster, or to specify an existing one (perhaps including group). I wonder if we couldn't keep the explicit separation between a request and the specification, like e.g replace "builder" by "request" and "specification" and have a boolean in the builder which only allows the correct fields for that context?

hmusum:
That separation was one of the things I wondered if we should somehow keep. Thanks for the suggestion, I'll look into it.

hmusum:
@bratseth please take a look again

bratseth:
Great! You could also validate that a request don't set group (and vespa-version?).

PR #12590, opened 2020-03-17
Reduce log output from NodeMetricsDbMaintainer

mpolden:
Fails frequently on connection refused. Observed in test.us-east-1.

@bratseth


bratseth (approved):

bratseth:
Happens a lot more in test probably.

PR #12596, opened 2020-03-17
Bratseth/search to shema take 2.

bratseth:


bratseth:
Must be merged at the same time as https://github.com/vespa-engine/system-test/pull/601

hmusum (approved):

PR #12597, opened 2020-03-17
Act on the readable subset of applications when appropriate

mpolden:
@jonmv


mpolden:
Please confirm if we have the same views on what's appropriate. 🙂

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
```diff
@@ -939,4 +941,25 @@ public void testChangeEndpointCluster() {
...
+        db.curator().set(Path.fromString("/controller/v1/applications/t2:a2"), new byte[]{(byte) 0xDE, (byte) 0xAD});
+        assertEquals(1, tester.applications().readable().size());
+
+        try {
+            tester.applications().asList();
+            fail("Expected exception");
+        } catch (Exception ignored) {
+        }
+    }
+
```
jonmv: Perhaps add a test with a deployment scenario, where one app is unreadable?

jonmv (approved):
Looks good. A bit scary, so I'd like some more testing.

mpolden:
PTAL at two last commits.

jonmv (approved):
Great!

PR #12600, opened 2020-03-17
Order the bits so it is more suitable as a hash value.

baldersheim:
@toregge or @geirst PR
I think I did this some time ago too.
But I suspect it got lost in a rewrite.......


geirst (approved):
👍

vespalib/src/vespa/vespalib/datastore/entryref.h
```diff
@@ -32,9 +32,8 @@ class EntryRefT : public EntryRef {
     EntryRefT() : EntryRef() {}
     EntryRefT(size_t offset_, uint32_t bufferId_);
     EntryRefT(const EntryRef & ref_) : EntryRef(ref_.ref()) {}
-    uint32_t hash() const { return offset() + (bufferId() << OffsetBits); }
```
toregge: No replacement for hash() method ?

baldersheim: It was not used at all.

PR #12601, opened 2020-03-17
Bjorncs/implicit access control

bjorncs:


config-model/src/main/java/com/yahoo/vespa/model/container/http/xml/HttpBuilder.java
```diff
@@ -79,6 +82,28 @@ private AccessControl buildAccessControl(DeployState deployState, AbstractConfig
         return builder.build();
     }
 
+    // TODO Fail if domain is not provided through deploy properties
+    private static AthenzDomain getAccessControlDomain(DeployState deployState, Element accessControlElem) {
+        AthenzDomain tenantDomain = deployState.getProperties().athenzDomain().orElse(null);
```
tokle: will this fail the first redeploy (after config server is upgraded) ? or will this only be evaluated when application is deployed with the new version?

tokle (approved):

config-model/src/main/java/com/yahoo/vespa/model/container/http/xml/HttpBuilder.java
bjorncs: Only if `domain` attribute is missing from `access-control`. This attribute has been mandatory up to now, so it should not be the case.

tokle (approved):

PR #12603, opened 2020-03-18
Revert "Missing node specification allocates one node per cluster"

hmusum:
Reverts vespa-engine/vespa#12580

Breaks when prod clusters are required to have at least 2 nodes


bratseth (approved):

mpolden:
👍

PR #12616, opened 2020-03-18
Avoid trailing query delimiter when no query

jonmv:
@mpolden please review.


mpolden (approved):

jonmv:
Seems it should handle the mixed case too, just to be user friendly. It's more forgiving that way, and shouldn't cause any surprises.

jonmv:
PTAL.

mpolden (approved):

PR #12617, opened 2020-03-18
use RPC directly instead of calling external process

arnej27959:

I think this is more straightforward than going via vespa-sentinel-cmd utility.
@baldersheim please review
@gjoranv @hmusum FYI


baldersheim:
@bjorncs FYI

baldersheim (approved):
Very good :)

bjorncs (approved):
👍

PR #12618, opened 2020-03-18
Balder/use fast map for few elements.

baldersheim:
@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/features/dotproductfeature.cpp
+++ b/searchlib/src/vespa/searchlib/features/dotproductfeature.cpp
@@ -196,10 +196,9 @@ DotProductExecutorByEnum::execute(uint32_t docId) {
     outputs().set_number(0, val);
 }
 
-class SingleDotProductByEnum final : public fef::FeatureExecutor {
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/features/dotproductfeature.cpp
```diff
@@ -584,6 +655,10 @@ createForDirectWSetImpl(const IAttributeVector * attribute, V && vector, vespali
     if (!attribute->isImported() && (iattr != nullptr) && supportsGetRawValues<A, VT>(*iattr)) {
         auto * exactA = dynamic_cast<const ExactA *>(iattr);
         if (exactA != nullptr) {
+            if (extractSize(vector) == 1) {
+                const auto & elem = extractElem(vector, 0ul);
```
havardpe: could let auto resolve to a value here and move it into the constructor

havardpe (approved):

searchlib/src/vespa/searchlib/features/dotproductfeature.cpp
baldersheim: Just removed the temporary.

PR #12621, opened 2020-03-19
change NNS rawscore so it is a closeness not a distance

arnej27959:


searchlib/src/tests/queryeval/nearest_neighbor/nearest_neighbor_test.cpp
```diff
@@ -17,6 +17,8 @@
 #include <vespa/log/log.h>
 LOG_SETUP("nearest_neighbor_test");
 
+#define EPS 1.0e-6
```
geirst: Prefer constexpr instead

geirst (approved):
👍

PR #12625, opened 2020-03-19
Reuse Supervisor instance

bjorncs:
This will allow the client to reuse existing SSL sessions

FYI @baldersheim @gjoranv


arnej27959 (approved):
this looks good to me (i considered doing something like this already, but didn't want to change to much in one go...)

gjoranv (approved):

PR #12626, opened 2020-03-19
Serialize and deserialize docker image repo

hmusum:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
@@ -187,8 +187,7 @@ public class NodeSerializer {
         object.setLong(currentRestartGenerationKey, allocation.restartGeneration().current());
         object.setBool(removableKey, allocation.isRemovable());
         object.setString(wantedVespaVersionKey, allocation.membership().cluster().vespaVersion().toString());
-        // TODO serialize dockerImageRepo
-        //object.setString(wantedDockerImageRepoKey, allocation.membership().cluster().dockerImageRepo().orElse(""));
+        object.setString(wantedDockerImageRepoKey, allocation.membership().cluster().dockerImageRepo().orElse(""));
         allocation.networkPorts().ifPresent(ports -> NetworkPortsSerializer.toSlime(ports, object.setArray(networkPortsKey)));
     }
 
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
@@ -187,7 +187,7 @@ public class NodeSerializer {
         object.setLong(currentRestartGenerationKey, allocation.restartGeneration().current());
         object.setBool(removableKey, allocation.isRemovable());
         object.setString(wantedVespaVersionKey, allocation.membership().cluster().vespaVersion().toString());
-        object.setString(wantedDockerImageRepoKey, allocation.membership().cluster().dockerImageRepo().orElse(""));
+        allocation.membership().cluster().dockerImageRepo().ifPresent(repo -> object.setString(wantedDockerImageRepoKey, repo));
         allocation.networkPorts().ifPresent(ports -> NetworkPortsSerializer.toSlime(ports, object.setArray(networkPortsKey)));
     }
 
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodesResponse.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodesResponse.java
@@ -157,10 +157,9 @@ class NodesResponse extends HttpResponse {
             toSlime(allocation.membership(), object.setObject("membership"));
             object.setLong("restartGeneration", allocation.restartGeneration().wanted());
             object.setLong("currentRestartGeneration", allocation.restartGeneration().current());
-            String wantedVespaVersion = allocation.membership().cluster().vespaVersion().toFullString();
             object.setString("wantedDockerImage", allocation.membership().cluster().dockerImage()
-                    .orElse(nodeRepository.dockerImage(node).withTag(allocation.membership().cluster().vespaVersion()).asString()));
-            object.setString("wantedVespaVersion", wantedVespaVersion);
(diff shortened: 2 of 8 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
```diff
@@ -187,8 +187,7 @@ private void toSlime(Allocation allocation, Cursor object) {
         object.setLong(currentRestartGenerationKey, allocation.restartGeneration().current());
         object.setBool(removableKey, allocation.isRemovable());
         object.setString(wantedVespaVersionKey, allocation.membership().cluster().vespaVersion().toString());
-        // TODO serialize dockerImageRepo
-        //object.setString(wantedDockerImageRepoKey, allocation.membership().cluster().dockerImageRepo().orElse(""));
+        object.setString(wantedDockerImageRepoKey, allocation.membership().cluster().dockerImageRepo().orElse(""));
```
freva: I realize the deserializer handles empty string, but consider just to serializing dockerImageRepo if it's not set.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodesResponse.java
```diff
@@ -157,8 +157,11 @@ private void toSlime(Node node, boolean allFields, Cursor object) {
             toSlime(allocation.membership(), object.setObject("membership"));
             object.setLong("restartGeneration", allocation.restartGeneration().wanted());
             object.setLong("currentRestartGeneration", allocation.restartGeneration().current());
-            object.setString("wantedDockerImage", nodeRepository.dockerImage(node).withTag(allocation.membership().cluster().vespaVersion()).asString());
-            object.setString("wantedVespaVersion", allocation.membership().cluster().vespaVersion().toFullString());
+            String wantedVespaVersion = allocation.membership().cluster().vespaVersion().toFullString();
+            Optional<String> dockerImageRepo = allocation.membership().cluster().dockerImageRepo();
```
freva: Consider having a method on `ClusterSpec` that returns the full docker image

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/ProvisioningTest.java
```diff
@@ -156,6 +157,33 @@ public void nodeVersionIsReturnedIfSet() {
...
+
+        // deploy
+        ApplicationId application1 = tester.makeApplicationId();
+        SystemState state1 = prepare(application1, tester, 1, 1, 1, 1, defaultResources, "1.2.3");
+        String dockerImageRepo = "docker.domain.tld/my/image";
+        prepare(application1, tester, 1, 1, 1 , 1 , false, defaultResources, "1.2.3", Optional.of(dockerImageRepo));
+        tester.activate(application1, state1.allHosts);
+
+        HostSpec host1 = state1.container0.iterator().next();
+        assertFalse(host1.version().isPresent());
```
freva: Why is this the case?

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/ProvisioningTest.java
```diff
@@ -156,6 +157,33 @@ public void nodeVersionIsReturnedIfSet() {
...
+        Node node1 = tester.nodeRepository().getNode(host1.hostname()).get();
+        tester.nodeRepository().write(node1.with(node1.status().withDockerImage(DockerImage.fromString(dockerImageRepo))), () -> {});
```
freva: This actually shouldn't be allowed, current docker image should always include the version as well.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
hmusum: Yeah, fixed

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodesResponse.java
hmusum: Done

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/ProvisioningTest.java
hmusum: Should not have been there, removed

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/ProvisioningTest.java
hmusum: test code updated

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/v2/NodesResponse.java
```diff
@@ -157,8 +157,10 @@ private void toSlime(Node node, boolean allFields, Cursor object) {
             toSlime(allocation.membership(), object.setObject("membership"));
             object.setLong("restartGeneration", allocation.restartGeneration().wanted());
             object.setLong("currentRestartGeneration", allocation.restartGeneration().current());
-            object.setString("wantedDockerImage", nodeRepository.dockerImage(node).withTag(allocation.membership().cluster().vespaVersion()).asString());
-            object.setString("wantedVespaVersion", allocation.membership().cluster().vespaVersion().toFullString());
+            String wantedVespaVersion = allocation.membership().cluster().vespaVersion().toFullString();
+            object.setString("wantedDockerImage", allocation.membership().cluster().dockerImage()
+                    .orElse(nodeRepository.dockerImage(node).withTag(allocation.membership().cluster().vespaVersion()).asString()));
+            object.setString("wantedVespaVersion", wantedVespaVersion);
```
freva:
```suggestion
            object.setString("wantedDockerImage", allocation.membership().cluster().dockerImage()
                    .orElseGet(() -> nodeRepository.dockerImage(node).withTag(allocation.membership().cluster().vespaVersion()).asString()));
            object.setString("wantedVespaVersion", allocation.membership().cluster().vespaVersion().toFullString());
```
(Changing back to `orElseGet()` and setting `wantedVespaVersion` directly)

freva (approved):
:+1: 

Is there any plan to default to some `dockerImageRepo` so `ClusterSpec` always has it? Or maybe just get rid of `docker-image-override` flag and use default for node type?

hmusum:
Yes, I want to get rid of `docker-image-override` flag and use default for node type after this goes in and we have start using this instead

PR #12629, opened 2020-03-19
Use a common accelrator instance

baldersheim:
@havardpe PR

### DIFF
--- a/vespalib/src/vespa/vespalib/hwaccelrated/iaccelrated.cpp
+++ b/vespalib/src/vespa/vespalib/hwaccelrated/iaccelrated.cpp
@@ -106,7 +106,7 @@ class RuntimeVerificator
 public:
     RuntimeVerificator();
 private:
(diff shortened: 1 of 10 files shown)


vespalib/src/vespa/vespalib/hwaccelrated/iaccelrated.cpp
```diff
@@ -156,9 +156,16 @@ RuntimeVerificator _G_verifyAccelrator;
 
 
 IAccelrated::UP
-IAccelrated::getAccelrator()
+IAccelrated::createAccelrator()
```
havardpe: can we just remove this?

havardpe (approved):

vespalib/src/vespa/vespalib/hwaccelrated/iaccelrated.cpp
baldersheim: Done

PR #12635, opened 2020-03-19
Revert "Satisfy capacity requirement when allocating without nodes tag"

hmusum:
Reverts vespa-engine/vespa#12610

Short story: We cannot do topology changes.


bratseth (approved):

bratseth:
Won't this just make us go thrugh the whole thing again on the next release though?

hmusum:
The apps that haven no `nodes` element for containers are able to upgrade now (they fail in staging). In staging applications will not converge, since the code checking node convergence will find that a node reserved to the app has not its expected Vespa version (a node that is allocated in the newest config model, but not in the version the nodes will be running on). More info in ticket about 7.195 broken release

bratseth:
Ok then ...

PR #12636, opened 2020-03-19
Fix HK2 service reification failure

hakonhall:
The jaxrs_client_utils tests used to dump lots of warnings of the format:

```
WARNING: HK2 service reification failed for [org.glassfish.jersey.message.internal.DataSourceProvider] with an exception:
MultiException stack 1 of 2
java.lang.NoClassDefFoundError: javax/activation/DataSource
	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
```

### DIFF
--- a/jaxrs_client_utils/pom.xml
+++ b/jaxrs_client_utils/pom.xml
@@ -81,6 +81,16 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.sun.activation</groupId>
+      <artifactId>javax.activation</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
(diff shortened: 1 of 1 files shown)


hakonhall:
This adds them to the test scope. It's not entirely clear to me whether 1. another scope would be better (provided I presume), or 2. whether there are other tests that need similar treatment.

toregge (approved):
@gjoranv : FYI

gjoranv:
👍I assume that those deps are provided in all runtime scenarios, at least inside a Jdisc container, so that's probably why you don't see those warnings there. Test scope is probably fine.

hakonhall:
it's compile scope in jdisc_core, which I found a bit surprising, but it might be enough depending on which bundle that reflection stuff is happening at run time. And the *-dependencies-enforcer has lists e.g. com.sun.activation as provided!?

gjoranv:
jdisc_core classes are provided from the classpath.

PR #12637, opened 2020-03-19
Bratseth/nicer autoscaling logging

bratseth:


bratseth:
Just some better formatting to make these easier to read

freva (approved):

PR #12639, opened 2020-03-20
Use system-dependent timeouts in deployment jobs

jonmv:


hmusum (approved):
Great!

We might want to have different settings for different environments too, but we can look at that later (e.g. we wait 2 hours for a container in test environment to come up (`noNodesDownTimeout`), we should probably give up after 30 minutes or an hour)

jonmv:
Yes, definitely. Adding that should be easy now.

jonmv:
.

PR #12654, opened 2020-03-20
Revert "Hold a JVM-wide reentrant lock to grab mutex — helps ZK stale reads?" 

jonmv:
Reverts vespa-engine/vespa#11815

@hakonhall let's revert if this doesn't help. We should know within a week or two.


hakonhall (approved):

jonmv:
Done elsewhere.

PR #12655, opened 2020-03-20
Wait for installReal step before doing report step

jonmv:
@hmusum please review.


hmusum (approved):

jonmv:
Report step has been run too early since separation of production test and deployments.

PR #12656, opened 2020-03-20
Revert "Revert "Since the most important aspect our use of hash maps/sets is for spee…""

baldersheim:
Reverts vespa-engine/vespa#12653


baldersheim:
@hmusum PR
Only the last one matters.

hmusum (approved):

PR #12657, opened 2020-03-20
Revert "Set phrase-segmenting default off"

hakonhall:
Reverts vespa-engine/vespa#12649

This might be the cause of the specialtokens_dotnet__ELASTIC system test failure?

```
<"my_str:\"yahoo .net\""> expected but was
<nil>.
```

https://factory-web.vespa.corp.yahoo.com/versions/1/builds/54647/products/1/tests/279720/testruns/39494814


arnej27959:
most likely this test needs updating, and should test with phrase segmenting both on and off

bratseth:
Phrase segmentation is just required spuriously by this test. I fixed the test so reverting this.

PR #12663, opened 2020-03-23
Add benchmark for xxhash32, xxhash64 and legacy

baldersheim:
@vekterli PR

### DIFF
--- a/vespalib/src/tests/stllike/lookup_benchmark.cpp
+++ b/vespalib/src/tests/stllike/lookup_benchmark.cpp
@@ -2,11 +2,11 @@
 #include <cstddef>
 #include <cstdlib>
 #include <cstdio>
-#include <map>
 #include <set>
 #include <unordered_set>
 #include <vector>
-#include <algorithm>
+//#define XXH_INLINE_ALL
+#include <xxhash.h>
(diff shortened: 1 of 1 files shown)


baldersheim:
Ouch, pushed green button to soon.

vespalib/src/tests/stllike/lookup_benchmark.cpp
```diff
@@ -95,6 +95,41 @@ size_t benchHashMapVespaLib2(size_t sz, size_t numLookups)
...
+}
+
+size_t benchLegacyHash(size_t sz, size_t numLookups) {
+    auto data = createData(sz);
+    size_t sum(0);
+    for (size_t i(0); i < numLookups; i++) {
+        sum += vespalib::hashValue(data.get(), sz);
+    }
+    return sum;
+}
```
vekterli: Since all these benchmarks are the same aside from the evaluation call, could consider factoring out as template argument to a generic benchmark loop function.

vekterli (commented):
:+1:

vespalib/src/tests/stllike/lookup_benchmark.cpp
baldersheim: Yes, I will make one more round.

PR #12665, opened 2020-03-23
fix provisioning race condition

andreer:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/Yum.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/Yum.java
@@ -118,25 +118,25 @@ public class Yum {
         //       - "Nothing to do"
         //     And in case we need to downgrade and return true from converge()
 
-        CommandLine commandLine = terminal.newCommandLine(context).add("yum", "install");
-        for (String repo : repos) commandLine.add("--enablerepo=", repo);
-        commandLine.add("--assumeyes", yumPackage.toName());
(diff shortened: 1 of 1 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/Yum.java
```diff
@@ -118,9 +118,9 @@ public boolean installFixedVersion(TaskContext context, YumPackageName yumPackag
         //       - "Nothing to do"
         //     And in case we need to downgrade and return true from converge()
 
-        CommandLine commandLine = terminal
-                .newCommandLine(context)
-                .add("yum", "install", "--assumeyes", yumPackage.toName());
+        CommandLine commandLine = terminal.newCommandLine(context).add("yum", "install");
+        for (String repo : repos) commandLine.add("--enablerepo=", repo);
```
freva: This will result in tokens like
"yum" "install" "--enablerepo=" "non-core" "--asumeyes" "nano"
which it parses as enable repo empty string, and `non-core` as yum command.

freva: These repos also need to be handled when downgrade is required (Line 131)

freva (commented):
I think `installFixedVersion()` should return an instance similar to `GenericYumCommand` that has methods to enable (and maybe in the future disable repos, etc)

andreer:
Was not able to refactor installFixedVersion today, but this is super annoying so I did a quick fix for now. PTAL

freva (approved):

PR #12666, opened 2020-03-23
Remove this from public packages

bratseth:
By our contract this requires a major release, but we should
make an exception because:
- This is not useful outside the cloud environment
- On the cloud environment there are just 2 usages, both of which are legacy
  and won't be impacted (as we're still exporting the packages).

The replacement is ai.vespa.cloud.SystemInfo

@kkraune AND @hmusum please review


hmusum (changes requested):
abi-spec.json needs to be updated

kkraune (approved):

hmusum (approved):

PR #12669, opened 2020-03-23
Use c++11 for loops and use std::move.

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/distributor/externaloperationhandler.cpp
+++ b/storage/src/vespa/storage/distributor/externaloperationhandler.cpp
@@ -241,7 +241,8 @@ void ExternalOperationHandler::bounce_or_invoke_read_only_op(
 
 IMPL_MSG_COMMAND_H(ExternalOperationHandler, Put)
 {
-    auto& metrics = getMetrics().puts[cmd->getLoadType()];
+    const documentapi::LoadType & loadType = cmd->getLoadType();
+    auto& metrics = getMetrics().puts[loadType];
     if (!checkTimestampMutationPreconditions(*cmd, getBucketId(cmd->getDocumentId()), metrics)) {
         return true;
     }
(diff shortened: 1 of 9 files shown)


storage/src/vespa/storage/distributor/externaloperationhandler.cpp
```diff
@@ -252,9 +253,10 @@ IMPL_MSG_COMMAND_H(ExternalOperationHandler, Put)
 
     auto handle = _mutationSequencer.try_acquire(cmd->getDocumentId());
     if (allowMutation(handle)) {
+        document::BucketSpace bucketSpace = cmd->getBucket().getBucketSpace();
         _op = std::make_shared<PutOperation>(*this,
-                                             _bucketSpaceRepo.get(cmd->getBucket().getBucketSpace()),
-                                             cmd, getMetrics().puts[cmd->getLoadType()], std::move(handle));
+                                             _bucketSpaceRepo.get(bucketSpace),
+                                             std::move(cmd), getMetrics().puts[loadType], std::move(handle));
```
vekterli: For all these callsites, `cmd` is a `const std::shared_ptr<T>&`, so the move won't have any effect in practice.

This is due to how the legacy storage links/chains work, where the message is always passed to handlers (as a const ref to the `shared_ptr`) and the handler returns whether it _handled_ the message after the fact. Since ownership is not known a-priori, moves cannot be done optimistically and using by-value would be too expensive.

vekterli (approved):
:+1: see comment re: moves not having much effect as things currently are in the distributor. This part of the code would do well with a redesign that explicitly takes ownership into account, though it's not a minor task since it touches a _lot_ of components on both the distributor and content node.

storage/src/vespa/storage/distributor/externaloperationhandler.cpp
baldersheim: Yes, I figured that out. I just wanted to pull it as far out as possible, so when code is rewritten it will work optimal :)
At least one move into the PendingTrackerImpl takes effect.

PR #12670, opened 2020-03-23
Add wanted and current docker image to Node

hmusum:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
@@ -3,6 +3,7 @@ package com.yahoo.vespa.hosted.controller.api.integration.configserver;
 
 import com.yahoo.component.Version;
 import com.yahoo.config.provision.ApplicationId;
+import com.yahoo.config.provision.DockerImage;
 import com.yahoo.config.provision.HostName;
 import com.yahoo.config.provision.NodeResources;
 import com.yahoo.config.provision.NodeType;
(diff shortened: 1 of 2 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/NodeRepository.java
```diff
@@ -122,7 +123,9 @@ private static Node toNode(NodeRepositoryNode node) {
                         clusterTypeOf(node.getMembership()),
                         node.getWantToRetire(),
                         node.getWantToDeprovision(),
-                        Optional.ofNullable(node.getReservedTo()).map(TenantName::from));
+                        Optional.ofNullable(node.getReservedTo()).map(TenantName::from),
+                        DockerImage.fromString(node.getWantedDockerImage()),
```
freva: Current & wanted docker image fields are nullable, this needs to be handled similar to `versionFrom()`

hmusum: Yes, thanks, fixed

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
```diff
@@ -324,6 +341,16 @@ public Builder wantedOsVersion(Version wantedOsVersion) {
             return this;
         }
 
+        public Builder currentOsVersion(DockerImage currentDockerImage) {
```
freva:
```suggestion
        public Builder currentDockerImage(DockerImage currentDockerImage) {
```

freva (approved):

PR #12681, opened 2020-03-24
Remove more user tenant leftovers 

jonmv:
@mpolden please review.


jonmv:
I guess the handling code should go as well, but we'll kill the dashboard when we do ... :)

mpolden (approved):
Dashboard disappears real soon now. Right, @andreer? 🙂

andreer:
Yes!

jonmv:
.

PR #12682, opened 2020-03-24
Support api key variable for signature based test framework controlle…

jonmv:
@kkraune please review and merge.


kkraune (approved):
thanks!

PR #12685, opened 2020-03-24
- Avoid adding empty traces that you need to carry on.

baldersheim:
- Avoid copying shared pointers when not necessary.
- Some c++11 improvements.

@vekterli PR


messagebus/src/vespa/messagebus/routing/routingnode.cpp
```diff
@@ -67,10 +67,8 @@ RoutingNode::~RoutingNode()
 void
 RoutingNode::clearChildren()
 {
-    for (std::vector<RoutingNode*>::iterator it = _children.begin();
-         it != _children.end(); ++it)
-    {
-        delete *it;
+    for (auto child : _children) {
```
vekterli: Could consider `auto*` here to make it more immediately obvious it's a raw pointer (also goes for other instances of for-each iteration over pointer containers)

vekterli (approved):
:+1:

messagebus/src/vespa/messagebus/routing/routingnode.cpp
baldersheim: Done

PR #12690, opened 2020-03-24
rank setup stack usage

havardpe:
With large models comes large stack usage.

- compile in separate thread for deterministic stack size
  (verify rank setup vs. proton config setup)

- warn about stack usage when over 128K
  (note that recursive decent parsing of rank expressions comes on top
   of the usage reported by the blueprint resolver)

@baldersheim please review
@lesters FYI


baldersheim (approved):
👍

baldersheim:
Merge at will when green.

havardpe:
the log claims the build is ok, but it is still yellow

PR #12694, opened 2020-03-24
Use name of the consumer instead of the metric set.

gjoranv:
I assume that you intended to use the Vespa consumer (the one with a lot of metrics). The current name is invalid, so you get the 'default' instead (the one with the "public" metrics).

FYI: @yngveaasheim

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/NodeMetricsFetcher.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/NodeMetricsFetcher.java
@@ -59,7 +59,7 @@ public class NodeMetricsFetcher extends AbstractComponent implements NodeMetrics
                                                           .stream()
                                                           .findFirst();
         if (metricsV2Container.isEmpty()) return Collections.emptyList();
-        String url = "http://" + metricsV2Container.get().hostname() + ":" + 4080 + apiPath + "?consumer=Vespa";
+        String url = "http://" + metricsV2Container.get().hostname() + ":" + 4080 + apiPath + "?consumer=default";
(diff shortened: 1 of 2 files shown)


bratseth:
Hmm, I really just need the "system" metrics now but I'll need the "vespa" ones later. So this was my attempt at requesting one which contained both, since we have this:

```
public class VespaMetricsConsumer {

    public static final String VESPA_CONSUMER_ID = VespaMetrics.VESPA_CONSUMER_ID.id;

    private static final MetricSet vespaConsumerMetrics = new MetricSet("vespa-consumer-metrics",
                                                                        emptyList(),
                                                                        ImmutableList.of(vespaMetricSet,
                                                                                         systemMetricSet,
                                                                                         networkMetricSet));

    public static MetricsConsumer getVespaMetricsConsumer() {
        return new MetricsConsumer(VESPA_CONSUMER_ID, vespaConsumerMetrics);
    }

}
```

But maybe that isn't enough to make this metric set exist?

bratseth:
Ok, I get it now. Just needed to dig deeper into the complexities of this. So, the metrics consumer has an id, which is not actually mentioned in its consumer class, and it also defines a metric set, which has a different id, which should not be used. I think I should request "default" instead for now then, to avoid unnecessarily transferring so many metrics, and probably we should create a separate consumer for this at one point.

Thanks!

bratseth (approved):

gjoranv:
The consumer classes are just a container for two static fields (consumer id and the metric set) and a static getter for the consumer itself. I guess we could use the same id for both, although they are different concepts. A consumer must have a metric set (potentially with children), but a metric set does not necessarily map to a consumer.

PR #12708, opened 2020-03-25
Bjorncs/access control

bjorncs:


tokle (approved):

bjorncs:
@gjoranv Please do a post-merge review (we need this fix out quickly to production). I will follow up with a new PR for any requested changes from review.

config-model/src/test/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilderTest.java
```diff
@@ -857,6 +858,34 @@ public void access_control_is_implicitly_added_for_hosted_apps() {
         assertThat(accessControl.domain, equalTo(tenantDomain.value()));
     }
 
+    @Test
+    public void access_control_is_implicitly_added_for_hosted_apps_with_existing_http_element() {
```
gjoranv: Please move to AccessControlTest

gjoranv:
I would prefer using volatiles instead of synchronized where appropriate. It generates less code noise, and is less prune to deadlocks, although the latter is not relevant here.

PR #12710, opened 2020-03-25
Remove dangling rpath and add /opt/vespa/jdk symlink

hakonhall:


hakonhall:
FYI @aressem

dist/vespa.spec
```diff
@@ -330,6 +332,7 @@ fi
 %{_prefix}/etc/systemd
 %{_prefix}/etc/vespa
 %{_prefix}/include
+%{_prefix}/jdk
```
toregge: This should be moved to the files section for vespa-base.

toregge (approved):

hakonhall:
The travis build is green - not sure why the color isn't able to update.

PR #12711, opened 2020-03-25
Add control over mbus params

baldersheim:
- num network threads
 - num connections per target
 - tcpnodelay
Defaults are unchanged.

@vekterlior @havardpe  PR

### DIFF
--- a/container-messagebus/src/main/resources/configdefinitions/container-mbus.def
+++ b/container-messagebus/src/main/resources/configdefinitions/container-mbus.def
@@ -13,7 +13,7 @@ numconnectionspertarget int default=1
 numthreads int default=2
 
 # Optimize for latency, or throughput.
-optimization enum {LATENCY, THROUGHPUT} default=LATENCY
+optimize_for enum {LATENCY, THROUGHPUT} default=LATENCY
 
 # Everying below is deprecated and will go away very soon.
 # Dynamic throttling is used, and works better than anything else.
--- a/container-messagebus/src/main/resources/configdefinitions/container-mbus.def
+++ b/container-messagebus/src/main/resources/configdefinitions/container-mbus.def
@@ -2,9 +2,24 @@
 namespace=container.jdisc
 
 #settings for message bus in container
-enabled bool default=false
+
+# Which network port is used
 port int default=0
+
+# Number of connections per target
+numconnectionspertarget int default=1
+
+# Number network threads
+numthreads int default=2
+
+# Optimize for latency, or throughput.
+optimization enum {LATENCY, THROUGHPUT} default=LATENCY
+
+# Everying below is deprecated and will go away very soon.
+# Dynamic throttling is used, and works better than anything else.
 maxpendingcount int default=2048
+
+enabled bool default=false
 #maxpendingsize is set in megabytes!
 maxpendingsize int default=100
 
--- a/jrt/src/com/yahoo/jrt/Connection.java
+++ b/jrt/src/com/yahoo/jrt/Connection.java
@@ -36,6 +36,7 @@ class Connection extends Target {
     private final Buffer output  = new Buffer(WRITE_SIZE * 2);
     private int maxInputSize  = 64*1024;
     private int maxOutputSize = 64*1024;
+    private final boolean tcpNoDelay;
     private final Map<Integer, ReplyHandler> replyMap = new HashMap<>();
     private final Map<TargetWatcher, TargetWatcher> watchers = new IdentityHashMap<>();
     private int activeReqs = 0;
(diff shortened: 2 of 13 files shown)


jrt/src/com/yahoo/jrt/Connection.java
```diff
@@ -89,21 +90,23 @@ private void setState(int state) {
...
         owner.sessionInit(this);
     }
 
-    public Connection(TransportThread parent, Supervisor owner, Spec spec, Object context) {
+    public Connection(TransportThread parent, Supervisor owner, Spec spec, Object context, boolean tcpNoDelay) {
         super(context);
         this.parent = parent;
         this.owner = owner;
         this.spec = spec;
+        this.tcpNoDelay = tcpNoDelay;
```
vekterli: Do we have any numbers that indicate disabling `TCP_NODELAY` is worth it? We unconditionally enabled nodelay for everything to get rid  of spurious latency spikes incurred by the in-kernel deferred packet sending.

messagebus/src/main/java/com/yahoo/messagebus/network/rpc/RPCNetwork.java
```diff
@@ -82,7 +82,7 @@ private static int getNumThreads() {
     public RPCNetwork(RPCNetworkParams params, SlobrokConfigSubscriber slobrokConfig) {
         this.slobroksConfig = slobrokConfig;
         identity = params.getIdentity();
-        orb = new Supervisor(new Transport(2));
+        orb = new Supervisor(new Transport(params.getNumNetworkThreads(), params.getOptimization() == RPCNetworkParams.Optimization.LATENCY));
```
vekterli: Consider wrapping in a `shouldEnableNoDelay(params)` method or similar to make semantics more obvious

vekterli (approved):
:+1: Overall LGTM but I'm slightly wary of letting the Nagle algorithm back into the playground. Deferring final say to @havardpe 🙂

jrt/src/com/yahoo/jrt/Connection.java
baldersheim: I saw significantly lower system cpu when hardcoding it yesterday.
However I did a lot of different tests then, and I am not putting more into it than it is a dimension that should be tested further. And for that to happen I need config control over it.

messagebus/src/main/java/com/yahoo/messagebus/network/rpc/RPCNetwork.java
baldersheim: Fixed

container-messagebus/src/main/resources/configdefinitions/container-mbus.def
```diff
@@ -2,9 +2,24 @@
...
 port int default=0
+
+# Number of connections per target
+numconnectionspertarget int default=1
+
+# Number network threads
+numthreads int default=2
+
+# Optimize for latency, or throughput.
+optimization enum {LATENCY, THROUGHPUT} default=LATENCY
```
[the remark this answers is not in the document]
baldersheim: So be it, that was one of my options.

container-messagebus/src/main/resources/configdefinitions/container-mbus.def
```diff
@@ -2,9 +2,24 @@
...
 port int default=0
+
+# Number of connections per target
+numconnectionspertarget int default=1
+
+# Number network threads
+numthreads int default=2
+
+# Optimize for latency, or throughput.
+optimize_for enum {LATENCY, THROUGHPUT} default=LATENCY
```
havardpe: this feels a bit generic

If it only controls tcp_nodelay, we can just be more specific. If we add more code alternatives choosing between latency and throughput later we might not be able to test their effects separately, or they need more configuration than a single flag.

storage/src/vespa/storage/config/stor-communicationmanager.def
```diff
@@ -33,6 +33,8 @@ mbus.rpctargetcache.ttl double default = 600
 ## Any value below 1 will be 1.
 mbus.num_threads int default=4
 
+mbus.optimize_for enum {LATENCY, THROUGHPUT} default = LATENCY
```
havardpe: this feels a bit generic

If it only controls tcp_nodelay, we can just be more specific. If we add more code alternatives choosing between latency and throughput later we might not be able to test their effects separately, or they need more configuration than a single flag.

havardpe (approved):

container-messagebus/src/main/resources/configdefinitions/container-mbus.def
baldersheim: I do not want to expose too much details. We should figure out a set that works for all.

storage/src/vespa/storage/config/stor-communicationmanager.def
baldersheim: I do not want to expose too much details. We should figure out a set that works for all.

PR #12712, opened 2020-03-25
Revert "Don't take combined clusters of size 1 down"

hmusum:
Reverts vespa-engine/vespa#12704

Breaks a system test, might be that the test just needs an update...


baldersheim (approved):
👍

PR #12717, opened 2020-03-25
Bratseth/autoscaling ranges 1 .

bratseth:
Just some preparation work, no behavior changes. It should work well to look at the commits in order.


config-provisioning/src/main/java/com/yahoo/config/provision/Capacity.java
```diff
@@ -11,41 +11,56 @@
...
     private final boolean canFail;
 
-    private final Optional<NodeResources> nodeResources;
-
     private final NodeType type;
 
-    private Capacity(int nodeCount, Optional<NodeResources> nodeResources, boolean required, boolean canFail, NodeType type) {
-        this.nodeCount = nodeCount;
+    private Capacity(ClusterResources min, ClusterResources max, boolean required, boolean canFail, NodeType type) {
+        this.min = min;
```
freva: Verify that `min <= max`?

freva (approved):
Great! :+1:

config-provisioning/src/main/java/com/yahoo/config/provision/Capacity.java
bratseth: Ok, done.

PR #12718, opened 2020-03-25
Revert "Revert "Revert "Don't take combined clusters of size 1 down"""

bratseth:
Reverts vespa-engine/vespa#12714

Still failing


hakonhall:
👍

PR #12719, opened 2020-03-26
Use image for child node type if set 

hmusum:
If docker image is set for a node type, use that image when preloading docker image by host-admin for the corresponding host type. Avoids unnecessarily downloading an image that will not be used.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/DockerImages.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/DockerImages.java
@@ -16,7 +16,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Logger;
 
 /**
- * Multithread safe class to get and set docker images for given host types.
+ * Multithread safe class to get and set docker images for given node types.
  *
  * @author freva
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/DockerImages.java
```diff
@@ -57,9 +57,30 @@ private void createCache() {
...
+                    typeToUseForLookup = NodeType.tenant;
+                    break;
+                case proxyhost:
+                    typeToUseForLookup = NodeType.proxy;
+                    break;
+                default:
+                    break; // do not change
+            }
+        }
+
```
freva:
```suggestion
        NodeType typeToUseForLookup = type.isDockerHost() ? type.childNodeType() : type;
```

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/DockerImages.java
hmusum: Much better, thanks

PR #12725, opened 2020-03-26
Bratseth/autoscaling ranges 2

bratseth:
Ranges can be specified but has no effect (we always take the min).

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/NodesSpecification.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/NodesSpecification.java
@@ -377,7 +377,7 @@ public class NodesSpecification {
             }
         }
         catch (IllegalArgumentException e) {
-            throw new IllegalArgumentException("Expected a number or range on the form [min, max], but got '" + s + "'");
+            throw new IllegalArgumentException("Expected a number or range on the form [min, max], but got '" + s + "'", e);
         }
     }
 
--- a/config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/NodesSpecification.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/NodesSpecification.java
@@ -28,11 +28,9 @@ import java.util.Optional;
  */
 public class NodesSpecification {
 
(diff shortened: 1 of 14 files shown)


config-model/src/test/schema-test-files/services-hosted.xml
```diff
@@ -27,4 +27,11 @@
     </nodes>
   </content>
 
+  <content id="ml" version="1.0">
+    <redundancy>2</redundancy>
+    <nodes count="[10-20]" flavor="large" groups="[1-3]">
```
freva:
```suggestion
    <nodes count="[10,20]" flavor="large" groups="[1,3]">
```

config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/NodesSpecification.java
```diff
@@ -344,9 +363,28 @@ private static IllegalArgumentException clusterReferenceNotFoundException(String
...
+            if (s.startsWith("[") && s.endsWith("]")) {
+                String[] numbers = s.substring(1, s.length() - 1).split(",");
+                if (numbers.length != 2) throw new IllegalArgumentException();
+                return new Pair<>(valueParser.apply(numbers[0].trim()), valueParser.apply(numbers[1].trim()));
+            } else {
+                return new Pair<>(valueParser.apply(s), valueParser.apply(s));
+            }
+        }
+        catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Expected a number or range on the form [min, max], but got '" + s + "'");
```
freva: Consider adding the message from `e`, useful when `valueParser` is `parseGbAmount` and users specify invalid unit.

freva (commented):
You also need to update
https://github.com/vespa-engine/vespa/blob/77fa3b877833b8dc5957039ee02b1068cf0b0be6/config-model/src/main/resources/schema/common.rnc#L27
to allow string

config-model/src/test/schema-test-files/services-hosted.xml
bratseth: thanks!

freva (approved):

PR #12741, opened 2020-03-27
move distance-metric specification, part 1

arnej27959:
* move distance-metric in grammar, and change
  wiring within config-model.  It still has the
  same place in generated config and only has
  effect if hnsw index is specified.

@geirst please review


geirst (approved):
👍

PR #12742, opened 2020-03-27
Increase ZooKeeper test server tick time

vekterli:
@geirst please review

Tests running under heavy concurrent CI build load appear
to struggle maintaining ZK sessions, causing flakiness. This
is an attempt to mitigate that by bumping the ZK server's tick
interval (and therefore session timeout) by 20x.

According to local testing, this does not increase testing time
in the common case.


geirst (approved):
👍

PR #12744, opened 2020-03-27
UnixPath support for atomicWrite, getFilename, getParent, setLastModifiedTime

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/UnixPath.java
```diff
@@ -135,6 +159,10 @@ public Instant getLastModifiedTime() {
         return getAttributes().lastModifiedTime();
     }
 
+    public void setLastModifiedTime(Instant instant) {
+        uncheck(() -> Files.setLastModifiedTime(path, FileTime.from(Instant.now())));
```
freva:
```suggestion
        uncheck(() -> Files.setLastModifiedTime(path, FileTime.from(instant)));
```

freva (commented):
I suggest always returning `UnixPath` (`this`) rather than `void`

freva (approved):

PR #12745, opened 2020-03-27
Use xxhash for most out hashing. It is significantly better quality h…

baldersheim:
…ash.

From 0.7.3 is is also significantly faster.

@toregge or @vekterli PR


vekterli (approved):
:+1: once it's out of pre-release stage, we should consider moving to XXH3 since it's even faster. Also when using `XXH_INLINE_ALL` we could perhaps even consider using it for integers, as constant sized input will be compiler-optimized really well.

baldersheim:
XXH_INLINE_ALL is pending some spec file changes coming soon.

PR #12749, opened 2020-03-28
Reduce warmup time with one second to avoid lockstep with tick

hmusum:
Reduce the chance of using an extra tick when there are some milliseconds
left of warmup time

I very often see logs like this:
` Refusing to resume until warm up period ends (in PT0.096547S)`

this PR tries to avoid spending an extra tick (30 seconds) when there are just some ms left before warm up period ends


freva (approved):

hakonhall:
🥇

freva:
On a second thought we should probably increase, this was initially 30 sec, which due to the issue you describe made it effectively 1 min. That was not enough and it was increased to 1 min - effectively 1 min 30 sec. With this, we will essentially go back to pre- 16f7515017f67b9d939a19c33de7a96da391b20b result. 

So this should probably be between 1 min - 1 min 29 sec, if on the higher end, that will actually increase it for CD (tick interval 15 sec). Or use a separate time for CD.

hakonhall:
@baldersheim can you confirm?

baldersheim:
If possible this should only be done in prod, in cd and test this could be 0.

hmusum:
Changes based on discussion above should be handled in https://github.com/vespa-engine/vespa/pull/12752

PR #12752, opened 2020-03-29
Set warmup duration based on zone

hmusum:
No warmup time necessary in cd or test environments. Set
default warmup time to 89 seconds, as the previous value of 60
seconds (which ended up being ~90 seconds) was the wanted one based
on experience with prod applications.

Merge needs to be coordinated with change in internal repo, please review only.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
@@ -4,7 +4,6 @@ package com.yahoo.vespa.hosted.node.admin.nodeagent;
 import com.yahoo.config.provision.DockerImage;
 import com.yahoo.config.provision.Environment;
 import com.yahoo.config.provision.NodeType;
-import com.yahoo.config.provision.SystemName;
 import com.yahoo.config.provision.zone.ZoneApi;
 import com.yahoo.log.LogLevel;
 import com.yahoo.vespa.flags.DoubleFlag;
@@ -103,9 +102,9 @@ public class NodeAgentImpl implements NodeAgent {
     public NodeAgentImpl(NodeAgentContextSupplier contextSupplier, NodeRepository nodeRepository,
                          Orchestrator orchestrator, DockerOperations dockerOperations, StorageMaintainer storageMaintainer,
                          FlagSource flagSource, Optional<CredentialsMaintainer> credentialsMaintainer,
-                         Optional<AclMaintainer> aclMaintainer, Optional<HealthChecker> healthChecker, Clock clock, ZoneApi zone) {
+                         Optional<AclMaintainer> aclMaintainer, Optional<HealthChecker> healthChecker, Clock clock) {
         this(contextSupplier, nodeRepository, orchestrator, dockerOperations, storageMaintainer, flagSource, credentialsMaintainer,
-             aclMaintainer, healthChecker, clock, warmUpDurationForZone(zone));
+             aclMaintainer, healthChecker, clock, DEFAULT_WARM_UP_DURATION);
     }
 
     public NodeAgentImpl(NodeAgentContextSupplier contextSupplier, NodeRepository nodeRepository,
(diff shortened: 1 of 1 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -605,4 +606,10 @@ public void createSymlink(Path symlink, Path target) {
     protected Optional<CredentialsMaintainer> credentialsMaintainer() {
         return credentialsMaintainer;
     }
+
+    private static Duration warmUpDurationForZone(ZoneApi zone) {
+        return zone.getSystemName() == SystemName.cd || zone.getEnvironment().isTest()
```
freva: Any reason this shouldn't apply to `PublicCd`?
```suggestion
        return zone.getSystemName().isCd() || zone.getEnvironment().isTest()
```

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -605,4 +606,10 @@ public void createSymlink(Path symlink, Path target) {
...
+                ? Duration.ZERO
```
freva: Duration of zero still starts the container unconstrained, but then immediately attempts to suspend against orch. to set limits. Negative will start the container with the constraints https://github.com/vespa-engine/vespa/blob/44e2be1ea393b7556ab59e7f77a757d8d25be546/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java#L215
```suggestion
                ? Duration.ofSeconds(-1)
```

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -102,9 +103,9 @@
     public NodeAgentImpl(NodeAgentContextSupplier contextSupplier, NodeRepository nodeRepository,
                          Orchestrator orchestrator, DockerOperations dockerOperations, StorageMaintainer storageMaintainer,
                          FlagSource flagSource, Optional<CredentialsMaintainer> credentialsMaintainer,
-                         Optional<AclMaintainer> aclMaintainer, Optional<HealthChecker> healthChecker, Clock clock) {
+                         Optional<AclMaintainer> aclMaintainer, Optional<HealthChecker> healthChecker, Clock clock, ZoneApi zone) {
         this(contextSupplier, nodeRepository, orchestrator, dockerOperations, storageMaintainer, flagSource, credentialsMaintainer,
-                aclMaintainer, healthChecker, clock, DEFAULT_WARM_UP_DURATION);
+             aclMaintainer, healthChecker, clock, warmUpDurationForZone(zone));
```
freva: Consider just resolving this when needed via the `ZoneApi` in `NodeAgentContext`

hmusum: NodeAgentContextSupplier.nextContext() says `blocks until the next context is ready`and that sounded like something I wouldn't do in a constructor, I think I'll leave this for now at least

freva: No, I meant in the methods that actually use this variable (`startContainer()`, `updateContainerIfNeeded()` and `doConverge()`).

Another option is to pass the initial context to the constructor instead of just the zone, since `NodeAgentFactory` already has it https://github.com/vespa-engine/vespa/blob/44e2be1ea393b7556ab59e7f77a757d8d25be546/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentFactory.java#L9

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
hmusum: Yes, I initially did resolving when needed, let's go back to it. Thanks, done

freva (approved):
:100:

PR #12753, opened 2020-03-29
Handle newer gtest versions where the legacy API is deprecated.

toregge:
@vekterli : please review


vekterli (approved):
:+1: since the macro arguments are the same, consider adding e.g. `VESPA_GTEST_INSTANTIATE_TEST_SUITE` and `VESPA_GTEST_TYPED_TEST_SUITE` in `vespalib/gtest/gtest.h` that expands to the correct macro instead of needing to duplicate the arguments in the test code.

vekterli (approved):
:+1:

PR #12761, opened 2020-03-30
Compare correctly

bratseth:


freva (commented):
Missing a negation here...

PR #12762, opened 2020-03-30
Compare even more correctly

bratseth:


bratseth:
Thanks! My tests never got to this condition  :-/

freva (commented):
:+1:

PR #12768, opened 2020-03-30
Remove unused argument

hmusum:
Unused, so unless this is component is required to be constructed before ConfigServerBootstrap in some way I cannot see it can be removed.


hakonhall (approved):
Thanks, a left-over from my changes.

PR #12772, opened 2020-03-30
Configure initially out of rotation for config servers

bratseth:
How about this?


hmusum (approved):
Looks good to me, thanks!

PR #12777, opened 2020-03-31
Reduce timeout for nodesDown and noNodeDown in cd

hmusum:
When there are issues preventing progress in cd, 30 minutes is too long
to wait for the errors to show up, trying with 20 minutes. Hard to know the right number, but we should rather increase it again if there is any issue.


hmusum:
@jonmv FYI

hakonhall (approved):

PR #12780, opened 2020-03-31
- Add testing of enumerated weighted sets.

baldersheim:
- Add testing that query can use floating point numbers for the weight.
- Stay with the type feature_t for query. Do not assume th ethe weight type used in the attribute.

@havardpe or @geirst PR


havardpe (approved):

geirst (commented):
👍

PR #12784, opened 2020-03-31
Add explicit limits to backend document selection parsing

vekterli:
@geirst please review

Adds the following (very generous) limits:
- Max AST depth of 1024
- Max input selection string size of 1 MiB

Have to track AST depth manually, as there is no exposed way of
doing this natively via Bison.

Also removed a regex that had the potential of catastrophic
backtracking in case of massive inputs. It wasn't removed during
the previous purge due to being used with capture groups, which are
not supported by our current vespalib regex wrapper.

### DIFF
--- a/document/src/vespa/document/select/parser.cpp
+++ b/document/src/vespa/document/select/parser.cpp
@@ -13,7 +13,9 @@ namespace {
 
 void verify_expression_not_too_large(const std::string& expr) {
     if (expr.size() > ParserLimits::MaxSelectionByteSize) {
-        throw ParsingFailedException("expression is too large to be parsed");
+        throw ParsingFailedException(vespalib::make_string(
+                "expression is too large to be parsed (max %zu bytes)",
+                ParserLimits::MaxSelectionByteSize));
     }
 }
 
--- a/document/src/vespa/document/select/parser.cpp
+++ b/document/src/vespa/document/select/parser.cpp
@@ -1,5 +1,6 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #include "parser.h"
+#include "parser_limits.h"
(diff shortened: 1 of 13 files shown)


document/src/vespa/document/select/parser.cpp
```diff
@@ -8,7 +9,18 @@
 
 namespace document::select {
 
+namespace {
+
+void verify_expression_not_too_large(const std::string& expr) {
+    if (expr.size() > ParserLimits::MaxSelectionByteSize) {
+        throw ParsingFailedException("expression is too large to be parsed");
```
geirst, line 16: Consider adding the size and limit to the exception message.

document/src/vespa/document/select/parser_limits.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#include "parser_limits.h"
+#include "parsing_failed_exception.h"
+
+namespace document::select {
+
+void throw_max_depth_exceeded_exception() {
+    throw ParsingFailedException("expression is too deeply nested");
```
geirst, line 8: Consider adding the depth and max depth limit to the exception message.

geirst (approved):
👍  Consider adding some more details to the exception messages.

document/src/vespa/document/select/parser.cpp
vekterli: Done

document/src/vespa/document/select/parser_limits.cpp
vekterli: Added limit to message. We cannot know the actual depth of the input AST as we terminate parsing immediately when hitting the limit.

geirst (approved):
👍

PR #12789, opened 2020-04-01
Improve error handling in VespaDocumentOperation Pig UDF

rafan:
- Fix NoClassDefFoundError on shaded/vespa/commons/lang/exception/ExceptionUtils
  The ExceptionUtils from vespajlib would be a good replacement if it was
  client library (JDK 8 compatible). To minimize dependency, simply grab
  the implementation.
- Emit UDF error logs to stderr so it can be examined
- Use different PigWarning to make triage easier


lesters (approved):
LGTM!

PR #12790, opened 2020-04-01
Revert "Add explicit limits to backend document selection parsing"

aressem:
Reverts vespa-engine/vespa#12784

Broke the build.


aressem:
Should be fixed by #12791

PR #12798, opened 2020-04-01
Default to epoch 0 with YumPackageName::toVersionLockName

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumPackageName.java
```diff
@@ -72,6 +72,15 @@ public Builder(YumPackageName packageName) {
...
+        /**
+         * Set the epoch of the YUM package.
+         *
+         * <p>WARNING: Should only be invoked if the YUM package actually has an epoch. Typically
+         * YUM packages doesn't have one explicitly set, and in case "0" will be used with
+         * {@link #toVersionLockName()} (otherwise it fails), but it will be absent from an
+         * install with {@link #toName()} (otherwise it fails). This typically means that
+         * you should set this only if the epoch is != "0".</p>
+         */
         public Builder setEpoch(String epoch) { this.epoch = Optional.of(epoch); return this; }
```
freva, line 84: Throw if 0?

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumPackageName.java
hakonhall: I'm not sure. I'm guessing the critical thing is whether an epoch has been set or not in the RPM spec?  One could theoretically set epoch 0, in case it might work and could serve to document it?

In any case I would like to avoid touching https://git.ouroath.com/vespa/vespa-yahoo/blob/master/hosted/host-admin/src/main/java/com/yahoo/vespa/hosted/hostadmin/task/network/FirewallTask.java#L71

PR #12799, opened 2020-04-01
Copy the grouping expression string

bratseth:


arnej27959 (approved):
i guess this is preparation for something.

bratseth:
No, just a bug.

PR #12800, opened 2020-04-02
ensure consistent offset in logcontrol file

arnej27959:

This is in preparation for using append instead of seeking to EOF and then doing write.  This change is enough to ensure consistent alignment for all logcontrol lines.  I have tested manually that both roll-forward and roll-back seems to work (luckily the old code was quite robust to small deviations in the file format).

@gjoranv please review and merge
@bjorncs FYI


gjoranv:
Shouldn't you also add 2 in line 142 (extendMapping)? If so, couldn't you just add 2 to the `maxPrefix` and `controlFileHeaderLength` constants?

arnej27959:
> Shouldn't you also add 2 in line 142 (extendMapping)?

no, because we need in-place version upgrades to work - the new code must be able to use the old file format.  So it will start looking for lines in the old place and that will handle both old and new file formats.  We should do cleanup later, but possibly not until next major version bump...

gjoranv (approved):

PR #12806, opened 2020-04-02
Add more fine grained pecentiles add 98, 99.5,99.6, 99.7, 99.8, 99.9

baldersheim:
@havardpe or @geirst PR

### DIFF
--- a/fbench/src/fbench/fbench.cpp
+++ b/fbench/src/fbench/fbench.cpp
@@ -213,7 +213,7 @@ std::string
 fmtPercentile(double percentile) {
     char buf[32];
     if (percentile <= 99.0) {
-        snprintf(buf, sizeof(buf), "%2d", int(percentile));
+        snprintf(buf, sizeof(buf), "%2d  ", int(percentile));
     } else {
         snprintf(buf, sizeof(buf), "%2.1f", percentile);
     }
--- a/fbench/src/fbench/fbench.cpp
+++ b/fbench/src/fbench/fbench.cpp
@@ -227,13 +227,6 @@ FBench::PrintSummary()
     actualRate = (status._realTime > 0) ?
                  realNumClients * 1000.0 * status._requestCnt / status._realTime : 0;
 
-    double p25 = status.GetPercentile(25);
-    double p50 = status.GetPercentile(50);
-    double p75 = status.GetPercentile(75);
-    double p90 = status.GetPercentile(90);
-    double p95 = status.GetPercentile(95);
-    double p99 = status.GetPercentile(99);
-
     if (_keepAlive) {
         printf("*** HTTP keep-alive statistics ***\n");
         printf("connection reuse count -- %" PRIu64 "\n", status._reuseCnt);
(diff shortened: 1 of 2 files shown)


fbench/src/fbench/fbench.cpp
```diff
@@ -250,24 +243,24 @@ FBench::PrintSummary()
...
-        printf("95 percentile:          %8.2f ms (approx)\n", p95);
-    else         printf("95 percentile:          %8.2f ms\n", p95);
-    if (p99 > status._timetable.size() / status._timetableResolution - 1)
-        printf("99 percentile:          %8.2f ms (approx)\n", p99);
-    else         printf("99 percentile:          %8.2f ms\n", p99);
+
+    for (double percentile : {25.0, 50.0, 75.0, 90.0, 98.0, 99.0, 99.5, 99.6, 99.7, 99.8, 99.9}) {
+        if (percentile <= 99.0) {
+            //Backwards compatible printing
+            if (percentile > (status._timetable.size() / status._timetableResolution - 1)) {
```
havardpe, line 250: you need to check the result of the percentile calculation against the sampled data to see if it is approximate, not the input parameter of the percentile calculation...

fbench/src/fbench/fbench.cpp
```diff
@@ -250,24 +243,24 @@ FBench::PrintSummary()
...
+                printf("%2d percentile:          %8.2f ms (approx)\n", int(percentile), status.GetPercentile(percentile));
+            } else {
+                printf("%2d percentile:          %8.2f ms\n", int(percentile), status.GetPercentile(percentile));
+            }
+        } else {
+            if (percentile > (status._timetable.size() / status._timetableResolution - 1)) {
+                printf("%2.1f percentile:          %8.2f ms (approx)\n", percentile, status.GetPercentile(percentile));
```
havardpe, line 257: factoring out the formatting of the percentile number would probably be useful here...

fbench/src/fbench/fbench.cpp
baldersheim: Good catch, fixed

fbench/src/fbench/fbench.cpp
baldersheim: Factored out both optional (approx) and formating of percentile.

baldersheim:
@havardpe Now you have another look.

fbench/src/fbench/fbench.cpp
```diff
@@ -200,6 +200,28 @@ FBench::StopClients()
...
+    return (latency > (status._timetable.size() / status._timetableResolution - 1))
+           ? "ms (approx)"
+           : "ms";
+}
+
+std::string
+fmtPercentile(double percentile) {
+    char buf[32];
+    if (percentile <= 99.0) {
+        snprintf(buf, sizeof(buf), "%2d  ", int(percentile));
```
havardpe, line 216: I think this will add 2 extra spaces in the 'backwards' case...

havardpe (commented):
looks good, but I think the old-style formatting is a bit off.

fbench/src/fbench/fbench.cpp
baldersheim: Yes, that is fine. I just wanted to avid decimals  in the old percentiles

PR #12810, opened 2020-04-02
Add distributor bucket db memory usage metrics

vekterli:
@geirst please review.

One question up for debate is the metric dimension key name of the bucket DB metrics. There is one dimension for the mutable DBs and one for the read-only DBs. I've just gone with "type" as the key here, but we might have something more suitable.


metrics/src/vespa/metrics/common/memory_usage_metrics.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include "memory_usage_metrics.h"
+#include <vespa/vespalib/util/memoryusage.h>
+
+namespace metrics {
+
+MemoryUsageMetrics::MemoryUsageMetrics(metrics::MetricSet* parent)
+    : MetricSet("memory_usage", {}, "The memory usage for a given component", parent),
+      _allocated_bytes("allocated_bytes", {}, "The number of allocated bytes", this),
+      _used_bytes("used_bytes", {}, "The number of used bytes (<= allocatedbytes)", this),
```
geirst, line 11: Consider changing description to match metric name: "<= allocated_bytes"

metrics/src/vespa/metrics/common/memory_usage_metrics.cpp
```diff
@@ -0,0 +1,28 @@
...
+      _dead_bytes("dead_bytes", {}, "The number of dead bytes (<= usedbytes)", this),
```
geirst, line 12: Consider changing description to match metric name: "<= used_bytes"

storage/src/vespa/storage/distributor/distributormetricsset.cpp
```diff
@@ -2,11 +2,19 @@
...
 #include <vespa/metrics/loadmetric.hpp>
 #include <vespa/metrics/summetric.hpp>
+#include <vespa/vespalib/util/memoryusage.h>
 
 namespace storage::distributor {
 
 using metrics::MetricSet;
 
+BucketDbMetrics::BucketDbMetrics(const vespalib::string& db_type, metrics::MetricSet* owner)
+    : metrics::MetricSet("bucket_db", {{"type", db_type}}, "", owner),
```
geirst, line 12: I think "type" could interfere with other dimension names. Consider "bucket_db_type" instead?

geirst (approved):
👍 Some comments to consider, especially the dimension name for bucket db as you already pointed out.

metrics/src/vespa/metrics/common/memory_usage_metrics.cpp
vekterli: Done

storage/src/vespa/storage/distributor/distributormetricsset.cpp
vekterli: Done

metrics/src/vespa/metrics/common/memory_usage_metrics.cpp
vekterli: Done

geirst (approved):
👍

PR #12811, opened 2020-04-02
Revert "Revert "Balder/rearrange threads""

baldersheim:
As servicepool is now thread safe this one should get another go,
@vekterli PR


vekterli (approved):
:+1: presumably `optimize_for` (and `num_threads`?) will be reverted back after testing?

baldersheim:
I do expect fallout from this so we will see what brings next.
I will wait to merge this one until the service pool has been properly verified.

PR #12813, opened 2020-04-02
Bypass communicationmanager Q

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/common/storagelink.cpp
+++ b/storage/src/vespa/storage/common/storagelink.cpp
@@ -123,9 +123,9 @@ void StorageLink::sendDown(const StorageMessage::SP& msg)
         default:
             LOG(error, "Link %s trying to send %s down while in state %s",
                 toString().c_str(), msg->toString().c_str(), stateToString(getState()));
-            assert(false);
+            return;
     }
-    assert(msg.get());
+    assert(msg);
     LOG(spam, "Storage Link %s to handle %s", toString().c_str(), msg->toString().c_str());
     if (isBottom()) {
         LOG(spam, "Storage link %s at bottom of chain got message %s.", toString().c_str(), msg->toString().c_str());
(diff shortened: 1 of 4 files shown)


baldersheim:
Merging to get a nightly stress test burn in.

storage/src/vespa/storage/storageserver/communicationmanager.cpp
```diff
@@ -492,8 +489,8 @@ void
 CommunicationManager::enqueue(std::shared_ptr<api::StorageMessage> msg)
 {
     assert(msg);
-    LOG(spam, "Enq storage message %s, priority %d", msg->toString().c_str(), msg->getPriority());
-    _eventQueue.enqueue(std::move(msg));
+    LOG(spam, "Process storage message %s, priority %d", msg->toString().c_str(), msg->getPriority());
+    process(msg);
```
vekterli, line 493: We should probably rename `CommunicationManager::enqueue` in due time as well 🙂

vekterli (commented):
:+1: one question re: asserts

storage/src/vespa/storage/common/storagelink.cpp
```diff
@@ -165,9 +165,9 @@ void StorageLink::sendUp(const shared_ptr<StorageMessage> & msg)
         default:
             LOG(error, "Link %s trying to send %s up while in state %s",
                 toString().c_str(), msg->toString(true).c_str(), stateToString(getState()));
-            assert(false);
+            return;
```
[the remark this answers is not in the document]
baldersheim: There was one test that tested enque/deque ordering that triggered this. It posted N tasks BEFORE doing open(), and then checking that they were received in same order as posted.
With a queue it would wait in the queue until open was called. That would now trigger the assert as it is open once constructed. So I concluded the assert could just be removed.

storage/src/vespa/storage/storageserver/communicationmanager.cpp
baldersheim: Yes and some other  stuff could go away too.

PR #12821, opened 2020-04-03
Control mbus worker threads and network threads separately.

baldersheim:
@vekterli PR


storage/src/vespa/storage/config/stor-communicationmanager.def
```diff
@@ -31,7 +31,11 @@ mbus.rpctargetcache.ttl double default = 600
 
 ## Number of threads for network.
 ## Any value below 1 will be 1.
-mbus.num_threads int default=1
+mbus.num_network_threads int default=1
+
+## Number ofworkers threads for messagebus.
```
vekterli, line 36: ofworkers -> of worker

vekterli (approved):
:+1:

storage/src/vespa/storage/config/stor-communicationmanager.def
baldersheim: Fixed

PR #12822, opened 2020-04-03
Compute DeploymentStatistics on demand, from fresh data

jonmv:
@mpolden please review.


jonmv:
Controllers on old and new version may disagree about confidence for controller and system versions for a short while. I doubt that's a problem.

mpolden (approved):
Great!

PR #12830, opened 2020-04-04
Create YumTester

freva:
Must be merged together with internal PR

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/systemd/SystemCtlTester.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/systemd/SystemCtlTester.java
@@ -3,8 +3,7 @@ package com.yahoo.vespa.hosted.node.admin.task.util.systemd;
 
 import com.yahoo.vespa.hosted.node.admin.task.util.process.TestTerminal;
 
-import java.util.HashSet;
-import java.util.Set;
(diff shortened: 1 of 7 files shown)


hakonhall (commented):
What about making a method

```
boolean install(TaskContext context, String package, String... enableRepos)
```
and an overloaded method taking a `List<String> packages`.  These two covers all invocations of install, as there is no need to expose YumPackageName methods as only `toName()` is called anyways. Thus we can avoid fluent-style method invocation.

The same can be done to upgrade and remove.  `installFixedVersion` can be kept as-is. This would make all yum commands similar to other non-fluent-like methods, and tests can use e.g. Mockito.

If some tests really have to verify the order between Yum and terminal commands that becomes awkward with Mockito, derive a YumTester from Yum and override all public methods to make a single invocation on TestTerminal, such that each single invocation can be tested for with e.g. terminal.expect("yum installfixed --enablerepo=foo bar-2.3.4 2>&1").

The same can be done with SystemCtl: Make methods e.g. `boolean enable(TaskContext context, String unit)` etc. The benefit of having a SystemCtlTester only invoke a single TestTerminal command is more obvious with SystemCtl as it nearly always actually invokes more than one command.

Make a SystemCtl constructor that enables sudo. Any SystemCtlTester, if we still need one, would not see any such `sudo` in the `terminal.expectCommand()`.

hakonhall (approved):

PR #12831, opened 2020-04-04
Balder/move sequenced task executors to staging vespalib

baldersheim:
@havardpe PR


havardpe (approved):
We might need to replace the optimize_for config abstraction with something that chooses implementation rather than goal and also lets us specify additional parameters.

havardpe:
@geirst should also have a say. Are we moving too much to vespalib/util?

baldersheim:
Better granularity is coming soon.
I would suggest that we move executor related stuff from util till a separate vespalib/execute or similar module. That is a separate step.

geirst:
Moving executor related code to a separate sub-library sounds good.

PR #12840, opened 2020-04-05
Revert "Control mbus worker threads and network threads separately."

baldersheim:
Reverts vespa-engine/vespa#12821


vekterli (commented):
:+1:

PR #12845, opened 2020-04-06
Add verbose property to print docId when set

jingtinghuang:

- add a property to enable printing docid when there is an error parsing the document

### DIFF
--- a/vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
+++ b/vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
@@ -67,7 +67,7 @@ public class VespaDocumentOperation extends EvalFunc<String> {
     private static final String PROPERTY_CREATE_IF_NON_EXISTENT = "create-if-non-existent";
     private static final String PROPERTY_ID_TEMPLATE = "docid";
     private static final String PROPERTY_OPERATION = "operation";
-    private static final String PROPERTY_PRINT_DOCID_WHEN_ERROR = "print-docid-when-error";
+    private static final String PROPERTY_VERBOSE = "verbose";
     private static final String BAG_AS_MAP_FIELDS = "bag-as-map-fields";
     private static final String SIMPLE_ARRAY_FIELDS = "simple-array-fields";
     private static final String SIMPLE_OBJECT_FIELDS = "simple-object-fields";
@@ -81,7 +81,7 @@ public class VespaDocumentOperation extends EvalFunc<String> {
     private static final String PARTIAL_UPDATE_ASSIGN = "assign";
     private static final String PARTIAL_UPDATE_ADD = "add";
     private static final String PARTIAL_UPDATE_REMOVE = "remove";
-    private static boolean printIdOnError;
+    private static boolean printId;
 
     private static Map<String, String> mapPartialOperationMap;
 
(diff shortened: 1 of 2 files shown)


vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
```diff
@@ -157,6 +160,12 @@ public String exec(Tuple tuple) throws IOException {
             if (statusReporter != null) {
                 statusReporter.incrCounter("Vespa Document Operation Counters", "Document operation failed", 1);
             }
+            if (printIdOnError) {
+                Schema inputSchema = getInputSchema();
+                Map<String, Object> fields = TupleTools.tupleMap(inputSchema, tuple);
+                String docId = TupleTools.toString(fields, template);
+                System.out.println("Error occur when processing document with docID: " +docId);
```
lesters, line 167: This should probably not be a system out here. Use the warnlog function. But, isn't this information already in the line below as the entire tuple is output to warnlog? If so, I don't see the point of adding additional output here. If not, rather change the exception out below to explicitly add the docid. No configuration should be necessary for this.

lesters (changes requested):
See comments.

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
jingtinghuang: Make sense. BTW, if we want to always print the docID out (System.out.println() will be at line 151, before the create function is called). Is it acceptable that we just print it out or should we make it a property just like in this PR? Thanks!

jingtinghuang: @lesters

lesters: Well, we definitely shouldn't be stdout'ing the document id for each document by default, as grid feeds usually have quite a large number of documents. If you want such a feature, add a `verbose` or `debug` property, and if this property is set you can output it. Make sure this property is by default off.

jingtinghuang:
Hi @lesters ,

I've added the "verbose" property and add 2 tests to ensure it's printing what I want & not printing anything when not set.

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
```diff
@@ -80,6 +81,7 @@ public static boolean valid(String text) {
     private static final String PARTIAL_UPDATE_ASSIGN = "assign";
     private static final String PARTIAL_UPDATE_ADD = "add";
     private static final String PARTIAL_UPDATE_REMOVE = "remove";
+    private static boolean printId;
```
lesters, line 84: To be consequent with other code here, move this to the section below for other properties that are retrieved, e.g. line 104ish. Also, make this a `private final boolean`, not a `static`. I guess, to be consequent, call this `verbose` so it can be used for other things as well. So, in sum, `private final boolean verbose;`.

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/pig/VespaDocumentOperation.java
```diff
@@ -145,6 +148,9 @@ public String exec(Tuple tuple) throws IOException {
             Schema inputSchema = getInputSchema();
             Map<String, Object> fields = TupleTools.tupleMap(inputSchema, tuple);
             String docId = TupleTools.toString(fields, template);
+            if (printId) {
```
lesters, line 151: `if (verbose) {`

lesters:
Hi @jingtinghuang Sorry for taking my time here, we've had easter here. Just consider my last here comments and we should be good to go.

jingtinghuang:
Hi @lesters ,
Thanks for reviewing. I've fix the lines mentioned in the review but I had aa accident where I had to hard reset to a previous commit and I'm not sure  if this cause a fail in one of the check, as the "Merge Stop Enforcer " is showing a red cross right now.
If yes, how do I fix it?

lesters:
@jingtinghuang The merge stop enforcer is something we have in place to stop merges to the code when we are fixing issues to ensure we have a release candidate. It is only temporary and will probably be removed soon.

lesters (approved):
LGTM!

PR #12847, opened 2020-04-06
Measure failing maintenance deployments 

bratseth:
We should probably graph this as an unusual uptick indicates that something is wrong.


bratseth:
Ping @yngveaasheim

bratseth:
@mpolden then?

mpolden (approved):

PR #12849, opened 2020-04-06
Add histogram for links on level 0 

arnej27959:
* also add stats for current config

@geirst please review


geirst (approved):
👍 Now I think it's time to add a unit test for HnswIndex::get_state(). Information that is less deterministic (like memory usage), we can ignore in the test.

PR #12852, opened 2020-04-06
Call failed handler for io error on non-linux platform.

toregge:
@baldersheim : please review

### DIFF
--- a/fastos/src/vespa/fastos/unix_file.cpp
+++ b/fastos/src/vespa/fastos/unix_file.cpp
@@ -26,6 +26,13 @@ ssize_t
 FastOS_UNIX_File::Read(void *buffer, size_t len)
 {
     ssize_t nRead = read(_filedes, buffer, len);
+    if (nRead < 0 && _failedHandler != nullptr) {
+        int error = errno;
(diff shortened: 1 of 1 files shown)


fastos/src/vespa/fastos/unix_file.cpp
```diff
@@ -58,6 +72,12 @@ void FastOS_UNIX_File::ReadBuf(void *buffer, size_t length,
     ssize_t readResult;
 
     readResult = pread(_filedes, buffer, length, readOffset);
+    if (readResult < 0 && _failedHandler != nullptr) {
+        int error = errno;
+        const char *fileName = GetFileName();
+        _failedHandler("read", fileName, error, readOffset, length, readResult);
```
baldersheim, line 78: should it be "pread" ?

baldersheim (approved):

fastos/src/vespa/fastos/unix_file.cpp
toregge: This reflects how it is done on linux and what the ioerrorhandler unit test expects.
The error reporting does not differentiate between pread and read, nor between pwrite and write.

PR #12853, opened 2020-04-06
Invoke SubRef on RPC request before exiting process

vekterli:
@arnej27959 please review

Is reported as memory leak by AddressSanitizer, but slips
between the cracks when run under Valgrind.


arnej27959 (approved):
LGTM

PR #12856, opened 2020-04-06
Add control for kind_of_watermark and reaction time.

baldersheim:
@havardpe PR


searchcore/src/vespa/searchcore/proton/server/threading_service_config.h
```diff
@@ -17,21 +18,25 @@ class ThreadingServiceConfig {
...
-    uint32_t    _indexingThreads;
-    uint32_t    _defaultTaskLimit;
-    uint32_t    _semiUnboundTaskLimit;
-    OptimizeFor _optimize;
+    uint32_t           _indexingThreads;
+    uint32_t           _defaultTaskLimit;
+    uint32_t           _semiUnboundTaskLimit;
+    OptimizeFor        _optimize;
+    uint32_t           _kindOfWatermark;
+    vespalib::duration _reactionTime;         // Minimum reaction time to new tasks
```
havardpe, line 26: ... or maximum?

havardpe (approved):
I am a bit worried that this is getting complicated and entangled... Ideally I would prefer separate and specific config settings related to each implementation choice (separate configs might be too heavy, maybe something similar to std::variant as a new config structured type...)

baldersheim:
Agree, but not so worried as I expect these to only have a short lifespan, and will be removed in Q2 after experimentation has converged

searchcore/src/vespa/searchcore/proton/server/threading_service_config.h
baldersheim: Yes, min and max are hard....

PR #12859, opened 2020-04-07
Avoid taking run lock and then application lock, to avoid deadlock

jonmv:


mpolden (approved):
Test failure in `jdisc_http_service` is unrelated.

PR #12860, opened 2020-04-07
Ensure RPC target is freed if exception thrown in ctor 

vekterli:
@geirst please review

`RpcForwarder::ping_logserver()` throws if the ping request fails.
Since this happens at construction time the `RpcForwarder` destructor
is not invoked, and therefore `SubRef()` won't be called on the
target. Replace with field-level RAII guard instead


baldersheim (approved):

geirst (approved):
👍

PR #12864, opened 2020-04-07
Balder/let feeder control throttlepolicy

baldersheim:
@vekterli PR

This should hopefully reduce variation on feed performance tests, by not being so eager on the backoff. No change in defaults for anything but vespa-feed-perf.

### DIFF
--- a/vespa_feed_perf/src/test/java/com/yahoo/vespa/feed/perf/FeederParamsTest.java
+++ b/vespa_feed_perf/src/test/java/com/yahoo/vespa/feed/perf/FeederParamsTest.java
@@ -101,6 +101,32 @@ public class FeederParamsTest {
     }
 
     @Test
+    public void requireThatWindowSizeIncrementIsParsed() throws ParseException, FileNotFoundException {
+        assertEquals(20, new FeederParams().getWindowIncrementSize());
+        assertEquals(17, new FeederParams().parseArgs("--window_incrementsize", "17").getWindowIncrementSize());
+    }
+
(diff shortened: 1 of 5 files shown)


vespa_feed_perf/src/test/java/com/yahoo/vespa/feed/perf/FeederParamsTest.java
```diff
@@ -100,6 +100,32 @@ public void requireThatNumMessagesToSendAreParsed() throws ParseException, FileN
         assertEquals(19, new FeederParams().parseArgs("--nummessages", "19").getNumMessagesToSend());
     }
 
+    @Test
+    public void requireThatWindowSizeIncrementIsParsed() throws ParseException, FileNotFoundException {
+        assertEquals(20, new FeederParams().getWindowIncrementSize());
+        assertEquals(17, new FeederParams().parseArgs("--window_incrementsize", "17").getWindowIncrementSize());
+    }
+
+    static final double SMALL_NUMBER = 0.000000000001;
```
vekterli, line 109: "epsilon" might be a more familiar term for this when used with floating point comparisons

vekterli (approved):
:+1:

vespa_feed_perf/src/test/java/com/yahoo/vespa/feed/perf/FeederParamsTest.java
baldersheim: Done

PR #12867, opened 2020-04-07
Only update bucket DB memory statistics at certain intervals

vekterli:
@geirst please review
@baldersheim FYI

B-tree/datastore stats can be expensive to sample, so don't do
this after every full DB iteration. For now, wait at least 30s.


geirst (approved):
👍

PR #12868, opened 2020-04-07
Tests must be after last prod job, with any status

jonmv:
@mpolden please review.


jonmv:
Corner-case where an application had cancelled prod job after deployment, and cancelled upgrade, but still had an application change rolling out (I don't really know how, though ...!) — the test job runs on versions that are actually deployed, but when test job was successful, it was still filtered out, as it couldn't find any successes on the versions it had run on. 

No point in considering only the successful production deployment jobs — the last completed is the one we want anyway, as it can only be later than the last successful one, but new production jobs shouldn't normally run again without changes, and new changes would require a successful production deployment before the tests would be eligible ot run again.

PR #12869, opened 2020-04-07
Provide all (relevant runs of all) jobs in deployment API

jonmv:
@freva please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
@@ -33,6 +33,7 @@ import java.util.stream.Stream;
 
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toUnmodifiableMap;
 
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
```diff
@@ -129,25 +124,73 @@ private HttpResponse root(HttpRequest request) {
...
+                      instanceObject.setString("instance", instance.instance().value());
+                      instanceObject.setLong("productionJobCount", jobsByInstance.get(instance).production().size());
+                      instanceObject.setString("upgradePolicy", toString(deploymentStatuses.matching(status -> status.application().id().equals(TenantAndApplicationId.from(instance)))
+                                                                                           .first().map(status -> status.application().deploymentSpec())
+                                                                                           .flatMap(spec -> spec.instance(instance.instance()).map(DeploymentInstanceSpec::upgradePolicy))
+                                                                                           .orElse(DeploymentSpec.UpgradePolicy.defaultPolicy)));
+                      Cursor allJobsObject = instanceObject.setObject("allJobs");
+                      Cursor upgradeJobsObject = instanceObject.setObject("upgradeJobs");
+                      for (JobType type : JobType.allIn(controller.system())) {
```
freva, line 171: Would be nice if we could just list relevant jobs, and maybe list all the jobs once in a list for order, if we need that.

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
jonmv: I guess that's an option that saves some JSON over the wire :)

jonmv:
PTAL.

freva (approved):
:100:

PR #12876, opened 2020-04-08
remove latest vespa.log if it is filling the disk too quickly.

arnej27959:

@baldersheim please review

tested manually by running a system test and then turning on spam logging.  not sure if the limits here are reasonable (5 minutes to fill a log file, and filled to 110% of configured rotate limit).

### DIFF
--- a/logd/src/logd/watcher.cpp
+++ b/logd/src/logd/watcher.cpp
@@ -117,6 +117,7 @@ Watcher::watchfile()
 {
     struct donecache already;
     char newfn[FILENAME_MAX];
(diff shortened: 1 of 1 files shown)


baldersheim (changes requested):
I am not sure if this is a good solution. It happens quite often that  logfile  can pass its size during bursts.
I think this is too aggressive. Where does this requirement come from ?
I suggest we discuss tomorrow.

arnej27959:
@baldersheim PTAL

baldersheim (approved):
Now it looks quite a bit safer

PR #12877, opened 2020-04-08
Hold config lock when modifying load balancers

mpolden:
@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
@@ -5,7 +5,6 @@ import com.google.inject.Inject;
 import com.yahoo.collections.ListMap;
 import com.yahoo.component.AbstractComponent;
 import com.yahoo.component.Version;
-import com.yahoo.vespa.hosted.provision.Node.State;
 import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.DockerImage;
 import com.yahoo.config.provision.Flavor;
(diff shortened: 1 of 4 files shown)


bratseth (approved):
LGTM but I don't see why we need node locks as you mention in the NodeRepository TODO: All changes to nodes are locked by the application lock for allocated nodes and the unprovisioned lock for unallocated nodes.

mpolden:
> I don't see why we need node locks as you mention in the NodeRepository TODO.

Yes, but as far as I can tell the appropriate lock isn't held at the point in time when the transaction is committed. It's only held while appending operations to the transaction. Example: Nodes can be changed at any point after the lock is released here https://github.com/vespa-engine/vespa/blob/master/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java#L44-L47 and before the commit in https://github.com/vespa-engine/vespa/blob/master/configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java#L134-L139.

 It's the same problem as with load balancers.

bratseth:
Yes, very good point, but I don't see that we need more fine-grained locks, we just need to share the same application lock on both sides, which is what you achieve here for the load balancer case. Let's just call that the application lock and use it everywhere?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -761,6 +762,7 @@ else if (node.flavor().getType() == Flavor.Type.DOCKER_CONTAINER) { // removing
     public Zone zone() { return zone; }
 
     /** Create a lock which provides exclusive rights to making changes to the given application */
+    // TODO(mpolden): Make this delegate to CuratorDatabaseClient#lockConfig instead
```
mpolden, line 765: > Let's just call that the application lock and use it everywhere?

Yes, that's this TODO, but baby steps... :)

PR #12882, opened 2020-04-08
Validate deserialization of flag data

hakonhall:


controller-api/src/test/resources/system-flags/flags/flag-with-empty-data/main.controller.json
```diff
@@ -1,4 +1,3 @@
 {
-  "id" : "flag-with-empty-data",
-  "comment": "empty data using only id field"
```
freva, line 3: How about we just ignore all key-pairs if key == "comment"?

freva (approved):

vespajlib/src/main/java/com/yahoo/text/JSON.java
```diff
@@ -1,15 +1,23 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.util.Map;
+import java.util.Objects;
+
+import static com.yahoo.yolean.Exceptions.uncheck;
 
 /**
- * Static methods for working with the map textual format which is parsed by {@link MapParser}
```
bjorncs, line 7: "The text format which name should not be spoken" 🤔

vespajlib/pom.xml
```diff
@@ -38,6 +38,11 @@
       <artifactId>guava</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
bjorncs, line 41: Note: Several modules uses `vespajlib` in compile scope - some bundles may require additional Jackson packages.

bjorncs (approved):

freva (approved):

vespajlib/src/main/java/com/yahoo/text/JSON.java
hakonhall: what do you mean? the name is "map textual format", but this class is now more than just utils for that.

vespajlib/pom.xml
hakonhall: Good point, I have removed the dependency and use Slime instead.

I have discussed the Slime changes with @havardpe .

vespajlib/src/main/java/com/yahoo/text/JSON.java
bjorncs: Would be simpler with "Static methods for working with JSON" ;)

bjorncs (approved):

PR #12902, opened 2020-04-14
Bratseth/faster qp compiling

bratseth:


arnej27959 (approved):
i don't know this code well enough to review properly - we need to hire someone who can work on this with you...

PR #12904, opened 2020-04-14
Hold config lock when modifying load balancers

mpolden:
Moves the lock map to `Curator` to allow re-entrant lock across modules. First
commit is the same as #12877.

@bratseth or @hmusum


bratseth (approved):
LGTM!

PR #12911, opened 2020-04-14
Ensure that documentid is legal.

baldersheim:
@vekterli PR

### DIFF
--- a/document/src/main/java/com/yahoo/document/idstring/IdString.java
+++ b/document/src/main/java/com/yahoo/document/idstring/IdString.java
@@ -42,6 +42,7 @@ public abstract class IdString {
     private final String namespace;
     private final String namespaceSpecific;
     private Utf8String cache;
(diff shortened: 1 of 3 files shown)


bratseth (approved):

document/src/main/java/com/yahoo/document/idstring/IdString.java
```diff
@@ -42,6 +42,7 @@ public String getGroup() {
     private final String namespace;
     private final String namespaceSpecific;
     private Utf8String cache;
+    static final int MAX_LENGTH_EXCEPT_NAMESPACE_SPECIFIC = 0xfffe;
```
vekterli, line 45: `0xffff - 1` due to actual offset being the next char after `:`, I presume? A tiny comment would be nice for future reference 🙂

vekterli (commented):
:+1:

PR #12915, opened 2020-04-15
Upgrade jackson-databind to 2.8.11.6

bjorncs:
FYI @gjoranv


kkraune (approved):
I guess there is a reason for not doing 2.10 ?

gjoranv (approved):

bjorncs:
Jersey must be upgraded before we can upgrade Jackson further.

PR #12918, opened 2020-04-15
Introduce top-k-probability and use it to fetch correct proper amount…

baldersheim:
… of hits from each partition

@bratseth PR
@jobergum FYI

### DIFF
--- a/configdefinitions/src/vespa/dispatch.def
+++ b/configdefinitions/src/vespa/dispatch.def
@@ -23,11 +23,13 @@ distributionPolicy enum { ROUNDROBIN, ADAPTIVE } default=ROUNDROBIN
 ## don't use it if you don't (really) mean it.
 maxHitsPerNode int default=2147483647
 
-## Probability for getting the correct topK documents.
-## A value of 1.0 will ask all partitions for topK documents.
-## Any value between <0, 1> will use a Student T fith 30 degrees freedom and compute a K value that
-## will give you the topK documents according to this formulae.
+## Probability for getting the K best hits (topK).
+## A value of 1.0 will ask all N partitions for K hits.
+## Any value between <0, 1> will use a Student T with 30 degrees freedom and compute a value Q that
+## will give you the globally K best hits according to this formula with the desired probability.
 ## q = k/n + qT (p',30) x √(k × (1/n) × (1 − 1/n))
+## With a probability of 0.999 and K=200 and N=10 will give a Q of 38, meaning that you only need to fetch 19% compared to
+## default setting of 1.0. This is a significant optimisation with with very little loss in presicion.
 topKProbability double default=1.0
 
 # Is multi-level dispatch configured for this cluster
--- a/configdefinitions/src/vespa/dispatch.def
+++ b/configdefinitions/src/vespa/dispatch.def
@@ -23,6 +23,13 @@ distributionPolicy enum { ROUNDROBIN, ADAPTIVE } default=ROUNDROBIN
 ## don't use it if you don't (really) mean it.
 maxHitsPerNode int default=2147483647
 
+## Probability for getting the correct topK documents.
+## A value of 1.0 will ask all partitions for topK documents.
+## Any value between <0, 1> will use a Student T fith 30 degrees freedom and compute a K value that
+## will give you the topK documents according to this formulae.
+## q = k/n + qT (p',30) x √(k × (1/n) × (1 − 1/n))
+topKProbability double default=1.0
+
 # Is multi-level dispatch configured for this cluster
 # Deprecated, will go away soon, NOOP
 useMultilevelDispatch bool default=false
--- a/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
@@ -11,6 +11,7 @@ import com.yahoo.prelude.Pong;
 import com.yahoo.search.cluster.ClusterMonitor;
 import com.yahoo.search.cluster.NodeManager;
 import com.yahoo.vespa.config.search.DispatchConfig;
+import org.apache.commons.math3.distribution.TDistribution;
 
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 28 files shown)


configdefinitions/src/vespa/dispatch.def
```diff
@@ -23,6 +23,13 @@ distributionPolicy enum { ROUNDROBIN, ADAPTIVE } default=ROUNDROBIN
 ## don't use it if you don't (really) mean it.
 maxHitsPerNode int default=2147483647
 
+## Probability for getting the correct topK documents.
+## A value of 1.0 will ask all partitions for topK documents.
+## Any value between <0, 1> will use a Student T fith 30 degrees freedom and compute a K value that
+## will give you the topK documents according to this formulae.
```
bratseth, line 29: Not: It's just a single formula

configdefinitions/src/vespa/dispatch.def
```diff
@@ -23,6 +23,13 @@ distributionPolicy enum { ROUNDROBIN, ADAPTIVE } default=ROUNDROBIN
 ## don't use it if you don't (really) mean it.
 maxHitsPerNode int default=2147483647
 
+## Probability for getting the correct topK documents.
+## A value of 1.0 will ask all partitions for topK documents.
```
bratseth, line 27: Here perhaps we should say "for the number of documents specified by the hits parameter".
In general it's a bit strange that "hits" is suddenly called "topK" just because we optimize, but otoh it makes it easier to search for this :-)

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -76,6 +96,9 @@ public SearchCluster(String clusterId, DispatchConfig dispatchConfig, int contai
         for (Node node : nodes)
             nodesByHostBuilder.put(node.hostname(), node);
         this.nodesByHost = nodesByHostBuilder.build();
+        hitEstimator = ((0.0 < dispatchConfig.topKProbability()) && (dispatchConfig.topKProbability() < 1.0))
```
bratseth, line 99: Would be better to move this into the estimator I think. (And then also move it to the top level as a package-visible class).

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -240,6 +263,12 @@ private void setInRotationOnlyIf(boolean inRotation) {
             vipStatus.removeFromRotation(clusterId);
     }
 
+    public int estimateHitsToFetch(int wantedHits, int numPartitions) {
+        return ((hitEstimator == null) || (numPartitions <= 1))
```
bratseth, line 267: Same, why not move this into the estimator.

bratseth: Might be quite useful to control this by query? How much you care can depend on use case ...

bratseth (approved):
LGTM! Consider my nits.

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
baldersheim: Agree, done.

configdefinitions/src/vespa/dispatch.def
baldersheim: Fixed

configdefinitions/src/vespa/dispatch.def
baldersheim: Improved

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
baldersheim: Moved.
Will add query too.

bratseth (approved):

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
baldersheim: Now I have added query side control. dispatch.top-k-probability.
Can '-' be used in query without requiring any escaping ?
I used that to have it identical with services.xml.

bratseth:
The convention for query parameters is camelCase: https://docs.vespa.ai/documentation/reference/search-api-reference.html

baldersheim:
@gjoranv Could you verify the changes to the 3 pom.xml files if they are OK.
I am pulling in or.apache.commons.math3.

baldersheim:
@gjoranv I think I got it now. I had to hide common-math3 in container-dev.
A bit complicated, when you have not done it before.

baldersheim:
@hmusum or @gjoranv What is missing or wrong here ?
[ERROR] container_and_referenced_content(com.yahoo.application.ApplicationTest)  Time elapsed: 0.511 s  <<< ERROR!
com.yahoo.container.di.componentgraph.core.ComponentNode$ComponentConstructorException: Error constructing 'dispatcher.foo' of type 'com.yahoo.search.dispatch.Dispatcher': null
Caused by: java.lang.NoClassDefFoundError: org/apache/commons/math3/distribution/TDistribution
	at com.yahoo.search.dispatch.TopKEstimator.<init>(TopKEstimator.java:20)
	at com.yahoo.search.dispatch.searchcluster.SearchCluster.<init>(SearchCluster.java:81)
Caused by: java.lang.ClassNotFoundException: org.apache.commons.math3.distribution.TDistribution
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)

gjoranv:
It seems to me that the `commons-math3` artifact is only used internally in container-search. It is not leaked from any apis, and should not be visible when building user projects, or from user bundles (or other vespa bundles) at runtime. 

If the above assumptions are correct:
* Don't export the package, i.e. remove the package-info.java file
* Don't add it to the enforcer pom. (It's only for artifacts that are exposed from jdisc runtime and hence should be visible in provided scope for user projects.)
* Exclude it from container-search in `container-dev/pom.xml`
* To make it available for 'application' based unit tests, add it in compile scope to `application/pom.xml` below the comment saying:

>  

* Also, please add a version property in `parent/pom.xml` to make it easier to detect new versions of commons-math3.

baldersheim:
Ah, thanks @gjoranv. It was the application/pom.xml that I did not get.

PR #12928, opened 2020-04-15
Bratseth/vespa http client improvements

bratseth:
@vekterli please review. The last commit is the one to pay attention to, the earlier ones are just cleanup.

We're observing uneven load on containers during feeding on some applications, and this is the only reason for it that I can see. Once we let connections pass through the routing layer I guess this becomes more likely.

### DIFF
--- a/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
+++ b/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
@@ -17,7 +17,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.Random;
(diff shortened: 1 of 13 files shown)


vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
```diff
@@ -17,7 +17,9 @@
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
+import java.util.Random;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ThreadLocalRandom;
```
vekterli, line 22: This import does not seem to have been used

vekterli (approved):
:+1: a couple of nitpicks for consideration

bratseth:
Thanks!

PR #12930, opened 2020-04-15
Allow expressions as arguments 

lesters:
@bratseth Please take a look. This is not suggested for merge, it is for discussion whether or not this approach is viable until we have proper functions in the backend. Generates ranking features in `ExpressionFunction.expand`.Some disruption with the code for shadowing built-in functions that needs to be looked at. Passes all unit tests at least.


bratseth:
This looks right to me. Could well be subtleties here I'm forgetting to think about though, especially around the order in which things are done.

PR #12932, opened 2020-04-15
Revert "Take same lock as config server when modifying nodes"

hmusum:
Reverts vespa-engine/vespa#12919

Config servers in several cd zones fail to start after upgrade with:
` com.yahoo.container.di.componentgraph.core.ComponentNode$ComponentConstructorException: Error constructing 'NodeRepository' of type 'com.yahoo.vespa.hosted.provision.NodeRepository': null
        Caused by: com.yahoo.config.provision.ApplicationLockException: com.google.common.util.concurrent.UncheckedTimeoutException: Timed out after waiting PT2M to acquire lock '/provision/v1/locks/hosted-vespa/routing/default'`


baldersheim (approved):

aressem:
👍

PR #12934, opened 2020-04-15
Unset IP addresses on deprovisioning 

bjormel:


bjormel:
@mpolden I hope that this is sufficient

mpolden (approved):
LGTM! What is the reason we need to do this?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -629,8 +629,10 @@ public Node markNodeAvailableForNewAllocation(String hostname, Agent agent, Stri
                 List<Node> removed = new ArrayList<>(children);
                 if (zone.cloud().value().equals("aws"))
                     db.removeNodes(List.of(node));
-                else
+                else {
+                    node = node.with(IP.Config.EMPTY);
```
mpolden, line 633: Consider extending one of the deprovision tests in `NodeRepositoryTest` to test this behaviour.

mpolden:
.

PR #12941, opened 2020-04-16
Add metrics for enqueued and processed coredumps 

olaaun:


node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandlerTest.java
```diff
@@ -206,13 +209,28 @@ public void process_single_coredump_test() throws IOException {
...
+        assertEquals(1, updatedMetrics.size());
+        Map<String, Number> values = updatedMetrics.get(0).getMetrics();
+        assertEquals(2, values.get("coredumps.enqueued").intValue());
+        assertEquals(1, values.get("coredumps.processed").intValue());
     }
 
     @Before
     public void setup() throws IOException {
-        Files.createDirectories(donePath);
+        Files.createDirectories(donePath.resolve("container-123"));
```
freva, line 232: I'd recommend using `UnixPath` in the test methods to easily create parent directories and/or files you need, the before method should only create directories are guaranteed to always exist, otherwise the tests might not catch that a certain directory is not created as expected.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -194,4 +204,47 @@ Path findCoredumpFileInProcessingDirectory(Path coredumpProccessingDirectory) {
                 .orElseThrow(() -> new IllegalStateException(
                         "No coredump file found in processing directory " + coredumpProccessingDirectory));
     }
+
+    void updateMetrics(NodeAgentContext context, Path containerCrashPathOnHost) {
+        Dimensions dimensions = generateDimensions(context);
+
+        // Unprocessed coredumps
+        int numberOfUnprocessedCoredumps = FileFinder.files(containerCrashPathOnHost)
```
freva, line 212: The `processing` directory may contain multiple files, at least the core file and `metadata.json`. You probably want to limit max-depth here, but then you'll be off by 1 if there is a coredump being processed right now...

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -194,4 +204,47 @@ Path findCoredumpFileInProcessingDirectory(Path coredumpProccessingDirectory) {
...
+                .match(nameStartsWith(".").negate())
+                .list().size();
+
+        metrics.declareGauge(Metrics.APPLICATION_NODE, "coredumps.enqueued", dimensions, Metrics.DimensionType.PRETAGGED).sample(numberOfUnprocessedCoredumps);
+
+        // Processed coredumps
+        Path processedCoredumpsPath = doneCoredumpsPath.resolve(context.containerName().asString());
+        int numberOfProcessedCoredumps = FileFinder.files(processedCoredumpsPath)
```
freva, line 220: Same problem here, should be `.directories()` instead of `.files()` and limit max-depth to 1

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
olaaun: This is for number of coredumps in crash path, not the `processing` path. So I'll need to filter out hs_err files as well

freva: `processing` is inside the crash path
https://github.com/vespa-engine/vespa/blob/8bebe88526e449bbca0514d8bdfc87390cb26db4/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java#L87

olaaun: Aha 👍

freva (commented):
Looks good, only issue is processing core dump(s) fails (f.ex. full disk), the metric will not be updated. Maybe move updating the metric up, or catch and re-throw exception?

freva (approved):

PR #12944, opened 2020-04-16
Add flag control over softstart seconds

baldersheim:
@hmusum or @hakonhall PR Is this enough to get host and applicationlevel control of the default-soft-start-seconds ?
@yngveaasheim FYI

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
@@ -149,6 +149,7 @@ public class ModelContextImpl implements ModelContext {
         private final double defaultTopKprobability;
         private final Optional<EndpointCertificateSecrets> endpointCertificateSecrets;
         private final double defaultTermwiseLimit;
+        private final double defaultSoftStartSeconds;
         private final boolean useBucketSpaceMetric;
         private final String proxyProtocol;
         private final Optional<AthenzDomain> athenzDomain;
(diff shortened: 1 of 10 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -183,6 +184,8 @@ public Properties(ApplicationId applicationId,
             this.endpointCertificateSecrets = endpointCertificateSecrets;
             defaultTermwiseLimit = Flags.DEFAULT_TERM_WISE_LIMIT.bindTo(flagSource)
                     .with(FetchVector.Dimension.APPLICATION_ID, applicationId.serializedForm()).value();
+            defaultSoftStartSeconds = Flags.DEFAULT_SOFT_START_SECONDS.bindTo(flagSource)
+                    .with(FetchVector.Dimension.APPLICATION_ID, applicationId.serializedForm()).value();
```
hakonhall, line 188: Since only the application ID dimension is set here, the value can only depend on the application ID, not hostname.

baldersheim: So how can I control this with hostname too ?
I see that there is a Dimension.HOSTNAME, but I do not have the hostname accessible here.
Any example ?

hakonhall: All current examples of HOSTNAME are trivial: the hostname would be given and you would invoke `with(HOSTNAME, hostname)` here like you did for APPLICATION_ID.

However this will not work here, as all we have are data structures that will apply during the deployment of a given application, irrespective of the hosts involved.

So could we wait with resolving the flag until closer to the place in the config model where we know the involved host?  No, because changing the flag wouldn't bump the config version and wouldn't trigger a new fetch, and possibly leaving half of the nodes with one view of the flags and the other half with another view.

One possibility is to freeze the flag values (fork the flag source), and pass that through to the config model. Another faster time-to-market would be to have a special value (say a String) which is specially handled in the config model (split on ";" in to a set of hostnames).

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
baldersheim: OK, then the host dimension needs another solution.

PR #12945, opened 2020-04-16
isWorking is tri-state MERGEOK

bratseth:


container-search/src/test/java/com/yahoo/search/dispatch/MockSearchCluster.java
```diff
@@ -40,7 +40,6 @@ public MockSearchCluster(String clusterId, DispatchConfig dispatchConfig, int gr
             List<Node> nodes = new ArrayList<>();
             for (int node = 0; node < nodesPerGroup; node++) {
                 Node n = new Node(distributionKey, "host" + distributionKey, group);
-                n.setWorking(true);
                 nodes.add(n);
```
baldersheim, line 43: Was this one intentional, or was it just temporary testing ?

bratseth: Intentional, as it reproduces the true initial state (which would have surfaced this bug).

PR #12950, opened 2020-04-16
Remove redundant bucket DB lookup in persistence reply handling.

vekterli:
@geirst please review

Bucket DB updating happened unconditionally anyway; this was
only used for failing operations in an overly pessimistic way.

Removing this lookup has two benefits:
- Less CPU spent in (and around) DB
- Less impact expected during feeding during node state
  transitions since fewer operations will have to be needlessly
  retried by the client.

Rationale: an operation towards a given bucket completes (i.e.
is ACKed by all its replica nodes) at time _t_ and the bucket is
removed from the DB at time _T_. There is no fundamental change
in correctness or behavior from the client's perspective if
the order of events is _tT_ or _Tt_. Both are equally valid, as
the state transition edge happens independently of any reply
processing.

### DIFF
--- a/storage/src/tests/distributor/putoperationtest.cpp
+++ b/storage/src/tests/distributor/putoperationtest.cpp
@@ -182,7 +182,7 @@ TEST_F(PutOperationTest, do_not_send_inline_split_if_not_configured) {
               _sender.getCommands(true, true));
 }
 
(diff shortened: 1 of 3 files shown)


baldersheim (approved):

geirst (approved):
👍

PR #12951, opened 2020-04-16
Don't resize retired nodes

bratseth:


bratseth:
@freva could you take a look again? I think this is complete now.

freva (approved):
Great!

PR #12954, opened 2020-04-16
Gjoranv/bundle loader unit test

gjoranv:


gjoranv:
Replaced by #12955 due to wrong order of commits. Force pushing did not help.

PR #12955, opened 2020-04-16
Gjoranv/bundle loader refactor and unit test

gjoranv:
Replaces #12954 due to wrong order of commits in that PR.

The order of commits changed after the PR was created here as well. The first commit ('Add unit test') should have been number 3 (between 'Allow setting ...' and 'Simplify code flow') but the end result is still the same.


container-core/src/test/java/com/yahoo/container/core/config/BundleLoaderTest.java
```diff
@@ -0,0 +1,90 @@
+package com.yahoo.container.core.config;
```
bjorncs: Copyright header?

bjorncs (approved):
Nice improvements to code readability 👍

PR #12956, opened 2020-04-17
Stop using deprecated constructor

hmusum:
Found that this code uses a XmlPreProcessor constructor that we should not use anymore


jonmv (approved):
:)

PR #12962, opened 2020-04-17
Since throttling now is on the outside, and has been for 12 years

baldersheim:
we can remove this hack.

@hmusum or @lesters PR


hmusum (approved):
LGTM!

PR #12970, opened 2020-04-17
Temporarily inhibit maintenance scans when node has heavy message load

vekterli:
@geirst please review.

If requests or responses from external sources are being constantly
processed as part of the distributor ticking, allow for up to _N_ ticks
to skip maintenance scanning, where _N_ is a configurable number.
This reduces the amount of CPU time spent on maintenance operations
when the node has a lot of incoming data to deal with.

Haven't seen much of a difference when testing locally, but I suspect that
my laptop is too scrawny to make the distributor the bottleneck when testing
end-to-end. Would like to get the factory performance tests' opinion on this.

### DIFF
--- a/storage/src/vespa/storage/config/stor-distributormanager.def
+++ b/storage/src/vespa/storage/config/stor-distributormanager.def
@@ -237,3 +237,10 @@ use_weak_internal_read_consistency_for_client_gets bool default=false
 ## feature, so it's not required to set that config to true, nor will setting it
 ## to false actually disable the feature.
 enable_metadata_only_fetch_phase_for_inconsistent_updates bool default=false
+
+## If a distributor main thread tick is constantly processing requests or responses
+## originating from other nodes, setting this value above zero will prevent implicit
+## maintenance scans from being done as part of the tick for up to N rounds of ticking.
+## This is to reduce the amount of CPU spent on ideal state calculations and bucket DB
+## accesses when the distributor is heavily loaded with feed operations.
+max_consecutively_inhibited_maintenance_ticks int default=20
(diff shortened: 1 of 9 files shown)


storage/src/vespa/storage/config/stor-distributormanager.def
```diff
@@ -236,4 +236,11 @@ use_weak_internal_read_consistency_for_client_gets bool default=false
...
-enable_metadata_only_fetch_phase_for_inconsistent_updates bool default=false
\ No newline at end of file
+enable_metadata_only_fetch_phase_for_inconsistent_updates bool default=false
+
+## If a distributor main thread tick is constantly processing requests or responses
+## originating from other nodes, setting this value above zero will prevent implicit
+## maintenance scans from being done as part of the tick for up to N rounds of ticking.
+## This is to reduce the amount of CPU spent on ideal state calculations and bucket DB
+## accesses when the distributor is heavily loaded with feed operations.
+max_consecutively_inhibited_maintenance_ticks int default=20
```
baldersheim, line 246: What is the strange sign at the end of the line ?

baldersheim:
The assumptions are correct, on the branch I have modifying the thread model. I manually just added 1 in 4 if more to do, and that improved performance significantly. This will make it even better.

storage/src/vespa/storage/config/stor-distributormanager.def
vekterli: That would be me forgetting to add a newline 😬

vekterli:
Travis build was caught in the midst of our GCC 9 transition
```
Status: Downloaded newer image for vespaengine/vespa-build-centos7:latest
/source/travis/travis-build-full.sh: line 9: /etc/profile.d/enable-devtoolset-9.sh: No such file or directory
travis_time:end:0753fe36:start=1587136518407317906,finish=1587136658124084347,duration=139716766441,event=script
[0K[31;1mThe command "./travis/travis.sh" exited with 1.[0m
```

baldersheim:
Now it fails with 
storageframework_gtest_runner_app: /source/storageframework/src/vespa/storageframework/generic/thread/tickingthread.cpp:159: virtual void storage::framework::TickingThreadPoolImpl::start(storage::framework::ThreadPool&): Assertion `!_tickers.empty()' failed.

baldersheim (approved):

geirst (approved):
👍

PR #12972, opened 2020-04-17
Retry once if the acquired file bundle is not readable.

gjoranv:
Should we actually retry (like it is now), or simply wait for a period after a single call to waitFor?

FYI: @bjorncs


hmusum:
We should probably wait for some seconds

gjoranv:
@hmusum PTAL

hmusum (approved):
Great!

PR #4422, opened 2017-12-12
Enable glibc lock elision.

baldersheim:
See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.4_release_notes/new_features_compiler_and_tools
@vekterli PR Only review, I will merge once performance tests have stabilised. Will only take effect on centos 7.
@toregge @havardpe @aressem FYI

PR #12974, opened 2020-04-18
Revert "Enable glibc lock elision."

baldersheim:
Reverts vespa-engine/vespa#4422

I will revert this one as it did not help on the rhel7 lock degradation issue
@vekterli or @toregge PR


vekterli (approved):
:+1:

PR #12975, opened 2020-04-19
Since gcc does not provide lock free 16 byte access we must do so our…

baldersheim:
…selves.

@toregge and @vekterli PR

### DIFF
--- a/vespamalloc/src/vespamalloc/malloc/allocchunk.h
+++ b/vespamalloc/src/vespamalloc/malloc/allocchunk.h
@@ -23,9 +23,14 @@ struct TaggedPtr {
 #if defined(__x86_64__)
 #define VESPA_USE_ATOMIC_TAGGEDPTR
     TaggedPtr load(std::memory_order = std::memory_order_seq_cst) {
+        // Not that this is NOT an atomic load. The current use as the initial load
+        // in a compare_exchange loop is safe as a teader load will just give a retry.
         return *this;
     }
     void store(TaggedPtr ptr) {
+        // Not that this is NOT an atomic store. The current use is in a unit test as an initial
+        // store before any threads are started. Just done so keep api compability with std::atomic as
+        // that is the preferred implementation..
         *this = ptr;
     }
     bool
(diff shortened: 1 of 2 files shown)


vespamalloc/src/vespamalloc/malloc/allocchunk.h
```diff
@@ -20,16 +20,48 @@ struct TaggedPtr {
     TaggedPtr() noexcept : _ptr(nullptr), _tag(0) { }
     TaggedPtr(void *h, size_t t) noexcept : _ptr(h), _tag(t) {}
 
+#if defined(__x86_64__)
+#define VESPA_USE_ATOMIC_TAGGEDPTR
+    TaggedPtr load(std::memory_order = std::memory_order_seq_cst) {
```
toregge, line 25: This reintroduces the ABA problem since we are not guaranteed that the load is atomic.

vespamalloc/src/vespamalloc/malloc/allocchunk.h
```diff
@@ -20,16 +20,48 @@ struct TaggedPtr {
...
+        return *this;
+    }
+    void store(TaggedPtr ptr) {
```
toregge, line 28: This reintroduces the ABA problem since we are not guaranteed that the store is atomic.

toregge (changes requested):

toregge (approved):

vespamalloc/src/vespamalloc/malloc/allocchunk.h
baldersheim: Agree, I will do that.

vespamalloc/src/vespamalloc/malloc/allocchunk.h
baldersheim: I also noticed that it was used in a uint test so I not look at the details. I will do that and see if it is necessary.

toregge (approved):

vespamalloc/src/vespamalloc/malloc/allocchunk.h
```diff
@@ -20,16 +20,53 @@ struct TaggedPtr {
...
+        // store before any threads are started. Just done so to keep api compatible with std::atomic as
+        // that is the preferred implementation..
+        *this = ptr;
+    }
+    bool
+    compare_exchange_weak(TaggedPtr & oldPtr, TaggedPtr newPtr, std::memory_order, std::memory_order) {
+        char result;
+        __asm__ volatile (
+        "lock ;"
+        "cmpxchg16b %6;"
```
vekterli, line 41: What does `%6` signify in this case?

vekterli (commented):
:+1:

PR #12978, opened 2020-04-19
Use std::conditional to selct correct type depending on if std::atomi…

baldersheim:
…c<TaggedPtr> is lock free.

@toregge and @vekterli I think this is cleaner and more correct.

### DIFF
--- a/vespamalloc/src/vespamalloc/malloc/allocchunk.h
+++ b/vespamalloc/src/vespamalloc/malloc/allocchunk.h
@@ -62,7 +62,7 @@ struct AtomicTaggedPtr {
     size_t _tag;
 } __attribute__ ((aligned (16)));
 #else
-    using AtomicTaggedPtr = TagTaggedPtr;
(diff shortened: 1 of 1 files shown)


vespamalloc/src/vespamalloc/malloc/allocchunk.h
```diff
@@ -50,21 +57,20 @@ struct TaggedPtr {
         );
         return result;
     }
-#endif
 
     void *_ptr;
     size_t _tag;
 } __attribute__ ((aligned (16)));
+#else
+    using AtomicTaggedPtr = TagTaggedPtr;
```
toregge, line 65: TagTaggedPtr ?

baldersheim: Fixed

toregge (approved):

vekterli (commented):
:+1:

PR #12981, opened 2020-04-20
Require libcgroup (cgget is used in startup scripts)

hmusum:


aressem (approved):

hmusum:
@baldersheim FYI

baldersheim:
libcgroup-tools is the rpm for rhel7.
Evenrything was in libcgroup for rhel6.

hmusum:
See https://github.com/vespa-engine/vespa/pull/12984 for adding libcgroup-tools where needed

PR #12985, opened 2020-04-20
Handle version of echo not understanding -n option.

toregge:
@baldersheim : please review


baldersheim (changes requested):
I think it is better to change the test to something that does not require an ifdef.
Is -n necessary, or could we use something that is truly universal ?

baldersheim (approved):

PR #12987, opened 2020-04-20
Require libcgroup-tools for all platforms.

toregge:
@hmusum : please review

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -118,7 +118,7 @@ Requires: epel-release
 %endif
 Requires: which
 Requires: initscripts
-Requires: libcgroup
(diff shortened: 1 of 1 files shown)


hmusum:
Seems like there is no libcgroup-tools for YLiInux6 (`cgget` is in `libcgroup`RPM):
`CD-US-CENTRAL-1 [root@17490301-v6-1:/]$ rpm -ql libcgroup| grep cgget
/bin/cgget
/usr/share/man/man1/cgget.1.gz
/bin/cgget
/usr/share/man/man1/cgget.1.gz`

hmusum (approved):

PR #12988, opened 2020-04-20
There are so many combinations that the libstdc++ library can be buil…

baldersheim:
…t that the performance

you get from std::chrono::system_clock and stxad::chrono::steady_clock has a dramatic performance difference.
On RHEL7/Centos with an witout this patch is 18ns vs 550ns.

@vekterli and @toregge @havardpe PR
Any other suggestions ?

### DIFF
--- a/vespalib/src/vespa/vespalib/util/time.cpp
+++ b/vespalib/src/vespa/vespalib/util/time.cpp
@@ -52,3 +52,29 @@ Timer::waitAtLeast(duration dur, bool busyWait) {
 }
 
 }
+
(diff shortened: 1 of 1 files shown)


havardpe (approved):
I did not even know this was possible...

vekterli (commented):
🤯👍 a quick explanation of how this works would be great. Does it take precedence over the libstdc++ symbols?

baldersheim:
To be honest, I am entirely sure. But I assume that library load order matters here.
I think @toregge might provide the gory details.

PR #12989, opened 2020-04-20
Do not add empty children.

baldersheim:
@vekterli PR


vekterli (approved):
:+1: We should consider adding a conditional add-method to the trace class itself, as this is a common pattern

baldersheim:
@vekterli PTAL I added another one before I saw your approval.
Yes, I agree a conditional add would be good, but to do that with transfer by value ?

vekterli (approved):
:+1:

PR #12990, opened 2020-04-20
Revert "There are so many combinations that the libstdc++ library can be buil…"

baldersheim:
Reverts vespa-engine/vespa#12988
@vekterli or @toregge or @havardpe it needs some spcialhandling as clock_gettime is in non-standard library there.


vekterli (commented):
:+1:

PR #12991, opened 2020-04-21
Revert "Revert "There are so many combinations that the libstdc++ library can be buil…""

baldersheim:
Reverts vespa-engine/vespa#12990


baldersheim:
@havardpe @vekterli @toregge PR
It is the same but with a link adjustment for rhel 6 and more elaboration in the comment.

havardpe (approved):
If I remember correctly, this issue was fixed on RHEL6 by linking with the rt library (and still using the std::chrono functions). Is this an option for RHEL7 as well? (or was it moved from rt to libc and disabled due to compile options...?)

toregge (approved):

toregge (approved):

PR #12994, opened 2020-04-21
GC unused code.

baldersheim:
@vekterli PR
@toregge FY
Btw, what about the tools
vespa-storage-cmd
storage_analyzedistribution_app
storage_generatedistributionbits_app
storage_getidealstate_app


vekterli (approved):
:+1: most of the tools mentioned were presumably made during the design and implementation phase of the distribution algorithm, so we probably don't need these anymore

PR #12998, opened 2020-04-21
Lesters/bert model integration

lesters:
@bratseth Please review.

These are changes primarily in `model-integration` needed to parse, convert and evaluate our BERT/transformer test model correctly. More changes are incoming later in `searchlib` and `config-model` to actually be able to deploy the model.


bratseth (approved):
Amazing work!

PR #13003, opened 2020-04-21
A failed attempt at unlinking a directory causes another exception

toregge:
on Darwin due to unlink system call returning a different error
code.

@baldersheim : please review


baldersheim (approved):
could much of fileutil be replaced with std::filesystem stuff.

PR #13005, opened 2020-04-22
Create flag for rolling out nlb proxy protocol

tokle:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -241,7 +241,13 @@
...
             "nginx-upstream-proxy-protocol", false,
             "Whether the nginx should enable proxy-protocol for all upstreams",
-            "Takes effect immediately");
+            "Takes effect immediately",
+            ZONE_ID, APPLICATION_ID);
+
+    public static final UnboundBooleanFlag NLB_PROXY_PROTOCOL = defineFeatureFlag(
+            "nlb-proxy-protocol", false,
+            "Configure NLB to use proxy protocol",
+            "Takes effect on next application redeploy");
```
mpolden, line 250: No dimensions?

mpolden (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
tokle: Nor sure how that was removed... should definitely have application_id dimension

mpolden (approved):

PR #13009, opened 2020-04-22
Add component

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
```diff
@@ -183,6 +183,7 @@ public ContainerCluster(AbstractConfigProducer<?> parent, String subId, String n
         addSimpleComponent("com.yahoo.container.jdisc.state.StateMonitor");
         addSimpleComponent("com.yahoo.container.jdisc.ContainerThreadFactory");
         addSimpleComponent("com.yahoo.container.handler.VipStatus");
+        addSimpleComponent("ai.vespa.cloud.SystemInfo");
```
gjoranv, line 186: Is this relevant for e.g. metrics-proxy, clustercontroller and logserver? If not, it should be moved to ApplicationContainerCluster.

bratseth: Done - thanks!

gjoranv (approved):

PR #13015, opened 2020-04-22
Use DockerImage instead of String

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
```diff
@@ -153,6 +154,12 @@ public Builder endpointCertificateMetadata(String serialized) {
         }
 
         public Builder dockerImageRepository(String dockerImageRepository) {
+            if (dockerImageRepository == null) return this;
```
freva, line 157: The `null` behaviors in this class are all over the place. `endpointCertificateMetadata(String)`, `containerEndpoints(String)` ignores it, as does `athenzDomain(AthenzDomain)`, while `athenzDomain(String)`, and `tlsSecretsKeyName(String)` (correctly, IMO) resets the state.

freva (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
hmusum: Yes, good point, I'll try to clean it up

freva (approved):
:1st_place_medal:

PR #13019, opened 2020-04-22
Measure platform and OS change duration per host 

mpolden:
@bratseth


mpolden:
FYI @yngveaasheim

mpolden:
merge unfreeze

bratseth (approved):
LGTM. 

I wish you'd tune down your use of "var" though. It makes things harder to read because the code doesn't say what things are.

PR #13020, opened 2020-04-22
Increase streaming search default timeout trace sampling probabilities/rates

vekterli:
@geirst please review
@yngveaasheim FYI

May increase these further (and decrease required timeout multiplier) later, but let's see what this gets us first.


geirst (approved):
👍

PR #13022, opened 2020-04-22
Extend lifetime of string containing namespace specific portion of document id.

toregge:
@baldersheim : please review
@vekterli : FYI


baldersheim:
I think it is better to change String2ResultNode to use vespalib::stringref instead instead of using a const vespalib::string & reference instead of adapting to the bad code.

toregge:
Obsoleted by upcoming better fix from @baldersheim.

PR #13027, opened 2020-04-22
Split suggesting and autoscaling.

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Application.java
```diff
@@ -52,7 +56,7 @@ public Application with(Cluster cluster) {
     public Application withClusterLimits(ClusterSpec.Id id, ClusterResources min, ClusterResources max) {
         Cluster cluster = clusters.get(id);
         if (cluster == null)
-            cluster = new Cluster(id, min, max, Optional.empty());
+            cluster = new Cluster(id, min, max, Optional.empty(), Optional.empty());
         else
             cluster = cluster.withLimits(min, max);
```
freva, line 61: Is this safe? Looks like this is called during deployment and deciding the targeted resources. If this is not a new deployment that has run for some time, then it probably has target resources. If the user then changes the limits so that the target is outside, this will throw?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Application.java
bratseth: It will not throw but wipe the target. I think that is correct.

In the case where new limits are set which are outside the current size we'll pick the lowest limit. That's not the right choice if the entire new window is *below* the current size, but this isn't perfectly straightforward since it's multidimensional ... so I plan to do some more work on making that safe, but it's orthogonal to this.

freva: Where does it wipe it? `Cluster::withLimits` creates `new Cluster` with the new `min`, `max` while keeping old `target`?

bratseth: In the Cluster constructor.

PR #13028, opened 2020-04-23
Extend the timeout for config-proxy to start from 1m to 3m.

baldersheim:
@hmusum PR
@hakonhall FYI


hmusum (approved):

hakonhall:
In the example we have looked at, it would have to wait 10m, so increase to 15m?

baldersheim:
I do not think that is necessary. There will be progress in the whole period. They will all be fighting for the same resources and all will get some. We just give 3x more to complete, and that should help. If not we should consider giving it more.

PR #13030, opened 2020-04-23
Unwrap reference wrappers to avoid extra indirections via invalid memory.

toregge:
@havardpe : please review
@vekterli : FYI


havardpe (approved):
maybe we should consider using TensorFunction instead of auto as well (which would have avoided the issue in the first place)...

PR #13034, opened 2020-04-23
Add support for aggregating over B-tree keys

vekterli:
@toregge @geirst please review carefully. Only a subset of aggregation tests are implemented explicitly for keys; it's not obvious how to create any kind of parameterized fixture for testing both without lots of duplication and the test file tests a _lot_ of various cases, not all related to aggregation. Not super happy with how keys vs. value branching is done, but it's a binary choice so I felt introducing enums would be a bit overkill.

This complements the existing support for aggregating over values.

Let aggregate calculator specify whether it expects to be invoked
with keys or values.

In the current implementation there is some code duplication
that could have been removed by using e.g. tag dispatch, but
this is a pragmatic choice intended to guarantee these changes
do not introduce any performance regressions for existing
code using value aggregation. Can be refactored later once we
have a functional baseline with this for both keys and values.

### DIFF
--- a/vespalib/src/vespa/vespalib/btree/btreeremover.hpp
+++ b/vespalib/src/vespa/vespalib/btree/btreeremover.hpp
@@ -110,11 +110,17 @@ remove(BTreeNode::Ref &root,
     NodeAllocatorType &allocator(itr.getAllocator());
     AggrT oldca(AggrCalcT::hasAggregated() ? lnode->getAggregated() : AggrT());
     AggrT ca;
-    if (AggrCalcT::hasAggregated() &&
-        aggrCalc.remove(lnode->getAggregated(),
-                        aggrCalc.getVal(lnode->getData(idx)))) {
+    if constexpr (AggrCalcT::hasAggregated()) {
+        bool removed;
+        if constexpr (AggrCalcT::aggregate_over_values()) {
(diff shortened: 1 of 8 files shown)


vespalib/src/vespa/vespalib/btree/btreeremover.hpp
```diff
@@ -110,11 +110,17 @@ remove(BTreeNode::Ref &root,
     NodeAllocatorType &allocator(itr.getAllocator());
     AggrT oldca(AggrCalcT::hasAggregated() ? lnode->getAggregated() : AggrT());
     AggrT ca;
-    if (AggrCalcT::hasAggregated() &&
-        aggrCalc.remove(lnode->getAggregated(),
-                        aggrCalc.getVal(lnode->getData(idx)))) {
+    if constexpr (AggrCalcT::hasAggregated()) {
+        bool removed;
```
toregge, line 114: Consider renaming removed to need_aggregation_recalc.

toregge (approved):

vespalib/src/vespa/vespalib/btree/btreeremover.hpp
vekterli: Fixed

toregge (approved):

geirst:
This is pretty unknown territory for me, so @toregge should do the main review on this one. I can have a look afterwards.

geirst (approved):
👍

PR #13037, opened 2020-04-23
To avoid deadlock when doing synchrounous execution in other chains, …

baldersheim:
…we use one threadpool per chain.

@bratseth PR
@jobergum and @kkraune FYI


bratseth (approved):
LGTM, although I think it would be more logical to put it in the DocprocExecutor, and yes this has grown from something very simple to ... something not quite so simple.

PR #13041, opened 2020-04-23
Use java.util.logging.Level instead of vespalog's LogLevel.

gjoranv:
- The purpose is to allow using config in vespalog.


bratseth (approved):

bratseth (approved):

hmusum:
👍

PR #13044, opened 2020-04-23
Revert "Use java.util.logging.Level instead of vespalog's LogLevel."

hmusum:
Reverts vespa-engine/vespa#13041

Build broke


bratseth:
I already fixed that

bratseth (approved):
Hm, no there was a lot more here.

PR #13046, opened 2020-04-23
Gjoranv/dont depend on vespalog take 2

bratseth:
@gjoranv I did some more here, but a lot more is still needed.

### DIFF
--- a/config/pom.xml
+++ b/config/pom.xml
@@ -53,7 +53,7 @@
     </dependency>
     <dependency>
       <groupId>com.yahoo.vespa</groupId>
-      <artifactId>vespalog</artifactId>
+      <artifactId>defaults</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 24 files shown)


gjoranv:
Thanks for trying! I'll have another go. We should only need to change these modules, that will now be pulled in by vespalog:
> [INFO] +- com.yahoo.vespa:config:jar:7-SNAPSHOT:compile
> [INFO] |  +- com.yahoo.vespa:config-lib:jar:7-SNAPSHOT:compile
> [INFO] |  +- com.yahoo.vespa:configgen:jar:7-SNAPSHOT:compile
> [INFO] |  +- com.yahoo.vespa:jrt:jar:7-SNAPSHOT:compile
> [INFO] |  |  \- com.yahoo.vespa:security-utils:jar:7-SNAPSHOT:compile
> [INFO] |  +- com.yahoo.vespa:yolean:jar:7-SNAPSHOT:compile
> [INFO] |  +- com.yahoo.vespa:http-utils:jar:7-SNAPSHOT:compile

For the other failing modules, they probably just need to have vespalog explicitly added as a dependency, as it was earlier pulled in transitively by `config`.

bratseth:
Imo it would be better to just change them to use the java.util.logging API.

gjoranv:
Replaced by #13061

PR #13047, opened 2020-04-23
Extract test only code in DocprocService to a packake private class

baldersheim:
@bratseth PR


bratseth (commented):
I guess this will break almost all unit tests of docprocs so I don't think we can just merge it.
The original though with this class, way back, was that it shouldn't do any of the threading stuff but just provide an API for submitting and taking tasks. Then it would be easy to test. Can't we just use it like that? How are people supposed to test now?

baldersheim:
Ah,  so this was intended as test utility class. The issue here is that we did not make a smaller interface for use in our code.
Then I will se if that is doable without breaking any api's.

bratseth:
Thanks. It was intended as a layer that provided processing of documents, without having an opinion on environment/thread model, for both runtime environments and tests. Tests works best if they use the same logic as production, and the production code gets better if it is easily testable. I'm not entirely sure that it is separable like this, but it seems to me it was built out in breadth (more parallel ways to do almost the same) rather than depth (layering) for no particular good reason. Also not sure it can be untangled from where we are now :-)

PR #13053, opened 2020-04-24
Support inspection of partition size

mpolden:
@freva


mpolden:
Duplicates recentely added internal code.

PR #13055, opened 2020-04-24
Optimize B-tree bucket DB lookup with used-bits aggregation

vekterli:
@geirst please review

By tracking the minimum used bits count across all buckets in
the database we can immediately start seeking at that implicit
level in the tree, as we know no parent buckets can exist above
that level.

Local synthetic benchmarking shows the following results with a
DB size of 917,504 buckets and performing `getParents` for all
buckets in sequence:

Before optimization:
 - B-tree DB: **0.593321** seconds
 - Legacy DB: 0.227947 seconds

After optimization:
 - B-tree DB: **0.213738** seconds
 - Legacy DB: (unchanged)


geirst (approved):
Nice 👍

PR #13062, opened 2020-04-25
Revert "Optimize B-tree bucket DB lookup with used-bits aggregation"

vekterli:
@baldersheim please review

This caused system test regressions. Need to be investigated properly.

Reverts vespa-engine/vespa#13055


toregge (approved):

baldersheim (commented):
👍

PR #13063, opened 2020-04-25
Extend BTree frozen view with empty() and getAggregated() methods.

toregge:
@geirst : please review
@vekterli : FYI


geirst (approved):
👍

PR #13064, opened 2020-04-25
Let size of thread pool and Q follow number of cores on the machine w…

baldersheim:
…here it is running.

If auto detected num worker threads will use number of #cores * 4, quesize will use #numWorkers * 4.
No changes of the default value in this commit.

@bratseth and @bjorncs PR


bratseth (approved):
👍

bjorncs (commented):
👍

PR #13067, opened 2020-04-26
Lesters/bert searchlib and config model

lesters:
@bratseth Please review.

The main changes here are for resolving types and serializing functions. Both would break down for large rank profiles (e.g. imported ml models) with a large degree of diamond-shaped dependencies on functions without arguments.


bratseth (approved):
Great!

PR #13072, opened 2020-04-26
Use fewer threads to get next generation of config

hmusum:
Check for next config generation in one thread.
Simplify Subscriber.
Use a ConcurrentHashMap for active subscribers.
Reduces the number of threads used by config proxy from ~230 to ~30
Had to reduce interval between nextGeneration calls down to 10 milliseconds to avoid slowdown getting config when running a system test

Please review only, will merge later.


config-proxy/src/main/java/com/yahoo/vespa/config/proxy/RpcConfigSourceClient.java
```diff
@@ -69,14 +74,10 @@
         this.memoryCache = memoryCache;
         this.delayedResponses = new DelayedResponses();
         checkConfigSources();
-        exec = Executors.newCachedThreadPool(new DaemonThreadFactory("subscriber-"));
+        nextConfigFuture = nextConfigScheduler.scheduleAtFixedRate(this, 0, 10, MILLISECONDS);
```
baldersheim, line 77: Every 10ms sounds like it is often enough :)

baldersheim (approved):
That was quite simple....

PR #13074, opened 2020-04-26
Use gcc toolset 9 on CentOS 8 and RHEL 8

toregge:
@baldersheim : please review


build_settings.cmake
```diff
@@ -79,7 +84,11 @@ if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -latomic -ldl")
   endif()
 else()
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id -latomic -ldl -Wl,-E")
+  if(VESPA_ATOMIC_LIB STREQUAL "")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id -ldl -Wl,-E")
+  else()
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--build-id -latomic -ldl -Wl,-E")
+  endif()
```
baldersheim, line 91: Is it not simpler to set VESPA_ATOMIC_LIB to "-latomic" and do a simple append instead of 5 lines.

baldersheim: Ignore comment I see it is used further down in the PR

baldersheim (approved):

PR #13079, opened 2020-04-27
Create policies for billing API resources

oyving:
More fine-grained access policies for billing API.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
@@ -142,7 +142,32 @@ enum Policy {
     /** Access to /payment/notification */
     paymentProcessor(Privilege.grant(Action.create)
                                 .on(PathGroup.paymentProcessor)
-                                .in(SystemName.PublicCd));
+                                .in(SystemName.PublicCd)),
+
+    /** Read your own instrument information */
+    paymentInstrumentRead(Privilege.grant(Action.read)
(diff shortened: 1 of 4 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
```diff
@@ -142,7 +142,32 @@
...
+
+    /** Ability to update tenant payment instrument */
+    paymentInstrumentUpdate(Privilege.grant(Action.update)
+                                     .on(PathGroup.billingInstrument)
+                                     .in(SystemName.PublicCd)),
+
+    /** Ability to remove your own payment instrument */
+    paymentInstrumentDelete(Privilege.grant(Action.delete)
+                                     .on(PathGroup.billingInstrument)
+                                     .in(SystemName.PublicCd)),
```
olaaun, line 160: What's the benefit of having these as three separate policies?

oyving: They are assigned differently in roles?

olaaun:
Tests need fixing

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
olaaun: Alright

olaaun (approved):

jonmv (commented):
A bit late to the party here, but looks good to me :)

PR #13083, opened 2020-04-27
Add container name to metadata

olaaun:


freva (commented):
`coredump_path` should be the path to the actual directory instead of the directory to all coredumps for a container, how about:
```
metadata.put("coredump_path", doneCoredumpsPath.resolve(context.containerName().asString()).resolve(coredumpDirectory.getFileName()).toString());
```
here: https://github.com/vespa-engine/vespa/blob/7dd1c5d8705885b9819f5b46dec41f0a2f2335c4/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java#L167
?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -255,4 +256,7 @@ private Dimensions generateDimensions(NodeAgentContext context) {
         return dimensionsBuilder.build();
     }
 
+    public Path getDoneCoredumpsPath() {
```
freva, line 259: Not used

freva (approved):

PR #13084, opened 2020-04-27
Optimize B-tree bucket DB lookup with used-bits aggregation

vekterli:

Fix and optimisation based on @toregge’s suggestions (slightly modified).

Previous iteration of the code had a regression that would trigger under the following scenario:
- A non-empty bucket DB had a minimum aggregated used bit count of X
- The queried bucket had a used bit count of Y < X

Since our bucket resolving previously started with the first bucket in the DB in the first iteration of the seek loop, for all other queries than for the very first bucket we’d end up invoking `next_parent_bit_seek_level`. This function would sanity check that the bit counts of the buckets it computed over were at least as big as the current working tree depth level (to ensure  progress down the tree). Since the query bucket was an input here and it would have a bit count lower than the working bit count, this sanity check would fail and the distributor would blow up in a beautiful fireworks display.

Moving to a lower bound seek based on the minimum number of bits in the DB _before_ the loop has two major benefits:
- The loop body (and thus `next_parent_bit_seek_level`) cannot be invoked when the query bucket’s bit count is < the DB bit count, as the lower bound bucket will have a greater key than the query bucket’s key by definition.
- We skip the entire loop body and re-seeking in the common case where the bucket tree has no inconsistent splits. This further speeds up the B-tree database compared to the legacy implementation.

By tracking the minimum used bits count across all buckets in
the database we can immediately start seeking at that implicit
level in the tree, as we know no parent buckets can exist above
that level.

Local synthetic benchmarking shows the following results with a
DB size of 917504 buckets and performing getParents for all
buckets in sequence:

Before optimization:
 - B-tree DB: 0.593321 seconds
 - Legacy DB: 0.227947 seconds

After optimization:
 - B-tree DB: 0.191971 seconds
 - Legacy DB: (unchanged)

### DIFF
--- a/storage/src/tests/distributor/bucketdatabasetest.cpp
+++ b/storage/src/tests/distributor/bucketdatabasetest.cpp
@@ -202,66 +202,84 @@ TEST_P(BucketDatabaseTest, find_parents) {
     // test that it is so for now to avoid breaking the world.
     EXPECT_EQ(
             std::string("0"),
-            doFindParents(toVector(document::BucketId(17, 0xcafe)),
-                          document::BucketId(17, 0xcafe)));
+            doFindParents({BucketId(17, 0xcafe)},
+                          BucketId(17, 0xcafe)));
 
     EXPECT_EQ(
             std::string("1,2"),
-            doFindParents(toVector(document::BucketId(1, 0x0),
-                                   document::BucketId(1, 0x1),
-                                   document::BucketId(2, 0x1)),
-                          document::BucketId(16, 0x1)));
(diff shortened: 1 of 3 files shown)


toregge (approved):

vekterli:
Unrelated Travis failure:
```
[ERROR] Errors: 
[ERROR]   HttpServerTest.requireThatConnectorWithProxyProtocolMixedEnabledAcceptsBothProxyProtocolAndHttps:727->sendJettyClientRequest:740
```

geirst (approved):
👍

PR #13087, opened 2020-04-28
Use /tmp/ for maven local repository

hmusum:
Using $HOME/.m2/repository (default) does not work on ylinux7, use /tmp instead, which is better anyway.


jonmv (approved):

freva:
Consider using `$VESPA_HOME/tmp` instead to avoid writing to thinpool

jonmv:
Right, that's probably something we should do instead.

hmusum:
Thanks, fixed in https://github.com/vespa-engine/vespa/pull/13095

PR #13089, opened 2020-04-28
Take both legacy lock and application lock

mpolden:
@bratseth


bratseth (approved):

mpolden:
Will merge this myself (when I can keep an eye on CD).

PR #13090, opened 2020-04-28
Indicate pending runs and whether they are upgrades

jonmv:
@freva please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
@@ -193,7 +193,7 @@ public class DeploymentApiHandler extends LoggingRequestHandler {
                                                                       .allMatch(versions -> versions.sourcePlatform()
                                                                                                     .map(versions.targetPlatform()::equals)
                                                                                                     .orElse(true))
-                                                             ? "revision" : "upgrade");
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/DeploymentApiHandler.java
```diff
@@ -186,6 +188,13 @@ private HttpResponse root(HttpRequest request) {
                           jobStatus.pausedUntil().ifPresent(until -> jobObject.setLong("pausedUntil", until.toEpochMilli()));
                           jobStatus.coolingDownUntil(status.application().require(instance.instance()).change())
                                    .ifPresent(until -> jobObject.setLong("coolingDownUntil", until.toEpochMilli()));
+                          if (jobsToRun.containsKey(job)) {
+                              jobObject.setString("pending", jobsToRun.get(job).stream()
+                                                                      .allMatch(versions -> versions.sourcePlatform()
+                                                                                                    .map(versions.targetPlatform()::equals)
+                                                                                                    .orElse(true))
+                                                             ? "revision" : "upgrade");
```
freva, line 196: Use "application" and "platform" instead or is this something else?

jonmv: Hmm, ok.

freva (approved):

PR #13094, opened 2020-04-28
Implement hasReply avoid copying the shared_ptr just to peak at the r

baldersheim:
…esult.

@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/persistence/persistenceutil.h
+++ b/storage/src/vespa/storage/persistence/persistenceutil.h
@@ -49,7 +49,7 @@ public:
     api::StorageReply & getReply() {
         return *_reply;
     }
-    api::StorageReply::SP stealReplySP() {
+    api::StorageReply::SP && stealReplySP() && {
         return std::move(_reply);
     }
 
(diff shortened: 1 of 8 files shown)


storage/src/vespa/storage/persistence/persistenceutil.h
```diff
@@ -43,20 +42,27 @@ class MessageTracker : protected Types {
...
-    api::StorageReply::SP getReply() {
-        return _reply;
+    bool hasReply() const { return bool(_reply); }
+    const api::StorageReply & getReply() const {
+        return *_reply;
+    }
+    api::StorageReply & getReply() {
+        return *_reply;
+    }
+    api::StorageReply::SP getReplySP() {
```
vekterli, line 52: I'd prefer this to be called something like `steal_reply_sp`, as the current name implies the function doesn't have any side effects

vekterli (approved):
:+1: one comment for consideration

storage/src/vespa/storage/persistence/persistenceutil.h
toregge, line 52: Please use an rvalue qualifier for this method.

baldersheim: Done.

baldersheim (commented):
?

PR #13097, opened 2020-04-29
Add method to fetch constants as tensors

lesters:
@bratseth Please review.

During import of BERT model, over 50 seconds was spent in converting tensors to strings and back. This reduces that to effectively 0.


bratseth (approved):
Great

PR #13106, opened 2020-04-29
Wait longer for servers to reach barrier

hmusum:
1. Wait up to 2 seconds for all to reach barrier.
2. If not, wait up to 4 seconds for the server that waits for the barrier to
   be one of the respondents AND a majority of servers have reached barrier.
3. If not, wait for a majority of servers to have reached barrier.

The timeouts are probably wrong, need some experience before adjusting them.

Please review only, will merge later.


jonmv (approved):
LGTM, but what is the idea?

hmusum:
We once talked about the fact that the server that receives the deployment request sometimes ends up not reaching the barrier, so it will not have prepared and/or activated a deployment when it returns. Which If that happens, we sometimes see failures or issues with deleting the app. This is more of an experiment to gather some data about that and it might give some clues about some of the deployment failures we see.

PR #13114, opened 2020-04-29
Bratseth/allow non allocatable limits

bratseth:
Well this was a lot more work than anticipated but now it's possible to just configure any limits that contains an AWS flavor. Another use for autoscaling: When you don't want to think about which flavors exist.


freva (approved):
LGTM, but requires PR in internal repo to follow the API changes

bratseth:
Thanks!

PR #13115, opened 2020-04-29
vespaVersion -> vespaversion

aressem:


jonmv (changes requested):
This is a different `vespaVersion` >_< This is the reason I wanted to alias them. Really, I would like to rename them entirely, just didn't dare do it yet. I have a ticket to follow this up.

aressem:
Ok. A bit confusing :)

PR #13119, opened 2020-04-30
If we enter blocking state we should sample every second in order to …

baldersheim:
…quickly detect when we are out of it.

@geirst PR


geirst (approved):
👍

PR #13127, opened 2020-04-30
Mount /etc/machine-id and /var/log/journal

hakonhall:
When /var/log/journal is mounted explicitly (/var/log is already mounted, but
that's not sufficient), Docker/systemd will not mount the container's
/var/log/journal/MACHINE-ID on the host's /var/log/journal/MACHINE-ID.

Also, upgrades and restarts of a container will now reuse the same machine ID,
as /etc/machine-id is written once and mounted in.  Dirtying the container node
means a new machine ID must be written for the same hostname later.


freva (approved):
We should just consider adding `TaskContext::fileSystem` or something to make file system usage more consistent everywhere...

hakonhall:
An alternative is for the node agent context to reuse the filesystem belonging to any Path parameter, and otherwise only return paths as Strings.

freva:
That would require to replace pretty much all usages of `Paths::get` with call to `FileSystem` just to get the tests working, and even then there is still potential for some nasty surp

PR #13130, opened 2020-05-01
 Revert "Also map /opt/vespa/logs/vespa" and "Mount /etc/machine-id and /var/log/journal"

hmusum:
Reverts #13127 and #13129

Container seem to be unable to start, all fail with something like
` com.yahoo.vespa.hosted.node.admin.task.util.process.ChildProcessFailureException: Command 'nsenter --net=/proc/396780/ns/net -- ip6tables-restore /tmp/iptables-restore12860967353789305634.nat 2>&1' terminated with exit code 1: stdout/stderr: 'nsenter: cannot open /proc/396780/ns/net: No such file or directory
        '


freva (approved):
Must be merged with an internal revert (vespa-yahoo#4388)

freva:
@hakonhall The issue is that the mounts for YLinux6 now symmetric (`container-storage/home/y/*` -> `/home/y/*` instead of `container-storage/opt/vespa/*` -> `/home/y/*`)

hakonhall:
Thank you

PR #13133, opened 2020-05-01
When using multiple threads per search we will never have 100% dutycy…

baldersheim:
…cle to to the law of Amdahl.

To be on the safe side we ensure that we have enough threads to reach full cpu utilization even with a 0% cuty cycle.

@bratseth or @geirst PR


hmusum (approved):

bratseth (approved):

geirst (commented):
👍

PR #13134, opened 2020-05-02
- Reserve space upfront in the vectors.

baldersheim:
- Remove virtual interface from some very frequently called methods that are not necessary.
- Use std::move for vector and shared_ptr.
- Some c++11ification.

@geirst or @havardpe PR


geirst (approved):
👍

PR #13140, opened 2020-05-03
Require the same FileSystem provider

hakonhall:
TL;DR  Require all paths with NodeAgentContext[Impl] to use the same file system provider, and add a `FileSystem fileSystem()` method.

NodeAgentContext[Impl] requires paths to have specific providers, otherwise a ProviderMismatchException will be thrown, which it often does when adding tests involving ths class.  The current API is best described by defining a few terms:

* Let PathDefault be a Path with a file system provider matching that of FileSystems.getDefault().provider().
* Let PathContainerStorage be a Path with the same file system provider as the context's pathToContainerStorage constructor param
* Let PathVespaHome be a Path with the same file system provider as the context's pathToVespaHome constructor param.

The context methods handling paths have these requirements:

* `Path pathOnHostFromPathInNode(Path pathInNode)` maps a PathDefault to PathContainerStorage
* `Path pathInNodeFromPathOnHost(Path pathOnHost)` maps a PathContainerStorage to PathDefault
* `Path pathInNodeUnderVespaHome(Path relativePath)` maps a PathVespaHome to PathVespaHome
* `Path rewritePathInNodeForWantedDockerImage(Path path)` maps a PathDefault to PathDefault, and requires PathVespaHome == PathDefault, if it is to work under all circumstances.

This PR requires all paths with NAC[I] to have the same provider, otherwise we'll detect this and throw as early as possible.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextImpl.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextImpl.java
@@ -15,7 +15,7 @@ import com.yahoo.vespa.hosted.node.admin.docker.DockerNetworking;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Path;
-import java.nio.file.Paths;
+import java.nio.file.ProviderMismatchException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
@@ -26,7 +26,6 @@ import java.util.logging.Logger;
  * @author freva
  */
 public class NodeAgentContextImpl implements NodeAgentContext {
-    private static final Path ROOT = Paths.get("/");
 
     private final String logPrefix;
     private final NodeSpec node;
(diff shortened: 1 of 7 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextImpl.java
```diff
@@ -296,8 +300,9 @@ public String getCloudNativeRegionName() {
                             return getId().region().value();
                         }
                     }),
-                    Optional.ofNullable(pathToContainerStorage).orElseGet(() -> fileSystem.getPath("/home/docker")),
-                    Optional.ofNullable(pathToVespaHome).orElseGet(() -> fileSystem.getPath("/opt/vespa")),
+                    fileSystem,
+                    fileSystem.getPath("/home/docker/container-storage"),
```
freva, line 304: The `container-storage` addition means more typing in tests... :(

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextImpl.java
hakonhall: yes, but it matches the expected path.  I was hoping this could lead to simplifying the NodeAgentContextImpl constructor by removing 2 path parameters - but I ended up keeping them.

PR #13150, opened 2020-05-04
Remove deprecated bucket cross-disk move functionality

vekterli:
@baldersheim please review
@geirst FYI

The notion of multiple disks hasn't been supported since we
removed VDS, and likely won't be in the future either.


baldersheim (approved):

baldersheim:
Note that a c++ test did not complete, either hanging or eternal loop.

PR #13157, opened 2020-05-05
Downgrade log level when node watch races with internal state resets

vekterli:
@geirst please review. Causes spurious system test failures due to log error checks kicking in.

This appears to happen if pending data node watches are triggered with
failures after the cluster controller has already reset the internal
voting state due to connectivity issues. Internal state is rebuilt
automatically, so no need to scream errors into the logs for this.


geirst (approved):
👍

PR #13162, opened 2020-05-06
Revert "Add OS version parameter to HostProvisioner"

hmusum:
Reverts vespa-engine/vespa#13156

This broke code in internal repo.


freva:
The hosted part is in `vespa/vespa-yahoo#4411`, which was merged at the same time, but for some reason wasn't included in the same build... Next build should be OK

hmusum:
👍

PR #13165, opened 2020-05-06
Revert "Do nothing if session status is unchanged"

hmusum:
Reverts vespa-engine/vespa#13142

Seems like this broke system tests


jonmv:
Interesting ... it's the sort of thing that _shouldn't_ change anything, but I was afraid it might.

PR #13168, opened 2020-05-06
Reduce default maintainer interval for config server maintainers

hmusum:
No reason to wait so long between each run.

Please review only, will merge after changes to expiry time for sessions have been rolled out.


jonmv (commented):
What could possibly go wrong?

... 

except things. Application locks held all the time, perhaps?

hmusum:
Nah, this isn't as important as I thought and there is some risk, closing.

PR #13169, opened 2020-05-06
Model cloud features explicitly

mpolden:
Hold merge as host-admin must be changed to generate the correct config.

@bratseth

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
@@ -63,7 +63,11 @@ public class Zone {
     }
 
     /** Returns the current cloud */
-    public Cloud cloud() { return cloud; }
+    // TODO: For compatibility with older config models. Remove after June 2020
+    public CloudName cloud() { return cloud.name(); }
+
+    /** Returns the current cloud */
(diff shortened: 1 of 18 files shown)


mpolden:
FYI @freva

config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
```diff
@@ -38,29 +41,29 @@ public Zone(Environment environment, RegionName region) {
...
-        this.cloudName = cloudName;
+        this.cloud = cloud;
         this.systemName = systemName;
         this.environment = environment;
         this.region = region;
         this.nodeFlavors = Optional.ofNullable(nodeFlavors);
     }
 
     /** Returns the current cloud */
-    public CloudName cloud() { return cloudName; }
```
bratseth, line 63: Won't this break old model versions?

bratseth (commented):
Otherwise, great!

config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
mpolden: Yes, thanks!

bratseth (approved):

PR #13170, opened 2020-05-06
Reduce the time we keep expired remotes sessions and fix comment

hmusum:


bjormel (approved):

bjormel:
Was there any specific reason for keeping these for 30 days?

hmusum:
No, it was more of a "better be on the safe side and we are allowed 30 days, so use that".

PR #13171, opened 2020-05-06
Reduce expiry time in steps (was 30 days)

hmusum:
See https://github.com/vespa-engine/vespa/pull/13170


hmusum:
Will revert in https://github.com/vespa-engine/vespa/pull/13172 first instead

PR #13183, opened 2020-05-07
signal if a blueprint wants a global filter

arnej27959:

@havardpe FYI

next is adding the API for sending the global filter into a blueprint - i guess something like
setup_global_filter(BitVector::UP global_filter)
where the unique_ptr may be null to signal that there is no global filter (so we can assume all hits from ANN are actually returned as hits).

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.h
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.h
@@ -63,7 +63,7 @@ public:
         uint32_t          _cost_tier;
         uint32_t          _tree_size;
         bool              _allow_termwise_eval;
-        bool              _wants_global_filter;
+        bool              _want_global_filter;
 
     public:
         static constexpr uint32_t COST_TIER_NORMAL = 1;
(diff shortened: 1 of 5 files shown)


geirst (approved):
👍 I suggest we add a unit test in queryeval/intermediate_blueprints_test.cpp or queryeval/blueprint_test.cpp that verifies that the flag on an intermediate node is correctly inferred.

geirst:
Perhaps we should call it set_global_filter()?, as the filter itself is already calculated at this point. With _setup_ it sounds that we should do something on the filter. I also think we should use a shared_ptr as the filter can be shared between multiple blueprints.

searchlib/src/vespa/searchlib/queryeval/blueprint.h
```diff
@@ -63,6 +63,7 @@ class Blueprint
         uint32_t          _cost_tier;
         uint32_t          _tree_size;
         bool              _allow_termwise_eval;
+        bool              _wants_global_filter;
```
havardpe, line 66: please init in constructor

havardpe (approved):
I prefer 'want' over 'wants'

geirst (approved):
👍

PR #13185, opened 2020-05-07
Balder/handle operation with shorter vectors

baldersheim:
@havardpe PR
Only the second commit has any semantics. The first is just code maintenance.

### DIFF
--- a/searchlib/src/vespa/searchlib/common/bitvector.cpp
+++ b/searchlib/src/vespa/searchlib/common/bitvector.cpp
@@ -24,7 +24,7 @@ void verifyContains(const search::BitVector & a, const search::BitVector & b) __
 
 void verifyContains(const search::BitVector & a, const search::BitVector & b)
 {
-    if ((a.getStartIndex() < b.getStartIndex()) || (a.size() > b.size())) {
+    if (a.getStartIndex() < b.getStartIndex()) {
         throw IllegalArgumentException(make_string("[%d, %d] is not contained in [%d, %d]",
                                                    a.getStartIndex(), a.size(), b.getStartIndex(), b.size()),
                                        VESPA_STRLOC);
@@ -178,8 +178,17 @@ void
 BitVector::orWith(const BitVector & right)
 {
     verifyContains(*this, right);
-    IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
 
+    if (right.size() < size()) {
+        ssize_t commonBytes = numActiveBytes(getStartIndex(), right.size()) - sizeof(Word);
+        if (commonBytes > 0) {
+            IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+        }
+        Index last(right.size() - 1);
+        getWordIndex(last)[0] |= (right.getWordIndex(last)[0] & ~endBits(last));
+    } else {
+        IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    }
     repairEnds();
     invalidateCachedCount();
 }
@@ -201,9 +210,13 @@ BitVector::andWith(const BitVector & right)
 {
     verifyContains(*this, right);
 
-    IAccelrated::getAccelrator().andBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    uint32_t commonBytes = std::min(getActiveBytes(), numActiveBytes(getStartIndex(), right.size()));
+    IAccelrated::getAccelrator().andBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+    if (right.size() < size()) {
+        clearInterval(right.size(), size());
+    }
 
-    setGuardBit();
+    repairEnds();
     invalidateCachedCount();
 }
 
@@ -213,9 +226,18 @@ BitVector::andNotWith(const BitVector& right)
 {
     verifyContains(*this, right);
 
-    IAccelrated::getAccelrator().andNotBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    if (right.size() < size()) {
+        ssize_t commonBytes = numActiveBytes(getStartIndex(), right.size()) - sizeof(Word);
+        if (commonBytes > 0) {
+            IAccelrated::getAccelrator().andNotBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+        }
+        Index last(right.size() - 1);
+        getWordIndex(last)[0] &= ~(right.getWordIndex(last)[0] & ~endBits(last));
+    } else {
+        IAccelrated::getAccelrator().andNotBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    }
 
-    setGuardBit();
+    repairEnds();
     invalidateCachedCount();
 }
 
--- a/searchlib/src/vespa/searchlib/common/bitvector.cpp
+++ b/searchlib/src/vespa/searchlib/common/bitvector.cpp
@@ -94,7 +94,7 @@ void
 BitVector::clearIntervalNoInvalidation(Range range_in)
 {
     Range range = sanitize(range_in);
-    if ( ! range.validNonZero()) return;
+    if ( ! range.validNonZero()) { return; }
 
     Index last = range.end() - 1;
     Index startw = wordNum(range.start());
@@ -113,7 +113,7 @@ void
 BitVector::setInterval(Index start_in, Index end_in)
 {
     Range range = sanitize(Range(start_in, end_in));
-    if ( ! range.validNonZero()) return;
+    if ( ! range.validNonZero()) { return; }
 
     Index last = range.end() - 1;
     Index startw = wordNum(range.start());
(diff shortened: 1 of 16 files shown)


searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -178,8 +178,17 @@ void
 BitVector::orWith(const BitVector & right)
 {
     verifyContains(*this, right);
-    IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
 
+    if (right.size() < size()) {
```
havardpe, line 182: This should check if the right one ends before this one, since the right one could start before this one and also end before this one while still having greater size (should also add unit test for this case).

havardpe (commented):
Please consider my comment and also let @toregge have a look

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -20,11 +20,11 @@ using vespalib::alloc::Alloc;
...
 namespace {
 
-void verifyContains(const search::BitVector & a, const search::BitVector & b) __attribute__((noinline));
+void verifyInclusiveStart(const search::BitVector & a, const search::BitVector & b) __attribute__((noinline));
 
-void verifyContains(const search::BitVector & a, const search::BitVector & b)
+void verifyInclusiveStart(const search::BitVector & a, const search::BitVector & b)
 {
-    if ((a.getStartIndex() < b.getStartIndex()) || (a.size() > b.size())) {
+    if (a.getStartIndex() < b.getStartIndex()) {
```
toregge, line 27: Consider adjusting exception message to match new semantics.

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -177,9 +177,18 @@ BitVector::countInterval(Index start, Index end) const
...
-    verifyContains(*this, right);
-    IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    verifyInclusiveStart(*this, right);
 
+    if (right.size() < size()) {
+        ssize_t commonBytes = numActiveBytes(getStartIndex(), right.size()) - sizeof(Word);
+        if (commonBytes > 0) {
+            IAccelrated::getAccelrator().orBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+        }
+        Index last(right.size() - 1);
```
toregge, line 187: If right.size() is 0 then we get a wraparound and bad memory accesses. Note that bitvector have other methods that fails in a similar manner when size() is 0, e.g. clearIntervalNoInvalidation(), setInterval(), countInterval().

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -199,23 +208,36 @@ BitVector::repairEnds()
...
 
-    IAccelrated::getAccelrator().andBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    uint32_t commonBytes = std::min(getActiveBytes(), numActiveBytes(getStartIndex(), right.size()));
+    IAccelrated::getAccelrator().andBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+    if (right.size() < size()) {
+        clearInterval(right.size(), size());
+    }
 
-    setGuardBit();
+    repairEnds();
```
toregge, line 219: Note that repairEnds() fails to set guard bit if size() is 0.

toregge (approved):

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -199,23 +208,36 @@ BitVector::repairEnds()
...
-    verifyContains(*this, right);
+    verifyInclusiveStart(*this, right);
 
-    IAccelrated::getAccelrator().andNotBit(getActiveStart(), right.getWordIndex(getStartIndex()), getActiveBytes());
+    if (right.size() < size()) {
+        ssize_t commonBytes = numActiveBytes(getStartIndex(), right.size()) - sizeof(Word);
+        if (commonBytes > 0) {
+            IAccelrated::getAccelrator().andNotBit(getActiveStart(), right.getWordIndex(getStartIndex()), commonBytes);
+        }
+        Index last(right.size() - 1);
```
toregge, line 234: Wraparound when right.size() is 0, followed by bad memory access.

searchlib/src/vespa/searchlib/common/bitvector.cpp
toregge: I don't think this is a problem, since size() includes the portion before getStartIndex().
getEndIndex() doesn't make sense, and should be removed.

baldersheim: That was my conclusion after looking at the details. Removed getEndIndex.

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Fixed

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Guard added

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Not guarding setGuardBit.

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Guard added

baldersheim:
@toregge PTAL

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -351,8 +351,8 @@ BitVector::create(Index start, Index end)
 BitVector::UP
 BitVector::create(const BitVector & org, Index start, Index end)
 {
-    return (start == 0)
-           ? create(end)
+    return ((start == 0) && (end == org.size()))
```
toregge, line 361: Consider also checking that org.getStartIndex() is zero or adjusting constructor for AllocatedBitVector to handle
that case.

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -91,9 +91,10 @@ BitVector::clearInterval(Index start, Index end)
 }
 
 void
-BitVector::clearIntervalNoInvalidation(Index start, Index end)
+BitVector::clearIntervalNoInvalidation(Index start_in, Index end)
 {
-    if (start >= end) { return; }
+    Index start = std::max(start_in, getStartIndex());
+    if (start >= end || end == 0 || size() == 0) { return; }
```
toregge, line 97: This can be simplified if end is set to std::min(end, size()) before checking for start >= end.
if end or size() is  0 then start >= end will evaluate to true, triggering the early return.

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Fixed using first option.

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Refactored range checking.

baldersheim:
@toregge PTAL

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -133,28 +133,28 @@ BitVector::setInterval(Index start_in, Index end)
...
 }
 
 BitVector::Index
-BitVector::countInterval(Index start_in, Index end) const
+BitVector::countInterval(Range range_in) const
 {
-    Index start = std::max(start_in, getStartIndex());
-    if (start >= end || end == 0 || size() == 0) { return 0; }
+    Range range = sanitize(range_in);
+    if ( ! range.validNonZero()) return 0;
```
toregge, line 143: Consider adding braces around return 0.

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -85,46 +85,46 @@ BitVector::clear()
...
 }
 
 void
-BitVector::setInterval(Index start_in, Index end)
+BitVector::setInterval(Index start_in, Index end_in)
 {
-    Index start = std::max(start_in, getStartIndex());
-    if (start >= end || end == 0 || size() == 0) { return; }
+    Range range = sanitize(Range(start_in, end_in));
+    if ( ! range.validNonZero()) return;
```
toregge, line 116: Consider adding braces around return statement.

searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -85,46 +85,46 @@ BitVector::clear()
...
 }
 
 void
-BitVector::clearIntervalNoInvalidation(Index start_in, Index end)
+BitVector::clearIntervalNoInvalidation(Range range_in)
 {
-    Index start = std::max(start_in, getStartIndex());
-    if (start >= end || end == 0 || size() == 0) { return; }
+    Range range = sanitize(range_in);
+    if ( ! range.validNonZero()) return;
```
toregge, line 97: Consider adding braces around return statement.

baldersheim: Done

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Done

searchlib/src/vespa/searchlib/common/bitvector.cpp
baldersheim: Done

PR #13186, opened 2020-05-08
Enable async as default

baldersheim:
@vekterli PR


vekterli (approved):
:+1: let's try it out

PR #13192, opened 2020-05-08
add set_global_filter API w/wiring and unit test

arnej27959:

@geirst please review
@havardpe @toregge FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
@@ -104,6 +104,11 @@ Blueprint::get_replacement()
     return Blueprint::UP();
 }
 
+void
+Blueprint::set_global_filter(std::shared_ptr<BitVector>)
+{
+}
+
 const Blueprint &
 Blueprint::root() const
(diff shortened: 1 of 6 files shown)


searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
```diff
@@ -365,6 +370,16 @@ IntermediateBlueprint::optimize(Blueprint* &self)
     maybe_eliminate_self(self, get_replacement());
 }
 
+void
+IntermediateBlueprint::set_global_filter(std::shared_ptr<BitVector> global_filter)
```
havardpe, line 374: We should consider using a wrapper class (ex GlobalFilter) that owns the bitvector and implements shared_from_this to avoid over-copying the shared pointer.

geirst (approved):
👍

searchlib/src/vespa/searchlib/queryeval/blueprint.cpp
arnej27959: good catch, we'll do that soon.

PR #13198, opened 2020-05-08
Use a lock to ensure it is thread safe.

baldersheim:
@toregge or @vekterli PR

### DIFF
--- a/storage/src/tests/persistence/common/persistenceproviderwrapper.h
+++ b/storage/src/tests/persistence/common/persistenceproviderwrapper.h
@@ -19,6 +19,7 @@
 #include <vespa/persistence/spi/abstractpersistenceprovider.h>
 #include <vector>
 #include <string>
+#include <mutex>
 
 namespace storage {
 
@@ -50,8 +51,10 @@ public:
 private:
     spi::PersistenceProvider& _spi;
     spi::Result _result;
+    mutable std::mutex  _lock;
     mutable std::vector<std::string> _log;
     uint32_t _failureMask;
+    using Guard = std::unique_lock<std::mutex>;
 public:
     PersistenceProviderWrapper(spi::PersistenceProvider& spi);
     ~PersistenceProviderWrapper() override;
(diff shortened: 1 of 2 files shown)


storage/src/tests/persistence/common/persistenceproviderwrapper.h
```diff
@@ -50,8 +51,10 @@ class PersistenceProviderWrapper : public spi::AbstractPersistenceProvider
 private:
     spi::PersistenceProvider& _spi;
     spi::Result _result;
+    mutable std::mutex  _lock;
     mutable std::vector<std::string> _log;
     uint32_t _failureMask;
+    using Guard = std::unique_lock<std::mutex>;
```
toregge, line 57: std::lock_guard&lt;std::mutex&gt; should be sufficient here.

storage/src/tests/persistence/common/persistenceproviderwrapper.h
```diff
@@ -61,12 +64,16 @@ class PersistenceProviderWrapper : public spi::AbstractPersistenceProvider
      * return the given error without the wrapped SPI ever being invoked.
      */
     void setResult(const spi::Result& result) {
+        Guard guard(_lock);
         _result = result;
     }
     void clearResult() {
         _result = spi::Result(spi::Result::ErrorType::NONE, "");
     }
-    const spi::Result& getResult() const { return _result; }
```
toregge, line 69: clearResult() above doesn't seem to be used. Consider removing it.

toregge (approved):

storage/src/tests/persistence/common/persistenceproviderwrapper.h
baldersheim: Thanks, fixed

storage/src/tests/persistence/common/persistenceproviderwrapper.h
baldersheim: Good catch, done.

vekterli (commented):
:+1:

PR #13201, opened 2020-05-09
Avoid filling log with unused metrics events.

baldersheim:
@vekterli or @toregge PR


toregge (approved):

vekterli (approved):
:+1: in the age before good centralized metric collection this was a fairly decent poor man's solution for finding out what has happened historically on a node. But these days it's not needed anymore.

PR #13202, opened 2020-05-10
Try to hide implementation.

baldersheim:
@toregge PR There are no semantic changes here. Just formating and hiding of data members to clarify what is changed where.
@geirst FYI

Is FakeEGCompr64FilterOcc in use ? It noticed a commend about old format.
It is doing some dirty buffer stealing that would would like to avoid. I noticed a comment about old format.
/*
 * Old compressed posocc format.
 */
class FakeEGCompr64FilterOcc


toregge:
FakeEGCompr64FilterOcc is used by `postinglistbm` and `posting_list_test` and contains some posting lists formats with bitwise encoding of document id deltas and skip information. These formats are slower than the formats using byte based encoding of document id deltas and  skip information and are thus not used elsewhere.

toregge (approved):

PR #13204, opened 2020-05-10
Control num response threads with flag

baldersheim:
@vekterli or @hmusum PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/FileStorProducer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/FileStorProducer.java
@@ -11,8 +11,8 @@ import com.yahoo.vespa.model.content.cluster.ContentCluster;
 public class FileStorProducer implements StorFilestorConfig.Producer {
 
     public static class Builder {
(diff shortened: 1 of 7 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/FileStorProducer.java
```diff
@@ -11,8 +11,8 @@
 public class FileStorProducer implements StorFilestorConfig.Producer {
 
     public static class Builder {
-        protected FileStorProducer build(ContentCluster parent, ModelElement clusterElem) {
-            return new FileStorProducer(parent, getThreads(clusterElem));
+        protected FileStorProducer build(ContentCluster parent, Integer numResonseThreads, ModelElement clusterElem) {
```
hmusum, line 14: typo in variable name

hmusum (approved):

config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/FileStorProducer.java
baldersheim: Fixed

PR #13208, opened 2020-05-11
allow filter in HNSW index

arnej27959:

For discussion: we can reduce some code duplication by sending the filter by pointer, optionally nullptr, and testing for that in the appropriate place - what do you think?

missing for now: unit test

### DIFF
--- a/searchlib/src/tests/attribute/tensorattribute/tensorattribute_test.cpp
+++ b/searchlib/src/tests/attribute/tensorattribute/tensorattribute_test.cpp
@@ -186,6 +186,16 @@ public:
         (void) explore_k;
         return std::vector<Neighbor>();
     }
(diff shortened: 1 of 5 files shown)


geirst (commented):
Yes, I think we should try to avoid as much code duplication as possible, collapsing search_l0(),  find_top_k_with_filter() and top_k_candidates() (that takes filter) into generic ones that can optionally handle the filter. Passing the filter as a pointer to the underlying functions sounds good, and we can check whether to use it or not (nullptr indicating we are not using it). As calculation of distance is the costly part of searching, we will probably not notice these extra checks. If we observe it on profiling we can template the functions on whether to use filter or not. The new function on the index API seems good, there it makes sense to have an explicit function for using filter.

arnej27959:
@geirst PTAL - now with unit test and reduced code duplication

geirst (approved):
👍 I don't see the unit test, did you forget to push it?

geirst (approved):
👍

PR #13209, opened 2020-05-11
actually call set_global_filter (with empty SP for now)

arnej27959:

for discussion: i think this is the place to call set_global_filter, but we need to somehow wire in the whitelist bitvector (for now) as the global filter.

@geirst @toregge


geirst (approved):
👍

PR #13220, opened 2020-05-11
Add first 200 bytes of message to xml exception 

baldersheim:
@bratseth PR
@yngveaasheim FYI

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedReaderFactory.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedReaderFactory.java
@@ -3,10 +3,12 @@ package com.yahoo.vespa.http.server;
 
 import com.yahoo.document.DocumentTypeManager;
 import com.yahoo.document.json.JsonFeedReader;
+import com.yahoo.text.Utf8;
 import com.yahoo.vespa.http.client.config.FeedParams;
 import com.yahoo.vespaxmlparser.FeedReader;
 import com.yahoo.vespaxmlparser.VespaXMLFeedReader;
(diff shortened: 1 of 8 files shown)


baldersheim:
@bjorncs PR

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedReaderFactory.java
```diff
@@ -31,7 +33,14 @@ public FeedReader createReader(
...
                 } catch (Exception e) {
-                    throw new RuntimeException("Could not create VespaXMLFeedReader", e);
+                    byte [] peek;
+                    try {
+                        peek = new byte[Math.min(200, inputStream.available())];
+                        inputStream.read(peek);
+                    } catch (IOException  io) {
+                        peek = new byte [0];
+                    }
+                    throw new RuntimeException("Could not create VespaXMLFeedReader. First characters are: " + Utf8.toString(peek), e);
```
bjorncs, line 43: Will the bytes contain the actual first bytes? I would assume that the `VespaXMLFeedReader` consumes some bytes before throwing the exception.

bjorncs (approved):

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedReaderFactory.java
baldersheim: You are right. I added mark/reset support and made it default to peek the first 200 bytes.
Will consider if this should become a feature flag.

PR #13222, opened 2020-05-12
add default false require-all-up feature

kkraune:
FYI @yngveaasheim @bratseth 

Add a config for requiring all content clusters to be up for container to be in rotation

I assume adding a new config param is safe wrt deployments

### DIFF
--- a/container-core/src/main/resources/configdefinitions/vip-status.def
+++ b/container-core/src/main/resources/configdefinitions/vip-status.def
@@ -10,3 +10,6 @@ statusfile string default="share/qrsdocs/status.html"
 
 ## Not used TODO: Remove on Vespa 8
 initiallyInRotation bool default=true
+
+## Require all clusters UP to set container UP
+requireAllUp bool default=false
--- a/container-core/src/main/java/com/yahoo/container/handler/VipStatus.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/VipStatus.java
@@ -26,6 +26,9 @@ public class VipStatus {
 
     private final boolean initiallyInRotation;
 
+    /** Override default, require all content clusters up */
+    private final boolean requireAllUp;
+
     /** The current state of this */
     private boolean currentlyInRotation;
 
@@ -59,6 +62,7 @@ public class VipStatus {
         this.clustersStatus = clustersStatus;
         this.healthState = healthState;
         initiallyInRotation = vipStatusConfig.initiallyInRotation();
+        requireAllUp = vipStatusConfig.requireAllUp();
         healthState.status(StateMonitor.Status.initializing);
         clustersStatus.setContainerHasClusters(! dispatchers.searchcluster().isEmpty());
         updateCurrentlyInRotation();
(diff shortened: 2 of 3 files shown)


baldersheim:
I think it is better that @bratseth takes this one.

yngveaasheim:
Ideally this should be a per cluster config, indicating wether the cluster is mandatory or not.

container-core/src/main/resources/configdefinitions/vip-status.def
```diff
@@ -10,3 +10,6 @@ statusfile string default="share/qrsdocs/status.html"
 
 ## Not used TODO: Remove on Vespa 8
 initiallyInRotation bool default=true
+
+## Require all clusters UP to set container UP
+requireAllUp bool default=false
```
baldersheim, line 15: Is this a public config ? If so we should be very careful about it. Remember that we are not able to give any warnings or deprecations or anything to users ahead of time.

I would guess that we need a a list of mandatory clusters and a matchtype
require {ANY, ALL}

I think that this requires an architect meeting.

bratseth:
This looks fine. I agree it would be even better to have a setting per cluster but it is also quite a bit more work and I suggest it is not worth it and we should just merge this and move on.

bratseth (approved):

kkraune:
yes, the per-cluster config is only useful for apps with 3 or more clusters - and one can configure "Non-essential Providers" to cover some of the use cases relevant here. so I think this is good enough.

question is, do we want to expose this in services.xml or not? Agree we should have a review of config then, in this case not so sure, but people can feed back on name used maybe

baldersheim:
We shall not make a config that we want to expose to users.
 There MUST be services.xml control of it.

kkraune:
ok. I think the VipStatusHandler is set up if configuring &lt;search/&gt; in a container, so this config should then got somewhere in https://docs.vespa.ai/documentation/reference/services-search.html - like 

&lt;search&gt;
    &lt;status require-all-up="true" file="/tmp/status.html" /&gt;

https://github.com/vespa-engine/vespa/blob/master/container-core/src/main/resources/configdefinitions/vip-status.def has two other relevant parameters, accessdisk and statusfile that should be added while at it, disregarding unused initiallyInRotation

I don't really get the need for accessfile - I think the existence of statusfile is sufficient, but maybe I did not get all the details.

regardless, the services.xml changes could go in a different PR, please feed back on config suggestion above then

bratseth:
In general, I don't agree to that. It's not like we're check mate if we want to deprecate config, we can add a warn-on-deprecation mechanism whenever we want.

yngveaasheim:
@kkraune A per-cluster config is useful also for applications with two content clusters where one of them is mandatory.

baldersheim:
If we let config loose then we are doomed. When we decide that we did need some logic and some services.xml syntax for it we are not able to solve the gory details in the config-model. We can then either choose to ignore what is stated in config overrides, or we need to add logic where we use the config. We have had plenty of issues with this in the backend. That is still a mess as we have not cleaned out all config usage around.

Aside from the general experience based hostility towards anything but generated config,
I also think that this is should be on a per cluster basis.
A list of mandatory clusters, or a mandatory tag on the clusters.

bratseth:
On the former: Adding every config to another layer just to pass it through is also a lot of work, and nobody's taking on this work for all the existing config.

On the latter: Yes, it's better but it's not clear to me why we should do it now Who will use it/why should we prioritize making that effort?

yngveaasheim:
> On the latter: Yes, it's better but it's not clear to me why we should do it now Who will use it/why should we prioritize making that effort?
I believe that for >95% of use-cases there is one content cluster that must be up for the result page to have value for an end user. Hence "all" applications with more than one content cluster  want this feature. That said, it is likely not much more important today than it was last month.

kkraune:
> Hence "all" applications with more than one content cluster want this feature.

maybe. none has asked so far - this is a request from ourselves.

As I am not going to write the more complex per-cluster feature, I am going to clean up the documentation (it is incorrect), make it smaller and point to code instead. Teams can always write their own handler for healthchecks, based on their custom needs

yngveaasheim:
> maybe. none has asked so far - this is a request from ourselves.

You won't ask for what you don't realise that you need. And the cases where this would be useful are pretty rare.

kkraune:
true. I cleaned up documentation in https://github.com/vespa-engine/documentation/pull/975, hopefully this is good for now

PR #13223, opened 2020-05-12
Distribute application package to other config servers

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
@@ -44,7 +44,9 @@ import com.yahoo.vespa.config.server.tenant.EndpointCertificateMetadataSerialize
 import com.yahoo.vespa.config.server.tenant.EndpointCertificateMetadataStore;
 import com.yahoo.vespa.config.server.tenant.EndpointCertificateRetriever;
 import com.yahoo.vespa.curator.Curator;
+import com.yahoo.vespa.flags.BooleanFlag;
 import com.yahoo.vespa.flags.FlagSource;
+import com.yahoo.vespa.flags.Flags;
 import org.xml.sax.SAXException;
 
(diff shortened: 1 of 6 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
```diff
@@ -221,6 +223,16 @@ void checkTimeout(String step) {
             }
         }
 
+        void distributeApplicationPackage() {
+            FileRegistry fileRegistry = fileDistributionProvider.getFileRegistry();
+            FileReference fileReference = fileRegistry.addFile("");
```
gjoranv, line 234: What does the empty string mean?

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
```diff
@@ -112,6 +116,7 @@ public ConfigChangeActions prepare(SessionContext context, DeployLogger logger,
                                        Optional<ApplicationSet> currentActiveApplicationSet, Path tenantPath, 
                                        Instant now) {
         Preparation preparation = new Preparation(context, logger, params, currentActiveApplicationSet, tenantPath);
+        preparation.distributeApplicationPackage();
```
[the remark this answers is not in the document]
hmusum: I think so, but it depends on how it will be used and deployed, as you say. My plan was to end up building it the same way on all config servers, but there's much to do before we are in that state. Something to keep in mind

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
hmusum: It's the relative path in the application package, so it basically means add "add everything in the application package" directory. Should probably add a new method to make this clearer, I'll look into it in the next PR.

PR #13227, opened 2020-05-12
Increase default feed concurrency from 0.4 to 0.5

baldersheim:
@geirst PR
This is the last step on the way from 0.2 to 0.5.


geirst (approved):
👍

PR #13228, opened 2020-05-12
 Remove the TenantCost interface and the tenant cost API 

oyving:
* Removed the tenant cost REST API as I don't think it's in use any longer
* Removed the TenantCost interface, merged it with other components


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -207,8 +207,6 @@ private HttpResponse handleGET(Path path, HttpRequest request) {
         if (path.matches("/application/v4/")) return root(request);
         if (path.matches("/application/v4/tenant")) return tenants(request);
         if (path.matches("/application/v4/tenant/{tenant}")) return tenant(path.get("tenant"), request);
-        if (path.matches("/application/v4/tenant/{tenant}/cost")) return tenantCost(path.get("tenant"), request);
-        if (path.matches("/application/v4/tenant/{tenant}/cost/{month}")) return tenantCost(path.get("tenant"), path.get("month"), request);
```
olaaun, line 211: These paths should also be removed from `com.yahoo.vespa.hosted.controller.api.role.PathGroup`

olaaun (approved):

PR #13230, opened 2020-05-12
Use syncronized to make the loadbalancer semantically thread safe.

baldersheim:
Use a cache to avoid parsing a string to get an index.
Move test to same package to avoid public access to internal details.

@vekterli PR
@yngveaasheim FYI

### DIFF
--- a/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/LoadBalancer.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/LoadBalancer.java
@@ -56,12 +56,7 @@ public class LoadBalancer {
         }
     }
     private int getCachedIndex(String nodeName) {
-        Integer index = cachedIndex.get(nodeName);
-        if (index == null) {
-            index = getIndex(nodeName);
-            cachedIndex.put(nodeName, index);
-        }
-        return index;
+        return cachedIndex.computeIfAbsent(nodeName, key -> getIndex(key));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/LoadBalancer.java
```diff
@@ -68,24 +77,24 @@ public Node getRecipient(List<Mirror.Entry> choices) {
...
+                NodeMetrics nodeMetrics = getNodeMetrics(entry);
 
-            weightSum += nodeMetrics.weight;
+                weightSum += nodeMetrics.weight;
 
-            if (weightSum > position) {
-                selectedNode = new Node(entry, nodeMetrics);
-                break;
+                if (weightSum > position) {
+                    selectedNode = new Node(entry, nodeMetrics);
```
vekterli, line 87: The returned `NodeMetrics` object is shared and thus may be read outside the lock whilst concurrently being updated by another thread. Is this a problem? Should we deep-copy the metrics to make sure they're internally consistent? This is not something introduced by this PR, but just a general observation.

vekterli (approved):
:+1: I need some time to digest the existing algorithm for this, but synchronization changes look good and also should make it easier to reason about the behavior in general.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/LoadBalancer.java
baldersheim: That was why I made it package private. I checked all other usage. That was only in tests.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/LoadBalancer.java
```diff
@@ -54,6 +55,14 @@ public int getIndex(String nodeName) {
             throw new IllegalArgumentException(err, e);
         }
     }
+    private int getCachedIndex(String nodeName) {
+        Integer index = cachedIndex.get(nodeName);
+        if (index == null) {
+            index = getIndex(nodeName);
+            cachedIndex.put(nodeName, index);
+        }
```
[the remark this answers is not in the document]
baldersheim: Done, I think the JIT compiler will solve the performance for us.

PR #13232, opened 2020-05-13
- Update metrics less often by removing the forceEventLogging alltoge…

baldersheim:
…ther.

- Let default bucket iteration work in smaller chunks with shorter waits.
@vekterli PR


vekterli (approved):
:+1: metrics code still too complicated for its own good...

PR #13239, opened 2020-05-13
Bjorncs/policy description

bjorncs:


security-utils/src/main/java/com/yahoo/security/tls/policy/PeerPolicy.java
```diff
@@ -48,12 +59,18 @@ public boolean equals(Object o) {
...
 
     @Override
     public int hashCode() {
-        return Objects.hash(policyName, assumedRoles, requiredCredentials);
+        return Objects.hash(policyName, description, assumedRoles, requiredCredentials);
+    }
+
+    @Override
+    public int compareTo(PeerPolicy o) {
+        return policyName.compareTo(o.policyName);
```
vekterli, line 74: Do we enforce policy name presence and/or uniqueness today? If not, what happens if someone adds multiple policies without setting names? Should this be a multi-dimensional comparison if this is an issue?

vekterli (commented):
Quick question about `compareTo` vs `equals` semantics to resolve first, otherwise LGTM

security-utils/src/main/java/com/yahoo/security/tls/policy/PeerPolicy.java
bjorncs: Both are enforced in the constructor of `AuthorizedPeer`.
I just discovered that I have to rewrite `compareTo` to be consistent with `equals`. `TreeSet.equals` is currently reporting false positives 😨

bjorncs: There is no simple way to implement `Comparable` to be "consistent with equals" (ref Javadoc for `Comparable`). I have removed the commit - it was really not necessary (only intended to improve assertion error messages in a related unit test).

bjorncs:
@vekterli PTAL

vekterli (approved):

PR #13244, opened 2020-05-13
Ensure that we hold the lock when calling mark_controller_as_having_o…

baldersheim:
…bserved_explicit_node_state.

@vekterli PR


vekterli (approved):
:+1:

PR #13255, opened 2020-05-14
add a special API to get a whitelist bitvector

arnej27959:

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/documentmetastore/white_list_provider.h
+++ b/searchcore/src/vespa/searchcore/proton/documentmetastore/white_list_provider.h
@@ -1,13 +1,16 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
-#include <vespa/searchlib/common/bitvector.h>
+#include <memory>
 
 #pragma once
 
+namespace search { class BitVector; }
+
 namespace proton::documentmetastore {
 
+/** Interface for fetching a copy of the white list bitvector */
 struct WhiteListProvider {
-    virtual search::BitVector::UP get_white_list_filter() const = 0;
+    virtual std::unique_ptr<search::BitVector> get_white_list_filter() const = 0;
 protected:
     ~WhiteListProvider() = default;
 };
--- a/searchcore/src/vespa/searchcore/proton/documentmetastore/white_list_provider.h
+++ b/searchcore/src/vespa/searchcore/proton/documentmetastore/white_list_provider.h
@@ -1,0 +1,15 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/searchlib/common/bitvector.h>
+
+#pragma once
+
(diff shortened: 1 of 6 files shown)


arnej27959:
one thing i don't know: do we need to use "invalidateCachedCount" on the bitvector?  because NN blueprint uses countTrueBits.

searchcore/src/vespa/searchcore/proton/documentmetastore/white_list_provider.h
```diff
@@ -0,0 +1,15 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/searchlib/common/bitvector.h>
+
+#pragma once
+
+namespace proton::documentmetastore {
+
+struct WhiteListProvider {
+    virtual search::BitVector::UP get_white_list_filter() const = 0;
```
toregge, line 10: Consider forward declaring search::BitVector and using std::unique_ptr<search::BitVector>.

searchcore/src/vespa/searchcore/proton/matching/query.cpp
```diff
@@ -156,6 +157,11 @@ void
 Query::setWhiteListBlueprint(Blueprint::UP whiteListBlueprint)
 {
     _whiteListBlueprint = std::move(whiteListBlueprint);
+    using proton::documentmetastore::WhiteListProvider;
+    auto wlf = dynamic_cast<WhiteListProvider *>(_whiteListBlueprint.get());
+    if (wlf) {
+        _global_white_list = wlf->get_white_list_filter();
```
toregge, line 163: This adds a cost (creating a copy of the white list bitvector) even when this copy won't be used.

toregge (changes requested):

searchcore/src/vespa/searchcore/proton/matching/query.cpp
geirst: Agree with Tor. Can we do this in optimize() instead, where we know we need it? In that case we need to store a pointer to WhiteListProvider in Query as the white list blueprint are moved into the blueprint tree between this call and the call to optimize().

geirst (commented):
We need to find a solution to avoid creating the global white list when we don't need it, otherwise LGTM.

searchcore/src/vespa/searchcore/proton/matching/query.h
```diff
@@ -19,11 +21,12 @@ class Query
 {
 private:
     using Blueprint=search::queryeval::Blueprint;
+    using WhiteListProvider=proton::documentmetastore::WhiteListProvider;
     search::query::Node::UP _query_tree;
     Blueprint::UP           _blueprint;
     search::fef::Location   _location;
     Blueprint::UP           _whiteListBlueprint;
-    search::BitVector::UP   _global_white_list;
+    WhiteListProvider      *_white_list_provider;
```
geirst, line 29: This could be const as get_white_list_filter() is const.

geirst (approved):
👍

toregge (approved):

PR #13265, opened 2020-05-15
Added hosted accountant role

olaaun:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
```diff
@@ -22,12 +22,12 @@
 
     /** Full access to everything. */
     operator(Privilege.grant(Action.all())
-                      .on(PathGroup.all())
+                      .on(PathGroup.allExcept(PathGroup.hostedAccountant))
```
smorgrav, line 25: Should we let operators and supports being able to view this data? Is that distinction easy to make? So all reads are ok for supporter but you have to be accountant to write?

smorgrav: @oyving I guess this is a product decision?

oyving: I agree we should open this up for reading for operators and supporters. It should be easy by splitting it up into two different `Privilege` statements.

smorgrav (approved):

PR #13273, opened 2020-05-18
Persist application roles

tokle:
@oyving fyi


configserver/src/main/java/com/yahoo/vespa/config/server/tenant/ApplicationRolesStore.java
```diff
@@ -0,0 +1,64 @@
...
+            Slime slime = new Slime();
+            ApplicationRolesSerializer.toSlime(applicationRoles, slime.setObject());
+            curator.set(applicationRolesPath(application), SlimeUtils.toJsonBytes(slime));
+        } catch (Exception e) {
+            throw new RuntimeException("Could not write application roles of " + application, e);
+        }
+    }
+
+    /** Returns a transaction which deletes application roles if they exist */
+    public CuratorTransaction delete(ApplicationId application) {
```
hmusum, line 55: This method is never called, it should be added in `ApplicationRepository.delete()`

hmusum (changes requested):

hmusum (approved):

PR #13274, opened 2020-05-18
Shuffle upgrades

mpolden:
* Avoids instances always upgrading in the same order
* Prevents clumping together upgrades of instances under the same tenant

@jonmv


jonmv (approved):
Hope no one relies on this :)

PR #13278, opened 2020-05-18
Skip rebalancing of app if it has been recently deployed.

hmusum:
Wait some time after deployment with doing rebalancing of node resources.

In CD we have seen that Rebalancer has changed nodes for short-lived applications while the test runs, making test runs unstable, since we assume that no such changes happen while running the tests. This means that more than one change to nodes for an app might happen in a short time, which is probably not what we want, so I think this change makes sense in general.


bratseth (approved):
LGTM (although I don't like fields in test methods ... first there are fields, then we get shared methods setting them up, and then everything becomes a mess where changing something for one test breaks ten others. This happens for testers too but then it's all explicit and so easier to fix.)

hmusum:
I agree that testers are better, https://github.com/vespa-engine/vespa/pull/13284 is an attempt to make up for being lazy ...

PR #13286, opened 2020-05-18
Handle missing history event

mpolden:
@freva


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
```diff
@@ -394,6 +394,7 @@ private Agent eventAgentFromSlime(Inspector eventAgentField) {
             case "ProvisionedExpirer" : return Agent.ProvisionedExpirer;
             case "ReservationExpirer" : return Agent.ReservationExpirer;
             case "DynamicProvisioningMaintainer" : return Agent.DynamicProvisioningMaintainer;
+            case "RetiringUpgrader" : return Agent.RetiringUpgrader;
```
freva, line 397: Must be added in https://github.com/vespa-engine/vespa/blob/master/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeHistory.java as well

freva (approved):

mpolden:
Java build passed.

PR #13287, opened 2020-05-18
Bratseth/autoscale groupsize

bratseth:
Here it is probably best to follow the commits.


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Limits.java
```diff
@@ -17,7 +17,7 @@
 
     private final ClusterResources min, max;
 
-    private Limits(ClusterResources min, ClusterResources max) {
+    public Limits(ClusterResources min, ClusterResources max) {
```
freva, line 20: Need to verify that either both arguments are `null` or neither and generalize `isEmpty()`

freva (approved):

PR #13289, opened 2020-05-18
Handle when an instance has only declared tests

jonmv:
@mpolden please review. 

An instance with only declared tests would not qualify as an upgrade target to the Upgrader, making such a pipeline stop. This fixes that by adding declared test jobs also when there aren't (yet, because the next instance with the production job can't upgrade before the current instance, with only the tests) any production jobs waiting tot have their change tested. Not super-pretty, but I couldn't find a better way.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/Upgrader.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/Upgrader.java
@@ -107,7 +107,7 @@ public class Upgrader extends ControllerMaintainer {
     /** Returns a list of all production application instances, except those which are pinned, which we should not manipulate here. */
     private InstanceList instances() {
         return InstanceList.from(controller().jobController().deploymentStatuses(ApplicationList.from(controller().applications().readable())))
-                           .withProductionDeployment()
+                           .withDeclaredJobs()
                            .unpinned();
     }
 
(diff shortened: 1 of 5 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/Upgrader.java
```diff
@@ -105,7 +104,7 @@ public void maintain() {
     /** Returns a list of all production application instances, except those which are pinned, which we should not manipulate here. */
     private InstanceList instances() {
         return InstanceList.from(controller().jobController().deploymentStatuses(ApplicationList.from(controller().applications().readable())))
-                           .withProductionDeployment()
+                           .withDeclaredJobs()
```
jonmv, line 107: This _should_ be safe ...

mpolden:
LGTM, but build failed:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project controller-server: Compilation failure
[ERROR] /source/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/InstanceList.java:[71,31] cannot find symbol
[ERROR]   symbol:   variable statuses
[ERROR]   location: class com.yahoo.vespa.hosted.controller.application.InstanceList
```

jonmv:
There is no reference to 'statuses' on line 71 ...

mpolden:
L68 I presume. Did you rebase on latest master?

jonmv:
Ah, thanks.

PR #13292, opened 2020-05-18
Shutdown docproc thread executors.

baldersheim:
@jonmv or @hmusum or @bratseth PR

### DIFF
--- a/docproc/src/main/java/com/yahoo/docproc/jdisc/DocumentProcessingHandler.java
+++ b/docproc/src/main/java/com/yahoo/docproc/jdisc/DocumentProcessingHandler.java
@@ -121,7 +121,8 @@ public class DocumentProcessingHandler extends AbstractRequestHandler {
 
     @Override
     protected void destroy() {
-        //threadPoolMap.values().forEach( pool -> pool.shutdown());  //calling shutdownNow() seems like a bit of an overkill
+        laterExecutor.shutdown();
+        docprocServiceRegistry.allComponents().forEach(docprocService -> docprocService.deconstruct());
     }
 
     public ComponentRegistry<DocprocService> getDocprocServiceRegistry() {


hmusum (approved):

bratseth (approved):

docproc/src/main/java/com/yahoo/docproc/jdisc/DocumentProcessingHandler.java
```diff
@@ -121,7 +121,8 @@ public DocumentProcessingHandler(ComponentRegistry<DocumentProcessor> documentPr
 
     @Override
     protected void destroy() {
-        //threadPoolMap.values().forEach( pool -> pool.shutdown());  //calling shutdownNow() seems like a bit of an overkill
+        laterExecutor.shutdown();
+        docprocServiceRegistry.allComponents().forEach(docprocService -> docprocService.deconstruct());
```
jonmv, line 125: I would have thought the one create by this would suffice?

baldersheim: I think that all are created by this one.

jonmv: But not by this instance? Wouldn’t this also destruktivt the next gen create by DI, or isn’t the registre shared?

jonmv: Ok, the registry is owned by the handler. All good, then.

baldersheim: Well, after sleeping on this one I am no longer certain about the life time about this one.
I looked at the ComponentRegistry and saw that it was freezeable and assumed it would be frozen once component graph was set up, and hence not being possible to unregister anything. But that assumption might be wrong. No tests failed due to this, but that might be due that the 60s grace period.
@gjoranv could you have a look at this one.

gjoranv: This will deconstruct all DocprocServices, regardless of whether they survive the next reconfig or not. The injected registry contains _all_ of them, no matter which generation they were created in.

baldersheim: That was the the creeping feeling I had.

jonmv: That was my objection yesterday, but it seems the registry is created in the constructor of this handler, not injected.

jonmv: The other registries are injected, but not the one used for the `DocprocService`es, which are the ones with the thread pools that weren't shut down.

gjoranv: Sorry, I looked at the wrong constructor. As @jonmv says, the registry is not injected. Without knowing the code in detail, it looks like all the DocprocServices are created by the handler instance. So this should be all good.

PR #13293, opened 2020-05-18
Reporting metrics from audit log with user as dimension

bjormel:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
@@ -78,15 +78,33 @@ public class MetricsReporter extends ControllerMaintainer {
 
     private void reportAuditLog() {
         AuditLog log = controller().auditLogger().readLog();
+        HashMap<String, HashMap<String, Integer>> metricCounts = new HashMap<>();
+
         for (AuditLog.Entry entry : log.entries()) {
             String[] resource = entry.resource().split("/");
-            String operationMetric;
-            if(resource.length > 1 && resource[1] != null) {
+            if((resource.length > 1) && (resource[1] != null)) {
                 String api = resource[1];
-                operationMetric = OPERATION_PREFIX + api;
-                Metric.Context context = metric.createContext(Map.of("operator", entry.principal()));
+                String operationMetric = OPERATION_PREFIX + api;
 
-                metric.add(operationMetric, 1, context);
+                if (metricCounts.containsKey(operationMetric)) {
+                    HashMap<String, Integer> dimension = metricCounts.get(operationMetric);
+                    if (dimension.containsKey(entry.principal())) {
+                        Integer count = dimension.get(entry.principal());
+                        dimension.replace(entry.principal(), ++count);
+                    } else {
+                        dimension.put(entry.principal(), 1);
+                    }
+
+                } else {
+                    HashMap<String, Integer> dimension = new HashMap<>();
+                    dimension.put(entry.principal(),1);
+                    metricCounts.put(operationMetric, dimension);
+                }
+            }
+        }
+        for (String operationMetric : metricCounts.keySet()) {
+            for (String userDimension : metricCounts.get(operationMetric).keySet()) {
+                metric.set(operationMetric, (metricCounts.get(operationMetric)).get(userDimension), metric.createContext(Map.of("operator", userDimension)));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


bjormel:
@mpolden this might be enough for the first pass? I got lost in the dimensions so I ended up with just one

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -71,6 +73,24 @@ public void maintain() {
         reportRemainingRotations();
         reportQueuedNameServiceRequests();
         reportInfrastructureUpgradeMetrics();
+        reportAuditLog();
+    }
+
+    private void reportAuditLog() {
+        final String OPERATOR = "operator";
```
mpolden, line 80: Variable should be lowercase, and `final` can be removed (we typically don't use final for local variables).

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -71,6 +73,24 @@ public void maintain() {
...
+
+        AuditLog log = controller().auditLogger().readLog();
+        for (AuditLog.Entry entry : log.entries()) {
+            String[] resource = entry.resource().split("/");
+            String operationMetric;
+            if(resource[1] != null) {
```
mpolden, line 86: Array length should be checked.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -71,6 +73,24 @@ public void maintain() {
...
+                String api = resource[1];
+                operationMetric = OPERATION_PREFIX + api;
+                Metric.Context context = metric.createContext(Map.of(OPERATOR, entry.principal()));
+
+                metric.add(operationMetric, 1, context);
```
mpolden, line 91: Since `MetricsReporter` runs at an interval and this loops over all entries every time, this metric will end up counting the same entries multiple times.

I suggest that you instead calculate the total count and use `metric.set`. You can verify that the metric is as expected by calling `metricsReporter.maintain()` more than once in the test.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
```diff
@@ -43,6 +44,34 @@
...
+                "http://localhost:8080/zone/v2/prod/us-west-1/nodes/v2/state/dirty/le04614.ostk.bm2.prod.gq1.yahoo.com",
+                com.yahoo.jdisc.http.HttpRequest.Method.PUT
+        );
+        req1.getJDiscRequest().setUserPrincipal(() -> "user.bjormel");
+        tester.controller().auditLogger().log((req1));
+        HttpRequest req2 = HttpRequest.createTestRequest(
+                "http://localhost:8080/routing/v1/inactive/tenant/partner_publishing/application/contentindexing/instance/default/environment/prod/region/us-west-1",
+                com.yahoo.jdisc.http.HttpRequest.Method.POST
+        );
+        req2.getJDiscRequest().setUserPrincipal(() -> "user.bjormel");
```
mpolden, line 64: Test should use fake hostnames, applications and users.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
```diff
@@ -43,6 +44,34 @@
...
+        tester.controller().auditLogger().log((req2));
+
+        // End data to audit log
+        metricsReporter.maintain();
+        getMetric(MetricsReporter.OPERATION_PREFIX + "zone", "user.bjormel");
+        getMetric(MetricsReporter.OPERATION_PREFIX + "routing", "user.bjormel");
```
mpolden, line 70: This should assert the metric value.

mpolden (changes requested):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -78,15 +78,33 @@ public void maintain() {
...
+            if((resource.length > 1) && (resource[1] != null)) {
                 String api = resource[1];
-                operationMetric = OPERATION_PREFIX + api;
-                Metric.Context context = metric.createContext(Map.of("operator", entry.principal()));
-
-                metric.add(operationMetric, 1, context);
+                String operationMetric = OPERATION_PREFIX + api;
+
+                if (metricCounts.containsKey(operationMetric)) {
+                    HashMap<String, Integer> dimension = metricCounts.get(operationMetric);
```
mpolden, line 90: You can simplify this and drop a key lookup by always doing `metricCounts.get` and checking for `null` value  instead of the explicit `containsKey`. Same for the code below.

mpolden (approved):
👍

PR #13295, opened 2020-05-18
Set top-k-probability to 10 ppm

baldersheim:
@yngveaasheim PR
I will merge once we have fully confirmed performance


yngveaasheim (approved):
👍 Note: You use five 9s rather than four 9s as we are currently testing with.

baldersheim:
Clsoing as it is superceeded by #13313

PR #13304, opened 2020-05-19
Avoid data race from pending visitor send task

vekterli:
@geirst please review

It's currently possible for the visitor session to complete
even if there's a `SendCreateVisitors` task scheduled. This
will usually happen if there's been an error or if a sufficient
number of documents has been retrieved, triggering an early
exit of the session. In this case we must ensure that we do
not mutate any shared data structures from the send task when
it finally executes, as they may be read concurrently by the
session client thread.

The `done` variable is written under the same mutex as that held
by the send task, so visibility is guaranteed.

Also add `synchronized` to the binary serialization method for
`ProgressToken` to match the existing non-binary serialization
methods. This should not be required with the main race condition
fix, but is included for completion. Shall not break ABI compatibility.

### DIFF
--- a/documentapi/abi-spec.json
+++ b/documentapi/abi-spec.json
@@ -293,7 +293,7 @@
       "public void <init>(int)",
       "public void <init>(java.lang.String)",
       "public void <init>(byte[])",
-      "public byte[] serialize()",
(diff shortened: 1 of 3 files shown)


geirst (approved):
👍

PR #13311, opened 2020-05-19
Gjoranv/create registry from config

gjoranv:
@bratseth, please review the new constructor code carefully.

Next, we can have this injected to SearchHandler instead of creating a new registry there.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/profile/compiled/CompiledQueryProfileRegistry.java
+++ b/container-search/src/main/java/com/yahoo/search/query/profile/compiled/CompiledQueryProfileRegistry.java
@@ -1,9 +1,14 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.search.query.profile.compiled;
 
+import com.google.inject.Inject;
 import com.yahoo.component.ComponentSpecification;
 import com.yahoo.component.provider.ComponentRegistry;
-import com.yahoo.search.query.profile.types.QueryProfileType;
+import com.yahoo.search.query.profile.QueryProfile;
+import com.yahoo.search.query.profile.QueryProfileCompiler;
+import com.yahoo.search.query.profile.QueryProfileRegistry;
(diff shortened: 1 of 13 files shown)


baldersheim (changes requested):
Looks good to me. I would just include the change in SearchHandler to avoid blink-dpa being the first one using it.

baldersheim (approved):

gjoranv:
Done, but that was a _painful_ reminder about the integration style tests in container-search. These were written before the DI framework, but were adapted to it by converting the test files for configs that are long gone now (handlers and chains) into the new `components` config.

IMO, these tests should be rewritten to plain unit tests, and some should just be removed. No need to test DI here anymore.

baldersheim (approved):
LGTM. Sometimes pain is necessary go get going :)

container-search/src/main/java/com/yahoo/search/query/profile/compiled/CompiledQueryProfileRegistry.java
```diff
@@ -18,6 +23,15 @@
     
     private final QueryProfileTypeRegistry typeRegistry;
 
+    @Inject
+    public CompiledQueryProfileRegistry(QueryProfilesConfig config) {
+        QueryProfileRegistry registry = QueryProfileConfigurer.createFromConfig(config);
+        typeRegistry = registry.getTypeRegistry();
+        for (QueryProfile inputProfile : registry.allComponents()) {
```
bratseth, line 30: Why not just do 
QueryProfileCompiler.compile(registry)
here?

bratseth (approved):
LGTM but what is it that you want to achieve, recompiling less?

container-search/src/main/java/com/yahoo/search/query/profile/compiled/CompiledQueryProfileRegistry.java
gjoranv: I didn't want to create a temporary object of the same type as the class itself. There is room for improvement, but it requires creating even more public methods and/or constructors.

PR #13315, opened 2020-05-20
change splunkd startup logic

arnej27959:
* doing "start" then "restart" seems to trigger splunkd core dumps,
  so we will try "stop" + wait + "start" instead.
* log logforwarder actions on info level to make it easier to
  track splunk activity.

### DIFF
--- a/logforwarder/src/apps/vespa-logforwarder-start/child-handler.cpp
+++ b/logforwarder/src/apps/vespa-logforwarder-start/child-handler.cpp
@@ -27,7 +27,7 @@ runSplunk(const vespalib::string &prefix, std::vector<const char *> args)
         dbg.append(arg);
         dbg.append("'");
     }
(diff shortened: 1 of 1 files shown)


logforwarder/src/apps/vespa-logforwarder-start/child-handler.cpp
```diff
@@ -76,14 +76,15 @@ void
 ChildHandler::startChild(const vespalib::string &prefix)
 {
     if (! _childRunning) {
+        // it is possible that splunk was already running anyway, so
+        // make sure we restart it to get new config activated:
+        runSplunk(prefix, {"stop"});
+        sleep(1);
```
jonmv, line 82: It looked from the logs like stopping `splunkd` would take roughly 15s. I don't know what happens if we don't wait for the previous instance to shut down, but it's probably OK, since we don't know otherwise by now.

jonmv (approved):

PR #13320, opened 2020-05-20
Default dispatch policy is now adapative meaning that it will loadbal…

baldersheim:
…ance groups based on latency.

@hmusum PR
@yngveaasheim FYI


hmusum:
Should you change default value in dispatch.def as well?

baldersheim:
Yes, that is intentional.

hmusum (approved):

PR #13323, opened 2020-05-20
Specify round-robin dispatch policy explisit as default has changed t…

baldersheim:
…o adaptive.

@hmusum or @bratseth PR
I did not notice the red build on #13320


hmusum:
👍

PR #13325, opened 2020-05-20
Revert "Gjoranv/create registry from config"

hmusum:
Reverts vespa-engine/vespa#13311

`java.lang.IllegalArgumentException: Could not create a component with id 'com.yahoo.search.query.profile.compiled.CompiledQueryProfileRegistry'. Tried to load class directly, since no bundle was found for spec: com.yahoo.search.query.profile.compiled.CompiledQueryProfileRegistry. If a bundle with the same name is installed, there is a either a version mismatch or the installed bundle's version contains a qualifier string. `


bratseth (approved):

gjoranv:
Thanks, I missed some class -> bundle mapping magic.

PR #13332, opened 2020-05-20
Hmusum/use remote session repo for finding active session

hmusum:
This is a fix for issues seen where we do e.g. do not restart nodes when there are changes requiring it (we do not compare with previous deployment, since we do not find it).

(First commit is from https://github.com/vespa-engine/vespa/pull/13326)

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
@@ -1,11 +1,15 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.config.server.session;
 
+import com.yahoo.component.Version;
+import com.yahoo.config.application.api.ApplicationMetaData;
 import com.yahoo.config.application.api.ApplicationPackage;
 import com.yahoo.config.provision.AllocatedHosts;
+import com.yahoo.config.provision.ApplicationId;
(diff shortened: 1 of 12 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
```diff
@@ -135,4 +139,34 @@ public void delete() {
...
+        return createSetStatusTransaction(Status.DEACTIVATE);
+    }
+
+    private Transaction createSetStatusTransaction(Status status) {
+        return zooKeeperClient.createWriteStatusTransaction(status);
+    }
+
+    public ApplicationMetaData getMetaData() {
+        return zooKeeperClient.loadApplicationPackage().getMetaData();
+    }
```
jonmv, line 169: Could probably remove most of these from LocalSession, if they're not to be used on that?

jonmv (commented):
So the idea is that the remote session is more up to date? Local sessions are set only when deployed locally, while remotes are set whenever deployed? If so, this is absolutely right, and horrible it's been wrong for so long :)

configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
hmusum: I think I removed those that could be removed. I want to do more refactoring in this area, but let's try this first.

hmusum:
Local session for one id exists only on one config server today, so, yes, that's the idea. I think this has been so all the time.

PR #13337, opened 2020-05-20
Revert "Hmusum/use remote session repo for finding active session"

hmusum:
Reverts vespa-engine/vespa#13332

In case the PR breaks something. Please review only,


hmusum:
#13332 failed when redeploying applications after an upgrade

PR #13341, opened 2020-05-22
The prometheus metrics returns duplicate HELP and TYPE lines.

takamabe:
The prometheus metrics returns duplicate HELP and TYPE lines.
We got an error when using prometheus input plugin for Telegraf to collect the metrics from the /prometheus/v1/values.

Error of Telegraf:
```
2020-05-17T23:45:00Z E! [inputs.prometheus]: Error in plugin: error reading metrics for http://localhost:19092/prometheus/v1/values?consumer=Vespa: reading text format failed: text format parsing error in line 125: second TYPE line for metric name "memory_virt", or TYPE reported after samples
```

We think there should be just one HELP and TYPE line per metric name.
Could you please check if there is any problems with this pull request.


gjoranv:
I verified this manually by adding the "cpu" metric to the DummyService.getMetrics, and visually inspecting the handler output before and after this change.

takamabe:
Thank you so much for the code review.

PR #13343, opened 2020-05-22
only throttle node failures when nodes are still in state "failed"

andreer:


smorgrav:
So simple, so elegant!

smorgrav (approved):

PR #13347, opened 2020-05-22
Factor out portions of field match computer to field match computer shared state.

toregge:
@havardpe : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/features/fieldmatch/computer.cpp
+++ b/searchlib/src/vespa/searchlib/features/fieldmatch/computer.cpp
@@ -1,8 +1,10 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "computer.h"
+#include "computer_shared_state.h"
 #include <vespa/searchlib/features/utils.h>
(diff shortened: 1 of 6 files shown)


searchlib/src/vespa/searchlib/features/fieldmatch/computer.cpp
```diff
@@ -15,55 +17,24 @@ using namespace search::fef;
...
-    _currentMetrics(this),
-    _finalMetrics(this),
-    _simpleMetrics(params),
-    _segments(),
-    _alternativeSegmentationsTried(0),
-    _cachedHits()
+Computer::Computer(const ComputerSharedState& shared_state, const PhraseSplitter& splitter)
+    : _shared_state(shared_state),
+      _splitter(splitter),
+      _fieldId(_shared_state.get_field_id()),
```
baldersheim, line 23: Any reason why you copy the fields in the shared state, and not use it where they are ?

baldersheim (approved):

searchlib/src/vespa/searchlib/features/fieldmatch/computer.cpp
toregge: To keep the data localized while also minimizing the size of the PR diff.

PR #13348, opened 2020-05-22
Bratseth/real resources in model

bratseth:


config-model/src/main/java/com/yahoo/config/model/provision/SingleNodeProvisioner.java
```diff
@@ -28,12 +32,21 @@
...
 
     public SingleNodeProvisioner(Flavor flavor) {
         host = new Host(HostName.getLocalhost());
-        this.hostSpec = new HostSpec(host.hostname(), host.aliases(), flavor);
+        /*
+        this.hostSpec = new HostSpec(host.hostname(),
+                                     List.of(),
+                                     flavor.resources(), flavor.resources(),
+                                     Optional.empty(), Optional.empty(), Optional.empty(), Optional.empty(), Optional.empty());
+         */
```
freva, line 45:
```suggestion
```

freva (approved):
Very nice! :100:

PR #13353, opened 2020-05-23
Revert "Bratseth/real resources in model"

hmusum:
Reverts vespa-engine/vespa#13348

Nodes are attempted to be failed out in cd (most/all nodes, I guess, throttling has kicked in) due to errors like:
`The number of CPUs from cpuinfo (48.0) is less than the minimum guaranteed by the flavor (65.3)`, seems like cpu speedup factor is used/not used in some calculations.

### DIFF
--- a/config-model/src/main/java/com/yahoo/config/model/provision/HostsXmlProvisioner.java
+++ b/config-model/src/main/java/com/yahoo/config/model/provision/HostsXmlProvisioner.java
@@ -7,7 +7,6 @@ import com.yahoo.vespa.model.container.Container;
 
 import java.io.Reader;
 import java.util.List;
-import java.util.Optional;
 
 /**
  * A host provisioner based on a hosts.xml file.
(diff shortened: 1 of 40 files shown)


freva (approved):

hmusum:
Note that due to the API changes in HostSpec cd and publiccd was broken for a long time after this PR was merged (since there existed config models calling the methods in HostSpec that were reverted) and required manual interventions in cd and publiccd to get back to the state it was before this. It would be good to keep those changes separate to avoid this happening again.

Also, 7.220.14 will probably be with us for many months, so `Remove after June 2020` or similar will be misleading at that time.

bratseth:
> It would be good to keep those changes separate to avoid this happening again.

Thanks. Yes that was the intention, but I assumed this tiny change was innocuous. @freva do you think it is it worth keeping the raw cpu count (not scaled for speed) around for this check? No big issue if so just don't want extra stuff for no good reason.

> Also, 7.220.14 will probably be with us for many months

Urk, yes excellent point. We'll have to go back to the more cumbersome practice of specifying the version number in these for the time being.

freva:
Host-admin code has access to `VespaFlavor`, so it could get it directly from there instead of REST API, then we shouldn't need to send the unscaled value to config server at all

bratseth:
Yes, ok. Done in https://git.ouroath.com/vespa/vespa-yahoo/pull/4539

PR #13355, opened 2020-05-24
Increase stack size when using address sanitizer.

toregge:
@vekterli : please review

### DIFF
--- a/searchlib/src/tests/query/streaming_query_large_test.cpp
+++ b/searchlib/src/tests/query/streaming_query_large_test.cpp
@@ -27,7 +27,11 @@ void setMaxStackSize(rlim_t maxStackSize)
 // a stack overflow if the stack usage increases.
 TEST("testveryLongQueryResultingInBug6850778") {
     const uint32_t NUMITEMS=20000;
+#ifdef __SANITIZE_ADDRESS__
+    setMaxStackSize(12 * 1024 * 1024);
+#else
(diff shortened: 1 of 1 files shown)


searchlib/src/tests/query/streaming_query_large_test.cpp
```diff
@@ -27,7 +27,11 @@ void setMaxStackSize(rlim_t maxStackSize)
 // a stack overflow if the stack usage increases.
 TEST("testveryLongQueryResultingInBug6850778") {
     const uint32_t NUMITEMS=20000;
+#ifdef __SANITIZE_ADDRESS__
+    setMaxStackSize(12 * 1024 * 1024);
+#else
     setMaxStackSize(4 * 1024 * 1024);
```
baldersheim, line 33: What is it that demands so large stack size ?
Is it recursion or large stack allocations ?
We do not use that in production anyway.

baldersheim (approved):

searchlib/src/tests/query/streaming_query_large_test.cpp
vekterli: AddressSanitizer will instrument stack allocations, presumably to have sentinel zones for catching out of bounds writes. According to https://clang.llvm.org/docs/AddressSanitizer.html "AddressSanitizer uses more stack memory. We have seen up to 3x increase."

vekterli (approved):
:+1:

PR #13359, opened 2020-05-25
Delete expired non-active sessions

hmusum:
Due to a bug fixed in https://github.com/vespa-engine/vespa/pull/13338
we ended up with sessions that are in state ACTIVATED in zookeeper,
but that was not the latest active session.  These would not be
cleaned up, the changes here will delete these sessions (if they are
expired and not the current active session for an application).

Also deletes a test that was covered by tests in ApplicationRepositoryTest.


hakonhall (approved):

hakonhall:
This could potentially be a lot of data and there may be a need to partition the deletion like was done with the old sessions flag?

hmusum:
Yes, that might be a good idea, I'll look into it

hakonhall (approved):

hmusum:
@hakonhall please take a look

PR #13362, opened 2020-05-25
Include instances with only tests in confidence computation if app ha…

jonmv:
…s other prod deployment

@mpolden please review.

The idea is that we can use the tests for one of our own applications, which will have _good_ tests, to verify new releases, but without upgrading the production instances of the application too early.


mpolden:
👍

PR #13368, opened 2020-05-25
Detect use of clang version of address sanitizer.

toregge:
@vekterli : please review


vekterli (approved):
:+1: later on maybe we could add our own higher level definition set by CMake that works across both compilers

PR #13371, opened 2020-05-25
Move fastlib testsuite into namespace fast::testsuite.

toregge:
@vekterli : please review

This prevents address sanitizer for clang from complaining about odr-violations.


vekterli (approved):
:+1:

PR #13376, opened 2020-05-26
extend blueprint API

arnej27959:

@havardpe @geirst @toregge for consideration - does this look reasonable?
Note: we cannot re-use TrueSearch, it is only for unit tests.

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.h
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.h
@@ -164,7 +164,11 @@ public:
         virtual bool check(const Blueprint & bp) const = 0;
     };
 
-    enum class FilterBound { UPPER, LOWER };
+    // Signal if createFilterSearch should ensure the returned
+    // iterator is an upper bound (yielding a hit on at least
+    // all matching documents) or a lower bound (never yielding a
+    // hit that isn't certain to be a match).
+    enum class FilterConstraint { UPPER_BOUND, LOWER_BOUND };
 
     Blueprint();
     Blueprint(const Blueprint &) = delete;
@@ -201,7 +205,7 @@ public:
     bool frozen() const { return _frozen; }
 
     virtual SearchIteratorUP createSearch(fef::MatchData &md, bool strict) const = 0;
-    virtual SearchIteratorUP createFilterSearch(bool strict, FilterBound bound) const;
+    virtual SearchIteratorUP createFilterSearch(bool strict, FilterConstraint constraint) const;
 
     // for debug dumping
     vespalib::string asString() const;
--- a/searchlib/src/vespa/searchlib/queryeval/blueprint.h
+++ b/searchlib/src/vespa/searchlib/queryeval/blueprint.h
@@ -164,6 +164,8 @@ public:
         virtual bool check(const Blueprint & bp) const = 0;
     };
 
+    enum class FilterBound { UPPER, LOWER };
+
     Blueprint();
     Blueprint(const Blueprint &) = delete;
     Blueprint &operator=(const Blueprint &) = delete;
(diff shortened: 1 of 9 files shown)


searchlib/src/vespa/searchlib/queryeval/blueprint.h
```diff
@@ -164,6 +164,8 @@ class Blueprint
         virtual bool check(const Blueprint & bp) const = 0;
     };
 
+    enum class FilterBound { UPPER, LOWER };
```
havardpe, line 167: I think the 'bound' part should be part of the enum values and not the enum type, since it smells like past tense 'bind' on its own. Maybe FilterType or FilterConstraint, with UPPER_BOUND and LOWER_BOUND

searchlib/src/vespa/searchlib/queryeval/full_search.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    // nop
+}
+
+BitVector::UP
+FullSearch::get_hits(uint32_t begin_id)
+{
+    auto result = BitVector::create(begin_id, getEndId());
+    result->setInterval(begin_id, getEndId());
+    result->invalidateCachedCount();
```
havardpe, line 36: setInterval also invalidates the cached count

searchlib/src/vespa/searchlib/queryeval/full_search.cpp
```diff
@@ -0,0 +1,46 @@
...
+void
+FullSearch::doSeek(uint32_t docid)
+{
+    setDocId(docid);
+}
+
+void
+FullSearch::doUnpack(uint32_t)
+{
+    abort();
```
havardpe, line 16: I think empty is better than aborting

havardpe (commented):
EmptySearch seems to be lacking, since it claims to be strict but refuses to update the docid in doSeek. Doing setAtEnd in doSeek should be sufficient and seems the simplest (the alternative is to do it in init_range).

havardpe:
Also; a big question is whether we could get rid of the strict parameter to the createFilterSearch function...

searchlib/src/vespa/searchlib/queryeval/blueprint.h
geirst: Also add a comment describing the difference between the two.

searchlib/src/vespa/searchlib/queryeval/full_search.h
```diff
@@ -0,0 +1,25 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include "searchiterator.h"
+#include <vespa/searchlib/common/bitvector.h>
+
+namespace search::queryeval {
+
+class FullSearch : public SearchIterator
```
geirst, line 10: Please add class comment.

geirst (commented):
Overall looks good.

arnej27959:
EmptySearch does indeed do "setAtEnd" in initRange, I did not change that.
Please take another look now...

geirst (approved):
👍

searchlib/src/vespa/searchlib/queryeval/emptysearch.cpp
```diff
@@ -15,6 +14,26 @@ EmptySearch::doUnpack(uint32_t)
...
+EmptySearch::and_hits_into(BitVector &result, uint32_t begin_id)
+{
+    result.clearInterval(begin_id, getEndId());
+}
+
+BitVector::UP
+EmptySearch::get_hits(uint32_t begin_id)
+{
+    auto result = BitVector::create(begin_id, getEndId());
+    result->clearInterval(begin_id, getEndId());
```
havardpe, line 33: I still do not think you need to clear an empty bitvector

havardpe:
> EmptySearch does indeed do "setAtEnd" in initRange, I did not change that.
> Please take another look now...

I am blind :-p

Issue #13356, opened 2020-05-25
DELETE request of document API may not be reflected when one of replica nodes is initializing


riida:
Hi.

We have a question about DELETE operation of document API.

## Environment and Setting
- Vespa version is 7.207.47
  - [vespaengine/vespa:7.207.47](https://hub.docker.com/layers/vespaengine/vespa/7.207.47/images/sha256-42cdb695bdc490d3cc2fd4fe216ae09d6b005719cf5184e49d3859f05ecd16b8?context=explore)
- content redundancy is 2
- we check that `.idealstate.merge_bucket.pending` is 0 before stop content node(container)
- we set `Maximum storage transition time` is 20 minutes to avoid pending bucket in release
- document prune age was 2 weeks (default) in our setting

## Problem
We found DELETE operation of document API would not be reflected.
This seems to be happened only when one of replica nodes is initializing (vespa process is running but storage state is not up).
Applying partial update (UPDATE document API and `create: false`) to the document that DELETE operation would be not be reflected, the update was reflected.
This partial update was for the field `indexing: attribute` and `attribute: fast-search`.
The time series was as follows.

```
Container up (Storage Node State is Maintenance) -> DELETE Request -> Storage Node State is Up -> UPDATE Request

* these were in about 30 minutes
```

Document timestamp is the same as UPDATE Request time.
We checked access log and found that both DELETE and UPDATE request returned 200 (StatusOK).

## Question

We expect that the document will not be revival when partial update was applyed after DELETE request returns 200.
Is there any solution for this problem ?

Thanks.

jobergum:
>Applying partial update (UPDATE document API and create: false) to the document that DELETE >operation would be not be reflected, the update was reflected.

Could you please try to clarify this sentence, preferably with the sequence of operations, e.g time a send delete for doc x, time b, send update for doc x with create:false. Etc.

riida:
Thank you for your quick reply.

Our sequence of operations was as follows.

- [05/08/2020 20:27:06.326] [DELETE](https://docs.vespa.ai/documentation/reference/document-v1-api-reference.html#delete) request for `/document/v1/namespace/doctype/docid/1.hoge`
- [05/08/2020 20:46:19.723] [UPDATE](https://docs.vespa.ai/documentation/reference/document-v1-api-reference.html#update) request for `/document/v1/namespace/doctype/docid/1.hoge`
  - http method is PUT
  - send data-binary is like a below, create field is not specify actually (we expect default is false)
[...]

PR #13379, opened 2020-05-26
Properly handle tombstones during read and write repair

vekterli:

This alters the internal semantics Get-replies for removed documents to include the timestamp rather than always be zero. This lets higher level logic make explicit choices based on whether the document was not found or deleted (and thus possibly hiding older versions).

Get operation logic in the distributor has been changed to reflect this, allowing newer deletes to properly override older document versions. Since multi-phase updates use the Get operation logic internally, this also transitively fixes its behavior in this scenario.

Protocol serialization logic for tombstones is implemented in a forwards and backwards compatible way; old nodes will interpret tombstone replies as if they were "not found"-replies, as is the legacy behavior.

This is related to #13356


toregge (approved):

geirst (approved):
👍

PR #13380, opened 2020-05-26
Add hack for using same version as applications are pinned to

hmusum:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
@@ -194,7 +194,12 @@ public class InternalStepRunner implements StepRunner {
     }
 
     private Optional<RunStatus> deployTester(RunId id, DualLogger logger) {
-        Version platform = controller.systemVersion();
+        Version targetPlatform = controller.jobController().run(id).get().versions().targetPlatform();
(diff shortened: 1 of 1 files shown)


tokle (approved):

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -194,7 +194,12 @@ public InternalStepRunner(Controller controller) {
     }
 
     private Optional<RunStatus> deployTester(RunId id, DualLogger logger) {
-        Version platform = controller.systemVersion();
+        Version targetPlatform = controller.jobController().run(id).get().versions().targetPlatform();
+        // TODO: Hack to make sure some pinned applications get this version for tester deployments
+        //       Remove when no applications are pinned to this version anymore
+        final Version platform = targetPlatform.equals(Version.fromString("7.220.14"))
```
jonmv, line 200: There are issues both with using the target platform version, and with using the system version. Sometimes you want one, sometimes the other. We _could_ use system version if the application wasn't _pinned_ to a particular version, and that particular version if it was, to address this in a better way. Probably not the answer to everything either, but should solve this.

hmusum: Sounds like a pragmatic solution to me

PR #13378, opened 2020-05-26
Setup field match executor shared state.

toregge:
@geirst : please review
@baldersheim : FYI

PR #13381, opened 2020-05-26
Revert "Setup field match executor shared state."

hmusum:
Reverts vespa-engine/vespa#13378

About 15 system tests started failing with variants of this error:
`Feature 'fieldMatch(field24).matches' does not have expected score. Expected: 2.0 ~ 0.0001. Actual: 0.0`


hmusum:
@aressem FYI

baldersheim (approved):

PR #13384, opened 2020-05-26
Increase ccache max size to 1600M

hmusum:

### DIFF
--- a/travis/travis-build-full.sh
+++ b/travis/travis-build-full.sh
@@ -9,7 +9,7 @@ export MAVEN_OPTS="-Xss1m -Xms128m -Xmx2g"
 source /etc/profile.d/enable-devtoolset-9.sh
 source /etc/profile.d/enable-rh-maven35.sh
 
-ccache --max-size=1250M
+ccache --max-size=1600M
 ccache --set-config=compression=true
 ccache -p
 


toregge (approved):

travis/travis-build-full.sh
```diff
@@ -9,7 +9,7 @@ export MAVEN_OPTS="-Xss1m -Xms128m -Xmx2g"
 source /etc/profile.d/enable-devtoolset-9.sh
 source /etc/profile.d/enable-rh-maven35.sh
 
-ccache --max-size=1250M
+ccache --max-size=1600M
```
baldersheim, line 12: Ouch, I am using 20G on my laptop.
I would add ccahe --show-stats to see how the stats are And bump it to at least 2G.

hmusum: There is a `ccache --show-stats` on line 22, I guess that is what @toregge based his suggestion on

toregge: I performed an minor experiment in a CentOS 7 environment on my work laptop  by enabling compression, clearing the cache, and building vespa once. Afterwards ccache reported 1.3GB used, which is more than 1250 MB.  Some additional headroom is needed since we don't always compile files in the same order each time. But too much headroom slows down cache download/upload.

baldersheim: OK

PR #13385, opened 2020-05-26
Make zookeeper session cache a separate class

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
```diff
@@ -1,17 +1,17 @@
...
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.config.server.session;
 
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 
 /**
- * A generic session repository that can store any type of session that extends the abstract interface.
+ * A session cache that can store any type of session that extends the abstract interface.
```
gjoranv, line 9: We're probably talking about the Session class here, so I suggest to add a javadoc link to the class instead of the slightly confusing 'session that extends the abstract interface'.

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
hmusum: Fixed

PR #13387, opened 2020-05-27
Reduce expiry from 10 to 1 minute

tokle:
@jonmv / @mpolden: Do we still need this cache?

@bratseth fyi


jonmv:
Yes, we still want the cache. A single console page view has to pass through this multiple times.

PR #13394, opened 2020-05-27
Do not schedule NodeAgentContext if not waiting

freva:
See VESPA-18319


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextManager.java
```diff
@@ -76,7 +77,6 @@ public NodeAgentContext nextContext() throws InterruptedException {
             }
 
             currentContext = nextContext;
-            nextContext = null;
```
hakonhall, line 79: This meant that if NA spent >1 HA tick, the context it eventually picked up on its next NA tick was >1 HA tick old?

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextManager.java
freva: No, NA would pick up the context set during the last HA tick.

PR #13395, opened 2020-05-27
wrapper for iterators needing TFMDA in constructor

arnej27959:

To get started with the createFilterSearch() implementations, I looked at disktermblueprint, and found it could help to have a wrapper (based on BooleanMatchIteratorWrapper) that owns some unused TermFieldMatchData and make wiring in existing iterators easier.

Longer term we may want to avoid this by making a pure matching filter-only iterator on the lower levels directly, without any match data at all, but this can be used as a bridge for now.

@havardpe @toregge @geirst please take a look

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
+++ b/searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
@@ -1,0 +1,63 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include "searchiterator.h"
+#include "blueprint.h"
+#include <vespa/searchlib/fef/termfieldmatchdata.h>
+#include <vespa/searchlib/fef/termfieldmatchdataarray.h>
+
+namespace search::queryeval {
+
+/**
+ * Wraps an iterator for use as a filter search.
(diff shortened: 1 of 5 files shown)


geirst (approved):
👍

toregge (approved):

searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
```diff
@@ -0,0 +1,63 @@
...
+ * Owns TermFieldMatchData the wrapped iterator
+ * can wire to, and write to if necessary.
+ **/
+class FilterWrapper : public SearchIterator {
+private:
+    std::vector<fef::TermFieldMatchData> _unused_md;
+    fef::TermFieldMatchDataArray _tfmda;
+    std::unique_ptr<SearchIterator> _wrapped_search;
+public:
+    FilterWrapper(const Blueprint::State &state)
```
havardpe, line 23: prefer sending the number of fields directly instead of the state to reduces unneeded coupling

havardpe (commented):
looks good. should probably add some kind of test as well

geirst:
> looks good. should probably add some kind of test as well
I agree. We can add that in an upcoming PR as I depend on this for ongoing work in AttributeBlueprintFactory.

PR #13398, opened 2020-05-27
Add a test to show how topk affects as hits grows.

baldersheim:
@bratseth PR
@thigm85 @yngveaasheim FYI


bratseth (commented):
Were you planning to add some asserts to the second half of the test?
Otherwise, I suggest just commenting it out.

baldersheim:
I added a larger table as documentation and assert on it.

PR #13400, opened 2020-05-27
Add SystemMonitor and report to it from VersionStatusUpdater

jonmv:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VersionStatusUpdater.java
```diff
@@ -27,10 +33,30 @@ protected void maintain() {
...
+                    log.log(Level.INFO, "Failed reporting system version status to monitor: " + Exceptions.toMessageString(e));
+                }
+            });
         } catch (Exception e) {
             log.log(Level.WARNING, "Failed to compute version status: " + Exceptions.toMessageString(e) +
                                    ". Retrying in " + interval());
         }
     }
 
+    private static SystemMonitor.Confidence convert(VespaVersion.Confidence confidence) {
```
freva, line 52: Add a test that verifies that all `VespaVersion.Confidence` can be converted?

jonmv: Fair enough.

freva (approved):

PR #13401, opened 2020-05-27
Support provisioning exact capacity

mpolden:
Depends on internal PR.

@freva

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -122,10 +122,11 @@ public class Flags {
             "Takes effect on the next run of RetiredExpirer.",
             HOSTNAME);
 
-    public static final UnboundListFlag<HostCapacity> PREPROVISION_CAPACITY = defineListFlag(
+    public static final UnboundListFlag<HostCapacity> TARGET_CAPACITY = defineListFlag(
             "preprovision-capacity", List.of(), HostCapacity.class,
-            "List of node resources and their count that should be unallocated in zone to receive new deployments. When a " +
-            "preprovisioned host is taken, another one will be provisioned within next iteration of maintainer.",
+            "List of node resources and their count that should be provisioned." +
+            "In a dynamically provisioned zone this specifies the unallocated (i.e. pre-provisioned) capacity. " +
+            "Otherwise it specifies the total (unallocated or not) capacity.",
             "Takes effect on next iteration of DynamicProvisioningMaintainer.");
 
     public static final UnboundDoubleFlag DEFAULT_TERM_WISE_LIMIT = defineDoubleFlag(
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -3,7 +3,7 @@ package com.yahoo.vespa.flags;
 
 import com.yahoo.component.Vtag;
 import com.yahoo.vespa.defaults.Defaults;
-import com.yahoo.vespa.flags.custom.PreprovisionCapacity;
+import com.yahoo.vespa.flags.custom.HostCapacity;
 
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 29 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DynamicProvisioningMaintainer.java
```diff
@@ -93,31 +94,78 @@ void updateProvisioningNodes(NodeList nodes, Mutex lock) {
...
+     * @return Excess hosts that can be deprovisioned.
+     */
+    private List<Node> provisionExactCapacity(List<Node> removableHosts) {
+        List<NodeResources> capacity = capacityFrom(exactCapacityFlag);
+        if (capacity.isEmpty()) return List.of();
+
+        List<Node> excessHosts = new ArrayList<>(removableHosts);
+        List<NodeResources> unsatisfiedCapacity = new ArrayList<>(capacity);
+        for (var resources : capacity) {
+            for (var host : removableHosts) {
```
freva, line 127: Sort by cost, otherwise you risk counting a large host as satisfying a small HostCapacity, but then you wont find match for a larget HostCapacity request

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DynamicProvisioningMaintainer.java
```diff
@@ -93,31 +94,78 @@ void updateProvisioningNodes(NodeList nodes, Mutex lock) {
...
-        List<NodeResources> preProvisionCapacity = preprovisionCapacityFlag.value().stream()
-                .flatMap(cap -> {
-                    NodeResources resources = new NodeResources(cap.getVcpu(), cap.getMemoryGb(), cap.getDiskGb(), 1);
-                    return IntStream.range(0, cap.getCount()).mapToObj(i -> resources);
-                })
-                .sorted(NodeResourceComparator.memoryDiskCpuOrder().reversed())
-                .collect(Collectors.toList());
+    /** Converge zone to wanted capacity */
+    private void convergeToCapacity(NodeList nodes) {
+        List<Node> removableHosts = removableHostsOf(nodes);
```
freva, line 99: `removableHosts()` only contains "free" hosts, so if you were to allocate some of the host, this would lead to provisioning more hosts with exactly-provision-capacity?

Generally, I think `preprovisionCapacity()` and `provisionExactCapacity()` can be merged, instead you should have 2 variants of `removableHostsOf()`

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -122,11 +122,17 @@
...
-            "List of node resources and their count that should be present in zone to receive new deployments. When a " +
-            "preprovisioned is taken, new will be provisioned within next iteration of maintainer.",
-            "Takes effect on next iteration of HostProvisionMaintainer.");
+    public static final UnboundListFlag<HostCapacity> PREPROVISION_CAPACITY = defineListFlag(
+            "preprovision-capacity", List.of(), HostCapacity.class,
+            "List of node resources and their count that should be unallocated in zone to receive new deployments. When a " +
+            "preprovisioned host is taken, another one will be provisioned within next iteration of maintainer.",
+            "Takes effect on next iteration of DynamicProvisioningMaintainer.");
+
+    public static final UnboundListFlag<HostCapacity> EXACT_PROVISION_CAPACITY = defineListFlag(
```
freva, line 131: I'm not sure about introducing another effectively permanent flag, especially when these 2 flags are mutually exclusive...

mpolden: Agree. Do you have something in mind? I suppose the other flag could be reused, but then we overload its meaning.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DynamicProvisioningMaintainer.java
mpolden: > this would lead to provisioning more hosts with exactly-provision-capacity?

Good catch, will fix.

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
freva: > Do you have something in mind? 

I was hoping we could reuse the other flag. It's not ideal, but I think in the long run it will be less confusing than having 2 flags and needing to validate the state (only 1 of them should be set at a given time).

mpolden:
PTAL at last commit.

freva (approved):
Elegant solution! :100:

PR #13402, opened 2020-05-27
Bratseth/lowest real resources allocating

bratseth:
Account for resource tax when autoscaling.

I went a bit back and forth so probably best top read the end result.

### DIFF
--- a/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTest.java
+++ b/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTest.java
@@ -367,7 +367,7 @@ public class AutoscalingTest {
     }
 
     /**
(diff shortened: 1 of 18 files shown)


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTest.java
```diff
@@ -325,4 +366,40 @@ public void testAutoscalingAws() {
                                tester.autoscale(application1, cluster1.id(), min, max));
     }
 
+    /**
+     * This calculator subtracts the memory tac when forecasting overhead, but not when actually
```
freva, line 370:
```suggestion
     * This calculator subtracts the memory tax when forecasting overhead, but not when actually
```

freva (approved):

PR #13408, opened 2020-05-27
Record need for normal term field match data for phrase splitter input.

toregge:
@geirst : please review

This replaces #13407


baldersheim (approved):

geirst (commented):
👍

PR #13420, opened 2020-05-28
Avoid creating empty instance specs

jonmv:
@bratseth please review. 
@tokle FYI.

Allows an application with the deployment spec
```xml
<deployment some-attribute='some-value'>
</deployment>
```
to be deleted — currently this is disallowed, because it has a `"default"` instance.


bratseth (approved):

bratseth:
You need to fix some unit test in config-model-api though.

jonmv:
Thanks!

jonmv:
Yeah, this didn't work at all >_< Wonder how I crossed that unit test off as successful, in my head. Anyway, this should be better. PTAL.

bratseth (approved):

PR #13422, opened 2020-05-28
Allow from 5% skew in document distribution and still get good result…

baldersheim:
…s when asking for many hits.

@thigm85 and @bratseth PR
@vekterli and @yngveaasheim FYI

Sets p_max = 1.0 + skew_factor (5%)
Computes a modified n' = 1/p_max


bratseth (approved):
👍 nice

thigm85:
LGTM 👍

PR #13426, opened 2020-05-28
Get rid of SessionContext

hmusum:
The context just makes it harder to see what is going on and what
the dependencies really are.


gjoranv (approved):
💯 The GlobalComponentRegistry is a similar construct, but probably a lot more work to remove.

hmusum:
Yeah, I think that one will be a lot harder. We'll see.

bratseth:
🥇

PR #13431, opened 2020-05-29
add "isCd" to user api output

andreer:


flags/src/main/java/com/yahoo/vespa/flags/FetchVector.java
```diff
@@ -57,7 +57,10 @@
          * <p>NOTE: There is seldom any need to set ZONE_ID, as all flags are set per zone anyways. The controller
          * could PERHAPS use this where it handles multiple zones.
          */
-        ZONE_ID
+        ZONE_ID,
+
+        /** Email address of user - provided by auth0 in console. */
+        CONSOLE_USER_EMAIL
```
hakonhall, line 63: put it above ZONE_ID, which is a bit special.

hakonhall:
I presume the red travis is because the new dimension needs serialization code.

smorgrav:
Get it green and I'm good

hakonhall (approved):
Consider verifying the flag value in https://github.com/vespa-engine/vespa/blob/master/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java#L202

smorgrav (approved):

PR #13445, opened 2020-06-02
add check on console user feature flag dimension

andreer:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
```diff
@@ -206,6 +206,12 @@ private static void verifyValues(JsonNode root) {
                     // Throws exception if not recognized
                     NodeType.valueOf(nodeTypeString);
                 });
+            } else if (dimension.isEqualTo(DimensionHelper.toWire(FetchVector.Dimension.CONSOLE_USER_EMAIL))) {
+                condition.get("values").forEachArrayElement(conditionValue -> {
+                    String consoleUserEmailString = conditionValue.asString()
+                            .orElseThrow(() -> new IllegalArgumentException("Non-string email address: " + conditionValue));
+                    NodeType.valueOf(consoleUserEmailString);
```
hakonhall, line 213: Add unit test: This would then fail.

andreer: Yes, looks like I copy pasted a little too much :-)

hakonhall (approved):
👍

PR #13447, opened 2020-06-02
Allow gdb path based on image name.

aressem:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoreCollector.java
```diff
@@ -49,6 +52,17 @@ Path readBinPathFallback(NodeAgentContext context, Path coredumpPath) {
         return Paths.get(matcher.group("path").split(" ")[0]);
     }
 
+    Path GDBPath(NodeAgentContext context) {
+        DockerImage image = context.node().currentDockerImage().get();
```
freva, line 56: `currentDockerImage` is not guaranteed to be set (f.ex. if core is dumped while starting or stopping the container), I suggest defaulting to the RHEL 7 variant.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoreCollector.java
```diff
@@ -49,6 +52,17 @@ Path readBinPathFallback(NodeAgentContext context, Path coredumpPath) {
...
+
+        if (image.tag().get().startsWith("vespa/ci")) {
+            return vespaHomeInContainer.resolve("bin64/gdb");
```
freva, line 59: I suggest using `context.pathInNodeUnderVespaHome()` and remove the option from constructor

hmusum (approved):

freva (approved):

hmusum (approved):

PR #13449, opened 2020-06-02
Expose Cloud through ZoneApi

mpolden:
Depends on internal PR.

@freva


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ControllerMaintenance.java
```diff
@@ -102,10 +102,10 @@ public void deconstruct() {
     /** Create one OS upgrader per cloud found in the zone registry of controller */
     private static List<OsUpgrader> osUpgraders(Controller controller) {
         return controller.zoneRegistry().zones().controllerUpgraded().zones().stream()
-                         .map(ZoneApi::getCloudName)
+                         .map(ZoneApi::getCloud)
                          .distinct()
                          .sorted()
```
freva, line 107: Will this work? `Cloud` is not `Comparable`?

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ControllerMaintenance.java
mpolden: You're right, but I also don't see why they need to be sorted. Will remove it.

PR #13454, opened 2020-06-02
Get rid of RemoteSessionFactory

hmusum:
Please review only


gjoranv (approved):

gjoranv:
Restarted the build, as the failure was in a test with a 60 SECONDS timeout. The author at least had the decency to use caps lock for the time unit ;)
``` QueueBoundsTest.requireThatFullIntermediateQueueBlocksAndUnblocks:183->assertResultQueueSize:274 Queue never reached size 4 before timeout of 60 SECONDS. Size now: 0```

PR #13460, opened 2020-06-02
To avoid issues during feeding and resize stick to using the docidLim…

baldersheim:
…it used for the search.

@toregge or @geirst or @havardpe PR


geirst:
@toregge please have a look at this

toregge (approved):
We might need extra code to cap docid limit in the base class by a docid limit from the data structure to ensure safe access (in case the docid limit in the search context was too high for this data structure). That docid limit should be sampled after the read guard for the data structure has been taken. If we can guarantee that the search context always provides a low enough limit, then this extra code should not be needed.

Related to this, MatchView::match() should be changed to take the document meta store read guard before creating the match context.

baldersheim:
We are now taking the read guard prior to creating the match context.
The docid_limit is taken from the searchcontext and injected into all blueprints, and should hence have correct guarantees. That is the limit we are relying on all over in blueprints and should then be safe.

PR #13461, opened 2020-06-02
As the source bit vector might change in a different thread,

baldersheim:
ensure that we sample size once to avoid incosistent read outs during copy.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
+++ b/searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
@@ -32,6 +32,7 @@ extract_size_capacity(const AllocatedBitVector & bv) {
     while (capacity < size) {
         // Since size and capacity might be changed in another thread we need
         // this fallback to avoid inconsistency during shrink.
+        std::atomic_thread_fence(std::memory_order_seq_cst);
         size = bv.size();
         capacity = bv.capacity();
     }
--- a/searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
+++ b/searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
@@ -29,6 +29,12 @@ std::pair<BitVector::Index, BitVector::Index>
 extract_size_capacity(const AllocatedBitVector & bv) {
     BitVector::Index size = bv.size();
     BitVector::Index capacity = bv.capacity();
+    while (capacity < size) {
+        // Since size and capacity might be changed in another thread we need
+        // this fallback to avoid inconsistency during shrink.
+        size = bv.size();
+        capacity = bv.capacity();
+    }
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
```diff
@@ -17,6 +17,21 @@ size_t computeCapacity(size_t capacity, size_t allocatedBytes) {
...
+extract_size_size(const BitVector & bv) {
+    BitVector::Index size = bv.size();
+    return std::pair<BitVector::Index, BitVector::Index>(size, size);
+}
+
+std::pair<BitVector::Index, BitVector::Index>
+extract_size_capacity(const AllocatedBitVector & bv) {
+    BitVector::Index size = bv.size();
+    BitVector::Index capacity = bv.capacity();
+    return std::pair<BitVector::Index, BitVector::Index>(size, capacity);
```
toregge, line 32: If size and capacity is reduced at the same time on the underlying bitvector then capacity can become less than size here.

toregge (approved):

searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
baldersheim: capacity is updated before size when growing.
However upon shrink that will not be safe.
So I guess a loop and a comment is needed here.

baldersheim: Fixed, @toregge PTAL

searchlib/src/vespa/searchlib/common/allocatedbitvector.cpp
```diff
@@ -29,6 +29,12 @@ std::pair<BitVector::Index, BitVector::Index>
 extract_size_capacity(const AllocatedBitVector & bv) {
     BitVector::Index size = bv.size();
     BitVector::Index capacity = bv.capacity();
+    while (capacity < size) {
+        // Since size and capacity might be changed in another thread we need
+        // this fallback to avoid inconsistency during shrink.
+        size = bv.size();
+        capacity = bv.capacity();
+    }
```
toregge, line 37: both size() and capacity() are very small inlined methods.  Thus the compiler can just cache the result in a register. You need to add a std::atomic_thread_fence og replace the loop with `size = std::min(size, capacity);`

baldersheim: Added sequential consistency.

toregge:
Note that there is still a crash risk. If the original vector is shrunk after reading size and capacity then the AllocatedBitVector constructor might read past the end of the memory buffer allocated for the original vector.

baldersheim:
Yes, but that situation should be guarded at a higher level.

toregge:
There is no such guard there now.

PR #13468, opened 2020-06-03
Require wantToRetire when setting wantToDeprovision

mpolden:
Most of the code assumes this and after further discussion with @freva we
couldn't think of any other cases where `wantToDeprovision=true` and
`wantToRetire=false` is useful.

Some nodes exist in states `failed` and `deprovisioned` that violate the
constraint so we can't enforce it just yet.

@freva


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/restapi/NodesV2ApiTest.java
```diff
@@ -207,7 +207,7 @@ public void test_requests() throws Exception {
                                    Utf8.toBytes("{\"modelName\": \"foo\"}"), Request.Method.PATCH),
                        "{\"message\":\"Updated dockerhost1.yahoo.com\"}");
         assertResponse(new Request("http://localhost:8080/nodes/v2/node/dockerhost1.yahoo.com",
-                        Utf8.toBytes("{\"wantToDeprovision\": true}"), Request.Method.PATCH),
+                        Utf8.toBytes("{\"wantToRetire\": true, \"wantToDeprovision\": true}"), Request.Method.PATCH),
```
freva, line 210: So I'm guessing the JSON also has to specify `wantToRetire` before `wantToDeprovision`? I'm not sure if our clients do this, or even if there is a way to force certain order for some of them?

mpolden: Good point. PTAL

freva (approved):
:100:

PR #13473, opened 2020-06-03
Gjoranv/remove local session loader

gjoranv:
The removal of the SessionFactory interface was badly handled by git (or me), so the diff is larger than it could have been.


hmusum (approved):
Great!

PR #13476, opened 2020-06-03
Bjorncs/container thread pool

bjorncs:

### DIFF
--- a/container-core/src/main/resources/configdefinitions/threadpool.def
+++ b/container-core/src/main/resources/configdefinitions/threadpool.def
@@ -16,3 +16,6 @@ queueSize int default=0
 # get out of a bad state. This should be set a bit higher than the expected max execution
 # time of each request when in a state of overload, i.e about "worst case execution time*2"
 maxThreadExecutionTimeSeconds int default=190
+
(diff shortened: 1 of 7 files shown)


container-core/src/main/resources/configdefinitions/threadpool.def
```diff
@@ -16,3 +22,6 @@ queueSize int default=0
 # get out of a bad state. This should be set a bit higher than the expected max execution
 # time of each request when in a state of overload, i.e about "worst case execution time*2"
 maxThreadExecutionTimeSeconds int default=190
+
+# Prefix for the name of the threads
+name string default="default-pool"
```
baldersheim, line 27: missing EOL

baldersheim (approved):
Fix the comments and you are good to go.

baldersheim:
BTW I see these from the previous PR
Container.com.yahoo.container.di.componentgraph.core.ComponentNode	warning	Multiple public constructors found in class com.yahoo.container.handler.ThreadPoolProvider, there should only be one. If more than one public constructor is needed, the primary one must be annotated with @Inject.
I think this should be addressed in this PR too.

PR #13486, opened 2020-06-04
Bjorncs/container thread pool

bjorncs:


baldersheim:
It looks fine to me, but I think @gjoranv should have a look too.

baldersheim (approved):

gjoranv:
This looks good. The comment on the first commit is not accurate, as removing the inject annotation and inheritance doesn't change how the ContainerThreadPool is created. The deciding factor is which component is set up in the model, and that must have been the TPProvider. I know because the pool class itself takes a config class, and they no longer have a default constructor so that guice is unable to create them.

gjoranv (approved):

PR #13488, opened 2020-06-04
use vector of UP as MultiSearch::Children

arnej27959:
* add helper class for constructing MultiSearch::Children
  (mostly for unit tests)
* rewrite as needed to adapt

@havardpe @geirst @toregge please review

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/sourceblendersearch.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/sourceblendersearch.cpp
@@ -158,14 +158,14 @@ SourceBlenderSearch::setChild(size_t index, SearchIterator::UP child) {
     _sources[_children[index]] = child.release();
 }
 
-SourceBlenderSearch *
+SearchIterator::UP
 SourceBlenderSearch::create(std::unique_ptr<sourceselector::Iterator> sourceSelector,
                             const Children &children, bool strict)
 {
     if (strict) {
-        return new SourceBlenderSearchStrict(std::move(sourceSelector), children);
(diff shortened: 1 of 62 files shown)


havardpe (approved):

searchlib/src/vespa/searchlib/queryeval/sourceblendersearch.cpp
```diff
@@ -158,14 +158,14 @@ SourceBlenderSearch::setChild(size_t index, SearchIterator::UP child) {
...
 }
 
-SourceBlenderSearch *
+SearchIterator::UP
 SourceBlenderSearch::create(std::unique_ptr<sourceselector::Iterator> sourceSelector,
                             const Children &children, bool strict)
 {
     if (strict) {
-        return new SourceBlenderSearchStrict(std::move(sourceSelector), children);
+        return SearchIterator::UP(new SourceBlenderSearchStrict(std::move(sourceSelector), children));
```
baldersheim, line 166: Should you not use std::make_unique instead ?

baldersheim (approved):
This is a long awaited modernisation. I just had one comment on SearchIterator::UP versus std::make_unique.

searchlib/src/vespa/searchlib/queryeval/sourceblendersearch.cpp
arnej27959: I wanted to handle Strict and non-Strict the same way, and the other constructor wasn't public; will adjust with a NonStrict variant and use make_unique.

geirst (commented):
👍

PR #13489, opened 2020-06-04
Config server refactoring, part 6

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
```diff
@@ -38,32 +53,43 @@
...
+    private final HostRegistry<ApplicationId> hostRegistry;
+    private final ApplicationMapper applicationMapper = new ApplicationMapper();
+    private final MetricUpdater tenantMetricUpdater;
+    private final Clock clock = Clock.systemUTC();
 
-    private TenantApplications(Curator curator, ReloadHandler reloadHandler, TenantName tenant,
-                               ExecutorService zkCacheExecutor, StripedExecutor<TenantName> zkWatcherExecutor) {
-        this.curator = curator;
+
+    private TenantApplications(TenantName tenant, GlobalComponentRegistry registry) {
```
gjoranv, line 75: If all the things are directly available for injection, I suggest to do that instead of using the `GlobalComponentRegistry`.

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
hmusum: Yes, agreed, just need to do it in steps

PR #13492, opened 2020-06-05
Make sure to set previous active generation when one exists

hmusum:
Please review only

### DIFF
--- a/configserver/src/test/java/com/yahoo/vespa/config/server/ApplicationRepositoryTest.java
+++ b/configserver/src/test/java/com/yahoo/vespa/config/server/ApplicationRepositoryTest.java
@@ -142,6 +142,23 @@ public class ApplicationRepositoryTest {
     }
 
     @Test
+    public void redeploy() {
+        PrepareResult result = deployApp(testApp);
+
+        long firstSessionId = result.sessionId();
+
+        PrepareResult result2 = deployApp(testApp);
+        long secondSessionId = result2.sessionId();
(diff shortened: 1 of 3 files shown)


configserver/src/test/java/com/yahoo/vespa/config/server/ApplicationRepositoryTest.java
```diff
@@ -141,6 +141,23 @@ public void createAndPrepareAndActivate() {
...
+
+        PrepareResult result2 = deployApp(testApp);
+        long secondSessionId = result2.sessionId();
+        assertNotEquals(firstSessionId, secondSessionId);
+
+        TenantName tenantName = applicationId().tenant();
+        Tenant tenant = tenantRepository.getTenant(tenantName);
+        LocalSession session = tenant.getLocalSessionRepo().getSession(
+                tenant.getApplicationRepo().requireActiveSessionOf(applicationId()));
+        assertEquals(firstSessionId, session.getMetaData().getPreviousActiveGeneration());
```
gjoranv, line 158: Perhaps we should consider to stop using 'generation' if it really is the sessionId.

gjoranv (approved):

configserver/src/test/java/com/yahoo/vespa/config/server/ApplicationRepositoryTest.java
hmusum: Agreed, it's just an implementation detail that they are equal. I'll try to improve this in a later PR.

PR #13493, opened 2020-06-05
Disable connection reuse for VIPs

mpolden:
This should reduce the number of failed connections we see during upgrades.

@freva


freva (approved):
:100: 

Hopefully we can simplify all of this by always talking via VIP when configserver application package distribution WP is done?

PR #13500, opened 2020-06-08
Revert "Revert "When we pull in a cacheline, we should use it too.""

baldersheim:
Reverts vespa-engine/vespa#13499

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
@@ -5,7 +5,6 @@
 #include "andnotsearch.h"
 #include "sourceblendersearch.h"
 #include <vespa/searchlib/common/bitvectoriterator.h>
-#include <vespa/searchlib/fef/termfieldmatchdata.h>
 #include <vespa/searchlib/fef/termfieldmatchdataarray.h>
 #include <vespa/vespalib/util/optimized.h>
 #include <vespa/vespalib/hwaccelrated/iaccelrated.h>
@@ -27,8 +26,9 @@ public:
           _accel(IAccelrated::getAccelerator()),
           _lastWords()
     {
-        static_assert(sizeof(_lastWords) == 64, "Latswords should have 64 byte size");
-        memset(&_lastWords, 0, sizeof(_lastWords));
+        static_assert(sizeof(_lastWords) == 64, "Lastwords should have 64 byte size");
+        static_assert(NumWordsInBatch == 8, "Batch size should be 8 words.");
+        memset(_lastWords, 0, sizeof(_lastWords));
     }
 protected:
     void updateLastValue(uint32_t docId);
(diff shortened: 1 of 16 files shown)


baldersheim:
Depends on #13515

baldersheim:
This is a pure reapply. The issue causing the revert is addressed in #13515. It was triggered by having alignment above the standard which is 16.

searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
```diff
@@ -1,61 +1,91 @@
...
-            _lastMaxDocIdLimit = (index + 1) * WordLen;
-        } else {
+        if (__builtin_expect(docId >= _numDocs, false)) {
             setAtEnd();
+            return;
+        }
+        const uint32_t index(wordNum(docId));
+        if (docId >= _lastMaxDocIdLimitRequireFetch) {
+            uint32_t baseIndex = index & ~(sizeof(_lastWords)/sizeof(Word) - 1);
+            _update(_accel, baseIndex, _bvs, _lastWords);
```
havardpe, line 84: I think the magic should be done outside the update call and the update call just forward to the appropriate function.

searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
```diff
@@ -1,61 +1,91 @@
...
+        : MultiBitVectorIterator<Update>(std::move(children))
+    { }
 private:
     void doSeek(uint32_t docId) override { this->strictSeek(docId); }
     Trinary is_strict() const override { return Trinary::True; }
 };
 
+struct And {
+    using Word = BitWord::Word;
+    void operator () (const IAccelrated & accel, size_t offset, const std::vector<std::pair<const void *, bool>> & src, Word *dest) {
```
havardpe, line 59: offset should already be bytes here, and destination void.

searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
```diff
@@ -1,61 +1,91 @@
...
+        accel.and64(offset*sizeof(uint64_t), src, dest);
```
havardpe, line 60: offset is probably in Words and not in uint64_t, but I think this magic should be done together with the other magic anyways.

baldersheim: Fixed

searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
baldersheim: Fixed

searchlib/src/vespa/searchlib/queryeval/multibitvectoriterator.cpp
baldersheim: Fixed

PR #13501, opened 2020-06-08
Arnej/global filter after fetch postings

arnej27959:

@geirst @toregge @havardpe please review

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_blueprint.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_blueprint.cpp
@@ -101,11 +101,12 @@ NearestNeighborBlueprint::set_global_filter(const GlobalFilter &global_filter)
                 _approximate = false;
                 return;
             }
+            est_hits = std::min(est_hits, max_hits);
         }
-        perform_top_k();
-        est_hits = _found_hits.size();
-        LOG(debug, "perform_top_k found %u hits", est_hits);
+        est_hits = std::min(est_hits, _target_num_hits);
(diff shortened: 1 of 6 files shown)


havardpe (commented):
Not sure what happens when the strictness of fetch postings and create iterator do not match...

searchlib/src/vespa/searchlib/queryeval/nearest_neighbor_blueprint.cpp
```diff
@@ -84,10 +87,15 @@ NearestNeighborBlueprint::set_global_filter(const GlobalFilter &global_filter)
...
+    LOG(debug, "set_global_filter with: %s / %s / %s",
+        (_approximate ? "approximate" : "exact"),
+        (nns_index ? "nns_index" : "no_index"),
+        (_global_filter->has_filter() ? "has_filter" : "no_filter"));
     if (_approximate && nns_index) {
         uint32_t est_hits = _attr_tensor.getNumDocs();
         if (_global_filter->has_filter()) {
             uint32_t max_hits = _global_filter->filter()->countTrueBits();
+            LOG(debug, "set_global_filter getNumDocs: %u / max_hits %u", est_hits, max_hits);
             if (max_hits * 10 < est_hits) {
```
geirst, line 99: Until we have data on performance tests, perhaps we should _not_ do fallback to brute force. That will make it harder to interpret the results.

geirst (approved):
👍

hmusum:
Seems like system tests tests/search/explain/explain.rb started failing after this PR, you might want to have a look

PR #13503, opened 2020-06-08
Split cd test module in implementation and api modules

tokle:


cloud-tenant-base/pom.xml
```diff
@@ -0,0 +1,383 @@
...
+
+    <artifactId>cloud-tenant-base</artifactId>
+    <name>Vespa Cloud tenant base</name>
+    <version>7-SNAPSHOT</version>
+    <description>Parent POM for all Vespa Cloud applications.</description>
+    <url>https://github.com/vespa-engine</url>
+    <packaging>pom</packaging>
+
+    <parent>
+        <artifactId>tenant-base</artifactId>
```
bjorncs, line 16: Should this be `hosted-tenant-base`?

cloud-tenant-cd/pom.xml
```diff
@@ -20,6 +18,11 @@
     </parent>
 
     <dependencies>
```
bjorncs, line 20: We'll need to clean up these dependencies when we redesign this as a bundle

hosted-tenant-base/pom.xml
```diff
@@ -0,0 +1,408 @@
...
+            <artifactId>container</artifactId>
+            <version>${vespaversion}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.yahoo.vespa</groupId>
+            <artifactId>container-test</artifactId>
+            <version>${vespaversion}</version>
+            <scope>test</scope>
```
bjorncs, line 85: We'll have to override the scope to runtime for `cloud-tenant-base`, perhaps we need to move this definition down.

hosted-tenant-base/pom.xml
```diff
@@ -0,0 +1,408 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
```
bjorncs, line 88: No idea why these exclusions are here.

hosted-tenant-base/pom.xml
tokle: can't we just set the scope to runtime in hosted-tenant-base?

hosted-tenant-base/pom.xml
tokle: 🤷

bjorncs (approved):

bjorncs (approved):

PR #13508, opened 2020-06-08
Use xxhash64 for bucket-to-stripe distribution

vekterli:
@baldersheim please review
@geirst FYI

Existing naive prime-based solution was susceptible to scheduling
operations for the subtree of a superbucket in one strand alone,
despite previous attempts to disperse this using prime number
multiplication. This would put a serious limiter on parallelism
for super bucket locality-sensitive reads such as streaming search
visitors.


vekterli:
Local testing shows vastly better distribution for a case with large sub-bucket fan-out and  ~1200 operations to schedule across 8 persistence threads (4 stripes of 2 threads each).

Before:
```
    T0: 714
    T1: 500
    T2 - T7: nada
```

After:
```
    T0: 166
    T1: 202
    T2: 112
    T3: 156
    T4: 136
    T5: 132
    T6: 156
    T7: 192
```

baldersheim (approved):

PR #13509, opened 2020-06-08
Update return and throws javadoc for SyncSession::update

hakonhall:


hakonhall:
FYI @jvenstad

documentapi/src/main/java/com/yahoo/documentapi/SyncSession.java
```diff
@@ -6,6 +6,7 @@
 import com.yahoo.document.DocumentPut;
 import com.yahoo.document.DocumentRemove;
 import com.yahoo.document.DocumentUpdate;
+import com.yahoo.document.TestAndSetCondition;
```
vekterli, line 9: Is it preferable to use an explicit import or to use a fully qualified class name in the JavaDoc instead?

vekterli (approved):
:+1: one question regarding additional import

documentapi/src/main/java/com/yahoo/documentapi/SyncSession.java
hakonhall: I have never discussed this with anyone, but I prefer it this way.

PR #13516, opened 2020-06-09
store entry docid/level in an atomic value

arnej27959:

note that entry docid and level need to be atomic together, so i stored them in the same 64-bit value.

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hnsw_graph.h
+++ b/searchlib/src/vespa/searchlib/tensor/hnsw_graph.h
@@ -38,8 +38,8 @@ struct HnswGraph {
     NodeRefVector node_refs;
     NodeStore     nodes;
     LinkStore     links;
-    uint32_t      entry_docid;
-    int32_t       entry_level;
+
+    std::atomic<uint64_t> entry_docid_and_level;
 
     HnswGraph();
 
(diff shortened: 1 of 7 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_graph.h
```diff
@@ -63,9 +63,25 @@ struct HnswGraph {
     
     void set_link_array(uint32_t docid, uint32_t level, const LinkArrayRef& new_links);
 
-    void set_entry_node(uint32_t docid, int32_t level) {
-        entry_docid = docid;
-        entry_level = level;
+    struct EntryNode {
+        uint32_t docid;
+        int32_t level;
+        EntryNode() : docid(0), level(-1) {}
```
geirst, line 69: Consider adding a constructor that takes docid and level, so we can keep that one liner `set_entry_node(EntryNode(my_docid, my_level))`, instead of the four lines it has changed into now in various places.

geirst (approved):
Overall looks good. One comment to consider to reduce code bloat.

arnej27959:
@geirst PTAL

geirst (approved):
👍 Very nice

PR #13517, opened 2020-06-09
Revert "Reapply "Create local session based on remote session and distributed applicat…""

hmusum:
Reverts vespa-engine/vespa#13512

In case something fails with the PR. Please review only


gjoranv (approved):

hmusum:
Not needed

PR #13522, opened 2020-06-09
Clean up some visiting code. No functional changes.

vekterli:
@geirst please review

Also changed logger ID of visitor component, as its existing ID is a prefix of a bunch of other things (such as `searchvisitor`), making it a pain to only enable debug logging for the visitor component itself.


geirst (approved):
👍

PR #13526, opened 2020-06-09
Add APIs to support setting a tensor and update nearest neighbor inde…

geirst:
…x as two-phase operations.

This will enable using multiple threads to update the nearest neighbor index.

@toregge and @arnej27959 please review

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/nearest_neighbor_index.h
+++ b/searchlib/src/vespa/searchlib/tensor/nearest_neighbor_index.h
@@ -3,6 +3,7 @@
 #pragma once
 
 #include "distance_function.h"
+#include "prepare_result.h"
 #include <vespa/eval/tensor/dense/typed_cells.h>
 #include <vespa/vespalib/util/generationhandler.h>
 #include <vespa/vespalib/util/memoryusage.h>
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/tensor/nearest_neighbor_index.h
```diff
@@ -36,6 +38,38 @@ class NearestNeighborIndex {
...
+    }
+
+    /**
+     * Performs the complete step in a two-phase operation to add a document to the index.
+     *
+     * This function is only called by the attribute writer thread.
+     * It must wait for the result from the prepare step (via the future) before it does the modifying changes.
+     */
+    virtual void complete_add_document(uint32_t docid,
+                                       std::future<std::unique_ptr<PrepareResult>> prepare_result) {
```
toregge, line 67: promise and future handling should be at a layer above this code. Consider using `std::unique_ptr<PrepareResult>`or `PrepareResult&`instead of `std::future<std::unique_ptr<PrepareResult>>`.

geirst: Fixed

geirst:
@toregge and @arnej27959 PTAL

arnej27959 (approved):
looks good to me

PR #13531, opened 2020-06-10
Add feature flag control of default jvm gc options

baldersheim:
@bjorncs PR
@yngveaasheim FYI

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
@@ -89,6 +89,8 @@ public interface ModelContext {
 
         String docprocLoadBalancerType();
 
+        String jvmGCOptions();
+
         boolean useDistributorBtreeDb();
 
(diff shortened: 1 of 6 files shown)


config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
```diff
@@ -89,6 +89,8 @@ default double defaultTopKProbability() {
 
         String docprocLoadBalancerType();
 
+        String jvmGCOptions();
```
bjorncs, line 92: Consider adding a default implementation (you'll have to do that later anyway) + adding a comment on its purpose.

bjorncs (approved):

config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
baldersheim: I think we will keep this flag for a very long time for rolling out chnages to GC  as we discover better combinations.

PR #13533, opened 2020-06-10
protect RNG with mutex

arnej27959:

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/inv_log_level_generator.h
+++ b/searchlib/src/vespa/searchlib/tensor/inv_log_level_generator.h
@@ -19,19 +19,23 @@ class InvLogLevelGenerator : public RandomLevelGenerator {
     std::mt19937_64 _rng;
     std::mutex _mutex;
     std::uniform_real_distribution<double> _uniform;
-    double _levelMultiplier;
+    const double _levelMultiplier;
+
+    double get_uniform() {
+        std::lock_guard<std::mutex> guard(_mutex);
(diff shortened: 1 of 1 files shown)


searchlib/src/vespa/searchlib/tensor/inv_log_level_generator.h
```diff
@@ -16,16 +17,19 @@ namespace search::tensor {
...
     InvLogLevelGenerator(uint32_t m)
       : _rng(0x1234deadbeef5678uLL),
+        _mutex(),
         _uniform(0.0, 1.0),
         _levelMultiplier(1.0 / log(1.0 * m))
     {}
 
     uint32_t max_level() override {
+        std::lock_guard<std::mutex> guard(_mutex);
         double unif = _uniform(_rng);
```
baldersheim, line 33: Consider guarding only the random call, log might be expensive.

searchlib/src/vespa/searchlib/tensor/inv_log_level_generator.h
```diff
@@ -16,16 +17,19 @@ namespace search::tensor {
 
 class InvLogLevelGenerator : public RandomLevelGenerator {
     std::mt19937_64 _rng;
+    std::mutex _mutex;
     std::uniform_real_distribution<double> _uniform;
     double _levelMultiplier;
```
baldersheim, line 22: add const to signal that it does not change during the lifetime of the object.

baldersheim (approved):

PR #13541, opened 2020-06-10
Add draft LocalVisitorSession

jonmv:
@vekterli please review for discussion only. Will add tests and wire into the `LocalDocumentAccess` if it looks OK.


jonmv:
PTAL. Think it's ready now.

documentapi/src/main/java/com/yahoo/documentapi/local/LocalVisitorSession.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    @Override
+    public boolean waitUntilDone(long timeoutMs) throws InterruptedException {
+        return control.waitUntilDone(timeoutMs);
+    }
+
+    @Override
+    public void ack(AckToken token) {
+        outstanding.remove((DocumentId) token.ackObject);
```
vekterli, line 140: Since `outstanding` is a `ConcurrentSkipListMap` I'm presuming it operates on a logical snapshot of the set of initial document IDs where concurrent removes can't mess anything up? It's been a while since I've looked into its semantics 🙂

documentapi/src/main/java/com/yahoo/documentapi/local/LocalVisitorSession.java
```diff
@@ -0,0 +1,164 @@
...
+            try {
+                // Iterate through all documents and pass on to data handler
+                outstanding.forEach((id, document) -> {
+                    if (selector.accepts(new DocumentPut(document)) != Result.TRUE)
+                        return;
+
+                    Document copy = new Document(document.getDataType(), document.getId());
+                    for (Field field : document.getDataType().getFields())
+                        if (fieldSet.contains(field))
+                            copy.setFieldValue(field, document.getFieldValue(field));
```
vekterli, line 85: I'm one of those filthy `{}` lovers, so consider adding braces around at least the outer for-loop since it's easy to accidentally put things in the wrong scope when using multiple nested control flow statements

vekterli: Alternatively, I see that this functionality is already provided in `FieldSetRepo.copyFields`

vekterli (approved):
:+1: the visitor session behavior is a complex beast to model accurately, but I think this captures the most important parts

documentapi/src/main/java/com/yahoo/documentapi/local/LocalVisitorSession.java
jonmv: Its iterators are [weakly consistent](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#Weakly) — while this is weakly defined, they _can_ reflect changes during iteration.
This is the case for the last part of the unit test. I suppose this test would break, were the implementation of `ConcurrentSkipListMap` ever to change.

documentapi/src/main/java/com/yahoo/documentapi/local/LocalVisitorSession.java
jonmv: Sweet!

documentapi/src/test/java/com/yahoo/documentapi/local/LocalDocumentApiTestCase.java
```diff
@@ -0,0 +1,241 @@
...
+
+        // Select all music documents where the "artist" field is set
+        VisitorParameters parameters = new VisitorParameters("music.artist");
+        parameters.setFieldSet("music:artist");
+        VisitorControlHandler control = new VisitorControlHandler();
+        parameters.setControlHandler(control);
+        List<Document> received = new ArrayList<>();
+        parameters.setLocalDataHandler(new DumpVisitorDataHandler() {
+            @Override public void onDocument(Document doc, long timeStamp) {
+                received.add(doc);
```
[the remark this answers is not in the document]
jonmv: Will use a concurrent list.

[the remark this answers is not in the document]
jonmv: Well, that would have to be a set, then.

documentapi/src/main/java/com/yahoo/documentapi/local/LocalVisitorSession.java
jonmv: Put a guard on state being `RUNNING` in the iteration.

jonmv:
(Sorry about force-push — had to rebase on master for some local dev workz.)

PR #13543, opened 2020-06-10
Revert "Merge SessionFactory into SessionRepository"

baldersheim:
Reverts vespa-engine/vespa#13534
@hmusum I suspect this one is root cause for HostHandler.get_application_for_host failing.
I am not sure if test needs an update, or something else.


hmusum (approved):
Let's revert, I'll look into it

PR #13545, opened 2020-06-11
Provision load balancer for config server cluster

mpolden:
Depends on internal PR.

@tokle

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -287,6 +287,13 @@ public class Flags {
             CONSOLE_USER_EMAIL
     );
 
+    public static final UnboundBooleanFlag CONFIGSERVER_PROVISION_LB = defineFeatureFlag(
+            "configserver-provision-lb", false,
+            "Provision load balancer for config server cluster",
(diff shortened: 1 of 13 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -287,6 +287,13 @@
             CONSOLE_USER_EMAIL
     );
 
+    public static final UnboundBooleanFlag CONFIGSERVER_PROVISION_LB = defineFeatureFlag(
+            "configserver-provision-lb", false,
+            "Provision load balancer for config server cluster",
+            "Takes effect when zone-config-servers application is redeployed",
+            ZONE_ID
```
tokle, line 294: This is likely not necessary as this code will run on the config server? The zone_id dimension is only relevant on flags used in the controller.

tokle (approved):

tokle (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
mpolden: Yes, but I want to use the same flag in the controller when implementing DNS updates.

PR #13550, opened 2020-06-11
common code for operation inlining

havardpe:
- add common code to make selecting the appropriate template function
  easier (vespa/vespalib/util/typify.h)
- enable detection of lambda functions matching all low-level
  operations. (lookup_op1, lookup_op2)
- add typifiers to decide which low-level operations should be inlined
  (TypifyOp1, TypifyOp2)
- integrate into dense_simple_join as a pilot customer

@arnej27959 please review
@lesters FYI


arnej27959 (approved):
very nice :-)

PR #13554, opened 2020-06-11
Remove duplicate TenantApplications arguments

gjoranv:
.. by adding a method to return itself as a HostValidator


hmusum (approved):
Nice way of doing it 👍

PR #13560, opened 2020-06-12
revert back to getting random number in the [0,1) range

arnej27959:
* for whatever reason, std::uniform_real_distribution doesn't
  allow a>b when specifying the [a,b) half-open range.  So we
  need to get a number from [0,1) and then do (1.0-unif) to
  convert to the (0,1] range that we want.
* extend the unit test with more initial numbers, and write
  it in a more compact way.


geirst (approved):
👍

PR #13565, opened 2020-06-12
Import Tensorflow models vis ONNX conversion

lesters:
@bratseth Please review.
@aressem FYI.

This switches TensorFlow import to use `tf2onnx` to convert to an ONNX model first. It tries to convert to ONNX using 3 different opsets, and fails if none can be converted (see `TensorFlowImporter`). Since this can't be tested in unit test (it requires the python tf2onnx script), I've removed all TensorFlow models from the config-model and application directories and moved those tests to be performed on ONNX models instead.

There is some residual work after this PR:

1. There is now a lot of dead code in model-integration for TensorFlow. Still some stuff there that is tested, so a follow up is to clean this up.
2. Remove dependencies on TensorFlow in our poms etc.
3. We should update the 'tf2onnx' python script and the TensorFlow python import to a TensorFlow versjon >= 2.1

I've tested all models in the systemtests that they convert successfully - but I might have missed some stuff like possible changes in names of outputs etc. I'll follow them up at next test run.

Don't merge until it passes PR build :-)


bratseth (approved):
Great

PR #13570, opened 2020-06-12
Arnej/use typify invoke for select

arnej27959:

@havardpe please review
(these are the simple cases)

### DIFF
--- a/eval/src/vespa/eval/tensor/dense/dense_xw_product_function.cpp
+++ b/eval/src/vespa/eval/tensor/dense/dense_xw_product_function.cpp
@@ -76,35 +76,6 @@ void my_cblas_float_xw_product_op(eval::InterpretedFunction::State &state, uint6
     state.pop_pop_push(state.stash.create<DenseTensorView>(self.result_type, TypedCells(dst_cells)));
 }
 
-template <bool common_inner>
-struct MyXWProductOp {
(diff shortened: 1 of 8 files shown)


eval/src/vespa/eval/tensor/dense/dense_xw_product_function.cpp
```diff
@@ -92,7 +92,9 @@ eval::InterpretedFunction::op_function my_select2(CellType lct, CellType rct) {
             return my_cblas_float_xw_product_op<common_inner>;
         }
     }
-    return select_2<MyXWProductOp<common_inner>>(lct, rct);
+    using Target = MyXWProductOp<common_inner>;
+    using MyTypify = eval::TypifyCellType;
+    return typify_invoke<2,MyTypify,Target>(lct, rct);
```
havardpe, line 97: could probably typify more? (common_inner)

havardpe (approved):

eval/src/vespa/eval/tensor/dense/dense_xw_product_function.cpp
havardpe: please use same strategy as suggested for matmul in
https://github.com/vespa-engine/vespa/pull/13571

havardpe (approved):

PR #13571, opened 2020-06-12
use typify_invoke to select matmul implementation

arnej27959:
* instead of a chain of select functions, use typify_invoke;
  use partial specialization on a struct to handle the
  double*double and float*float cases specially.

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/tensor/dense/dense_matmul_function.cpp
+++ b/eval/src/vespa/eval/tensor/dense/dense_matmul_function.cpp
@@ -80,47 +80,6 @@ void my_cblas_float_matmul_op(eval::InterpretedFunction::State &state, uint64_t 
     state.pop_pop_push(state.stash.create<DenseTensorView>(self.result_type, TypedCells(dst_cells)));
 }
 
-template <bool lhs_common_inner, bool rhs_common_inner>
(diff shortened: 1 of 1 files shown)


eval/src/vespa/eval/tensor/dense/dense_matmul_function.cpp
```diff
@@ -160,6 +119,35 @@ const TensorFunction &create_matmul(const TensorFunction &a, const TensorFunctio
...
+};
+
+template <bool lhs_common_inner, bool rhs_common_inner>
+struct MyMatMulOp<float, float, lhs_common_inner, rhs_common_inner> {
+    static auto get_fun() {
+        return my_cblas_float_matmul_op<lhs_common_inner, rhs_common_inner>;
+    }
+};
+
+struct MyTarget {
```
havardpe, line 143: can we eliminate this struct by letting the others capture the TypifyBool::Result instead of bool?

havardpe (approved):

eval/src/vespa/eval/tensor/dense/dense_matmul_function.cpp
havardpe: alternative suggestion added below

PR #13573, opened 2020-06-12
Use scope 'provided' for builtin jdisc artifacts.

gjoranv:
- These are already pulled in by the vespaclient-java
  jar via its compile scoped dependency on container-dev.

@freva Please review, do not merge.

### DIFF
--- a/documentapi/pom.xml
+++ b/documentapi/pom.xml
@@ -27,6 +27,7 @@
       <groupId>com.yahoo.vespa</groupId>
       <artifactId>component</artifactId>
       <version>${project.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


freva (approved):

gjoranv:
Alas, this would break the document api guide:
https://docs.vespa.ai/documentation/document-api-guide.html

Need to find another way..

gjoranv:
@freva, I might revive this by adding a new pom artifact for users to pull in, instead of `documentapi`. It will be something like the `container` artifact for container plugin development.

kkraune:
thanks, feel free to change https://docs.vespa.ai/documentation/document-api-guide.html at will, this is not widely used outside docprocs I presume

PR #13579, opened 2020-06-13
auto-detect actual type typify results

havardpe:

@arnej27959 please review


arnej27959 (approved):
looks good to me; any special reason you did not keep "has_type_member" name?

havardpe:
I had has_type_member first, but realized that we want to have a 'type' member that is a type. Then it was has_type_type_member. 'Has' implies a member or similar relation and has_type_type felt unique enough. These are my random reasons. I am open to changing the name.

PR #13580, opened 2020-06-14
avoid deadlock when using compile cache with blocking executors

havardpe:
- avoid taking cache lock in compile task destructor
- avoid holding cache lock while posting compile task
- add unit test trying to provoke deadlock
- bonus: fix blocking executor implementation
  (did not work with threads <= task limit)

@baldersheim please review


havardpe:
edit: did not work with threads >= task limit

PR #13583, opened 2020-06-15
Add local session to session cache

hmusum:
If we managed to created a local session, add it to the session cache.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
@@ -16,9 +16,7 @@ public class SessionCache<SESSIONTYPE extends Session> {
     private final HashMap<Long, SESSIONTYPE> sessions = new HashMap<>();
 
     public synchronized void addSession(SESSIONTYPE session) {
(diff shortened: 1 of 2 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
```diff
@@ -16,9 +16,7 @@
     private final HashMap<Long, SESSIONTYPE> sessions = new HashMap<>();
 
     public synchronized void addSession(SESSIONTYPE session) {
-        if (sessions.containsKey(session.getSessionId()))
-            throw new IllegalArgumentException("There already exists a session with id '" + session.getSessionId() + "'");
-        sessions.put(session.getSessionId(), session);
+        sessions.putIfAbsent(session.getSessionId(), session);
```
baldersheim, line 19: Intentional that already existing session no longer causes an exception ?

baldersheim (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
hmusum: Yes, I could not find any issues with changing this

PR #13592, opened 2020-06-15
Two phase put in attribute writer

geirst:
@toregge please review
@baldersheim @arnej27959 FYI

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/attribute/attribute_writer.cpp
+++ b/searchcore/src/vespa/searchcore/proton/attribute/attribute_writer.cpp
@@ -146,13 +146,14 @@ complete_put_to_attribute(SerialNum serial_num,
                           uint32_t docid,
                           AttributeVector& attr,
                           const FieldValue::SP& field_value,
-                          std::unique_ptr<PrepareResult> prepare_result,
+                          std::future<std::unique_ptr<PrepareResult>>& result_future,
                           bool immediate_commit,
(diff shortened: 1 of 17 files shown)


searchcore/src/vespa/searchcore/proton/attribute/attribute_writer.cpp
```diff
@@ -303,6 +351,101 @@ PutTask::run()
...
+{
+}
+
+CompletePutTask::~CompletePutTask() = default;
+
+void
+CompletePutTask::run()
+{
+    if (_attr.getStatus().getLastSyncToken() < _serial_num) {
+        auto result = _result_future.get();
```
toregge, line 443: This will hang if promise is never set, e.g. when PreparePutTask::run() doesn't set promise due to field value not being set.

toregge (changes requested):

searchcore/src/vespa/searchcore/proton/attribute/attribute_writer.cpp
geirst: Fixed

toregge (approved):

geirst:
Failed Travis build is not related: `NodeAgentContextManagerTest.blocks_in_nextContext_until_one_is_scheduled:69->Object.wait:328->Object.wait:-2 Â» TestTimedOut, [INFO] node-admin ......................................... FAILURE [ 33.302 s]`

PR #13599, opened 2020-06-16
[VESPA-18430] Stabilize unit test

freva:


node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextManagerTest.java
```diff
@@ -141,51 +141,43 @@ public void setFrozen_is_successful_if_converged_in_time() throws InterruptedExc
         assertEquals(Optional.of(true), asyncScheduler.response);
     }
 
+    private void waitUntilWaitingForNextContext() {
```
hakonhall, line 144: This should be a method on NodeAgentContext using Object::wait and notify with no sleep necessary.

hakonhall (approved):

hakonhall (approved):

PR #13609, opened 2020-06-17
Change default value, all zones already use 'true'

hmusum:
Seems to me that we can remove overrides (and eventually get rid of the flag) after this change is released


hakonhall (approved):
Yes, this can be removed, both the flag values and this flag definition. I can send out the additional PRs.

PR #13612, opened 2020-06-17
Randomness is also needed when having only 2 nodes. If not it will al…

baldersheim:
…ways select [0] when pending is equal.

@vekterli or @yngveaasheim PR


vekterli (approved):
:+1:

PR #13613, opened 2020-06-17
Support "user" and "group" as identifiers in C++ selection parser 

vekterli:
@geirst please review

These were not included in the explicit list of allowed tokens
and would therefore cause parse failures if e.g. a document type
was called "user".

Java parser already has the necessary tokens added.


geirst (approved):
👍

PR #13626, opened 2020-06-18
observe the results of multiply-add across implementations

havardpe:
- add specific test showing diverging results when using FMA
- re-write eval test to not fail due to FMA
- re-write hitcollector test to not fail due to FMA

@arnej27959 please review
@baldersheim FYI


arnej27959 (approved):
👍

PR #13631, opened 2020-06-18
added 'erf' function

havardpe:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

@lesters please review and merge
we should enable erf in the cross-language conformance test once it is supported by Java

PR #13635, opened 2020-06-18
Lesters/additional onnx operators

lesters:
@bratseth Please review.  Required to support some transformer models.


lesters:
the `erf` function has also been added to the backend: https://github.com/vespa-engine/vespa/pull/13631

bratseth (approved):

PR #13641, opened 2020-06-19
Arnej/target num hits alias

arnej27959:

@bratseth please review
@geirst FYI


arnej27959:
random aside: SelectParser code does (int) annotation.asDouble(), but I think it's better to do (int) annotation.asLong() since it's almost certainly a long internally.

bratseth (approved):

PR #13647, opened 2020-06-19
Replace joins with same input with a map

lesters:
@bratseth Please review. Most often used when squaring tensors. Allows this to be done in place.

### DIFF
--- a/model-integration/src/main/java/ai/vespa/rankingexpression/importer/operations/IntermediateOperation.java
+++ b/model-integration/src/main/java/ai/vespa/rankingexpression/importer/operations/IntermediateOperation.java
@@ -312,6 +312,14 @@ public abstract class IntermediateOperation {
         return -1;
     }
 
+    /** Removes outputs if they point to the same operation */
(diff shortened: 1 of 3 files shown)


bratseth (commented):
LGTM but shouldn't this be done in searchlib/../TensorOptimizer?

lesters:
The `TensorOptimizer` is for optimizing tensor evaluation Java side. This is for evaluating the graph on the backend, as it is difficult to recognize if the two inputs to the join are the same. So this is most easily done during graph import.

bratseth:
> The TensorOptimizer is for optimizing tensor evaluation Java side.

Yes you are right of course; this should be an ExpressionTransformer.

> it is difficult to recognize if the two inputs to the join are the same

How so?

lesters:
That is a good point that this could be an additional `ExpressionTransformer`, however the additional effect here was to not generate the function representing the inputs (when an operation is used more than once, we automatically create a function). The effect is that the map can then be done in-place, i.e. without requiring memory allocations. As far as I understand, this is difficult to resonate about on the backend.

PR #13653, opened 2020-06-22
Tgm/application package

thigm85:
This PR contains:
* Code and unit tests to create application package files (package.py and test_package.py) 
* Jinja 2 templates for schema.sd, hosts.xml and services.xml
* Jupyter notebook showing a full working example of creating a package and deploying it locally from python.
* Vespa Cloud deployment will be addressed in a future sprint.
* Implemented json serialization for every python object to be able to save and load the application from python.

### DIFF
--- a/python/vespa/vespa/package.py
+++ b/python/vespa/vespa/package.py
@@ -1,4 +1,79 @@
+from typing import List, Mapping, Optional
+
 import docker
+
+from vespa.json_serialization import ToJson, FromJson
+
+
+class Document(object):
+    def __init__(self) -> None:
+        """
+        Object representing a Vespa document.
+
+        """
+        self.fields = []
+
+    def add_fields(self, *fields):
+        """
+        Add Fields to the document.
+
+        :param fields: fields to be added
+        :return:
+        """
+        self.fields.extend(list(fields))
+
+
+class Field(ToJson, FromJson["Field"]):
+    def __init__(
+        self,
+        name: str,
+        type: str,
+        indexing: Optional[List[str]] = None,
+        index: Optional[List[str]] = None,
+    ) -> None:
+        """
+        Object representing a Vespa document field.
+
+        :param name: Field name.
+        :param type: Field data type.
+        :param indexing: Configures how to process data of a field during indexing.
+        :param index: Sets index parameters. Content in fields with index are normalized and tokenized by default.
+        """
+        self.name = name
+        self.type = type
+        self.indexing = indexing
+        self.index = index
+
+    @staticmethod
+    def from_dict(mapping: Mapping) -> "Field":
+        return Field(
+            name=mapping["name"],
+            type=mapping["type"],
+            indexing=mapping.get("indexing", None),
+            index=mapping.get("index", None),
+        )
+
+    @property
+    def to_dict(self) -> Mapping:
+        map = {"name": self.name, "type": self.type}
+        if self.indexing is not None:
+            map.update(indexing=self.indexing)
+        if self.index is not None:
+            map.update(index=self.index)
+        return map
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        return (
+            self.name == other.name
+            and self.type == other.type
+            and self.indexing == other.indexing
+            and self.index == other.index
+        )
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
 
 
 class ApplicationPackage(object):
@@ -21,16 +96,20 @@ class ApplicationPackage(object):
         :return:
         """
         client = docker.from_env()
-        self.container = client.containers.run(
-            "vespaengine/vespa",
-            detach=True,
-            mem_limit=container_memory,
-            name=self.name,
-            hostname=self.name,
-            privileged=True,
-            volumes={self.disk_folder: {"bind": "/app", "mode": "rw"}},
-            ports={8080: 8080, 19112: 19112},
-        )
+        if self.container is None:
+            try:
+                self.container = client.containers.get(self.name)
+            except docker.errors.NotFound:
+                self.container = client.containers.run(
+                    "vespaengine/vespa",
+                    detach=True,
+                    mem_limit=container_memory,
+                    name=self.name,
+                    hostname=self.name,
+                    privileged=True,
+                    volumes={self.disk_folder: {"bind": "/app", "mode": "rw"}},
+                    ports={8080: 8080, 19112: 19112},
+                )
 
     def check_configuration_server(self) -> bool:
         """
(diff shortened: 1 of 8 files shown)


python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+        self.fields = fields
+
+    def add_fields(self, *fields: Field):
+        """
+        Add Fields to the document.
+
+        :param fields: fields to be added
+        :return:
+        """
+        self.fields.extend(list(fields))
```
oyving, line 89: I think `list()` here is redundant.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+        map = {"fields": [field.to_envelope for field in self.fields]}
+        return map
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        return self.fields == other.fields
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
```
oyving, line 106: I would avoid `\n` in `__repr__`. It's better to have it on one line. `__repr__` should attempt to look like a Python expressions while `__str__` is for informal string representation. See https://docs.python.org/3/reference/datamodel.html#object.__repr__.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+            return False
+        return (
+            self.name == other.name
+            and self.type == other.type
+            and self.indexing == other.indexing
+            and self.index == other.index
+        )
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
```
oyving, line 68: See `__repr__` comment below.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+        map = {"name": self.name, "fields": self.fields}
+        return map
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        return self.name == other.name and self.fields == other.fields
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
```
oyving, line 140: See `__repr__` comment above.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+class Document(ToJson, FromJson["Document"]):
+    def __init__(self, fields: Optional[List[Field]] = None) -> None:
+        """
+        Object representing a Vespa document.
+
+        """
+        if not fields:
+            fields = []
+
+        self.fields = fields
```
oyving, line 80: Just a different style, it might not be more clear: `self.fields = [] if not fields else fields`.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+        if not isinstance(other, self.__class__):
+            return False
+        return (
+            self.name == other.name
+            and self.first_phase == other.first_phase
+            and self.inherits == other.inherits
+        )
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
```
oyving, line 182: See `__repr__` above.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+                rank_profile.name: rank_profile for rank_profile in rank_profiles
+            }
+
+    def add_rank_profile(self, rank_profile: RankProfile) -> None:
+        """
+        Add a `RankProfile` to the `Schema`.
+        :param rank_profile: `RankProfile` to be added.
+        :return: None.
+        """
+        self.rank_profiles.update({rank_profile.name: rank_profile})
```
oyving, line 220: Why `.update()` instead of just `self.rank_profiles[rank_profile.name] = rank_profile`?

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+            return False
+        return (
+            self.name == other.name
+            and self.document == other.document
+            and self.fieldsets == other.fieldsets
+            and self.rank_profiles == other.rank_profiles
+        )
+
+    def __repr__(self):
+        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
```
oyving, line 259: See `__repr__` above.

python/vespa/vespa/package.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+class ApplicationPackage(ToJson, FromJson["ApplicationPackage"]):
```
oyving, line 262: There are a few methods in this class that related to running the application. I would pull those out into a separate class or just have them as regular functions outside a class. Then it would be easier to have `VespaCloud.run(app)` and `VespaDocker.run(app)` and `ApplicationPackage` just becomes a data container.

python/vespa/vespa/templates/services.xml
```diff
@@ -0,0 +1,18 @@
...
+    <container id="{{ application_name }}_container" version="1.0">
+        <search></search>
+        <document-processing></document-processing>
+        <document-api></document-api>
+    </container>
+    <content id="{{ application_name }}_content" version="1.0">
+        <redundancy reply-after="1">1</redundancy>
+        <documents>
+            <document type="{{ document_name }}" mode="index"></document>
+            <document-processing cluster="{{ application_name }}_container"></document-processing>
```
oyving, line 12: I don't think this is necessary. The indexing docproc cluster is implicitly there.

python/vespa/vespa/templates/services.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<services version="1.0">
+    <container id="{{ application_name }}_container" version="1.0">
+        <search></search>
+        <document-processing></document-processing>
```
oyving, line 5: Unless you are adding document processors, you don't need this either.

oyving (changes requested):

python/vespa/vespa/package.py
thigm85: Good point.

python/vespa/vespa/package.py
```diff
@@ -65,7 +65,13 @@ def __eq__(self, other):
         )
 
     def __repr__(self):
-        return "{0}\n{1}".format(self.__class__.__name__, str(self.to_dict))
+        return "{0}({1}, {2}, {3}, {4})".format(
+            self.__class__.__name__,
+            str(self.name),
```
oyving, line 70: You should probably call `repr()` on these.

oyving (approved):
I think if you want to have `__repr__` on the object, you should probably called `repr()` on the members of the class. If you are just using these to make debugging easier and just need an informal string representation then maybe just implementing `__str__` would be easier.

oyving (approved):

PR #13657, opened 2020-06-22
Merge watcher classes for sessions into one class

hmusum:
Also make sure to add local session to watcher in the case where it is
added at a later time than when the watcher was created

Commit no. 2 from https://github.com/vespa-engine/vespa/pull/13636 and additional changes mentioned above. 

Please review only.


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.concurrent.Executor;
+import java.util.logging.Logger;
+
+/**
+ * Watches one particular session (/config/v2/tenants/&lt;tenantName&gt;/sessions/&lt;n&gt;/sessionState in ZooKeeper)
+ * The session must be in the session repo.
+ *
+ * @author Vegard Havdal
+ */
```
gjoranv, line 20: I suggest to add yourself as author since you have created the class/file.

gjoranv (approved):
🏆

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
hmusum: Yes, good point, will do

PR #13669, opened 2020-06-23
Add source target per module for generated sources.

toregge:
@geirst : please review
@aressem, @vekterli : FYI


geirst:
@aressem please review instead

aressem (approved):

PR #13673, opened 2020-06-23
If the estimated rate of matching documents is less than global filte…

toregge:
…r limit then

don't build a global filter.

@geirst : please review
@arnej27959 : FYI


geirst (approved):
👍

arnej27959 (approved):
looks good to me

PR #13679, opened 2020-06-23
Prefer hosts at least twice the size of the node

bratseth:


bratseth:
@freva I merged this by accident. Please review though.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/PrioritizableNode.java
```diff
@@ -128,6 +133,15 @@ public int compareTo(PrioritizableNode other) {
     /** Returns the allocation skew of the parent of this after adding this node to it */
     double skewWithThis() { return skewWith(node.resources()); }
 
+    private boolean lessThanHalfTheHost(PrioritizableNode node) {
```
freva, line 136: The method name is the opposite of what it returns

freva (commented):
:+1:

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/PrioritizableNode.java
bratseth: No.

PR #13688, opened 2020-06-24
Set spare count 0 where we can provision on demand MERGEOK

freva:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -124,7 +123,7 @@ public NodeRepository(NodeRepositoryConfig config,
              new DnsNameResolver(),
              DockerImage.fromString(config.dockerImage()), config.useCuratorClientCache(),
              provisionServiceProvider.getHostProvisioner().isPresent(),
-             zone.environment() == Environment.prod ? 1 : 0);
+             zone.environment().isProduction() && provisionServiceProvider.getHostProvisioner().isEmpty() ? 1 : 0);
```
bratseth, line 126:
```suggestion
             zone.environment().isProduction() && ! zone.getCloud().dynamicProvisioning() ? 1 : 0);
```

bratseth (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
freva: This wont exclude `cd-aws-us-east-1a`

bratseth: Because we pre-provision there? I think we should correct the definition for it then. dynamicProvisioning() definition:

> /** Returns whether this can provision hosts dynamically */

And that looks to me like the exact right condition for when we should have spares.

freva:
Merging to fix integration tests in AWS.

FYI: @bjormel

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
bratseth: ... looks like we need to know both whether it can and if it wants to always provision dynamically ... or something. We should clean up this stuff, but independent of this PR.

PR #13696, opened 2020-06-24
Handle an existing spantree in exactmatch

bratseth:
This may happen if a field which is indexed is used as an input for another field indexed as exact match.


bratseth:
Ping @arnej27959

arnej27959 (approved):
looks good to me

PR #13697, opened 2020-06-24
make tenant node quota adjustable by feature flag 

andreer:
Perhaps too simple? But it will allow us to get rid of the strange hashcodes in the next step (for which one can easily produce other matching names ...)


oyving (approved):
I think this approach makes sense as a stopgap measure. Does this look like what you were thinking, @bratseth?

bratseth (approved):
LGTM!

PR #13699, opened 2020-06-24
Separate documentapi artifacts

gjoranv:
This replaces #13573, and is necessary to allow bundle-plugin to fail when a bundle embeds the `component` artifact. This is also a small step in the direction to build slim bundles.

We should now use `container-documentapi` for anything related to the container, while `documentapi` should only be used for standalone clients.

@aressem Is anything else requried to have the new module built on factory and published on maven-central?


gjoranv:
´´´
CMake Error at CMakeLists.txt:78 (add_subdirectory):
  The source directory
    /source/documentapi
  does not contain a CMakeLists.txt file.
´´´

Why does this module need a CMakeLists, and not the `container` module?

I didn't notice the C++ code in here until now. Why is this a mixed module? Perhaps we should use the name `documentapi-lib` instead.

bratseth (approved):

toregge:
Consider not moving the C++ code from documentapi module to container-documentapi module.

vekterli:
The documentapi directory contains all the C++ logic required for the content cluster to both receive and send DocumentAPI protocol messages from/to clients. It is a "legacy" module and these had a tendency to be mixed language. But the C++ code should not be moved to a Java-specific target directory.

geirst (changes requested):
As summarised by @toregge and @vekterli, the C++ code should not be moved.

gjoranv:
As agreed with Tor on Slack, I'll create a new PR that does not move the C++ code, and update the OWNERS to reflect reality.

PR #13704, opened 2020-06-25
Let the executor create an executor id to its liking.

baldersheim:
Taske full advantage over strands instead of being limited to threads.

@geirst and @havardpe PR

### DIFF
--- a/staging_vespalib/src/vespa/vespalib/util/adaptive_sequenced_executor.cpp
+++ b/staging_vespalib/src/vespa/vespalib/util/adaptive_sequenced_executor.cpp
@@ -256,6 +256,11 @@ AdaptiveSequencedExecutor::~AdaptiveSequencedExecutor()
     assert(_worker_stack.empty());
 }
 
+ISequencedTaskExecutor::ExecutorId
+AdaptiveSequencedExecutor::getExecutorId(uint64_t component) const {
+    return ExecutorId(component % _strands.size());
+}
+
 void
 AdaptiveSequencedExecutor::executeTask(ExecutorId id, Task::UP task)
 {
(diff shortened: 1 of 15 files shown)


havardpe:
The following tests FAILED:
	568 - staging_vespalib_adaptive_sequenced_executor_test_app (Failed)

staging_vespalib/src/vespa/vespalib/util/adaptive_sequenced_executor.cpp
```diff
@@ -256,6 +256,11 @@ AdaptiveSequencedExecutor::~AdaptiveSequencedExecutor()
     assert(_worker_stack.empty());
 }
 
+ISequencedTaskExecutor::ExecutorId
+AdaptiveSequencedExecutor::getExecutorId(uint64_t component) const {
```
havardpe, line 260: this bypasses a lot of magic implemented in the 'interface'; is that intentional? if yes, move magic to the other subclass. If no, move magic to utility function that can be called from this one.

havardpe (changes requested):
Please fix failing unit test.
Also consider the fact that syncing all strands will be more expensive than before.

staging_vespalib/src/vespa/vespalib/util/adaptive_sequenced_executor.cpp
baldersheim: Yes, that is intentional. I planned to move the magic out of the interface in a separate PR. I can put it in this one if you prefer that.

baldersheim:
@havardpe Test has been updated. PTAL

havardpe (approved):

PR #13706, opened 2020-06-25
Create generic B-tree bucket DB and content node DB implementation

vekterli:
@geirst please review. This is a big diff so it makes sense to have a separate show&tell session.

This is the first stage of removing the legacy DB implementation.
Support for B-tree specific functionality such as lock-free snapshot
reads will be added soon. This commit is just for feature parity.

Abstract away actual database implementation to allow it to
be chosen dynamically at startup. This abstraction does incur
some overhead via call indirections and type erasures of callbacks,
so it's likely it will be removed once the transition to the
new B-tree DB has been completed.

Since the algorithms used for bucket key operations is so similar
between the content node and distributor, a generic B-tree backed
bucket database has been created. The distributor DB will be rewritten
around this code very soon.

Due to the strong coupling between bucket locking and actual DB
implementation details, the new bucket DB has a fairly significant
code overlap with the legacy implementation. This is to avoid
spending time abstracting away and factoring out code for a
legacy implementation that is to be removed entirely anyway.

Remove existing `LockableMap` functionality not used or that's
only used by tests.


storage/src/vespa/storage/bucketdb/generic_btree_bucket_database.hpp
```diff
@@ -0,0 +1,494 @@
...
+    BTreeTrailingInserter(DBType& db, BTreeBuilderType& builder)
+        : _db(db),
+          _builder(builder)
+    {}
+
+    ~BTreeTrailingInserter() override = default;
+
+    void insert_at_end(const BucketId& bucket_id, const ValueType& e) override {
+        const uint64_t bucket_key = bucket_id.toKey();
+        const auto new_value = DataStoreTraitsT::store_and_wrap_value(_db.store(), e);
```
geirst, line 453: Consider rename to wrap_and_store_value.

storage/src/tests/bucketdb/lockablemaptest.cpp
```diff
@@ -641,24 +703,64 @@ TEST(LockableMapTest, find_all_inconsistent_below_16_bits) {
...
     map.insert(id2.stripUnused().toKey(), A(1,2,3), "foo", preExisted);
     {
-        Map::WrappedEntry entry(
-                map.get(id1.stripUnused().toKey(), "foo", true));
+        auto entry = map.get(id1.stripUnused().toKey(), "foo", true);
         EXPECT_FALSE(map.isConsistent(entry));
     }
 }
 
+TYPED_TEST(LockableMapTest, get_without_auto_create_does_implicitly_not_lock_bucket) {
```
geirst, line 724: Consider fixing function name (move 'not').

geirst (approved):
👍

storage/src/vespa/storage/bucketdb/btree_lockable_map.hpp
```diff
@@ -0,0 +1,509 @@
...
+        return EntryRef(value & 0xffffffffULL);
+    }
+    static ValueType make_invalid_value() {
+        return ValueType();
+    }
+    static uint64_t store_and_wrap_value(DataStoreType& store, const ValueType& value) noexcept {
+        return store.addEntry(value).ref();
+    }
+    static void remove_by_wrapped_value(DataStoreType& store, uint64_t value) noexcept {
+        store.freeElem(entry_ref_from_value(value), 1);
```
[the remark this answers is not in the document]
toregge: Use holdElem to ensure it goes through hold cycle (exists with stable value while any readers might access it).

vekterli:
@geirst PTAL

geirst (approved):
👍

PR #13708, opened 2020-06-25
Copy application package atomically

hmusum:
Fix bug where we don't do this atomically, looks like this can happend
when application package is copied at the same time as config server is shut down

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -41,6 +41,8 @@ import org.apache.curator.framework.recipes.cache.PathChildrenCacheEvent;
 import java.io.File;
 import java.io.FilenameFilter;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 import java.time.Clock;
 import java.time.Duration;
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -515,6 +517,20 @@ private ApplicationPackage createApplicationPackage(File applicationFile, Applic
         return applicationPackage;
     }
 
+    private void copyApp(File sourceDir, File destinationDir) throws IOException {
+        if (destinationDir.exists())
+            throw new RuntimeException("Destination dir " + destinationDir + " already exists");
+        if (! sourceDir.isDirectory())
+            throw new IllegalArgumentException(sourceDir.getAbsolutePath() + " is not a directory");
+
+        // Copy app it atomically: Copy to default tmp dir and move to destination
```
gjoranv, line 526: remove 'it'?

gjoranv (approved):
Great!

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: thanks, will fix later

PR #13709, opened 2020-06-25
Surrogate aware gram splitting

bratseth:


arnej27959 (approved):
please do not commit crimes against humanity :-)

PR #13711, opened 2020-06-25
Only sort once during ideal group calculations

vekterli:
@baldersheim please review
@geirst FYI

Avoids invoking `std::sort` _O(n)_ times in favor of just once.

Benchmark for 150 groups of 1 node each:

* Before: 0.0004381478 seconds per invocation
* After:  0.0000377917 seconds per invocation


baldersheim (approved):
👍 This was identical to the solution I ended up with. So this must be close to perfect.

PR #13714, opened 2020-06-26
Avoid adding local session to session cache twice

hmusum:
createLocalSessionUsingDistributedApplicationPackage already adds the
session to session cache.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -599,7 +599,7 @@ public class SessionRepository {
         return new TenantFileSystemDirs(componentRegistry.getConfigServerDB(), tenantName).getUserApplicationDir(sessionId);
     }
 
-    public void addWatcher(long sessionId, Curator.FileCache fileCache, RemoteSession remoteSession, Optional<LocalSession> localSession) {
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -596,7 +599,7 @@ private File getSessionAppDir(long sessionId) {
         return new TenantFileSystemDirs(componentRegistry.getConfigServerDB(), tenantName).getUserApplicationDir(sessionId);
     }
 
-    private void addWatcher(long sessionId, Curator.FileCache fileCache, RemoteSession remoteSession, Optional<LocalSession> localSession) {
+    public void addWatcher(long sessionId, Curator.FileCache fileCache, RemoteSession remoteSession, Optional<LocalSession> localSession) {
```
gjoranv, line 602: Intentional?

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: No, that was leftover from another take at this where I needed the method to be public. Thanks, will fix

gjoranv (approved):

PR #13716, opened 2020-06-26
Avoid going out of memory when deleting files

freva:
Host-admin would go out of memory when deleting a directory with 4M files.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
@@ -110,17 +110,19 @@ public class FileFinder {
      * @return true iff anything was matched and deleted
      */
     public boolean deleteRecursively(TaskContext context) {
+        final int limit = 20;
+        int[] numDeleted = { 0 };
         List<Path> deletedPaths = new ArrayList<>();
 
         try {
             forEach(attributes -> {
                 if (attributes.unixPath().deleteRecursively()) {
-                    deletedPaths.add(attributes.path());
+                    if (numDeleted[0]++ <= limit) deletedPaths.add(attributes.path());
                 }
             });
         } finally {
-            if (deletedPaths.size() > 20) {
-                context.log(logger, "Deleted " + deletedPaths.size() + " paths under " + basePath);
+            if (numDeleted[0] > limit) {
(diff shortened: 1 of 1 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
```diff
@@ -110,17 +110,19 @@ public FileFinder maxDepth(int maxDepth) {
      * @return true iff anything was matched and deleted
      */
     public boolean deleteRecursively(TaskContext context) {
+        final int limit = 20;
+        int[] numDeleted = { 0 };
```
bratseth, line 114: Why not use a for loop on line 118 instead to avoid this hack?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
```diff
@@ -110,17 +110,19 @@ public FileFinder maxDepth(int maxDepth) {
      * @return true iff anything was matched and deleted
      */
     public boolean deleteRecursively(TaskContext context) {
+        final int limit = 20;
```
bratseth, line 113: Maybe worth a comment saying why there is a limit?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
freva: At its core, this class only has a way a method to run a lambda on a matching file via `java.nio.file.Files::walkFileTree`, I don't see any simple way to create an `Iterable` out of that?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
freva: Renamed the variable to `maxNumberOfDeletedPathsToLog`

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/FileFinder.java
bratseth: Ah, right. Didn't look at the context. Then I suggest adding a comment, or using MutableInteger.

bratseth (approved):

PR #13720, opened 2020-06-26
Modernize to c++11 style loops.

baldersheim:
@vekterli PR

### DIFF
--- a/vdslib/src/vespa/vdslib/distribution/group.cpp
+++ b/vdslib/src/vespa/vdslib/distribution/group.cpp
@@ -144,8 +144,8 @@ Group::setNodes(const std::vector<uint16_t>& nodes)
 const Group*
 Group::getGroupForNode(uint16_t nodeIdx) const
 {
-    for (uint32_t i = 0; i < _nodes.size(); ++i) {
-        if (_nodes[i] == nodeIdx) {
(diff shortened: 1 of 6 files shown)


vdslib/src/vespa/vdslib/distribution/group.cpp
```diff
@@ -39,11 +39,9 @@ Group::Group(uint16_t index, vespalib::stringref name,
...
 Group::~Group()
 {
-    for (std::map<uint16_t, Group*>::iterator it = _subGroups.begin();
-         it != _subGroups.end(); ++it)
-    {
-        delete it->second;
-        it->second = 0;
+    for (auto & subGroup : _subGroups) {
+        delete subGroup.second;
+        subGroup.second = 0;
```
vekterli, line 44: Consider `nullptr` instead of literal 0 (or just remove it entirely since it's a destructor anyway...)

vekterli (approved):
:+1:

vdslib/src/vespa/vdslib/distribution/group.cpp
baldersheim: Fixed++ PTAL

vekterli (approved):

PR #13727, opened 2020-06-28
Config server refactoring, remove TenantHandlerProvider

hmusum:


gjoranv (commented):
🏆 It is unclear to my why the Tenant knows about the RequestHandler at all.

PR #13729, opened 2020-06-29
filter invalid UTF-8 (including encoded surrogates) to make protobuf …

arnej27959:
…happy

@geirst please review
@havardpe @vekterli FYI

if any of you know where we send "string" type fields through protobuf and it's possible that we can produce invalid UTF-8, we will need to add similar filtering there.

### DIFF
--- a/logd/src/logd/proto_converter.cpp
+++ b/logd/src/logd/proto_converter.cpp
@@ -1,6 +1,7 @@
 // Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "proto_converter.h"
+#include <vespa/vespalib/text/utf8.h>
 
 using ns_log::LogMessage;
 using ns_log::Logger;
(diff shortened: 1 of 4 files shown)


logd/src/logd/proto_converter.cpp
```diff
@@ -59,7 +60,20 @@ ProtoConverter::log_message_to_proto(const LogMessage& message, ProtoLogMessage&
     proto.set_service(message.service());
     proto.set_component(message.component());
     proto.set_level(convert_level(message.level()));
-    proto.set_payload(message.payload());
+    const std::string &payload = message.payload();
+    vespalib::Utf8Reader reader(payload.c_str(), payload.size());
```
geirst, line 64: Consider moving the filtering to a helper function to make the code easier to read.

geirst (approved):
👍

logd/src/logd/proto_converter.cpp
arnej27959: please take another look now

geirst (approved):
👍

PR #13735, opened 2020-06-29
Only sort once during ideal group calculations

vekterli:
@baldersheim please review. Same as https://github.com/vespa-engine/vespa/pull/13711 with two changes:

1. `tmpResults` is bounded by the calculated group redundancy factor rather than the parent redundancy factor. Means that we don't erroneously return too many groups back to the caller. This was a semantical oversight in the previous PR. A test has been added to verify the new code generates the same output as the old code for the internally failing system test.
2. Reserve `tmpResults` buffer equal to the number of sub-groups, as that's always the number of elements that will be appended prior to sorting and pruning.

Avoids invoking `std::sort` _O(n)_ times in favor of just once.

Benchmark for 150 groups of 1 node each:

* Before: 0.0004381478 seconds per invocation
* After:  0.0000377917 seconds per invocation


baldersheim (approved):

baldersheim:
Merge at will

PR #13737, opened 2020-06-29
Partially disable alloc test when using address sanitizer.

toregge:
@geirst : please review
@vekterli : FYI


toregge:
I suspect that when address sanitizer is enabled then extra mmap/munmap might occur due to the address sanitizer libraries or extra instrumentation code, breaking some of the assumptions in the disabled tests.

geirst (approved):
👍

PR #13741, opened 2020-06-29
Rename unique feature to globalsequence.

baldersheim:
Invert so that the first doc from the first node will sort first,
followed by the first doc from the second node, and the the last doc of the last node at the end.
It will produce number in the range (1 << 48) down to zero.
Sequence = (1 << 48) - ((docId << 16) | disributionKey)

@havardpe and @geirst PR


searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
```diff
@@ -0,0 +1,64 @@
...
+void
+GlobalSequenceBlueprint::visitDumpFeatures(const IIndexEnvironment &, IDumpFeatureVisitor &) const
+{
+}
+
+bool
+GlobalSequenceBlueprint::setup(const IIndexEnvironment & env, const ParameterList & )
+{
+    _distributionKey = env.getDistributionKey();
+    assert( _distributionKey < 0x80000);
```
havardpe, line 47: what about the 3 bits we allow to overlap between distribution key and docid?

searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
```diff
@@ -0,0 +1,64 @@
...
+    uint32_t _distributionKey;
+
+public:
+    GlobalSequenceExecutor(uint32_t distributionKey)
+        : _distributionKey(distributionKey)
+    {
+    }
+
+    void execute(uint32_t docId) override {
+        outputs().set_number(0, ((1ul << 48u) - ((uint64_t(docId) << 16u) | _distributionKey)));
```
havardpe, line 26: consider moving the magic to an inlined function in the header that can also be called by the test

searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+};
+
+}
+
+GlobalSequenceBlueprint::GlobalSequenceBlueprint() :
+    Blueprint("globalsequence"),
```
havardpe, line 33: consider camel case to match naming of other features

geirst: Yes, the name should use came case.

searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
baldersheim: Ouch, thanks. That should be 0x10000

searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
baldersheim: globalsequence -> globalSequence

searchlib/src/vespa/searchlib/features/global_sequence_feature.cpp
baldersheim: Fixed

PR #13743, opened 2020-06-29
Separate documentapi artifacts 2

gjoranv:
This replaces #13699 which moved a lot of code files. The original purpose was to make the bundle-plugin fail when a bundle embeds the `component` artifact. 

This simplifies the dependency graph even further than the previous PR, and it should now be possible to deploy `container-documentapi` as a slim bundle. The work required would be to set its scope to 'provided' in container-core (or even removing it) and adding it to container-dev and all the modules that now get it transitively from container-core.

### DIFF
--- a/config-model/pom.xml
+++ b/config-model/pom.xml
@@ -137,7 +137,7 @@
     </dependency>
     <dependency>
       <groupId>com.yahoo.vespa</groupId>
-      <artifactId>documentapi</artifactId>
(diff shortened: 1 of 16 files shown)


bjorncs (approved):

gjoranv:
FYI: @vekterli, please note the pom.xml change. All dependencies should now be added to `documenapi-dependencies` to avoid having to duplicate.

PR #13747, opened 2020-06-30
Handle app being deleted while maintainer runs

hmusum:


gjoranv (approved):
Good catch. But can the session change state between the time we check for null and until we retrieve the app reference? Can the app reference then be null, or something other bad?

hmusum:
It can change state and the app reference can be null, but we handle that.

PR #13749, opened 2020-06-30
Flip distributor B-tree DB config flag to true by default

vekterli:
@geirst please review. Also need to remove the config model feature flag before this has any effect in practice, but one step at a time.


geirst (approved):
👍

baldersheim (approved):
You should also change the default for the feature flag in Flags.java and/or remove the feature flag too. You can remove everything except removing the interface method which you should just make default and return true.

PR #13754, opened 2020-06-30
Add bindings for prometheus handler in application containers.

gjoranv:
+ Simplify by using common path for json and prometheus in the
  internal applicationmetrics api.


Oracien (approved):
Looks good. The final endpoint for Metrics being named METRICS_VALUES_PATH and METRICS_V1_PATH also being the endpoint for both prometheus and metrics makes it a bit confusing, but not really possible to change that without renaming exposed metrics.

PR #13758, opened 2020-06-30
Add support for serializing and compressing once, instead of once per…

baldersheim:
… backend node.

@bratseth PR


bratseth (approved):
LGTM!

Did some minor changes

baldersheim:
Good, thanks.

PR #13765, opened 2020-07-01
Use find_package to find gtest library.

toregge:
@geirst : please review
@arnej27959 : FYI

### DIFF
--- a/searchlib/src/tests/queryeval/wrappers/CMakeLists.txt
+++ b/searchlib/src/tests/queryeval/wrappers/CMakeLists.txt
@@ -1,11 +1,12 @@
 # Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
+find_package(GTest REQUIRED)
 vespa_add_executable(searchlib_wrappers_test_app TEST
     SOURCES
     wrappers_test.cpp
     DEPENDS
     searchlib
     searchlib_test
-    gtest
+    GTest::GTest
 )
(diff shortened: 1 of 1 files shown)


geirst (approved):
👍

arnej27959:
do we really need to do this for every unit test that use gtest? if so, that's not good...

toregge:
vespa_add_target_dependency fails if ${OTHER_TARGET} is not a proper target. When trying to speed up compilation of vespa on a machine with many cpu cores, one strategy is to make more object libraries.  Doing it automatically based on static libraries, shared libraries or executables might end up with code like the following snippet from a WIP version of vespa_add_target_dependency to avoid cmake errors due to trying to retrieve target properties from a non-target.
```
string(REGEX REPLACE ".*TARGET_OBJECTS:.*" "TARGET" OTHER_TARGET_REPLACED ${OTHER_TARGET})
if(NOT OTHER_TARGET STREQUAL dl AND NOT OTHER_TARGET STREQUAL atomic AND NOT
    OTHER_TARGET STREQUAL gtest AND NOT OTHER_TARGET STREQUAL m AND NOT
    OTHER_TARGET STREQUAL rt AND NOT OTHER_TARGET_REPLACED STREQUAL "TARGET" AND NOT
    OTHER_TARGET STREQUAL "stdc++fs")
    target_include_directories(${TARGET} PRIVATE $<TARGET_PROPERTY:${OTHER_TARGET},INTERFACE_INCLUDE_DIRECTORIES>)
endif()
```
At this time, all targets are not defined, so we cannot use `if(TARGET ${OTHER_TARGET})`constructs to detect what is a target.

PR #13766, opened 2020-07-01
Skip dispatching on reply in FileStorManager

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
@@ -28,7 +28,7 @@ namespace storage {
 FileStorManager::
 FileStorManager(const config::ConfigUri & configUri, const spi::PartitionStateList& partitions,
                 spi::PersistenceProvider& provider, ServiceLayerComponentRegister& compReg)
-    : StorageLinkQueued("File store manager", compReg),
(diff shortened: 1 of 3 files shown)


storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
```diff
@@ -765,13 +765,13 @@ FileStorManager::sendReply(const std::shared_ptr<api::StorageReply>& reply)
     // Currently we need to dispatch due to replies sent by remapQueue
     // function in handlerimpl, as filestorthread keeps bucket db lock
     // while running this function
-    dispatchUp(reply);
+    sendUp(reply);
```
vekterli, line 768: The comment above this call implies that not dispatching could cause a deadlock. Has the threading/locking been changed sufficiently that this no longer is a problem? If so, the comment should be removed. If not, some more pondering might be needed.

vekterli (commented):
Should resolve an existing dispatch-related comment first

storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
baldersheim: You are right. Reverting and just doing GC of members.

PR #13767, opened 2020-07-01
filter log messages from plugins

arnej27959:
* if the "--internal" (or "-i") flag is specified,
  check log messages coming from a "Container" component
  and only print them if they match one of our internal
  package name prefixes.

@bratseth please review


bratseth (approved):
Great!

PR #13768, opened 2020-07-01
Bucket manager also dispatches directly.

baldersheim:
@vekterli PR
Just review, I will coordinate with #13766


vekterli (approved):
:+1: This code is a bit finicky as it involves intercepting mutation replies and enqueuing them if there is at least one concurrent bucket info fetch in progress, sending it all in bulk once all fetches are done. But from what I can tell this _shouldn't™_ change any ordering semantics.

baldersheim:
That was my conclusion too, so I think this one should be fine.

PR #13773, opened 2020-07-01
Since there is not a well defined contract for the the sendReply inte…

baldersheim:
…rface,

I add a stricter sendReplyDirectly interface where the caller guarantees that he has no hidden
requirements that the calle should be aware of.
This will avoid a task switch when propagating the reply.

@vekterli PR
This is a safer alternative to #13766


vekterli (approved):
:+1:

PR #13779, opened 2020-07-02
Bratseth/healt check improvements

bratseth:
@baldersheim see the commit message on the first commit for an explanation

### DIFF
--- a/config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
+++ b/config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
@@ -423,21 +423,26 @@ public class ConfigSubscriber implements AutoCloseable {
      * @return The handle of the config
      * @see #startConfigThread(Runnable)
      */
-    public <T extends ConfigInstance> ConfigHandle<T> subscribe(final SingleSubscriber<T> singleSubscriber, Class<T> configClass, String configId) {
-        if (!subscriptionHandles.isEmpty())
-            throw new IllegalStateException("Can not start single-subscription because subscriptions were previously opened on this.");
-        final ConfigHandle<T> handle = subscribe(configClass, configId);
-        if (!nextConfig())
-            throw new ConfigurationRuntimeException("Initial config of " + configClass.getName() + " failed.");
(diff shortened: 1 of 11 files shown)


container-core/src/main/java/com/yahoo/container/handler/ClustersStatus.java
```diff
@@ -37,11 +39,15 @@ public ClustersStatus() { }
...
     private final Map<String, Boolean> clusterStatus = new HashMap<>();
 
-    public void setContainerHasClusters(boolean containerHasClusters) {
+    /** Sets the current clusters of this container */
+    public void setClusters(Set<String> clusters) {
         synchronized (mutex) {
-            this.containerHasClusters = containerHasClusters;
-            if ( ! containerHasClusters)
-                clusterStatus.clear(); // forget container clusters which was configured away
+            this.containerHasClusters = clusters.size() > 0;
```
baldersheim, line 45: Use isEmpty ?

config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
```diff
@@ -423,21 +423,26 @@ public boolean isClosed() {
...
         startConfigThread(() -> {
                 while (!isClosed()) {
                     try {
-                        if (nextConfig()) {
-                            if (handle.isChanged()) singleSubscriber.configure(handle.getConfig());
-                        }
-                    } catch (Exception e) {
-                        log.log(SEVERE, "Exception from config system, continuing config thread: " + Exceptions.toMessageString(e));
+                        if (nextConfig() && handle.isChanged())
+                            singleSubscriber.configure(handle.getConfig());
```
baldersheim, line 440: You probably need a separate catch for nextConfig() that is SEVERE to ensure you you only blame the subscriber when it is truly to blame.

baldersheim (approved):

container-core/src/main/java/com/yahoo/container/handler/ClustersStatus.java
bratseth: I actually think this is slightly clearer by avoiding negation.

config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
bratseth: Good point, done.

PR #13783, opened 2020-07-02
Consider currentDockerImage/wantedDockerImage if container is empty

olaaun:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/StorageMaintainer.java
```diff
@@ -199,4 +201,14 @@ private String getMicrocodeVersion() {
 
         return results[1].trim();
     }
+
+    private Optional<String> getDockerImage(NodeAgentContext context, Optional<Container> container) {
+        return Stream.of(
+                    container.map(c -> c.image.asString()),
+                    context.node().currentDockerImage().map(DockerImage::asString),
+                    context.node().wantedDockerImage().map(DockerImage::asString))
```
mpolden, line 209: Consider returning `<none>` instead of wanted, as the latter might be misleading.

mpolden (approved):

mpolden (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/StorageMaintainer.java
freva: FWIW, current is potentially equally misleading. If we are upgrading Vespa and the container crashes on startup with the new version, current will still be the old version.

PR #13784, opened 2020-07-02
Create generic flags for IPs we want to block

oyving:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -302,6 +302,12 @@ public class Flags {
             APPLICATION_ID
     );
 
+    public static final UnboundBooleanFlag ONLY_PUBLIC_ACCESS = defineFeatureFlag(
+            "enable-public-only", false,
+            "Only access public hosts from container",
+            "Takes effect on next tick"
+    );
+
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -295,19 +295,25 @@
             APPLICATION_ID
     );
 
-    public static final UnboundBooleanFlag ONLY_PUBLIC_ACCESS = defineFeatureFlag(
```
hakonhall, line 298: The overrides for this flag in the public systems should be removed first?

oyving: I was thinking to remove the flag once the code change went out so that it remains valid until we have released this code.

hakonhall: The common strategy to remove a flag is 1. stop using the flag and wait for that to roll out everywhere, 2. remove flag overrides, and 3. remove flag def.  (2) and (3) can be reversed but I believe it may break the push of flag pushes and/or make it harder to remove the flag.

hakonhall (approved):

andreer (approved):

hakonhall (approved):

PR #13786, opened 2020-07-02
Unify content node and distributor B-tree databases

vekterli:
@geirst please review

* Rewrite distributor B-tree DB in terms of generic implementation
* Move out merger and read guard types so they can be used with
  both content and distributor databases.


geirst (approved):
👍  @toregge please take a look as well

toregge (approved):

PR #13791, opened 2020-07-02
Add feature flag for content node B-tree bucket DB (and remove distributor B-tree flag)

vekterli:
@baldersheim please review. Does this suffice for safely removing the deprecated distributor flag?


baldersheim:
This looks good. However I would probably remove wait with this one until the flag has been removed from hosted-feature-flags. That should you can do as soon as the release containing the change of default value rolled out to hosted infrastructure, which I think happened this morning.

baldersheim:
Now this one should be ready to roll.

PR #13800, opened 2020-07-04
Log time used in distributeApplicationPackage step when preparing

hmusum:


gjoranv (commented):
Where is it logged?

gjoranv (approved):

hmusum:
It it logged after the steps are run, https://github.com/vespa-engine/vespa/blob/master/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java#L144-L145

PR #13807, opened 2020-07-06
use fixed port for rebinding test

arnej27959:

@vekterli please review and merge


vekterli (approved):
:+1: is this port within the "reserved" range for cluster controller tests? I seem to recall us having such a list of ranges, but not where it is anymore...

PR #13811, opened 2020-07-06
Take session lock in more places

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -142,7 +142,7 @@ public class SessionRepository {
             try {
                 addLocalSession(createSessionFromId(Long.parseLong(session.getName())));
             } catch (IllegalArgumentException e) {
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -206,15 +206,13 @@ private boolean isActiveSession(LocalSession candidate) {
 
     public void deleteLocalSession(LocalSession session) {
         long sessionId = session.getSessionId();
-        try (Lock lock = lock(sessionId)) {
```
gjoranv, line 209: Why was taking the lock removed here?

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: I mistakenly thought the lock was not reentrant, so I think it is OK to add it back again

gjoranv (approved):

PR #13818, opened 2020-07-07
Async handling of gid/lid changes.

baldersheim:
@toregge and @geirst PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
+++ b/searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
@@ -25,23 +25,23 @@ GidToLidChangeListener::~GidToLidChangeListener()
 }
 
 void
-GidToLidChangeListener::notifyPutDone(document::GlobalId gid, uint32_t lid)
+GidToLidChangeListener::notifyPutDone(Context context, document::GlobalId gid, uint32_t lid)
 {
-    std::promise<void> promise;
-    auto future = promise.get_future();
     _attributeFieldWriter.executeLambda(_executorId,
-                                        [this, &promise, gid, lid]() { _attr->notifyReferencedPut(gid, lid); promise.set_value(); });
-    future.wait();
+                                        [this, context=std::move(context), gid, lid]() {
+                                            (void) context;
+                                            _attr->notifyReferencedPut(gid, lid);
+                                        });
 }
 
 void
-GidToLidChangeListener::notifyRemove(document::GlobalId gid)
+GidToLidChangeListener::notifyRemove(Context context, document::GlobalId gid)
 {
-    std::promise<void> promise;
-    auto future = promise.get_future();
     _attributeFieldWriter.executeLambda(_executorId,
-                                        [this, &promise, gid]() { _attr->notifyReferencedRemove(gid); promise.set_value(); });
-    future.wait();
+                                        [this, context = std::move(context), gid]() {
+                                            (void) context;
+                                            _attr->notifyReferencedRemove(gid);
+                                        });
 }
 
 void
@@ -50,7 +50,10 @@ GidToLidChangeListener::notifyRegistered()
     std::promise<void> promise;
     auto future = promise.get_future();
     _attributeFieldWriter.executeLambda(_executorId,
-                                        [this, &promise]() { _attr->populateTargetLids(); promise.set_value(); });
+                                        [this, &promise]() {
+                                            _attr->populateTargetLids();
+                                            promise.set_value();
+                                        });
     future.wait();
 }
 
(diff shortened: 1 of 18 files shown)


searchcore/src/vespa/searchcore/proton/server/storeonlyfeedview.cpp
```diff
@@ -632,8 +633,10 @@ StoreOnlyFeedView::adjustMetaStore(const DocumentOperation &op, const GlobalId &
                 putMetaData(_metaStore, docId, op, _params._subDbType == SubDbType::REMOVED);
             }
         } else if (op.getValidPrevDbdId(_params._subDbId)) {
-            _gidToLidChangeHandler.notifyRemove(gid, serialNum);
+            vespalib::Gate gate;
```
toregge, line 636: Consider using std::promise&lt;void&gt;, std::future&lt;void&gt;, and a context that just sets the promise in the destructor.

searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
```diff
@@ -25,23 +25,23 @@ GidToLidChangeListener::~GidToLidChangeListener()
...
 void
-GidToLidChangeListener::notifyPutDone(document::GlobalId gid, uint32_t lid)
+GidToLidChangeListener::notifyPutDone(Context context, document::GlobalId gid, uint32_t lid)
 {
-    std::promise<void> promise;
-    auto future = promise.get_future();
     _attributeFieldWriter.executeLambda(_executorId,
-                                        [this, &promise, gid, lid]() { _attr->notifyReferencedPut(gid, lid); promise.set_value(); });
-    future.wait();
+                                        [this, context=std::move(context), gid, lid]() {
```
toregge, line 31: There might be problems if the listener can be destroyed before the task has been completed. Consider extending the destructor to sync the executor (_attributeFieldWriter) in order to ensure that all previously queued tasks have a valid listener. An alternate approach might be capture a copy of `_attr` in the lambda.

searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
```diff
@@ -25,23 +25,23 @@ GidToLidChangeListener::~GidToLidChangeListener()
...
 void
-GidToLidChangeListener::notifyRemove(document::GlobalId gid)
+GidToLidChangeListener::notifyRemove(Context context, document::GlobalId gid)
 {
-    std::promise<void> promise;
-    auto future = promise.get_future();
     _attributeFieldWriter.executeLambda(_executorId,
-                                        [this, &promise, gid]() { _attr->notifyReferencedRemove(gid); promise.set_value(); });
-    future.wait();
+                                        [this, context = std::move(context), gid]() {
```
toregge, line 41: Same issue has in notifyPutDone.

toregge (changes requested):

searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
baldersheim: Fixed by sync in destructor.

searchcore/src/vespa/searchcore/proton/reference/gid_to_lid_change_listener.cpp
baldersheim: Fixed by sync in destructor.

toregge (approved):

toregge (approved):

searchcore/src/vespa/searchcore/proton/reference/i_gid_to_lid_change_listener.h
```diff
@@ -16,9 +16,10 @@ namespace proton {
 class IGidToLidChangeListener
 {
 public:
+    using Context = std::shared_ptr<search::IDestructorCallback>;
```
geirst, line 19: I would prefer "DestructorCallbackSP" (or similar) instead of "Context", which is very generic. Then it is possible to look at a function signature and better understand what the first parameter is representing.

geirst (commented):
👍

searchcore/src/vespa/searchcore/proton/reference/i_gid_to_lid_change_listener.h
baldersheim: Fixed in #13828

searchcore/src/vespa/searchcore/proton/server/storeonlyfeedview.cpp
baldersheim: I think it is clearer to use the Gate than the general promise/future trick.

PR #13819, opened 2020-07-07
Vekterli/basic snapshot support for content node bucket db

vekterli:
@geirst please review. With these changes, metric and status aggregation will be lock-free when running with the B-tree DB implementation.

* Expose `ReadGuard` via `AbstractLockableMap` interface
* Add B-tree snapshot read guard impl
* Add placeholder wrapper read guard for legacy DB
* Enforce value const-ness of existing `for_each_chunked` iteration API
* Return read guard entries by value instead of modifying ref argument
* Replace chunked iteration with read guards for metric and status aggregation


geirst (approved):
👍  @toregge FYI

toregge (approved):

PR #13823, opened 2020-07-07
Don't log invalid requests for filter execution on servlets

bjorncs:
Fix bug where a malformed uri to servlet/JAX-RS resource would trigger
an unhandled request exception. Jetty logs unhandled exceptions from
filters/servlets as warning with full stack trace.

FYI @gjoranv


gjoranv (commented):
👍

PR #13825, opened 2020-07-07
New configs for bundles

gjoranv:
This is work in progress, but it's ok to review what's been done so far. The goal is to simplify the bundle loading code in the container.


config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
```diff
@@ -108,6 +110,11 @@ public void getConfig(BundlesConfig.Builder builder) {
         }
     }
 
+    @Override
+    public void getConfig(PlatformBundlesConfig.Builder builder) {
+        bundles.forEach(builder::bundles);
```
bjorncs, line 115: Will the final config produced contain the changes from same method in `ContainerCluster`?

bjorncs (approved):

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
gjoranv: Yes, this is a container, which is a child of the cluster. The new method is just a copy of the method above that produced the old config.

PR #13837, opened 2020-07-08
Handle no active session existing better

hmusum:
* Add check for application existing and throw with exception
  about missing application id if missing
* Do not throw exception in activeSessionOf(), requireActiveSessionOf() does
  that and can be used when necessary
* tenant.getApplicationRepo() can never return null


baldersheim (approved):

gjoranv (commented):
👍

PR #13841, opened 2020-07-08
Skip communication manager thread for requests.

baldersheim:
@vekterli PR
@geirst FYI


storage/src/vespa/storage/common/storagelink.cpp
```diff
@@ -165,9 +165,9 @@ void StorageLink::sendUp(const shared_ptr<StorageMessage> & msg)
         default:
             LOG(error, "Link %s trying to send %s up while in state %s",
                 toString().c_str(), msg->toString(true).c_str(), stateToString(getState()));
-            assert(false);
+            return;
```
vekterli, line 168: Is it safe to remove these abort-branches? If the link is not in a state where it's safe to process certain types of messages, the behavior is likely undefined

messagebus/src/vespa/messagebus/messagebus.cpp
```diff
@@ -112,7 +112,7 @@ MessageBus::MessageBus(INetwork &net, const MessageBusParams &params) :
     _routingTables(),
     _sessions(),
     _protocolRepository(std::make_unique<ProtocolRepository>()),
-    _msn(std::make_unique<Messenger>()),
+    _msn(std::make_unique<Messenger>(true, true)),
```
vekterli, line 115: Is it currently safe to have thread-skipping be the default behavior for this constructor overload? Or should this be `false, false` instead?

vekterli (approved):
:+1: a couple of questions that would be good to resolve prior to merging

storage/src/vespa/storage/storageserver/communicationmanager.cpp
```diff
@@ -437,6 +443,18 @@ CommunicationManager::process(const std::shared_ptr<api::StorageMessage>& msg)
     }
 }
 
+void
+CommunicationManager::optionalEnqueue(std::shared_ptr<api::StorageMessage> msg)
```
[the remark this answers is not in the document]
baldersheim: Fixed

messagebus/src/vespa/messagebus/messagebus.cpp
baldersheim: This constructor is only used by tests. My thought is that everything except distributor and proton should run with future defaults.

storage/src/vespa/storage/common/storagelink.cpp
baldersheim: It was initially removed, as there was a test failing in communication manager. However there is no longer need for this as the test itself was fixed.

PR #13846, opened 2020-07-09
Delete expired locks (older than 1 day)

hmusum:
* There has never been any cleanup of locks, but until we started taking
  a lock per session this has not been seen as an issue, since there were
  few of them
* Add code to maintainer that deletes locks older than 1 day

Due to the way locks are implemented and (they are removed if there is just one process that takes the lock and releases it) there are no test added for this change.  Tested in a config server cluster in CD.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
@@ -38,6 +38,6 @@ public class SessionsMaintainer extends ConfigServerMaintainer {
 
         Duration lockExpiryTime = Duration.ofDays(1);
         int deleted = applicationRepository.deleteExpiredLocks(lockExpiryTime);
-        log.log(LogLevel.INFO, "Deleted " + deleted + " expired locks, expiry time " + lockExpiryTime);
+        log.log(LogLevel.INFO, "Deleted " + deleted + " locks older than " + lockExpiryTime);
     }
 }
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
@@ -35,5 +35,9 @@ public class SessionsMaintainer extends ConfigServerMaintainer {
             int deleted = applicationRepository.deleteExpiredRemoteSessions(expiryTime);
             log.log(LogLevel.FINE, "Deleted " + deleted + " expired remote sessions, expiry time " + expiryTime);
         }
+
+        Duration lockExpiryTime = Duration.ofDays(1);
+        int deleted = applicationRepository.deleteExpiredLocks(lockExpiryTime);
+        log.log(LogLevel.INFO, "Deleted " + deleted + " expired locks, expiry time " + lockExpiryTime);
     }
 }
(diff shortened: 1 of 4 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
```diff
@@ -35,5 +35,9 @@ protected void maintain() {
             int deleted = applicationRepository.deleteExpiredRemoteSessions(expiryTime);
             log.log(LogLevel.FINE, "Deleted " + deleted + " expired remote sessions, expiry time " + expiryTime);
         }
+
+        Duration lockExpiryTime = Duration.ofDays(1);
+        int deleted = applicationRepository.deleteExpiredLocks(lockExpiryTime);
+        log.log(LogLevel.INFO, "Deleted " + deleted + " expired locks, expiry time " + lockExpiryTime);
```
gjoranv, line 41: nit: the middle part of the message could be simplified to " locks older than "

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -665,7 +687,7 @@ public Lock lock(long sessionId) {
         return curator.lock(lockPath(sessionId), Duration.ofMinutes(1)); // These locks shouldn't be held for very long.
     }
 
-    private Path lockPath(long sessionId) {
+    public Path lockPath(long sessionId) {
```
gjoranv, line 690: intended? I don't see any new usage in this PR.

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -281,6 +282,27 @@ public int deleteExpiredRemoteSessions(Clock clock, Duration expiryTime) {
...
+            if (zooKeeperNodeCreated(path).isBefore(clock.instant().minus(expiryTime))) {
+                log.log(Level.INFO, "Lock  " + path + " has expired, deleting it");
+                curator.delete(path);
+                deleted++;
+            }
+        }
+        return deleted;
+    }
+
+    private Instant zooKeeperNodeCreated(Path path) {
```
gjoranv, line 298: perhaps this should return an Optional instead, and empty where it now returns the current time?

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: I tried to write a test that needed this (but that was abandoned), so not anymore, will fix

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
hmusum: thanks, will fix

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Yes, probably better, thanks

hmusum:
@gjoranv please take a look again

gjoranv (commented):
👍

PR #13849, opened 2020-07-09
Control maintenance jobs with feature flag

mpolden:
All jobs are currently active in all production environments.

Depends on internal PR.

@hmusum


hmusum (approved):
Looks great

I really want to get https://github.com/vespa-engine/vespa/pull/13846 merged before this one (as contains an important fix and there will be a conflict between these 2 PRs)

PR #13856, opened 2020-07-09
If flush has been triggered due to spread, we need to prioritize that…

baldersheim:
… correctly.

We should never prioritize bloat of single file, unless global bloat is exceeded.
This will significantly reduce number of compaction operations.

@toregge or @geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/docstore/logdatastore.cpp
+++ b/searchlib/src/vespa/searchlib/docstore/logdatastore.cpp
@@ -313,16 +313,23 @@ LogDataStore::compact(uint64_t syncToken)
     }
 }
 
+bool
+LogDataStore::isTotalDiskBloatExceeded() const {
+    const size_t diskFootPrint = getDiskFootprint();
+    const size_t maxConfiguredDiskBloat = diskFootPrint * _config.getMaxDiskBloatFactor();
+    return getDiskBloat() > maxConfiguredDiskBloat;
+}
+
 size_t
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/docstore/logdatastore.cpp
```diff
@@ -313,16 +313,23 @@ LogDataStore::compact(uint64_t syncToken)
...
+    const size_t maxConfiguredDiskBloat = diskFootPrint * _config.getMaxDiskBloatFactor();
+    return getDiskBloat() > maxConfiguredDiskBloat;
+}
+
 size_t
 LogDataStore::getMaxCompactGain() const
 {
+    size_t bloat = 0;
+    if ( isTotalDiskBloatExceeded() ) {
+        bloat = getDiskBloat();
```
toregge, line 328: Consider a strategy to avoid having to call getDiskBloat() again here, e.g. return std::pair<bool, size_t> or just a size_t from a renamed isTotalDiskBloatExceeded().

toregge (approved):

geirst (approved):
👍

searchlib/src/vespa/searchlib/docstore/logdatastore.cpp
baldersheim: Done, in addition to avoid deadlock.

geirst (commented):
👍

PR #13859, opened 2020-07-10
Remove configuration in maven-bundle-plugin of parent/pom.xml file since the related issue had been already fixed. 

markxiaotao:
According to the link https://issues.apache.org/jira/browse/FELIX-5879, I observed that such issue had already been fixed in maven-bundle-plugin (4.1.0). And based on the comment "TODO: remove when bundle-plugin understands java 10". This configuration is not needed in my opinion.

The configuration and comment:
https://github.com/vespa-engine/vespa/blob/a82b26c3b61b01537439fd2bc1f85e04d899e6f1/parent/pom.xml#L85-L90

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -82,12 +82,6 @@
                     <groupId>org.apache.felix</groupId>
                     <artifactId>maven-bundle-plugin</artifactId>
                     <version>${maven-bundle-plugin.version}</version>
-                    <configuration>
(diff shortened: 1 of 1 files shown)


gjoranv:
@markxiaotao Thanks for the reminder! Upgrading the plugin in on my TODO list for the summer, so I'll look into it shortly. The plugin version must be updated before removing the config.

PR #13863, opened 2020-07-10
Use config parameter to find correct path for 'artifacts' subdirectory

bjorncs:
Introduce new config for junit testrunner.
Remove use of ConfigserverConfig and knowledge about system names.

FYI @tokle


hmusum (approved):
Great!

PR #13869, opened 2020-07-13
Require vespa-lz4 and vespa-zstd

aressem:


aressem:
Depends on https://git.vzbuilders.com/vespa/build-dependencies/pull/357 and RPM distribution.

toregge:
Should we drop the dependencies on lz4 and zstd due to having vespa-lz4 and vespa-zstd ? Or do we need both variants ?

aressem:
Pushed a commit now removing them

toregge (approved):

PR #13870, opened 2020-07-13
Shutdown state server and metric manager before service layer

vekterli:
@geirst @toregge please review. I believe this is the most likely reason for a core observed during shutdown on an internal test (I could not get access to the actual core file itself).

Avoids a very small race condition window where metric update hooks
may point into the service layer components even after they have
been destroyed, as these are not explicitly unregistered today.

Another option would be to add unregistering on component destruction,
but that adds another race condition where an external client may
observe partial metric existence during this time window. By shutting
down the metric exporting interfaces first, we should avoid this.

Side note: automatic metric hook unregistering should be added at some point
anyway since they after this change cannot cause partial metric existence.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/proton.cpp
+++ b/searchcore/src/vespa/searchcore/proton/server/proton.cpp
@@ -223,6 +223,7 @@ Proton::Proton(const config::ConfigUri & configUri,
       _initStarted(false),
       _initComplete(false),
       _initDocumentDbsInSequence(false),
+      _has_shut_down_config_and_state_components(false),
       _documentDBReferenceRegistry(std::make_shared<DocumentDBReferenceRegistry>()),
       _nodeUpLock(),
       _nodeUp()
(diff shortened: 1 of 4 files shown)


toregge (approved):

searchcore/src/vespa/searchcore/proton/server/proton.cpp
```diff
@@ -462,6 +462,15 @@ Proton::~Proton()
     LOG(debug, "Explicit destructor done");
 }
 
+void
+Proton::perform_pre_service_layer_shutdown_steps()
+{
+    _stateServer.reset();
```
geirst, line 468: We need to reset _customComponentRootToken and _customComponentBindToken first, as these have reference to a JsonHandlerRepo owned by the StateServer, and will access this when destructed.

geirst: In Proton::~Proton() we do `_executor.sync();` before any of this. I am not sure if we need something similar here. @toregge can you comment on that?

toregge: That depends on what queued tasks can contain references to the structure being destroyed.  We should probably close the config fetcher, disable reconfig and sync the executor, to avoid new DocumentDB instances arriving after we've decided to take down the state server.

vekterli: I have moved config, token, state server and metrics shutdown into a dedicated method. Explicit executor sync is done in both places. PTAL

geirst (approved):
👍

vekterli:
@toregge could you take a look at the last commit as well and see if you spot any potential issues?

toregge (approved):

PR #13876, opened 2020-07-13
Emit log warning on repeated content node bucket info fetch failures

vekterli:
@geirst please review

Currently requires a certain number of repeated failures for a given
cluster state transition. Rationale is that problematic nodes usually
fail for a prolonged amount of time, so it's wise to reduce log noise
from more transient failures. Threshold to be adjusted later as needed.


geirst (approved):
👍

vekterli:
Unrelated Travis error due to move to custom builds of zstd/lz4 (presumably):
```
cmake3: error while loading shared libraries: libzstd.so.1: cannot open shared object file: No such file or directory
```

PR #13884, opened 2020-07-14
Make sure to close input stream

hmusum:

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/handler/LogReader.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/LogReader.java
@@ -96,12 +96,14 @@ class LogReader {
         private final BufferedReader reader;
         private final double from;
         private final double to;
+        private final InputStream inputStream;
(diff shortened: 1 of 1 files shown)


container-core/src/main/java/com/yahoo/container/handler/LogReader.java
```diff
@@ -122,6 +124,7 @@ public LineWithTimestamp next() {
         @Override
         public void close() throws IOException {
             reader.close();
+            inputStream.close();
```
mpolden, line 127: I don't think this is necessary though as closing a `BufferedReader` also closes the underlying reader.

mpolden (approved):

container-core/src/main/java/com/yahoo/container/handler/LogReader.java
hmusum: Yes, you are right. Closing PR

PR #13896, opened 2020-07-15
Add per-operation metrics for puts and removes that are part of merges

vekterli:
@geirst please review
@yngveaasheim @baldersheim FYI

Move all merge-related metrics out into a separate wrapper for convenience.


geirst (approved):
👍

PR #13898, opened 2020-07-15
Load platform bundles separately

gjoranv:
FYI: @baldersheim 

There is still some cleanup that can be left to a future PR.

The bundles.def file will be removed after its use has rolled out of hosted Vespa.


container-di/src/main/java/com/yahoo/container/di/Osgi.java
```diff
@@ -25,6 +25,10 @@ default BundleClasses getBundleClasses(ComponentSpecification bundle, Set<String
         return new BundleClasses(new MockBundle(), Collections.emptySet());
     }
 
+    default void installPlatformBundles(Collection<FileReference> bundles) {
```
bjorncs, line 28: The interface also behaves as a mock implementation of itself? Never seen that before.

bjorncs (approved):

container-di/src/main/java/com/yahoo/container/di/Osgi.java
gjoranv: Perhaps it made sense in Scala :)

gjoranv:
@bjorncs I added a few commits after your review. The only significant one is "Throw exception if ..."

PR #13917, opened 2020-07-17
Revert "Gjoranv/allow preinstall per bundle"

hmusum:
Reverts vespa-engine/vespa#13913


hmusum:
Merged revert in https://github.com/vespa-engine/vespa/pull/13916 and in internal repo instead

PR #13929, opened 2020-07-21
Add endpoint to delete all prod deployments

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackage.java
```diff
@@ -184,4 +189,27 @@ private static String withoutLegacyDir(String name) {
 
     }
 
+    /** Creates a valid application package that will remove all application's deployments */
+    public static ApplicationPackage createEmptyForDeploymentRemoval() {
```
mpolden, line 193: Consider simplifying to `deploymentRemoval()`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -284,6 +284,7 @@ private HttpResponse handleDELETE(Path path, HttpRequest request) {
         if (path.matches("/application/v4/tenant/{tenant}")) return deleteTenant(path.get("tenant"), request);
         if (path.matches("/application/v4/tenant/{tenant}/key")) return removeDeveloperKey(path.get("tenant"), request);
         if (path.matches("/application/v4/tenant/{tenant}/application/{application}")) return deleteApplication(path.get("tenant"), path.get("application"), request);
+        if (path.matches("/application/v4/tenant/{tenant}/application/{application}/deployment")) return removeProdAllDeployments(path.get("tenant"), path.get("application"));
```
mpolden, line 287: What about test/staging? If this indeed only applies to prod, we can perhaps use the path `/application/v4/tenant/{tenant}/application/{application}/environment/prod`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1901,12 +1902,17 @@ private HttpResponse submit(String tenant, String application, HttpRequest reque
                                                             sourceRevision,
                                                             authorEmail,
                                                             sourceUrl,
-                                                            commit,
                                                             projectId,
                                                             applicationPackage,
                                                             dataParts.get(EnvironmentResource.APPLICATION_TEST_ZIP));
     }
 
+    private HttpResponse removeProdAllDeployments(String tenant, String application) {
```
mpolden, line 1910: Consider naming it `removeAllProdDeployments`.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
```diff
@@ -974,6 +971,36 @@ public void testMeteringResponses() {
...
+        app.submit(applicationPackage).deploy();
+        tester.controller().jobController().deploy(app.instanceId(), JobType.devUsEast1, Optional.empty(), applicationPackage);
+
+        assertEquals(Set.of(ZoneId.from("prod.us-west-1"), ZoneId.from("prod.us-east-3"), ZoneId.from("prod.eu-west-1"), ZoneId.from("dev.us-east-1")),
+                app.instance().deployments().keySet());
+
+        tester.assertResponse(request("/application/v4/tenant/tenant1/application/application1/deployment", DELETE)
+                        .userIdentity(USER_ID)
+                        .oktaAccessToken(OKTA_AT).oktaIdentityToken(OKTA_IT),
+                "{\"message\":\"Application package version: 1.0.2-unknown\"}");
```
mpolden, line 999: Would be better if this said what happened (i.e. "all deployments removed"), but not important.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
freva: My understanding is that this will submit the empty application, which will notify the JobController which will asynchronously at some point see that prod deployments have been removed and start making DELETE calls against config servers. It will also figure out that the system/staging tests it might have been running are no longer needed and abort them + delete deployment (similar to what happens when you submit a new package while tests are running).

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
freva: Done

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
freva: Done

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/ApplicationPackage.java
freva: Done

mpolden (approved):

PR #13935, opened 2020-07-22
Rename endpoint scope weighted -> region

mpolden:
No functional changes, just a better name for this kind of endpoints.

@bjorncs


bjorncs (approved):
Nice 👍

PR #13938, opened 2020-07-22
Don't verbose print documents at debug log level

vekterli:
@geirst please review

Causes way too much output to be useful. Move to separate logging that is only visible at spam level.


geirst (approved):
👍

PR #13946, opened 2020-07-29
Upgrade to Curator 2.13.0

hmusum:


zkfacade/src/main/java/com/yahoo/vespa/curator/mock/MockCurator.java.orig
```diff
@@ -0,0 +1,1199 @@
+// Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
```
hakonhall, line 1: Remove stray file?

hmusum: Yes, thanks, fixed

hakonhall (approved):

PR #13952, opened 2020-07-30
Avoid unnecessary fixation of manual clock

hakonhall:


hakonhall:
This commit was supposed to be part of #13950

hmusum (approved):

PR #9954, opened 2019-07-03
Bratseth/dont wait

bratseth:
@vekterli please review

The IO thread both sends documents and handles responses.
At the start of each cycle it would wait for at most 100 ms for incoming documents before progressing to handling responses. This cause the min time of a cycle to be 100 ms. This is usually ok, but not when the client uses the sync API and is both latency and throughput sensitive (depending on whether it writes few or many documents, in the same code path). 

I don't see a good reason to wait 100 ms so this PR just lowers this to 1 ms at line 338 in IOThread.

I initially thought I needed to do something more complicated so there are more changes here but no other changes to functionality.

PR #13966, opened 2020-08-03
Change default poll frequency when idle form 1000hz to 10hz.

baldersheim:
The cost of a poll is huge as it implies sending an http post.
Synchrous usage might want to reduce this.

@bratseth or @bjorncs PR
This reverts the effective change in #9954, but makes it configurable.
Sync usage should use a sync protocol instead.

### DIFF
--- a/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/SyncFeedClient.java
+++ b/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/SyncFeedClient.java
@@ -25,9 +25,16 @@ public class SyncFeedClient implements AutoCloseable {
     private final FeedClient wrappedClient;
     private final Callback callback;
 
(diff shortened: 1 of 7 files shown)


bratseth (commented):
> Sync usage should use a sync protocol instead.

It's not that simple since there is one code path, which some times happen to get many updates, and some times a few for a given poll to the system delivering updates. 

This will bring back the problem fixed by the previous PR ... unless people know which lever to pull to fix it, but that's very hard. Could we think about whether this could be done in a smarter way, how about keeping track of how long it was since we polled?

baldersheim:
@bratseth Now you can have a look again.
Another option would be to use exponential backoff in the range [1-100], Fx 1, 2, 4 in the sync case.

bratseth (approved):
💯

PR #13978, opened 2020-08-04
Avoid doing a full get for metadata only get

baldersheim:
@vekterli or @geirst or @toregge or @arnej27959 or @havardpe PR


baldersheim:
@hakonhall PR Another small one, but not without semantics.

hakonhall (approved):

PR #13983, opened 2020-08-04
Disallow deleting tenants with outstanding charges

olaaun:


oyving:
Can we add a test for this API behavior?

oyving (approved):

PR #13985, opened 2020-08-04
Balder/minor fieldset cleanup

baldersheim:
@bratseth PR


bratseth (commented):
Great, but this breaks public API and should wait for Vespa8, or otherwise we need to at least check if  anybody internal are using it (I guess it's a safe assumption nobody external are). 

Does it help something in particular to do this now?

Also, I think there was something stopping us from doing this on Vespa 7 - do you remember what it was?

baldersheim:
I checked for any internal usage - none found. It should never be as header/body concept was internal vds concepts.
As far as I remember it was API conservatism, we had not deprecated all methods/classes that had to go, nor had we cleaned up all internal usage. There has since been a few rounds doing that. The last one with #13601. I wanted to clean this up now as we have to rework some fieldset handling in the backend. To avoid doing that on stuff that is going away too, I wanted to clean it up first.

bratseth (approved):
Alright, let's do it then

PR #13994, opened 2020-08-05
Stop reading BillingInfo but write the default value

oyving:
Preparing to completely remove BillingInfo as it was never used. First step is to stop reading it.  Once that has propagated we can remove the class completely.

Please take an extra look to make sure we are not going to get lots of issues with deserialising existing objects :)


jonmv (approved):
👍

PR #13995, opened 2020-08-05
Turn off lint for test marco definitions we can actually use clang-ti…

baldersheim:
…dy to cleanup our code.

@hakonhall PR
@vekterli @havardpe @geirst @toregge FYI


baldersheim:
BTW, No semantic changes, just cleanup to reduce number of complaints from Clion so we can actually make use of its quite useful hints.

hakonhall (approved):

PR #13999, opened 2020-08-06
Allow references when expecting numbers

bratseth:


geirst:
@arnej27959 please review after the fact

arnej27959 (commented):
looks good to me, unit test shows what it does

PR #14003, opened 2020-08-06
Extend IndexedSegmentItem instead of SegmentItem in AndSegmentItem

bratseth:
@arnej27959 do you know if there was some thinking behind the decision to not make AndSegmentItem indexed?

I stumbled upon this since it breaks an assumption in the serializer leading to a serialization exception.
If this is ok, we should probably also move most of the And/PhraseSegmentItem code into a common superclass.

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -313,7 +313,7 @@
     "fields": []
   },
   "com.yahoo.prelude.query.AndSegmentItem": {
-    "superClass": "com.yahoo.prelude.query.SegmentItem",
+    "superClass": "com.yahoo.prelude.query.IndexedSegmentItem",
     "interfaces": [
       "com.yahoo.prelude.query.BlockItem"
     ],
@@ -327,6 +327,9 @@
       "public com.yahoo.prelude.query.Item$ItemType getItemType()",
       "public java.lang.String getName()",
       "public java.lang.String getIndexName()",
+      "public void addItem(com.yahoo.prelude.query.Item)",
+      "public void setIndexName(java.lang.String)",
+      "public java.lang.String getIndexedString()",
       "public void setWeight(int)"
     ],
     "fields": []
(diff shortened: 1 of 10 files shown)


arnej27959:
i have looked at this too, and couldn't figure out how the various Segment items are supposed to work at all.  Worse, the connectivity is modeled directly in the Item superclass, but it only makes sense to connect together taggable items, so it should be there instead.

The problem here is both in the modeling (which lets you make inconsistent query trees) and the parser.
My understanding is that this code in StructuredParser:
```
           Item word = phraseWord(indexName, quoted, (firstWord != null) || (composite != null));
[returns AndSegment("m", "s")}
 ...
            if (composite != null) {
                composite.addItem(word);
                connectLastTermsIn(composite);
```
ends up connecting the previous term "m" to the AndSegment.
This worked when phraseWord returned a phrase, but here you should connect the "m" term to the first term inside the AndSegment instead.

So this can be fixed in connectLastTermsIn(), adding something like
```
if (nextToLast instanceof AndSegmentItem) {
    CompositeItem tmp = (CompositeItem) nextToLast;
    nextToLast = tmp.getItem(tmp.getItemCount() - 1);
}
if (last instanceof AndSegmentItem) {
    last = ((AndSegmentItem)last).getItem(0);
}
```

but I don't know if this is enough, or if there are more similar cases that also needs taking care of.

arnej27959 (changes requested):
i don't think the backend will be happy with this, it will serialize as AND but with index and unique id, which isn't expected at all.
@havardpe do you know if the backend can handle that?

havardpe:
> i don't think the backend will be happy with this, it will serialize as AND but with index and unique id, which isn't expected at all.
> @havardpe do you know if the backend can handle that?

Looking at the code (for AND); it seems a unique id will simply be ignored while an index name will break the binary format. This is because the unique id is sent as a tagged value that can be attached to any item while the existence of the index name is determined by the schema for each item type, and AND only has 'arity' in its schema.

arnej27959:
replaced by https://github.com/vespa-engine/vespa/pull/14019

PR #14017, opened 2020-08-10
Avoid creating session path when creating waiters

jonmv:
@hakonhall please review.


hakonhall (approved):

hmusum:
I am not sure if I understand the change, won't both the old and the new code create` .../sessions/n/waiterPath`? `create()` javdoc says `Creates an empty node at a path, creating any parents as necessary.`

PR #14018, opened 2020-08-10
[VESPA-18652] Fix retirement loop

freva:
Analysis of the issue in ticket.

It's a bit tricky to write a unit test since the previous retirement followed the reverse allocation priority, so the node with the largest index would be on the host with most free resources, which often is the spare host.


bratseth (approved):
👍 excellent detective work. 

Abstracting a bit the problem is that the ordering of PrioritizableNode is inconsistent with the ordering used when deciding what to retire. To protect against all such problems I think we should use the same order in both (but reversed in the latter of course). Seems like this could be done simply by adding the index comparison to active nodes in PrioritizableNode?

Anyway, feel free to merge, as this fixes the concrete problem.

PR #14020, opened 2020-08-10
[VESPA-18632] Fix recursive wantToRetire & wantToDeprovision patch

freva:
After #13468, patching a host with both `wantToRetire` and `wantToDeprovision` only works for the host (updating the node silently fails).


bratseth (approved):
Nice ... and yet another case of we should have split nodes and hosts. I think we could do it gradually by creating a Host subclass of Node, and make sure hosts are created as Hosts. Then we could move hosts logic to use Host until we're ready to move host fields and methods down.

PR #14030, opened 2020-08-11
Revert "Reduce the use of hamcrest when normal assertEquals/assertTrue does a…"

baldersheim:
Reverts vespa-engine/vespa#14011
@hmusum PR


hmusum:
👍

PR #14031, opened 2020-08-12
Revert "Reduce the use of hamcrest when normal assertEquals/assertTrue does a…"

aressem:
Reverts vespa-engine/vespa#14028

Suspecting that this broke the config server.


aressem:
Already reverted.

PR #14032, opened 2020-08-12
Revert "Revert "Reduce the use of hamcrest when normal assertEquals/assertTrue does a…""

baldersheim:
Reverts vespa-engine/vespa#14030
@bratseth or @hmusum Seems unrelated to build failure last night.


bratseth:
👍

PR #14039, opened 2020-08-12
- Do not do maintenance in the logger thread. That blocks both delive…

baldersheim:
…ry to the bound Q and holds objects in memory.

  Instead just signal that it might be a good idea to do maintenance.

@arnej27959 PR

### DIFF
--- a/logserver/src/main/java/com/yahoo/logserver/handlers/HandlerThread.java
+++ b/logserver/src/main/java/com/yahoo/logserver/handlers/HandlerThread.java
@@ -2,7 +2,6 @@
 package com.yahoo.logserver.handlers;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.logging.Logger;
 
@@ -65,17 +64,7 @@ public class HandlerThread extends Thread implements LogHandler {
     }
 
     private final BlockingQueue<ItemOrList> queue;
-    private final List<LogHandler> handlers = new ArrayList<LogHandler>();
-    private long count;
-    @SuppressWarnings("unused")
-    private long droppedCount = 0;
-    @SuppressWarnings("unused")
-    private boolean queueWasFull = false;
-    @SuppressWarnings("unused")
-    private long lastDropLogMessage = 0;
-    @SuppressWarnings("unused")
-    private long lastAcceptingLogMessage = 0;
-
+    private final List<LogHandler> handlers = new ArrayList<>();
     public HandlerThread(String name) {
         super(name);
         queue = new LinkedBlockingQueue<>(queueSize);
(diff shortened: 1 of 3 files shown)


logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
```diff
@@ -30,21 +28,28 @@
...
+        synchronized (mutex) {
+            mutex.wait(milliS);
+        }
+    }
 
     private void run() {
         try {
-            Thread.sleep(125000); // 2 m 5 s
             while (true) {
                 while (maintenance()) {
```
arnej27959, line 49: now you do the same anyway, this could be simpler:
```
        while (true) {
                maintenance();
                waitForTrigger(2000);
            }
```

logserver/src/main/java/com/yahoo/logserver/handlers/HandlerThread.java
```diff
@@ -201,10 +182,12 @@ public void run() {
                 drainList.add(queue.take());
                 queue.drainTo(drainList);
 
-                for (ItemOrList o : drainList) {
+                for (int i = 0; i < drainList.size(); i++) {
                     // we can get two types of elements here: single log
                     // messages or lists of log messages, so we need to
                     // handle them accordingly.
+                    ItemOrList o = drainList.get(i);
```
arnej27959, line 189: why do you get() then set() the same thing?

arnej27959 (approved):
functionally seems ok, please re-check two minor details

logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
baldersheim: Good point.

logserver/src/main/java/com/yahoo/logserver/handlers/HandlerThread.java
baldersheim: Argh, null it should be.

baldersheim:
@arnej27959 PTAL

arnej27959 (approved):
👍

logserver/src/main/java/com/yahoo/logserver/handlers/HandlerThread.java
```diff
@@ -191,8 +172,8 @@ public void run() {
             throw new NullPointerException("channel is not allowed to be null");
         }
 
-        // TODO: Make the legmessage elements some kind of composite structure to handle both individual messages and lists uniformly.
```
jonmv, line 194: RIP legmessage :'(

PR #14076, opened 2020-08-18
Move methods to Session and ApplicationRepository

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/session/Session.java
```diff
@@ -27,13 +31,24 @@
...
+    protected final Optional<ApplicationPackage> applicationPackage;
 
     protected Session(TenantName tenant, long sessionId, SessionZooKeeperClient sessionZooKeeperClient) {
         this.tenant = tenant;
         this.sessionId = sessionId;
         this.sessionZooKeeperClient = sessionZooKeeperClient;
+        this.applicationPackage = Optional.empty();
     }
 
+    protected Session(TenantName tenant, long sessionId, SessionZooKeeperClient sessionZooKeeperClient,
```
gjoranv, line 43: This has duplicate code from the constructor above, but assuming it's temporary while refactoring the Session classes.

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/Session.java
hmusum: Yes, will clean up

PR #14077, opened 2020-08-18
Skip logging only for IllegalInputException

bratseth:
- Add IllegalInputException to signal cases where we know the exception
  is caused by illegal input received from the requestor.
- Only skip logging for IllegalInputException instead of the superclass
  IllegalArgumentException as that is also used to signal illegal
  arguments to methods due to bugs which are otherwise hard to debug.
- Throw IllegalInputException rather than IllegalArgumentException
  where appropriate.
- Deprecated QueryException as it was only used to be able to separate
  between query string and query parameter exceptions, and not doing
  that consistently, and is in a package we don't want more use of.
- Clean up some cases where the wrong exception was thrown.


gjoranv (approved):

gjoranv:
Build fails with:
```
[ERROR] /source/config-model/src/test/java/com/yahoo/searchdefinition/derived/NearestNeighborTestCase.java:[30,18] cannot find symbol
  symbol:   class IllegalInputException
  location: class com.yahoo.searchdefinition.derived.NearestNeighborTestCase
```

gjoranv (approved):

Issue #13989, opened 2020-08-04
Debian build


ygrek:
ftr, here is a patch that makes vespa build on debian 10: 

```
diff --git a/build_settings.cmake b/build_settings.cmake
index d7dd26f5ee..d1f28776b0 100644
--- a/build_settings.cmake
+++ b/build_settings.cmake
@@ -25,7 +25,7 @@ set(RUN_BENCHMARKS FALSE CACHE BOOL "If TRUE, benchmarks are run together with t
 set(AUTORUN_UNIT_TESTS FALSE CACHE BOOL "If TRUE, tests will be run immediately after linking the test executable")
 # Warnings
-set(C_WARN_OPTS "-Wuninitialized -Werror -Wall -W -Wchar-subscripts -Wcomment -Wformat -Wparentheses -Wreturn-type -Wswitch -Wtrigraphs -Wunused -Wshadow -Wpointer-arith -Wcast-qual -Wcast-align -Wwrite-strings")
+set(C_WARN_OPTS "-Wuninitialized -Wall -W -Wchar-subscripts -Wcomment -Wformat -Wparentheses -Wreturn-type -Wswitch -Wtrigraphs -Wunused -Wshadow -Wpointer-arith -Wcast-qual -Wcast-align -Wwrite-strings")
 if (VESPA_USE_SANITIZER)
     # Instrumenting code changes binary size, which triggers inlining warnings that
     # don't happen during normal, non-instrumented compilation.
@@ -67,7 +67,7 @@ else()
 endif()
 # C and C++ compiler flags
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O3 -fno-omit-frame-pointer ${C_WARN_OPTS} -fPIC ${VESPA_CXX_ABI_FLAGS} -DXXH_INLINE_ALL -DBOOST_DISABLE_ASSERTS ${VESPA_CPU_ARCH_FLAGS} ${EXTRA_C_FLAGS}")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -O3 -fno-omit-frame-pointer ${C_WARN_OPTS} -fPIC ${VESPA_CXX_ABI_FLAGS} -DBOOST_DISABLE_ASSERTS ${VESPA_CPU_ARCH_FLAGS} ${EXTRA_C_FLAGS}")
 # AddressSanitizer/ThreadSanitizer work for both GCC and Clang
 if (VESPA_USE_SANITIZER)
[...]

PR #14082, opened 2020-08-18
Port to Debian 10.

toregge:
@baldersheim : please review

This contains most changes needed for issue #13989


storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -14,9 +14,11 @@
 #include <vespa/storageapi/message/stat.h>
 #include <vespa/vespalib/stllike/hash_map.hpp>
 #include <vespa/vespalib/util/exceptions.h>
+#ifndef XXH_NO_INLINE_ALL
```
baldersheim, line 17: Since we are controlling this in the build_settings.cmake I think the whole XXH_INLINE stuff here can just go away.

baldersheim (changes requested):
Just remove the XXH_INLINE stuff in filestorhandlerimpl.cpp

baldersheim (approved):

PR #14093, opened 2020-08-19
Config server refactoring, part 27

hmusum:


gjoranv (approved):
Can't we also remove `ApplicationSet.getAllVersions` now?

hmusum:
No, that one is still needed, the `getAllVersions()` methods don't do the same

PR #14095, opened 2020-08-19
Reapply "Throw exception instead of returning null when reading application id"

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/Session.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/Session.java
@@ -135,7 +135,10 @@ public abstract class Session implements Comparable<Session>  {
     }
 
     /** Returns application id read from ZooKeeper. Will throw RuntimeException if not found */
-    public ApplicationId getApplicationId() { return sessionZooKeeperClient.readApplicationId(); }
+    public ApplicationId getApplicationId() {
+        return sessionZooKeeperClient.readApplicationId()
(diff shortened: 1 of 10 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/Session.java
```diff
@@ -134,8 +134,18 @@ public void setAthenzDomain(Optional<AthenzDomain> athenzDomain) {
         sessionZooKeeperClient.writeAthenzDomain(athenzDomain);
     }
 
+    /** Returns application id read from ZooKeeper. Will throw RuntimeException if not found */
     public ApplicationId getApplicationId() { return sessionZooKeeperClient.readApplicationId(); }
 
+    /** Returns application id read from ZooKeeper. Will return Optional.empty() if not found */
+    public Optional<ApplicationId> getOptionalApplicationId() {
+        try {
+            return Optional.of(getApplicationId());
```
jonmv, line 143: It would perhaps be better if the `readApplicationId()` returned an optional, and `getApplicationId()` threw on empty? I fear something else may go wrong, and we then believe that was because the id didn't exist, even though it does.

hmusum: Yes, good point, that's better

hmusum: Fixed

jonmv (approved):

PR #14100, opened 2020-08-19
Wrap tf2onnx to add output aliases

lesters:
@bratseth Please review. Wraps the tensorflow to onnx conversion script to add output aliases that are lost in the conversion process. Should fix the main problem that caused us to revert previous attempt at tf2onnx: https://github.com/vespa-engine/system-test/pull/786/files

Only enabled for models under a special directory (`tf_2_onnx` under `models`) for now for testing to avoid user service interruptions.


bratseth (approved):
👍

PR #14104, opened 2020-08-19
return Vespa connection when deploying Vespa app.

thigm85:
Deploying with `VespaDocker.deploy` now return a `Vespa` app connection in addition to the deployment message.

### DIFF
--- a/python/vespa/vespa/package.py
+++ b/python/vespa/vespa/package.py
@@ -464,5 +464,7 @@ class VespaDocker(object):
             raise RuntimeError(deployment_message)
 
         return Vespa(
-            url="http://localhost", port=8080, deployment_message=deployment_message
+            url="http://localhost",
+            port=self.local_port,
+            deployment_message=deployment_message,
         )
--- a/python/vespa/vespa/package.py
+++ b/python/vespa/vespa/package.py
@@ -1,4 +1,5 @@
 import os
+import re
 from time import sleep
 from typing import List, Mapping, Optional
 from pathlib import Path
@@ -440,8 +441,7 @@ class VespaDocker(object):
         :param disk_folder: Disk folder to save the required Vespa config files.
         :param container_memory: Docker container memory available to the application.
 
-        :return: A tuple where the first element is the deployment message and the second element is a Vespa connection
-            instance.
+        :return: a Vespa connection instance.
         """
 
         self.application_package.create_application_package_files(dir_path=disk_folder)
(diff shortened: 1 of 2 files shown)


python/vespa/vespa/package.py
```diff
@@ -446,4 +457,8 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
         deployment = self.container.exec_run(
             "bash -c '/opt/vespa/bin/vespa-deploy prepare /app/application && /opt/vespa/bin/vespa-deploy activate'"
         )
-        return deployment.output.decode("utf-8").split("\n")
+
+        return (
+            deployment.output.decode("utf-8").split("\n"),
+            Vespa(url="http://localhost", port=8080),
```
oyving, line 463: Is it idiomatic to return a tuple with message and a handle? It seems odd, and how do you programatically see the difference between successes and failures here?

You should probably use self.local_port instead of directly referencing port 8080?

oyving (changes requested):

python/vespa/vespa/package.py
thigm85: 1. Yes, you are right. The tuple seems odd. I suggest to store the deployment message as an attribute of the `Vespa` instance. What do you think?

2. About the difference between successes and failures: This info needs to be extracted from the deployment message. My plan is to improve this part of the code as I experience deployment failures as I do not have a full overview of the type of errors I can face at this moment.

3. I agree about `self.local_port`. I will make the change after we settle about the points above.

oyving: I guess the idiomatic way would be to throw and exception on failure and return the handle on success? It can lead to a bit noisy code, but I think that's the way the Python standard libraries does it (e.g. opening a file failures). Then an error message could be encoded in the exception. Messages on success can be an attribute on the Vespa object.

thigm85: Yes, I agree. I will definitely throw an exception if a deployment failure occur. I will implement the exceptions as soon as I face a deployment failure. But I think we are agreeing that the deployment message will be included on the `Vespa` instance when it is returned (which will only happen if deployment goes through.)

thigm85: Also, I found a way to detect when vespa deployment fails and then the code raises RunTimeError exception with the deployment message showing what went wrong.

oyving: Suddenly I wonder why port is a separate parameter to the constructor and not part of the URL. I'm slow to catch that :)

thigm85: Not sure why I chose that besides the fact that I guess url and port often comes as two different arguments in other libraries. I myself always wonder if port should be included or not when specifying urls. Do you suggest a change?

thigm85: But if we agree to unify url and port into one I can do so in another PR.

oyving: In general the URL should only omit the port number if it's the default port number for the protocol specified in the URL. But yes, let's keep that for a different PR then :)

oyving (approved):

PR #14107, opened 2020-08-19
Log and skip application entries that have wrong tenant

hmusum:


hakonhall (approved):

hakonhall:
Could these just be deleted from ZK's /config/... ?

hmusum:
Yes, I plan to do that in a later PR. And also verify before writing such entries

PR #14113, opened 2020-08-20
Add method to feed one data point

thigm85:

### DIFF
--- a/python/vespa/vespa/application.py
+++ b/python/vespa/vespa/application.py
@@ -3,6 +3,7 @@
 from typing import Optional, Dict, Tuple, List
 from requests import post
 from pandas import DataFrame
+from requests import post
 
 from vespa.query import Query, VespaResult
 from vespa.evaluation import EvalMetric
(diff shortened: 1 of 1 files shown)


python/vespa/vespa/application.py
```diff
@@ -76,6 +77,22 @@ def query(
...
+        """
+        Feed a data point to a Vespa app.
+
+        :param schema: The schema that we are sending data to.
+        :param data_id: Unique id associated with this data point.
+        :param fields: Dict containing all the fields required by the `schema`.
+        :return: Response of the HTTP POST request.
+        """
+        end_point = "{}/document/v1/{}/{}/docid/{}".format(
+            self.end_point, schema, schema, str(data_id)
```
oyving, line 90: Schema and cluster are two different things. I am not sure if it matters specifically here.

oyving (approved):

python/vespa/vespa/application.py
thigm85: They are different things but my (possibly wrong) understanding is that they must have the same name. At least I believe I read that in the docs and all our tutorials follow this pattern.

PR #14117, opened 2020-08-20
Revert "Revert "Balder/use an actual fieldset repo""

baldersheim:
Reverts vespa-engine/vespa#14106


baldersheim:
@havardpe The last commit is new.

document/src/vespa/document/fieldset/fieldsetrepo.cpp
```diff
@@ -124,7 +128,13 @@ FieldSetRepo::configureDocumentType(const DocumentType & documentType) {
     for (const auto & entry : documentType.getFieldSets()) {
         vespalib::string fieldSetName(documentType.getName());
         fieldSetName.append(':').append(entry.first);
-        _configuredFieldSets[fieldSetName] = parse(_doumentTyperepo, fieldSetName);
+        try {
+            auto fieldset = parse(_doumentTyperepo, fieldSetName);
+            _configuredFieldSets[fieldSetName] = std::move(fieldset);
+        } catch (const FieldNotFoundException & ex) {
```
havardpe, line 134: It is probably better to just ignore these silently (since they are legal when used in context of the proton index schema) and rather warn about usage later if needed.

PR #14120, opened 2020-08-20
Write tenant metadata

hmusum:
We want metadata to be able to find unused tenants, there is no way of
tracking this today without tracking all sessions for a tenant and we allow
sessions some lifetime, so this makes it impossible to find out in systems
where there are a lot of short-lived deployments (test systems).

Writing metadata is guarded by a feture flag.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -691,6 +691,8 @@ public class SessionRepository {
         return curator.lock(lockPath(sessionId), Duration.ofMinutes(1)); // These locks shouldn't be held for very long.
     }
 
+    public Clock clock() { return clock; }
+
     private Path lockPath(long sessionId) {
         return locksPath.append(String.valueOf(sessionId));
     }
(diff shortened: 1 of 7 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -691,6 +691,8 @@ public Lock lock(long sessionId) {
         return curator.lock(lockPath(sessionId), Duration.ofMinutes(1)); // These locks shouldn't be held for very long.
     }
 
+    public Clock clock() { return clock; }
```
jonmv, line 694: Not where I'd necessarily expect to find a `Clock`, but I guess there may be reasons it's here.

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantMetaData.java
```diff
@@ -0,0 +1,65 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to encode metadata", e);
+        }
+    }
+
+    public static TenantMetaData fromJsonString(String jsonString) {
+        try {
+            Slime data = new Slime();
+            new JsonDecoder().decode(data, Utf8.toBytes(jsonString));
+            Inspector root = data.get();
```
jonmv, line 49: `SlimeUtils` lets you easily convert JSON/bytes/Strings, for later :)

jonmv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantMetaData.java
hmusum: Thanks, forgot about `SlimeUtils`, I'll do a pass and convert some other code that should use SlimeUtils as well

PR #14127, opened 2020-08-21
add metadata

kkraune:
- I got Unrecognized runtime "vespa"; defaulting to "python3" when running on colab - maybe this helps, pasted a subset of the getting started notebook here
- no idea if this works though


thigm85:
I see the issues you reported here. I will fix the "Unrecognized runtime "vespa"; defaulting to "python3" when running on colab " in another PR.

The failure on the pipeline have been solved.

Issue #14122, opened 2020-08-21
Could not instantiate query from YQL due to special characters


yehzu:
Hi, we found that some special characters like ½ would break YQL. 
for example: `select * from item where title contains "juice ½ oz";`
and the response got from Vespa is
```
{
    "root": {
        "id": "toplevel",
        "relevance": 1.0,
        "fields": {
            "totalCount": 0
        },
        "errors": [
            {
                "code": 4,
                "summary": "Invalid query parameter",
                "message": "Could not instantiate query from YQL",
[...]

PR #14143, opened 2020-08-24
handle plugin tokenizer returning tokens with empty original string

arnej27959:

should fix:
https://github.com/vespa-engine/vespa/issues/14122

@bratseth please review


bratseth (approved):
👍

PR #14145, opened 2020-08-24
Event entries have the highest int value

jonmv:
@olaaun please review and merge.


olaaun (approved):
Travis fails on unrelated test

PR #14157, opened 2020-08-25
Add delay before running maintenance the first time 

hmusum:
Unit tests depend on files not being deleted immediately after
FilesArchived has been started

Issue introduced in https://github.com/vespa-engine/vespa/pull/14039

### DIFF
--- a/logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
+++ b/logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
@@ -45,6 +45,8 @@ public class FilesArchived {
 
     private void run() {
         try {
+            // Sleep some time before first maintenance, unit test depend on files not being removed immediately
+            Thread.sleep(1000);
             while (true) {
                 maintenance();
                 waitForTrigger(2000);


logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
```diff
@@ -45,6 +45,8 @@ private void waitForTrigger(long milliS) throws InterruptedException {
 
     private void run() {
         try {
+            // Sleep some time before first maintenance, unit test depend on files not being removed immediately
+            Thread.sleep(1000);
```
baldersheim, line 49: Sounds like imperfect unit tests.
Is the delay a little short ?

baldersheim (approved):

logserver/src/main/java/com/yahoo/logserver/handlers/archive/FilesArchived.java
hmusum: Running the test failed every time for me until I added the delay and worked every time afterwards. On a fast machine I guess it usually works.

PR #14159, opened 2020-08-25
Allow deployment of application package without deployOptions 

jonmv:
@bratseth please review.
@kkraune @thigm85 FYI.


bratseth (approved):
Nice

PR #14165, opened 2020-08-26
Deploy to Vespa Cloud from python, talk to endpoints with mTLS

jonmv:
@thigm85


jonmv:
I think I'm done here. Further python specific changes left to @thigm85, and review to be done by @oyving :)

python/vespa/vespa/application.py
```diff
@@ -87,7 +90,7 @@ def query(
         if debug_request:
             return VespaResult(vespa_result={}, request_body=body)
         else:
-            r = post(self.search_end_point, json=body)
+            r = post(self.search_end_point, json=body, cert=self.cert)
```
jonmv, line 93: Not sure if this works against the local docker instance.

python/vespa/vespa/application.py
```diff
@@ -103,7 +106,7 @@ def feed_data_point(self, schema: str, data_id: str, fields: Dict) -> Response:
             self.end_point, schema, schema, str(data_id)
         )
         vespa_format = {"fields": fields}
-        response = post(end_point, json=vespa_format)
+        response = post(end_point, json=vespa_format, cert=self.cert)
```
jonmv, line 109: Same here.

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,160 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+    def read_private_key(key_location: str) -> ec.EllipticCurvePrivateKey:
+        with open(key_location, 'rb') as key_data:
+            key = serialization.load_pem_private_key(key_data.read(), None, default_backend())
+            if not isinstance(key, ec.EllipticCurvePrivateKey):
+                raise TypeError("Key at " + key_location + " must be an elliptic curve private key")
+            return key
+
+    @staticmethod
+    def write_private_key_and_cert(key: ec.EllipticCurvePrivateKey, cert: x509.Certificate) -> (str, str):
+        cert_file = tempfile.NamedTemporaryFile('wt')
```
jonmv, line 522: This file can only be read by the user writing it, and is deleted immediately when closed. The `VespaCloud` closes this file on _its_ close, and is meant to be used like
```python
with VespaCloud('tenant', 'application', '/path/to/api_key') as cloud:
    ...
```

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,160 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+            return parsed
+
+    def get_dev_region(self) -> str:
+        return self.request('GET', '/zone/v1/environment/dev/default')['name']
+
+    def get_endpoint(self, instance: str, region: str) -> str:
+        endpoints = self.request('GET',
+                                 '/application/v4/tenant/{}/application/{}/instance/{}/environment/dev/region/{}' \
+                                 .format(self.tenant, self.application, instance, region))['endpoints']
+        container_url = [endpoint['url'] for endpoint in endpoints if endpoint['cluster'] == 'test_app_container']
```
jonmv, line 577: Hard-coded container name here. Remember this is the container name changes (in `services.xml`).

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,160 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+        region = self.get_dev_region()
+        job = 'dev-' + region
+        run = self.start_deployment(instance, job, application_package)
+        self.follow_deployment(instance, job, run)
+        endpoint_url = self.get_endpoint(instance, region)
+        return Vespa(url = endpoint_url, cert = self.data_cert_file.name)
+
+    def close(self):
+        self.connection.close()
+        self.data_cert_file.close()
```
jonmv, line 643: This deletes the data key/cert _forever_. Consider allowing the user to specify the certificate file, for persistent storage, if it's meaningful to allow the user to keep talking to its endpoint outside of the python code, or if the `VespaCloud` variable can't be kept in scope.

jonmv:
Ok, so the lifecycles of the `VespaCloud` class and the actual deployment are a bit weird together. The actual deployment may live on (unless you `delete(...)` it), but the mTLS creds are lost with the `VespaCloud` object. At the same time, it makes sense to not `delete`, because deployment time from scratch is 15 minutes, but only 15 seconds if you already have a running instance. So, as long as all python code starts with deploying (which probably is a good idea anyway, to keep the instance from disappearing), and then keeps the `VespaCloud` object in scope for as long as it runs, this isn't a problem.

python/vespa/vespa/application.py
```diff
@@ -15,13 +15,15 @@ def __init__(
...
         deployment_message: Optional[List[str]] = None,
+        cert: Optional[str] = None,
     ) -> None:
         """
         Establish a connection with a Vespa application.
 
         :param url: URL
         :param port: Port
         :param deployment_message: Message returned by Vespa engine after deployment.
+        :param cert: Certificate and key pair used for mTLS.
```
oyving, line 26: Please indicate the type of the string. Is it a filename, two PEM encoded objects, etc.

python/vespa/vespa/package.py
oyving: Is there a way to use an in-memory file object instead of a temporary on-disk file?

python/vespa/vespa/application.py
oyving, line 26: This should indicate the type in the `cert`. Is it a file name or the object itself?

oyving (changes requested):

python/vespa/vespa/application.py
oyving: This is a duplicate of the comment above 🤷

python/vespa/vespa/package.py
jonmv: Absolutely not, unfortunately. It's been filed as a "bug" for 8 years, but nothing has happened.

jonmv: There were even made PRs to address the issue, but the maintainers rejected them because they wanted a "perfect" solution, but this solution has, of course, not arrived, nor does it seem like it will.

python/vespa/vespa/application.py
jonmv: Fixed.

jonmv: Yeah 🤷‍♂️

python/vespa/vespa/package.py
oyving: How pythonic

thigm85: Another option is to use 
```
vespa_cloud = VespaCloud("vespa-team", "ms-marco", "/Users/tmartins/sample_application/tmartins.vespa-team.pem")
app = vespa_cloud.deploy('from-notebook', app_package)
```
Then continue  doing all the operation like `app.feed_data_point(...)` and then use `app.close()` when you are done.

Does that makes sense @jonmv?

python/vespa/vespa/package.py
thigm85: I believe that using
```
vespa_cloud = VespaCloud("vespa-team", "ms-marco", "/Users/tmartins/sample_application/tmartins.vespa-team.pem")
app = vespa_cloud.deploy('from-notebook', app_package)
# run as many commands as one needs
app.close()
```
solves the case where the `VespaCloud` variable can't be kept in scope.

thigm85: I think it is best to let them generate the certificate themselves if they want to talk to the end-point outside of the python code. As the intent of this library is to make experiments from a python environment.

python/vespa/vespa/package.py
jonmv: Do you mean if it makes sense to move generation of data place key/cert to when you deploy, so it lives only while  a particular deployment (`Vespa` object) is in scope? I think that makes sense, but there's not much practical difference. You can avoid keeping the `VespaCloud` object in scope, but that's it. 
And for the certificates, it's possible to just never `.close()` the object, of course. But you'd have a hard time finding the certificate file without printing its name somewhere ... :)

jonmv:
```python
with VespaCloud(...) as cloud:
    with cloud.deploy(...) as app:
        # both variables are in scope now, as you can't have the Vespa in scope without the VespaCloud anyway
        # and the VespaCloud is intended to be closed regardless of where the certs are, due to its HTTPConnection
```

jonmv: I guess you could perhaps do 
```python
with VespaCloud(...) as cloud:
    app = cloud.deploy(...)

with app:
    # do stuff
```

jonmv: Anyway, these are all details. Do whatever you think is better :)

python/vespa/vespa/package.py
```diff
@@ -504,122 +513,208 @@ def __init__(self, tenant: str, application: str, key_location: str) -> None:
...
+                elif status == "endpointCertificateTimeout":
+                    raise RuntimeError(
+                        "Endpoint certificate not ready in time; please contact Vespa team"
+                    )
+                elif status == "testFailure":
+                    raise RuntimeError(
+                        "Unexpected status; tests are not run for manual deployments"
+                    )
+                else:
+                    raise RuntimeError("Unexpected status '" + status + "'")
```
jonmv, line 710: Hehe, this is why I hate autoformatters :)

python/vespa/vespa/package.py
```diff
@@ -468,3 +499,264 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+        region = self.get_dev_region()
+        job = "dev-" + region
+        run = self.start_deployment(instance, job, application_package)
+        self.follow_deployment(instance, job, run)
+        endpoint_url = self.get_endpoint(
+            instance=instance,
+            region=region,
+            application_package_name=application_package.name,
+        )
+        return Vespa(url=endpoint_url, cert=self.data_cert_file.name)
```
thigm85, line 737: @jonmv It would be nice to include the `deployment_message` argument when creating the `Vespa` instance. What would best represent the deployment message in the Vespa Cloud case?

thigm85 (changes requested):

python/vespa/vespa/package.py
jonmv: I guess that depends. What is it used for? The whole log could be it. It's usually not that long for dev deployments. So, everything that's printed in `follow_deployment`.

python/vespa/vespa/application.py
jonmv: I initially believed it would be possible to send the actual cert and key, of course ... >_<

python/vespa/vespa/package.py
jonmv: Yes, you'd need to move the key/cert stuff to the `deploy(...)` call, but that's possible.

jonmv: You should prefer `with ...:`, though, to ensure resources are closed when exceptions are thrown.

python/vespa/vespa/package.py
thigm85: Ok, for now it is just for informational purposes.

jonmv: The deployment message from the config server is in there, if you do substring search, etc..

python/vespa/vespa/package.py
thigm85: No, I don't need to change anything. My code above is wrong and I meant `vespa_cloud.close()` instead of `app.close()`

I don't need to move anything to `.deploy()` as the `Vespa` instance knows the location of the cert.

python/vespa/vespa/package.py
thigm85: @oyving sold me on `black` a while ago, now it is hard to live without it.

jonmv: Yeah, pros and cons :)

python/vespa/vespa/application.py
thigm85: Yes, it works.

thigm85 (approved):

thigm85:
@oyving Can you make a final review before we merge?

oyving (changes requested):
There are some methods on the class that are very internal to the class itself and not part of the documented API. Would it make sense to prefix those with `_` (according to PEP-8) to make the intent of their internal-ness clear?

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,289 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+        with open(os.path.join(disk_folder, "application.zip"), "wb") as zipfile:
+            zipfile.write(application_zip_bytes.getvalue())
+
+        response = self.request(
+            "POST",
+            deploy_path,
+            application_zip_bytes,
+            {"Content-Type": "application/zip"},
+        )
+        print(response["message"])
```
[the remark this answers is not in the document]
jonmv: No, it contains a message intended for the user.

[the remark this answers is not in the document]
oyving: Wouldn't it be better to return that value from the method instead? Library code printing stuff is a bit iffy. I know there's a contention here between regular library code and interactive notebooks, but it would be nice to see if there was a better way.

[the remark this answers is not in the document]
jonmv: Yes, moving it up one level is a good idea.

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,289 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+                    + ":\n"
+                    + parsed["message"]
+                )
+            return parsed
+
+    def get_dev_region(self) -> str:
+        return self.request("GET", "/zone/v1/environment/dev/default")["name"]
+
+    def get_endpoint(
+        self, instance: str, region: str, application_package_name: str
```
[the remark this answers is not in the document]
thigm85: `application_package_name` is unnecessary here as the info is available through `self.application_package.name`. Made the changes.

python/vespa/vespa/package.py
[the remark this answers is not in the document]
thigm85: I don't think this was a left over. The idea is to print the deployment messages so that the user knows which stage he is. This `response["message"]` contains the following
```
Deployment started in run 6 of dev-aws-us-east-1c for vespa-team.ms-marco.from-notebook. This may take about 15 minutes the first time.
```
The entire message including the updates looks something like:
```
Deployment started in run 6 of dev-aws-us-east-1c for vespa-team.ms-marco.from-notebook. This may take about 15 minutes the first time.
INFO    [08:08:03]  Deploying platform version 7.276.8 and application version unknown ...
INFO    [08:08:04]  No services requiring restart.
INFO    [08:08:04]  Deployment successful.
INFO    [08:08:05]  Session 13618 for tenant 'vespa-team' prepared and activated.
INFO    [08:08:05]  ######## Details for all nodes ########
INFO    [08:08:05]  h711a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud: expected to be UP
INFO    [08:08:05]  --- platform docker.ouroath.com:4443/vespa/centos-tenant:7.276.8
INFO    [08:08:05]  --- container on port 4080 has config generation 13618, wanted is 13618
INFO    [08:08:05]  h712a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud: expected to be UP
INFO    [08:08:05]  --- platform docker.ouroath.com:4443/vespa/centos-tenant:7.276.8
INFO    [08:08:05]  --- logserver-container on port 4080 has config generation 13618, wanted is 13618
INFO    [08:08:05]  h713a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud: expected to be UP
INFO    [08:08:05]  --- platform docker.ouroath.com:4443/vespa/centos-tenant:7.276.8
INFO    [08:08:05]  --- container-clustercontroller on port 19050 has config generation 13618, wanted is 13618
INFO    [08:08:05]  --- storagenode on port 19102 has config generation 13618, wanted is 13618
INFO    [08:08:05]  --- searchnode on port 19107 has config generation 13618, wanted is 13618
INFO    [08:08:05]  --- distributor on port 19111 has config generation 13618, wanted is 13618
INFO    [08:08:05]  Found endpoints:
INFO    [08:08:05]  - dev.aws-us-east-1c
INFO    [08:08:05]   |-- https://msmarco-container.from-notebook.ms-marco.vespa-team.aws-us-east-1c.dev.public.vespa.oath.cloud/ (cluster 'msmarco_container')
INFO    [08:08:06]  Installation succeeded!
```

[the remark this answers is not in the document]
thigm85: @oyving Would you change it?

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,289 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+            )
+        )
+
+        application_zip_bytes = self.to_application_zip()
+
+        self.write_private_key_and_cert(
+            self.data_key, self.data_certificate, disk_folder
+        )
+        with open(os.path.join(disk_folder, "application.zip"), "wb") as zipfile:
+            zipfile.write(application_zip_bytes.getvalue())
```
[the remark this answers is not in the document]
thigm85: This is a good question. It is not necessary but I thought it would be nice to have a copy in case the user wants to check the config files that got deployed.

A future change would include the option to deploy a user generated .zip file to Cloud from pyvespa, instead of building the .zip file from our limited application package API. 

That way assuming there will always be a .zip file in the `disk_folder` would tie those two use cases together. But for now it is mostly for the user convenience of checking the config files.

python/vespa/vespa/package.py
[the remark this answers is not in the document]
oyving: I think it should be changed, but it might be over-engineering it. It's your call. But as a minimum I would have a `self.output` file-handle that you write these messages to (that default to `sys.stdout`). That way the output can at least be controlled by the user if they wish.

thigm85:
> There are some methods on the class that are very internal to the class itself and not part of the documented API. Would it make sense to prefix those with `_` (according to PEP-8) to make the intent of their internal-ness clear?

Yes, I agree. Made the changes for the classes involved in this PR.

python/vespa/vespa/package.py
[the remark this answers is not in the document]
oyving: I think that's a mixing of concerns. Also, wouldn't it be better for the user to just inspect the application package object and get the files printed to stdout?

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,289 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+        response = self.request(
+            "POST",
+            deploy_path,
+            application_zip_bytes,
+            {"Content-Type": "application/zip"},
+        )
+        print(response["message"])
+        return response["run"]
+
+    def follow_deployment(self, instance: str, job: str, run: int):
```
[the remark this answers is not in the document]
thigm85: Made the change. Check if it was that what you had in mind.

python/vespa/vespa/package.py
```diff
@@ -468,3 +490,289 @@ def deploy(self, disk_folder: str, container_memory: str = "4G"):
...
+                elif status == "endpointCertificateTimeout":
+                    raise RuntimeError(
+                        "Endpoint certificate not ready in time; please contact Vespa team"
+                    )
+                elif status == "testFailure":
+                    raise RuntimeError(
+                        "Unexpected status; tests are not run for manual deployments"
+                    )
+                else:
+                    raise RuntimeError("Unexpected status '" + status + "'")
```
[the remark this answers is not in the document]
thigm85: Done.

python/vespa/vespa/package.py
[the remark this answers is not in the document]
thigm85: I removed the step that write zip file to disk. I can revisit that if needed.

python/vespa/vespa/package.py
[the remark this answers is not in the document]
thigm85: Done.

oyving (approved):

PR #14167, opened 2020-08-26
Treat replica minority test-and-set failure as success

vekterli:
@geirst please review
@baldersheim FYI

Since Put/Remove operations do not have a write-repair pipeline
when replicas are out of sync/missing, any such TaS mutation hitting
a stale replica will fail. This will normally fail the entire operation
back to the client with a TaS failure even though the mutation will
eventually become visible in the cluster. This is particularly noticeable
for documents that are globally distributed across all nodes.

This commit introduces a pragmatic approach to this issue where TaS
failures will be effectively ignored iff a _majority_ of replicas have
ACKed the mutation successfully. This is a tradeoff where we don't want
to "lie" to the client about the level of redundancy of the ACKed operation,
but also don't want to introduce too many client resends due to spurious
failures. Requiring a majority is a good middle ground (and is what we
would like to move towards in the future with a majority consensus system
anyway).

### DIFF
--- a/storage/src/tests/distributor/putoperationtest.cpp
+++ b/storage/src/tests/distributor/putoperationtest.cpp
@@ -72,11 +72,11 @@ public:
     }
 
     void sendPut(std::shared_ptr<api::PutCommand> msg) {
-        op.reset(new PutOperation(getExternalOperationHandler(),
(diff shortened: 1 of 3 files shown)


geirst (approved):
👍

PR #14174, opened 2020-08-27
Do not throw if a session has already been removed

hmusum:
A session that has already been removed from cache is OK


jonmv (approved):
Ok ... :)

PR #14177, opened 2020-08-27
- If document type has been resolved, use if for test-and-set too.

baldersheim:
- Inline small frequently called methods.

@vekterli PR


document/src/vespa/document/datatype/structureddatatype.cpp
```diff
@@ -48,9 +48,11 @@ int32_t StructuredDataType::createId(vespalib::stringref name)
     // ASCII characters. Probably screwed up otherwise, but generated ids
     // should only be used in testing anyways. In production this will be
     // set from the document manager config.
-    vespalib::asciistream ost;
-    ost << name << ".0";  // Hardcode version 0 (version is not supported).
-    return crappyJavaStringHash(ost.str());
+    char *bufOnStack = static_cast<char *>(alloca(name.size() + 2));
```
vekterli, line 51: I always get a bit of a cold sweat when I see `alloca`, in particular since our threads are usually created from pools with stack size configurations that aren't all that big, and I'm not sure if type names are bounded in size. Could the `alloca` be guarded behind an `if` on the name size where it will only be used if the size is below a fixed threshold? Say, 256 bytes or so? Could even just use a fixed buffer for this case, which as a bonus is more portable. Fallback to `asciistream` in the extremely unlikely case where the name is longer.

vekterli (approved):
:+1: one comment for consideration re: stack allocation

document/src/vespa/document/datatype/structureddatatype.cpp
baldersheim: Agreed.

PR #14180, opened 2020-08-27
Arnej/count used memory

arnej27959:

@geirst @havardpe please review

### DIFF
--- a/eval/src/vespa/eval/eval/simple_tensor.cpp
+++ b/eval/src/vespa/eval/eval/simple_tensor.cpp
@@ -769,5 +769,14 @@ SimpleTensor::decode(nbostream &input)
     return builder.build();
 }
 
+size_t
+SimpleTensor::count_memory_used() const {
+    size_t result = sizeof(SimpleTensor);
+    size_t addr_size = sizeof(Address) * _type.dimensions().size();
+    size_t cell_size = sizeof(Cell) + addr_size;
+    result += _cells.size() * cell_size;
+    return result;
+}
+
 } // namespace vespalib::eval
 } // namespace vespalib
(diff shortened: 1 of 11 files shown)


eval/src/vespa/eval/tensor/wrapped_simple_tensor.cpp
```diff
@@ -54,6 +54,17 @@ WrappedSimpleTensor::accept(TensorVisitor &visitor) const
     }
 }
 
+size_t
+WrappedSimpleTensor::count_memory_used() const
+{
+    size_t result = sizeof(WrappedSimpleTensor);
+    if (_space) {
+        result += sizeof(SimpleTensor);
+        result += _tensor.cells().size() * sizeof(SimpleTensor::Cell);
```
havardpe, line 63: this will greatly underestimate the actual memory used since most of the address will not be included. probably not a big deal.

havardpe (approved):

eval/src/vespa/eval/tensor/wrapped_simple_tensor.cpp
havardpe: on second thought; the size of the Label struct multiplied by the number of dimensions in the tensor would be a good estimation for space used for the address part.

geirst (approved):
👍

eval/src/vespa/eval/eval/simple_tensor.cpp
```diff
@@ -769,5 +769,14 @@ SimpleTensor::decode(nbostream &input)
     return builder.build();
 }
 
+size_t
+SimpleTensor::count_memory_used() const {
+    size_t result = sizeof(SimpleTensor);
+    size_t addr_size = sizeof(Address) * _type.dimensions().size();
```
havardpe, line 775: Address -> Label

PR #14181, opened 2020-08-27
Group commits to TLS and sync (resurrected).

baldersheim:
@toregge PR
This is the resurrection of a very old branch that we tried to get in a few years ago, but it failed as we did not have enough parallelism in the backend. Now that bottleneck has been removed when persistence layer became async.

I suggest you have a brief look at it and we can discuss how to commence.


baldersheim:
This will be broken down into smaller parts.

baldersheim:
With #14304 this should be fully merged.

PR #14187, opened 2020-08-27
infer unknown onnx dimension sizes

havardpe:
also pre-allocate output onnx tensors and generally try to resolve as
much as possible up-front to reduce per-eval overhead.

@arnej27959 please review
@lesters FYI


arnej27959 (approved):
looks good to me!

PR #14188, opened 2020-08-27
Revert "Revert "Only wait for th elids that you are interested in.""

baldersheim:
Reverts vespa-engine/vespa#14186


baldersheim:
@toregge or @geirst PR
When adding commitAndWait in the reconfigureIndexSearchable it is run in the master thread which will never complete as the sign out task is posted on the same thread. But signing out the committed lids can happen in any task, so I just moved it to the ForceDoneContext destructor instead.

toregge (approved):

PR #14190, opened 2020-08-28
Move code from RemoteSession to SessionRepository

hmusum:
Simplify and move in the direction of unifying remote and local sessions
Move tests and avoid a lot of low-level setup code


configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
```diff
@@ -26,95 +15,46 @@
...
 
-    public synchronized ApplicationSet ensureApplicationLoaded() {
-        return applicationSet == null ? applicationSet = loadApplication() : applicationSet;
+    Optional<ApplicationSet> applicationSet() {
+        return applicationSet;
     }
 
-    public synchronized void deactivate() {
-        applicationSet = null;
+    public synchronized RemoteSession deactivate() {
```
jonmv, line 52: Would like a name change here, to `deactivated()`

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -275,13 +280,25 @@ public int deleteExpiredRemoteSessions(Clock clock, Duration expiryTime) {
...
+                deleteSession(session);
                 deleted++;
             }
         }
         return deleted;
     }
 
+
+    public void deactivate(RemoteSession remoteSession) {
+        remoteSessionCache.addSession(remoteSession.deactivate());
```
jonmv, line 292: `addSession` should perhaps also be named `putSession`?

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -358,27 +375,19 @@ void activate(RemoteSession session) {
...
     }
 
-    public void deactivate(RemoteSession remoteSession) {
-        remoteSession.deactivate();
-    }
-
-    public void delete(RemoteSession remoteSession, Optional<LocalSession> localSession) {
+    public void deleteSession(RemoteSession remoteSession, Optional<LocalSession> localSession) {
         localSession.ifPresent(this::deleteLocalSession);
         remoteSession.deactivate();
```
jonmv, line 388: Shouldn't this also be `deactivate(remoteSession)`? This does nothing now.

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -394,13 +403,84 @@ private void loadSessionIfActive(RemoteSession session) {
...
+        RemoteSession newSession = new RemoteSession(session.getTenantName(),
+                                    session.getSessionId(),
+                                    session.getSessionZooKeeperClient(),
+                                    Optional.of(newApplicationSet));
+        remoteSessionCache.addSession(newSession);
+        return newApplicationSet;
+    }
+
+    void confirmUpload(RemoteSession session) {
+        Curator.CompletionWaiter waiter = session.getSessionZooKeeperClient().getUploadWaiter();
```
jonmv, line 436: Why `session.getSessionZooKeeperClient()` here, and `createSessionZooKeeperClient(session.getSessionId())` elsewhere?

jonmv (changes requested):
Yes, this will be much better :)

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Yes, both are used, I want to clean up that, will do in a later PR

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Agreed, will fix

configserver/src/main/java/com/yahoo/vespa/config/server/session/RemoteSession.java
hmusum: 👍

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Yes, thanks

jonmv (approved):

PR #14199, opened 2020-08-30
Use llvm 11 on Fedora 33.

toregge:
@baldersheim : please review

### DIFF
--- a/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
+++ b/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
@@ -200,7 +200,12 @@ struct FunctionBuilder : public NodeVisitor, public NodeTraverser {
         }
         assert(pass_params == PassParams::LAZY);
         assert(params.size() == 2);
+#if LLVM_VERSION_MAJOR >= 11
+        return builder.CreateCall(llvm::cast<llvm::FunctionType>(params[0]->getType()->getPointerElementType()),
+                                  params[0], {params[1], builder.getInt64(idx)}, "resolve_param");
+#else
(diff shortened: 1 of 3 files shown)


eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
```diff
@@ -200,7 +200,12 @@ struct FunctionBuilder : public NodeVisitor, public NodeTraverser {
         }
         assert(pass_params == PassParams::LAZY);
         assert(params.size() == 2);
+#if LLVM_VERSION_MAJOR >= 11
+        return builder.CreateCall(llvm::cast<llvm::FunctionType>(params[0]->getType()->getPointerElementType()),
+                                  params[0], {params[1], builder.getInt64(idx)}, "resolve_param");
+#else
```
baldersheim, line 206: When was this signature available ? In 11 or in an earlier version ?

baldersheim (approved):

eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
toregge: 7 or earlier. The other signature was deprecated in llvm 8 and removed in llvm 11.

baldersheim: Then I suggest you change to test for the version where the new signature was available. What is the oldest llvm version we require ?

toregge: 7

baldersheim: But then we can just avoid the ifdef and use the new signature directly.

PR #14203, opened 2020-08-31
Run maintainers every 30 seconds

hmusum:


hakonhall (approved):

hakonhall:
What happens if the previous run takes more than 30s?

hmusum:
Maintainer uses `ScheduledExecutorService.scheduleAtFixedRate` which in its doc says `If any execution of this task takes longer than its period, then subsequent executions may start late, but will not concurrently execute`, so that should be no problem

hakonhall:
👍

PR #14207, opened 2020-08-31
Reduce time to keep unused file references from 4 to 2 hours

hmusum:
Remove unused config field

### DIFF
--- a/configdefinitions/src/vespa/configserver.def
+++ b/configdefinitions/src/vespa/configserver.def
@@ -51,9 +51,7 @@ ztsUrl string default=""
 
 # Maintainers
 maintainerIntervalMinutes int default=30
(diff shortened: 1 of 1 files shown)


configdefinitions/src/vespa/configserver.def
```diff
@@ -51,9 +51,7 @@ ztsUrl string default=""
 
 # Maintainers
 maintainerIntervalMinutes int default=30
-# TODO: Default set to a high value (1 year) => maintainer will not run, change when maintainer verified out in prod
-tenantsMaintainerIntervalMinutes int default=525600
```
jonmv, line 55: Unused?

jonmv (approved):

configdefinitions/src/vespa/configserver.def
hmusum: Yes, not in use anymore, it can be removed since it is only used by config server

PR #14208, opened 2020-08-31
Bratseth/drain

bratseth:
@baldersheim I think you might know this code best now?


baldersheim (approved):
I think I got the intention, and the implementations seemed sane.
But the size of the PR made created stack overflow a few times so I can not guarantee the correctness.

PR #14209, opened 2020-08-31
Use an allocator that automagically allocates large buffer by using m…

baldersheim:
…map.

@vekterli or @toregge PR

While tracing allocation above 16G I found some huge ones that can do better by using mmap on the large ones.


vekterli (approved):
:+1:

PR #14213, opened 2020-08-31
Jonmv/log handler with blocking flush

jonmv:

### DIFF
--- a/container-core/abi-spec.json
+++ b/container-core/abi-spec.json
@@ -114,7 +114,8 @@
     ],
     "methods": [
       "public void <init>(java.util.concurrent.Executor, com.yahoo.container.core.LogHandlerConfig)",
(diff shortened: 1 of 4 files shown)


jonmv:
I decided `flush()` was an appropriate place to synchronise here.

jonmv:
Or @bratseth

bratseth (approved):

baldersheim:
Is the LogHandler involved with logging in a container application ?
It sounds like it will the request threads block. If so this sounds like a very bad idea.

jonmv:
The threads that will block are those serving logs from the logserver archive. You're afraid these are a shared resource, used for things that will now block, and cause havoc elsewhere? 

This handler is hardly ever used, so I don't think we'll notice anything now, but to be future proof, we should probably consider if this could be a problem.

jonmv:
Conclusion: it's fine.

PR #14214, opened 2020-08-31
Explicitly enable freelists for content node B-tree DB data store

vekterli:

Freelists were unintentionally not enabled due to the difference
in how `ArrayStore` and `DataStore` freelist setup is done. The
existing distributor `ArrayStore` traits set this up implicitly via
the constructor config, while the content node `DataStore` needs to
have freelists explicitly enabled.


toregge (approved):

geirst (approved):
👍

PR #14218, opened 2020-09-01
Include test report in job run details

tokle:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -190,6 +190,26 @@ public void updateTestLog(RunId id) {
         });
     }
 
+    public void updateTestReport(RunId id) {
+        locked(id, run -> {
+            if(run.stepStatus(endTests)
+                    .stream().peek(status -> System.out.println("endTests status: " + status.name()))
```
jonmv, line 196:
```suggestion
                    .stream()
```

tokle: removed

jonmv (approved):

jonmv:
Very good!

PR #14231, opened 2020-09-01
Add and use API for storing deployment meeta data

jonmv:


freva (approved):
Consider also adding other XML files referenced by `services.xml`/`deployment.xml` to the meta ZIP

jonmv:
Urgh, yes.

PR #14232, opened 2020-09-01
Decode once

bratseth:


arnej27959 (approved):
this looks OK but is it really any faster to do caching of doubles?

bratseth:
I didn't really measure. Slime uses 2 hash lookups, but so do the replacement since I want to be able to store nulls. Let me check it tomorrow.

bratseth:
Yes, a FeatureData became about 3-5 times faster for doubles with this.
I made it completely on par with Map here: https://github.com/vespa-engine/vespa/pull/14258

PR #14238, opened 2020-09-02
Ignore response filter chain bindings when constructing access contro…

bjorncs:
…l filter chains

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/http/AccessControl.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/http/AccessControl.java
@@ -144,8 +144,7 @@ public class AccessControl {
         for (FilterBinding binding : http.getBindings()) {
             if (binding.chainId().toId().equals(chainId)) {
                 for (FilterBinding otherBinding : http.getBindings()) {
-                    if (!binding.chainId().equals(otherBinding.chainId())
-                            && effectivelyDuplicateOf(binding.binding(), otherBinding.binding())) {
+                    if (effectivelyDuplicateOf(binding, otherBinding)) {
                         duplicateBindings.add(binding);
                     }
                 }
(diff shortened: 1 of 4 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/http/AccessControl.java
```diff
@@ -154,14 +153,17 @@ private void removeDuplicateBindingsFromChain(Http http, ComponentId chainId) {
         duplicateBindings.forEach(http.getBindings()::remove);
     }
 
-    private static boolean effectivelyDuplicateOf(BindingPattern accessControlBinding, BindingPattern other) {
-        return accessControlBinding.equals(other)
-                || (accessControlBinding.path().equals(other.path()) && other.matchesAnyPort());
+    private static boolean effectivelyDuplicateOf(FilterBinding accessControlBinding, FilterBinding other) {
+        if (accessControlBinding.chainId().equals(other.chainId())) return false; // Same filter chain
+        if (other.type() == FilterBinding.Type.RESPONSE) return false;
```
jonmv, line 158: How about comparing the types, instead of assuming this is a request type?

config-model/src/main/java/com/yahoo/vespa/model/container/http/FilterBinding.java
```diff
@@ -11,16 +11,20 @@
...
-    private FilterBinding(ComponentSpecification chainId, BindingPattern binding) {
+    private FilterBinding(Type type, ComponentSpecification chainId, BindingPattern binding) {
+        this.type = type;
         this.chainId = chainId;
         this.binding = binding;
     }
 
-    public static FilterBinding create(ComponentSpecification chainId, BindingPattern binding) {
-        return new FilterBinding(chainId, binding);
+    public static FilterBinding create(Type type, ComponentSpecification chainId, BindingPattern binding) {
```
jonmv, line 26: I assume this code isn't called by older versions of code, in the config server ... ?

config-model/src/main/java/com/yahoo/vespa/model/container/http/xml/HttpBuilder.java
```diff
@@ -121,7 +121,8 @@ private static AthenzDomain getAccessControlDomain(DeployState deployState, Elem
 
                 for (Element bindingSpec: XML.getChildren(child, "binding")) {
                     String binding = XML.getValue(bindingSpec);
-                    result.add(FilterBinding.create(chainId, UserBindingPattern.fromPattern(binding)));
+                    FilterBinding.Type type = tagName.equals("request-chain") ? FilterBinding.Type.REQUEST : FilterBinding.Type.RESPONSE;
```
jonmv, line 124: Would prefer explicit check for `"response-chain"` and throw otherwise, just to be absolutely future-proff.

config-model/src/main/java/com/yahoo/vespa/model/container/http/AccessControl.java
bjorncs: The first parameter is assumed to be an access control binding, meaning it cannot be a response filter binding.

jonmv (approved):
Some details, but logic LGTM.

config-model/src/main/java/com/yahoo/vespa/model/container/http/FilterBinding.java
bjorncs: Nope, this method is only used inside `config-model` module.

config-model/src/main/java/com/yahoo/vespa/model/container/http/xml/HttpBuilder.java
bjorncs: I have improved the logic in new commit.

config-model/src/main/java/com/yahoo/vespa/model/container/http/AccessControl.java
jonmv: Well, it's a private in the `AccessControl` class, so I guess that's OK.

jonmv (approved):
:)

PR #14239, opened 2020-09-02
Memory usage tracking in direct tensor store

geirst:
@toregge please review
@arnej27959 FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/direct_tensor_store.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/direct_tensor_store.cpp
@@ -57,7 +57,7 @@ EntryRef
 DirectTensorStore::store_tensor(std::unique_ptr<Tensor> tensor)
 {
     assert(tensor);
-    return add_entry(TensorSP(tensor.release()));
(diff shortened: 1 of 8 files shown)


toregge (approved):

searchlib/src/vespa/searchlib/tensor/direct_tensor_store.cpp
```diff
@@ -31,8 +57,7 @@ EntryRef
 DirectTensorStore::store_tensor(std::unique_ptr<Tensor> tensor)
 {
     assert(tensor);
-    // TODO: Account for heap allocated memory
-    return _concrete_store.addEntry(TensorSP(tensor.release()));
+    return add_entry(TensorSP(tensor.release()));
```
arnej27959, line 60: I would prefer making the shared_ptr as TensorSP(std::move(tensor)) instead of using release().

arnej27959 (approved):
👍

searchlib/src/vespa/searchlib/tensor/direct_tensor_store.cpp
geirst: Fixed

PR #14241, opened 2020-09-02
- Add test for sequencing.

baldersheim:
- Also sequence multiple operations to the same id as the originating request.

@bratseth PR

### DIFF
--- a/docproc/src/main/java/com/yahoo/docproc/jdisc/messagebus/MbusRequestContext.java
+++ b/docproc/src/main/java/com/yahoo/docproc/jdisc/messagebus/MbusRequestContext.java
@@ -51,6 +51,7 @@ public class MbusRequestContext implements RequestContext, ResponseHandler {
     // document being processed is a resource and is then grabbing more resources of
     // the same type without releasing its own resources.
     public final static String internalNoThrottledSource = "internalNoThrottledSource";
+    private final static String internalNoThrottledSourcePath = "/" + internalNoThrottledSource;
 
     public MbusRequestContext(MbusRequest request, ResponseHandler responseHandler,
                               ComponentRegistry<DocprocService> docprocServiceComponentRegistry,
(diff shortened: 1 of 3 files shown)


baldersheim:
Test failure is unrelated.

docproc/src/main/java/com/yahoo/docproc/jdisc/messagebus/MbusRequestContext.java
```diff
@@ -96,12 +97,16 @@ public void processingDone(List<Processing> processings) {
         }
         long inputSequenceId = requestMsg.getSequenceId();
         ResponseMerger responseHandler = new ResponseMerger(requestMsg, messages.size(), this);
+        int numMsgWithOriginalSequenceId = 0;
         for (Message message : messages) {
+            if (message.getSequenceId() == inputSequenceId) numMsgWithOriginalSequenceId++;
+        }
+        for (Message message : messages) {
+            String path = internalNoThrottledSourcePath;
+            if ((numMsgWithOriginalSequenceId == 1) && (message.getSequenceId() == inputSequenceId))
```
bratseth, line 106: Shouldn't this be numMsgWithOriginalSequenceId >= 1 ?

bratseth (commented):
otherwise LGTM!

docproc/src/main/java/com/yahoo/docproc/jdisc/messagebus/MbusRequestContext.java
baldersheim: No, you shall only continue on the original path if there is exactly one operation with the same id as the incoming operation. If there are multiple operations with the same id they must always be sequenced. If they are to the original id or not does not matter.

baldersheim: The if is actually the normal path :)

bratseth: 👍

PR #14257, opened 2020-09-02
Allow setting a request type explicitly

bratseth:
This lets handler authors control the requestType explicitly
by setting it on the HttpResponse, which is useful to avoid
misclassification of POST requests to reading handlers as writes.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/jdisc/HttpResponse.java
+++ b/container-core/src/main/java/com/yahoo/container/jdisc/HttpResponse.java
@@ -3,6 +3,7 @@ package com.yahoo.container.jdisc;
 
 import com.yahoo.container.logging.AccessLogEntry;
 import com.yahoo.jdisc.HeaderFields;
+import com.yahoo.jdisc.Request;
 import com.yahoo.jdisc.Response;
 import com.yahoo.processing.execution.Execution.Trace.LogValue;
 
@@ -18,21 +19,18 @@ import java.util.Collections;
  */
 public abstract class HttpResponse {
 
-    /**
-     * Default response content type; text/plain.
-     */
+    /** Default response content type; text/plain. */
     public static final String DEFAULT_MIME_TYPE = "text/plain";
 
-    /**
-     * Default encoding/character set of a HTTP response; UTF-8.
-     */
+    /** Default encoding/character set of a HTTP response; UTF-8. */
     public static final String DEFAULT_CHARACTER_ENCODING = "UTF-8";
-
 
     private final Response parentResponse;
 
+    private Request.RequestType requestType;
+
     /**
-     * Create a new HTTP response.
+     * Creates a new HTTP response
      *
      * @param status the HTTP status code to return with this response (may be changed later)
      * @see Response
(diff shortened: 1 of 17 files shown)


container-core/src/main/java/com/yahoo/container/jdisc/HttpResponse.java
```diff
@@ -129,4 +125,13 @@ public void complete() { }
         return Collections::emptyIterator;
     }
 
+    /** Sets the type classification of this request for metric collection purposes */
+    public void setRequestType(Request.RequestType requestType) { this.requestType = requestType; }
```
bjorncs, line 129: Why is request type as member of `HttpResponse`? I would assume `HttpRequest` is more suitable as the "type" is a property of the request, not the response. It's also stored in the request type for the jdisc core abstraction layer.

jdisc_core/src/main/java/com/yahoo/jdisc/Request.java
```diff
@@ -48,6 +48,11 @@
     private boolean serverRequest;
     private Long timeout;
     private URI uri;
+    private RequestType requestType;
+
+    public enum RequestType {
```
bjorncs, line 53: `RequestType` is a too generic term in my opinion. Consider naming the concept something more specific.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyHttpServer.java
```diff
@@ -208,32 +203,26 @@ private static void initializeJettyLogging() {
...
 
-    private HandlerCollection getHandlerCollection(
-            ServerConfig serverConfig,
-            ServletPathsConfig servletPathsConfig,
-            List<JDiscServerConnector> connectors,
-            ServletHolder jdiscServlet,
-            ComponentRegistry<ServletHolder> servletHolders,
-            FilterHolder jDiscFilterInvokerFilter) {
-
+    private HandlerCollection getHandlerCollection(ServerConfig serverConfig,
```
bjorncs, line 220: I actually prefer the previous indentation style. The new style requires a re-indentation of all parameters if the name or return type of the method is changed.

bjorncs (commented):
See comments. Consider adding the new attribute to the access log as well.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyHttpServer.java
bratseth: But the IDE does that for you automatically ...

bjorncs: Not mine, probably missing some non-default configuration.

jdisc_core/src/main/java/com/yahoo/jdisc/Request.java
bratseth: That's what it's already called in the metric so using a different name here increases entropy a bit. Did you have a suggestion in mind?

bjorncs: Not really, good naming ain't easy :)

container-core/src/main/java/com/yahoo/container/jdisc/HttpResponse.java
bratseth: That allows the handler/processors/searchers to process the request and decide the right classification based on its content.

bjorncs: Jdisc filters cannot perform this classification easily if it's not part of the request type.

bratseth: Well there's lots of things filters cannot perform easily - they are special purpose. Is there a good use case for filters doing this classification?

bjorncs: Access control

bjorncs (approved):

PR #14258, opened 2020-09-02
Optimize for looking up values that are present

bratseth:
At the cost of always attempting to decode a nonexisting value
we can skip the containsKey lookup, making double lookups in FeatureData
about 50% faster and on par with HashMap.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/result/FeatureData.java
+++ b/container-search/src/main/java/com/yahoo/search/result/FeatureData.java
@@ -74,10 +74,14 @@ public class FeatureData implements Inspectable, JsonProducer {
      *                                  (that is, if it is a tensor with nonzero rank)
      */
     public Double getDouble(String featureName) {
-        if (decodedDoubles != null && decodedDoubles.containsKey(featureName))
-            return decodedDoubles.get(featureName);
-        Double value = decodeDouble(featureName);
-        if (decodedDoubles == null)
+        Double value = null;
+        if (decodedDoubles != null)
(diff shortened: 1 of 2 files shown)


container-search/src/main/java/com/yahoo/search/result/FeatureData.java
```diff
@@ -74,10 +74,14 @@ public StringBuilder writeJson(StringBuilder target) {
...
-        Double value = decodeDouble(featureName);
-        if (decodedDoubles == null)
+        Double value = null;
+        if (decodedDoubles != null)
+            value = decodedDoubles.get(featureName);
+        if (value != null)
+            return value;
+
+        value = decodeDouble(featureName);
+        if (value != null && decodedDoubles == null)
```
arnej27959, line 84: this must still be just:
```
    if (decodedDoubles == null)
            decodedDoubles = new HashMap<>();	
```
same with decodedTensors below.
but since in all common cases you will always create it, I would suggest doing that first in the method.  That should make the code easier to read.

arnej27959 (changes requested):

container-search/src/main/java/com/yahoo/search/result/FeatureData.java
bratseth: Good suggestion, I'll do that.

bratseth:
PTAL @arnej27959

arnej27959 (approved):
👍

PR #14269, opened 2020-09-03
Use the tensor type to switch tensor binary format

bratseth:
The binary format of a tensor should depend on the tensor type,
not the implementation type as the API permits the user choosing that
(and it may not be 1-1 anyway).

This makes this change for sparse tensors using the mixed implementation
type but not dense tensors using the mixed implementation type as that
would be more work given the unfinished state of the mixed implementation.


lesters (approved):
LGTM!

PR #14290, opened 2020-09-04
Arnej/optimize copy cells

arnej27959:

@baldersheim as the hash_map / optimization expert please review
@geirst  and @toregge - feel free to veto the second commit if you think it's too "clever"

### DIFF
--- a/eval/src/vespa/eval/tensor/sparse/sparse_tensor.cpp
+++ b/eval/src/vespa/eval/tensor/sparse/sparse_tensor.cpp
@@ -29,11 +29,11 @@ using Cells = SparseTensor::Cells;
 void
 copyCells(Cells &cells, const Cells &cells_in, Stash &stash)
 {
-    cells.resize(cells_in.capacity());
-    for (const auto &cell : cells_in) {
+    cells = cells_in;
+    for (auto &cell : cells) {
         SparseTensorAddressRef oldRef = cell.first;
         SparseTensorAddressRef newRef(oldRef, stash);
-        cells[newRef] = cell.second;
+        cell.first = newRef;
     }
 }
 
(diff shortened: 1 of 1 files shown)


toregge (approved):

eval/src/vespa/eval/tensor/sparse/sparse_tensor.cpp
```diff
@@ -29,10 +29,11 @@ using Cells = SparseTensor::Cells;
 void
 copyCells(Cells &cells, const Cells &cells_in, Stash &stash)
 {
-    for (const auto &cell : cells_in) {
+    cells = cells_in;
+    for (auto &cell : cells) {
         SparseTensorAddressRef oldRef = cell.first;
         SparseTensorAddressRef newRef(oldRef, stash);
-        cells[newRef] = cell.second;
+        cell.first = newRef;
```
baldersheim, line 36: 1 - A little comment about this not changing the hash value would be good.
2 - I guess you can can also write it as 
cell.first = SparseTensorAddressRef(cell.first, stash);
You should see what effect it has on speed. The compiler might find it easier or harder. 
noexcept on the SparseTensorAddressRef constructors might also make it easier on the optimizer.

baldersheim (changes requested):
At least a comment about dirty trick would be good.

eval/src/vespa/eval/tensor/sparse/sparse_tensor.cpp
arnej27959: added some comments.
doing 2) or adding noexcept didn't have any impact on the produced code, so I left it as-is.

arnej27959:
@baldersheim Please Take Another Look.

baldersheim (approved):

baldersheim (approved):

PR #14294, opened 2020-09-04
Jonmv/feed client error code handling

jonmv:
@bratseth please review.

### DIFF
--- a/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
+++ b/vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
@@ -347,11 +347,16 @@ public class IOThread implements Runnable, AutoCloseable {
                     currentConnection.handshake();
                     successfulHandshakes.getAndIncrement();
                 } catch (ServerResponseException ser) {
+                    int code = ser.getResponseCode();
+                    if (code == 401 || code == 403) {
+                        drainDocumentQueueWhenFailingPermanently(new Exception("Denied access by endpoint:" + ser.getResponseString()));
+                        log.log(Level.SEVERE, "Failed authentication or authorization with '" + endpoint + "': " + Exceptions.toMessageString(ser));
+                        return ConnectionState.CONNECTED; // Should ideally exit immediately, instead of doing this per X documents :/
+                    }
 
                     executeProblemsCounter.incrementAndGet();
                     log.log(Level.INFO, "Failed talking to endpoint. Handshake with server endpoint '" + endpoint +
-                                        "' failed. Will re-try handshake.",
-                            ser);
+                                        "' failed -- will re-try handshake: " + Exceptions.toMessageString(ser));
 
                     drainFirstDocumentsInQueueIfOld();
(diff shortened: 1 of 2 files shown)


vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
```diff
@@ -348,11 +347,16 @@ private ConnectionState cycle(ConnectionState connectionState) {
                     currentConnection.handshake();
                     successfulHandshakes.getAndIncrement();
                 } catch (ServerResponseException ser) {
+                    int code = ser.getResponseCode();
+                    if (code == 401 || code == 403) {
+                        drainDocumentQueueWhenFailingPermanently(new Exception("Denied access by endpoint:" + ser.getResponseString()));
+                        log.log(Level.SEVERE, "Failed authentication or authorization with '" + endpoint + "': " + Exceptions.toMessageString(ser));
+                        return ConnectionState.CONNECTED; // Should ideally exit immediately, instead of doing this per X documents :/
```
bratseth, line 354: If we want to we could exit, but I'm not sure it's always the best thing to do since there could be a fix to the problem rolling out simultaneously, we could be feeding to multiple clusters etc. I think in general this client wants to keep trying until timeout, for better or worse.

vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
```diff
@@ -595,12 +599,9 @@ private boolean timeToPoll(GatewayConnection connection) {
...
             // Exponential (2^x) dropoff:
-            double unit = pollIntervalUS / 1000.0;
-            double x = (  clock.millis() - connection.connectionTime().plus(connectionTimeToLive).toEpochMilli() ) / unit;
-            double lastX = (  connection.lastPollTime().toEpochMilli() - connection.connectionTime().plus(connectionTimeToLive).toEpochMilli() ) / unit;
-
-            double currentDelayDoublings = Math.log(lastX)/Math.log(2);
-            return (x - lastX) > Math.pow(2, currentDelayDoublings);
+            double connectionEndOfLife = connection.connectionTime().plus(connectionTimeToLive).toEpochMilli();
+            double connectionLastPolled = connection.lastPollTime().toEpochMilli();
+            return clock.millis() - connectionEndOfLife > 2 * (connectionLastPolled - connectionEndOfLife);
```
bratseth, line 604: Thanks for dissolving my math! I just wrote down the function I had in mind :stuck_out_tongue_closed_eyes:

bratseth (approved):

vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
jonmv: 😜

vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/IOThread.java
jonmv: Yep.

PR #14295, opened 2020-09-04
Add content bucket DB metrics and reduce memory footprint

vekterli:
@geirst please review.

The 2nd commit reduces the initial memory footprint of a content node B-tree bucket DB by three orders of magnitude and then some 🙂 This adds up since there are always two of them present (global/non-global).

### DIFF
--- a/storage/src/vespa/storage/bucketdb/lockablemap.hpp
+++ b/storage/src/vespa/storage/bucketdb/lockablemap.hpp
@@ -85,6 +85,15 @@ LockableMap<Map>::getMemoryUsage() const noexcept
         sizeof(std::mutex) + sizeof(std::condition_variable);
 }
 
+template <typename Map>
+vespalib::MemoryUsage LockableMap<Map>::detailed_memory_usage() const noexcept {
+    // We don't have any details for this map type, just count everything as "allocated".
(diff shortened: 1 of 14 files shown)


storage/src/vespa/storage/bucketdb/lockablemap.hpp
```diff
@@ -85,6 +85,15 @@ LockableMap<Map>::getMemoryUsage() const noexcept
         sizeof(std::mutex) + sizeof(std::condition_variable);
 }
 
+template <typename Map>
+vespalib::MemoryUsage LockableMap<Map>::detailed_memory_usage() const noexcept {
+    // We don't have any details for this map type, just count everything as "allocated".
+    size_t used = getMemoryUsage();
+    vespalib::MemoryUsage mem_usage;
+    mem_usage.incAllocatedBytes(used);
```
geirst, line 93: You should probably count everything as 'used' as well.

geirst (approved):
👍

storage/src/vespa/storage/bucketdb/lockablemap.hpp
vekterli: Good point, done

geirst (approved):
👍

PR #14299, opened 2020-09-04
add "Hamming" distance metric for ANN

arnej27959:

@geirst please review
note: only tested manually, but this is very straight-forward.
we need someone (the people who requested the feature, preferably) to check if there's any speed-up at all before adding systemtest and doc for this.

@bratseth FYI


bratseth (approved):
👍

geirst (commented):
👍  But I think we should extend searchlib/src/tests/tensor/distance_functions/distance_functions_test.cpp with new a test case first.

geirst (approved):
👍

PR #14303, opened 2020-09-06
use functions, not objects

havardpe:

@arnej27959 please review
@lesters FYI

### DIFF
--- a/eval/src/vespa/eval/tensor/dense/onnx_wrapper.cpp
+++ b/eval/src/vespa/eval/tensor/dense/onnx_wrapper.cpp
@@ -26,8 +26,17 @@ using vespalib::make_string_short::fmt;
 
 namespace vespalib::tensor {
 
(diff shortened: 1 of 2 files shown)


arnej27959 (approved):
this looks good.  I suggest for reading flow to move the structs to after the definition of the methods their invoke() returns, and then an extra anonymous namespace with the trampolines.
We could also consider doing this:
```
struct Onnx::EvalContext::SelectAdaptParam {
    template <typename ...Ts> static auto invoke() { return adapt_param<Ts...>; }
    static auto from(eval::ValueType::CellType ct) {
        return typify_invoke<1,MyTypify,SelectAdaptParam>(ct);
    }
};
...
            _param_binders.push_back(SelectAdaptParam::from(vespa.cell_type()));
```
putting the trampolines inside the structs instead.

havardpe:
@arnej27959 PTAL

PR #14286, opened 2020-09-04
Balder/balder/group commits to tls 2 remaining

baldersheim:
@toregge PR
This is the last PR which will actually bundle commits together and add value.
I think it requires some extra thought...

PR #14304, opened 2020-09-07
Balder/group commits to tls 2 remaining rebased

baldersheim:
@toregge PR
This is #14286 pluss the last commit that uses the scheduler for regular commit.

### DIFF
--- a/searchlib/src/vespa/searchlib/transactionlog/translogserver.cpp
+++ b/searchlib/src/vespa/searchlib/transactionlog/translogserver.cpp
@@ -1,9 +1,9 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #include "translogserver.h"
+#include "domain.h"
 #include <vespa/searchlib/common/gatecallback.h>
 #include <vespa/vespalib/util/stringfmt.h>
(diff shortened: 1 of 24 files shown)


searchlib/src/vespa/searchlib/transactionlog/translogserver.cpp
```diff
@@ -144,6 +167,9 @@ TransLogServer::TransLogServer(const vespalib::string &name, int listenPort, con
         throw std::runtime_error(make_string("Failed creating tls base dir %s r(%d), e(%d). Requires manual intervention.", _baseDir.c_str(), retval, errno));
     }
     start(*_threadPool);
+    double chunkAgeLimit = vespalib::to_s(_domainConfig.getChunkAgeLimit());
+    auto staleCommitTask = new StaleCommitTask(_supervisor->GetScheduler(), *this, chunkAgeLimit);
+    staleCommitTask->ScheduleNow();
```
toregge, line 172: This looks like a memory leak. You'll need to remember the task, and call Kill() as part of teardown.

toregge (changes requested):

searchlib/src/vespa/searchlib/transactionlog/translogserver.cpp
baldersheim: Plugged

PR #14308, opened 2020-09-07
 Move code from RemoteSession to SessionRepository, take 2

hmusum:
Move tests and avoid a lot of low-level setup code

Does not use `synchronized` for `ensureApplicationLoaded()`, otherwise same as https://github.com/vespa-engine/vespa/pull/14190 (which led to worse performance, the theory is that it was because of an unnecesary `synchronized` qualifier on `ensureApplicationLoaded()`


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java.orig
```diff
@@ -0,0 +1,1100 @@
...
+import static java.nio.file.Files.readAttributes;
+
+/**
+ * The API for managing applications.
+ *
+ * @author bratseth
+ */
+// TODO: Move logic for dealing with applications here from the HTTP layer and make this the persistent component
+//       owning the rest of the state
+public class ApplicationRepository implements com.yahoo.config.provision.Deployer {
```
jonmv, line 111: Does not compile 🤖

hmusum: Urgh, thanks for fixing.

jonmv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -394,13 +402,84 @@ private void loadSessionIfActive(RemoteSession session) {
...
+        if (applicationSet.isPresent()) {
+            return applicationSet.get();
+        }
+
+        ApplicationSet newApplicationSet = loadApplication(session);
+        RemoteSession newSession = new RemoteSession(session.getTenantName(),
+                                    session.getSessionId(),
+                                    session.getSessionZooKeeperClient(),
+                                    Optional.of(newApplicationSet));
+        remoteSessionCache.putSession(newSession);
```
jonmv, line 430: Sure this shouldn't be synchronized, but more fine-grained than on the session repository shared by all sessions?

hmusum: Yes, I think you are right. Need to look into that.

hmusum:

jonmv (approved):

PR #14309, opened 2020-09-07
Move to container engine neutral names

hakonhall:


kkraune (approved):
Great! It is always like that ...

config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
```diff
@@ -37,7 +37,7 @@
     default Optional<File> appDir() { return Optional.empty();}
 
     /** The Docker image repo we want to use for images for this deployment (optional, will use default if empty) */
-    default Optional<DockerImage> wantedDockerImageRepo() { return Optional.empty(); }
```
bratseth, line 40: This breaks the API between the config server and config model. You need to keep the old method with a TODO specifying when it can be removed . Unfortunately, since I think we have pinned applications now that should be by version and not just something like "Remove in October 2020".

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
```diff
@@ -115,7 +115,7 @@ public static ClusterMembership from(String stringValue, Version vespaVersion) {
         return new ClusterMembership(stringValue, vespaVersion, Optional.empty());
     }
 
-    public static ClusterMembership from(String stringValue, Version vespaVersion, Optional<DockerImage> dockerImageRepo) {
```
bratseth, line 118: This also breaks that API since all of config-provisioning is an API for the config model, but based on current usage this one is probably ok.

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterSpec.java
```diff
@@ -49,7 +49,7 @@ private ClusterSpec(Type type, Id id, Optional<Group> groupId, Version vespaVers
     public Id id() { return id; }
 
     /** Returns the docker image repository part of a docker image we want this cluster to run */
-    public Optional<DockerImage> dockerImageRepo() { return dockerImageRepo; }
```
bratseth, line 52: This one also breaks the API but is probably ok.

config-provisioning/src/main/java/com/yahoo/config/provision/ContainerImage.java
```diff
@@ -11,14 +11,14 @@
  *
  * @author mpolden
  */
-public class DockerImage {
```
bratseth, line 14: This renaming breaks the API. Not sure you can do it, maybe by duplicating and subtyping?

config-provisioning/src/main/java/com/yahoo/config/provision/HostSpec.java
```diff
@@ -50,7 +50,7 @@ public HostSpec(String hostname,
                     ClusterMembership membership,
                     Optional<Version> version,
                     Optional<NetworkPorts> networkPorts,
-                    Optional<DockerImage> dockerImageRepo) {
```
bratseth, line 53: More API breakage.

bratseth (changes requested):
There could be more, I didn't review to the end. I think you need to do this in smaller steps and more carefully unfortunately.

hakonhall:
Good catch, I'll reduce the scope of the PR.

hakonhall:
I left DockerImage untouched since it is used by the config model.

docker-api/src/main/java/com/yahoo/vespa/hosted/dockerapi/DockerImpl.java
```diff
@@ -40,10 +39,11 @@
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Stream;
 
-public class DockerImpl implements Docker {
+public class DockerImpl implements ContainerEngine {
```
mpolden, line 46: MIght as well simplify this to `Docker` now.

node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integrationTests/DockerMock.java
```diff
@@ -23,7 +23,7 @@
  *
  * @author freva
  */
-public class DockerMock implements Docker {
+public class DockerMock implements ContainerEngine {
```
mpolden, line 26: `ContainerEngineMock`?

hakonhall: Fixed

docker-api/src/main/java/com/yahoo/vespa/hosted/dockerapi/DockerImpl.java
hakonhall: DockerEngine?

mpolden (approved):

PR #14315, opened 2020-09-07
guess batch dimension when inference fails

havardpe:

@arnej27959 please review
@lesters FYI

### DIFF
--- a/eval/src/vespa/eval/tensor/dense/onnx_wrapper.h
+++ b/eval/src/vespa/eval/tensor/dense/onnx_wrapper.h
@@ -12,6 +12,7 @@
 #include <vespa/eval/eval/value_type.h>
 #include <vector>
 #include <map>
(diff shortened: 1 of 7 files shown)


eval/src/vespa/eval/tensor/dense/onnx_wrapper.h
```diff
@@ -83,8 +84,9 @@ class Onnx {
     private:
         std::map<vespalib::string,eval::ValueType> _input_types;
         std::map<vespalib::string,size_t> _symbolic_sizes;
+        std::set<size_t> _unknown_sizes;
```
arnej27959, line 87: consider renaming to "_resolved_sizes" since they are no longer unknown when added to the set...

arnej27959 (approved):
looks good, one minor idea for naming to consider...

PR #14316, opened 2020-09-07
Remove redundant logging

bratseth:
There is another info message produced elsewhere on session activation with more detail


hmusum:
This message is produced on all config servers in a cluster, the other message only on the config server that receives the HTTP  deployment request (if it is the one I think you refer to), so it is useful. But I am working on getting rid of remote sessions and this message altogether...

bratseth:
Yes that's right, ok

PR #14321, opened 2020-09-08
Cluster size independent autoscaling reaction time

bratseth:
This makes the window considered about 1 hour regardless of cluster size.

@rafan fyi

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
@@ -20,15 +20,7 @@ import java.util.stream.Collectors;
  */
 public class Autoscaler {
 
-    /*
-     TODO:
-     - Scale group size
-     - Consider taking spikes/variance into account
-     - Measure observed regulation lag (startup+redistribution) and take it into account when deciding regulation observation window
-     - Scale by performance not just load+cost
-     */
-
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -113,8 +105,9 @@ private boolean similar(double r1, double r2, double threshold) {
                                                           resource,
                                                           clusterNodes.stream().map(Node::hostname).collect(Collectors.toList()));
 
-        if (window.measurementCount() < minimumMeasurements) return Optional.empty();
-        if (window.hostnames() != clusterNodes.size()) return Optional.empty(); // Regulate only when all nodes are measured
+        // Require a total number of measurements scaling with the number of nodes but don't require them from all nodes
```
freva, line 108: The second part of this reads more like we don't require having any measurements from all nodes, instead of not requiring at least `minimumMeasurementsPerNode` for all nodes

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
bratseth: Better now?

freva (approved):
:100:

PR #14327, opened 2020-09-08
Revert "set omp num threads to 1"

jonmv:
Reverts vespa-engine/vespa#14320

@aressem @havardpe @arnej27959 possibly killed performance tests.


aressem (approved):

aressem:
@havardpe @arnej27959 We probably have some 3rdparty libraries that are affected by this environment variable. Maybe openblas ?

PR #14337, opened 2020-09-09
Arnej/more time for prepare for restart 

arnej27959:

NOTE: this changes (indirectly) the waiting time for vespa-stop-services, which means that restarting vespa on a node may take 10-15 minutes instead of 1-1.5 minutes.

However, in those cases it will also be much more likely that proton will manage to actually flush all the data it needs to flush in order to restart quickly afterwards, so the total restart time should not become any worse.

@baldersheim please review

### DIFF
--- a/vespalog/src/logger/runserver.cpp
+++ b/vespalog/src/logger/runserver.cpp
@@ -390,16 +390,16 @@ int main(int argc, char *argv[])
             } else {
                 fprintf(stdout, "%s was running with pid %d, sending SIGTERM\n",
                     service, pid);
-                if (killpg(pid, SIGTERM) != 0) {
+                if (kill(pid, SIGTERM) != 0) {
                     fprintf(stderr, "could not signal %d: %s\n", pid,
                             strerror(errno));
                     return 1;
                 }
             }
-            fprintf(stdout, "Waiting for exit (up to 60 seconds)\n");
-            for (int cnt(0); cnt < 1800; cnt++) {
+            fprintf(stdout, "Waiting for exit (up to 15 minutes)\n");
+            for (int cnt(0); cnt < 86400; cnt++) {
                 usleep(100000); // wait 0.1 seconds
-                if ((cnt > 300) && (cnt % 100 == 0)) {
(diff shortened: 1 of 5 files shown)


baldersheim (approved):
Will this give extra time only for prepare restart, or for all processes ?
If latter, would it be possible to differentiate ?

vespalog/src/logger/runserver.cpp
```diff
@@ -390,16 +390,16 @@ int main(int argc, char *argv[])
...
-                if (killpg(pid, SIGTERM) != 0) {
+                if (kill(pid, SIGTERM) != 0) {
                     fprintf(stderr, "could not signal %d: %s\n", pid,
                             strerror(errno));
                     return 1;
                 }
             }
-            fprintf(stdout, "Waiting for exit (up to 60 seconds)\n");
-            for (int cnt(0); cnt < 1800; cnt++) {
+            fprintf(stdout, "Waiting for exit (up to 15 minutes)\n");
```
geirst, line 399: @arnej27959 and @baldersheim: If we are not able to differentiate between processes, perhaps we should be conservative at first and increase this to only 120 seconds?

arnej27959:
differentiation: yes and no.
The outermost stopping command (in runserver) doesn't know what it's running at all, so it needs a longer timeout than anything that is useful inside.
config-sentinel has a 1 minute (or to be exact, 58 seconds) timeout.  But it starts the timer *after* running any pre-shutdown hooks (implicitly).

So the sentinel first sends a SIGTERM to some random services, then hits the synchronous call to system() when calling terminate() for the searchnode (proton) service, where it waits until that is finished, then signals the rest of the services.

So for nodes that doesn't run proton the timeout is still 58 seconds inside the sentinel.
For nodes with proton it's "however long the prepareShutdown RPC call takes" plus 58 seconds.

I'm going to move code slightly so prepare_for_shutdown is its own phase before sending any signals,
but it would be even better to run it async with its own timer.

baldersheim (approved):

baldersheim (approved):
Thanks for the explanation. This semes like a better solution.

PR #14338, opened 2020-09-09
Support 'id' as field name in C++ document selection lexing/parsing

vekterli:
@geirst please review
@kkraune FYI

'id' is normally a reserved keyword and wasn't explicitly allowed for
the purpose of field name identifiers. Change `ID` to be lexed as a
string token to allow for preserving the original casing when used as
an identifier.

Also verify existing semantics of 'id' as field name in Java parser.


geirst (approved):
👍

PR #14340, opened 2020-09-09
Move storageserver rpc unit tests to separate subfolder with its own …

geirst:
…gtest runner.

@vekterli please review


vekterli (approved):
:+1:

PR #14341, opened 2020-09-09
Explicit sync and async dispatch

vekterli:
@geirst please review.

Delete now unused `MessageEnqueuer` class. Also purge old, unused MessageDispatcher concept that was dropped on the factory floor at some point.


geirst (approved):
👍

PR #14364, opened 2020-09-10
set omp num threads to 1 when running *proton*

havardpe:

It seems that the java container uses open mp for something (not sure what). This tries setting the number of threads to 1 for proton only.

@arnej27959 please review and merge
@aressem @toregge @lesters FYI


arnej27959 (approved):
👍

baldersheim (commented):
This should have been added in the config model instead. There we build the command to be executed and we do stuff like that already.

havardpe:
> This should have been added in the config model instead. There we build the command to be executed and we do stuff like that already.

here you go; https://github.com/vespa-engine/vespa/pull/14372

PR #14372, opened 2020-09-10
set env variable in the config model

havardpe:

Set environment variable in the config model instead.

Note that we only need this until we are able to make our own onnxruntime without openmp support. Not sure if we can just remove it again or if we are stuck with it until Vespa 8 once it is part of the public api...

@baldersheim please review


havardpe:
@arnej27959 please have a look as well

baldersheim (approved):

baldersheim (approved):
Merge at will

PR #14374, opened 2020-09-11
Reduce visibility to avoid having to see everything.

baldersheim:
@toregge PR
Here should be no semantic changes.

### DIFF
--- a/searchlib/src/vespa/searchlib/transactionlog/translogserver.h
+++ b/searchlib/src/vespa/searchlib/transactionlog/translogserver.h
@@ -88,7 +88,6 @@ private:
     std::unique_ptr<FastOS_ThreadPool>  _threadPool;
     std::unique_ptr<FNET_Transport>     _transport;
     std::unique_ptr<FRT_Supervisor>     _supervisor;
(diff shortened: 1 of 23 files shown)


searchlib/src/vespa/searchlib/transactionlog/translogserver.h
```diff
@@ -87,12 +88,15 @@ class TransLogServer : public document::Runnable, private FRT_Invokable, public
     std::unique_ptr<FastOS_ThreadPool>  _threadPool;
     std::unique_ptr<FNET_Transport>     _transport;
     std::unique_ptr<FRT_Supervisor>     _supervisor;
+    std::unique_ptr<std::thread>        _staleCommitThread;
```
toregge, line 91: This variable seems to be unused.

toregge (approved):

searchlib/src/vespa/searchlib/transactionlog/translogserver.h
baldersheim: Fixed

PR #14375, opened 2020-09-11
Do not start config server maintainers until bootstrapping is done

hmusum:


hakonhall (approved):
But some of the maintainers should halt a successful bootstrap if they fail, perhaps?

hmusum:
Off the top of my head I don't know of any scenario where halting a bootstrap because a maintainer fails would be of benefit. Did you have anything particular in mind?

hakonhall:
The two maintainers this PR removes: file distribution and session manager. If they fail it may because of full disk and we'd rather not start the config server?

PR #14377, opened 2020-09-11
Correct allocated memory accounting for vespalib::string.

toregge:
@arnej27959 : please review


baldersheim:
You might want to add a unit test.

toregge:
The unit test was compiling with clang.

PR #14378, opened 2020-09-11
Bratseth/request type 2

bratseth:
@bjorncs by taking the statistics when we have the response we can let handlers executing in separate threads determine the request type. 

This makes the code simpler as well so I don't see why it was done the way it was, but maybe there's something I don't get?


bjorncs (commented):
Jdisc will no longer aggregate response metrics for requests not processed by a jdisc request handler. E.g responses generated by other servlets, jaxrs resources and some Jetty handlers.

bratseth:
We'll try something else for now.

PR #14387, opened 2020-09-13
Create remote session only when needed

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionCache.java
```diff
@@ -30,5 +26,7 @@ public synchronized SESSIONTYPE getSession(long id) {
     public synchronized List<SESSIONTYPE> getSessions() {
         return new ArrayList<>(sessions.values());
     }
+
+    public synchronized boolean sessionExists(long id) { return sessions.containsKey(id); }
     
 }
```
jonmv, line 32: Soon time to just replace this class with a `ConcurrentHashMap`? :)

jonmv (approved):

PR #14389, opened 2020-09-14
Make sure to delete remote session if local session does not exist

hmusum:
Bugfix: If request is received on a server that does not have the
local session (e.g. if it has not been created yet, this might take
some time) we still need to delete the remote session


jonmv:
Isn't this called by watchers? Ah, but I see the only caller of `RemoteSession.delete()` today is GC of expired sessions. OK, fine.

jonmv (approved):

jonmv:
But why should it be called only conditionally?

hmusum:
Because if a local session exists a watcher will see that it has been deleted and removed it. This is a mess and I want to clean it up, but history has shown that we need to take small steps to fix and try to understand why the fix doesn't work when we break something

PR #14395, opened 2020-09-14
Add feature flag to use direct RPC for Storage API communication betw…

geirst:
…een distributor and content node.

@vekterli please review

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -196,6 +196,12 @@ public class Flags {
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundBooleanFlag USE_DIRECT_STORAGE_API_RPC = defineFeatureFlag(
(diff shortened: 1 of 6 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -196,6 +196,12 @@
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundBooleanFlag USE_DIRECT_STORAGE_API_RPC = defineFeatureFlag(
+            "use-direct-storage-api-rpc", false,
+            "Whether to use direct RPC for Storage API communication between distributor and content node.",
```
vekterli, line 201: Could just leave this as "(...) communication between content cluster nodes.", as it also involves storage<->storage communication.

vekterli (approved):
:+1:

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
geirst: Fixed

PR #14396, opened 2020-09-14
Use rwlock

baldersheim:
@toregge PR


toregge (approved):
Note that std::shared_mutex came with C++17.

baldersheim:
We are already using it. I also think that we are building with c++17 flags enabled.

PR #14400, opened 2020-09-14
Initial direct StorageAPI RPC functionality

vekterli:
@geirst please review. Only 1cd5272 added after last review round.
@baldersheim could you take a look at this as well to see if there are any obvious missing parts or dents in the bodywork? In particular for the `StorageApiRpcService` code which takes over for the MessageBus `RPCSendV2` functionality.

Note that this is _not yet_ considered production ready, so still quite a few TODOs/FIXMEs in code paths that aren't enabled by default. But it's sufficiently functional that we can very soon begin internal system/performance testing (we need a configurable RPC thread pool size before that, though).


geirst (approved):
👍

baldersheim (approved):
Merge at will

PR #14410, opened 2020-09-15
Configurable number of network threads used by the shared rpc resource.

geirst:
@vekterli please review
@baldersheim FYI

### DIFF
--- a/storage/src/vespa/storage/config/stor-communicationmanager.def
+++ b/storage/src/vespa/storage/config/stor-communicationmanager.def
@@ -59,3 +59,6 @@ skip_thread bool default=false
 ## Whether to use direct P2P RPC protocol for all StorageAPI communication
 ## instead of going via MessageBus.
 use_direct_storageapi_rpc bool default=false
+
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/config/stor-communicationmanager.def
```diff
@@ -59,3 +59,6 @@ skip_thread bool default=false
 ## Whether to use direct P2P RPC protocol for all StorageAPI communication
 ## instead of going via MessageBus.
 use_direct_storageapi_rpc bool default=false
+
+## The number of network (FNET) threads used by the shared rpc resource.
+rpc.num_network_threads int default=1
```
vekterli, line 64: Consider throwing in a newline at the end

vekterli (approved):
:+1:

storage/src/vespa/storage/config/stor-communicationmanager.def
geirst: Fixed

PR #14420, opened 2020-09-16
Add explicit I/O config for ONNX models config

lesters:
@havardpe For discussion. Allows explicit config of inputs and outputs.


havardpe (commented):
looks good to me

PR #14422, opened 2020-09-16
Make RPC compression configurable and use better defaults

vekterli:
@geirst please review

Introduces a dedicated RPC compression sub-config. Default values are the same as for MessageBus compression.


geirst (approved):
👍

PR #14426, opened 2020-09-16
Add benchmark for feeding at SPI layer.

toregge:
@baldersheim : please review

### DIFF
--- a/searchcore/src/apps/vespa-spi-feed-bm/vespa_spi_feed_bm.cpp
+++ b/searchcore/src/apps/vespa-spi-feed-bm/vespa_spi_feed_bm.cpp
@@ -1,0 +1,464 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/vespalib/testkit/testapp.h>
+
+#include <tests/proton/common/dummydbowner.h>
+#include <vespa/config-imported-fields.h>
+#include <vespa/config-rank-profiles.h>
+#include <vespa/config-summarymap.h>
+#include <vespa/fastos/file.h>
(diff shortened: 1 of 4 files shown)


baldersheim:
@geirst and @vekterli Please also take a look

baldersheim (approved):
This looks good as an initial start.
I would guess that the following improvements should be considered.
1 - Documents should made pre generated and serialised and then being created from serialised form to mimic how it happens in a live system. Due to lazyness deserialisation does not happen on the way to the TLS, but when forked off to the writer threads.
2 - Parameterise some of the test values. Number of unique documents to feed, number of refeeds, number of spi threads, etc.

geirst (commented):
👍

searchcore/src/apps/vespa-spi-feed-bm/vespa_spi_feed_bm.cpp
```diff
@@ -0,0 +1,464 @@
...
+    
+    PersistenceProviderFixture();
+    ~PersistenceProviderFixture();
+    void create_document_db();
+    uint32_t num_buckets() const { return (1u << _bucket_bits); }
+    Bucket make_bucket(uint32_t i) const { return Bucket(document::Bucket(_bucket_space, BucketId(_bucket_bits, i & (num_buckets() - 1))), PartitionId(0)); }
+    BucketId get_bucket_id(const Document &document) const {
+        auto &id = document.getId();
+        auto &gid = id.getGlobalId();
+        return BucketId(_bucket_bits, gid.convertToBucketId().getRawId()).stripUnused();
```
vekterli, line 253: Could also use `document::BucketIdFactory` to do this

vekterli (commented):
:+1:

PR #14429, opened 2020-09-16
Fix content-types

freva:
`?recursive=true` is actually already supported... :tada: 

This could all be reduced to just `Files.probeContentType()`, which would basically read from `/etc/mime.types`, but unfortunately that file is quite different between different hosts and it would be hard to have any tests with that. It would also be tricky to extend it if we wanted to handle some special types.

Just out of curiosity, I counted file extensions for all deployment in main, here's ones with >= 5 occurrences:
```
 11094 xml
  3064 expression
  1348 template
  1276 sd
   667 json
   299 jar
   155 properties
    70 py
    65 txt
    35 NONE
    27 model
    18 regex
    15 woff
    15 woff2
    12 meta
    10 map
    10 js
     9 pig
     9 mod
     9 30
     9 config
     9 schema
     6 fsa
     5 png
```


hmusum (approved):

hmusum:
I guess this might break some system tests, I haven't checked.

freva:
Thanks!

aressem:
Seems to have broken a few.

PR #14438, opened 2020-09-17
Arnej/add stored mappings 6

arnej27959:

@havardpe please take a look - preview without class comments for now


arnej27959:
@havardpe PTAL

havardpe (approved):
I think this is a good starting point, where we can add some microbenchmarks and see how the performance is compared to the traditional sparse/dense tensor implementations.

However, I do think even more things should use packed mixed tensor as the concept name;
packed_mixed_tensor_test, PackedMixedTensorBuilder, PackedMixedTensorBuilderFactory

havardpe:
@geirst should probably also take a look. I think we want to clean up and improve naming, but not everything needs to be perfect in the first iteration.

PR #14441, opened 2020-09-17
Take lock before deleting

hmusum:


jonmv (approved):

jonmv:
Probably a good idea :)

PR #14451, opened 2020-09-18
Bjorncs/feed handler cleanup

bjorncs:
@baldersheim  or @jonmv please review


vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/ClientFeederV3.java
```diff
@@ -191,30 +169,14 @@ private int getOverloadReturnCode(HttpRequest request) {
...
-            }
-            if (result.isAccepted()) {
-                return result;
-            }
-            Thread.sleep(100);
-        }
-        return result;
+    private Result sendMessage(DocumentOperationMessageV3 msg) {
+        msg.getMessage().pushHandler(feedReplyHandler);
+        return sourceSession.getResource().sendMessage(msg.getMessage());
```
jonmv, line 174: Did you intend for this to be `sendBlocking`, as we discussed?

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/ClientFeederV3.java
```diff
@@ -244,27 +205,21 @@ private void feed(FeederSettings settings,
                 repliesFromOldMessages.add(createOperationStatus(message.get().getOperationId(),
                                                                  result.getError().getMessage(),
                                                                  ErrorCode.TRANSIENT_ERROR,
-                                                                 false,
-                                                                 message.get().getMessage()));
+                        message.get().getMessage()));
             } else {
-                // should probably not happen, but everybody knows stuff that
-                // shouldn't happen, happens all the time
-                boolean isConditionNotMet = result.getError().getCode() == DocumentProtocol.ERROR_TEST_AND_SET_CONDITION_FAILED;
```
jonmv, line 252: 👍

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
```diff
@@ -44,31 +43,40 @@
...
-    public FeedHandlerV3(Executor executor,
+    public FeedHandlerV3(ContainerThreadPool threadpool,
                          Metric metric,
                          AccessLog accessLog,
                          DocumentmanagerConfig documentManagerConfig,
                          SessionCache sessionCache,
-                         ThreadpoolConfig threadpoolConfig,
                          DocumentApiMetrics metricsHelper) {
+        this(threadpool.executor(),
+                () -> threadpool.queuedTasks() > 0,
```
jonmv, line 57: I believe this would oftentimes be true, even though there's no need to throttle. Consider a larger number than 0.

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/ClientFeederV3.java
jonmv: NVM.

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
bjorncs: @baldersheim Any input?

baldersheim: Hard to tell what is the correct value here. Perhaps queuesize/2 is a more conservative number. Might be better to start with a larger number and reduce. I guess we can  look at queue size afterwards and see how it behaves.

baldersheim (approved):
I guess this is good to with a more conservative throttling limit. Especially to handle feeding from the grid with many clients.

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/FeedHandlerV3.java
bjorncs: I'm not really a fan of this ad-hoc throttling. I think is better to reuse the rejection logic in the threaded request handler base class. The response code must be overridden to 429, but that should be easy to implement.

PR #14454, opened 2020-09-18
Jonmv/async document v1

jonmv:
@bjorncs or @vekterli please review.

### DIFF
--- a/documentapi/src/main/java/com/yahoo/documentapi/DocumentAccess.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/DocumentAccess.java
@@ -55,8 +55,11 @@ public abstract class DocumentAccess {
      * while attempting to create such an object, this method will throw an
      * exception.
      *
+     * @deprecated Inject a DocumentManagerConfig and create a MessageBusDocumentAccess from this instead.
+     *
      * @return a running document access object with all default configuration
      */
+    @Deprecated(since = "7")
     public static DocumentAccess createDefault() {
         return new com.yahoo.documentapi.messagebus.MessageBusDocumentAccess();
     }
@@ -67,12 +70,11 @@ public abstract class DocumentAccess {
      * @param params the parameters to use for setup
      */
     protected DocumentAccess(DocumentAccessParams params) {
-        super();
         if (params.documentmanagerConfig().isPresent()) { // our config has been injected into the creator
             documentTypeManager = new DocumentTypeManager(params.documentmanagerConfig().get());
             documentTypeConfigSubscriber = null;
         }
-        else { // fallback to old style subscription
+        else { // fallback to old style subscription — this should be avoided
             documentTypeManager = new DocumentTypeManager();
             documentTypeConfigSubscriber = DocumentTypeManagerConfigurer.configure(documentTypeManager, params.getDocumentManagerConfigId());
         }
--- a/documentapi/src/main/java/com/yahoo/documentapi/DocumentIdResponse.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/DocumentIdResponse.java
@@ -16,8 +16,7 @@ public class DocumentIdResponse extends Response {
 
     /** Creates a successful response */
     public DocumentIdResponse(long requestId) {
-        super(requestId);
-        documentId = null;
+        this(requestId, null);
     }
 
     /**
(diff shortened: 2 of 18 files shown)


documentapi/src/main/java/com/yahoo/documentapi/DocumentAccess.java
```diff
@@ -55,8 +55,11 @@
      * while attempting to create such an object, this method will throw an
      * exception.
      *
+     * @deprecated Inject a DocumentManagerConfig and create a MessageBusDocumentAccess from this instead.
```
bratseth, line 58: But that's more difficult ... doesn't this just work?

bratseth (approved):
LGTM apart from that question

documentapi/src/main/java/com/yahoo/documentapi/DocumentAccess.java
jonmv: @baldersheim wqs of the opinion this should go.

baldersheim: It uses self subscription.
1 - It means that at will change out of sync with the other components also using the same config.
   - Components will be out of sync in for a short time.
   - It is either very costly due to extra synchronisation. However I think it is worse, they will never agree, as there are thread visibility issues.
2 - It leaks resources. Every heapdump I have checked where the customer has done this, it is leaking threads and other resources.

3 - I think that having the config injected is a simple remediation for this mess. I thought that was the whole reason for having config injected and creating final components.

This is one og our bad past patterns.

bratseth: Yes DocumentAccess is a factory for clients, if you're in a container you should have everything injected. This used to be convenient when running a client on a node which can reach config servers. It's not relevant to cloud deployments, so fine I guess.

Btw, we should have an object you can have injected directly to make it more obvious what to do when running in a container, perhaps that could be done as part of this work.

jonmv: Yes, I thought the same thing. The document access is perhaps something we wish to create only when needed, so a factory which is always present, based on the relevant config?

bratseth: Yes, or preferably something like a session (or, two - one sync and one async) which is always present so we don't need to destroy it - but otoh I'm not sure if we want to create the connections if we don't need them.

jonmv: But sync sessions are often used one per client? So rather a session factory, i.e. a document access? How about a lazy doc access, then? If one of those is always enough.

jonmv: A doc access is also what you need for visitor sessions.

jonmv: Hmm, why the sync sessions are used one per thread/operation in the current document/v1, I don't know ... Seems like an error to me. So you could indeed have injectable `Sync` and `Async`-`Session`s. That doesn't cover the other types of sessions, though.

jonmv: If anything, using multiple `SyncSessions` to handle a feed of operations breaks operation sequencing ...

bratseth: Yes, that should work if we're ok with always having instances of these sessions. 

What other types of sessions?

jonmv: VisitorSession. I guess SubscriptionSession never got an implementation. See #14461 for a suggestion of a lazy `DocumentAccess` provider.

documentapi/src/main/java/com/yahoo/documentapi/DocumentIdResponse.java
```diff
@@ -36,9 +35,9 @@ public DocumentIdResponse(long requestId, DocumentId documentId) {
      * @param textMessage the message to encapsulate in the Response
      * @param success     true if the response represents a successful call
      */
+    @Deprecated(since = "7") // TODO: Remove on Vespa 8
```
bjorncs, line 38: Consider setting `forRemoval=true`. IDE/compiler gives stronger feedback if true.

bjorncs (commented):
👍

PR #14455, opened 2020-09-18
30s down-moratorium before allowing suspension

hakonhall:
If all services of a node are down, we used allow suspension.  If those services are monitored with /state/v1/health, we have the timestamp the service became unhealthy - the "since" timestamp.  Now we will require the services to have been down for at least 30s before allowing suspension based on unhealthiness.

Also, for config servers only, we will log all healthiness transitions to track down some orchestrator issues.

### DIFF
--- a/application-model/src/main/java/com/yahoo/vespa/applicationmodel/ClusterId.java
+++ b/application-model/src/main/java/com/yahoo/vespa/applicationmodel/ClusterId.java
@@ -12,6 +12,9 @@ public class ClusterId {
 
     public static final ClusterId CONFIG_SERVER = new ClusterId("zone-config-servers");
     public static final ClusterId CONTROLLER = new ClusterId("controller");
+    public static final ClusterId CONFIG_SERVER_HOST = new ClusterId("configserver-host");
+    public static final ClusterId CONTROLLER_HOST = new ClusterId("configserver-host");
+    public static final ClusterId TENANT_HOST = new ClusterId("tenant-host");
 
     private final String id;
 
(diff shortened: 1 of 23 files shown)


application-model/src/main/java/com/yahoo/vespa/applicationmodel/ClusterId.java
```diff
@@ -12,6 +12,9 @@
 
     public static final ClusterId CONFIG_SERVER = new ClusterId("zone-config-servers");
     public static final ClusterId CONTROLLER = new ClusterId("controller");
+    public static final ClusterId CONFIG_SERVER_HOST = new ClusterId("configserver-host");
+    public static final ClusterId CONTROLLER_HOST = new ClusterId("configserver-host");
```
hmusum, line 16:
```suggestion
    public static final ClusterId CONTROLLER_HOST = new ClusterId("controller-host");
```

hmusum (changes requested):
One suggested change, otherwise LGTM

application-model/src/main/java/com/yahoo/vespa/applicationmodel/ClusterId.java
hakonhall: Thank you!

hmusum (approved):

hakonhall:
I'm unable to reproduce the Travis failure (ControllerTest, testDevDeployment), and I no longer have the ability to trigger another run on Travis.  I'll merge and make a revert just in case.

PR #14456, opened 2020-09-18
Revert "30s down-moratorium before allowing suspension"

hakonhall:
Reverts vespa-engine/vespa#14455


hakonhall:
In case original pr breaks the build

PR #14461, opened 2020-09-21
Jonmv/async document v1

jonmv:
@bratseth or @bjorncs please review.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
@@ -181,6 +181,7 @@ public abstract class ContainerCluster<CONTAINER extends Container>
         addSimpleComponent(com.yahoo.metrics.simple.jdisc.JdiscMetricsFactory.class.getName(), null, MetricProperties.BUNDLE_SYMBOLIC_NAME);
         addSimpleComponent("com.yahoo.container.jdisc.state.StateMonitor");
         addSimpleComponent("com.yahoo.container.jdisc.ContainerThreadFactory");
+        addSimpleComponent(com.yahoo.container.core.documentapi.MessageBusDocumentAccessProvider.class.getName());
         addSimpleComponent("com.yahoo.container.handler.VipStatus");
         addSimpleComponent(com.yahoo.container.handler.ClustersStatus.class.getName());
         addJaxProviders();
--- a/container-core/src/main/java/com/yahoo/container/core/documentapi/MessageBusDocumentAccessProvider.java
+++ b/container-core/src/main/java/com/yahoo/container/core/documentapi/MessageBusDocumentAccessProvider.java
@@ -1,0 +1,55 @@
+package com.yahoo.container.core.documentapi;
+
+import com.google.inject.Inject;
+import com.yahoo.component.AbstractComponent;
+import com.yahoo.container.di.componentgraph.Provider;
+import com.yahoo.document.config.DocumentmanagerConfig;
+import com.yahoo.documentapi.DocumentAccess;
+import com.yahoo.documentapi.messagebus.MessageBusDocumentAccess;
+import com.yahoo.documentapi.messagebus.MessageBusParams;
+import com.yahoo.documentapi.messagebus.loadtypes.LoadTypeSet;
+import com.yahoo.vespa.config.content.LoadTypeConfig;
+
+/**
+ * Has a lazily populated reference to a {@link MessageBusDocumentAccess}.
+ *
+ * @author jonmv
+ */
+public class MessageBusDocumentAccessProvider extends AbstractComponent implements Provider<DocumentAccess> {
+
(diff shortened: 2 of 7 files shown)


bratseth:
Just to mention it: I think we probably want to keep not_found a success for remove but change the other one.

Regarding providing this or sessions: This API might make people create a lot of sessions, and I think the sessions you create need to be destroyed.

bratseth (approved):

jonmv:
Yes, they should be destroyed, but how is this different from using a `DocumentAccess` obtained through `DocumentAccess.createDefault()`?

bratseth:
It's not - so feel free to merge. Just wondering if we couldn't do something better.

jonmv:
Aha, yes, agreed ... :)

container-core/src/main/java/com/yahoo/container/core/documentapi/MessageBusDocumentAccessProvider.java
```diff
@@ -0,0 +1,55 @@
...
+        this.documentmanagerConfig = documentmanagerConfig;
+        this.loadTypeConfig = loadTypeConfig;
+    }
+
+    @Override
+    public DocumentAccess get() {
+        synchronized (monitor) {
+            if (access == null) {
+                access = new MessageBusDocumentAccess((MessageBusParams) new MessageBusParams(new LoadTypeSet(loadTypeConfig)).setDocumentmanagerConfig(documentmanagerConfig));
+                if (shutDown)
```
baldersheim, line 37: Why creating it at all if shutdown is set. Can you just add it to the outer if as '&& !shutDown'

jonmv: Then you'd return `null`. I thought it better to preserve the existing error message you'd get when trying to use a shut down document access.

jonmv: Alternatively could throw something custom that was clear on this.

baldersheim: Throw instead, no point in constructing mbus access and shut it down just to preserve error message in this case.
This should never happen, but throwing is probably a good idea.

jonmv: Can do. 'twas my first impulse as well. And, yes, this should never happen, unless someone breaks DI.

container-core/src/main/java/com/yahoo/container/core/documentapi/MessageBusDocumentAccessProvider.java
```diff
@@ -0,0 +1,55 @@
...
+                    access.shutdown();
+            }
+            return access;
+        }
+    }
+
+    @Override
+    public void deconstruct() {
+        synchronized (monitor) {
+            if ( ! shutDown) {
```
bjorncs, line 47: IMHO, throwing an exception for duplicate calls to `deconstruct()` is ok - it should not happen. This is too defensive.

container-core/src/main/java/com/yahoo/container/core/documentapi/MessageBusDocumentAccessProvider.java
```diff
@@ -0,0 +1,55 @@
...
+public class MessageBusDocumentAccessProvider extends AbstractComponent implements Provider<DocumentAccess> {
+
+    private final DocumentmanagerConfig documentmanagerConfig;
+    private final LoadTypeConfig loadTypeConfig;
+    private final Object monitor = new Object();
+    private boolean shutDown = false;
+    private DocumentAccess access = null;
+
+    @Inject
+    public MessageBusDocumentAccessProvider(DocumentmanagerConfig documentmanagerConfig, LoadTypeConfig loadTypeConfig) {
```
bjorncs, line 27: Is this config available for all types of container clusters (e.g clustercontroller, metricsproxy, logserver)?

config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
```diff
@@ -181,6 +181,7 @@ public ContainerCluster(AbstractConfigProducer<?> parent, String subId, String n
         addSimpleComponent(com.yahoo.metrics.simple.jdisc.JdiscMetricsFactory.class.getName(), null, MetricProperties.BUNDLE_SYMBOLIC_NAME);
         addSimpleComponent("com.yahoo.container.jdisc.state.StateMonitor");
         addSimpleComponent("com.yahoo.container.jdisc.ContainerThreadFactory");
+        addSimpleComponent(com.yahoo.container.core.documentapi.MessageBusDocumentAccessProvider.class.getName());
```
bjorncs, line 184: Should this component be available for non-application-container-clusters?

bjorncs (commented):
See comments

PR #14463, opened 2020-09-21
Create remote session in addLocalSession() only when it does not exist

hmusum:
Yet another attempt. Please review only


jonmv (approved):
Look good :)

PR #14484, opened 2020-09-22
- Group commits to the TLS and sync to disk before acking.

baldersheim:
- Add zstd as compression.
- No immediate commit. Delay ack until fully committed instead.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/transactionlog/domain.cpp
+++ b/searchlib/src/vespa/searchlib/transactionlog/domain.cpp
@@ -376,9 +376,9 @@ Domain::commitChunk(std::unique_ptr<CommitChunk> chunk, const vespalib::MonitorG
 
 void
 Domain::doCommit(std::unique_ptr<CommitChunk> chunk) {
-    if (chunk->empty()) return;
-
     const Packet & packet = chunk->getPacket();
+    if (packet.empty()) return;
+    
     vespalib::nbostream_longlivedbuf is(packet.getHandle().data(), packet.getHandle().size());
     Packet::Entry entry;
     entry.deserialize(is);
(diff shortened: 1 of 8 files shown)


searchlib/src/vespa/searchlib/transactionlog/domain.cpp
```diff
@@ -318,22 +323,78 @@ Domain::optionallyRotateFile(SerialNum serialNum) {
...
 
 void
-Domain::append(const Packet & packet, Writer::DoneCallback onDone)
-{
-    (void) onDone;
+Domain::commitIfFull(const vespalib::MonitorGuard &guard) {
+    if (_currentChunk->sizeBytes() > _config.getChunkSizeLimit()) {
+        auto completed = std::move(_currentChunk);
+        _currentChunk = std::make_unique<CommitChunk>(_config.getChunkSizeLimit(), completed->stealCallbacks());
+        if (completed) {
```
toregge, line 358: This if should be unneeded, we know that completed is set, otherwise previous line would have caused a crash.

searchlib/src/vespa/searchlib/transactionlog/domain.cpp
```diff
@@ -318,22 +323,78 @@ Domain::optionallyRotateFile(SerialNum serialNum) {
...
+    assert(guard.monitors(_currentChunkMonitor));
+    auto chunk = std::move(_currentChunk);
+    _currentChunk = createCommitChunk(_config);
+    return chunk;
+}
+
+bool
+Domain::commitChunk(std::unique_ptr<CommitChunk> chunk, const vespalib::MonitorGuard & chunkOrderGuard) {
+    assert(chunkOrderGuard.monitors(_currentChunkMonitor));
+    if ( ! chunk->getPacket().empty()) {
```
toregge, line 375: This guard should be moved into the following lambda. We want to keep chunk alive until previous chunks have been committed.

toregge (changes requested):

searchlib/src/vespa/searchlib/transactionlog/domain.cpp
baldersheim: Correct, fixed

searchlib/src/vespa/searchlib/transactionlog/domain.cpp
baldersheim: Correct, fixed.

toregge (approved):

PR #14489, opened 2020-09-22
Avoid address stringification in common lookup path

vekterli:
@geirst please review

Precompute internal address hash over <cluster, type, index> tuple.
No other fields are included in the hash, as this is only used for storage API lookups.

Remove automatic retry field from address, as we never use MBus retry functionality in the backend communication protocols.


geirst (approved):
👍

baldersheim (approved):

PR #14493, opened 2020-09-22
Revert "Revert "- Group commits to the TLS and sync to disk before acking.""

baldersheim:
Reverts vespa-engine/vespa#14490


baldersheim:
Superceeded by #14495

PR #14499, opened 2020-09-23
Add zone dimension to use-config-server-vip flag

hmusum:


freva (approved):
Also add it to https://github.com/vespa-engine/vespa/blob/ddabd7f27e885d2d04c7d80523ab69b65bbc7af1/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/zone/v2/ZoneApiHandler.java#L116
?

hmusum:
Yes, working on it, I can of course add it to this PR, will do.

freva (approved):
:100:

PR #14506, opened 2020-09-23
Arnej/add new encode decode

arnej27959:

@havardpe please review


eval/src/vespa/eval/eval/value_codec.h
```diff
@@ -0,0 +1,33 @@
...
+#include <vespa/vespalib/stllike/string.h>
+
+namespace vespalib { class nbostream; }
+
+namespace vespalib::eval {
+
+/**
+ * encode a value to binary format
+ **/
+void new_encode(const Value &value, nbostream &output);
```
havardpe, line 15: I suggest we call this encode_value. Also, we could mention that this requires new api compliance in the comment.

eval/src/vespa/eval/eval/value_codec.h
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * decode a value from binary format
+ **/
+std::unique_ptr<Value> new_decode(nbostream &input, const ValueBuilderFactory &factory);
```
havardpe, line 20: I suggest we call this decode_value

eval/src/vespa/eval/eval/codec.h
```diff
@@ -0,0 +1,111 @@
...
+    abort();
+}
+
+/**
+ * Meta information about how a type can be decomposed into mapped and
+ * indexed dimensions and also how large each block is. A block is a
+ * dense-subspace consisting of all indexed dimensions that is
+ * uniquely specified by the labels of all mapped dimensions.
+ **/
+struct TypeMeta {
```
havardpe, line 39: this is a vital part of the SimpleTensor implementation that was re-used for encoding purposes, I do not think this should be moved out of like this. I think we should keep this one in SimpleTensor as-is and add a similar DecodePlan to value_codec.cpp

eval/src/vespa/eval/eval/simple_value.cpp
```diff
@@ -45,6 +46,10 @@ struct CreateValueFromTensorSpec {
             }
             map[sparse_key][dense_key] = entry.second;
         }
+        // hack for passing some (invalid?) unit tests
```
havardpe, line 49: not really a hack since converting from a spec should use 0.0 as default value for all existing non-defined cells. good catch though.

eval/src/vespa/eval/eval/simple_value.cpp
```diff
@@ -45,6 +46,10 @@ struct CreateValueFromTensorSpec {
...
+        if (spec.cells().empty() && type.count_mapped_dimensions() == 0) {
+            map[SparseKey()][0] = 0;
```
havardpe, line 51: prefer using {} or similar instead of SparseKey()

eval/src/vespa/eval/eval/simple_value.cpp
```diff
@@ -45,6 +46,10 @@ struct CreateValueFromTensorSpec {
             }
             map[sparse_key][dense_key] = entry.second;
         }
+        // hack for passing some (invalid?) unit tests
+        if (spec.cells().empty() && type.count_mapped_dimensions() == 0) {
```
havardpe, line 50: prefer checking 'map' over spec

havardpe (commented):
I suggest duplicating the code instead (since we want to remove SimpleTensor alltogether) and put everything from codec.h/cpp into value_codec.cpp (and possibly mutate it to better fit the code)

eval/src/vespa/eval/eval/value_type.h
```diff
@@ -60,10 +60,12 @@ class ValueType
     bool is_tensor() const { return (_type == Type::TENSOR); }
     bool is_sparse() const;
     bool is_dense() const;
+    size_t count_indexed_dimensions() const;
```
havardpe, line 63: should add unit test for this

eval/src/vespa/eval/eval/value_type.h
```diff
@@ -60,10 +60,12 @@ class ValueType
...
     size_t count_mapped_dimensions() const;
     size_t dense_subspace_size() const;
     const std::vector<Dimension> &dimensions() const { return _dimensions; }
     std::vector<Dimension> nontrivial_indexed_dimensions() const;
+    std::vector<Dimension> indexed_dimensions() const;
```
havardpe, line 68: should add unit test for this

eval/src/vespa/eval/eval/value_codec.cpp
```diff
@@ -0,0 +1,304 @@
...
+                    }
+                }
+                spec.add(addr, cells[(subspace_size * subspace_id) + i]);
+            }
+        }
+        return spec;
+    }
+};
+
+} // namespace <unnamed>
```
havardpe, line 239: should also contain classes below

havardpe (commented):
some minor comments, otherwise looks very good

havardpe (approved):

PR #14509, opened 2020-09-23
Run some maintainers with higher interval

hmusum:
When deployments exist on all config servers each server will deploy,
so do it with 3 times higher interval

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java
@@ -9,7 +9,9 @@ import com.yahoo.config.provision.HostLivenessTracker;
 import com.yahoo.config.provision.InfraDeployer;
 import com.yahoo.config.provision.Zone;
 import com.yahoo.jdisc.Metric;
+import com.yahoo.vespa.flags.BooleanFlag;
 import com.yahoo.vespa.flags.FlagSource;
+import com.yahoo.vespa.flags.Flags;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
 import com.yahoo.vespa.hosted.provision.autoscale.NodeMetrics;
 import com.yahoo.vespa.hosted.provision.autoscale.NodeMetricsDb;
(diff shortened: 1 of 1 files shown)


hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java
```diff
@@ -178,14 +180,14 @@ public void deconstruct() {
             nodeMetricsCollectionInterval = Duration.ofMinutes(1);
             operatorChangeRedeployInterval = Duration.ofMinutes(1);
             osUpgradeActivatorInterval = zone.system().isCd() ? Duration.ofSeconds(30) : Duration.ofMinutes(5);
-            periodicRedeployInterval = Duration.ofMinutes(30);
+            periodicRedeployInterval = deploymentExistsOnAllConfigServers ? Duration.ofMinutes(90) : Duration.ofMinutes(30);
```
freva, line 183: `PeriodicApplicationMaintainer` runs every minute (`redeployMaintainerInterval`), this is the `minTimeBetweenRedeployments`, so I don't think we should change this.

freva (commented):
In addition, I suggest scaling 3x `operatorChangeRedeployInterval`, `retiredInterval`, `autoscalingInterval` (going by usage of `Deployer`).

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java
hmusum: Ouch, yes, thanks

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeRepositoryMaintenance.java
```diff
@@ -180,12 +181,12 @@ public void deconstruct() {
...
             periodicRedeployInterval = Duration.ofMinutes(30);
             provisionedExpiry = Duration.ofHours(4);
-            rebalancerInterval = Duration.ofMinutes(40);
+            rebalancerInterval = deploymentExistsOnAllConfigServers ? Duration.ofMinutes(120) : Duration.ofMinutes(40);
             redeployMaintainerInterval = Duration.ofMinutes(1);
             // Need to be long enough for deployment to be finished for all config model versions
             // Should be equal to timeout for deployments
             reservationExpiry = zone.system().isCd() ? Duration.ofMinutes(5) : Duration.ofMinutes(30);
-            scalingSuggestionsInterval = Duration.ofMinutes(31);
+            scalingSuggestionsInterval = deploymentExistsOnAllConfigServers ? Duration.ofMinutes(91) : Duration.ofMinutes(31);
```
freva, line 189: This can also be reduced back now, it does not deploy (assume it was increased because of periodic interval change?)

hmusum: Yes, thanks, fixed

freva (approved):
LGTM, not sure if you saw my suggestion for other maintainers that could be have their interval increased.

hmusum:
No, didn't see it, will have a look

freva (approved):
:100:

PR #14510, opened 2020-09-23
Reorder equality checks so that common case is cheaper

vekterli:
@geirst please review.

Node index and type are much more likely to differ than cluster.


geirst (approved):
👍

PR #14512, opened 2020-09-23
Ignore force commit with serial number not greater than attribute vector create serial number.

toregge:
@baldersheim : please review
@geirst : FYI

### DIFF
--- a/searchcore/src/tests/proton/attribute/attribute_test.cpp
+++ b/searchcore/src/tests/proton/attribute/attribute_test.cpp
@@ -192,6 +192,11 @@ public:
     void assertExecuteHistory(std::vector<uint32_t> expExecuteHistory) {
         EXPECT_EQ(expExecuteHistory, _attributeFieldWriter->getExecuteHistory());
     }
+    SerialNum test_force_commit(AttributeVector &attr, SerialNum serialNum) {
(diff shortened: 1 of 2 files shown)


baldersheim (approved):

searchcore/src/tests/proton/attribute/attribute_test.cpp
```diff
@@ -975,6 +980,15 @@ TEST_F(AttributeWriterTest, forceCommit_clears_search_cache_in_imported_attribut
     EXPECT_EQ(0u, _mgr->getImportedAttributes()->get("imported_b")->getSearchCache()->size());
 }
 
+TEST_F(AttributeWriterTest, ignores_force_commit_serial_not_greater_than_create_serial)
+{
+    auto a1 = addAttribute("a1");
+    a1->setCreateSerialNum(100);
+    EXPECT_EQ(0u, test_force_commit(*a1, 50u));
+    EXPECT_EQ(0u, test_force_commit(*a1, 100u));
+    EXPECT_EQ(150u, test_force_commit(*a1, 150u));
```
geirst, line 989: In this case I would test with 101, as that is the first value with changed behaviour.

geirst (commented):
👍

PR #14517, opened 2020-09-23
Since only a simple single decimal floatingpoint number is needed avo…

baldersheim:
…id using string and floating point formatting. Use a simpler variant you get a 15 X speedup of this frequently called code.

@bratseth PR
@hmusum FYI


bratseth (approved):
Frequently called? That doesn't sound right.

PR #14518, opened 2020-09-24
instruction-level generic join

havardpe:
also move some stuff out of simple_value

@arnej27959 please review

### DIFF
--- a/eval/src/vespa/eval/eval/tensor_instructions.h
+++ b/eval/src/vespa/eval/eval/tensor_instructions.h
@@ -1,0 +1,24 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include "interpreted_function.h"
+
+namespace vespalib {
+
+class Stash;
+
+namespace eval {
+
+struct JoinPlan;
+
+namespace tensor_instruction {
+
+using join_fun_t = double (*)(double, double);
+
+InterpretedFunction::Instruction make_join(const ValueType &lhs_type, const ValueType &rhs_type, join_fun_t function,
+                                           const ValueBuilderFactory &factory, Stash &stash);
+
+}
+}
+}
--- a/eval/src/vespa/eval/eval/tensor_plans.h
+++ b/eval/src/vespa/eval/eval/tensor_plans.h
@@ -1,0 +1,77 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include "value_type.h"
+#include <vespa/vespalib/stllike/string.h>
+#include <vector>
+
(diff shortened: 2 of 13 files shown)


eval/src/vespa/eval/eval/tensor_instructions.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include "interpreted_function.h"
+
+namespace vespalib {
+
+class Stash;
+
+namespace eval {
+
+struct JoinPlan;
```
arnej27959, line 13: There is no "JoinPlan".

eval/src/vespa/eval/eval/tensor_instructions.h
```diff
@@ -0,0 +1,24 @@
...
+
+namespace tensor_instruction {
+
+using join_fun_t = double (*)(double, double);
+
+InterpretedFunction::Instruction make_join(const ValueType &lhs_type, const ValueType &rhs_type, join_fun_t function,
+                                           const ValueBuilderFactory &factory, Stash &stash);
+
+}
+}
```
arnej27959, line 23: ending three nested namespaces like this feels confusing to me.
I suggest something like:

```
namespace vespalib { class Stash; }

namespace vespalib::eval::tensor_instruction {
... rest of file ...
}
```

eval/src/vespa/eval/eval/tensor_plans.h
```diff
@@ -0,0 +1,77 @@
...
+struct DenseJoinPlan {
+    size_t lhs_size;
+    size_t rhs_size;
+    size_t out_size;
+    std::vector<size_t> loop_cnt;
+    std::vector<size_t> lhs_stride;
+    std::vector<size_t> rhs_stride;
+    DenseJoinPlan(const ValueType &lhs_type, const ValueType &rhs_type);
+    ~DenseJoinPlan();
+    template <typename F> void execute(size_t lhs, size_t rhs, F &&f) const {
```
arnej27959, line 30: is it possible to move the various execute() implementations to the .cpp file? or would that lead to less inlining opportunities?

arnej27959 (approved):
👍 only cosmetic comments 👍

eval/src/vespa/eval/eval/tensor_plans.h
havardpe: I think we want it in a separate header, for cross-use in different plans. We do not want to call the function via pointer, so I think we should avoid having it in a cpp file.

eval/src/vespa/eval/eval/tensor_instructions.h
havardpe: it got renamed to JoinParam and made local, I will clean it up

eval/src/vespa/eval/eval/tensor_instructions.h
havardpe: agree

PR #14519, opened 2020-09-24
Update javadoc for DocumentAccess and add convencience for outside co…

jonmv:
…ntainers

@bratseth please review.


bratseth (approved):

jonmv:
Urgh abi spec.

PR #14522, opened 2020-09-24
Store vespa version in ZooKeeper as well

hmusum:
Use version in ZooKeeper if configserver-distribute-application-package is true,
fallback to version in file. Will always write to ZooKeeper so this can be
the only source in the future.

This means that if configserver-distribute-application-package is true
and version data is found in ZooKeeper, redeployment of appplication packages
will only be done on one config server.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
@@ -5,45 +5,78 @@ import com.google.inject.Inject;
 import com.yahoo.cloud.config.ConfigserverConfig;
 import com.yahoo.component.Version;
 import com.yahoo.io.IOUtils;
+import com.yahoo.path.Path;
+import com.yahoo.text.Utf8;
+import com.yahoo.vespa.curator.Curator;
 import com.yahoo.vespa.defaults.Defaults;
+import com.yahoo.vespa.flags.BooleanFlag;
+import com.yahoo.vespa.flags.FlagSource;
+import com.yahoo.vespa.flags.Flags;
 
 import java.io.File;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.Optional;
 
 /**
- * Contains version information for this configserver.
+ *
+ * Contains version information for this configserver. Stored both in file system and in ZooKeeper (uses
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
```diff
@@ -5,45 +5,78 @@
...
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
     }
 
     public Version storedVersion() {
+        if (distributeApplicationPackage.value()) {
+            Optional<byte[]> version = curator.getData(versionPath);
+            if(version.isPresent()) {
+                System.out.println("Found version in zk ");
```
hakonhall, line 70: Remove or log

configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
```diff
@@ -5,45 +5,78 @@
...
+                try {
+                    return Version.fromString(Utf8.toString(version.get()));
+                } catch (Exception e) {
+                    return new Version(0, 0, 0); // Use an old value to signal we don't know
```
hakonhall, line 74: Consider falling back to the file.

configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
```diff
@@ -5,45 +5,78 @@
...
+                }
+            }
+        }
         try (FileReader reader = new FileReader(versionFile)) {
+            System.out.println("Found version in file ");
```
hakonhall, line 79: Remove or log

configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
hmusum: Good point, fixed

configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
hmusum: Fixed

configserver/src/main/java/com/yahoo/vespa/config/server/version/VersionState.java
hmusum: Fixed

hakonhall (approved):

PR #14523, opened 2020-09-24
Cache node objects

mpolden:
I suggest we do this first and consider handling it in `CuratorDatabase` if we
can think of a safe way to update that cache.

Reduces test run time in this module by 5 seconds on my machine.

@bratseth or @baldersheim

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
@@ -1,7 +1,11 @@
 // Copyright 2018 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.persistence;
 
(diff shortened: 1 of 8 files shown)


bratseth (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
```diff
@@ -107,6 +113,12 @@
     // Network port fields
     private static final String networkPortsKey = "networkPorts";
 
+    // A cache of deserialized Node objects. The cache is keyed on the hash of serialized node data.
+    //
+    // Deserializing a Node from slime is expensive, and happens frequently. Node instances that have already been
+    // deserialized are returned from this cache instead of being deserialized again.
+    private final Cache<Long, Node> cache = CacheBuilder.newBuilder().maximumSize(1000).build();
```
baldersheim, line 120: I guess the size should be configurable.

baldersheim (approved):
Apart from configurable max size this looks good.

mpolden:
Java build passed.

PR #14524, opened 2020-09-24
Add StatBucket storage protocol (de-)serialization support

vekterli:
@geirst please review

Adds both protobuf schema and type ID handling to protocol codec.
Old protocol versions are not expected to handle this message, as
StatBucket will run over the Document protocol instead of the
Storage protocol in such cases.


geirst (approved):
👍

PR #14529, opened 2020-09-24
Internal restart

freva:
The last commit is not needed, and I can remove if you don't like it. I wanted to create a test for this and needed a way to create `DeployTester` with `Provisioner` instead of `HostProvisioner`, and there was no easy way without also creating a bunch of other stuff. In the end, I couldn't get any test to work anyway... :/


hmusum (approved):
Looks good to me! 

DeployTester needed a builder anyway, it had way too many constructors, so that's good

PR #14544, opened 2020-09-24
Jonmv/more doc api parameters

jonmv:
@bratseth please review. 

I realised after doing all this that the interface `MessageBusSession` lets you set route and trace level, and the `MessageBusAsyncSession` obviously implements this. If the `DocumentAccess` is supposed to be the "starting point of the document API", however, then this is not reachable. You end up having to use the concrete implementation. Setting route and trace level on the whole session is also less flexible, and requires in practice a separate session whenever any of these are to be set — or a session per route/tracelevel. 
To make fieldSet, route and traceLevel available to the users of the document API, I think they should be parameters to each operation, and reflected in the `Response`. This PR does that, but it's quite a bit of code to achieve this, cluttering APIs somewhat — in particular the `Response` hierarchy. 

Soooo ... PTAL :)


documentapi/src/main/java/com/yahoo/documentapi/AsyncSession.java
```diff
@@ -158,6 +194,23 @@ default Result get(DocumentId id, DocumentProtocol.Priority priority) {
...
      */
     default Result remove(DocumentId id, DocumentProtocol.Priority priority) {
+        return remove(id, parameters().withPriority(priority));
+    }
+
+    /**
+     * <p>Removes a document if it is present. This method returns immediately.</p>
+     *
+     * <p>If this result is a success, this
+     * call will cause one or more {@link DocumentIdResponse} objects to apprear within the timeout time of this session.
```
bratseth, line 204:
```suggestion
     * call will cause one or more {@link DocumentIdResponse} objects to appear within the timeout time of this session.
```

documentapi/src/main/java/com/yahoo/documentapi/DocumentOperationParameters.java
```diff
@@ -0,0 +1,66 @@
...
+import com.yahoo.document.fieldset.FieldSet;
+import com.yahoo.document.fieldset.FieldSetRepo;
+import com.yahoo.documentapi.messagebus.protocol.DocumentProtocol;
+
+import java.util.Optional;
+import java.util.OptionalInt;
+
+import static java.util.Objects.requireNonNull;
+
+/** Optional parameters for a document operation. Immutable class. */
```
bratseth, line 12:
```suggestion
/** 
 * Optional parameters for a document operation. Immutable class.
 *
 * @author jonmv 
 */
```

documentapi/src/main/java/com/yahoo/documentapi/DocumentOperationParameters.java
```diff
@@ -0,0 +1,66 @@
+package com.yahoo.documentapi;
```
bratseth, line 1:
```suggestion
// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
package com.yahoo.documentapi;
```

bratseth (approved):
Great!

jonmv:
Thanks!

PR #14550, opened 2020-09-25
Jonmv/async doc v1

jonmv:
@bjorncs or @baldersheim please review.

PR #14554, opened 2020-09-25
Jonmv/async doc v1 implementation

jonmv:
@bjorncs please review. Will add tests as they're ready. Note that first 5 commits are already in #14550

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
@@ -37,7 +37,14 @@ import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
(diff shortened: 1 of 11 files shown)


jonmv:
Hmm, last commit wasn't good. Should use `Condition` for this.

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
```diff
@@ -0,0 +1,598 @@
...
+                    // Ready for retry — remove from queue and send.
+                    if (operation.readyAt().isBefore(clock.instant())) {
+                        action.accept(operation);
+                        operations.remove();
+                        continue;
+                    }
+                    // Not yet ready for retry — track earliest time to wake up again.
+                    waitUntilMillis = Math.min(waitUntilMillis, operation.readyAt.toEpochMilli());
+                }
+                synchronized (queue) {
```
bjorncs, line 264: Do not use a concurrent collection as an external monitor (unless `ConcurrentLinkedQueue` says it's okay).

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -0,0 +1,603 @@
...
+        this.executor = executor;
+        this.parser = parser;
+        this.metric = metric;
+        this.metrics = new DocumentApiMetrics(metricReceiver, "documentV1");
+        this.handlers = defineApi();
+    }
+
+    @Override
+    public ContentChannel handleRequest(Request rawRequest, ResponseHandler rawResponseHandler) {
+        metric.add("handled.requests", 1, HandlerMetricContextUtil.contextFor(rawRequest, metric, getClass()));
```
bjorncs, line 145: Consider moving the metric aggregation to `HandlerMetricContextUtil` and replace this with a e.g `HandlerMetricContextUtil.onHandledRequest(...)`.

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
jonmv: Good point. Already changed to allow tests to notify maintainers, though :)

jonmv:
I think we're almost ready to roll. May as well have this merged. Only need some tests for the REST API handler now!

PR #14556, opened 2020-09-25
Internal restart on internal deployments

freva:


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -395,9 +382,10 @@ public PrepareResult deploy(File applicationPackage, PrepareParams prepareParams
         SessionRepository sessionRepository = tenant.getSessionRepository();
         LocalSession newSession = sessionRepository.createSessionFromExisting(activeSession, logger, true, timeoutBudget);
         sessionRepository.addLocalSession(newSession);
+        boolean internalRestart = deployWithInternalRestart.with(FetchVector.Dimension.APPLICATION_ID, application.serializedForm()).value();
```
freva, line 385: Without `ZONE_ID` here since the flag will be set on the configserver for internal deployments. For external deployments the flag still must be set on the controller. That could be changed, but would need to re-create `PrepareParams`. 

I think we can remove `internalRestart` parameter and this flag in a week or so anyway and just always do internal restarts if the `provisioner` is set?

hmusum (approved):
Great!

PR #14557, opened 2020-09-25
Don't call commit() on visibility handler during replay.

toregge:
@baldersheim : please review
@geirst : FYI


baldersheim (approved):
Perhaps a small comment as to why.

baldersheim (approved):

PR #14558, opened 2020-09-25
Add feature flag to enable new restapi handler

bjorncs:
FYI @baldersheim

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/clients/ContainerDocumentApi.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/clients/ContainerDocumentApi.java
@@ -41,13 +41,17 @@ public class ContainerDocumentApi {
 
 
     private static void addRestApiHandler(ContainerCluster<?> cluster, Options options) {
-        var handler = newVespaClientHandler(
-                "com.yahoo.document.restapi.resource.RestApi", "/document/v1/*", options);
+        String handlerClassName = options.useNewRestapiHandler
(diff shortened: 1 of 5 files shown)


config-model/src/main/java/com/yahoo/vespa/model/clients/ContainerDocumentApi.java
```diff
@@ -41,13 +41,17 @@ private static void addFeedHandler(ContainerCluster<?> cluster, Options options)
...
+        var handler = newVespaClientHandler(handlerClassName, "/document/v1/*", options);
         cluster.addComponent(handler);
-        var executor = new Threadpool(
-                "restapi-handler", cluster, options.restApiThreadpoolOptions, options.feedThreadPoolSizeFactor);
-        handler.inject(executor);
-        handler.addComponent(executor);
+        if (!options.useNewRestapiHandler) {
+            var executor = new Threadpool(
+                    "restapi-handler", cluster, options.restApiThreadpoolOptions, options.feedThreadPoolSizeFactor);
+            handler.inject(executor);
```
jonmv, line 52: Explicit injection 😱

jonmv (approved):
No idea

config-model/src/main/java/com/yahoo/vespa/model/clients/ContainerDocumentApi.java
bjorncs: This is how the cool kids on the street do it

PR #14561, opened 2020-09-25
Automatically fallback to MessageBus if direct storage RPC not supported

vekterli:
@geirst please review

To avoid need to track this per peer, currently falls back to MBus
for _all_ peers if any of them indicates it's on an old version that
does not understand the new RPC method.

Also fix test teardown race condition where RPC threads could still
be running and invoking callbacks when the API RPC service had already
been destroyed.


geirst (approved):
👍

geirst:
@baldersheim FYI

PR #14566, opened 2020-09-25
Run NodeFailer with larger interval

hmusum:
If all config servers have application package run NodeFailer with
3 times larger interval.


bratseth:
We should do this with all maintainers doing deployments ...

freva (approved):

freva:
@bratseth Others done in #14509

bratseth:
> @bratseth Others done in #14509

👍

PR #14574, opened 2020-09-26
Avoid clang warnings.

toregge:
@baldersheim : please review


eval/src/vespa/eval/tensor/mixed/packed_mappings_builder.cpp
```diff
@@ -54,8 +54,8 @@ PackedMappingsBuilder::target_memory(char *mem_start, char *mem_end) const
     ssize_t avail_sz = mem_end - mem_start;
     assert(needs_sz <= avail_sz);
 
-    uint32_t * int_store_mem = (uint32_t *) mem_start;
-    uint32_t * offsets_mem = (uint32_t *) (mem_start + int_store_size);
+    uint32_t * int_store_mem = (uint32_t *) (void *) mem_start;
+    uint32_t * offsets_mem = (uint32_t *) (void *) (mem_start + int_store_size);
```
baldersheim, line 58: Is it not better that mem_start is void ?

eval/src/vespa/eval/tensor/mixed/packed_mixed_tensor_builder.cpp
```diff
@@ -39,7 +39,7 @@ PackedMixedTensorBuilder<T>::build(std::unique_ptr<ValueBuilder<T>>)
 
     // copy cells:
     memcpy(cells_mem, &_cells[0], cells_size);
-    ConstArrayRef<T> cells((T *)cells_mem, _cells.size());
+    ConstArrayRef<T> cells((T *)(void *) cells_mem, _cells.size());
```
baldersheim, line 42: Is it not better that mem_start is void * ?

baldersheim (approved):

eval/src/vespa/eval/tensor/mixed/packed_mappings_builder.cpp
toregge: If `mem_start` is `void*` then any code using `mem_start + offset` would need to use `((char *) mem_start) + offset`

storage/src/vespa/storage/storageserver/rpc/storage_api_rpc_service.cpp
```diff
@@ -117,7 +117,8 @@ void compress_and_add_payload_to_rpc_params(mbus::BlobRef payload,
 
     params.AddInt8(comp_type);
     params.AddInt32(static_cast<uint32_t>(to_compress.size()));
-    params.AddData(buf.stealBuffer(), buf.getDataLen());
+    auto buffer_len = buf.getDataLen();
+    params.AddData(buf.stealBuffer(), buffer_len);
```
baldersheim, line 121: This is not the first time this treacherous interface gives us trouble.
Fixed in #14575

PR #14585, opened 2020-09-28
Gjoranv/preinstall bundles cleanup

gjoranv:
FYI: @bjorncs


hmusum (approved):
I don't know which version was the last one with preinstall, but LGTM

gjoranv:
It was removed on July 17, so I think 7.254 was the last version to use preinstall.

hmusum:
👍

PR #14599, opened 2020-09-28
Add application and cluster to log message

mpolden:
@hakonhall or @hmusum

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
@@ -131,7 +131,8 @@ public class NodePrioritizer {
                 allocation = host.ipConfig().pool().findAllocation(allNodes, nodeRepository.nameResolver());
                 if (allocation.isEmpty()) continue; // No free addresses in this pool
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


hmusum (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -131,7 +131,8 @@ private void addNewDockerNodesOn(LockedNodeList candidates) {
                 allocation = host.ipConfig().pool().findAllocation(allNodes, nodeRepository.nameResolver());
                 if (allocation.isEmpty()) continue; // No free addresses in this pool
             } catch (Exception e) {
-                log.log(Level.WARNING, "Failed allocating IP address on " + host.hostname(), e);
+                log.log(Level.WARNING, "Failed allocating IP address on " + host.hostname() + " to " +
+                                       application + ", cluster " + clusterSpec.id(), e);
```
hakonhall, line 135: could we cut the stack trace?

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
mpolden: Done.

hakonhall (approved):

PR #14619, opened 2020-09-29
Delete session without lock

hmusum:
Two things has changed:
* Do not use lock when deleting, it's not necessary, nobody else takes the lock
* Delete remote session data in zookeeper immediately, watchers on other config servers will be notified and internal state cleaned up (also local session state)

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -478,7 +478,7 @@ public class ApplicationRepository implements com.yahoo.config.provision.Deploye
             Optional<Long> activeSession = tenantApplications.activeSessionOf(applicationId);
             if (activeSession.isEmpty()) return false;
 
-            // Deleting an application is done by deleting the remote session, other config
+            // Deleting an application is done by deleting the remote session, all config
             // servers will pick this up and clean up through the watcher in this class
             try {
                 RemoteSession remoteSession = getRemoteSession(tenant, activeSession.get());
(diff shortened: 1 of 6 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -469,47 +469,21 @@ static void checkIfActiveIsNewerThanSessionToBeActivated(long sessionId, long cu
...
             Optional<Long> activeSession = tenantApplications.activeSessionOf(applicationId);
             if (activeSession.isEmpty()) return false;
 
-            // Deleting an application is done by deleting the remote session and waiting
-            // until the config server where the deployment happened picks it up and deletes
-            // the local session
-            long sessionId = activeSession.get();
-
-            RemoteSession remoteSession;
+            // Deleting an application is done by deleting the remote session, other config
```
hakonhall, line 481:
```suggestion
            // Deleting an application is done by deleting the remote session, all config
```

This config server will also delete through the watcher?

hmusum: Yes

hakonhall: I now think the original comment is correct - delete(RemoteSession) only called in this config server appears to do delete and deactivate the remote session, the same as sessionRemoved(long) which seems to be invoked from watcher.

hakonhall (approved):

hakonhall (approved):

PR #14621, opened 2020-09-29
Check if we can deploy now before putting on queue

hmusum:
Avoids metric pendingDeployments showing deployments that will not be done
because they have been deployed recently


bratseth (commented):
Shouldn't we keep both checks?

hmusum:
I see that `canDeployNow()` is called right before `deploy()`(in `applicationsNeedingMaintenance()`), so this is not necessary, and, yes, we should keep the second one. But we need to make sure to remove from `pendingDeployments`, so see https://github.com/vespa-engine/vespa/pull/14625

PR #14623, opened 2020-09-29
Bratseth/defer ip address allocation

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
```diff
@@ -159,76 +186,174 @@ private double skewWith(NodeResources resources) {
...
-            return this;
+        /** Called when the node described by this candidate must be created */
+        public NodeCandidate withNode() {
+            Optional<IP.Allocation> allocation;
+            try {
+                allocation = parent.get().ipConfig().pool().findAllocation(allNodes, nodeRepository.nameResolver());
+                if (allocation.isEmpty())
+                    throw new IllegalStateException("No free ip addresses on " + parent.get() + ": Cannot allocate node");
+            } catch (Exception e) {
+                throw new IllegalStateException("Failed allocating IP address on " + parent.get() +": ", e);
```
freva, line 327: The old code would just advance to the next node if resolving failing. If we get a node with broken IP, the allocation will basically not succeed until manual intervention since it would keep trying the same node? We should catch this and try next node if this fails.

Also, lets drop the stack trace for this, see #14599

bratseth: Done, please ptal. I think it would be better if we could expect this not to fail though - perhaps we should have a maintainer looking for it.

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
mpolden: It's the product of countless cases of DNS/IP misconfiguration upstream from us. It will continue to happen. 😞

bratseth: I would like to stop using DNS but not sure if that would avoid all of that?

mpolden: IP assignment and DNS issues typically correlate, at least in our internal infrastructure: If foo.example.com resolves to 1.1.1.1, but 1.1.1.1 resolves to bar.example.com, there's likely an IP conflict where two distinct hosts have been assigned the same IP.

Vespa itself also has strict hostname validation on startup, where hostname and IP is resolved and compared. That's another reason we do this check during allocation as the node would refuse to start in case of misconfiguration anyway.

PR #14627, opened 2020-09-30
Avoid reading nodes from irrelevant states 

mpolden:
@freva


freva (approved):
Great! :1st_place_medal:

mpolden:
.

PR #14630, opened 2020-09-30
Use BigDecimal for budget inside the Quota

oyving:
- To allow budgets below $1/hour we change the internal representation of budget
  to a decimal number.  Some interfaces that assume integers are kept to keep the
  API stable.

- Created a nicer method .unlimited() instead of .empt() to better show semantics.

- Added some serialisation tests to make sure we support integers and decimals.


config-model-api/src/main/java/com/yahoo/config/model/api/Quota.java
```diff
@@ -14,42 +16,57 @@
...
 
     public static Quota fromSlime(Inspector inspector) {
         var clusterSize = SlimeUtils.optionalLong(inspector.field("clusterSize"));
-        var budget = SlimeUtils.optionalLong(inspector.field("budget"));
-        return new Quota(clusterSize.map(Long::intValue), budget.map(Long::intValue));
+        var budget = budgetFromSlime(inspector.field("budget"));
+        return new Quota(clusterSize.map(Long::intValue), budget, true);
+    }
+
+    public Quota withBudget(Optional<BigDecimal> budget) {
```
mpolden, line 43: Consider `BigDecimal` parameter instead of `Optional`. That way you avoid the initial problem where signatures with generic types cannot be overloaded. Same for the other method.

config-model-api/src/test/java/com/yahoo/config/model/api/QuotaTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.yahoo.slime.SlimeUtils;
+import org.junit.Test;
+
+
+import java.math.BigDecimal;
+
+import static org.junit.Assert.assertEquals;
+
+public class QuotaTest {
```
mpolden, line 12: Missing `@author`.

mpolden (approved):

mpolden (approved):

PR #14633, opened 2020-09-30
 Comment where payload is checkted in SignatureFilter

oyving:


jonmv (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/filter/SignatureFilter.java
```diff
@@ -65,6 +65,11 @@ public SignatureFilter(Controller controller) {
     }
 
     private boolean keyVerifies(PublicKey key, DiscFilterRequest request) {
+        /* This method only checks that the content hash has been signed by the provided public key, but
+         * does not verify the content of the request.  jDisc request filters do not allow inspecting the
+         * request body, so this responsibility falls on the handler consuming the body instead.  For this
+         * specific case the request body is validated in {@link ApplicationApiHandler.parseDataParts}.
```
jonmv, line 71:
```suggestion
         * request body, so this responsibility falls on the handler consuming the body instead.  For the
         * deployment cases, the request body is validated in {@link ApplicationApiHandler.parseDataParts}.
```

jonmv (approved):

PR #14638, opened 2020-09-30
Revert "Delete session without lock"

hmusum:
Reverts vespa-engine/vespa#14619

Changing this code has historically not worked all times, so this is in case something breaks. Please review only


hmusum:
@aressem FYI

hakonhall (approved):

PR #14639, opened 2020-09-30
Templated sparse tensor 3

arnej27959:

@havardpe please review
@toregge @geirst FYI


eval/src/vespa/eval/tensor/sparse/sparse_tensor_join.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "sparse_tensor_address_combiner.h"
+#include "direct_sparse_tensor_builder.h"
+
+namespace vespalib::tensor::sparse {
+
+template <typename LCT, typename RCT, typename Function>
+std::unique_ptr<Tensor>
+join(const SparseTensor &lhs_in, const SparseTensor &rhs_in, Function &&func)
+{
+    using OCT = typename eval::UnifyCellTypes<LCT,RCT>::type;
```
havardpe, line 16: if a sparse tensor can be of type double, this might not be correct in all cases (joining double with tensor)

eval/src/vespa/eval/tensor/sparse/sparse_tensor_index.cpp
```diff
@@ -220,38 +229,58 @@ SparseTensorValueIndex::create_view(const std::vector<size_t> &dims) const
...
-                                        const SparseTensorValueIndex &index_in,
-                                        const std::vector<T> &cells_in)
-    : _type(type_in),
-      _index(index_in),
-      _cells(cells_in)
+    
+size_t
+SparseTensorIndex::lookup_or_add(SparseTensorAddressRef tmp_ref)
 {
+#if 0
```
havardpe, line 243: prefer this form, add comment and experiment with structured bindings

arnej27959:
made a new PR with all these changes squished together.

eval/src/vespa/eval/tensor/sparse/sparse_tensor_t.cpp
```diff
@@ -0,0 +1,251 @@
...
+    if (function == eval::operation::Mul::f) {
+        if (lhs_type.dimensions() == rhs_type.dimensions()) {
+            return typify_invoke<1,eval::TypifyCellType,FastSparseJoin<T>>(
+                    rhs_type.cell_type(),
+                    *this, *rhs, std::move(res_type));
+        }
+    }
+    return typify_invoke<2,eval::TypifyCellType,GenericSparseJoin<T>>(
+            rhs_type.cell_type(), res_type.cell_type(),
+            *this, *rhs, std::move(res_type), function);
```
toregge, line 197: While this likely works due to `res_type._cell_type` not being cleared due to the move constructor being called, it would likely be better to adjust the code to ensure that `res_type.cell_type()` is evaluated first.

havardpe: evaluating std::move(value) will not change the value, since it 'only' casts the value to an rvalue. I thought that all parameters were evaluated (in undefined order) before the function was actually called, and that the copying of parameters into the local scope of the function was part of evaluating the function itself. I might be wrong, but changing the code will make it clearer and safer any way, so I am good with that. Maybe @vekterli can answer what the standard says about this?

vekterli: Had a quick rummage through the C++20 ISO draft.

**§ 6.9.1.10**

> Except where noted, evaluations of operands of individual operators and of subexpressions of individual expressions are unsequenced. (...) If a side effect on a memory location (6.7.1) is unsequenced relative to either another side effect on the same memory location or a value computation using the value of any object in the same memory location, and they are not potentially concurrent (6.9.2), the behavior is undefined.

**§ 6.9.1.11**
> When calling a function (whether or not the function is inline), every value computation and side effect
associated with any argument expression, or with the postfix expression designating the called function, is
sequenced before execution of every expression or statement in the body of the called function.

**§ 7.6.1.2.7** (emphasis mine)

> The initialization and destruction of each parameter occurs within the context of the _calling_ function. 

Unless there are some more sneaky sequencing constraints I'm not aware of, this implies to me that the object move-construction from the rvalue (and subsequent invalidation of the object pointed to by the rvalue) _may_ take place before it's used as part of another parameter.

PR #14641, opened 2020-09-30
Close connection early if no inflight operations MERGEOK

bjorncs:
FYI @baldersheim


vespa-http-client/src/main/java/com/yahoo/vespa/http/client/core/communication/ApacheGatewayConnection.java
```diff
@@ -427,7 +427,6 @@ public CloseableHttpClient createClient() {
             clientBuilder.setMaxConnTotal(1);
             clientBuilder.setUserAgent(String.format("vespa-http-client (%s)", Vtag.V_TAG_COMPONENT));
             clientBuilder.setDefaultHeaders(Collections.singletonList(new BasicHeader(Headers.CLIENT_VERSION, Vtag.V_TAG_COMPONENT)));
-            clientBuilder.disableContentCompression();
```
baldersheim, line 430: This sounds like a separate PR.
Have you checked the history on this one ?

baldersheim (approved):
Looks, good to me. But I would probably take the compression stuff in a separate PR

PR #14644, opened 2020-09-30
Handle ReferenceFieldValue in document selection expression.

toregge:
@vekterli : please review


vekterli (approved):
:+1: an added test case would be nice

baldersheim (approved):

PR #14647, opened 2020-09-30
Record locks taken for external deploys

hakonhall:
- Information about a lock attempt now includes a list of lock attempts done
   while holding the lock, forming a tree (forest) structure.
 - Records the duration and locking attempts done as part of an external
   deploy, forming a tree of locks with timing info.  The currently active
   external deploys are shown in an "ongoing-recording" field of
   /nodes/v2/locks.
 - The 3 longest external deploys are kept in "recordings" in /nodes/v2/locks.
 - Extracts the global process-wide parts of ThreadLockStats into separate
   class for clarity.


bratseth (approved):

mpolden (approved):
🌲

hakonhall:
Also showing nested locks may drastically increase the size of the response, e.g. internal redeployment takes one maintainer lock and would include deployment locks for many apps.  I'll check if this is a problem or not.

hakonhall:
Looks good

PR #14650, opened 2020-09-30
Add API for rerank-count

bratseth:


lesters (approved):
LGTM! Build fails on ABI checker...

PR #14656, opened 2020-10-01
Arnej/cleanup value api 1

arnej27959:

@havardpe please review


havardpe (commented):
looks good, but the PR says that there are conflicts. Also, we probably want to use array refs for 'create_view' and 'add_subspace' as well.

havardpe (approved):

PR #14660, opened 2020-10-01
Add StorageAPI RPC trace events

vekterli:
@geirst please review

Trace event set should roughly match that of MessageBus, but contains
more information (such as hostnames and payload size) for debugging
purposes.


geirst (approved):
👍

PR #14663, opened 2020-10-01
Clock.systemUTC() instead of non-existent injected Clock

jonmv:
@bjorncs please review :)


freva (approved):

jonmv:
What happen?

Someone set up us the bomb ...

PR #14669, opened 2020-10-01
Catch Throwable instead of just RuntimeException

jonmv:
@bjorncs please review.


freva (approved):

bratseth:
I think you should only catch LinkageError+RuntimeException and let the rest pass through.

jonmv:
Yes, this is only for debugging.

PR #14670, opened 2020-10-01
Let handlers define a default request type

bratseth:


container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
```diff
@@ -106,6 +106,18 @@ public final ContentChannel handleRequest(Request request, ResponseHandler respo
...
+     * <p>Returns the request type classification to use for requests to this handler.
+     * This overrides the default classification based on request method, and can in turn
+     * be overridden by setting a request type on individual responses in handleRequest
+     * whenever it is invoked (i.e not for requests that are rejected early e.g due to overload).</p>
+     *
+     * <p>This default implementation returns null.</p>
+     *
+     * @return the request type to set, or null to not override the default classification based on request method
+     */
+    protected Request.RequestType getRequestType() { return null; }
```
bjorncs, line 119: Consider using `Optional`.

bjorncs (approved):

bjorncs:
```

[INFO] --- abi-check-plugin:7-SNAPSHOT:abicheck (default) @ container-core ---
[ERROR] Class com.yahoo.container.jdisc.ThreadedRequestHandler: Extra method protected com.yahoo.jdisc.Request$RequestType getRequestType()
[INFO] ------------------------------------------------------------------------
```

container-core/src/main/java/com/yahoo/container/jdisc/ThreadedRequestHandler.java
bratseth: Good suggestion, thanks

PR #14671, opened 2020-10-01
Revert "Catch Throwable instead of just RuntimeException"

jonmv:
Reverts vespa-engine/vespa#14669

@bratseth please review.


jonmv:
https://github.com/vespa-engine/vespa/pull/14672 instead

PR #14672, opened 2020-10-01
Jonmv/catch only linkage error and runtime exception

jonmv:
@bratseth please review.


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutorImpl.java
```diff
@@ -148,11 +148,11 @@ private static VisitorParameters asParameters(VisitorOptions options, Map<String
     /** Assumes this stops receiving operations roughly when this is called, then waits up to 50 seconds to drain operations. */
```
jonmv, line 148:
```suggestion
    /** Assumes this stops receiving operations roughly when this is called, then waits up to 20 seconds to drain operations. */
```

bratseth (approved):

PR #14675, opened 2020-10-02
Add timeout to cluster controller test 

hmusum:

### DIFF
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/MasterElectionTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/MasterElectionTest.java
@@ -16,6 +16,7 @@ import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestRule;
+import org.junit.rules.Timeout;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/MasterElectionTest.java
```diff
@@ -226,24 +230,20 @@ public void testClusterStateVersionIncreasesAcrossMasterElections() throws Excep
         startingTest("MasterElectionTest::testClusterStateVersionIncreasesAcrossMasterElections");
         FleetControllerOptions options = defaultOptions("mycluster");
         options.masterZooKeeperCooldownPeriod = 1;
-        setUpFleetController(5, false, options);
+        setUpFleetController(3, false, options);
```
vekterli, line 233: I'm assuming this is to reduce the amount of load generated by the test?

vekterli (approved):
:+1: thank you for the cleanup!

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/MasterElectionTest.java
hmusum: Yes, and we usually have 3 and not 5 controllers. But this is not important, I think I just tried it out

PR #14679, opened 2020-10-02
Don't reactivate on retired hosts

bratseth:
With this it became less legal to have tenant nodes with nonexisting parents so I had to update some tests, but this is an improvement I think.


mpolden (approved):
Great!

PR #14682, opened 2020-10-02
Remove custom Utf8.toString implementation

mpolden:
`String::new` is now faster for both ASCII and Unicode strings:

```
Utf8::toString of ascii string took 132 ms
String::new of ascii string took 59 ms
Change = -55.30%
Utf8::toString of unicode string took 410 ms
String::new of unicode string took 280 ms
Change = -31.71%
```

There's at least two reasons for this:

* Java 9 introduced compact strings, which means that `String` is now backed by
a byte array to reduce the memory footprint of ASCII strings.
* Detection of Unicode strings may use HotSpot intrinsics.

@baldersheim


baldersheim (approved):
Finally....

PR #14683, opened 2020-10-02
Add support for multiple RPC targets per node when using Storage API …

geirst:
…over RPC.

This should allow for better parallelization and higher feed throughput.
The bucket id associated with a message is used to select the RPC target.
This ensures the same RPC target is used for all messages to the same bucket to the same node,
and the RPC target itself handles sequencing of these messages.

@vekterli please review
@baldersheim @toregge FYI


storage/src/vespa/storage/storageserver/rpc/storage_api_rpc_service.cpp
```diff
@@ -216,7 +220,7 @@ void StorageApiRpcService::send_rpc_v1_request(std::shared_ptr<api::StorageComma
         cmd->getType().getName().c_str(), cmd->getAddress()->toString().c_str());
 
     assert(cmd->getAddress() != nullptr);
-    auto target = _target_resolver->resolve_rpc_target(*cmd->getAddress());
+    auto target = _target_resolver->resolve_rpc_target(*cmd->getAddress(), cmd->getBucketId().getId());
```
vekterli, line 223: Before we can start using this in production I believe we must derive and use the superbucket here rather than the sub-bucket. This is because operations in flight can be remapped to other buckets when their original destination bucket is split or joined. Since RPC requests are bound to the target for the original, _non-remapped_ bucket it's possible for sequencing to be broken for a single bucket when the replies are sent. Remapped buckets are always contained within the superbucket, so they will always be bound to the same RPC target if we use it instead.

vekterli (approved):
:+1: just one note for future work

storage/src/vespa/storage/storageserver/rpc/storage_api_rpc_service.cpp
geirst: Fixed in https://github.com/vespa-engine/vespa/pull/15617.

PR #14691, opened 2020-10-03
Move lock metrics to MetricsReporter

hakonhall:
Adds two new metrics:

 - The load of acquiring each lock path: The average number of threads waiting
   to acquire the lock within the last minute (or unit of time).  Aka the lock
   queue (depth).

 - The load of the lock for each lock path: The average number of threads
   holding the lock within the last minute (or unit of time).  This is always
   <= 1.  Aka the lock utilization.

Changes the LockCounters to LockMetrics, and exporting those once every minute
through MetricReporter which is designed for this.


hakonhall:
The load of an ongoing lock is reported only at the end, and causes a spike there.  Working on smoothing that out.

mpolden (approved):
👍

hakonhall:
Let me merge to try to get into the release tomorrow - please post-review...

mpolden:
👍

PR #14697, opened 2020-10-03
Add maxpending which by default is 1000 to control max inflight.

baldersheim:
@toregge PR

### DIFF
--- a/searchcore/src/apps/vespa-feed-bm/vespa_feed_bm.cpp
+++ b/searchcore/src/apps/vespa-feed-bm/vespa_feed_bm.cpp
@@ -237,7 +237,7 @@ public:
           _num_buckets((num_buckets_in / _threads) * _threads)
     {
     }
-    uint64_t operator()(uint32_t i) {
+    uint64_t operator()(uint32_t i) const {
         return (static_cast<uint64_t>(i) * _threads + _thread_id) % _num_buckets;
     }
 };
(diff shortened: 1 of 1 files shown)


searchcore/src/apps/vespa-feed-bm/vespa_feed_bm.cpp
```diff
@@ -1313,6 +1318,7 @@ App::get_options()
     enum longopts_enum {
         LONGOPT_CLIENT_THREADS,
         LONGOPT_INDEXING_SEQUENCER,
+        LONGOPT_MAX_PENDING,
```
toregge, line 1321: Usage message is not updated, and order is inconsistent with option array.

toregge (changes requested):

searchcore/src/apps/vespa-feed-bm/vespa_feed_bm.cpp
baldersheim: Fixed

toregge: "documents" is before  "max-pending" in long_opts array, thus LONGOPT_MAX_PENDING should be after LONGOPT_DOCUMENTS in longopts_enum.

toregge (changes requested):

searchcore/src/apps/vespa-feed-bm/vespa_feed_bm.cpp
baldersheim: Now they should all be in alphabetical ordering.

toregge (approved):

PR #14711, opened 2020-10-05
Revert "Move code out of RemoteSesion"

hmusum:
Reverts vespa-engine/vespa#14690

In case this breaks something. Please review only.


jonmv (approved):

hmusum:
System tests are good, but performance tests show that deployment is taking more time, need to look into that.

PR #14712, opened 2020-10-05
Separate error code for insufficient storage

jonmv:
@bjorncs or @vekterli please review.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
@@ -52,8 +52,9 @@ public interface DocumentOperationExecutor {
         NOT_FOUND,
         PRECONDITION_FAILED,
         BAD_REQUEST,
+        ERROR,
         TIMEOUT,
-        ERROR;
+        INSUFFICIENT_STORAGE;
     }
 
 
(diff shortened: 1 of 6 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
```diff
@@ -52,8 +52,9 @@ default void shutdown() { }
         NOT_FOUND,
         PRECONDITION_FAILED,
         BAD_REQUEST,
+        ERROR,
```
bjorncs, line 55: Reordering an enum should in practice not be a problem, but any usage of `.ordinal()` may result in unexpected behaviour.

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
jonmv: Which is why the use of `ordinal()` is discouraged, and definitely not used for this (brand new) enum :)

vekterli:
Fails Travis build with
```
[ERROR] Failed to execute goal com.yahoo.vespa:abi-check-plugin:7-SNAPSHOT:abicheck (default) on project documentapi: ABI spec mismatch. To update run 'mvn package -Dabicheck.writeSpec' -> [Help 1]
```

jonmv:
https://github.com/vespa-engine/vespa/pull/14716
`@PublicApi` :)))))))

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutor.java
bjorncs: Both `.name()` and `.ordinal()` is bad practice, yet we do it (at least the former) in our own code ;)

PR #14721, opened 2020-10-05
Remove unused Utf8 methods

mpolden:
Only used in a few tests, and are now equivalent to `toString`/`toBytes`.

@baldersheim


baldersheim (approved):

bratseth:
Don't break public API's please.

mpolden:
Sorry, my fault. We do make a lot of cruft public though...

PR #14723, opened 2020-10-05
Remove node-object-cache-size flag

mpolden:

@hakonhall


hakonhall (approved):
Why do you make it a config?

mpolden:
To avoid having a long-lived feature flag, but still allowing a per-zone value if needed.

PR #14727, opened 2020-10-05
Log at lower level when moving file and file already exists

hmusum:
No need to log at info level, this is not a filure and will happen
occasionally. Also minor change so that we log if deleting a file
fails.

### DIFF
--- a/filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReceiver.java
+++ b/filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReceiver.java
@@ -56,7 +56,7 @@ public class FileReceiver {
         private final long fileSize;
         private long currentFileSize;
         private long currentPartId;
-        private long currentHash;
+        private final long currentHash;
         private final File fileReferenceDir;
         private final File tmpDir;
         private final File inprogressFile;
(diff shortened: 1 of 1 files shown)


filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReceiver.java
```diff
@@ -96,9 +96,12 @@ void addPart(int partId, byte [] part) {
             try {
                 Files.write(inprogressFile.toPath(), part, StandardOpenOption.WRITE, StandardOpenOption.APPEND);
             } catch (IOException e) {
-                log.log(Level.SEVERE, "Failed writing to file (" + inprogressFile.toPath() + "): " + e.getMessage(), e);
-                inprogressFile.delete();
-                throw new RuntimeException("Failed writing to file (" + inprogressFile.toPath() + "): ", e);
+                String message = "Failed writing to file (" + inprogressFile.toPath() + "): ";
+                log.log(Level.SEVERE, message + e.getMessage(), e);
```
hakonhall, line 100: Consider catching the particular subclass of IOException thrown when the file already exists (FileAlreadyExistsException?), and only try deletion etc then.

hakonhall (approved):

filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReceiver.java
hmusum: If we fail to write to the file we need to cleanup (we append to the file, so if it fails we have no idea of what has been written). So the change is just logging if we were unable to delete the file, we always need to delete no matter what the IOException was.

hakonhall: 👍

PR #14739, opened 2020-10-06
Don't use request headers for remote address/port in hosted Vespa

bjorncs:
Control which headers are used for remote address/port in access log through config model.


baldersheim (approved):
It looks fine to me, but I suggest letting it pass by someone else too.

PR #14740, opened 2020-10-06
Stop taking config lock exclusively in node-repository

mpolden:
We believe we've identified how locking should behave, at least in the
short-term:

1) Migrate existing use of `/config/v2/locks/<application-id>` inside
node-repository to only use `/provision/v1/locks/<application-id>`. This ensures
that prepare only acquires `/provision/v1/locks/<application-id>` on the
node-repository side. Locks and node repo writes always happen together, i.e.
we're not building a list of transactions that are committed at some later point.

2) Ensure that transactional operations such as `active` and `deactivate` holds
both `/config/v2/locks/<application-id>` and
`/provision/v1/locks/<application-id>` until the transaction is committed.

@hakonhall

FYI @hmusum

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/LoadBalancerProvisioner.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/LoadBalancerProvisioner.java
@@ -60,8 +60,11 @@ public class LoadBalancerProvisioner {
         // Read and write all load balancers to make sure they are stored in the latest version of the serialization format
         for (var id : db.readLoadBalancerIds()) {
             try (var lock = db.lock(id.application())) {
-                var loadBalancer = db.readLoadBalancer(id);
-                loadBalancer.ifPresent(db::writeLoadBalancer);
+                // TODO(mpolden): Remove inner lock
+                try (var innerLock = db.configLock(id.application())) {
+                    var loadBalancer = db.readLoadBalancer(id);
+                    loadBalancer.ifPresent(db::writeLoadBalancer);
+                }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/LoadBalancerProvisioner.java
```diff
@@ -60,8 +60,11 @@ public LoadBalancerProvisioner(NodeRepository nodeRepository, LoadBalancerServic
         // Read and write all load balancers to make sure they are stored in the latest version of the serialization format
         for (var id : db.readLoadBalancerIds()) {
             try (var lock = db.lock(id.application())) {
-                var loadBalancer = db.readLoadBalancer(id);
-                loadBalancer.ifPresent(db::writeLoadBalancer);
+                // TODO(mpolden): Remove inner lock
+                try (var innerLock = db.configLock(id.application())) {
```
hakonhall, line 64: This code is executed on bootstrap, so if a big deployment is happening concurrently the above db.lock() will time out.  But bootstrap redeployments are also vulnerable to this.

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/LoadBalancerProvisioner.java
mpolden: Yes. This is also why I left out locking in the `NodeRepository` constructor even though it's wrong. If a lock timeout happens at construction time it should recover though as the container will restart and try to construct the components again, and by then deployments are hopefully done.

PR #14741, opened 2020-10-06
Use push_back() instead of emplace_back()

toregge:
@baldersheim : please review
@havardpe : FYI


havardpe:
why is the stringref conversion operator on string better than the string constructor on stringref?

havardpe (approved):

havardpe:
my bad, I thought you could create (construct) a stringref from a string.

PR #14746, opened 2020-10-06
Remove unneeded noexcept specifiers.

toregge:
@baldersheim : please review

This supersedes #14742 and #14745


baldersheim (changes requested):
I do not think this solves anything. This will choose copy constructs instead of move constructors in stl structures. We must figure out what the underlying problem is.

toregge:
The warnings went away after upgrading compiler (AppleClang) from 11.0.0. to 12.0.0.

PR #14747, opened 2020-10-06
Use noexcept as advise by gcc -Wnoexcept

baldersheim:
@toregge PR


toregge (approved):

toregge:
Note that some existing `noexcept` specifiers for `vespalib::small_string` might have to be revisited (i.e. do we want to keep them or do we want to match std::string semantics).

PR #14749, opened 2020-10-07
Refactor session repo

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
```diff
@@ -138,7 +138,7 @@ public boolean exists(ApplicationId id) {
                 : data.map(bytes -> Long.parseLong(Utf8.toString(bytes)));
     }
 
-    public boolean hasLocalSession(long sessionId) {
+    public boolean sessionExistsInFileSystem(long sessionId) {
```
jonmv, line 141: Muuuuch better :)

jonmv (approved):
Nice :)

PR #14757, opened 2020-10-07
Add tensor partial update 2

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/tests/tensor/partial_remove/partial_remove_test.cpp
+++ b/eval/src/tests/tensor/partial_remove/partial_remove_test.cpp
@@ -60,11 +60,8 @@ Value::UP try_partial_remove(const TensorSpec &a, const TensorSpec &b) {
 
 TensorSpec perform_partial_remove(const TensorSpec &a, const TensorSpec &b) {
     auto up = try_partial_remove(a, b);
-    if (up) {
-        return spec_from_value(*up);
-    } else {
-        return TensorSpec(a.type());
-    }
+    EXPECT_TRUE(up);
+    return spec_from_value(*up);
 }
 
 TensorSpec perform_old_remove(const TensorSpec &a, const TensorSpec &b) {
@@ -82,7 +79,7 @@ TensorSpec perform_old_remove(const TensorSpec &a, const TensorSpec &b) {
 }
 
 
-TEST(PartialAddTest, partial_remove_works_for_simple_values) {
+TEST(PartialRemoveTest, partial_remove_works_for_simple_values) {
     ASSERT_TRUE((remove_layouts.size() % 2) == 0);
     for (size_t i = 0; i < remove_layouts.size(); i += 2) {
         TensorSpec lhs = spec(remove_layouts[i], N());
@@ -94,7 +91,7 @@ TEST(PartialAddTest, partial_remove_works_for_simple_values) {
     }
 }
 
-TEST(PartialAddTest, partial_remove_works_like_old_remove) {
+TEST(PartialRemoveTest, partial_remove_works_like_old_remove) {
     ASSERT_TRUE((remove_layouts.size() % 2) == 0);
     for (size_t i = 0; i < remove_layouts.size(); i += 2) {
         TensorSpec lhs = spec(remove_layouts[i], N());
@@ -103,8 +100,6 @@ TEST(PartialAddTest, partial_remove_works_like_old_remove) {
         auto expect = perform_old_remove(lhs, rhs);
         auto actual = perform_partial_remove(lhs, rhs);
         EXPECT_EQ(actual, expect);
-        // printf("%s remove %s -> %s\n", lhs.to_string().c_str(), rhs.to_string().c_str(), actual.to_string().c_str());
-
     }
 }
 
--- a/eval/src/tests/tensor/partial_remove/partial_remove_test.cpp
+++ b/eval/src/tests/tensor/partial_remove/partial_remove_test.cpp
@@ -1,0 +1,106 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/eval/eval/simple_value.h>
+#include <vespa/eval/eval/test/tensor_model.hpp>
+#include <vespa/eval/eval/value_codec.h>
+#include <vespa/eval/tensor/cell_values.h>
+#include <vespa/eval/tensor/default_tensor_engine.h>
+#include <vespa/eval/tensor/partial_update.h>
+#include <vespa/eval/tensor/sparse/sparse_tensor.h>
+#include <vespa/eval/tensor/tensor.h>
+#include <vespa/vespalib/util/stringfmt.h>
+#include <vespa/vespalib/gtest/gtest.h>
+#include <optional>
+
+using namespace vespalib;
+using namespace vespalib::eval;
+using namespace vespalib::eval::test;
+
+using vespalib::make_string_short::fmt;
+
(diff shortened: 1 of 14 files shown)


eval/src/vespa/eval/eval/tensor_spec.cpp
```diff
@@ -46,9 +46,14 @@ TensorSpec & TensorSpec::operator = (const TensorSpec &) = default;
 TensorSpec::~TensorSpec() { }
 
 TensorSpec &
-TensorSpec::set(Address address, double value) {
-    auto res = _cells.emplace(std::move(address), value);
-    if (!res.second) { assert(res.first->second.value == value); }
+TensorSpec::add(Address address, double value) {
+    auto [iter, inserted] = _cells.emplace(std::move(address), value);
+    if (! inserted) {
+        // to simplify reference concat implementation, allow
```
havardpe, line 52: I suggest dropping concat specifically, since others may also have use of this

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,388 @@
...
+                how.push_back(a.is_mapped() ? DimCase::MAPPED_MATCH : DimCase::CONV_TO_INDEXED);
+            }
+        };
+        const auto & input_dims = input_type.dimensions();
+        const auto & modifier_dims = modifier_type.dimensions();
+        visit_ranges(visitor,
+                     input_dims.begin(), input_dims.end(),
+                     modifier_dims.begin(), modifier_dims.end(),
+                     [](const auto &a, const auto &b){ return (a.name < b.name); });
+        if ((! valid) ||
```
havardpe, line 127: consider checking number of dimensions together with modifier sparseness and only checking valid here.

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,388 @@
...
+        assert(target_coords.dim == target_coords.dim_sizes.size());
+    }
+
+    ~AddressHandler();
+};
+AddressHandler::~AddressHandler() = default;
+
+template <typename CT>
+Value::UP
+copy_tensor(const Value &input, const ValueType &input_type, Addresses &helper, const ValueBuilderFactory &factory)
```
havardpe, line 165: consider doing simple encode/decode for value copy

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+//-----------------------------------------------------------------------------
+
+template <typename ICT>
+Value::UP
+my_remove_cells(const Value &input, const Value &modifier, const ValueBuilderFactory &factory)
+{
+    const ValueType &input_type = input.type();
+    const ValueType &modifier_type = modifier.type();
+    if (input_type.mapped_dimensions() != modifier_type.mapped_dimensions()) {
```
havardpe, line 307: compare to modifier.dimensions() directly to ensure there are no dense ones as well.

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,388 @@
...
+    }
+    if (input_type.mapped_dimensions().size() == 0) {
+        LOG(error, "cannot remove cells from a dense tensor");
+        return Value::UP();
+    }
+    const auto input_cells = input.cells().typify<ICT>();
+    const size_t num_mapped_in_input = input_type.count_mapped_dimensions();
+    const size_t dsss = input_type.dense_subspace_size();
+    Addresses addrs(num_mapped_in_input);
+    std::set<size_t> removed_subspaces;
```
havardpe, line 319: no need to precalculate this

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,388 @@
...
+        if (lookup_view->next_result({}, input_subspace_index)) {
+            removed_subspaces.insert(input_subspace_index);
+        }
+    }
+    const size_t expected_subspaces = input.index().size() - removed_subspaces.size();
+    auto builder = factory.create_value_builder<ICT>(input_type, num_mapped_in_input, dsss, expected_subspaces);
+    auto input_view = input.index().create_view({});
+    input_view->lookup({});
+    size_t input_subspace_index;
+    while (input_view->next_result(addrs.next_result_refs, input_subspace_index)) {
```
havardpe, line 336: just look up directly in the modify view to find out which ones not to include

arnej27959:
@havardpe PTAL
@geirst please review also

havardpe (approved):
looks good, but I still think the remove case would be a bit simpler if you looked up the sparse address during copy directly in the modify value (what to remove) rather than building an explicit subspace filter up front... If you are worried about code re-use, the explicit std::set<size_t> could be replaced with a templated predicate that takes both the sparse address and the subspace as input.

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -0,0 +1,347 @@
...
+    ~AddressHandler();
+};
+AddressHandler::~AddressHandler() = default;
+
+template <typename CT, typename ICT = CT, typename SkipFun>
+void copy_tensor_with_filter(const Value &input,
+                             size_t dsss,
+                             SparseCoords &addrs,
+                             ValueBuilder<CT> &builder,
+                             SkipFun && skip_subspaces)
```
havardpe, line 158: I suggest switching from skip to keep semantic on the predicate. Also, I strongly suggest passing both the address and the subspace to avoid depending too much on side-effect on objects passed into this function.

eval/src/vespa/eval/tensor/partial_update.cpp
```diff
@@ -261,17 +263,17 @@ PerformAdd::invoke(const Value &input, const Value &modifier, const ValueBuilder
...
             overwritten[input_subspace_index] = true;
         }
-        return false;
+        return true;
     };
     copy_tensor_with_filter<ICT, MCT>(modifier, dsss, addrs, *builder, remember_subspaces);
-    auto filter = [&] (size_t input_subspace) {
-        return overwritten[input_subspace];
+    auto filter = [&] (size_t input_subspace, const auto &) {
+        return ! overwritten[input_subspace];
```
havardpe, line 276: consider not adding more entries when checking old entries

eval/src/tests/tensor/partial_remove/partial_remove_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+TEST(PartialAddTest, partial_remove_works_like_old_remove) {
+    ASSERT_TRUE((remove_layouts.size() % 2) == 0);
+    for (size_t i = 0; i < remove_layouts.size(); i += 2) {
+        TensorSpec lhs = spec(remove_layouts[i], N());
+        TensorSpec rhs = spec(remove_layouts[i + 1], Div16(N()));
+        SCOPED_TRACE(fmt("\n===\nLHS: %s\nRHS: %s\n===\n", lhs.to_string().c_str(), rhs.to_string().c_str()));
+        auto expect = perform_old_remove(lhs, rhs);
+        auto actual = perform_partial_remove(lhs, rhs);
+        EXPECT_EQ(actual, expect);
+        // printf("%s remove %s -> %s\n", lhs.to_string().c_str(), rhs.to_string().c_str(), actual.to_string().c_str());
```
geirst, line 106: Consider removing remains of debug printing.

geirst (commented):
👍

PR #14758, opened 2020-10-07
fast value to enable inlined sparse operations

havardpe:
use full overlap sparse join as initial test of full inlining.

also improve simple sparse map performance by pre-calculating string
hashes and using hash values for equality checks.

@arnej27959 please review
@geirst FYI

NB: I think we want to fork the SimpleValue and FastValue implementations and use a string-based std::map for the simple one, and rename SimpleSparseMap to FastSparseMap. I will do that as a follow-up.

### DIFF
--- a/eval/src/tests/eval/simple_sparse_map/simple_sparse_map_test.cpp
+++ b/eval/src/tests/eval/simple_sparse_map/simple_sparse_map_test.cpp
@@ -24,7 +24,7 @@ public:
         }
     }
     ~StringList();
-    const std::vector<vespalib::string> direct_str() const { return _str_list; }
+    ConstArrayRef<vespalib::string> direct_str() const { return _str_list; }
     ConstArrayRef<vespalib::stringref> direct_ref() const { return _ref_list; }
     ConstArrayRef<const vespalib::stringref *> indirect_ref() const { return _ref_ptr_list; }
 };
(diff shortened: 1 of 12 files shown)


eval/src/tests/eval/simple_sparse_map/simple_sparse_map_test.cpp
```diff
@@ -55,10 +55,12 @@ TEST(SimpleSparseMapTest, simple_sparse_map_basic_usage_works) {
     EXPECT_EQ(map.lookup(a4.direct_ref()), map.npos());
     EXPECT_EQ(map.lookup(a4.indirect_ref()), map.npos());
     EXPECT_EQ(SimpleSparseMap::npos(), map.npos());
-    SL expect_labels({"a","a","a",
-                      "a","a","b",
-                      "a","b","a"});
-    EXPECT_EQ(map.labels(), expect_labels.direct_str());
+    EXPECT_EQ(map.labels().size(), 9);
+    auto dump = [&](auto addr_tag, auto subspace, auto hash) {
+        auto addr = map.make_addr(addr_tag);
```
arnej27959, line 60: dumping to stderr should be temporary

arnej27959 (approved):
the map seems actually easier to understand, good work 👍

PR #14759, opened 2020-10-07
Track test-and-set condition mismatches as own backend metric

vekterli:
@geirst please review

Avoids polluting the generic failure metric with condition mismatches,
making it easier to discern when real service problems are occurring.


geirst (approved):
👍

PR #14761, opened 2020-10-07
Destroy visitor session in other thread to avoid deadlock, and 1000 q…

jonmv:
@bjorncs please review. This will let us know early if there are more things to fix.


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutorImpl.java
```diff
@@ -85,6 +86,7 @@
     private final DelayQueue throttled;
     private final DelayQueue timeouts;
     private final Map<VisitorControlHandler, VisitorSession> visits = new ConcurrentHashMap<>();
+    private final Executor visitSessionShutdownExecutor = Executors.newSingleThreadExecutor();
```
bjorncs, line 89: The executor must be shutdown in `close`/`deconstruct`

bjorncs (changes requested):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/DocumentOperationExecutorImpl.java
bjorncs: Also give the thread a name

bjorncs (approved):

PR #14763, opened 2020-10-07
Remove broken copy and move constructor and assignment operatoos on v…

baldersheim:
…espalib::Lock and vespalib::Monitor.

Also repair broken usages of the same.

@toregge or @vekterli PR


toregge (approved):

vekterli (commented):
:+1: those Monitor/Lock ctors/assignment operators were scarier than I thought 😬

PR #14765, opened 2020-10-07
Zero delay — just the 10ms throttle wait — for throttle queue

jonmv:
@bjorncs please review.


jonmv:
Without this, once the queue is used, all requests will be delayed by 100ms. Silly.

hakonhall (commented):
👍

PR #14770, opened 2020-10-08
Arnej/less needless casting

arnej27959:

now that we have cells() on value, casting to DenseTensorView in order to call cellsRef() seems ugly

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/tensor/dense/dense_tensor_view.h
+++ b/eval/src/vespa/eval/tensor/dense/dense_tensor_view.h
@@ -52,10 +52,10 @@ public:
     }
 
     template <typename T> static ConstArrayRef<T> typify_cells(const eval::Value &self) {
-        return static_cast<const DenseTensorView &>(self).cellsRef().typify<T>();
+        return self.cells().typify<T>();
     }
     template <typename T> static ConstArrayRef<T> unsafe_typify_cells(const eval::Value &self) {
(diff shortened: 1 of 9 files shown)


eval/src/vespa/eval/tensor/dense/dense_tensor_view.h
```diff
@@ -52,10 +52,10 @@ class DenseTensorView : public Tensor
     }
 
     template <typename T> static ConstArrayRef<T> typify_cells(const eval::Value &self) {
-        return static_cast<const DenseTensorView &>(self).cellsRef().typify<T>();
+        return self.cells().typify<T>();
```
havardpe, line 55: prefer removing this and fixing callers instead

eval/src/vespa/eval/tensor/dense/dense_tensor_view.h
```diff
@@ -52,10 +52,10 @@ class DenseTensorView : public Tensor
...
     }
     template <typename T> static ConstArrayRef<T> unsafe_typify_cells(const eval::Value &self) {
-        return static_cast<const DenseTensorView &>(self).cellsRef().unsafe_typify<T>();
+        return self.cells().unsafe_typify<T>();
```
havardpe, line 58: prefer removing this and fixing callers instead

havardpe (approved):
consider removing the 'cellsRef' function from DenseTensorView and making the 'cell' function final.

this looks good. All comments can be resolved in a follow-up PR

PR #14777, opened 2020-10-08
Emit restart actions on node resource change

bratseth:
The actions should track what changes will actually cause restarts
such that we can reason about the consequences of restarts based on them.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/StorageCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/StorageCluster.java
@@ -45,13 +45,12 @@ public class StorageCluster extends AbstractConfigProducer<StorageNode>
         }
     }
 
-    private Integer bucketMoverMaxFillAboveAverage = null;
-    private String clusterName;
-    private FileStorProducer fileStorProducer;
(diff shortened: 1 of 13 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/StorageCluster.java
```diff
@@ -71,9 +70,6 @@ protected StorageCluster doBuild(DeployState deployState, AbstractConfigProducer
 
     @Override
     public void getConfig(StorBucketmoverConfig.Builder builder) {
-        if (bucketMoverMaxFillAboveAverage != null) {
-            builder.max_target_fill_rate_above_average(bucketMoverMaxFillAboveAverage);
```
freva, line 75: :man_shrugging:

freva (approved):

config-model/src/main/java/com/yahoo/vespa/model/content/storagecluster/StorageCluster.java
bratseth: Not used

bratseth: (IntelliJ advice)

PR #14788, opened 2020-10-08
add optimization for the simplest dense-only concat case

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/generic_concat.cpp
+++ b/eval/src/vespa/eval/instruction/generic_concat.cpp
@@ -3,6 +3,7 @@
 #include "generic_concat.h"
 #include "generic_join.h"
 #include <vespa/eval/eval/value.h>
+#include <vespa/eval/tensor/dense/dense_tensor_view.h>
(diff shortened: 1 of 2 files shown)


eval/src/vespa/eval/instruction/generic_concat.cpp
```diff
@@ -93,8 +94,36 @@ void my_generic_concat_op(State &state, uint64_t param_in) {
     state.pop_pop_push(result_ref);
 }
 
+template <typename LCT, typename RCT>
+void my_dense_simple_concat_op(State &state, uint64_t param_in) {
+    using OCT = typename eval::UnifyCellTypes<LCT, RCT>::type;
```
havardpe, line 99: if you concat a double and a vector of floats, the result type will be float and not double. This issue seems to also be in the generic version, and not tested in the unit test.

havardpe (commented):
please also extend unit test with double vs vector of float concat

havardpe (approved):

PR #14789, opened 2020-10-08
Record quota usage

oyving:
- Expose the maximum quota usage rate in the configuration server
- Query for the quota usage rate right after deploy in the controller
- Store the quota usage rate together with the deployment information in ZK


mpolden (approved):
👍

mpolden (approved):

PR #14790, opened 2020-10-08
remove typify_cells

arnej27959:
* replace: DenseTensorView::typify_cells<T>(foo)
     with: foo.cells().typify<T>()
* costs one extra virtual function call
* also remove DenseTensorView::unsafe_typify_cells (was unused)

@havardpe please review


havardpe (approved):
we could also consider removing the cellsRef function from DenseTensorView

PR #14792, opened 2020-10-08
require newer xxhash

arnej27959:

with the recent change to vespalib::hash some unit tests failed for me:
```
staging_vespalib_adaptive_sequenced_executor_test_app 
adaptive_sequenced_executor_test.cpp: info:  running test suite 'adaptive_sequenced_executor_test.cpp'
[...]
adaptive_sequenced_executor_test.cpp: info:  status_for_test 'require that task with same string component id are serialized': PASS
adaptive_sequenced_executor_test.cpp:177: error: check failure #1: 'tryCnt < 100' in thread '0(1)' (adaptive_sequenced_executor_test.cpp)
    STATE[0]: 'require that task with different string component ids are not serialized' (adaptive_sequenced_executor_test.cpp:174)
adaptive_sequenced_executor_test.cpp: ERROR: status_for_test 'require that task with different string component ids are not serialized': FAIL
```

dist/vespa.spec claimed xxhash 0.7.3 should work, but I had to upgrade to 0.8.0.

@vekterli please review
@aressem @geirst FYI

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -110,7 +110,7 @@ BuildRequires: gtest-devel
 BuildRequires: gmock-devel
 %endif
 %endif
-BuildRequires: xxhash-devel >= 0.7.3
+BuildRequires: xxhash-devel >= 0.8.0
 BuildRequires: openblas-devel
 BuildRequires: zlib-devel
 BuildRequires: re2-devel
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -151,7 +151,7 @@ Requires: valgrind
 %endif
 Requires: Judy
 Requires: xxhash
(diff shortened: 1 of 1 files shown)


vekterli (approved):
:+1: XXH3 was moved into main `xxhash.h` in 0.8.0, so that makes sense. I thought we were already on that version since that's what I had in my dev container.

vekterli:
Also, 0.8.0 was when XXH3 became stable across versions, so that might be the real reason.

toregge:
What about the lines requiring xxhash-libs ? Should they also be changed from 0.7.3 to 0.8.0 ?

arnej27959:
yes, i would guess xxhash-libs should also be updated.

toregge (approved):

PR #14797, opened 2020-10-09
pack FastSparseMap::Key tighter

arnej27959:

@havardpe somewhat experimental, please take a look

### DIFF
--- a/eval/src/vespa/eval/eval/fast_sparse_map.h
+++ b/eval/src/vespa/eval/eval/fast_sparse_map.h
@@ -66,7 +66,7 @@ public:
         Key() : hash(0) {}
         Key(uint64_t hash_in)
             : hash(hash_in) {}
-    } __attribute__((aligned(4),packed));
+    } __attribute__((packed,aligned(4)));
 
     struct Hash {
         uint64_t operator()(const Key &key) const { return key.hash; }
--- a/eval/src/vespa/eval/eval/fast_sparse_map.h
+++ b/eval/src/vespa/eval/eval/fast_sparse_map.h
@@ -62,12 +62,11 @@ public:
     }
 
     struct Key {
-        uint32_t start;
         uint64_t hash;
-        Key() : start(0), hash(0) {}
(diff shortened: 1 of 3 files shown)


eval/src/vespa/eval/eval/fast_sparse_map.h
```diff
@@ -148,7 +147,7 @@ class FastSparseMap
     void each_map_entry(F &&f) const {
         _map.for_each([&](const auto &entry)
                       {
-                          f(entry.first.start, entry.second, entry.first.hash);
+                          f(entry.second * _num_dims, entry.second, entry.first.hash);
```
havardpe, line 150: an alternative is to not pass start and redefine the make_addr function instead to get an address from a subspace directly. That would trade a divide with a multiply for the iterate case and we would save the multiply for the time we actually need the address.

eval/src/vespa/eval/eval/fast_sparse_map.h
```diff
@@ -62,12 +62,11 @@ class FastSparseMap
...
-        uint32_t start;
         uint64_t hash;
-        Key() : start(0), hash(0) {}
-        Key(uint32_t start_in, uint64_t hash_in)
-            : start(start_in), hash(hash_in) {}
-    };
+        Key() : hash(0) {}
+        Key(uint64_t hash_in)
+            : hash(hash_in) {}
+    } __attribute__((packed));
```
havardpe, line 69: if we pack, I think we should also use align 4

eval/src/vespa/eval/eval/fast_sparse_map.h
```diff
@@ -66,7 +66,7 @@ class FastSparseMap
         Key() : hash(0) {}
         Key(uint64_t hash_in)
             : hash(hash_in) {}
-    } __attribute__((packed));
+    } __attribute__((aligned(4),packed));
```
havardpe, line 69: I am not sure if attributes are order sensitive, but we should switch them just to be safe.

havardpe (approved):

PR #14801, opened 2020-10-09
Signal deference of config change application until restart

bratseth:
When a restart needs to be done, processes should not apply
new config until restarted. This signals to clusters in the config model
when this obtains, but this is a no-op for now.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
@@ -111,6 +111,7 @@ public class Validation {
     private static void deferConfigChangesForClustersToBeRestarted(List<ConfigChangeAction> actions, VespaModel model) {
         Set<ClusterSpec.Id> clustersToBeRestarted = actions.stream()
                                                            .filter(action -> action.getType() == ConfigChangeAction.Type.RESTART)
+                                                           .filter(action -> action.clusterId() != null) // TODO: Remove this line after October 2020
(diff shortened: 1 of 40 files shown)


config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
```diff
@@ -103,4 +108,20 @@ private static void validateFirstTimeDeployment(VespaModel model, DeployState de
         new AccessControlOnFirstDeploymentValidator().validate(model, deployState);
     }
 
+    private static void deferConfigChangesForClustersToBeRestarted(List<ConfigChangeAction> actions, VespaModel model) {
+        Set<ClusterSpec.Id> clustersToBeRestarted = actions.stream()
+                                                           .filter(action -> action.getType() == ConfigChangeAction.Type.RESTART)
+                                                           .map(action -> action.clusterId())
+                                                           .collect(Collectors.toSet());
+        for (var clusterToRestart : clustersToBeRestarted) {
+            var containerCluster = model.getContainerClusters().get(clusterToRestart.value());
```
freva, line 117: Wont this throw NPE?

freva (approved):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
bratseth: Yes, I forgot about that. Excellent catch!

freva (approved):
:100:

PR #14802, opened 2020-10-09
Reuse threadpool

bjorncs:
Reuse threadpool between metric requests


baldersheim (approved):
The intention seems good. But you might want to get another reviewer too.
Why is everything static ? Not a fan of that though.

bjorncs:
Spawning new ad-hoc is okay if latency is not a big problem and it's done relatively seldom. The JVM needs a GC to clean up old threads. The biggest issue with the original implementation was that it did not close the thread pool correctly. Existing tasks were not interrupted, causing threads to look if the http client blocked forever.

bjorncs:
@olaaun Do you remember the motivation for using an adhoc instead of a persistent threadpool?

PR #14803, opened 2020-10-09
Set RequestType both ways

bratseth:
This sets the requestType automagically also for those who use both don't extend SearchHandler.


bratseth:
@bjorncs

bjorncs (approved):

PR #14805, opened 2020-10-10
Take StateMonitor snapshot every 60s

hakonhall:
Using Executors simplifies the code and fixes the following 2 small problems:
 - A tiny drift of the 1m interval: It starts the next snapshot 1m + the time
   before wait() return + the time until currentTimeMillis().
 - May potentially (but unlikely) invoke wait() with negative (throws
   exception) or 0 argument (waits forever).


hakonhall:
ping @baldersheim

baldersheim (approved):

baldersheim:
Sorry, missed this one.

PR #14806, opened 2020-10-10
Optimistic check for bucket presence without lock.

baldersheim:
@vekterli @geirst @toregge PR

### DIFF
--- a/storage/src/vespa/storage/bucketdb/btree_lockable_map.hpp
+++ b/storage/src/vespa/storage/bucketdb/btree_lockable_map.hpp
@@ -366,6 +366,7 @@ public:
     std::vector<T> find_parents_self_and_children(const document::BucketId& bucket) const override;
     void for_each(std::function<void(uint64_t, const T&)> func) const override;
     [[nodiscard]] uint64_t generation() const noexcept override;
+    bool hasKey(uint64_t key) const override;
 };
 
 template <typename T>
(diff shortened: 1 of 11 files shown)


storage/src/vespa/storage/bucketdb/btree_lockable_map.hpp
```diff
@@ -179,6 +179,12 @@ void BTreeLockableMap<T>::acquireKey(const LockId& lid, std::unique_lock<std::mu
     }
 }
 
+template <typename T>
+bool BTreeLockableMap<T>::hasKey(const key_type& key) const {
+    auto readGuard = do_acquire_read_guard();
+    return _impl->find(key).valid();
```
toregge, line 185: That find method is only supposed to be used by the writer and is not protected by the read guard.

toregge (changes requested):

storage/src/vespa/storage/bucketdb/btree_lockable_map.hpp
baldersheim: That was what I suspected. new wiring added.

vekterli (commented):
This change seems to introduce some subtle new interactions between mutating operations and operations that change the structure of the bucket tree. Should probably discuss this before deciding on whether we should merge.

baldersheim:
Too risky...

PR #14807, opened 2020-10-11
Deadlock detection

hakonhall:
Just before Lock.acquire() is invoked, the locks within the process is queried
to see if a "deadlock" will occur: The current thread waiting to acquire lock
path P1, which is held by thread T1 waiting on acquiring a lock at path P2,
etc, until a thread is waiting for a lock already held by another thread in this chain.

This PR adds metrics to detect, and logging to debug deadlocks.  Eventually I hope to
put in place automatic deadlock resolution:  by throwing a timeout exception when detected,
but that may break tests.

Some tests are changed in this PR: Tests that take the lock but do not release it.  This breaks
deadlock detection since lock stats are global and the same VM runs later tests.  411aa57 broke on 
SD with this, but I was unable to reproduce it locally.  I'll take another stab at locating more
tests later.


mpolden (approved):
Nice! 👍

PR #14818, opened 2020-10-12
Rebalance nodes to exclusive switches

mpolden:
`SwitchRebalancer` drives the balancing using the same mechanism as the existing
`Rebalancer`.

Node movers may second-guess each other, but that appears to already be case for
`Rebalancer` and `SpareCapacityMaintainer`. Haven't thought of a good solution
for that yet.

@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
@@ -76,6 +76,15 @@ public class NodePrioritizer {
         this.isDocker = resources(requestedNodes) != null;
     }
 
+    /** Collects all node candidates for this application and returns them in the most-to-least preferred order */
+    List<NodeCandidate> collect(List<Node> surplusActiveNodes) {
+        addApplicationNodes();
+        addSurplusNodes(surplusActiveNodes);
+        addReadyNodes();
+        addNewDockerNodes();
+        return prioritize();
+    }
+
     /** Returns the list of nodes sorted by {@link NodeCandidate#compareTo(NodeCandidate)} */
     private List<NodeCandidate> prioritize() {
         // Group candidates by their cluster switch
@@ -109,7 +118,7 @@ public class NodePrioritizer {
      * Add nodes that have been previously reserved to the same application from
      * an earlier downsizing of a cluster
      */
-    void addSurplusNodes(List<Node> surplusNodes) {
+    private void addSurplusNodes(List<Node> surplusNodes) {
         for (Node node : surplusNodes) {
             NodeCandidate candidate = candidateFrom(node, true);
             if (!candidate.violatesSpares || isAllocatingForReplacement) {
@@ -119,7 +128,7 @@ public class NodePrioritizer {
     }
 
     /** Add a node on each docker host with enough capacity for the requested flavor  */
-    void addNewDockerNodes() {
+    private void addNewDockerNodes() {
         if ( ! isDocker) return;
 
         LockedNodeList candidates = allNodes
@@ -156,7 +165,7 @@ public class NodePrioritizer {
     }
 
     /** Add existing nodes allocated to the application */
-    void addApplicationNodes() {
+    private void addApplicationNodes() {
         EnumSet<Node.State> legalStates = EnumSet.of(Node.State.active, Node.State.inactive, Node.State.reserved);
         allNodes.asList().stream()
                 .filter(node -> node.type() == requestedNodes.type())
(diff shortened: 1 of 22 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -168,14 +171,38 @@ private NodeCandidate candidateFrom(Node node, boolean isSurplus) {
...
+        return unusedSwitch(host, node.allocation()
+                                      .map(allocation -> allocation.membership().cluster().id())
+                                      .orElseGet(clusterSpec::id));
+    }
+
+    /** Returns whether switch of host is unused by any existing candidates for given cluster */
+    private boolean unusedSwitch(Node host, ClusterSpec.Id cluster) {
+        if (host.switchHostname().isEmpty()) return true;
+        String switchHostname = host.switchHostname().get();
+        for (var candidate : nodes) {
```
freva, line 190: When we will try to add new nodes in `addNewDockerNodesOn()`, wont this lead to us only considering the node on the first host that we read from ZK (or whatever order the hosts in `allNodes` are)?

mpolden: How? Each node added is checked against all the (parents of) candidates already added.

bratseth:
> Node movers may second-guess each other

The actual decision about where to place a node is made by NodeCandidate.compareTo and NodeAllocation and is singular, so I think the worst that can happen from having many independent movers is that the movers fail to make progress as their chosen move keeps being rejected. This may happen for movers who care about the target (verifyTarget=true) when a different target is chosen due to some condition not taken into account by the mover, or when there are no alternatives available in the zone, for all movers.

The first condition (verifyTarget=true) should be fixed. It is more likely to happen the more conditions there are and the less changes there are to node allocations. I have it on my todo list to make the suggestions made by movers who care about the target probabilistic rather than deterministic (pick a random move with probability weighted by its goodness, with a cutoff where the goodness becomes lower than the cost of moving, rather than picking the best one). By now that is only the Rebalancer that care about the move target so I don't think it's that high priority.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
freva: Lets say we need to allocate 1 node, we have 2 hosts: `a` and `b`. `b` has massive skew, so allocation to `a` should be preferred. Both `a` and `b` are on the same switch, `s`.

Since we need to allocate 1 node, we start by creating a virtual reserved node on both `a` and `b`: `a1` and `b1`, as we create them, we check if `nodes` already has a node on `s` (initially, that is not the case). After we add the first of `a1` and `b1`, the second one will be deemed as not on exclusive switch, so skew is completely ignored.

mpolden: So given two hosts on the same switch, you want to prefer the candidate on the host with the least skew?

freva: The point is that while `a1` and `b1` are on the same switch, only 1 of them will be actually allocated anyway, so they cant be in conflict with each other

mpolden: > only 1 of them will be actually allocated anyway

The exclusive switch isn't a hard requirement though. If you ask for 2 nodes, you will allocate both `a1` and `b1` if no better candidate on an exclusive switch exists.

mpolden:
d89a5bc priorities nodes locally on each switch. PTAL.

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -191,9 +222,38 @@ private boolean canStillAllocateToParentOf(Node node) {
...
 
-    private static NodeResources resources(NodeSpec requestedNodes) {
-        if ( ! (requestedNodes instanceof NodeSpec.CountNodeSpec)) return null;
-        return requestedNodes.resources().get();
+    /** A cluster and its network switch */
+    private static class ClusterSwitch {
+
+        private static final ClusterSwitch unknown = new ClusterSwitch(null, null);
+
+        private final ClusterSpec.Id cluster;
```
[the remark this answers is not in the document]
mpolden: I'll look into it. 👍

PR #14825, opened 2020-10-12
No need for BucketInfo to be virtual just to make it printable.

baldersheim:
@vekterli PR

### DIFF
--- a/storageapi/src/vespa/storageapi/message/bucket.cpp
+++ b/storageapi/src/vespa/storageapi/message/bucket.cpp
@@ -548,7 +548,7 @@ NotifyBucketChangeCommand::print(std::ostream& out, bool verbose,
                                  const std::string& indent) const
 {
     out << "NotifyBucketChangeCommand(" << getBucketId() << ", ";
-    out << _info.toString();
+    out << _info;
     out << ")";
     if (verbose) {
         out << " : ";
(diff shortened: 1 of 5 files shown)


storageapi/src/vespa/storageapi/message/bucket.cpp
```diff
@@ -552,7 +548,7 @@ NotifyBucketChangeCommand::print(std::ostream& out, bool verbose,
                                  const std::string& indent) const
 {
     out << "NotifyBucketChangeCommand(" << getBucketId() << ", ";
-    _info.print(out, verbose, indent);
+    out << _info.toString();
```
vekterli, line 551: Could be simplified down to just `out << _info;` (unless we want stringification to be explicit)

vekterli (approved):
:+1: the number of pointless vtables that have historically been scattered around due to `Printable` and its friends makes me think that maybe we should remove the thing entirely.

storageapi/src/vespa/storageapi/message/bucket.cpp
baldersheim: Correct, will do

baldersheim: Fixed

PR #14836, opened 2020-10-13
Arnej/add generic map instruction rebase 1

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/tests/tensor/instruction_benchmark/instruction_benchmark.cpp
+++ b/eval/src/tests/tensor/instruction_benchmark/instruction_benchmark.cpp
@@ -27,6 +27,7 @@
 #include <vespa/eval/instruction/generic_join.h>
 #include <vespa/eval/instruction/generic_reduce.h>
 #include <vespa/eval/instruction/generic_rename.h>
+#include <vespa/eval/instruction/generic_map.h>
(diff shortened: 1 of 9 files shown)


eval/src/tests/tensor/instruction_benchmark/instruction_benchmark.cpp
```diff
@@ -612,6 +635,28 @@ TEST(MergeBench, mixed_merge) {
 
 //-----------------------------------------------------------------------------
 
+TEST(MapBench, dense_map) {
```
havardpe, line 638: please add case for simple number

eval/src/vespa/eval/instruction/generic_map.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace vespalib::eval::instruction {
+
+using State = InterpretedFunction::State;
+using Instruction = InterpretedFunction::Instruction;
+
+namespace {
+
+uint64_t to_param(map_fun_t value) { return (uint64_t)value; }
+map_fun_t to_map_fun(uint64_t param) { return (map_fun_t)param; }
```
havardpe, line 18: why not use wrap/unwrap?

havardpe (approved):

PR #14841, opened 2020-10-13
Jonmv/async doc v1 take2

jonmv:
@bjorncs please review. 

This does not compile, as tests are broken. I'll start working on that now.

I already thought of one missing thing: Jetty threads don't currently help in sending enqueued operations.


bjorncs (approved):

jonmv:
PTAL :)

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -406,18 +409,14 @@ private void enqueueAndDispatch(HttpRequest request, ResponseHandler handler, Su
             return;
         }
         Operation operation = Operation.lazilyParsed(request, handler, operationParser);
-        if (enqueued.get() == 0 && operation.dispatch()) // Bypass queue if it is empty.
+        if (enqueued.get() == 1 && operation.dispatch()) // Bypass queue if it is empty.
```
bjorncs, line 412: `enqueued` might be positive even when the actual queue is empty (when multiple threads executes `enqueueAndDispatch` simultaneously.

vespaclient-container-plugin/src/main/resources/configdefinitions/document-operation-executor.def
```diff
@@ -1,15 +1,15 @@
...
+resendDelayMillis     int default=10
 
 # Bound on number of document operations to keep in retry queue — further operations are rejected
 maxThrottled          int default=1000
 
+# Whether to perform dispatch from Jetty threads
+doDispatchWithEnqueue bool default=true
+
+# Number of workers in the resender thread pool
+resenderCount         int default=1
```
bjorncs, line 14: We will probably want to control resender threads from `ContainerThreadpool` later on.

bjorncs (approved):

vespaclient-container-plugin/src/main/resources/configdefinitions/document-operation-executor.def
jonmv: Yes, this isn't even used now >_< Let's clean up when we decide on how we do it.

bjorncs (commented):
👍

PR #14844, opened 2020-10-13
Do internal restart if provisioner is set

freva:


config-provisioning/src/main/java/com/yahoo/config/provision/HostFilter.java
```diff
@@ -27,15 +27,11 @@ private HostFilter(Set<String> hostnames,
...
-
-        this.hostnames = hostnames;
-        this.flavors = flavors;
-        this.clusterTypes = clusterTypes;
-        this.clusterIds = clusterIds;
+        this.hostnames = Objects.requireNonNull(hostnames, "Hostnames cannot be null, use an empty list");
+        this.flavors = Objects.requireNonNull(flavors, "Flavors cannot be null, use an empty list");
+        this.clusterTypes = Objects.requireNonNull(clusterTypes, "clusterTypes cannot be null, use an empty list");
+        this.clusterIds = Objects.requireNonNull(clusterIds, "clusterIds cannot be null, use an empty list");
+        System.out.println(hostnames);
```
hmusum, line 34: Remove debug line?

hmusum (approved):
One comment, otherwise LGTM

config-provisioning/src/main/java/com/yahoo/config/provision/HostFilter.java
freva, line 34:
```suggestion
```

freva: Thanks :+1:

PR #14848, opened 2020-10-14
Use std::mutex/std::condition instead of vespalib::Monitor.

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/visiting/visitormanager.cpp
+++ b/storage/src/vespa/storage/visiting/visitormanager.cpp
@@ -28,6 +28,7 @@ VisitorManager::VisitorManager(const config::ConfigUri & configUri,
       _visitorThread(),
       _visitorMessages(),
       _visitorLock(),
+      _visitorCond(),
       _visitorCounter(0),
       _configFetcher(configUri.getContext()),
       _metrics(new VisitorMetrics),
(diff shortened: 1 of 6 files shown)


storage/src/vespa/storage/visiting/visitormanager.cpp
```diff
@@ -208,39 +202,33 @@ VisitorManager::run(framework::ThreadHandle& thread)
...
                 thread.registerTick(framework::WAIT_CYCLE);
             } else {
-                uint64_t timediff = (_visitorQueue.tbegin()->_time
-                                     - currentTime.getTime())
-                                            / 1000000;
+                uint64_t timediff = (_visitorQueue.tbegin()->_time- currentTime.getTime()) / 1000000;
                 timediff = std::min(timediff, uint64_t(1000));
                 if (timediff > 0) {
-                    waiter.wait(timediff);
+                    _visitorCond.wait_for(waiter, std::chrono::milliseconds(timediff));
```
vekterli, line 231: I'm a bit puzzled by the original code here. Since the queued timestamp and `currentTime` are both in microseconds, dividing by 1,000,000 implies that `timediff` is in discrete seconds and not in milliseconds. Expected wait period will therefore be off by a few orders of magnitude... I don't think this has mattered in practice, since it just means we'll wake up the thread more frequently than strictly required.

vekterli (approved):
:+1:

storage/src/vespa/storage/visiting/visitormanager.cpp
baldersheim: Then we are 2. I think it is overdue to replace replace the framework times with vespalib::duration. I think that was one of things we postponed last time we tried to unify time.

PR #14849, opened 2020-10-14
Drain all messages on flush to avoid the odd signaling on every message.

baldersheim:
@vekterli and @geirst PR

### DIFF
--- a/storage/src/vespa/storage/persistence/persistencethread.cpp
+++ b/storage/src/vespa/storage/persistence/persistencethread.cpp
@@ -956,7 +956,7 @@ PersistenceThread::run(framework::ThreadHandle& thread)
 void
 PersistenceThread::flush()
 {
-    while (_env._fileStorHandler.getQueueSize() != 0) {
+    while (_env._fileStorHandler.getQueueSize(_stripeId) != 0) {
         std::this_thread::sleep_for(1ms);
     }
 }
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/persistence/persistencethread.cpp
```diff
@@ -950,22 +949,15 @@ PersistenceThread::run(framework::ThreadHandle& thread)
...
-    flushMonitorGuard.broadcast();
 }
 
 void
 PersistenceThread::flush()
 {
-    vespalib::MonitorGuard flushMonitorGuard(_flushMonitor);
-    if (!_closed) {
-        flushMonitorGuard.wait();
+    while (_env._fileStorHandler.getQueueSize() != 0) {
```
vekterli, line 959: Could potentially just check the queue size for the thread's stripe, but `getQueueSize()` is a superset of that stripe so shouldn't matter in practice.

vekterli (approved):
:+1: `flush()` seems to only be called at shutdown, so no point in having extra overhead per message just to be more "efficient" at that

geirst (approved):
👍

storage/src/vespa/storage/persistence/persistencethread.cpp
baldersheim: I will add an api for that after disk concept is removed.

PR #14851, opened 2020-10-14
increase maxAllowedFailures to 50

bjormel:


smorgrav (approved):
LGTM, but a comment on why increase this would be good for historical purposes.

PR #14852, opened 2020-10-14
Arnej/some eval additions

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/eval/test/value_compare.cpp
+++ b/eval/src/vespa/eval/eval/test/value_compare.cpp
@@ -1,20 +1,33 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "value_compare.h"
-#include <vespa/eval/eval/engine_or_factory.h>
+#include <vespa/eval/eval/tensor.h>
+#include <vespa/eval/eval/tensor_engine.h>
+#include <vespa/eval/eval/value_codec.h>
 
 namespace vespalib::eval {
 
+namespace {
+
+TensorSpec get_spec_from(const Value &value) {
+    auto tensor = dynamic_cast<const vespalib::eval::Tensor *>(&value);
+    if (tensor) {
+        return tensor->engine().to_spec(value);
+    } else {
+        return spec_from_value(value);
+    }
+}
+
+} // namespace <unnamed>
+
 bool operator==(const Value &lhs, const Value &rhs)
 {
-    auto engine = EngineOrFactory::get();
-    return engine.to_spec(lhs) == engine.to_spec(rhs);
+    return get_spec_from(lhs) == get_spec_from(rhs);
 }
 
 std::ostream &operator<<(std::ostream &out, const Value &value)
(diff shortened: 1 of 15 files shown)


eval/src/vespa/eval/eval/test/value_compare.cpp
```diff
@@ -0,0 +1,22 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "value_compare.h"
+#include <vespa/eval/eval/engine_or_factory.h>
+
+namespace vespalib::eval {
+
+bool operator==(const Value &lhs, const Value &rhs)
+{
+    auto engine = EngineOrFactory::get();
```
havardpe, line 10: we should not use EngineOrFactory::get in the eval library.

here we should keep and open mind about implementation, everywhere else, we should use EngineOrFactory::get to select the implementation.

This would not be useful for unit tests, since comparing values created by the SimpleTensorEngine would result in problems, since DefaultTensorEngine is the default tensor implementation.

eval/src/vespa/eval/eval/test/value_compare.cpp
```diff
@@ -0,0 +1,22 @@
...
+    return engine.to_spec(lhs) == engine.to_spec(rhs);
+}
+
+std::ostream &operator<<(std::ostream &out, const Value &value)
+{
+    auto engine = EngineOrFactory::get();
```
havardpe, line 16: same as above

havardpe:
I suggest you try to dynamic cast the value to eval::Tensor, which would give you access to the engine associated with the value. If the cast fails, you can use the value_coded directly, since everything that is not a tensor will support the new API needed to create a TensorSpec without needing a factory.

eval/src/vespa/eval/eval/test/value_compare.cpp
```diff
@@ -0,0 +1,21 @@
...
+namespace vespalib::eval {
+
+bool operator==(const Value &lhs, const Value &rhs)
+{
+    return TensorSpec::from_value(lhs) == TensorSpec::from_value(rhs);
+}
+
+std::ostream &operator<<(std::ostream &out, const Value &value)
+{
+    out << TensorSpec::from_value(value);
```
havardpe, line 17: I think return here is better for full forwarding

eval/src/vespa/eval/eval/test/value_compare.h
```diff
@@ -0,0 +1,13 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include <vespa/eval/eval/value.h>
+
+namespace vespalib::eval {
+
+bool operator==(const Value &lhs, const Value &rhs);
+std::ostream &operator<<(std::ostream &out, const Value &tensor);
```
havardpe, line 10: tensor -> value

havardpe (approved):

havardpe (approved):

PR #14867, opened 2020-10-14
vespalib::Monitor -> std:.mutex/std::condition_variable

baldersheim:
@havardpe PR
A small one

### DIFF
--- a/vbench/src/vbench/core/handler_thread.hpp
+++ b/vbench/src/vbench/core/handler_thread.hpp
@@ -7,9 +7,9 @@ void
 HandlerThread<T>::run()
 {
     for (;;) {
-        vespalib::MonitorGuard guard(_monitor);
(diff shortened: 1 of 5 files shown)


vbench/src/vbench/core/handler_thread.hpp
```diff
@@ -58,10 +59,10 @@ void
 HandlerThread<T>::join()
 {
     {
-        vespalib::MonitorGuard guard(_monitor);
+        std::lock_guard guard(_lock);
         _done = true;
-        guard.signal();
     }
+    _cond.notify_one();
```
havardpe, line 65: prefer signal while holding lock unless very special conditions

baldersheim: Done.

PR #14880, opened 2020-10-14
avoid using code that should be internal to TypedBinaryFormat

arnej27959:

@baldersheim please review.
Does this code path have a test somewhere? (unit test or system test)
@havardpe please take a look also


baldersheim:
I assume this is OK.
As to testing I am uncertain.
The tests that I know of are in prod_features.cpp

baldersheim (approved):

PR #14899, opened 2020-10-15
Avoid copy construction

baldersheim:
@geirst and @vekterli PR

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -651,8 +651,7 @@ FileStorHandlerImpl::remapQueueNoLock(Disk& from, const RemapInfo& source,
     // Find all the messages for the given bucket.
     for (BucketIdx::iterator i = range.first; i != range.second; ++i) {
         assert(i->_bucket == source.bucket);
-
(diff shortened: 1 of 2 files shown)


baldersheim:
Ping

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -651,8 +651,7 @@ FileStorHandlerImpl::remapQueueNoLock(Disk& from, const RemapInfo& source,
     // Find all the messages for the given bucket.
     for (BucketIdx::iterator i = range.first; i != range.second; ++i) {
         assert(i->_bucket == source.bucket);
-
-        entriesFound.push_back(std::move(*i));
+        entriesFound.push_back(std::move(i.get_node()->value()));
```
vekterli, line 654: This seems to be accessing Multi-Index internal structures and is likely not safe to do, as there may be other index structures using the information within. Removing it may corrupt these in interesting ways.

Looking at https://www.boost.org/doc/libs/1_74_0/boost/multi_index/detail/bidir_node_iterator.hpp, it says 
```c++
/* get_node is not to be used by the user */
```

vekterli (changes requested):
I'm not confident this is safe to do with the current queue data structure.

baldersheim:
Not possible....

PR #14903, opened 2020-10-15
There will always be a sequencer.

baldersheim:
@vekterli or @geirst PR

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
@@ -86,7 +86,7 @@ FileStorManager::print(std::ostream& out, bool verbose, const std::string& inden
 namespace {
 
 uint32_t computeNumResponseThreads(int configured) {
-    return (configured < 0) ? std::max(1u, std::thread::hardware_concurrency()/4) : configured;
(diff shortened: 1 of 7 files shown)


storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
```diff
@@ -86,7 +86,7 @@ FileStorManager::print(std::ostream& out, bool verbose, const std::string& inden
 namespace {
 
 uint32_t computeNumResponseThreads(int configured) {
-    return (configured < 0) ? std::max(1u, std::thread::hardware_concurrency()/4) : configured;
+    return (configured <= 0) ? std::max(1u, std::thread::hardware_concurrency()/4) : configured;
```
geirst, line 89: Intentional change?

geirst (approved):
👍

storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
baldersheim: Yes, I removed the possibility to specify 0 response threads which would give synchronous behaviour. It has been async by default for at least 6 months, and nobody has had any override.

PR #14907, opened 2020-10-15
Try to avoid race when creating directory

hmusum:
If e.g. several containers try to create this directory at the same time
there will be unnecessary warnings in log


arnej27959 (approved):
this change in itself looks fine, but won't there be other problems if several containers log to the same directory?

arnej27959:
forget my last question - now I remember after looking at the rest of the code.
Multiple containers etc are already handled by using append semantics when writing to the meta file.

PR #14916, opened 2020-10-16
reduce empty mixed to dense must add cells with default value

havardpe:

@arnej27959 please review
@geirst FYI

This should fix the reduce crash in the performance test. However, reducing an empty mixed tensor to a dense tensor with default cell values does not seem like a relevant use-case for real life (should there not be some values in the tensors?)


arnej27959 (approved):
👍

PR #14917, opened 2020-10-16
Greatly simplify bucket DB persistence provider bootstrap procedure

vekterli:
@geirst please review

The legacy bucket DB initialization logic was designed for the case
where bucket information was spread across potentially millions of
files residing on spinning rust drives. It was therefore async and
running in parallel with client operations, adding much complexity
in order to deal with a myriad of concurrency edge cases.

Replace this with a very simple, synchronous init method that expects the
provider to have the required information readily and cheaply available.
This effectively removes the concept of a node's "initializing" state,
moving directly from reported state Down to Up.

Even though a node still technically starts up in Initializing state,
we never end up reporting this to the Cluster Controller as the DB init
completes before the RPC server stack is set up.

Legacy bucket DB initializer code will be removed in a separate pass.

Also simplify bucket DB interface contract for mutating iteration,
indicating that it is done in an unspecified order.

### DIFF
--- a/persistence/src/vespa/persistence/dummyimpl/dummypersistence.cpp
+++ b/persistence/src/vespa/persistence/dummyimpl/dummypersistence.cpp
@@ -359,6 +359,21 @@ DummyPersistence::setModifiedBuckets(const BucketIdListResult::List& buckets)
     _modifiedBuckets = buckets;
 }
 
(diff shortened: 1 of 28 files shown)


geirst (approved):
👍

PR #14918, opened 2020-10-16
Increase timeout for ExecutorService

olaaun:
Why was it changed from 1 min to 1 sec?


bjorncs (approved):
Ouch, that was an accident

PR #14936, opened 2020-10-17
NodeList cleanup

mpolden:
@freva


freva:
:+1: on last 4 commits, rest is from #14818

PR #14938, opened 2020-10-19
Merge LocalSession and RemoteSession

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -320,28 +276,21 @@ void activate(RemoteSession session) {
...
-        if (localSession != null) {
-            // TODO: Change log level to FINE when debugging is finished
-            log.log(Level.INFO, () -> localSession.logPre() + "Deleting local session " + sessionId);
-            deleteLocalSession(localSession);
-        }
+    public void delete(Session session) {
+        long sessionId = session.getSessionId();
+        deleteSessionFromZooKeeper(session);
+        sessionCache.remove(sessionId);
+        deleteLocalSession(session);
```
jonmv, line 283: Perhaps rename this method as well, to be in line with others? `deleteSessionFromFileSystem`?

jonmv (approved):
LGTM!

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Yeah, there is more renaming, collapsing and simplification that can be done. Will follow up in a later PR

PR #14940, opened 2020-10-19
We're replacing when we're *not* fulfilled

bratseth:
Looks like we need a unit test where we're replacing and we only have a spare, but I don't have time for it right now.


freva (approved):
Doh! :man_facepalming:

PR #14941, opened 2020-10-19
Take correct lock for each node when retiring

mpolden:
@hakonhall

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
@@ -766,6 +766,11 @@ public class NodeRepository extends AbstractComponent {
         });
     }
 
+    /** Retire nodes matching given filter */
+    public List<Node> retire(NodeFilter filter, Agent agent, Instant instant) {
+        return performOn(filter, (node, lock) -> write(node.withWantToRetire(true, agent, instant), lock));
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -809,13 +814,19 @@ else if (!node.type().isHost()) { // removing a child node
         // perform operation while holding locks
         List<Node> resultingNodes = new ArrayList<>();
         try (Mutex lock = lockUnallocated()) {
-            for (Node node : unallocatedNodes)
-                resultingNodes.add(action.apply(node, lock));
+            for (Node node : unallocatedNodes) {
+                Optional<Node> currentNode = db.readNode(node.hostname()); // Re-read while holding lock
```
hakonhall, line 818: This is an improvement.

I'm a bit sceptical of the method API: Usually, some condition has led to the caller wanting to set this node to dirty.  I would guess that whole condition needs to be reevaluated under the unallocated lock?  More than just the presence of the node may have changed between testing the condition and coming to this point where the lock is held.

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
mpolden: Agree, the API is not ideal.

PR #14945, opened 2020-10-19
Arnej/add wrapped simple value 2

arnej27959:

@havardpe please review


eval/src/vespa/eval/eval/simple_value_engine.h
```diff
@@ -0,0 +1,37 @@
...
+#include "tensor_engine.h"
+
+namespace vespalib {
+namespace eval {
+
+/**
+ * This is a TensorEngine implementation for the SimpleValue
+ * reference implementation.
+ **/
+class SimpleValueEngine : public TensorEngine
```
havardpe, line 14: I do not think we actually need this when we also have the immediate generic implementations wired through the EngineOrFactory class.

Where we would use SimpleValueEngine, we can just use EngineOrFactory(SimpleValueBuilderFactory::get()) instead.

havardpe (approved):

PR #14946, opened 2020-10-19
Don't perform unsafe std::move of functor when it may be used again

vekterli:
@geirst please review

All callsites explicitly wrapped the functor in `std::ref`, which
is why this hasn't been a problem in practice (likely no-op move).

Also remove unused DB iteration methods.


baldersheim (approved):

geirst (commented):
👍

PR #14957, opened 2020-10-19
Remove allow host sharing

freva:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/Rebalancer.java
```diff
@@ -35,7 +35,7 @@ public Rebalancer(Deployer deployer,
     @Override
     protected boolean maintain() {
         boolean success = true;
-        if ( ! nodeRepository().zone().getCloud().allowHostSharing()) return success; // Rebalancing not necessary
+        if (nodeRepository().zone().getCloud().dynamicProvisioning()) return success; // Rebalancing not necessary
```
hakonhall, line 38: "not necessary" is not correct - we hope to enable rebalancing even with dynamic provisioning, but it's a todo.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/SpareCapacityMaintainer.java
```diff
@@ -68,7 +68,8 @@ public SpareCapacityMaintainer(Deployer deployer,
     @Override
     protected boolean maintain() {
         boolean success = true;
-        if ( ! nodeRepository().zone().getCloud().allowHostSharing()) return success;
+        // Don't need to maintain spare capacity in dynamically provisioned zones; can provision more on demand.
```
hakonhall, line 71: We'd like to maintain spare capacity in dynamically provisioned zones too - but it will depend on flags.

hakonhall (approved):

PR #14965, opened 2020-10-20
Balder/add thread local access to persistencehandler

baldersheim:
@geirst and @vekterli PR

### DIFF
--- a/storage/src/vespa/storage/config/stor-communicationmanager.def
+++ b/storage/src/vespa/storage/config/stor-communicationmanager.def
@@ -68,6 +68,8 @@ rpc.num_network_threads int default=1
 ## The bucket id associated with a message is used to select the RPC target.
 ## This ensures the same RPC target is used for all messages to the same bucket to the same node,
 ## and the RPC target itself handles sequencing of these messages.
+## NB !! It is vital that this number is kept in sync with stor-filestor:num_network_threads.
+## Only skilled vespa core devolpers should touch this.
 rpc.num_targets_per_node int default=1
 
 # Minimum size of packets to compress (0 means no compression)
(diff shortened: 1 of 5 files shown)


storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
```diff
@@ -127,6 +130,14 @@ FileStorManager::createRegisteredHandler(ServiceLayerComponent & component)
                                                  *_bucketOwnershipNotifier, *_metrics->disks[0]->threads[index]));
     return *_persistenceHandlers.back();
 }
+
+PersistenceHandler &
+FileStorManager::getThreadLocalHandler() {
+    if (_G_threadLocalHandler == nullptr) {
+        _G_threadLocalHandler = & createRegisteredHandler(_component);
```
vekterli, line 137: How do we ensure the `_persistenceHandlers.size() < _metrics->disks[0]->threads.size()` invariant holds when an "arbitrary" number of handlers can be registered?

vekterli (commented):
One question regarding metric vs handler cardinality that I'd like to resolve first

storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
baldersheim: The assumption is that this code will only be invoked from network threads and response threads.
Those are accounted for in the last commit.

vekterli (approved):
:+1:

storage/src/vespa/storage/config/stor-communicationmanager.def
```diff
@@ -68,6 +68,8 @@ rpc.num_network_threads int default=1
 ## The bucket id associated with a message is used to select the RPC target.
 ## This ensures the same RPC target is used for all messages to the same bucket to the same node,
 ## and the RPC target itself handles sequencing of these messages.
+## NB !! It is vital that this number is kept in sync with stor-filestor:num_network_threads.
+## Only skilled vespa core devolpers should touch this.
```
geirst, line 72: Typo: devolpers -> developers

geirst (commented):
👍

storage/src/vespa/storage/config/stor-communicationmanager.def
baldersheim: Fixed in https://github.com/vespa-engine/vespa/pull/14970

PR #14967, opened 2020-10-20
Reapply "Merge LocalSession and RemoteSession"

hmusum:
Please review only. Unable to test on a multinode setup, so will merge later.


jonmv (approved):
Hehe, seems like an important fix ...

PR #14969, opened 2020-10-20
Let zone decide container registry

mpolden:
Depends on internal PR.

@tokle

FYI @freva


config-provisioning/src/main/resources/configdefinitions/config.provisioning.node-repository.def
```diff
@@ -1,9 +1,11 @@
 # Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 namespace=config.provisioning
 
-# Docker image to use in REST API responses. This must be a fully qualified name, including registry, but excluding
-# version. Example: my-docker-registry.domain.tld:8080/dist/vespa
-dockerImage string default="dummyImage"
+# The container registry to use, excluding the repository. Example: my-registry.example.com:9999
+containerRegistry string default=""
```
freva, line 5: Does it make sense to pass these as separate? I think we should just pass 1 string and let `DockerImage::fromString` figure it out

config-provisioning/src/main/java/com/yahoo/config/provision/DockerImage.java
```diff
@@ -55,25 +76,36 @@ public boolean equals(Object o) {
...
-        if (name.isEmpty()) return EMPTY;
+    public static DockerImage from(String registry, String repository) {
+        return new DockerImage(registry, repository, Optional.empty());
+    }
+
+    public static DockerImage fromString(String s) {
+        if (s.isEmpty()) return EMPTY;
+
+        int firstPathSeparator = s.indexOf('/');
+        if (firstPathSeparator < 0) throw new IllegalArgumentException("Missing path separator in '" + s + "'");
```
freva, line 97: I think we should allow specifying docker image without registry

freva (commented):
:1st_place_medal: on docker -> container renaming! :100:

config-provisioning/src/main/java/com/yahoo/config/provision/DockerImage.java
mpolden: Explicit is better I think. Won't the default depend on the container engine? I believe Docker uses hub.docker.com without registry, but that might not be the case for e.g. Podman?

mpolden:
PTAL.

tokle (approved):

PR #14971, opened 2020-10-20
Keep retired nodes on bootstrap deployment

freva:
When replacing a node, we would have a node surplus while the old node was retiring. When the application was redeployed during config server bootstrap, retired nodes would be dropped (moved to `inactive`) before retirement was complete since we did not consider nodes for retirement in bootstrap (`NodeSpec::considerRetiring`).


bratseth (approved):
👍

PR #14973, opened 2020-10-20
optimize sparse only merge

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/generic_merge.cpp
+++ b/eval/src/vespa/eval/instruction/generic_merge.cpp
@@ -2,10 +2,12 @@
 
 #include "generic_merge.h"
 #include <vespa/eval/eval/inline_operation.h>
+#include <vespa/eval/eval/fast_value.hpp>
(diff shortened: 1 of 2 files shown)


eval/src/vespa/eval/instruction/generic_merge.cpp
```diff
@@ -140,6 +172,10 @@ GenericMerge::make_instruction(const ValueType &lhs_type, const ValueType &rhs_t
                                const ValueBuilderFactory &factory, Stash &stash)
 {
     const auto &param = stash.create<MergeParam>(lhs_type, rhs_type, function, factory);
+    if (param.dense_subspace_size == 1) {
```
havardpe, line 175: I would prefer sending param to the Select function and having that choose the implementation (like for join)

arnej27959:
@havardpe PTAL

havardpe (approved):
looks good, but we should extend the generic merge unit test to also test with fast values.

PR #14978, opened 2020-10-20
Bjorncs/jetty threadpool

bjorncs:
FYI @baldersheim @tokle

### DIFF
--- a/jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
+++ b/jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
@@ -1,7 +1,7 @@
 // Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.jdisc.http.server.jetty;
 
-import com.yahoo.jdisc.Response;
+import com.yahoo.concurrent.DaemonThreadFactory;
 import com.yahoo.jdisc.http.ConnectorConfig;
 import com.yahoo.security.SslContextBuilder;
 import com.yahoo.security.tls.TransportSecurityOptions;
@@ -15,6 +15,7 @@ import org.apache.http.client.methods.HttpGet;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.util.EntityUtils;
 import org.eclipse.jetty.server.DetectorConnectionFactory;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.SslConnectionFactory;
@@ -22,8 +23,10 @@ import org.eclipse.jetty.server.handler.HandlerWrapper;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 
 import javax.net.ssl.SSLContext;
+import javax.servlet.AsyncContext;
 import javax.servlet.ServletException;
 import javax.servlet.ServletOutputStream;
+import javax.servlet.WriteListener;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
```diff
@@ -49,6 +53,7 @@
 
     private static final String HEALTH_CHECK_PATH = "/status.html";
 
+    private final Executor executor = Executors.newSingleThreadExecutor(new DaemonThreadFactory("health-check-proxy-client"));
```
jonmv, line 56:
```suggestion
    private final Executor executor = Executors.newSingleThreadExecutor(new DaemonThreadFactory("health-check-proxy-client-"));
```

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
```diff
@@ -124,21 +115,86 @@ protected void doStop() throws Exception {
...
+        final AsyncContext asyncContext;
+        final ProxyTarget target;
+        final HttpServletResponse servletResponse;
+        final ServletOutputStream output;
+
+        ProxyRequestTask(AsyncContext asyncContext, ProxyTarget target, HttpServletResponse servletResponse, ServletOutputStream output) {
+            this.asyncContext = asyncContext;
+            this.target = target;
+            this.servletResponse = servletResponse;
+            this.output = output;
```
jonmv, line 129: Super-nitty: this field could just be a local variable in the `run` method.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
```diff
@@ -124,21 +115,86 @@ protected void doStop() throws Exception {
...
+                servletResponse.setHeader("Content-Type", statusResponse.contentType);
+            }
+            output.setWriteListener(new WriteListener() {
+                @Override
+                public void onWritePossible() throws IOException {
+                    if (output.isReady()) {
+                        if (statusResponse.content != null) {
+                            output.write(statusResponse.content);
+                        }
+                        asyncContext.complete();
```
jonmv, line 146: I don't like the spec here. _How much_ can you write without blocking? Generally, this should be a `while` loop, but since you only do a single write here, and then close, I guess this is fine.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
```diff
@@ -124,21 +115,86 @@ protected void doStop() throws Exception {
...
-            return client()
-                    .execute(new HttpGet("https://localhost:" + port + HEALTH_CHECK_PATH));
+        StatusResponse requestStatusHtml() {
+            StatusResponse response = lastResponse;
+            if (response != null && !response.isExpired()) {
+                return response;
+            }
+            StatusResponse statusResponse = getStatusResponse();
+            lastResponse = statusResponse;
+            return statusResponse;
```
jonmv, line 179:
```suggestion
            return lastResponse = getStatusResponse();
```
I find this easier to read, but personal preference I guess :)

jonmv (approved):
LGTM!

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
bjorncs: `getOuputStream()` may throw `IOException`, so I want to do it in the `handle` method.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
bjorncs: Yes, it's a bad interface. The servlet has no clue on how much data to produce. This method typically implemented as a `while` loop if the servlet had multiple byte buffers to write.

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
bjorncs: Fixed

jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HealthCheckProxyHandler.java
bjorncs: Done :)

PR #14980, opened 2020-10-20
Remove always-acquire-provision-lock flag

mpolden:
Review, but hold merge.

@hmusum

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
@@ -121,21 +121,6 @@ public class NodeRepositoryProvisioner implements Provisioner {
     }
 
     @Override
-    // TODO(mpolden): Remove
-    public void activate(NestedTransaction transaction, ApplicationId application, Collection<HostSpec> hosts) {
-        try (var lock = lock(application)) {
-            activate(hosts, new ActivationContext(0), new ApplicationTransaction(lock, transaction));
-        }
-    }
-
-    @Override
-    // TODO: Remove after November 2020
-    public void activate(NestedTransaction transaction, Collection<HostSpec> hosts, ProvisionLock lock) {
-        validate(hosts);
-        activator.activate(hosts, 0, new ApplicationTransaction(lock, transaction));
-    }
-
-    @Override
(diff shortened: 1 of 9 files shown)


hmusum (approved):

mpolden:
@bratseth PTAL

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
```diff
@@ -120,21 +120,6 @@ public NodeRepositoryProvisioner(NodeRepository nodeRepository, Zone zone,
...
-    @Override
-    // TODO(mpolden): Remove
-    public void activate(NestedTransaction transaction, ApplicationId application, Collection<HostSpec> hosts) {
-        try (var lock = lock(application)) {
-            activate(hosts, new ActivationContext(0), new ApplicationTransaction(lock, transaction));
-        }
-    }
-
-    @Override
-    // TODO: Remove after November 2020
```
bratseth, line 132: I don't think this one can be removed yet?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
```diff
@@ -146,20 +131,6 @@ public void restart(ApplicationId application, HostFilter filter) {
...
-    // TODO(mpolden): Remove
-    @Override
-    public void remove(NestedTransaction transaction, ApplicationId application) {
-        try (var lock = lock(application)) {
-            remove(new ApplicationTransaction(lock, transaction));
-        }
-    }
-
-    // TODO: Remove after November 2020
-    @Override
```
bratseth, line 158: I don't think this one can be removed yet?

bratseth (approved):
I think you should add back 2 methods, otherwise LGTM!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
mpolden: None of these are called from the config-model that I can tell. It uses `HostProvisioner` instead, which only has `prepare`.

bratseth: Ah, yes you're right.

bratseth (approved):

PR #14982, opened 2020-10-20
Fix forward declarations.

toregge:
@baldersheim : please review


baldersheim (approved):
Is there any warning we can enable for gcc to catch these so you do not have to be a watchdog ?

baldersheim:
So waiting for devtoolset-10 then :)

toregge:
More like gcc-toolset-10 on CentOS 8 / RHEL 8.

PR #14984, opened 2020-10-20
Add generation counter to StorageComponent to detect if repos have ch…

baldersheim:
…anged and need reload.

Happy path where you do not need any reload is a single atomic relaxed load as opposed to at least 3 atomic locked instructions.

@vekterli and @toregge PR
@vekterli if you think this is a feasible way I will add tests and complete the PR
This should significantly reduce cost for get() and test-and-set operations.

### DIFF
--- a/storage/src/tests/persistence/processalltest.cpp
+++ b/storage/src/tests/persistence/processalltest.cpp
@@ -3,9 +3,9 @@
 #include <vespa/document/base/testdocman.h>
 #include <vespa/storage/persistence/processallhandler.h>
 #include <vespa/storage/persistence/messages.h>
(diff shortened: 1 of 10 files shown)


toregge (approved):
Consider _type_repo_generation and _last_type_repo_generation instead of _generation and _lastGeneration.

vekterli (approved):
:+1:

vekterli:
Eventually we very much want to move away from the "the world changes under your feet" model of reconfiguration, but this is a good stop-gap before then.

baldersheim:
yes, it is a step on the way to a better world.
@toregge I kept a general name as I intend to also use it for some of the other synchronized members sometime in a not too distant future.

PR #14985, opened 2020-10-20
Add delete operators for PackedMappings and PackedMixedTensor.

toregge:
@arnej27959 : please review


arnej27959 (approved):
this looks fine to me.  My understanding that this is what would happen by default, but probably indirectly via `::delete(void * ptr,  size_t) ` and AddressSanitizer hooks the latter to check that the size matches what was allocated with new.

PR #14988, opened 2020-10-21
make generic reduce faster

havardpe:
* use faster map implementation
* use multiple aggregators (invert dense loops)
* handle reducing all cells separately
* extend nested loop benchmarking

@arnej27959 please review

sparse cases seem ok for now with good room for improvements (inline for fast value, use pre-calculated label hashes and avoid extra copy of result for aggregators that are op2_t)

we might need a better strategy for dense subspace reduce. This can be solved with specific optimizers, but a faster generic solution would be preferable. (this will not be done as part of this PR)

### DIFF
--- a/eval/CMakeLists.txt
+++ b/eval/CMakeLists.txt
@@ -12,6 +12,7 @@ vespa_define_module(
     TESTS
     src/tests/ann
     src/tests/eval/aggr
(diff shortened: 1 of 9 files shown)


arnej27959 (approved):
looks good to me

PR #14991, opened 2020-10-21
possible utility to recognize object implementation

arnej27959:

@havardpe does this look somewhat like you envisioned?


havardpe (approved):
yes, I like this form

we should consider adding another level with either a function or a deduction guide to support generic types for the pair of objects and also have seamless support for 1,2,3 values to be checked and converted.

havardpe (approved):

PR #15000, opened 2020-10-22
Update with new thin pool sizes

hmusum:
Needs to be merged with changes in internal repo

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
@@ -1,4 +1,4 @@
-// Copyright 2019 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.provisioning;
 
 import com.yahoo.config.provision.ClusterSpec;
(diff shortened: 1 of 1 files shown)


freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
```diff
@@ -72,26 +72,27 @@ private double minAdvertisedMemoryGb(ClusterSpec.Type clusterType) {
...
-        if (requested.storageType() == NodeResources.StorageType.local && zone().getCloud().dynamicProvisioning()) {
+        return minRealDiskGb() + getThinPoolSize(requested.storageType());
+    }
+
+    // TODO: Calculate thin pool size instead of hardcoding
+    private long getThinPoolSize(NodeResources.StorageType storageType) {
+        if (storageType == NodeResources.StorageType.local && zone().getCloud().dynamicProvisioning()) {
             if (zone().system() == SystemName.Public)
-                return 10 + minRealDiskGb();
+                return 12;
```
hakonhall, line 82: Will this code be exercised for currently provisioned hosts too, and that are using the old thin pool sizes?  How does that work?  AFAIK, only if we're currently enlarging (rounding up) the node resources will `prepare`  now instead require a larger disk, which may involve... migration I guess?

freva: If someone specified less resources than what is now legal, their deployment will fail.
If someone specified resource range and they are nearly not utilizing anything, then they will be scaled to minimum - which is now increased, so they will be migrated. The rest should not be changed.

PR #15004, opened 2020-10-22
Wait for DNS to resolve for dynamically provisioned hosts

mpolden:
@freva

FYI @oyving


freva (approved):
Great cleanup in tests! :clap:

PR #15008, opened 2020-10-22
Get quota usage from node repository

oyving:
The previous method for retrieving quota usage did not work consistently. Instead of finding a novel way of doing it on the controller side, I'm just copying the same way our cluster retrieval code does it by asking the node repository. To avoid some issues with system applications I had to move quota usage retrieval a bit higher up in the call path - which is probably a good thing.

@mpolden - some advise on how to test this would be appreciated.


oyving:
Question I asked @mpolden goes for you as well, @jonmv :)

mpolden:
Controller tests have a `NodeRepositoryMock` implementation of `NodeRepository` where you can put different values for `Application`. Then assert that the correct quota usage is stored in `Instance` after deployment.

You can start by using `DeploymentTester` in your test which will setup everything you need. `NodeRepositoryMock` is accessible through `DeploymentTester.configServer().nodeRepository()`.

jonmv (approved):
Yes, higher up is the right place.

jonmv (approved):

PR #15010, opened 2020-10-22
Add type resolving for ONNX models

lesters:
Adds model type resolving for ONNX models so that type resolving in Java passes expressions of the form `onnxModel(my_model).output{d0:1}`. This requires reading the ONNX model during config generation, and since the ONNX model files are protobuf, some dependencies have been added in `config-model/pom.xml` and `fat-model-dependencies/pom.xml`. I am not sure this is the best place to put these, but was the only method I found for importing correctly with OSGi. Please let me know if there is another way of doing this.

The model is read both during the prepare step, where the model file itself can be found inside the application package being deployed, and from a pre-built derived package (ZK) where the file is retrieved from the file registry. See `OnnxModelTypeResolver.java`.  Works well in single-node setup, but again I'm not sure if this violates any assumptions in a multi-node configserver setup.

@bratseth Please review.


config-model/src/main/java/com/yahoo/searchdefinition/processing/OnnxModelTypeResolver.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private static ApplicationFile getFile(Path path, ApplicationPackage app) {
+        return app.getFile(path);
+    }
+
+    private static InputStream openFromFileRepository(Path path, FileReference reference) throws FileNotFoundException {
+        return new FileInputStream(new File(getFileRepositoryPath(path, reference.value())));
+    }
+
+    public static String getFileRepositoryPath(Path path, String fileReference) {
```
bratseth, line 126: This seems like a hack to me but I guess it will work. I think we should be able to ask the application package to give us the file regardless of how it is stored in the implementation class. 

There's probably more than enough in the PR already, but we should follow it up somehow. @hmusum any thoughts on this?

bratseth (approved):
A lot of stuff here so hard to be sure of all of it but let's give it a try 👍 

I think the protobuf dependency is fine - it is small. The other way to do it is to keep it outside the config model (i.e a dependency of the config server), and create an API for it that the config model can access, but I don't think it's necessary here.

config-model/src/main/java/com/yahoo/searchdefinition/processing/OnnxModelTypeResolver.java
hmusum: I think we can just drop the `getFileReference` (and subsequent `openFromFileRepository`) calls. If `getFile` fails `getFileReference` will also fail. If there exists is a file reference for a file in the application package it will be added to the application package directory immediately afterwards and it will be found when doing `getFile`.

lesters: So, the reason this is here is the `getFile` fails to find the file when the application package is backed by `ZKApplicationPackage` because the actual file hasn't been added there. See `ZooKeeperClient.java`. During the prepare step, as far as I understand, at first the application package is a `FilesApplicationPackage`. Here the file exists. However, during config deriving, the ZK application package will be created, and the model file will not be put there, in the `.preprocessed` directory. However, we still require the actual file. That's why we look in the file repository, and at this stage the file reference exists and the file has been added there.

Now, I'm not sure what happens with this scenario when we have multiple config servers. I guess we'll just have to try?

PR #15012, opened 2020-10-22
improve generic dense reduce with more robust cell ordering

havardpe:
- unroll reduce all loop
- extend benchmark with optimized instructions (to catch slow ones)
- drop SimpleTensorEngine from benchmark (always very slow)
- tweak API on templated aggregators

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #15013, opened 2020-10-22
Bratseth/persist metrics rebased

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
```diff
@@ -53,41 +57,69 @@ public void activate(Collection<HostSpec> hosts, NestedTransaction transaction,
...
+        rememberResourceChange(transaction, generation,
+                               NodeList.copyOf(oldActive).not().retired(),
+                               NodeList.copyOf(newActive).not().retired());
         unreserveParentsOf(reservedToActivate);
     }
 
+    private void rememberResourceChange(ApplicationTransaction transaction, long generation,
+                                        NodeList oldNodes, NodeList newNodes) {
+        if (nodeRepository.applications().get(transaction.application()).isEmpty()) return; // infrastructure app, hopefully
+        Application application = nodeRepository.applications().get(transaction.application()).get();
```
freva, line 102: Read this once from ZK?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
```diff
@@ -179,6 +180,13 @@ public NodeList group(int index) {
                                                        .findFirst());
     }
 
+    public ClusterResources toResources() {
+        if (isEmpty()) return new ClusterResources(0, 0, NodeResources.unspecified());
+        return new ClusterResources(size(),
+                                    (int)stream().distinct().count(),
```
freva, line 186: Shouldn't there be a map to to group id here?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
bratseth: This is reading it once?

freva: It's reading it in the if statement above as well?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
bratseth: Yes, thanks!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
bratseth: That is true. Changed.

PR #15014, opened 2020-10-22
Refactor filestormanager tests

geirst:
@vekterli and / or @baldersheim please review

### DIFF
--- a/storage/src/tests/persistence/filestorage/filestormanagertest.cpp
+++ b/storage/src/tests/persistence/filestorage/filestormanagertest.cpp
@@ -243,8 +243,10 @@ struct FileStorHandlerComponents {
         filestorHandler = std::make_unique<FileStorHandlerImpl>(messageSender, metrics, test._node->getComponentRegister());
         filestorHandler->setGetNextMessageTimeout(50ms);
     }
-    ~FileStorHandlerComponents() {}
+    ~FileStorHandlerComponents();
 };
+
+FileStorHandlerComponents::~FileStorHandlerComponents() = default;
 
 struct PersistenceHandlerComponents : public FileStorHandlerComponents {
     ServiceLayerComponent component;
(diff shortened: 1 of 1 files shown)


storage/src/tests/persistence/filestorage/filestormanagertest.cpp
```diff
@@ -218,6 +219,57 @@ struct TestFileStorComponents {
...
+    {
+        top.push_back(std::unique_ptr<StorageLink>(dummyManager));
+        top.open();
+
+        metrics.initDiskMetrics(loadTypes.getMetricLoadTypes(), 1, threadsPerDisk);
+
+        filestorHandler = std::make_unique<FileStorHandlerImpl>(messageSender, metrics, test._node->getComponentRegister());
+        filestorHandler->setGetNextMessageTimeout(50ms);
+    }
+    ~FileStorHandlerComponents() {}
```
vekterli, line 246: Consider `= default`

storage/src/tests/persistence/filestorage/filestormanagertest.cpp
```diff
@@ -218,6 +219,57 @@ struct TestFileStorComponents {
...
+        persistenceHandler =
+                std::make_unique<PersistenceHandler>(test._node->executor(), component, cfg,
+                                                     test._node->getPersistenceProvider(),
+                                                     *filestorHandler, bucketOwnershipNotifier,
+                                                     *metrics.disk->threads[0]);
+    }
+    std::unique_ptr<DiskThread> make_disk_thread() {
+        return createThread(*persistenceHandler, *filestorHandler, component);
+    }
+    ~PersistenceHandlerComponents() {}
```
vekterli, line 270: Consider `= default`

vekterli (approved):
:+1:

storage/src/tests/persistence/filestorage/filestormanagertest.cpp
geirst: That was not enough in this case.

storage/src/tests/persistence/filestorage/filestormanagertest.cpp
geirst: That was not enough in this case.

vekterli: Move `PersistenceHandlerComponents::~PersistenceHandlerComponents() = default;` outside the class itself and the inliner becomes happy again

baldersheim: Yes, that is the correct approach. = default in class definition applies the same inliner heuristics as not defining the destructor. Defining and empty destructor with {} changes the heuristics, while doing outside the class definitions avoids the problem.

geirst: Fixed.

PR #15018, opened 2020-10-23
Do not add applications nodes outside cluster in NodePrioritizer

freva:
No point in adding application nodes outside the cluster we are allocating for, these are immediately filtered out in `NodeAllocation` and this also simplifies grouping switch hostname


mpolden (approved):
👍

PR #15029, opened 2020-10-26
partial duplicate of micro-bemchmark

arnej27959:
* use conditional compilation to let the installed
  vespa-tensor-instructions-benchmark run its
  measurements only for the two most important
  implementations (old prod and new prod).

@havardpe please review


havardpe (commented):
This looks a bit clunky. I would prefer a runtime flag over ifdefs. Lets talk about this one.

arnej27959:
@geirst FYI

havardpe (approved):

PR #15030, opened 2020-10-26
Fix corner case for system/staging test jobs

jonmv:

This fixes a corner case where we compute versions to run in system and staging tests, where the first production deployment used to compute the versions has a newer platform that what we are deploying — this confused the algorithm into believing it needed both a run which was a success both on the deploying change, AND on the newer version, which is obviously impossible. When a newer version exists in the production deployment, this is now used instead of, not in addition to, the deploying change.

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
@@ -1218,7 +1218,7 @@ public class DeploymentTriggerTest {
     }
 
     @Test
-    public void test() {
+    public void testTriggeringOfIdleTestJobsWhenFirstDeploymentIsOnNewerVersionThanChange() {
         ApplicationPackage applicationPackage = new ApplicationPackageBuilder().systemTest()
                                                                                .stagingTest()
                                                                                .region("us-east-3")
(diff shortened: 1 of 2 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
```diff
@@ -1216,4 +1217,42 @@ public void testEagerTests() {
         app.assertNotRunning(stagingTest);
     }
 
+    @Test
+    public void test() {
```
mpolden, line 1221: I think there's a better name to be found.

jonmv: 😂

jonmv: Had no idea what the problem was when I started writing the test ...

jonmv:
```suggestion
    public void testTriggeringOfIdleTestJobsWhenFirstDeploymentIsOnNewerVersionThanChange() {
```

mpolden (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
mpolden: Sounds familiar. 🙂

PR #15036, opened 2020-10-26
GC unused _maxBufferSize data member from ApplyBucketDiffReply

baldersheim:
@vekterli PR


vekterli (approved):
:+1: but should we GC it or should we actually honor it?

vekterli (approved):
:+1:

PR #15041, opened 2020-10-26
Add unit test for bucket merge with partially filled diff

toregge:
@vekterli : please review
@geirst , @baldersheim : FYI

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/mergestatus.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/mergestatus.cpp
@@ -3,10 +3,69 @@
 #include "mergestatus.h"
 #include <ostream>
 #include <vespa/log/log.h>
+#include <vespa/vespalib/stllike/hash_map.h>
+#include <cassert>
 
 LOG_SETUP(".mergestatus");
 
 namespace storage {
+
+namespace {
+
+/*
+ * Class for remapping bit masks from a partial set of nodes to a full
+ * set of nodes.
+ */
+class MaskRemapper
+{
+    std::vector<uint16_t> _mask_remap;
+
+public:
+    MaskRemapper(const std::vector<api::MergeBucketCommand::Node> &all_nodes,
+                 const std::vector<api::MergeBucketCommand::Node> &nodes);
+    ~MaskRemapper();
+
+    uint16_t operator()(uint16_t mask) const;
+};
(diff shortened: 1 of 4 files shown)


storage/src/vespa/storage/persistence/filestorage/mergestatus.cpp
```diff
@@ -21,12 +80,14 @@ MergeStatus::~MergeStatus() = default;
 bool
 MergeStatus::removeFromDiff(
         const std::vector<api::ApplyBucketDiffCommand::Entry>& part,
-        uint16_t hasMask)
+        uint16_t hasMask,
```
vekterli, line 83: Consider adding a function comment that clarifies that this is the _per-request_ mask, not the global mask (and maybe also why that is why we need to additionally provide the original node set)

storage/src/vespa/storage/persistence/filestorage/mergestatus.cpp
```diff
@@ -62,11 +123,14 @@ MergeStatus::removeFromDiff(
                 }
                 it = diff.erase(it);
                 altered = true;
-            } else if (it2->_entry._hasMask != it->_hasMask) {
-                // Hasmasks have changed, meaning bucket contents changed on
-                // one or more of the nodes during merging.
-                altered = true;
-                it->_hasMask = it2->_entry._hasMask;
+            } else {
+                uint16_t mask = remap_mask(it2->_entry._hasMask);
```
vekterli, line 127: Consider adding a comment on why we have to remap the masks here

vekterli (approved):
:+1: just have a couple of requests for additional comments that will make it easier to remember the hows and whys a few years down the line 🙂

vekterli (approved):
:+1:

storage/src/vespa/storage/persistence/filestorage/mergestatus.cpp
```diff
@@ -3,11 +3,70 @@
...
 
 namespace storage {
 
+namespace {
+
+/*
+ * Class for remapping bit masks from a partial set of nodes to a full
+ * set of nodes.
+ */
+class MaskRemapper
```
geirst, line 19: @vekterli and @toregge : Consider moving this out of the cpp file and then creating explicit unit tests for it.

PR #15043, opened 2020-10-27
Junit 4.13 and juptier 5.7.0

jonmv:
@tokle

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -517,7 +517,7 @@
             <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
-                <version>4.13.1</version>
+                <version>4.12</version>
             </dependency>
             <dependency>
                 <groupId>net.java.dev.jna</groupId>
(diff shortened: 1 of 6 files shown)


aressem (approved):

parent/pom.xml
```diff
@@ -517,7 +517,7 @@
             <dependency>
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
-                <version>4.12</version>
+                <version>4.13.1</version>
```
hmusum, line 520: You need to update code to be able to do this (since e.g. `assertThat` and `ExpectedException` are deprecated in 4.13.1 and give warnings, which we fail on). I am working on this in another branch. If upgrading works while keeping 4.12 I would go with that.

jonmv: Ok, thanks. So ... the 4.13 that was bumped elsewhere probably broke this somewhere?

jonmv, line 520:
```suggestion
                <version>4.12</version>
```

PR #15051, opened 2020-10-27
Report image pull duration metric

mpolden:
@tokle and/or @freva


docker-api/src/main/java/com/yahoo/vespa/hosted/dockerapi/DockerEngine.java
```diff
@@ -369,7 +379,16 @@ public void onComplete() {
                 numberOfDockerApiFails.increment();
                 throw new DockerClientException("Could not download image: " + dockerImage);
             }
+            sampleDuration();
         }
+
+        private void sampleDuration() {
+            Gauge gauge = metrics.declareGauge("docker.imagePullDurationSecs",
+                                               new Dimensions(Map.of("tag", dockerImage.tagAsVersion().toFullString())));
```
freva, line 387: Include repo in the tag value as well?

freva (approved):

PR #15055, opened 2020-10-27
Jonmv/wait for running steps when aborting and finishing ru

jonmv:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
@@ -55,6 +55,7 @@ import static com.yahoo.vespa.hosted.controller.deployment.Step.copyVespaLogs;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.deactivateTester;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.endStagingSetup;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.endTests;
+import static com.yahoo.vespa.hosted.controller.deployment.Step.report;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toUnmodifiableList;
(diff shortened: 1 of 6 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -353,7 +354,10 @@ public void setStartTimestamp(RunId id, Instant timestamp, LockedStep step) {
     }
 
     /** Changes the status of the given run to inactive, and stores it as a historic run. */
-    public void finish(RunId id) {
+    public void finish(RunId id) throws TimeoutException {
+        // Ensure no step is still running before we finish the run — report depends transitively on all the other steps.
+        locked(id.application(), id.type(), report, __ -> { });
```
hakonhall, line 359: Btw, I don't get how locking works in the JobRunner maintainer.  Shouldn't the maintainer
1. acquire a global zk lock
2. read all Runs and decide on some actions, then either do those actions or store to zk that those actions are in progress (using ephemeral nodes)
3. unlock the global zk lock

Instead it looks like e.g. two JobRunner may decide to finish concurrently?  Or one thread decide to abort a job whose step may already be processed in another thread.

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
hakonhall: or perhaps the read of Runs is fine, but for each Run, a lock must be held before deciding to abort() or advancement

jonmv: > Instead it looks like e.g. two JobRunner may decide to finish concurrently? Or one thread decide to abort a job whose step may already be processed in another thread.
Yes:
Finishing a run (the actual state change) is guarded by a job-specific lock, so is already OK.
The second case is exactly what this fixes: current code doesn't wait for a running step to complete before finishing the run, which means any further logging by that step is lost. 
Steps have separate locks so they can run concurrently. This _could_ be implemented as reading and writing the step status under lock as well, but I felt the lock-per-step was clearer. 
Finishing the run only when each step status is non-running requires the abort signal to cause this status to be set for each step, which can only happen after any currently running steps release their locks, and just requires more logic.

PR #15063, opened 2020-10-28
Compare .max() and .current() when calculating use

oyving:
* .max() only has a value when the user specified auto-scaling. Thus we need to compare .max() with .current() to make sure we get the correct values when calculating quota usage.

* Moved quota usage calculations into DeploymentQuotaCalculator.

* Added author and copyright notice.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/DeploymentQuotaCalculator.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/DeploymentQuotaCalculator.java
@@ -1,7 +1,9 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.application;
 
 import com.yahoo.config.application.api.DeploymentSpec;
 import com.yahoo.config.provision.ApplicationId;
+import com.yahoo.config.provision.ClusterResources;
 import com.yahoo.config.provision.zone.ZoneId;
 import com.yahoo.vespa.hosted.controller.Application;
 import com.yahoo.vespa.hosted.controller.api.integration.billing.Quota;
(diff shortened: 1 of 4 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/DeploymentQuotaCalculator.java
```diff
@@ -25,6 +32,23 @@ public static Quota calculate(Quota tenantQuota,
         return getMaximumAllowedQuota(tenantQuota, tenantApps, deployingApp, deployingZone);
     }
 
+    public static QuotaUsage calculateQuotaUsage(com.yahoo.vespa.hosted.controller.api.integration.configserver.Application application) {
+        // the .max() resources are only specified when the user has specified a max.  to make sure we enforce quotas
```
mpolden, line 36: Consider adding a unit test for these scenarios.

mpolden (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/response/application.json
```diff
@@ -0,0 +1,156 @@
+{
+  "url": "https://api.vespa-external-cd.aws.oath.cloud:4443/zone/v2/prod/aws-us-east-1c/nodes/v2/applications/vespa.album-recommendation",
```
mpolden, line 2: This should anonymized in some way.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/DeploymentQuotaCalculatorTest.java
```diff
@@ -48,4 +57,12 @@ public void zero_quota_remains_zero() {
         assertEquals(calculated.budget().get().doubleValue(), 0, 1e-5);
     }
 
+    @Test
+    public void using_highest_resource_use() throws IOException, URISyntaxException {
+        var content = new String(Files.readAllBytes(Paths.get("src/test/java/com/yahoo/vespa/hosted/controller/application/response/application.json")));
+        var mapper = new ObjectMapper();
+        var application = mapper.readValue(content, ApplicationData.class).toApplication();
```
mpolden, line 64: Assuming testing serialization isn't important here, I think it's better to construct `Application` directly.

mpolden (approved):

PR #15066, opened 2020-10-28
Handle exclusiveTo in allocation

freva:


bratseth (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -71,9 +71,14 @@
                 .filter(clusterSpec.group()::equals)
                 .count();
 
-        this.allocateForReplacement = isReplacement(nodesInCluster.size(),
-                                                    nodesInCluster.state(Node.State.failed).size());
-        this.canAllocateNew = requestedNodes instanceof NodeSpec.CountNodeSpec;
+        // In dynamically provisioned zones, we can always take spare hosts since we can provision new on-demand,
+        // NodeCandidate::compareTo will ensure that they will not be used until there is no room elsewhere.
+        // In non-dynamically provisioned zones, we only allow allocating to spare hosts to replace failed nodes.
+        this.canAllocateToSpaceHosts = isReplacement(
```
hakonhall, line 77: Space -> Spare

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -71,9 +71,14 @@
...
+                nodesInCluster.size(), nodesInCluster.state(Node.State.failed).size()) || dynamicProvisioning;
```
hakonhall, line 78: Nit:  put dynamicProvisioning as the first condition of the `||` to short-circuit isReplacement expression if true

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodePrioritizer.java
```diff
@@ -71,9 +71,14 @@
...
+        // Do not allocate new nodes for exclusive deployments in dynamically provisioned zones: provision new host instead.
+        this.canAllocateNew = requestedNodes instanceof NodeSpec.CountNodeSpec
+                && (!dynamicProvisioning || !requestedNodes.isExclusive());
```
hakonhall, line 81: `!dynamicProvisioning ||` --> `dynamicProvisioning &&` ?

freva: Then we wouldn't allocate new nodes in non-dynamically provisioned zones? This should say that we can allocate new nodes on `host`s if:
* This is a deployment with count spec
* This is not an exclusive deployment to dynamically provisioned zone

hakonhall: I see, I confused `canAllocateNew` with "can allocate new host" which has the opposite meaning in that case.

PR #15071, opened 2020-10-28
Add "documentType" to "documents_processed" metric

jonmv:
@bratseth or @baldersheim please review.

### DIFF
--- a/docproc/src/main/java/com/yahoo/docproc/DocprocExecutor.java
+++ b/docproc/src/main/java/com/yahoo/docproc/DocprocExecutor.java
@@ -187,7 +187,7 @@ public class DocprocExecutor {
         return documentType -> contextCache.computeIfAbsent(documentType, type -> {
             Map<String, String> dimensions = new HashMap<>(2);
             dimensions.put("chain", chainDimension);
-            dimensions.put("documentType", type);
+            dimensions.put("documenttype", type);
             return metric.createContext(dimensions);
         });
     }
--- a/docproc/src/main/java/com/yahoo/docproc/DocprocExecutor.java
+++ b/docproc/src/main/java/com/yahoo/docproc/DocprocExecutor.java
@@ -5,12 +5,19 @@ import com.yahoo.document.DocumentOperation;
 import com.yahoo.document.DocumentPut;
 import com.yahoo.document.json.JsonWriter;
 import com.yahoo.jdisc.Metric;
-import java.util.logging.Level;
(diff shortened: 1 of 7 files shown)


docproc/src/main/java/com/yahoo/docproc/jdisc/metric/MockMetric.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Metric implementation for tests.
+ *
+ * @author jonmv
+ */
+public class MockMetric implements Metric {
```
jonmv, line 14: Should probably put this elsewhere.

docproc/src/main/java/com/yahoo/docproc/DocprocExecutor.java
```diff
@@ -176,4 +181,15 @@ private void logProgress(Processing processing, DocumentProcessor.Progress progr
         }
         return progress;
     }
+
+    private Function<String, Metric.Context> cachedContexts(String chainDimension) {
+        Map<String, Metric.Context> contextCache = new ConcurrentHashMap<>();
+        return documentType -> contextCache.computeIfAbsent(documentType, type -> {
+            Map<String, String> dimensions = new HashMap<>(2);
+            dimensions.put("chain", chainDimension);
+            dimensions.put("documentType", type);
```
geirst, line 190: In all backend metrics this dimension is named "documenttype". We should use the same her I guess.

yngveaasheim: Geir is correct. The dimension name must be exactly "documenttype", or it cannot be used in combination with other metrics. The option would likely be to replicated "documenttype" into "documentType" for all other metrics having this already, if we conclude "documentType" is the better name.

jonmv: Thanhks.

bratseth (commented):
What's the purpose of the "once" thing?

Otherwise, LGTM

jonmv:
I believe that was there to avoid reporting each operation multiple times in case of async prosessing.

bratseth:
Ah right, it was there from before. It's not worse than it was and not public so we can clean it up later, so fine.

bratseth (approved):

jonmv:
🙉

PR #15081, opened 2020-10-29
Take provision lock before session lock

mpolden:
This avoid deadlocks with internal redeployments that may take the locks in a
different order.

@hmusum and @hakonhall

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -65,7 +65,6 @@ import com.yahoo.vespa.config.server.tenant.Tenant;
 import com.yahoo.vespa.config.server.tenant.TenantMetaData;
 import com.yahoo.vespa.config.server.tenant.TenantRepository;
 import com.yahoo.vespa.curator.Curator;
-import com.yahoo.vespa.curator.Lock;
 import com.yahoo.vespa.curator.stats.LockStats;
 import com.yahoo.vespa.curator.stats.ThreadLockStats;
 import com.yahoo.vespa.defaults.Defaults;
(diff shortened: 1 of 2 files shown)


hmusum (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -719,20 +721,26 @@ private ServiceInfo getTesterServiceInfo(ApplicationId applicationId) {
...
     // ---------------- Session operations ----------------------------------------------------------------
 
-    public CompletionWaiter activate(Session session, Session previousActiveSession, ApplicationId applicationId, boolean force) {
-        CompletionWaiter waiter = session.getSessionZooKeeperClient().createActiveWaiter();
+    public Activation activate(Session session, ApplicationId applicationId, Tenant tenant, boolean force) {
         NestedTransaction transaction = new NestedTransaction();
-        transaction.add(deactivateCurrentActivateNew(previousActiveSession, session, force));
-        if (hostProvisioner.isPresent()) {
-            try (var applicationTransaction = new ApplicationTransaction(hostProvisioner.get().lock(applicationId), transaction)) {
+        Optional<ApplicationTransaction> applicationTransaction = hostProvisioner.map(provisioner -> provisioner.lock(applicationId))
```
hakonhall, line 726: Consider making an AutoCloseable class to replace these lines and the tenant.getApplicationRepo().lock().

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
mpolden: Tried it, didn't make the flow clearer imo.

mpolden:
.

PR #15084, opened 2020-10-29
Provisioning of AWS host as exclusive, shared, or any, ++

hakonhall:
- When provisioning a host, specify whether to provision the host exclusive to
   the application, or require a shared host, or if any will do.
 - Set the exclusiveTo on the host in the node repo.


freva:
Merged in #15086

PR #15091, opened 2020-10-29
Use source-only nodes again during bucket merge if hasMask histogram …

toregge:
…shows that some diff entries are unavailable on all other nodes.

@vekterli : please review
@baldersheim , @geirst : FYI


vekterli (approved):
:+1:

PR #15094, opened 2020-10-29
Lesters/resolve onnx model types 2

lesters:
Second attempt at ONNX model type resolving. Previous attempt had issues with multiple config server and the availability of the ONNX model file. Now, if the model file is found, `OnnxModelInfo` (new class with a lot of previous logic refactored here) will write a file containing necessary info to the `models.generated/replicated` directory (similarly to `ConvertedModel`), which is replicated in Zookeeper. If the model is not available, this file is used instead.

@bratseth Please review.
@hmusum Please review.


bratseth (approved):
👍

hmusum (approved):

PR #15095, opened 2020-10-29
Smorgrav/tenantinfo

smorgrav:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
```diff
@@ -0,0 +1,106 @@
...
+        this.email = Objects.requireNonNull(email);
+        this.website = Objects.requireNonNull(website);
+        this.contactName = Objects.requireNonNull(contactName);
+        this.contactEmail = Objects.requireNonNull(contactEmail);
+        this.invoiceEmail = Objects.requireNonNull(invoiceEmail);
+        this.address = Objects.requireNonNull(address);
+        this.billingContact = Objects.requireNonNull(billingContact);
+    }
+
+    public static TenantInfo EmptyInfo = new TenantInfo("","","", "", "", "",
```
mpolden, line 34: Should be named simply `EMPTY`, same goes for the other classes.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -145,6 +147,67 @@ private CloudTenant cloudTenantFrom(Inspector tenantObject) {
         return keys.build();
     }
 
+    private TenantInfo tenantInfoFromSlime(Inspector infoObject) {
+        return TenantInfo.EmptyInfo
```
mpolden, line 157: This chould just call the constructor directly as the immutable builder methods will create a new object per field.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -145,6 +147,67 @@ private CloudTenant cloudTenantFrom(Inspector tenantObject) {
...
+                .withEmail(infoObject.field("email").asString())
+                .withWebsite(infoObject.field("website").asString())
+                .withContactName(infoObject.field("contactName").asString())
+                .withContactEmail(infoObject.field("contactEmail").asString())
+                .withAddress(tenantInfoAddressFromSlime(infoObject.field("address")))
+                .withBillingContact(tenantInfoBillingContactFromSlime(infoObject.field("billingContact")));
+    }
+
+    private TenantInfoAddress tenantInfoAddressFromSlime(Inspector addressObject) {
+        return TenantInfoAddress.EmptyAddress
```
mpolden, line 169: Same as above.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -145,6 +147,67 @@ private CloudTenant cloudTenantFrom(Inspector tenantObject) {
...
+                .withAddressLines(addressObject.field("addressLines").asString())
+                .withPostalCodeOrZip(addressObject.field("postalCodeOrZip").asString())
+                .withCity(addressObject.field("city").asString())
+                .withStateRegionProvince(addressObject.field("stateRegionProvince").asString())
+                .withCountry(addressObject.field("country").asString());
+    }
+
+    private TenantInfoBillingContact tenantInfoBillingContactFromSlime(Inspector billingObject) {
+        return TenantInfoBillingContact.EmptyBillingContact
```
mpolden, line 178: Same as above.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
```diff
@@ -0,0 +1,106 @@
...
+    public TenantInfo withAddress(TenantInfoAddress newAddress) {
+        return new TenantInfo(name, email, website, contactName, contactEmail, invoiceEmail, newAddress, billingContact);
+    }
+
+    public TenantInfo withBillingContact(TenantInfoBillingContact newBillingContact) {
+        return new TenantInfo(name, email, website, contactName, contactEmail, invoiceEmail, address, newBillingContact);
+    }
+
+    public boolean isEmpty() {
+        return (name + email + website + contactEmail + contactName + invoiceEmail).isEmpty()
```
mpolden, line 102: If this implements equals/hashCode you can simply return `this.equals(EMPTY)`. Same for the other classes. This will also simplify serialization testing.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializerTest.java
```diff
@@ -80,7 +81,8 @@ public void cloud_tenant() {
         CloudTenant tenant = new CloudTenant(TenantName.from("elderly-lady"),
                                              Optional.of(new SimplePrincipal("foobar-user")),
                                              ImmutableBiMap.of(publicKey, new SimplePrincipal("joe"),
-                                                               otherPublicKey, new SimplePrincipal("jane")));
+                                                               otherPublicKey, new SimplePrincipal("jane")),
+                                             TenantInfo.EmptyInfo);
```
mpolden, line 88: Should set a non-empty `TenantInfo` here, and assert the re-serialized value below to ensure that serialization is correct.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Objects;
+
+/**
+ * Tenant information beyond technical tenant id and user authorizations.
+ *
+ * This info is used to capture generic support information and invoiced billing information.
+ *
+ * All fields are non null but strings can be empty
+ */
+public class TenantInfo {
```
mpolden, line 12: Missing `@author` and license header.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfoAddress.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Address formats are quite diverse across the world both in therms of what fields are used, named and
+ * the order of them.
+ *
+ * To be generic a little future proof the address fields here are a mix of free text (address lines) and fixed fields.
+ * The address lines can be street address, P.O box, c/o name, apartment, suite, unit, building floor etc etc.
+ *
+ * All fields are mandatory but can be an empty string (ie. not null)
+ */
+public class TenantInfoAddress {
```
mpolden, line 14: Missing `@author` and license header.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfoBillingContact.java
```diff
@@ -0,0 +1,54 @@
+package com.yahoo.vespa.hosted.controller.tenant;
+
+import java.util.Objects;
+
+public class TenantInfoBillingContact {
```
mpolden, line 5: Missing `@author` and license header.

mpolden (changes requested):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
smorgrav: I wasn't aware of that. Is there a naming convention document that I have missed?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
smorgrav: Not sure I understand - what is the difference?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
smorgrav: Good catch!

Could you give a hint to what part of testing could be simplified due to this?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
mpolden: In general the Java convention is `UPPERCASE` for constants. In this module we typically name such constants either `EMPTY/NONE` or `empty/none`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
mpolden: The difference is the number of objects you create. In the current state its is multiplied by the number of fields in `TenantInfo` (since each `withXXX` creates a new object), this is then multiplied by the number of tenants you're deserializing.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
mpolden: Also, this should be `final`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
mpolden: If you add a fully populated `TenantInfo` to `@Test public void cloud_tenant()` you can do do `assertEquals(tenant.tenantInfo(), serialized.tenantInfo())` to test that serialization works as expected in both directions.

smorgrav: Ah you where referring to if it implemented equals/hashCode - not how to implement the isEmpty method?I think that is exactly what is done right?

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializerTest.java
smorgrav: I guess. The non empty case is covered below (albeit not in the context of a CloudTenant) - the normal case is that it is empty. I can create another test.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
smorgrav: Ah - an optimisation. Using builder methods adds some benefits that I think outweigh that - one is that the order of the constructor when using strings are prone to mistakes (and this eliminates that).

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
mpolden: Yes, never mind! You should still be able to simplify `isEmpty` though, by just comparing with empty.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializerTest.java
mpolden: Right, I didn't notice the other tests.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
smorgrav: fixed

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
smorgrav: fixed

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
smorgrav: fixed

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfoAddress.java
smorgrav: fixed

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfoBillingContact.java
smorgrav: fixed

mpolden (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializerTest.java
smorgrav: Fixed. Added one extra test there to cover that the info field is set correctly in the tenant object.

PR #15098, opened 2020-10-29
Handle timeout units in doc/v1, and catch parse errors in handler

jonmv:
@hmusum or @baldersheim please review.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
@@ -100,6 +100,7 @@ import static com.yahoo.jdisc.http.HttpRequest.Method.PUT;
 import static java.util.Objects.requireNonNull;
 import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.WARNING;
+import static java.util.stream.Collectors.collectingAndThen;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toUnmodifiableMap;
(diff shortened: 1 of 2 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -113,7 +114,11 @@
 
     private static final Logger log = Logger.getLogger(DocumentV1ApiHandler.class.getName());
     private static final Parser<Integer> integerParser = Integer::parseInt;
-    private static final Parser<Double> doubleParser = Double::parseDouble;
+    private static final Parser<Double> timeoutParser = value -> {
+        int suffixLength = value.endsWith("ms") ? 2 : value.endsWith("s") ? 1 : 0;
```
baldersheim, line 118: What about us and ns ?
Don't we have som utility method for this ?

jonmv: It doesn't need to be an integer, do I don't see the need for other units. The search handler recognises `ms`, and treats _everything else_ as `s`. I think this is strictly better.

jonmv: Actually, it recognise `ks` and `µs` as well, which is just obscene. I found the code, which is split out somewhere, but it's in the `search` module.

jonmv: May as well use that code, then.

jonmv: (It was already a dependency of this module.)

jonmv:
Merging to fix that test.

PR #15101, opened 2020-10-30
Use documented environment variable for validation script

hmusum:
Use what's documented in https://docs.vespa.ai/documentation/reference/files-processes-and-ports.html#pre-start-check


kkraune (approved):
Thanks!

PR #15103, opened 2020-10-30
Skip old metrics

bratseth:


freva (approved):
This should fix the problem, but wont this also lead to massive gaps in metrics coverage?

bratseth:
Gaps? Yes. Massive? Not obvious because this timestamp has second resolution. However, I looked at the metrics code and you are right: Metrics are cached for 30 seconds along the way which means we'll have something like 29/30 probability of this happening. However, we already falsify the actual measurement times:

```
    /**
     * If the metrics packet is less than one minute newer or older than the given startTime,
     * set its timestamp to the given startTime. This is done to ensure that metrics retrieved
     * from different sources for this invocation get the same timestamp, and a timestamp as close
     * as possible to the invocation from the external metrics retrieving client. The assumption
     * is that the client requests metrics periodically every minute.
     * <p>
     * However, if the timestamp of the packet is too far off in time, we don't adjust it because
     * we would otherwise be masking a real problem with retrieving the metrics.
     */
    static MetricsPacket.Builder adjustTimestamp(MetricsPacket.Builder builder, Instant startTime) {
        Duration age = Duration.between(startTime, builder.getTimestamp());
        if (age.abs().minusMinutes(1).isNegative())
            builder.timestamp(startTime.getEpochSecond());
        return builder;
    }
```

So the node repo can just do the same without loss of precision.

bratseth:
@freva please have another look

freva (approved):
:+1:

PR #15111, opened 2020-10-30
Add TENANT_ID as a flag dimension

oyving:
* Added `TENANT_ID` as a Flags dimension that matches only the tenant name.
* Updated the`TENANT_BUDGET_QUOTA` to use that dimension.


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
```diff
@@ -221,6 +221,9 @@ private static void verifyValues(JsonNode root) {
             } else if (dimension.isEqualTo(DimensionHelper.toWire(FetchVector.Dimension.CONSOLE_USER_EMAIL))) {
                 condition.get("values").forEachArrayElement(conditionValue -> conditionValue.asString()
                         .orElseThrow(() -> new IllegalArgumentException("Non-string email address: " + conditionValue)));
+            } else if (dimension.isEqualTo(DimensionHelper.toWire(FetchVector.Dimension.TENANT_ID))) {
+                condition.get("values").forEachArrayElement(conditionValue -> conditionValue.asString()
```
hakonhall, line 225: Consider adding some additional validation of tenant id

hakonhall (approved):

PR #15118, opened 2020-10-30
FeedView has not used SearchView since we had support for the remove

toregge:
batch feed operation which contained a query.

@baldersheim : please review
@geirst : FYI


baldersheim (approved):

baldersheim:
👍

PR #15119, opened 2020-10-30
added simple median aggregator

havardpe:

@arnej27959 please review
@bratseth FYI

### DIFF
--- a/eval/src/vespa/eval/eval/aggr.h
+++ b/eval/src/vespa/eval/eval/aggr.h
@@ -8,6 +8,7 @@
 #include <vector>
 #include <map>
 #include <algorithm>
+#include <cmath>
 
 namespace vespalib {
 
@@ -126,11 +127,17 @@ private:
     std::vector<T> _seen;
 public:
     constexpr Median() : _seen() {}
-    constexpr Median(T value) : _seen({value}) {}
-    constexpr void sample(T value) { _seen.push_back(value); }
+    constexpr Median(T value) : _seen() {
+        sample(value);
+    }
+    constexpr void sample(T value) {
+        if (!std::isnan(value)) {
+            _seen.push_back(value);
+        }
+    }
     constexpr void merge(const Median &rhs) {
         for (T value: rhs._seen) {
-            _seen.push_back(value);
+            sample(value);
         }
     };
     constexpr T result() const {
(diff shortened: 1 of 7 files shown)


eval/src/vespa/eval/eval/aggr.h
```diff
@@ -120,6 +121,32 @@ template <typename T> class Max {
...
+    };
+    constexpr T result() const {
+        if (_seen.empty()) {
+            return std::numeric_limits<T>::quiet_NaN();
+        }
+        std::vector<T> tmp = _seen;
+        std::sort(tmp.begin(), tmp.end());
+        size_t cnt = tmp.size();
+        if ((cnt % 2) == 0) {
+            return ((tmp[cnt/2] + tmp[(cnt-1)/2]) / T{2});
```
arnej27959, line 144: if you want, you can always do this, because if "cnt" is odd then (cnt/2) == (cnt-1)/2 and (x+x)/2 == x

arnej27959 (approved):
looks good to me

eval/src/vespa/eval/eval/aggr.h
toregge: isn't std::nth_element, with an extra std::max_element or std::min_element in case of even number of elements cheaper ?

havardpe: @arnej27959 : yes, I also thought of that, but could not get myself to do it, even if the performance will be 'the same'
@toregge : I went for simple over fast, since it will be slow no matter what. Also, I am still not friends with nth_element (we have history). More importantly, your comment reminds me that sort will not work with NaN values. Need a new sort that works with NaN, I guess... hmm

havardpe:
need to resolve sort vs. NaN issue.

arnej27959:
Since we copy to tmp anyway I suggest skipping NaN when copying.

arnej27959:
Also note that formally you can just pick tmp[cnt/2] and that is a median always 😃

havardpe:
The problem is that Python is the new formal, and it says we need to average if the size is even

havardpe:
@arnej27959 PTAL

arnej27959 (approved):
👍

havardpe:
@toregge PTAL

toregge (approved):
👍

havardpe:
turns out we have implemented 'nanmedian' (ignore NaN values) and not 'median' (you get NaN if you have any NaN values). I will change the behavior to 'median' since it better matches what we do for other aggregators and also was what was requested.

havardpe:
@arnej27959 @toregge PTAL, hopefully for the last time

toregge (approved):

PR #15120, opened 2020-10-30
avoid unrolling reduce all loop

havardpe:
This is to see if it helps the number stability of the BERT performance
test. Note that unrolling made it go 4 times faster, so we might want
to re-introduce it again later.

@geirst please review
@arnej27959 FYI


arnej27959 (approved):
this is probably worth checking, but if it gives a large difference in end results, it's very likely that the unrolled variant is more exact - at least for sum, summing lots of positive numbers into the same aggregator will gradually lose more and more precision, so the unrolled version is 4 times more precise also.

PR #15121, opened 2020-10-30
[VESPA-19119] Report Java heap dumps

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoreCollector.java
```diff
@@ -24,9 +24,11 @@
 public class CoreCollector {
     private static final Logger logger = Logger.getLogger(CoreCollector.class.getName());
 
+    private static final Pattern JAVA_CORE_PATTERN = Pattern.compile("java_pid.*\\.hprof");
```
hakonhall, line 27: Let's name this heap dump.  A Java core is a kernel dump of the process (we have those too I guess).

hakonhall (approved):

PR #15124, opened 2020-10-30
Bjorncs/detect reindexing config change

bjorncs:

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeAction.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeAction.java
@@ -4,7 +4,6 @@ package com.yahoo.config.model.api;
 import com.yahoo.config.provision.ClusterSpec;
 
 import java.util.List;
-import java.util.Optional;
 
 /**
  * Contains the action to be performed on the given services to handle a config change
@@ -15,7 +14,7 @@ import java.util.Optional;
 public interface ConfigChangeAction {
 
     enum Type {
-        RESTART("restart"), REFEED("refeed");
+        RESTART("restart"), REFEED("refeed"), REINDEXING("reindexing");
 
         private final String type;
 
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeAction.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeAction.java
@@ -14,7 +14,7 @@ import java.util.List;
 public interface ConfigChangeAction {
 
     enum Type {
-        RESTART("restart"), REFEED("refeed"), REINDEXING("reindexing");
+        RESTART("restart"), REFEED("refeed"), REINDEX("reindex");
 
         private final String type;
 
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
@@ -1,0 +1,20 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.config.model.api;
+
+import java.util.Optional;
+
+/**
+ * Represents an action to re-index a document type in order to handle a config change.
+ *
+ * @author bjorncs
+ */
+public interface ConfigChangeReindexAction extends ConfigChangeAction {
+
+    @Override default Type getType() { return Type.REINDEXING; }
+
+    /** @return name identifying this kind of change, used to identify names which should be allowed */
+    String name();
+
+    /** @return name of the document type that must bed re-indexed, or empty if all document types */
+    Optional<String> getDocumentType();
+}
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
@@ -10,7 +10,7 @@ import java.util.Optional;
  */
 public interface ConfigChangeReindexAction extends ConfigChangeAction {
 
-    @Override default Type getType() { return Type.REINDEXING; }
+    @Override default Type getType() { return Type.REINDEX; }
 
     /** @return name identifying this kind of change, used to identify names which should be allowed */
     String name();
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
@@ -1,8 +1,6 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.config.model.api;
 
-import java.util.Optional;
-
 /**
  * Represents an action to re-index a document type in order to handle a config change.
  *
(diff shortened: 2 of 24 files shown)


config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeAction.java
```diff
@@ -15,7 +14,7 @@
 public interface ConfigChangeAction {
 
     enum Type {
-        RESTART("restart"), REFEED("refeed");
+        RESTART("restart"), REFEED("refeed"), REINDEXING("reindexing");
```
jonmv, line 17: `REINDEX` would be consistent with the other names here.

config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
```diff
@@ -0,0 +1,20 @@
...
+ * @author bjorncs
+ */
+public interface ConfigChangeReindexAction extends ConfigChangeAction {
+
+    @Override default Type getType() { return Type.REINDEXING; }
+
+    /** @return name identifying this kind of change, used to identify names which should be allowed */
+    String name();
+
+    /** @return name of the document type that must bed re-indexed, or empty if all document types */
```
jonmv, line 18: Hmm, suggestions don't work today?

Just wanted to point out typo `bed --> be`

config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
```diff
@@ -0,0 +1,20 @@
...
+    Optional<String> getDocumentType();
```
jonmv, line 19: Hmm, is "all document types" an option here?

config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
jonmv: Ah, no suggestions in per-commit-view, of course.

config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/ConfigChangeTestUtils.java
```diff
@@ -37,6 +37,15 @@ public static VespaConfigChangeAction newRefeedAction(ClusterSpec.Id id, String
         return VespaRefeedAction.of(id, name, overrides, message, services, documentType, now);
     }
 
+    public static VespaConfigChangeAction newReindexingAction(ClusterSpec.Id id, String name, ValidationOverrides overrides, String message, Instant now) {
```
jonmv, line 40: s/indexing/index/g

jonmv (approved):
Very good! I'd love to see the enum value change name, but that's it :)

jonmv (commented):
Very good! I'd love to see the enum value change name.

I also think we need the document type in the action, lest we be unable to store anything in the data store, since we probably don't know (easily) all the document types of the application there.

jonmv:
Laggy github + laggy brain = two differing reviews.

config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeReindexAction.java
bjorncs: I'm not sure, most likely not. The programmatic API allows it to be null. It is first null and then later updated with document type (uncertain whether this always happen).

bjorncs:
@jonmv PTAL

bjorncs:
@jonmv PTAL

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
```diff
@@ -336,6 +336,28 @@ public void handleRedundancy(Redundancy redundancy) {
...
+                streamingDocTypes.add(type);
+            }
+        }
+        return streamingDocTypes;
+    }
+
+    public List<NewDocumentType> getDocumentTypesWithIndexedCluster() {
+        List<NewDocumentType> indexedDocTypes = new ArrayList<>();
+        for (NewDocumentType type : documentDefinitions.values()) {
+            if (findStreamingCluster(type.getFullName().getName()).isEmpty()
```
jonmv, line 352: Not sure this first part is needed?

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
```diff
@@ -336,6 +336,28 @@ public void handleRedundancy(Redundancy redundancy) {
                 .collect(Collectors.toList());
     }
 
+    public List<NewDocumentType> getDocumentTypesWithStreamingCluster() {
```
jonmv, line 339: I don't think we need this, as we should be happy checking whether a document changes between "index" and not "index".

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/IndexingModeChangeValidator.java
```diff
@@ -27,31 +35,56 @@
...
+        {
+            Set<String> currentIndexedTypes = toDocumentTypeNames(currentSearchCluster.getDocumentTypesWithIndexedCluster());
+            Set<String> nextStreamingTypes = toDocumentTypeNames(nextSearchCluster.getDocumentTypesWithStreamingCluster());
+            for (String type : nextStreamingTypes) {
+                if (currentIndexedTypes.contains(type)) {
+                    changes.add(createReindexAction(overrides, now, nextCluster, type, "indexed", "streaming"));
+                }
+            }
+        }
+        return changes;
```
jonmv, line 66: We'll miss changes from "index" to "store-only", or vv.

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
bjorncs: It mimics the semantics of `getConfig`

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
bjorncs: It's the transitions between indexing modes we want to react on. We have to check if the document type is new or removed otherwise.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/IndexingModeChangeValidator.java
bjorncs: Ok, I can add that as well

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
jonmv: Yeah, if we need to separate streaming and store-only, we need it.

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
jonmv: But a document type must be either indexed or not — it would suffice to check whether it went from indexed to not indexed, or the other way, I believe.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/IndexingModeChangeValidator.java
jonmv: Ok, later.

jonmv:
Unicorn

PR #15129, opened 2020-10-31
Use javadoc-no-fork and various fixes

hakonhall:
The `no-fork` avoids an extra generate-sources.
Removes support for clover (that is not used AFAIK) and a Util that was unnecessary.
Modernizes some code.


yolean/pom.xml
```diff
@@ -37,10 +37,6 @@
                 <artifactId>bundle-plugin</artifactId>
                 <extensions>true</extensions>
             </plugin>
-            <plugin>
```
bjorncs, line 40: Consider keeping the `maven-compiler-plugin` section. It helps IntelliJ to detect correct language level on import.

bjorncs (approved):

yolean/pom.xml
hakonhall: That's why...

bjorncs (approved):

PR #15134, opened 2020-11-02
Add required="true" for override without node children

hakonhall:


config-application-package/src/main/java/com/yahoo/config/application/OverrideProcessor.java
```diff
@@ -218,9 +218,17 @@ private int getNumberOfOverrides(Element child, Context context) {
     private void doElementSpecificProcessingOnOverride(List<Element> elements) {
         // if node capacity is specified explicitly for some combination we should require that capacity
         elements.forEach(element -> {
-            if (element.getTagName().equals("nodes"))
-                if (element.getChildNodes().getLength() == 0) // specifies capacity, not a list of nodes
+            if (element.getTagName().equals("nodes")) {
+                boolean hasNodeChild = false;
```
bratseth, line 222: Would be better to put this in a function imho

bratseth (approved):
👍

config-application-package/src/main/java/com/yahoo/config/application/OverrideProcessor.java
hakonhall: Fixed

PR #15140, opened 2020-11-02
Smorgrav/tenantinfo get api

smorgrav:


smorgrav:
This should implement the read functionality for tenant info at tenant/info

I also took the opportunity to start refactor these classes due to their size.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
```diff
@@ -51,7 +51,7 @@
     /** Paths used for user management on the tenant level. */
     tenantUsers(Matcher.tenant,
                 PathPrefix.api,
-                "/user/v1/tenant/{tenant}"),
+                "/user/v1/tenant/{tenant}", "/user/v1/tenant/{tenant}/info/"),
```
jonmv, line 54: One per line, please :)

jonmv: Should this be in the above group?

jonmv (approved):

PR #15142, opened 2020-11-02
Added 'breakfixed' node state

olaaun:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -476,7 +476,10 @@ public final class Node implements Nodelike {
         parked,
 
         /** This host has previously been in use but is now removed. */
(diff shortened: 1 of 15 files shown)


freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
```diff
@@ -476,7 +476,10 @@ public String toString() {
         parked,
 
         /** This host has previously been in use but is now removed. */
-        deprovisioned;
+        deprovisioned,
+
+        /** This host is currently undergoing repair. */
+        breakfixed;
```
mpolden, line 482: Why do we need a new state?

Valid state transitions to/from this should be documented in `NodeRepository` (at the top).

olaaun: Fixed

PR #15144, opened 2020-11-02
Increase priority for global bucket merges

vekterli:

To avoid global buckets competing with (and usually being starved by)
default bucket space merges, explicitly prioritize default bucket
merges above most other load.

Increases both distributor-internal maintenance schedulinr priority and
persistence-level operation priority.

### DIFF
--- a/storage/src/vespa/storage/distributor/statecheckers.cpp
+++ b/storage/src/vespa/storage/distributor/statecheckers.cpp
@@ -2,6 +2,7 @@
 
 #include "statecheckers.h"
 #include "activecopy.h"
+#include <vespa/document/bucket/fixed_bucket_spaces.h>
 #include <vespa/storage/distributor/operations/idealstate/splitoperation.h>
 #include <vespa/storage/distributor/operations/idealstate/joinoperation.h>
 #include <vespa/storage/distributor/operations/idealstate/removebucketoperation.h>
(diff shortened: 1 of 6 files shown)


storage/src/vespa/storage/distributor/statecheckers.cpp
```diff
@@ -861,11 +862,18 @@ SynchronizeAndMoveStateChecker::check(StateChecker::Context& c)
...
+        MaintenancePriority::Priority schedPri;
+        if (c.getBucketSpace() == document::FixedBucketSpaces::default_space()) {
+            schedPri = (result.needsMoveOnly() ? MaintenancePriority::LOW
+                                               : MaintenancePriority::MEDIUM);
+            op->setPriority(result.priority());
+        } else {
+            // Since the default bucket space has a dependency on the global bucket space,
+            // we prioritize scheduling of merges to global buckets over those for default buckets.
+            schedPri = MaintenancePriority::HIGH;
+            op->setPriority(c.distributorConfig.getMaintenancePriorities().mergeGlobalBuckets);
```
toregge, line 875: This might trigger another kind of starvation if new global documents are fed to a cluster where a node is full.

geirst (approved):
👍

storage/src/vespa/storage/distributor/statecheckers.cpp
toregge: After having looked at the maintenance scheduler, this doesn't seem to be a problem. Buckets are added to a bucket priority database by a maintenance scanner, and the scheduler just picks buckets from that bucket priority database. If a maintenance operation fails then it is not retried until the maintenance scanner checks the bucket again.

toregge (approved):

vekterli:
@geirst @toregge PTAL at the latest commit. I've added an escape valve for manually getting out of a resource exhaustion scenario where global merges are starving default merges. This is of course not optimal, but serves as a stop-gap until we have more robust and generalized resource exhaustion handling across nodes.

toregge (approved):

geirst (approved):
👍

PR #15145, opened 2020-11-02
Use field set of document type when not set, for document GETs

jonmv:
@bjorncs or @freva please review and merge.


freva (approved):
:shrug:

bjorncs (approved):

PR #15158, opened 2020-11-03
Don't enforce quota in test deployments

oyving:
FYI @kkraune


mpolden (approved):

andreer:
Perhaps we should still limit this to the tenant's max overall quota?

oyving:
I think that's an excellent idea!

PR #15172, opened 2020-11-04
Write content for files in diretory to file

hmusum:
Instead of storing data from file reference that is a directory in a byte array
(which might be huge and use a lot of memory), store in a temporary file
and delete file when calling close().


baldersheim (approved):
👍  Merge at will

PR #15178, opened 2020-11-04
Bjorncs/detect indexing mode change storeonly

bjorncs:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
@@ -3,10 +3,10 @@ package com.yahoo.vespa.model.content;
 
 import com.yahoo.config.model.api.ModelContext;
 import com.yahoo.config.model.deploy.DeployState;
+import com.yahoo.config.model.producer.AbstractConfigProducer;
+import com.yahoo.documentmodel.NewDocumentType;
 import com.yahoo.vespa.config.search.DispatchConfig;
 import com.yahoo.vespa.config.search.core.ProtonConfig;
-import com.yahoo.documentmodel.NewDocumentType;
-import com.yahoo.config.model.producer.AbstractConfigProducer;
 import com.yahoo.vespa.model.builder.UserConfigBuilder;
 import com.yahoo.vespa.model.builder.xml.dom.DomSearchTuningBuilder;
 import com.yahoo.vespa.model.builder.xml.dom.ModelElement;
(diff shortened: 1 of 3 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
```diff
@@ -357,6 +357,18 @@ public void handleRedundancy(Redundancy redundancy) {
         return indexedDocTypes;
     }
 
+    public List<NewDocumentType> getDocumentTypesWithStoreOnly() {
+        List<NewDocumentType> indexedDocTypes = new ArrayList<>();
```
jonmv: Strange variabble type for a store-only set ;)

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
```diff
@@ -357,6 +357,18 @@ public void handleRedundancy(Redundancy redundancy) {
...
+        for (NewDocumentType type : documentDefinitions.values()) {
+            if (findStreamingCluster(type.getFullName().getName()).isEmpty() &&
+                    (hasIndexedCluster() && !getIndexed().hasDocumentDB(type.getFullName().getName()) ||
+                            !hasIndexedCluster())) {
```
jonmv: Could be simplified to ```
( ! hasIndexedCluster() || ! getIndexed().hasDocumentDB(type.getFullName().getName()))
```

jonmv (approved):
Very nice!

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
bjorncs: 😛

config-model/src/main/java/com/yahoo/vespa/model/content/ContentSearchCluster.java
bjorncs: Good point!

kkraune:
I wonder - what is the difference between streaming and store-only? Both are really no-index - should we kill store-only (i.e. deprecate and kill in vespa-8)?

bratseth:
We set up the search middleware if you want to do streaming. Could also be nice to know whether you plan to do streaming to tune how we do things in the content layer (although I don't think we do that today).

If we deprecate something I think it should be streaming, but due to the above I don't think we should.

kkraune:
thanks! what _would_ be nice is a way to control the search middleware independently - say, you have a regular indexed search, but for some cases you would want to stream search the doc store (which i believe is the same in both cases). not that off if you think of it, if you want to experiment with match modes on your corpus and don't mind latency

PR #15179, opened 2020-11-04
Remove handling of tensor-type-change in AttributeChangeValidator

bjorncs:
This is no longer needed as tensor type now follows the field type, and such changes are handled in DocumentChangeValidator.


bjorncs:
@geirst

geirst (approved):
👍

arnej27959:
As far as I can see this is just wrong:

> This is no longer needed as tensor type now follows the field type, and such changes are handled in DocumentChangeValidator.

What has happened "now" that would make this work?

A field with a tensor has a "DataType" that is a TensorDataType, but the comparison in DataType works in such a way that all TensorDataTypes are equal, so specific validation is necessary.

PR #15183, opened 2020-11-04
Distinguish between "no opinion" and "keep current allocation"

bratseth:
This is necessary when multiple config servers runs autoscaling in
parallel and redeployment takes along time.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
@@ -39,29 +39,25 @@ public class Autoscaler {
      * without taking min and max limits into account.
      *
      * @param clusterNodes the list of all the active nodes in a cluster
-     * @return a new suggested allocation for this cluster, or empty if it should not be rescaled at this time
+     * @return scaling advice for this cluster
      */
-    public Optional<ClusterResources> suggest(Cluster cluster, List<Node> clusterNodes) {
-        return autoscale(cluster, clusterNodes, Limits.empty(), cluster.exclusive())
-                       .map(AllocatableClusterResources::toAdvertisedClusterResources);
-
+    public Advice suggest(Cluster cluster, List<Node> clusterNodes) {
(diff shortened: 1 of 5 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -70,14 +66,14 @@ public Autoscaler(MetricsDb metricsDb, NodeRepository nodeRepository) {
...
+        if (cpuLoad.isEmpty() || memoryLoad.isEmpty() || diskLoad.isEmpty()) return Advice.none();
         var target = ResourceTarget.idealLoad(cpuLoad.get(), memoryLoad.get(), diskLoad.get(), currentAllocation);
 
         Optional<AllocatableClusterResources> bestAllocation =
                 allocationOptimizer.findBestAllocation(target, currentAllocation, limits, exclusive);
-        if (bestAllocation.isEmpty()) return Optional.empty();
-        if (similar(bestAllocation.get(), currentAllocation)) return Optional.empty();
-        return bestAllocation;
+        if (bestAllocation.isEmpty()) return Advice.dontScale();
+        if (similar(bestAllocation.get(), currentAllocation)) return Advice.dontScale();
```
freva, line 75: Shouldn't this return `currentAllocation`?

In the case this is meant to fix:
1. cfg1 calculates a new target, `t1`, and deploys with it
2. cfg2 decides that the new target, `t2`, which is not sufficiently different from the current, so it returns empty, which means that target `t1` is cleared
3. On the next deploy, on any cfg: There is no target and the nodes provisioned with `t1` are not yet active, so it will revert back to the previous target (per `NodeRepositoryProvsioner::currentResources`)

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
bratseth: Current isn't the current target but the current allocation (active nodes).

PR #15191, opened 2020-11-05
Guard against timeout value that is too high and will cause overflow.

baldersheim:
@havardpe PR

### DIFF
--- a/vespalib/src/tests/time/time_test.cpp
+++ b/vespalib/src/tests/time/time_test.cpp
@@ -59,4 +59,9 @@ TEST(TimeTest, to_string_print_iso_time) {
     EXPECT_EQ("2019-12-20 02:47:35.768 UTC", to_string(system_time(1576810055768543us)));
 }
 
+TEST(TimeTest, conversion_of_max) {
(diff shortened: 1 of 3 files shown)


vespalib/src/tests/time/time_test.cpp
```diff
@@ -59,4 +59,9 @@ TEST(TimeTest, to_string_print_iso_time) {
     EXPECT_EQ("2019-12-20 02:47:35.768 UTC", to_string(system_time(1576810055768543us)));
 }
 
+TEST(TimeTest, conversion_of_max) {
+    EXPECT_EQ(-9223372036.8547764, vespalib::to_s(vespalib::duration::min()));
```
havardpe, line 63: maybe this should be changed to float semantics, as in min is the smallest positive duration

havardpe (approved):

vespalib/src/tests/time/time_test.cpp
baldersheim: This is std::chrono::duration::min/max, so not up to us.

PR #15192, opened 2020-11-05
Add node state expiry metrics

freva:
`expired.reserved` metric can be used to detect issues with provisioning, `expired.(provisioned|dirty)` could be used to detect issues with hosts.


mpolden (approved):
:+1:

PR #15193, opened 2020-11-05
Return empty map, not null from getReports()

bjormel:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
@@ -1,6 +1,7 @@
 // Copyright 2018 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.api.integration.noderepository;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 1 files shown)


bjormel:
@hakonhall 
@olaaun

hakonhall (commented):
(weird - this comment ended up in the wrong place and is undeletable)

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
```diff
@@ -357,8 +358,9 @@ public void setWantedFirmwareCheck(Long wantedFirmwareCheck) {
         this.wantedFirmwareCheck = wantedFirmwareCheck;
     }
 
+    @JsonIgnore
     public Map<String, JsonNode> getReports() {
```
hakonhall, line 362: If NodeRepositoryNode will be used to serialize to JSON, I suggest adding a method

```
@JsonGetter("reports")
public Map<...> getReportsOrNull() { return reports; }
```

that will be used by jackson to serialize, while our code will typically refer to the `getReports()` to avoid having to deal with `null`.

hakonhall: This would be in addition to the changes you've done in this PR so far.

bjormel: Not sure that I get what you mean. Should I revert getReports to its original state?

hakonhall: No, my proposed changes are in addition to the ones you have already done in this PR.

hakonhall (approved):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
bjormel: nice, the getter really saved me the null handling 👍

PR #15202, opened 2020-11-06
Report metrics on athenz client errors

tokle:


bjorncs (approved):
Do we want to handle 5xx responses as errors (in addition to connection failures)?

bjorncs (approved):

PR #15203, opened 2020-11-06
Fix minor stuff found while reading code, no functional changes

hmusum:


config-model-api/src/main/java/com/yahoo/config/model/api/ConfigChangeRefeedAction.java
```diff
@@ -12,8 +12,7 @@
     default Type getType() { return Type.REFEED; }
 
     /** Returns the name identifying this kind of change, used to identify names which should be allowed */
-    // Remove this default implementation when model versions earlier than 5.125 are gone
```
jonmv, line 15: About time! :D

jonmv (approved):

PR #15209, opened 2020-11-06
Kkraune/autoscale debug

kkraune:
Autoscale decisions are hard to debug when no autoscale happens - is it running, why does it not autoscale? I suggest we add debug logging for now.

Should maybe add a 'reason' enum to Advice and log on outside of autoscale instead?


bratseth (approved):

bratseth:
Yes, maybe ... if we keep this.

PR #15216, opened 2020-11-07
Use most specific, not latest, reindexing ready timestamp, in config

jonmv:


jonmv:
With latest, we'd do triple the reindexing if application, cluster and document type had separate timestamps, and they're all incremented independently by the maintainer. 

Using the most specific is the right change. 

We could also consider changing the maintainer to override more specific status when scheduling periodic reindexing.

We should definitely do this for user-input reindexing.

bjorncs (approved):

PR #15227, opened 2020-11-09
Rest api to patch collection method - only accessible for accountants

smorgrav:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
@@ -164,6 +164,12 @@ enum Policy {
             .on(PathGroup.billingPlan)
             .in(SystemName.PublicCd)),
 
+    /** Ability to update tenant collection method */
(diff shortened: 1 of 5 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
```diff
@@ -164,6 +164,12 @@
             .on(PathGroup.billingPlan)
             .in(SystemName.PublicCd)),
 
+    /** Ability to update tenant collection method */
+    collectionMethodUpdate(Privilege.grant(Action.update)
+            .on(PathGroup.billingCollection)
+            .in(SystemName.PublicCd)),
```
oyving, line 170: Just make this for all of public now?

oyving (approved):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/Policy.java
smorgrav: We could - I suggest we make a PR where we change all relevant policies in one go

PR #15240, opened 2020-11-09
Reapply "Jupiter 5.7.0" 

hmusum:
Reverts vespa-engine/vespa#15054


jonmv (approved):

hmusum:
@bjorncs could you comment on the change in  tenant-cd-api/pom.xml?  Is it OK to bump the version here and in internal repo or does the comment indicate that more needs to be done?

tenant-cd-api/pom.xml
```diff
@@ -25,7 +25,7 @@
             This version must match the string in all ExportPackage annotations.
```
bjorncs, line 25: @jonmv Read this!

bjorncs (changes requested):
See comment

tenant-cd-api/pom.xml
jonmv: Thanks!

hmusum: Where are the ExportPackage annotations we are talking about here?

bjorncs: In the package-info.java files in tenant-cd-api (for the the junit annotations).

hmusum: Updated (although the old version exported was 5.4.2, not the expected 5.6.2)

bjorncs (approved):

bjorncs:
FYI @tokle

PR #15241, opened 2020-11-09
Use 8 bits as minimum for "legacy" distribution type

vekterli:
@geirst please review

No one should use this anyway, but since it's currently possible to
configure, enforce a minimum distribution bit amount.


geirst (approved):
👍

geirst:
@baldersheim and @toregge FYI

PR #15252, opened 2020-11-10
log application instead of cluster name 

kkraune:


bratseth (commented):
Should you add both?

kkraune:
we can keep the cluster name logging as well, I removed it as its name will be logged for any Advice.dontScale/none but is is debug so does not matter

kkraune:
ping. easier to find log lines when also app name is logged

PR #15253, opened 2020-11-10
Allow creation of local session failing (might have been created already) 

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -129,6 +129,10 @@ public class SessionRepository {
         return localSessionCache.get(sessionId);
     }
 
+    public boolean localSessionExists(long sessionId) {
+        return localSessionCache.get(sessionId) != null;
+    }
+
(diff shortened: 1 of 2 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -628,7 +632,15 @@ public void createLocalSessionFromDistributedApplicationPackage(long sessionId)
             ApplicationId applicationId = sessionZKClient.readApplicationId()
                     .orElseThrow(() -> new RuntimeException("Could not find application id for session " + sessionId));
             log.log(Level.FINE, () -> "Creating local session for tenant '" + tenantName + "' with session id " + sessionId);
-            createLocalSession(sessionDir, applicationId, sessionId);
+            try {
+                createLocalSession(sessionDir, applicationId, sessionId);
+            } catch (RuntimeException e) {
```
jonmv, line 637: Isn't there a lock we can hold instead?

hmusum: Yes, that would be better, I'll look into it

PR #15260, opened 2020-11-10
Bjorncs/reindexing cluster controller in config model

bjorncs:


config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ReindexingController.java
```diff
@@ -0,0 +1,52 @@
...
+    private final String contentClusterName;
+    private final Collection<NewDocumentType> documentTypes;
+
+    ReindexingController(ReindexingContext context) {
+        super(new ComponentModel(
+                BundleInstantiationSpecification.getFromStrings(
+                        "reindexing-maintainer",
+                        "ai.vespa.reindexing.ReindexingMaintainer",
+                        REINDEXING_CONTROLLER_BUNDLE)));
+        this.reindexing = context.reindexing().orElse(null);
```
jonmv, line 30: Can use the default interface here?

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ReindexingController.java
```diff
@@ -0,0 +1,52 @@
...
+        this.contentClusterName = context.contentClusterName();
+        this.documentTypes = context.documentTypes();
+    }
+
+    @Override
+    public void getConfig(ReindexingConfig.Builder builder) {
+        builder.clusterName(contentClusterName);
+        if (reindexing == null || !reindexing.enabled()) {
```
jonmv, line 38: That eliminates the need for nullity checks.

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ReindexingController.java
```diff
@@ -0,0 +1,52 @@
...
+            builder.enabled(false);
+            return;
```
jonmv, line 40: And we _can_ send the timestamps for all documents when disabled, as the thing won't run anyway. 

I think we should probably settle on using either signal for disabling it.

jonmv (approved):

jonmv:
Let's consider using a `new Reindexing() { }` in place of `null`, and always sending the timestamps for each document type, since the `enabled bool` guards the actual reindexing anyway.

bjorncs:
@jonmv PTAL

jonmv (approved):
Unicorn

PR #15268, opened 2020-11-10
Support setting specific variants unoverridable 

bratseth:


container-search/src/main/resources/configdefinitions/search.query.profile.config.query-profiles.def
```diff
@@ -52,10 +52,14 @@ queryprofile[].queryprofilevariant[].inherit[] string
 queryprofile[].queryprofilevariant[].property[].name string
 # Content of profile variant
 queryprofile[].queryprofilevariant[].property[].value string
+# Whether the is overridable: "true", "false" or ""
```
arnej27959, line 55: "whether the" what? (missing, i would guess "property") "is overridable"

container-search/src/main/resources/configdefinitions/search.query.profile.config.query-profiles.def
```diff
@@ -52,10 +52,14 @@ queryprofile[].queryprofilevariant[].inherit[] string
...
+queryprofile[].queryprofilevariant[].property[].overridable string default=""
 # Content of profile variant
 queryprofile[].queryprofilevariant[].reference[].name string
 # Content of profile variant
 queryprofile[].queryprofilevariant[].reference[].value string
+# Whether the is overridable: "true", "false" or ""
```
arnej27959, line 61: "whether the" what? (missing, i would guess "reference") "is overridable"

arnej27959 (approved):
query profiles keep getting more and more complicated...
but apart from one very minor comment this looks legit to me

container-search/src/main/resources/configdefinitions/search.query.profile.config.query-profiles.def
bratseth: Fixed, thanks.

bratseth:
Yes, it would be nice to redo the whole thing, now that the full feature set is known, but I don't think I'll ever have the time.

PR #15269, opened 2020-11-10
We must never keep any DeployState instance around. They contain most…

baldersheim:
… of the world

and should be allowed to vanish when model has been built.

@jonmv or @bjorncs or @hmusum PR


bratseth (approved):

jonmv:
I _think_ there are more. Can take a pass.

PR #15270, opened 2020-11-11
Cut number of config rpc threads by half

baldersheim:
@hmusum PR
@jonmv FYI


hmusum (approved):

jonmv:
Nice if this is enough :)

PR #15277, opened 2020-11-11
Fix thread lock detection bug 

hakonhall:
The effect of the bug was that a deadlock would be reported as long as the
current thread T0 that tries to acquire the ZK path P0 is in the following
situation:
  1. Thread T0 tries to acquire ZK path P0, held by T1.
  2. Thread T1 tries to acquire ZK path P1, held by T2.

Instead, T2 would need to equal T0.  Or,
  3. Thread T2 tries to acquire ZK path P2, held by T3 = one of (T0, T1).
etc.


hmusum (approved):

hakonhall:
Thanks

PR #15278, opened 2020-11-11
Arnej/move dense xw

arnej27959:

@havardpe please review


eval/src/vespa/eval/instruction/dense_xw_product_function.cpp
```diff
@@ -1,23 +1,18 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
-#include "dense_xw_product_function.h"
-#include "dense_tensor_view.h"
+#include <vespa/eval/instruction/dense_xw_product_function.h>
```
havardpe, line 3: the header file should still be in the same directory as the cpp file and can be included with local syntax

havardpe (commented):
consider moving the test as well

arnej27959:
@havardpe please take a look at unit test changes also

havardpe (approved):

PR #15283, opened 2020-11-11
Enforce minimum 8 configured distributor split bits

vekterli:
@geirst please review

Even if the config model's distribution type now enforces a minimum
of 8 bits, it's still possible to violate this with explicit config
overrides. Ensure that we have a hard limit internally to be safe
even against this.


geirst (approved):
👍

geirst:
@baldersheim @toregge FYI

PR #15293, opened 2020-11-11
Revert "Remove implicit and potensially false dependency on jdisc_http_service"

bjorncs:
Reverts vespa-engine/vespa#15289


jonmv:
No idea

PR #15285, opened 2020-11-11
Bjorncs/install clustercontroller reindexer

bjorncs:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #15295, opened 2020-11-11
Revert "Bjorncs/install clustercontroller reindexer"

jonmv:
Reverts vespa-engine/vespa#15285

@bjorncs I suspect this broke some system tests, in a non-obvious way. Let's just revert to find out.


jonmv:
Looks like the failures were unrelated (⁎⚈᷀᷁ᴗ⚈᷀᷁⁎)

PR #15299, opened 2020-11-11
Set state for session instead of deleting it when deleting an applica…

hmusum:
…tion

Simplify by just setting the state of the active session belonging to an
application that is deleted. This is more in line with what we do for other
transitions.

Inactive sessions will be deleted by maintainers in config server after
some time.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
@@ -48,7 +48,9 @@ public class SessionStateWatcher {
         switch (newStatus) {
             case NEW:
             case NONE:
+                break;
             case DELETE:
+                sessionRepository.deactivateAndUpdateCache(session);
                 break;
             case PREPARE:
                 createLocalSession(sessionId);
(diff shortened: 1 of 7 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
```diff
@@ -48,6 +48,7 @@ private synchronized void sessionStatusChanged(Status newStatus) {
         switch (newStatus) {
             case NEW:
             case NONE:
+            case DELETE:
```
jonmv, line 51: Doing nothing on DELETE means we must be sure the DEACTIVATE has already been acted upon, and the session deactivated. How can we be sure this is the case here?

jonmv:
How long are inactive sessions kept? Do we need more aggressive GC to keep disk usage down?

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
hmusum: Thanks for reminding me, that was then one thing I was thinking "that needs to be done as well", but had forgotten about today.

hmusum:
Deleting an application is not done very often, whereas periodic redeployments are done often, so I don't think GC will need to change due to this

jonmv:
It's the opposite for CD systems, but I guess you're right anyway.

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionStateWatcher.java
hmusum: Fixed

jonmv:
👍

PR #15312, opened 2020-11-12
Unify syntax

baldersheim:
@vekterli PR
Syntax changes only


vekterli (approved):
:+1: I'm not a fan of enforced separate line for the return type, but it makes sense to stick with one style in a particular source file

PR #15321, opened 2020-11-12
Enforce minimum bucked used bits at document metastore load time

vekterli:
@toregge and @geirst please review


geirst (approved):
👍  @toregge please have a look as well

toregge (approved):

PR #15323, opened 2020-11-12
Reduce loglevel from error to info

baldersheim:
@arnej27959 PR
I was wondering if I should only lower the level for FRTE_RPC_METHOD_FAILED.
What is your take on that ?


arnej27959 (approved):
I think this is fine as-is.

PR #15324, opened 2020-11-12
Separate node failure maintenance from failing

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailureStatusUpdater.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.stream.Collectors;
+
+import static java.util.stream.Collectors.counting;
+
+/**
+ * Checks if nodes are responding and updates their status accordingly
+ *
+ * @author bratseth
+ */
+public class NodeFailureStatusUpdater extends NodeRepositoryMaintainer {
```
mpolden, line 34: Consider a shorter name. `NodeHealthTracker`?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -385,6 +385,18 @@ private NodeAcl getNodeAcl(Node node, NodeList candidates) {
...
+    /**
+     * Returns whether the zone managed by this node repository seems to be working.
+     * If too many nodes are not responding, there is probably some zone-wide issue
+     * and we should probably refrain from making changes to it.
+     */
+    public boolean isWorking() {
+        NodeList activeNodes = list(State.active);
+        if (activeNodes.size() <= 5) return true; // Not enough data to decide
+        NodeList downNodes = activeNodes.down();
+        return ! ( (double)downNodes.size() / (double)activeNodes.size() > 0.2 );
```
mpolden, line 397: Consider simplifying to `(double)downNodes.size() / (double)activeNodes.size() <= 0.2`.

mpolden (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
bratseth: I find "working if not more than 20% fails" is easier to get than the simpler one but that might just be me ...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailureStatusUpdater.java
bratseth: Good suggestion, I'll change it.

PR #15325, opened 2020-11-12
added spin lock with test

havardpe:
The micro-benchmark shows promise (run test manually with 'verbose' as
parameter). It indicates that taking and releasing the spin lock from
a single thread is cheaper than doing the same with a normal mutex (as
expected based on previous tests). However, it also indicates that it
is competitive with a normal mutex for short critical sections with
multiple threads involved. (on my laptop, only the cases using more
threads than cores run faster with a normal mutex, which is a better
result than expected).

@baldersheim @vekterli @arnej27959 @toregge please review

### DIFF
--- a/vespalib/src/tests/spin_lock/spin_lock_test.cpp
+++ b/vespalib/src/tests/spin_lock/spin_lock_test.cpp
@@ -1,0 +1,174 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/vespalib/util/spin_lock.h>
+#include <vespa/vespalib/util/benchmark_timer.h>
+#include <vespa/vespalib/util/time.h>
+#include <vespa/vespalib/testkit/test_kit.h>
+
+using namespace vespalib;
+
+bool verbose = false;
+double budget = 0.25;
+size_t thread_safety_work = 1000000;
+
+struct DummyLock {
+    void lock() {}
+    void unlock() {}
+};
+
+//-----------------------------------------------------------------------------
+
+struct MyState {
+    static constexpr size_t SZ = 5;
+    std::array<size_t,SZ> state = {0,0,0,0,0};
+    void update() {
+        std::array<size_t,SZ> tmp;
+        for (size_t i = 0; i < SZ; ++i) {
+            tmp[i] = state[i];
+        }
+        for (size_t i = 0; i < SZ; ++i) {
+            state[i] = tmp[i] + 1;
+        }
+    }
+    bool check(size_t expect) const {
+        for (size_t value: state) {
+            if (value != expect) {
+                return false;
+            }
+        }
+        return true;
+    }
+    void report(size_t expect, const char *name) const {
+        if (check(expect)) {
+            fprintf(stderr, "%s is thread safe\n", name);
+        } else {
+            fprintf(stderr, "%s is not thread safe\n", name);
+            fprintf(stderr, "    expected %zu, got [%zu,%zu,%zu,%zu,%zu]\n",
+                    expect, state[0], state[1], state[2], state[3], state[4]);
+        }
+    }
+};
+
+//-----------------------------------------------------------------------------
+
+template <typename T> void  basic_usage() {
+    T lock;
+    {
+        std::lock_guard guard(lock);
+    }
+    {
+        std::unique_lock guard(lock);
+    }
+}
+
+//-----------------------------------------------------------------------------
+
+template <typename T> size_t thread_safety_loop(T &lock, MyState &state, size_t thread_id, size_t thread_limit) {
+    size_t loop_cnt = (thread_safety_work / thread_limit);
+    TEST_BARRIER();
+    auto t0 = steady_clock::now();
+    TEST_BARRIER();
+    if (thread_id < thread_limit) {
+        for (size_t i = 0; i < loop_cnt; ++i) {
+            std::lock_guard guard(lock);
+            state.update();
+        }
+    }
+    auto t1 = steady_clock::now();
+    TEST_BARRIER();
+    if (thread_id == 0) {
+        auto t2 = steady_clock::now();
+        size_t total_ms = count_ms(t2 - t0);
+        fprintf(stderr, "---> thread_safety_loop with %zu threads used %zu ms\n", thread_limit, total_ms);
+    }
+    TEST_BARRIER();
+    if (verbose && (thread_id < thread_limit)) {
+        size_t local_ms = count_ms(t1 - t0);
+        fprintf(stderr, "    -- thread %zu used %zu ms\n", thread_id, local_ms);
+    }
+    TEST_BARRIER();
+    return (loop_cnt * thread_limit);
+}
+
+//-----------------------------------------------------------------------------
+
+template <typename T> void estimate_cost(const char *name) __attribute__((noinline));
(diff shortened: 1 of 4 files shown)


havardpe:
@toregge we might need to figure out a portable way of saying 'pause' if we are to compile on multiple platforms

vespalib/src/tests/spin_lock/spin_lock_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+    expect += thread_safety_loop(f1, f2, thread_id, 12);
+    expect += thread_safety_loop(f1, f2, thread_id, 6);
+    expect += thread_safety_loop(f1, f2, thread_id, 3);
+    if (thread_id == 0) {
+        f2.report(expect, "SpinLock");
+        EXPECT_TRUE(f2.check(expect));
+    }
+}
+
+TEST_MT_FF("require that SpinLock is thread safe", 24, std::mutex(), MyState()) {
```
arnej27959, line 148: SpinLock -> std::mutex

vespalib/src/vespa/vespalib/util/spin_lock.h
```diff
@@ -0,0 +1,33 @@
...
+namespace vespalib {
+
+/**
+ * A spin-lock implementation that favors uncontended performance.
+ * Some measures are taken to reduce the impact of threads waiting to
+ * get the lock since this will not affect the fast-path of obtaining
+ * the lock immediately.
+ *
+ * This implementation satisfies the BasicLockable requirements,
+ * making it work with things like std::lock_guard.
```
toregge, line 14: Note that the spin lock does not provide priority inheritance, a low priority thread preempted while holding the spin lock can cause excessive spinning in a high priority thread if the system has few (e.g. a single) cpu cores.

havardpe: does std::mutex provide this, even if std::thread does not provide priority?

vespalib/src/vespa/vespalib/util/spin_lock.h
```diff
@@ -0,0 +1,33 @@
...
+ **/
+class SpinLock {
+private:
+    std::atomic<bool> _lock;
```
vekterli, line 27: Could consider using [`std::atomic_flag`](https://en.cppreference.com/w/cpp/atomic/atomic_flag) here. It has a simpler interface than `std::atomic` but is the only atomic type guaranteed to be lock free (that means the `static_assert` below could be removed). The inner relaxed load loop seems to require a C++20 addition for it, so hopefully we have it..!

vespalib/src/vespa/vespalib/util/spin_lock.h
```diff
@@ -0,0 +1,33 @@
...
+public:
+    SpinLock() noexcept : _lock(false) {
+        static_assert(std::atomic<bool>::is_always_lock_free);
+    }
+    void lock() noexcept {
+        while (__builtin_expect(_lock.exchange(true, std::memory_order_acquire), false)) {
+            while (_lock.load(std::memory_order_relaxed)) {
+                __builtin_ia32_pause();
```
vekterli, line 35: `_mm_pause` may be more portable, but I believe it requires `xmmintrin.h` so if used maybe it should be wrapped away in its own translation unit to avoid header pollution.

vespalib/src/vespa/vespalib/util/spin_lock.h
```diff
@@ -0,0 +1,33 @@
...
+ **/
+class SpinLock {
+private:
+    std::atomic<bool> _lock;
+public:
+    SpinLock() noexcept : _lock(false) {
+        static_assert(std::atomic<bool>::is_always_lock_free);
+    }
+    void lock() noexcept {
+        while (__builtin_expect(_lock.exchange(true, std::memory_order_acquire), false)) {
```
vekterli, line 33: Are we on a new and cool™ enough version of GCC to be able to use the C++20 `[[likely]]` attribute?

vespalib/src/vespa/vespalib/util/spin_lock.h
havardpe: added more warnings to the comment

vespalib/src/tests/spin_lock/spin_lock_test.cpp
havardpe: fixed

havardpe:
@arnej27959 @toregge PTAL

arnej27959 (approved):
looks good to me

vekterli (approved):
:+1: a few comments for consideration ☕️

toregge (approved):
Consider occasionally calling std::this_thread::yield() while spinning on the lock.

vespalib/src/vespa/vespalib/util/spin_lock.h
```diff
@@ -0,0 +1,42 @@
...
+            while (_lock.load(std::memory_order_relaxed)) {
+                __builtin_ia32_pause();
```
baldersheim, line 35: Have you compared this to using yield ?

baldersheim (approved):

vespalib/src/vespa/vespalib/util/spin_lock.h
havardpe: I switched from atomic_flag to atomic_bool because I want to be able to read the variable without writing it to avoid totally messing up the cache. I do not think gcc is C++20 ready just yet, at least I do not think we have enabled it. It is highly unlikely that we want to run Vespa on a system that does not support atomic bool anyway.

vekterli: Yeah, strangely being able to do [`test(std::memory_order_relaxed)`](https://en.cppreference.com/w/cpp/atomic/atomic_flag/test) is very recent. Reading variables is cutting edge technology

vespalib/src/vespa/vespalib/util/spin_lock.h
havardpe: I could not figure out the subtly differences between _mm_pause and __builtin_ia32_pause across compilers, but concluded that if __builtin_ia32_pause is supported, it generates the appropriate code. I will let @toregge sink into this when he tries to compile Vespa on other platforms.

vespalib/src/vespa/vespalib/util/spin_lock.h
havardpe: I am not sure, I thought C++20 was off the table for now, but I might be wrong... @baldersheim ?

vespalib/src/vespa/vespalib/util/spin_lock.h
havardpe: turns out yield is much better than pause in all tested cases, I will switch. This also fixes portability; double win

vespalib/src/vespa/vespalib/util/spin_lock.h
baldersheim: We are on 9 still, but could try to enable c++20 if it has support for it.

havardpe: I could not find it on the list: (maybe it has some cool name)
https://gcc.gnu.org/projects/cxx-status.html

But it does seem some C++20 features are available in gcc 9 (although experimental). We might want to wait until it is called 20 and not 2a...

PR #15335, opened 2020-11-13
Use a thread safe class for holding header values

jonmv:
@bjorncs

### DIFF
--- a/jdisc_core/src/main/java/com/yahoo/jdisc/HeaderFields.java
+++ b/jdisc_core/src/main/java/com/yahoo/jdisc/HeaderFields.java
@@ -3,7 +3,12 @@ package com.yahoo.jdisc;
 
 import com.google.common.collect.ImmutableList;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentSkipListMap;
 
 /**
  * This is an encapsulation of the header fields that belong to either a {@link Request} or a {@link Response}. It is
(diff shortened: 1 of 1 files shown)


bjorncs (approved):

jdisc_core/src/main/java/com/yahoo/jdisc/HeaderFields.java
```diff
@@ -14,13 +19,7 @@
...
 public class HeaderFields implements Map<String, List<String>> {
 
-    private final TreeMap<String, List<String>> content = new TreeMap<>(new Comparator<String>() {
-
-        @Override
-        public int compare(String lhs, String rhs) {
-            return lhs.compareToIgnoreCase(rhs);
-        }
-    });
+    private final ConcurrentSkipListMap<String, List<String>> content = new ConcurrentSkipListMap<>(String::compareToIgnoreCase);
```
baldersheim, line 22: Won't this kill performance for high qps applications ?

jonmv: How many headers are there?

jonmv: We could synchronize the methods instead, I guess.

jonmv: Can't use a `syncrhonizedMap`, as this class implements `Map` itself, and exposes all views of the underlying map.

jonmv: Ah, there are unsynchronized `ArrayList` objects in the map, so need to do something about them as well.

PR #15337, opened 2020-11-13
Revert "Do not run in standalone compatibility mode for single ZooKeeper servers"

hmusum:
Reverts vespa-engine/vespa#15332


aressem (approved):

hmusum:
@mpolden FYI. id file is needed when `standaloneEnabled=false`, will fix in new PR

PR #15338, opened 2020-11-13
add instance, plan and planName to api response

andreer:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/billing/MockBillingController.java
```diff
@@ -31,6 +31,11 @@ public PlanId getPlan(TenantName tenant) {
         return plans.getOrDefault(tenant, PlanId.from("trial"));
     }
 
+    @Override
+    public String getPlanDisplayName(String planId) {
```
oyving, line 35: Use the type `PlanId`?

oyving (changes requested):

oyving (approved):

PR #15339, opened 2020-11-13
Synchronize instead, and copy defensively when returning results

jonmv:


jonmv:
Hmm, can still get a hold of the mutable internal objects here. Not too much point in copying on `get` then.

jonmv:
Hmm... Most fix without breaking things is probably to use a thread safe list in a thread safe map ...

PR #15340, opened 2020-11-13
Reapply "Do not run in standalone compatibility mode for single ZooKeeper servers"

hmusum:
Reverts vespa-engine/vespa#15337


mpolden (approved):
👍

PR #15348, opened 2020-11-15
Support mixed tensor type in Java evaluation

lesters:
@bratseth Please review.

Conformance system test (https://github.com/vespa-engine/system-test/blob/master/tests/search/tensor_eval/tensor_conformance.rb) passes with this. Seems like this flag (`supportsMixedTypes`) was a remnant from old.

This only fixes that the conformance test passes, not that mixed tensor evaluation in Java is efficient, as it falls back to mapped tensor implementations anyway.


lesters:
@arnej27959 FYI.

bratseth (approved):

PR #15351, opened 2020-11-16
Log more details

bratseth:
@kkraune I see no obvious cause for this and it works fine for other container clusters, so add more logging


kkraune (approved):
thanks!

PR #15356, opened 2020-11-16
Filter out metrics for nodes not in service

bratseth:


freva (approved):
What exactly is the issue, that `/state/v1/metrics` comes up long before `/status.html` or that metrics can be up without `/status.html` ever coming up?

bratseth:
The node sets itself in rotation when it has initialized completely, and has received an answer from content cluster nodes it should talk to. The former can take a long time, and the latter even longer with tx replay. In addition, the node can take itself out of rotation later, if backends become unresponsive (or if the local content node does, if this is a combined cluster with single-node-group content distribution).

PR #15357, opened 2020-11-16
Jonmv/reindexing metrics

jonmv:
@bjorncs please review.

### DIFF
--- a/clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/Reindexer.java
+++ b/clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/Reindexer.java
@@ -91,12 +91,16 @@ public class Reindexer {
             throw new IllegalStateException("Already shut down");
 
         try (Lock lock = database.lockReindexing()) {
+            Reindexing reindexing = updateWithReady(ready, database.readReindexing(), clock.instant());
+            database.writeReindexing(reindexing);
+            metrics.dump(reindexing);
+
             for (DocumentType type : ready.keySet()) { // We consider only document types for which we have config.
                 if (ready.get(type).isAfter(clock.instant()))
                     log.log(INFO, "Received config for reindexing which is ready in the future — will process later " +
                                   "(" + ready.get(type) + " is after " + clock.instant() + ")");
                 else
-                    progress(type);
+                    progress(type, new AtomicReference<>(reindexing), new AtomicReference<>(reindexing.status().get(type)));
 
                 if (phaser.isTerminated())
                     break;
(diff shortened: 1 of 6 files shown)


clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/ReindexingMetrics.java
```diff
@@ -0,0 +1,47 @@
...
+    private final String cluster;
+
+    ReindexingMetrics(Metric metric, String cluster) {
+        this.metric = metric;
+        this.cluster = cluster;
+    }
+
+    void dump(Reindexing reindexing) {
+        reindexing.status().forEach((type, status) -> {
+            metric.set("reindexing.percent.done",
```
bjorncs, line 29: Consider having "percent" removed from metric name, and instead represent it as a value between 0 and 1.

bjorncs (approved):

clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/ReindexingMetrics.java
geirst: Also, we should discuss the metric names in an architect meeting at some time.

jonmv: Yes, that’s the intention :) Feel free to reach out early with suggestions.

jonmv:
Actually metrics and names will wait until discussed properly.

clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/Reindexer.java
```diff
@@ -110,20 +107,20 @@ public void reindex() throws ReindexingLockException {
...
-                                                        .running()
-                                                        .successful(clock.instant()));
-        if (ready.get(type).isAfter(status.startedAt()))
-            status = Status.ready(clock.instant()); // Need to restart, as a newer reindexing is required.
-
-        database.writeReindexing(reindexing = reindexing.with(type, status));
-        metrics.dump(reindexing);
-
-        switch (status.state()) {
+        AtomicReference<Reindexing> reindexing = new AtomicReference<>(database.readReindexing());
```
bjorncs, line 110: You should probably use a shared monitor/lock for both `reindexing` and `status` (unless they are used completely  independently of each other).

bjorncs (approved):

clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/Reindexer.java
jonmv: They are used independently enough imo.

jonmv:
Second last commit makes sure the state of all pending reindexing are set to pending when a new directive is received, rather than exactly when reindexing of that type begins.

bjorncs (approved):

PR #15359, opened 2020-11-16
Handle session not existing when deleting app

hmusum:
Also make settings state for active session to DELETE part of the
nested transaction.

Add test for a scenario when deleting an application fails in provisioner,
failure is fixed and deleting the app afterwards should work.


jonmv (approved):
👍

PR #15360, opened 2020-11-17
Always include stack trace when throwing

hmusum:


baldersheim (commented):
I think this should be controllable by an environment setting or feature flag. I think it adds quite a bit of cost as a more expensive check must be done everytime it accesses an object.

bjorncs:
This should only have a penalty for construction of exceptions afaik. JVM will no longer throw a pre-constructed instance after disabling `OmitStackTraceInFastThrow`. I doubt we will see much impact unless we have hot spots where exceptions are part of the regular control flow.

bjorncs (approved):

hmusum:
Will merge this, since it should not have much impact in config server. We might want to introduce a feature flag so that this can be enabled in other containers as well (feature flag should probably be a generic flag for JVM options, so any option can be set)

PR #15366, opened 2020-11-17
Add curator config and populate it for config servers and cluster con…

hmusum:
…trollers


configdefinitions/src/vespa/curator.def
```diff
@@ -0,0 +1,7 @@
+# Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+namespace=cloud.config
+
+server[].hostname string
```
mpolden, line 4: Consider adding a comment explaining each field. E.g. that these are the servers curator will connect to.

mpolden (approved):

mpolden (approved):

PR #15369, opened 2020-11-17
Introduce node IP address pool

hakonhall:
A host that is supposed to run containers has a non-empty IP pool: A set of
IPv4 and/or IPv6 addresses that can be assigned to containers.

This PR adds a list of hostnames to this pool.  The intent is that the
hostnames and IPs match through resolving, but resolution may not yet be
available (until DNS changes propagate).

For now, only a list of hostnames are specified.  We may want to specify
(hostname, IP address) pairs or (hostname, IPv4, IPv6) triplets later, and the
serialization format allows for that by storing the hsotnames in an array of
objects, the object having a "hostname" field.

However the REST API is kept simpler for now: it exposes and allows patching of
an array of strings of a "additionalHostnames" field.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
@@ -26,6 +26,7 @@ import com.yahoo.slime.Slime;
 import com.yahoo.slime.SlimeUtils;
 import com.yahoo.slime.Type;
 import com.yahoo.vespa.hosted.provision.Node;
+import com.yahoo.vespa.hosted.provision.node.Address;
 import com.yahoo.vespa.hosted.provision.node.Agent;
 import com.yahoo.vespa.hosted.provision.node.Allocation;
 import com.yahoo.vespa.hosted.provision.node.Generation;
@@ -66,6 +67,8 @@ public class NodeSerializer {
     private static final String hostnameKey = "hostname";
     private static final String ipAddressesKey = "ipAddresses";
     private static final String ipAddressPoolKey = "additionalIpAddresses";
+    private static final String containersKey = "containers";
+    private static final String containerHostnameKey = "hostname";
     private static final String idKey = "openStackId";
     private static final String parentHostnameKey = "parentHostname";
     private static final String historyKey = "history";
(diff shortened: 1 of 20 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
```diff
@@ -358,6 +371,16 @@ private ClusterMembership clusterMembershipFromSlime(Inspector object) {
...
 
+    private List<Address> addressesFromSlime(Inspector object) {
+        Inspector addressesField = object.field(containersKey);
+        if (addressesField.children() == 0)
+            return List.of();
+        List<Address> addresses = new ArrayList<>(addressesField.children());
+        addressesField.traverse((ArrayTraverser) (i, elem) ->
+                addresses.add(new Address(elem.field(containerHostnameKey).asString())));
+        return addresses;
+    }
```
freva, line 382: Consider
```suggestion
    private List<Address> addressesFromSlime(Inspector object) {
        return SlimeUtils.entriesStream(object.field(containersKey))
                .map(elem -> new Address(elem.field(containerHostnameKey).asString()))
                .collect(Collectors.toList());
    }
```

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/AllocationSimulator.java
```diff
@@ -11,8 +11,9 @@
 import com.yahoo.vespa.hosted.provision.NodeList;
 import com.yahoo.vespa.hosted.provision.node.Allocation;
 import com.yahoo.vespa.hosted.provision.node.Generation;
+import com.yahoo.vespa.hosted.provision.node.IP;
 
-import javax.swing.JFrame;
+import javax.swing.*;
```
freva, line 16: Disable wildcard imports for `javax.swing.*`

freva (approved):

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/provisioning/AllocationSimulator.java
hakonhall: huh, that was a weird exception in IntelliJ - removed and fixed.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
hakonhall: Sure

mpolden (approved):

PR #15373, opened 2020-11-18
Activate only if all config change actions are allowed, for hosted de…

jonmv:
…ployment

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -321,11 +321,17 @@ public class ApplicationRepository implements com.yahoo.config.provision.Deploye
         return deploy(applicationPackage, prepareParams, DeployHandlerLogger.forPrepareParams(prepareParams));
     }
 
-    public PrepareResult deploy(File applicationPackage, PrepareParams prepareParams, DeployHandlerLogger logger) {
+    private PrepareResult deploy(File applicationPackage, PrepareParams prepareParams, DeployHandlerLogger logger) {
         ApplicationId applicationId = prepareParams.getApplicationId();
         long sessionId = createSession(applicationId, prepareParams.getTimeoutBudget(), applicationPackage);
         Deployment deployment = prepare(sessionId, prepareParams, logger);
-        deployment.activate();
+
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -321,11 +321,17 @@ public PrepareResult deploy(File applicationPackage, PrepareParams prepareParams
...
     }
 
-    public PrepareResult deploy(File applicationPackage, PrepareParams prepareParams, DeployHandlerLogger logger) {
+    private PrepareResult deploy(File applicationPackage, PrepareParams prepareParams, DeployHandlerLogger logger) {
         ApplicationId applicationId = prepareParams.getApplicationId();
         long sessionId = createSession(applicationId, prepareParams.getTimeoutBudget(), applicationPackage);
         Deployment deployment = prepare(sessionId, prepareParams, logger);
-        deployment.activate();
+
+        if (deployment.configChangeActions().getRefeedActions().getEntries().stream().anyMatch(entry -> ! entry.allowed()))
```
bratseth, line 329: I think this check should be down in validation somewhere, not in ApplicationRepository?
And isn't there a validate switch you should take into account?

jonmv: The actions are generated based on the validate switch down in validation. 
This is a higher-level evaluation. 
It could be in `Deployment.activate()`, but I think this is clearer, because that only conditionally runs `Deployment.prepare()`, which is the thing that sets these config change actions.

jonmv: This is just moving the logic that was previously in the controller (fail deployment if prepare reveals disallowed config change actions) closer to the action, so it actually works. Currently these deployments do happen, but the controller claims they didn't.

bratseth: Ok, I believe it was worse before. Even so, it could be that we can make it better than this.

[New!] Why is this warning and otherwise silently ignoring prepare? Prepare should either succeed or throw?

And why can't we do that down in the validation code like we do with everything else?

And should this really disregard the "validate" argument?

jonmv: Let's plug this hole now, and then do it properly. It may be some work.

hmusum (approved):

PR #15384, opened 2020-11-18
Non-functional-changes-only

bratseth:


container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
```diff
@@ -143,8 +144,12 @@ protected void configure() {
     }
 
     private static class SimpleContainerThreadpool implements ContainerThreadPool {
+
         private final Executor executor = Executors.newCachedThreadPool();
-        @Override public Executor executor() { return executor; }
```
bjorncs, line 147: I don't agree with this style change. The original style is more compact without impacting readability (imho).

container-core/src/main/java/com/yahoo/container/handler/threadpool/DefaultContainerThreadpool.java
```diff
@@ -49,9 +49,14 @@ public DefaultContainerThreadpool(ContainerThreadpoolConfig config, Metric metri
                 config.name(), config.queueSize());
     }
 
-    @Override public Executor executor() { return threadpool; }
-    @Override public void close() { closeInternal(); }
-    @Override public void deconstruct() { closeInternal(); super.deconstruct(); }
```
bjorncs, line 54: See earlier comment on style.

container-core/src/main/java/com/yahoo/container/handler/threadpool/ExecutorServiceWrapper.java
```diff
@@ -28,10 +28,12 @@
     private final Thread metricReporter;
     private final AtomicBoolean closed = new AtomicBoolean(false);
 
-    ExecutorServiceWrapper(
```
bjorncs, line 31: I don't like this style change either. The new style requires parameters to be reindented if `ExecutorServiceWrapper` is renamed or access modifier is introduced. It's annoying when a developer relies on automatic refactoring features in an IDE. It's also prone to create lines exceeding recommended max line length if either class name or parameter definitions are too long.

bjorncs (approved):
See comments. I don't agree with some of the style changes.

container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
bratseth: I agree with you actually but nobody else seem to like it so I've given up :-)

bjorncs: Why change it then? 😛

container-core/src/main/java/com/yahoo/container/handler/threadpool/ExecutorServiceWrapper.java
bratseth: I know you don't like it, and I see it as a taste thing where there's no right answer (the IDE can indent like this for you so it doesn't cost any time either way). But I'm an author on this file and you are not :-)

bjorncs: Does it matter who is formally listed as author of the file? Many source files are modified by several developers that are not listed as author.

container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
bratseth: If I'm the only one who wants some style I'll go for consistency instead.

bjorncs: FYI I have seen other Vespa devs using that style as well btw. Same applies to my indentation preference for parameter lists.

container-core/src/main/java/com/yahoo/container/handler/threadpool/ExecutorServiceWrapper.java
bratseth: I usually check authors if I want to change to a style I know people don't agree to. It just seems polite to me.

Btw, on author: I usually set myself as author if I make substantial changes to something.

None of these things are precise enough to be "rules" of course ...

container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
bratseth: Yes I think Valeri also like the wrong indentation style and you're both batshit crazy.
Let me change the @overrides back now that I know I have an ally on that :-)

container-core/src/main/java/com/yahoo/container/handler/threadpool/ExecutorServiceWrapper.java
bjorncs: I don't agree in changing code style if there is no intention to do functional changes to the same code - we do not have any agreed style convention or automatic linting for our Java code. It clutters the git history and creates potential merge conflicts. Is it constructive if several developers reformat each others code all the time because they don't the other style?

bratseth: No but I don't think that is an actual problem we have. It's more that we have lots of code that looks a bit old and crufty with lots of yellow when viewed in IntelliJ.

(Not that that's a big problem either - here I was killing time while in a meeting.)

PR #15392, opened 2020-11-19
Deprecate Curator

mpolden:
This moves all the implementation details of `Curator` to a non-public package,
while preserving the API.

The plan is to add a minimal interface for `VespaCurator` that we'll expose as a
public API.

Depends on internal PR due to the deprecation.

(The large diff is mainly due to refactoring of `MockCurator`)

@hmusum and @bratseth


bratseth:
I see that you have both moved the implementation of Curator into VespaCurator and also declares that the package containing Curator should be removed. Why do both? Seems like one or the other should be sufficient.

The Curator deprecation tells people to use com.yahoo.vespa.curator.impl.VespaCurator instead, but they can't since it isn't public.

What do you plan to remove from the API? It's hard to have an opinion on this without seeing the rest of the plan I think.

mpolden:
> Why do both? 

We want to reuse code, i.e. our own code should use the same implementation as users of the public API.

> The Curator deprecation tells people to use com.yahoo.vespa.curator.impl.VespaCurator instead

Yes, that comment is confusing. It was intended for usage in our own code. Will fix.

> What do you plan to remove from the API? It's hard to have an opinion on this without seeing the rest of the plan I think.

The plan is:
* Users inject `com.yahoo.vespa.curator.api.VespaCurator` (public API, implemented by `impl.VespaCurator`) that only exposes locking in the initial version. This interface can then grow based on user requirements.
* Our own code injects `com.yahoo.vespa.curator.impl.VespaCurator` and can access the full non-public API.
* `com.yahoo.vespa.curator.Curator` is removed in Vespa 8 (and package becomes non-public). The `impl` package could then be pulled up one-level, but that's not really important.
* On Vespa 8 `com.yahoo.vespa.curator.api` will be the only public API.

bratseth:
So the goal is 1) to provide just the locking part of the API as public, and 2) revoke public on com.yahoo.vespa.curator.
Makes sense, but wouldn't you achieve this just by adding com.yahoo.vespa.curator.api.VespaCurator (implemented by Curator), adding the deprecation com.yahoo.vespa.curator and nothing else?

mpolden:
Looks like that may be possible, but `Curator` needs to have `ConfigserverConfig` injected. I see now that the config-model already (and surprisingly) produces that config for container clusters: https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java#L559, but the initial plan was to use the new `CuratorConfig` for this.

bratseth:
Yes, that's a bit surprising, but I don't see the connection to this?

mpolden:
`VespaCurator` in this PR takes a `CuratorConfig` instead of `ConfigserverConfig`. Our idea was to change the config model to produce that `CuratorConfig` for container clusters instead adding `ConfigserverConfig` in even more non-ideal places, but it already does the latter... 😅

hmusum (approved):

bratseth:
Good plan but we can achieve that just by accepting both configs in Curator?

mpolden:
Sure, but does DI handle more than one constructor annotated with `@Inject`? If I remember correctly, it fails as it doesn't know which one to choose. I'm thinking that since the config-model already creates `ConfigserverConfig` we can keep using that for now.

bratseth:
> Sure, but does DI handle more than one constructor annotated with @Inject? 

No, but we can just change which one has the @Inject.

> I'm thinking that since the config-model already creates ConfigserverConfig we can keep using that for now.

Yes, it doesn't make it worse.

PR #15399, opened 2020-11-19
Jonmv/throw during prepare with disallowed config change actions

jonmv:
@bratseth please reivew. This will fails several system and integration tests, so let me merge when I'll be ready to watch these.

@geirst FYI.


jonmv:
I suppose it would have been easier to just throw on the first dissllowed action, though :) Perhaps that’s worth the reduced complexity.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
```diff
@@ -99,9 +105,17 @@
...
                 new NodeResourceChangeValidator()
         };
-        return Arrays.stream(validators)
-                .flatMap(v -> v.validate(currentModel, nextModel, overrides, now).stream())
-                .collect(toList());
+        List<ConfigChangeAction> actions = Arrays.stream(validators)
+                                                 .flatMap(v -> v.validate(currentModel, nextModel, overrides, now).stream())
+                                                 .collect(toList());
+        overrides.invalid(actions.stream()
+                                 .filter(DisallowableConfigChangeAction.class::isInstance)
```
bratseth, line 112: Given this usage, I think it would be simpler to just put validationId on ConfigChangeAction as an Optional.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
```diff
@@ -99,9 +105,17 @@
                 new ContainerRestartValidator(),
                 new NodeResourceChangeValidator()
         };
-        return Arrays.stream(validators)
-                .flatMap(v -> v.validate(currentModel, nextModel, overrides, now).stream())
-                .collect(toList());
+        List<ConfigChangeAction> actions = Arrays.stream(validators)
```
bratseth, line 108: Just to repeat my standard comment to your code: These stream statements that does everything are like a long sentence with lots of subsentences. It's easier to read shorter sentences, especially since it makes you name the thing they produce.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/VespaRefeedAction.java
```diff
@@ -22,37 +23,36 @@
...
      * which is model version independent.
      */
-    private final String name;
-
+    private final ValidationId validationId;
     private final String documentType;
     private final boolean allowed;
 
-    private VespaRefeedAction(ClusterSpec.Id id, String name, String message, List<ServiceInfo> services, String documentType, boolean allowed) {
+    private VespaRefeedAction(ClusterSpec.Id id, ValidationId validationId, String message, List<ServiceInfo> services, String documentType, boolean allowed) {
```
bratseth, line 30: 👍

bratseth (approved):
Great

bratseth:
You need to regen ABI spec on config-model-api

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
jonmv: I suppose that's equally not super-obvious, but good enough, and a bit simpler, yes. Can't even remember why I didn't do this. Probably no good reason. It's "you must provide a validatoin id for implementations which may block deployment", in any case.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
jonmv: But.
I.
Like.
Long.
Sentences.

I appreciate your comment on this; you're probably not the only one, so I will try to adjust for readability.

config-model/src/main/java/com/yahoo/vespa/model/application/validation/Validation.java
jonmv: Ah, so it's because making it optional makes you have to `.get()` or `.orElseThrow()` to get the `name()` in the child interfaces, which is ugly. It's clearer with a separate interface, although more verbose.

jonmv: Should just remove the `allowed()` altogether, since it no longer serves a purpose with these changes. Guess there'' be some more work here, then ...

jonmv:
@bratseth PTAL.

bratseth (approved):

PR #15404, opened 2020-11-20
Log every 10th run of SessionsMaintainer

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
```diff
@@ -25,6 +26,9 @@
 
     @Override
     protected boolean maintain() {
+        if (iteration % 10 == 0)
+            log.log(LogLevel.INFO, () -> "Running " + SessionsMaintainer.class.getSimpleName());
```
hakonhall, line 30: Consider including the iteration in the message, to make it clearer for the reader of the message it is actually running more often

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainer.java
hmusum: Good point, done

hakonhall (approved):

PR #15406, opened 2020-11-20
Get additional IP addresses and OSTK id for node

olaaun:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
```diff
@@ -57,14 +59,17 @@
     private final Optional<ApplicationId> exclusiveTo;
     private final Map<String, JsonNode> reports;
     private final List<NodeHistory> history;
+    public Set<String> additionalIpAddresses;
+    public String openStackId;
```
freva, line 63:
```suggestion
    private final Set<String> additionalIpAddresses;
    private final String openStackId;
```

freva (approved):

PR #15415, opened 2020-11-20
Allow allocating to a provisioned tenant host

hakonhall:
This PR changes (A) and (B) described below, which effectivelly allows
the allocation of nodes to a dynamically provisioned tenant host before
they have been assigned an IP pool.

A.  In NodePriotizer, when converting a VirtualNodeCandidate to a
ConcreteNodeCandidate with 'withNode()', primary IP addresses are currently
picked from the IP pool of the parent host and assigned to the node, along with
the corresponding hostname found using a DNS resolver.

This PR allows a ConcreteNodeCandidate to be created WITH a hostname, but
WITHOUT primary IP addresses, when 1. the parent host has no IP addresses in
the pool AND 2. there are hostnames in the address pool not yet assigned to any
child node.

This may happen for a brief period of time just after provisioning in a
dynamically provisioned zone: An asynchronous process is supposed to add IP
addresses to the pool (based on the pool hostnames) before the host can become
active, and also update the child nodes with their primary IP addresses
accordingly.  They both hold the unallocated lock to guarantee atomicity.

B.  In NodePriotizer.addCandidatesOnExistingHosts(), the
HostCapacity.hasCapacity() will now return true when (a) the parent host is
tenant, (b) there are NO IP addresses in the pool, and (c) there is a hostname
in the pool not assigned to

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -94,7 +94,7 @@ public final class Node implements Nodelike {
             requireNonEmpty(ipConfig.primary(), "Active node " + hostname + " must have at least one valid IP address");
 
         if (parentHostname.isPresent()) {
-            if (!ipConfig.pool().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
+            if (!ipConfig.pool().getIpSet().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
             if (modelName.isPresent()) throw new IllegalArgumentException("A child node cannot have model name set");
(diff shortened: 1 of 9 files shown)


freva (approved):
Not sure if you want to do this now or in a later PR, but can we stop using `ProvisionedHost::generateNode` and let `NodePrioritizer` make new node on the newly provisioned host the same way it will make for future deployments?

hakonhall:
I would need to look into that, so let me merge this now.

hakonhall:
Re suggestion: Currently
1. NodePriotizer is used to create a NodeAllocation,
2. if there is deficit, hosts are provisioned
3. and nodes put on those hosts.

Future deployments would allocate nodes on the same hosts in (1), but since that's before (2) I don't understand how we can "make new node on newly provisioned host the same way"?

freva:
Yeah, it would be to basically restart from 1) with the new hosts

PR #15416, opened 2020-11-21
Get method to return all entries from the compiled query profile object

srikanthnampelli:


bratseth (commented):
👍  could you run mvn install in the module and do what the error message tells you?

PR #15420, opened 2020-11-23
Remove duplicate validation messages

jonmv:
@bratseth or @bjorncs please review and merge. `? extends Collection<String>` should accommodate the existing `List<String>` used by current `config-model` code.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
@@ -14,6 +14,7 @@ import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


bratseth:
Wont that require recompilation of the old model?

jonmv:
Why? `Map<ValidationId, ? extends Collection<String>>` should be capable of accepting a `Map<ValidationId, List<String>>` ... ?

jonmv:
The caller is old, and passes an argument which is a `... List<String>` — this is fine with the old signature, which matches this exactly, _and_ with the new one.

bratseth:
Yes, but signature changes which are type-compatible may require recompiling the caller as the byte-code invoke is for the exact types of the signature it matches. I don't recall the exact rules so don't know if that is true in this case though.

jonmv:
I doubt it, since this is a generic signature.

jonmv:
Probably doesn't matter that the generic type matches, as it's erased at runtime anyway.

jonmv:
If this was a change from accepting a `List` to `Collection`, it could matter, and perhaps it wouldn't work. But the signature here is a `Map` in both cases, so I'd be very surprised if this wasn't OK.

bjorncs (approved):
The generic type of `Map` is erased at runtime.

PR #15435, opened 2020-11-23
Add VespaCurator interface

mpolden:
@hmusum

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
@@ -1,0 +1,19 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.curator.api;
+
+import com.yahoo.path.Path;
+
+import java.time.Duration;
+
+/**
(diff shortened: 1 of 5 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
```diff
@@ -0,0 +1,19 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.curator.api;
+
+import com.yahoo.path.Path;
```
hmusum, line 4: Could we use java.nio.file.Path instead? com.yahoo.path.Path is marked Beta and if we can use com.yahoo.path.Path internally only it would be nice. I haven't looked into if it's feasible, though

mpolden: `java.nio.file.Path` requires a `FileSystem` though, so not sure its suitable. `com.yahoo.path.Path` is better than `String` though.

hmusum: You are right, doesn't look like that will work. Agreed that it's better than `String`.

hmusum (approved):

PR #15437, opened 2020-11-23
Balder/some preliminary cleanup

baldersheim:
@vekterli PR


messagebus_test/src/tests/speed/cpp-client.cpp
```diff
@@ -76,7 +77,7 @@ Client::handleReply(Reply::UP reply) {
         for (uint32_t i = 0; i < reply->getNumErrors(); ++i) {
             fprintf(stderr, "ERR[%d]: code=%d, msg=%s\n", i,
                     reply->getError(i).getCode(),
-                    reply->getError(i).getMessage().c_str());
+                    reply->getError(i).getMessage().data());
```
vekterli, line 80: Consider wrapping in a temporary string, `data()` isn't guaranteed to be null-terminated

vekterli (approved):
:+1: one comment re: stringref usage with printf in a test

messagebus_test/src/tests/speed/cpp-client.cpp
baldersheim: Fixed

PR #15439, opened 2020-11-24
Log when running a maintainer

hmusum:


hakonhall (approved):
May want to log when returning too?

hakonhall (approved):

PR #15441, opened 2020-11-24
Report metric for non-active nodes

mpolden:
@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
@@ -3,6 +3,7 @@ package com.yahoo.vespa.hosted.provision.maintenance;
 
 import com.yahoo.component.Version;
 import com.yahoo.config.provision.ApplicationId;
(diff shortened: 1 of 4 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
```diff
@@ -70,9 +72,38 @@ public boolean maintain() {
         updateLockMetrics();
         updateDockerMetrics(nodes);
         updateTenantUsageMetrics(nodes);
+        updateAllocationMetrics(nodes);
         return true;
     }
 
+    private void updateAllocationMetrics(NodeList nodes) {
+        Map<ClusterKey, List<Node>> byApplication = nodes.stream()
```
bratseth, line 80: byCluster?

bratseth (approved):
👍

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
mpolden: Yes, thanks. Fixed!

PR #15443, opened 2020-11-24
Balder/gc loadtypes

baldersheim:
@vekterli PR


vekterli (approved):
🥳👍

kkraune:
so we should change doc as well? https://docs.vespa.ai/documentation/reference/services.html#clients

I assume we should state that "clients" is deprecated and will be removed in vespa 8 - and that using this has no effect?

baldersheim:
Not quite. The priority mapping will still work as before. The only difference is that you will not get a different metric per loadtype.

PR #15449, opened 2020-11-24
Move the error description to a separate allocation as it is rarely u…

baldersheim:
…sed.

This reduces the size of frequently used objects.

@geirst PR

### DIFF
--- a/storageapi/src/vespa/storageapi/messageapi/returncode.cpp
+++ b/storageapi/src/vespa/storageapi/messageapi/returncode.cpp
@@ -10,17 +10,39 @@ ReturnCode::ReturnCode()
       _message()
 {}
 
-ReturnCode::ReturnCode(const ReturnCode &) = default;
-ReturnCode & ReturnCode::operator = (const ReturnCode &) = default;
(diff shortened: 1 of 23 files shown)


geirst (approved):
👍  @vekterli FYI

storageapi/src/vespa/storageapi/messageapi/returncode.cpp
```diff
@@ -145,5 +167,21 @@ ReturnCode::isBucketDisappearance() const
             return false;
     }
 }
+namespace {
+    vespalib::stringref Empty("");
+}
+vespalib::stringref
+ReturnCode::getMessage() const {
+    return _message ? _message->operator vespalib::stringref() : Empty;
```
vekterli, line 175: Is the global object really needed here, vs. just returning a default-constructed `stringref`?

vekterli (commented):
:+1:

storageapi/src/vespa/storageapi/messageapi/returncode.cpp
baldersheim: No, you are right. Old habit....

PR #15450, opened 2020-11-24
Support marking configs as apply on restart

bratseth:
This replaces the hack using restartOnDeploy to defer config changes until restart when appropriate with a proper mechanism.

Next I want to remove internalRedeploy as that distinction has become outdated by autoscaling.


hmusum (approved):

jonmv (approved):
Nice!

PR #15462, opened 2020-11-25
Add exception to log message

hmusum:
Hard to figure out what's wrong when you only get the message, so add exception so that stack trace will logged


jonmv (approved):

jonmv:
Suspect we'll get a lot of logging from this, but can revert later if needed.

PR #15466, opened 2020-11-25
Simplify storage message address

geirst:
@vekterli please review
@baldersheim FYI

### DIFF
--- a/storageapi/src/vespa/storageapi/messageapi/storagemessage.cpp
+++ b/storageapi/src/vespa/storageapi/messageapi/storagemessage.cpp
@@ -154,13 +154,8 @@ createAddress(vespalib::stringref cluster, const lib::NodeType& type, uint16_t i
 size_t
 calculate_node_hash(const lib::NodeType& type, uint16_t index)
 {
-    struct NodeValue {
-        uint16_t type;
-        uint16_t index;
-        NodeValue(const lib::NodeType& type_in, uint16_t index_in) : type(type_in), index(index_in) {}
-    };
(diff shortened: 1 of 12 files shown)


storageapi/src/vespa/storageapi/messageapi/storagemessage.cpp
```diff
@@ -160,39 +151,50 @@ createAddress(vespalib::stringref cluster, const lib::NodeType& type, uint16_t i
...
+size_t
+calculate_node_hash(const lib::NodeType& type, uint16_t index)
+{
+    struct NodeValue {
+        uint16_t type;
+        uint16_t index;
+        NodeValue(const lib::NodeType& type_in, uint16_t index_in) : type(type_in), index(index_in) {}
+    };
+    NodeValue value(type, index);
+    return vespalib::hashValue(&value, sizeof(NodeValue));
```
vekterli, line 163: To avoid having to contemplate any struct alignment issues and how they may affect the resulting value, could be simplified down to
```c++
uint16_t buf[] = { type, index };
return vespalib::hashValue(&buf, sizeof(buf));
```

vekterli (approved):
:+1:

storageapi/src/vespa/storageapi/messageapi/storagemessage.cpp
geirst: Nice, will fix.

baldersheim (approved):
Perfect :)

PR #15472, opened 2020-11-25
Bjorncs/config convergence checker

bjorncs:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
@@ -1,26 +1,33 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.config.server.application;
 
-import ai.vespa.util.http.VespaClientBuilderFactory;
+import ai.vespa.util.http.VespaAsyncHttpClientBuilder;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
 import com.yahoo.component.AbstractComponent;
+import com.yahoo.concurrent.DaemonThreadFactory;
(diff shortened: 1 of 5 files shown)


hmusum (approved):
Great! 💯

configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
```diff
@@ -108,46 +117,67 @@ public JSONResponse getServiceConfigGenerationResponse(Application application,
...
+            if (statePort <= 0) continue;
+
+            URI uri = URI.create("http://" + service.getHostName() + ":" + statePort);
+            CompletableFuture<Void> inprogressRequest = getServiceGeneration(uri, timeout)
+                    .handle((result, error) -> {
+                        if (result != null) {
+                            temporaryResult.put(service, result);
+                        } else {
+                            log.log(
+                                    LogLevel.DEBUG,
```
jonmv: `Level.FINE`

jonmv:
Nice :)

configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
bjorncs: I think they are equivalent (`LogLevel` being the Vespa wrapper for `Level`). Any benefits using `Level` instead?

jonmv: I know @gjoranv went to great lengths to replace with `Level` values some time ago. I don't know why, though.

bjorncs: Ok, I removed use of `LogLevel` in new PR.

gjoranv: I had to remove it from config in order to use config inside `vespalog`, and then it was proposed to remove `LogLevel` usage from all modules. I think the motivation was that custom log levels are unnecessary and a bit odd.

PR #15481, opened 2020-11-26
Only the cluster name is ever a valid ID here anyway

jonmv:
@vekterli I would like to verify this assumption before I start the cleanup work :)

### DIFF
--- a/documentapi/abi-spec.json
+++ b/documentapi/abi-spec.json
@@ -1527,7 +1527,6 @@
     ],
     "methods": [
       "public void <init>(java.util.Map)",
-      "public java.lang.String getDistributionConfigId()",
       "public com.yahoo.documentapi.messagebus.protocol.StoragePolicy$SlobrokHostPatternGenerator createPatternGenerator()"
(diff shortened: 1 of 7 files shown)


jonmv:
Perhaps @baldersheim would like to review instead?

jonmv:
There's something in C++ as well ...

jonmv:
May need to update unit tests for the C++ part.

jonmv:
Hmm... The C++ implementation already correctly handles this for `content`.

baldersheim:
Yes, I will complete the c++ part and cleanup.

jonmv:
Thanks.

jonmv:
Will need to replace the use of :raw config subscription in unit tests in C++ for this to work. Far beyond the original scope >_< Closing for now. What I _really_ wanted was to fix the default in `StoragePolicy` so it actually works, so it doesn't have to be specified everywhere. But perhaps we can just use `ContentPolicy` instead, everywhere ...

PR #15486, opened 2020-11-26
Setup components and config for container cluster when zookeeper is c…

hmusum:
…onfigured

Feature flag probably needed only short-term the way it is done now, but might be useful when we start testing this


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -218,7 +218,8 @@ private void addZooKeeper(ApplicationContainerCluster cluster, Element spec) {
                                                MIN_ZOOKEEPER_NODE_COUNT + " and " + MAX_ZOOKEEPER_NODE_COUNT);
         }
         cluster.addSimpleComponent("com.yahoo.vespa.curator.Curator", null, "zkfacade");
-        // TODO: Add server component
+        cluster.addSimpleComponent("com.yahoo.vespa.curator.ReconfigurableVespaZooKeeperServer", null, "zkfacade");
```
mpolden, line 221: Wrong package and bundle?

hmusum: Yes, will fix

mpolden (approved):
👍

hmusum:
Build still pending (!?), merging

PR #15479, opened 2020-11-26
Bjorncs/support sanuri rules

bjorncs:
FYI @tokle

PR #15487, opened 2020-11-26
Revert "Bjorncs/support sanuri rules"

hmusum:
Reverts vespa-engine/vespa#15479

Build fails with several similar to this one:
vespa-yahoo/hosted/host-admin/src/main/java/com/yahoo/vespa/hosted/hostadmin/task/security/TlsAuthorizationRulesGenerator.java:[10,37] com.yahoo.security.tls.policy.HostGlobPattern is not public in com.yahoo.security.tls.policy; cannot be accessed from outside package


bjorncs:
Code in `TlsAuthorizationRulesGenerator.java` is now fixed

PR #15489, opened 2020-11-26
GC unused html metrics reporting.

baldersheim:
@vekterli PR - Pure cleanup


storage/src/vespa/storage/common/statusmetricconsumer.cpp
```diff
@@ -155,160 +75,76 @@ StatusMetricConsumer::reportStatus(std::ostream& out,
...
-    bool xml = (!html && path.getAttribute("format") == "xml");
-    bool json = (!html && path.getAttribute("format") == "json");
+    bool xml = (path.getAttribute("format") == "xml");
+    bool json = (path.getAttribute("format") == "json");
 
     int verbosity(path.get("verbosity", 0));
         // We have to copy unset values if using HTML as HTML version gathers
         // metrics for calculations and thus needs unset values.
-    bool copyUnset = (html || verbosity >= 2);
+    bool copyUnset = (verbosity >= 2);
```
baldersheim, line 87: I wonder about this verbosity >= 2 which triggers copying and constitutes 2/3 of the remaining complexity.
Is that necessary or could it be removed too ? Less is more.

storage/src/vespa/storage/common/statusmetricconsumer.cpp
```diff
@@ -62,85 +60,7 @@ StatusMetricConsumer::getReportContentType(const framework::HttpUrlPath& path) c
...
-                linked = true;
-                out << "<a href=\""
-                    << "?interval=" << interval
-                    << "&format=" << formatStr
-                    << "&consumer=" << consumer
-                    << "&verbosity=0"
-                    << "&pattern=.*"
-                    << "&callsnapshothooks=0"
-                    << "&tmpsnapshot=" << tmp
-                    << "\">"
```
vekterli, line 130: It would be good to have _some_ form of output on the main metrics page that gives the parameters to use for text metrics dump of the past 1 minute, 5 minutes and all-time. This can be hard-coded. It's just hard to remember the URLs by heart...!

vekterli (approved):
:+1: see comment on adding a few lines of output to the metrics index page that gives the parameters to do a manual text dump

storage/src/vespa/storage/common/statusmetricconsumer.cpp
baldersheim: K, I will leave this as is for now.

storage/src/vespa/storage/common/statusmetricconsumer.cpp
baldersheim: I will merge this one now, and we can discuss the details for a fixed text tomorrow.

PR #15499, opened 2020-11-27
Set id for each zookeeper server

hmusum:


config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainer.java
```diff
@@ -70,6 +73,13 @@ public String getJvmOptions() {
     }
 
     private boolean hasDocproc() {
-        return (parent instanceof ContainerCluster) && (((ContainerCluster)parent).getDocproc() != null);
+        return (parent instanceof ContainerCluster) && (((ContainerCluster<?>)parent).getDocproc() != null);
     }
+
+    @Override
+    public void getConfig(ZookeeperServerConfig.Builder builder) {
+        ((ApplicationContainerCluster)parent).getConfig(builder);
```
mpolden, line 81: Is the parent guaranteed to be this type and non-null?

hmusum: Well, better to check anyway, I'll fix.

mpolden (approved):

PR #15500, opened 2020-11-27
Handle event data being null

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
@@ -186,18 +186,17 @@ public class TenantApplications implements RequestHandler, HostValidator<Applica
 
     private void childEvent(CuratorFramework ignored, PathChildrenCacheEvent event) {
         zkWatcherExecutor.execute(() -> {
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/application/TenantApplications.java
```diff
@@ -185,7 +186,9 @@ public Lock lock(ApplicationId id) {
 
     private void childEvent(CuratorFramework ignored, PathChildrenCacheEvent event) {
         zkWatcherExecutor.execute(() -> {
-            ApplicationId applicationId = ApplicationId.fromSerializedForm(Path.fromString(event.getData().getPath()).getName());
+            ChildData data = event.getData();
+            if (data == null) return; // Node might have been deleted after we got event
```
hakonhall, line 190: null is passed when connection is suspended, lost, or on (re)connected, or when cache is initialized(?)  So I think the correct thing to do is to duplicate the current statement to within each `case` in the below switch (and document getData may be null on some other event types)

hmusum: Right, thanks, will do as suggested

hakonhall (approved):

PR #15501, opened 2020-11-27
Reduce reindexer dynamic throttle policy weight

jonmv:
@vekterli please marge.


jonmv:
Hmm, the min window size is also pretty large ... concurrency of 20 isn't nothing.

vekterli (approved):
:+1: Magic dials and buttons 🦄

PR #15507, opened 2020-11-27
Read stateful tag in ClusterMembership

mpolden:
The string format in `ClusterMembership` is a bit wonky, so have to deploy
read-support before doing more changes.

@hmusum


hmusum (approved):
👍

PR #15508, opened 2020-11-27
Bratseth/apply on restart take 3

bratseth:


hmusum (approved):

bratseth:
@hmusum please review ... next try :-)

hmusum (commented):
Probably ok, really hard figure what the correct way of doing the deps are

bratseth:
Yes, because configgen isn't a bundle. But the real source of trouble is that the object model of a def file is in configgen!

PR #15511, opened 2020-11-27
Havardpe/reference evaluation

havardpe:

@arnej27959 please review

### DIFF
--- a/eval/src/vespa/eval/eval/test/reference_operations.cpp
+++ b/eval/src/vespa/eval/eval/test/reference_operations.cpp
@@ -232,9 +232,6 @@ TensorSpec ReferenceOperations::peek(const TensorSpec &param, const PeekSpec &pe
             result.add(my_addr, cell.second);
         }
     }
-    if (result.cells().empty() && result_type.is_scalar()) {
-        result.add({}, 0.0);
-    }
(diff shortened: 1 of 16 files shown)


eval/src/vespa/eval/eval/test/reference_operations.cpp
```diff
@@ -283,5 +283,32 @@ TensorSpec ReferenceOperations::rename(const TensorSpec &a, const std::vector<ve
     return result;
 }
 
+TensorSpec ReferenceOperations::lambda(const vespalib::string &type_in, lambda_fun_t fun) {
+    ValueType type = ValueType::from_spec(type_in);
+    TensorSpec result(type.to_spec());
+    if (type.is_error()) {
+        return result;
+    }
+    std::vector<size_t> sizes;
```
arnej27959, line 292: adding `const auto & dimensions = type.dimensions()` here would improve readability;
note there's no need for the `sizes` vector, you can just use dimensions[idx].size directly.

arnej27959 (approved):
it all looks quite straight-forward, even the self-capturing lambda is quite easy to understand (but I could definitely not have written it).  Will merge this now and look at the peek renumbering.

PR #15517, opened 2020-11-28
Get method to return all variants of value

srikanthnampelli:


bratseth:
Sorry, I missed this one earlier.

I consider these indexes an implementation detail so I would prefer not to expose them in a public API, what will you use it for?

PR #15528, opened 2020-11-30
Generate correct zookeeper server config and fix setup of components

hmusum:
Copmonents need to created so that they will subscribe with a config id
that is unique per container (since config is different for each container)

Also make sure to to only generate needed ´myid´ in config for each container.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainer.java
@@ -78,9 +78,6 @@ public final class ApplicationContainer extends Container implements
 
     @Override
     public void getConfig(ZookeeperServerConfig.Builder builder) {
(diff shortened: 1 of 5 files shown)


mpolden (approved):
Great!

config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainer.java
```diff
@@ -81,8 +81,6 @@ public void getConfig(ZookeeperServerConfig.Builder builder) {
         AbstractConfigProducer<?> parent = getParent();
         if (parent == null) return;
```
mpolden, line 82: Perhaps obsolete now?

hmusum: Yes, that's right, thanks!

mpolden (approved):

PR #15530, opened 2020-11-30
Balder/wakeuptrigger java

baldersheim:
@havardpe PR

### DIFF
--- a/container-messagebus/src/main/resources/configdefinitions/container.jdisc.container-mbus.def
+++ b/container-messagebus/src/main/resources/configdefinitions/container.jdisc.container-mbus.def
@@ -15,6 +15,9 @@ numthreads int default=2
 # Optimize for latency, or throughput.
 optimize_for enum {LATENCY, THROUGHPUT} default=LATENCY
 
+# Number of writeevents before triggering wakeup of network thread.
+wakeup_trigger_count int default=1
+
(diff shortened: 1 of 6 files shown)


container-messagebus/src/main/resources/configdefinitions/container.jdisc.container-mbus.def
```diff
@@ -15,6 +15,9 @@ numthreads int default=2
 # Optimize for latency, or throughput.
 optimize_for enum {LATENCY, THROUGHPUT} default=LATENCY
 
+# Number of writeevents before triggering wakeup of network thread.
+wakeup_trigger_count int default=1
```
havardpe, line 19: transport_events_before_wakeup

havardpe (commented):
we should probably also start using a Params object in both Java and C++, probably with delayed resolving of crypto engine (and resolver for C++)

havardpe:
this way looks good, I think we should do it like this in C++ as well

container-messagebus/src/main/resources/configdefinitions/container.jdisc.container-mbus.def
baldersheim: Fixed

baldersheim:
I will followup and complete c++ in the same pattern.
Then I will do parameter cleanup in both.

PR #15538, opened 2020-11-30
Arnej/less simple tensor engine

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/eval/tensor_spec.cpp
+++ b/eval/src/vespa/eval/eval/tensor_spec.cpp
@@ -4,13 +4,13 @@
 #include "array_array_map.h"
 #include "function.h"
 #include "interpreted_function.h"
-#include "simple_value.h"
 #include "tensor.h"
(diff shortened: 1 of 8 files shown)


eval/src/vespa/eval/eval/tensor_spec.cpp
```diff
@@ -208,7 +208,8 @@ TensorSpec::from_expr(const vespalib::string &expr)
     NoParams no_params;
     auto fun = Function::parse(expr);
     if (!fun->has_error() && (fun->num_params() == 0)) {
-        InterpretedFunction ifun(SimpleTensorEngine::ref(), *fun, NodeTypes());
+        NodeTypes node_types(*fun, {});
+        InterpretedFunction ifun(SimpleValueBuilderFactory::get(), *fun, node_types);
```
havardpe, line 212: here I think we should use the reference evaluation instead

havardpe (approved):

PR #15539, opened 2020-11-30
Bratseth/filter unstable later

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
```diff
@@ -63,6 +63,7 @@ protected boolean maintain() {
     private Collection<Pair<String, MetricSnapshot>> filter(Collection<Pair<String, MetricSnapshot>> snapshots) {
```
freva, line 63: This is unused now?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -152,7 +154,8 @@ private void ensureExists(String tableName) {
 
         try (SqlCompiler compiler = new SqlCompiler(engine)) {
             compiler.compile("create table " + tableName +
-                             " (hostname string, at timestamp, cpu_util float, mem_total_util float, disk_util float, application_generation long)" +
+                             " (hostname string, at timestamp, cpu_util float, mem_total_util float, disk_util float," +
```
freva, line 157: What about the existing tables?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
bratseth: Yes, removed thanks

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
bratseth: They will get default values, which is not what we want here (we want true not false), but the only effect is no autoscaling for a while, so I think that's okay.

freva: So as soon as you add a row with 6th column, that is legal even though the table was created with 5 columns, and all existing rows will get a default value of the type of the 6th column in the new row?

bratseth: Apparently - I tested it. Couldn't find any documentation on this scenario.

PR #15545, opened 2020-11-30
Wakeup every 10 events

baldersheim:
@havardpe PR


havardpe:
ok, would this be something for @vekterli to review?

PR #15546, opened 2020-11-30
Support lower bound on number of shared hosts

hakonhall:
Adds a 'minCount' field to the shared host jackson flag, denoting the minimum
number of "shared hosts" that must exist, otherwise the deficit will be
provisioned by DynamicProvisioningMaintainer.

A "shared host" is one that is considered for allocation if current tenant node
allocations were removed: It must be a tenant host, cannot be an exclusiveTo
host, etc.

minCount requires the setting of (at least one) shared host.


freva (approved):
:100:

PR #15556, opened 2020-12-01
Fix reconfiguration

hmusum:

### DIFF
--- a/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
+++ b/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
@@ -27,6 +27,7 @@ public class Reconfigurer extends AbstractComponent {
     private static final int sessionTimeoutInSeconds = 30;
 
     private ZooKeeperRunner zooKeeperRunner;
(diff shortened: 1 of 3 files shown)


zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
```diff
@@ -88,10 +83,48 @@ private String connectionSpec(ZookeeperServerConfig config) {
...
 
-    private List<String> servers(ZookeeperServerConfig config) {
+    private static List<String> servers(ZookeeperServerConfig config) {
         return config.server().stream()
                 .map(server -> server.hostname() + ":" + server.quorumPort() + ":" + server.electionPort())
                 .collect(Collectors.toList());
     }
 
+    ZookeeperServerConfig existingConfig() {
+        return zooKeeperRunner.zookeeperServerConfig();
```
mpolden, line 93: Is this ever updated Looks like it's only set when starting the server, so `shouldReconfigure` is always comparing the new config to the first config it received, not the current.

hmusum: You are right, that's wrong.

mpolden (approved):

PR #15562, opened 2020-12-01
Bratseth/questdb robustness

bratseth:
I get this error consistently in CD but not in production ¯\_(ツ)_/¯
And I'm unable to reproduce it. In any case since this isn't critical data it is nice to have a recovery mechanism.


freva:
The first commit shouldn't be here...

bratseth:
Thanks, I'll rebase

PR #15564, opened 2020-12-01
Remove lock metrics not used in dashboard

hakonhall:
Reduces the number of lock metrics set in MetricsReporter from 19 to 8.

Most/all metrics in MetricsReporter ought to cause 1 metric to be emitted, like the lock metrics. However the framework emits 3 (last, average, and max), and it's not easily fixed.


hakonhall:
Reduces the number of metrics set in MetricsReporter from 19 to 8.

Most/all metrics in MetricsReporter ought to cause 1 metric to be emitted, like the lock metrics. However the framework emits 3 (last, average, and max), and it's not easily fixed.

mpolden (approved):

PR #15565, opened 2020-12-01
Move code in clustercontroller-apputils into clustercontroller-apps

hmusum:
Code in clustercontroller-apputils is now only used from clustercontroller-apps,
so those two modules can be merged


vekterli (approved):
:+1:

hakonhall (approved):

PR #15569, opened 2020-12-01
Only report non-zero lock metrics

hakonhall:
This should reduce the number of reported lock metrics to 3%.

Unfortunately this will mean the graphs will not drop to zero at these times, but will be completely missing.


mpolden (approved):
👍

PR #15571, opened 2020-12-01
- Separate out the tail of the EventLoop into separate method to ease…

baldersheim:
… readability.

- Also separate out timeout handling of old connections to separate method and unify on vespalib::time.

@havardpe PR

### DIFF
--- a/fnet/src/vespa/fnet/transport_thread.cpp
+++ b/fnet/src/vespa/fnet/transport_thread.cpp
@@ -18,7 +18,6 @@ LOG_SETUP(".fnet");
 using vespalib::ServerSocket;
 using vespalib::SocketHandle;
 using vespalib::SocketSpec;
-using OptimizeFor = vespalib::Executor::OptimizeFor;
 using vespalib::steady_clock;
 
 namespace {
--- a/fnet/src/vespa/fnet/transport_thread.cpp
+++ b/fnet/src/vespa/fnet/transport_thread.cpp
@@ -19,6 +19,7 @@ using vespalib::ServerSocket;
 using vespalib::SocketHandle;
 using vespalib::SocketSpec;
 using OptimizeFor = vespalib::Executor::OptimizeFor;
+using vespalib::steady_clock;
 
 namespace {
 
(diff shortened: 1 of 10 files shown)


fnet/src/vespa/fnet/transport_thread.cpp
```diff
@@ -19,6 +19,7 @@ using vespalib::ServerSocket;
 using vespalib::SocketHandle;
 using vespalib::SocketSpec;
 using OptimizeFor = vespalib::Executor::OptimizeFor;
```
havardpe, line 21: We no longer need OptimizeFor

baldersheim: Fixed

fnet/src/vespa/fnet/transport_thread.h
```diff
@@ -156,6 +154,9 @@ class FNET_TransportThread : public FastOS_Runnable
      **/
     bool InitEventLoop();
 
+    void endEventLoop();
```
havardpe, line 157: should not be public

fnet/src/vespa/fnet/transport_thread.h
```diff
@@ -156,6 +154,9 @@ class FNET_TransportThread : public FastOS_Runnable
...
+    void checkTimedoutComponents(vespalib::duration timeout);
```
havardpe, line 158: should not be public

fnet/src/vespa/fnet/transport_thread.h
baldersheim: It is in the private: section so I think this is good.

fnet/src/vespa/fnet/transport_thread.h
baldersheim: It is in the private: section so I think this is good.

havardpe: this should not be together with the event loop external API, then

fnet/src/vespa/fnet/transport_thread.h
havardpe: this should not be together with the event loop external API, then

havardpe: my bad, I forgot we removed the single-threaded option a long time ago

fnet/src/vespa/fnet/transport_thread.h
havardpe: my bad, I forgot we removed the single-threaded option a long time ago

PR #15572, opened 2020-12-01
Initial support for backend reindexing visitor functionality

vekterli:
@jonmv FYI

Introduces the concept of a read-for-write visitor operation which
blocks all mutating operations from starting for a bucket being
visited. This read-for-write mode is used if (and only if) the visitor
library being specified by the client is `"reindexingvisitor"`.

Since read-for-write visitors cannot race with concurrent write
operations, starting such visitors are deferred until no further
mutations are pending.

Also adds a basic `reindexingvisitor` implementation to the content node
which sends all documents as Puts containing a special TaS token
that will let the operation through even if a bucket is locked. This
token is cleared by the distributor before it is passed on to the
content nodes.

Note: this feature is not yet production ready. For now the following
caveats apply:
* Mutating vs non-mutating pending ops to a bucket are not tracked
  separately, so it’s possible to starve a reindexing visitor by
  sending constant pending read load, as read load is not blocked by
  the operation sequencer.
* Ideal state operations towards locked buckets are not blocked, so
  it's possible for e.g. a split to be sent for a bucket that is being
  visited.

### DIFF
--- a/storage/src/vespa/storage/distributor/pendingmessagetracker.h
+++ b/storage/src/vespa/storage/distributor/pendingmessagetracker.h
@@ -22,12 +22,21 @@
 
 namespace storage::distributor {
 
+/**
+ * Since the state a deferred task depends on may have changed between the
+ * time a task was scheduled and when it's actually executed, this enum provides
+ * a means of communicating if a task should be started as normal.
+ */
 enum class TaskRunState {
-    OK,
-    Aborted,
-    BucketLost
+    OK,        // Task may be started as normal
+    Aborted,   // Task should trigger an immediate abort behavior (distributor is shutting down)
+    BucketLost // Task should trigger an immediate abort behavior (bucket no longer present on node)
 };
 
+/**
+ * Represents an arbitrary task whose execution may be deferred until no
+ * further pending operations are present.
+ */
 struct DeferredTask {
     virtual ~DeferredTask() = default;
     virtual void run(TaskRunState state) = 0;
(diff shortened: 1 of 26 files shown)


toregge (approved):

storage/src/vespa/storage/distributor/pendingmessagetracker.h
```diff
@@ -15,22 +15,48 @@
...
 
-class PendingMessageTracker : public framework::HtmlStatusReporter
-{
+enum class TaskRunState {
+    OK,
+    Aborted,
+    BucketLost
+};
+
+struct DeferredTask {
```
geirst, line 31: Consider adding class comment.

storage/src/vespa/storage/distributor/pendingmessagetracker.h
```diff
@@ -15,22 +15,48 @@
...
 #include <set>
 #include <unordered_map>
 #include <chrono>
 #include <mutex>
 
 namespace storage::distributor {
 
-class PendingMessageTracker : public framework::HtmlStatusReporter
-{
+enum class TaskRunState {
```
geirst, line 25: Consider adding class comment.

storage/src/vespa/storage/visiting/reindexing_visitor.h
```diff
@@ -0,0 +1,27 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include "visitor.h"
+
+namespace storage {
+
+class ReindexingVisitor : public Visitor {
```
geirst, line 8: Consider adding class comment.

geirst (approved):
👍

vekterli:
@geirst PTAL at the two latest commits

geirst (approved):
👍

PR #15575, opened 2020-12-01
Bratseth/apply on restart take 9

bratseth:
@gjoranv fyi. 

I remembered wrongly - it's building a ConfigInstance from a ConfigInstance.Builder I moved out of the config model and into ConfigInstance.Builder. I tried to put it in configserver instead now to see if this makes it work. Otherwise I think I'll embark on a major cleanup, but would rather not couple that with this as it is enough here as it is.


hmusum:
👍

PR #15579, opened 2020-12-01
Jonmv/floating point window size

jonmv:
@baldersheim please review, but do not merge.


jonmv:
Uh, first two commits should have been one.

baldersheim (approved):
Looks fine, but I think that @vekterli should weigh in too

vekterli (commented):
The inner workings of this algorithm is about as transparent to me as a trying to stare through a plaster wall 😬 Adding some comments to the code base for the hows and whys of these tweaks would be very nice

jonmv:
Yes, working on that, and a thorough unit test/playground.

jonmv:
I have no further, significant changes, though, so may as well merge this now to see how it behaves in real scenarios.

PR #15588, opened 2020-12-02
Add host switch updater

mpolden:
@freva

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
@@ -61,6 +61,7 @@ public class Node {
     private final List<NodeHistory> history;
     private final Set<String> additionalIpAddresses;
     private final String openStackId;
+    private final Optional<String> switchHostname;
 
     public Node(HostName hostname, Optional<HostName> parentHostname, State state, NodeType type, NodeResources resources, Optional<ApplicationId> owner,
                 Version currentVersion, Version wantedVersion, Version currentOsVersion, Version wantedOsVersion,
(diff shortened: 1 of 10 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
```diff
@@ -495,12 +503,17 @@ public Builder openStackId(String openStackId) {
             return this;
         }
 
+        public Builder switchHostname(Optional<String> switchHostname) {
```
freva, line 506: I don't think builders should take Optionals, instead the caller should avoid calling if the field is not present.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
```diff
@@ -0,0 +1,52 @@
...
+        Map<String, NodeEntity> nodeEntities = controller().serviceRegistry().entityService().listNodes().stream()
+                                                           .collect(Collectors.toMap(NodeEntity::hostname,
+                                                                                     Function.identity()));
+        for (var zone : controller().zoneRegistry().zones().controllerUpgraded().all().ids()) {
+            for (var node : nodeRepository.list(zone)) {
+                NodeEntity nodeEntity = nodeEntities.get(node.hostname().value());
+                if (!shouldUpdate(node, nodeEntity)) continue;
+
+                NodeRepositoryNode updatedNode = new NodeRepositoryNode();
+                updatedNode.setSwitchHostname(nodeEntity.switchHostname().orElse(null));
```
freva, line 40: If we are guaranteed that this is set, this should be `.orElseThrow()`, otherwise we should filter this in `shouldUpdate()`

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
mpolden: NodeEntity isn't guaranteed to have a switch, and if it's removed we want to remove it from the node as well.

freva: I see, makes sense, but `NodeRepositoryNode` excludes `null`s, so I think this will just be a patch with `{}`?

PR #15589, opened 2020-12-02
Subclass QuorumPeerMain to be able to use initializeAndRun()

hmusum:
Cannot use `QuoeumPeerMain.main()`, since it calls `System.exit()` and will not work
when running unit tests, so subclass and use `initializeAndRun()`

### DIFF
--- a/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/ZooKeeperRunner.java
+++ b/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/ZooKeeperRunner.java
@@ -11,6 +11,7 @@ import org.apache.zookeeper.server.quorum.QuorumPeerMain;
 import java.io.IOException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


hmusum:
@aressem FYI: This should fix occasionally failing unit tests in `zookeeper-server`

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/ZooKeeperRunner.java
```diff
@@ -18,31 +25,42 @@
...
     }
 
     void shutdown() {
-        zkServerThread.interrupt();
-        try {
-            zkServerThread.join();
-        } catch (InterruptedException e) {
-            log.log(Level.WARNING, "Error joining server thread on shutdown", e);
-        }
+        executorService.shutdownNow();
```
mpolden, line 39: Wait for shutdown?

hmusum: Yep.

mpolden (approved):

PR #15595, opened 2020-12-02
Allow a grace period after node re-activation

freva:
Common scenario: Host (and nodes on it) are failed out because they are down. Operator fixes the host and sets everything to `active`. The `down` event is still in the history from last time and it may take a few minutes before the node is up again (f.ex. due to downloading the image, waiting for services). If `NodeFailer` runs again in the meantime, the node will be failed out again.

This adds a grace period if the node was re-activated, without clearing the down time (in case it doesn't work out, we will still have the time of first down).


bratseth (approved):
👍

PR #15596, opened 2020-12-02
Simplify symlink

hakonhall:
The symlink points to a file in the same directory.  Therefore, instead of
pointing to the absolute path, it should point to the filename.  This also
makes the symlink work on the host, if the symlink was made in a container (and
vice versa).

Plus, argument to Math.round() is a long, so not necessary.


hakonhall:
ping @baldersheim 

Travis failed on HEAD being broken (at the time)

hmusum (approved):

hakonhall:
Thank you

PR #15598, opened 2020-12-02
Bjorncs/feature flags

bjorncs:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/FlagDefinition.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/FlagDefinition.java
@@ -2,6 +2,7 @@
 package com.yahoo.vespa.flags;
 
 import javax.annotation.concurrent.Immutable;
+import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


flags/src/main/java/com/yahoo/vespa/flags/FlagDefinition.java
```diff
@@ -39,4 +53,26 @@ public String getDescription() {
...
+    public Instant getExpiresAt() { return expiresAt; }
+
+    private static void validate(List<String> owners, Instant createdAt, Instant expiresAt) {
+        if (expiresAt.isBefore(createdAt)) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Flag cannot expire before its creation date (createdAt='%s', expiresAt='%s')",
+                            createdAt, expiresAt));
+        }
+        if (owners.equals(PermanentFlags.OWNERS)) {
```
hakonhall, line 70: == should be used here?

bjorncs: I don't like that for anything except enums. Defensive copying of collections in constructors is a common pattern in our code base. That would break a reference equality check.

hakonhall: As it is now, the below if-block is unreachable because of this, since if `owners.equals(PermanentFlags.OWNERS)` then `owners.isEmpty()` is true.

This code should instead ask PermanentFlags if the Flag is owned by PermanentFlags, e.g. testing for reference-equals with a private static OWNERS, and then that class should validate the Flag instead of here.
If this code is supposed to

bjorncs: I agree, the code was confusing. I changed line 70 to use your suggestion to check if flag originates from `PermanentFlags`.

hakonhall (approved):

PR #15599, opened 2020-12-02
Report 0 metrics once

hakonhall:
A metric like configserver.lockAttempt.acquireTimedOut.average may be reported
as non-zero once, and then not reported for a while.  In the meantime, the
value reported for this metrics becomes 'last', carried over from earlier
periods.

This PR ensures that the value is set to 0 the first period after the value
becomes 0, but not thereafter (as carrying 0 is fine).

Ideally the metrics framework would expose a way to do this, but that would mean
changing the public metric APIs.  Also, I don't fully understand the current framework,
and changing it would be non-trivial.


mpolden (approved):
> Ideally the metrics framework would expose a way to do this

Yes, please!

PR #15605, opened 2020-12-02
Limit HostSwitchUpdater to relevant systems

mpolden:
@freva

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
@@ -5,6 +5,7 @@ import com.yahoo.config.provision.SystemName;
 import com.yahoo.vespa.hosted.controller.Controller;
 import com.yahoo.vespa.hosted.controller.api.integration.configserver.Node;
 import com.yahoo.vespa.hosted.controller.api.integration.configserver.NodeRepository;
+import com.yahoo.vespa.hosted.controller.api.integration.entity.EntityService;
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
```diff
@@ -31,14 +36,22 @@ protected boolean maintain() {
...
-            for (var node : nodeRepository.list(zone)) {
-                NodeEntity nodeEntity = nodeEntities.get(node.hostname().value());
-                if (!shouldUpdate(node, nodeEntity)) continue;
-
-                NodeRepositoryNode updatedNode = new NodeRepositoryNode();
-                updatedNode.setSwitchHostname(nodeEntity.switchHostname().get());
-                nodeRepository.patchNode(zone, node.hostname().value(), updatedNode);
+        int nodesUpdated = 0;
+        try {
+            for (var zone : controller().zoneRegistry().zones().controllerUpgraded().all().ids()) {
```
freva, line 41: Could also limit this to `yahoo` cloud zones...

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/HostSwitchUpdater.java
mpolden: Yes, but that's not available here so would have to use string. 😞

mpolden:
PTAL at last commit.

freva (approved):

PR #15611, opened 2020-12-02
Allow starting deferred tasks concurrently with reads

vekterli:
@geirst please review
* Test edge case where a bucket is removed before starting deferred visitor
* Allow starting deferred tasks concurrently with pending read ops


geirst (approved):
👍

PR #15621, opened 2020-12-02
Allocate once with the correct size

baldersheim:
@havardpe PR

### DIFF
--- a/vespalib/src/vespa/vespalib/stllike/hashtable.hpp
+++ b/vespalib/src/vespa/vespalib/stllike/hashtable.hpp
@@ -6,6 +6,28 @@
 
 namespace vespalib {
 
+namespace {
+
+/// TODO Currently we require that you have atleast one element in _nodes to avoid one extra branch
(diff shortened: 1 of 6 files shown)


baldersheim:
or @geirst

geirst (approved):
👍

vespalib/src/vespa/vespalib/stllike/hashtable.hpp
```diff
@@ -6,6 +6,28 @@
...
+template<typename Modulator>
+uint32_t
+computeModulo(size_t size) {
+    return (size > 0) ? Modulator::selectHashTableSize(roundUp2inN(size) / 3) : 1;
+}
+
+template <typename NodeStore>
+NodeStore
+createStore(size_t size, uint32_t modulo) {
+    size = (size > 0) ? roundUp2inN(std::max(size_t(modulo), roundUp2inN(size))) : 1;
```
havardpe, line 23: would it not be better to round up the memory used to (close to) a power of 2, instead of the number of entries?

PR #15626, opened 2020-12-03
Enable c++2a

baldersheim:
@arnej27959 or @toregge or @vekterli or @havardpe or @geirst PR


havardpe (commented):
which C++20 features supported by C++2a are finalized and which are experimental?
there are two concerns:
1 - features for which the implementation is experimental/flawed
2 - features that are implemented, but not yet finalized in the standard
If we can avoid these, we are good to go

baldersheim:
I think that must be checked in each use case. So far we are not using any.
The gcc 9 status can be checked at https://gcc.gnu.org/gcc-9/changes.html

havardpe:
> I think that must be checked in each use case. So far we are not using any.
> The gcc 9 status can be checked at https://gcc.gnu.org/gcc-9/changes.html

well, it does say that the 2a support is experimental in itself, so I guess all features we decide to use need to be tested pretty well with the current compiler to verify their exact behavior.

PR #15632, opened 2020-12-03
Add process_update member function to BucketDatabase.

toregge:
It is used for updating an entry with a single lookup.

@baldersheim : please review
@vekterli : FYI


baldersheim (approved):
This looks fine to me, but I think that @vekterli needs to verify it.

vekterli (approved):
:+1:

PR #15642, opened 2020-12-03
Record scaling event completion

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
```diff
@@ -87,13 +95,38 @@ private Applications applications() {
...
     }
 
+    /** Check if the last scaling event for this cluster has completed and if so record it in the returned instance */
+    private Cluster updateCompletion(Cluster cluster, NodeList clusterNodes) {
+        if (cluster.lastScalingEvent().isEmpty()) return cluster;
+        var event = cluster.lastScalingEvent().get();
+        if (event.completion().isPresent()) return cluster;
+
+        // Scaling event is complete if:
+        // - 1. no nodes which was retired by this are still present (which also implies data distribution is complete)
```
freva, line 105: How do we know when data distribution has to completed if we add new nodes? 
Since we know from the event what the target is, wouldn't it be better to identify the nodes that we are scaling to (i.e. the `count` nodes that are `active`, not retired and have the same resources as the `cluster.targetResources()`), then for each of those check with cluster controller if the node is in sync?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
bratseth: Yes, you are right; in that case we need to talk to the cluster controller. Not everything else is in place for content clusters yet either: They don't send the generation metric, and some times apply the new generation too quickly.

PR #15647, opened 2020-12-03
Don't increment or decrement volatile variables. It is deprecated when using C++20.

toregge:
@baldersheim : please review

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/querylimiter.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/querylimiter.cpp
@@ -25,14 +25,14 @@ QueryLimiter::grabToken(const Doom & doom)
             _cond.wait_for(guard, left);
         }
     }
-    _activeThreads++;
+    _activeThreads = _activeThreads + 1;
 }
 
 void
 QueryLimiter::releaseToken()
 {
     std::lock_guard<std::mutex> guard(_lock);
-    _activeThreads--;
+    _activeThreads = _activeThreads - 1;
     _cond.notify_one();
 }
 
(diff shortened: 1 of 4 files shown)


searchcore/src/vespa/searchcore/proton/matching/querylimiter.cpp
```diff
@@ -25,14 +25,14 @@ QueryLimiter::grabToken(const Doom & doom)
...
-    _activeThreads++;
+    _activeThreads = _activeThreads + 1;
 }
 
 void
 QueryLimiter::releaseToken()
 {
     std::lock_guard<std::mutex> guard(_lock);
-    _activeThreads--;
+    _activeThreads = _activeThreads - 1;
```
baldersheim, line 35: I think that it is better to remove volatile instead. It seems all use is while holding a lock.

vespalib/src/tests/array/array_test.cpp
```diff
@@ -27,16 +27,16 @@ std::ostream & operator << (std::ostream & os, const Array<T> & a)
 
 class Clever {
 public:
-    Clever() : _counter(&_global) { (*_counter)++; }
+    Clever() : _counter(&_global) { *_counter = *_counter + 1; }
```
baldersheim, line 30: I think it is better to replace with an atomic instead.

vespalib/src/tests/stllike/hash_test.cpp
```diff
@@ -171,16 +171,16 @@ TEST("test hash map iterator stability")
 
 class Clever {
 public:
-    Clever() : _counter(&_global) { (*_counter)++; }
+    Clever() : _counter(&_global) { *_counter = *_counter + 1; }
```
baldersheim, line 174: I think it is better to replace with an atomic instead.

baldersheim (changes requested):

baldersheim (approved):

PR #15660, opened 2020-12-04
Use std::to_chars for int to string conversion.

baldersheim:
@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/integerbase.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/integerbase.cpp
@@ -62,7 +62,11 @@ IntegerAttribute::getString(DocId doc, char * s, size_t sz) const {
     if (sz > 1) {
         largeint_t v = getInt(doc);
         auto res = std::to_chars(s, s + sz - 1, v, 10);
-        res.ptr[0] = 0;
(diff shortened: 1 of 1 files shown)


searchlib/src/vespa/searchlib/attribute/integerbase.cpp
```diff
@@ -49,18 +59,19 @@ uint32_t IntegerAttribute::get(DocId doc, WeightedConstChar * v, uint32_t sz) co
 }
 const char *
 IntegerAttribute::getString(DocId doc, char * s, size_t sz) const {
-    largeint_t v = getInt(doc);
-    snprintf(s, sz, "%" PRId64, v);
+    if (sz > 1) {
+        largeint_t v = getInt(doc);
+        auto res = std::to_chars(s, s + sz - 1, v, 10);
+        res.ptr[0] = 0;
```
havardpe, line 65: consider checking ec to avoid returning unspecified results.

baldersheim: Fixed

baldersheim:
@havardpe PTAL

havardpe (approved):

PR #15663, opened 2020-12-04
Separate loks so the same lock is not used for multiple things.

baldersheim:
@havardpe PR

### DIFF
--- a/fnet/src/vespa/fnet/transport_thread.cpp
+++ b/fnet/src/vespa/fnet/transport_thread.cpp
@@ -116,7 +116,7 @@ FNET_TransportThread::PostEvent(FNET_ControlPacket *cpacket,
 {
     size_t qLen;
     {
-        std::unique_lock<std::mutex> guard(_qLock);
+        std::unique_lock<std::mutex> guard(_lock);
         if (IsShutDown()) {
             guard.unlock();
             SafeDiscardEvent(cpacket, context);
@@ -217,9 +217,9 @@ FNET_TransportThread::FNET_TransportThread(FNET_Transport &owner_in)
       _selector(),
       _queue(),
       _myQueue(),
-      _qLock(),
       _lock(),
-      _cond(),
+      _shutdownLock(),
+      _shutdownCond(),
       _pseudo_thread(),
       _started(false),
       _shutdown(false),
@@ -233,9 +233,9 @@ FNET_TransportThread::FNET_TransportThread(FNET_Transport &owner_in)
 FNET_TransportThread::~FNET_TransportThread()
 {
     {
-        std::lock_guard<std::mutex> guard(_lock);
+        std::lock_guard<std::mutex> guard(_shutdownLock);
     }
-    if (_started.load(std::memory_order_relaxed) && !_finished) {
+    if (_started.load() && !_finished) {
         LOG(error, "Transport: delete called on active object!");
     } else {
         std::lock_guard guard(_pseudo_thread);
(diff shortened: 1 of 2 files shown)


fnet/src/vespa/fnet/transport_thread.cpp
```diff
@@ -234,7 +235,7 @@ FNET_TransportThread::~FNET_TransportThread()
     {
         std::lock_guard<std::mutex> guard(_lock);
     }
-    if (_started && !_finished) {
+    if (_started.load(std::memory_order_relaxed) && !_finished) {
```
havardpe, line 238: do not use relaxed memory order here.

fnet/src/vespa/fnet/transport_thread.h
```diff
@@ -43,10 +43,11 @@ class FNET_TransportThread : public FastOS_Runnable
     Selector                 _selector;       // I/O event generator
     FNET_PacketQueue_NoLock  _queue;          // outer event queue
     FNET_PacketQueue_NoLock  _myQueue;        // inner event queue
-    std::mutex               _lock;           // used for synchronization
-    std::condition_variable  _cond;           // used for synchronization
+    std::mutex               _qLock;          // protects the Q
+    std::mutex               _lock;           // used for synchronization during shutdown
```
havardpe, line 47: you should rename this lock instead of the other lock

fnet/src/vespa/fnet/transport_thread.h
```diff
@@ -43,10 +43,11 @@ class FNET_TransportThread : public FastOS_Runnable
...
+    std::condition_variable  _cond;           // used for synchronization during shutdown
```
havardpe, line 48: also rename this to make it match the renamed lock

fnet/src/vespa/fnet/transport_thread.cpp
baldersheim: Fixed

fnet/src/vespa/fnet/transport_thread.h
baldersheim: Fixed,

fnet/src/vespa/fnet/transport_thread.h
baldersheim: Fixed

baldersheim:
@havardpe PTAL

fnet/src/vespa/fnet/transport_thread.cpp
```diff
@@ -556,10 +549,10 @@ FNET_TransportThread::endEventLoop() {
            _myQueue.IsEmpty_NoLock());
 
     {
-        std::lock_guard<std::mutex> guard(_lock);
+        std::lock_guard<std::mutex> guard(_shutdownLock);
         _finished = true;
         if (_waitFinished) {
-            _cond.notify_all();
+            _shutdownCond.notify_all();
```
havardpe, line 555: consider some extra cleanup by removing _waitFinished altogether and always notifying all.

havardpe (approved):

fnet/src/vespa/fnet/transport_thread.cpp
baldersheim: Godd point, I was wondering about its actual value.

havardpe: It is the 'avoid signals unless someone is waiting' pattern used elsewhere, but no real need for it here.

PR #15617, opened 2020-12-02
Use the super bucket key when resolving which RPC target to use to ha…

geirst:
…ndle a command.

@vekterli please review
@baldersheim FYI

PR #15669, opened 2020-12-04
Use the super bucket key when resolving which RPC target to use to ha…

geirst:
…ndle a command.

@vekterli please review

This reapplies https://github.com/vespa-engine/vespa/pull/15617 with a fix that handles bucket id that is zero.


vekterli (approved):
:+1:

PR #15673, opened 2020-12-04
Retry reconfiguration if another one is in progress

hmusum:

### DIFF
--- a/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
+++ b/zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
@@ -27,6 +27,7 @@ public class Reconfigurer extends AbstractComponent {
 
     private static final Logger log = java.util.logging.Logger.getLogger(Reconfigurer.class.getName());
     private static final Duration sessionTimeout = Duration.ofSeconds(30);
+    private static final Duration retryReconfigurationPeriod = Duration.ofMinutes(5);
 
     private ZooKeeperRunner zooKeeperRunner;
     private ZookeeperServerConfig activeConfig;
@@ -87,10 +88,10 @@ public class Reconfigurer extends AbstractComponent {
         joiningServers = joiningServers.isEmpty() ? null : joiningServers;
         log.log(Level.INFO, "Will reconfigure ZooKeeper cluster. Joining servers: " + joiningServers +
                             ", leaving servers: " + leavingServers);
+
         String connectionSpec = connectionSpec(activeConfig);
         boolean reconfigured = false;
-        Instant end = Instant.now().plus(Duration.ofMinutes(5));
-
+        Instant end = Instant.now().plus(retryReconfigurationPeriod);
         // Loop reconfiguring since we might need to wait until another reconfiguration is finished before we can succeed
         while ( ! reconfigured && Instant.now().isBefore(end)) {
             try {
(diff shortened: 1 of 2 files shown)


zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
```diff
@@ -26,6 +27,7 @@
 
     private static final Logger log = java.util.logging.Logger.getLogger(Reconfigurer.class.getName());
     private static final Duration sessionTimeout = Duration.ofSeconds(30);
+    private static final Duration retryReconfigurationPeriod = Duration.ofMinutes(5);
```
mpolden, line 30: This seems too long. How about 30 seconds? Typically there will only be one config change and subsequent reconfigs on other nodes will effectively be no-ops.

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
```diff
@@ -86,7 +88,26 @@ private void reconfigure(ZookeeperServerConfig newConfig) {
...
+        while ( ! reconfigured && Instant.now().isBefore(end)) {
+            try {
+                zooKeeperReconfigure(connectionSpec, joiningServers, leavingServers);
+                reconfigured = true;
+            } catch (KeeperException e) {
+                if ( ! (e instanceof KeeperException.ReconfigInProgress))
+                    throw new RuntimeException(e);
+                log.log(Level.INFO, "Will retry in 1 second");
+                try {
+                    Thread.sleep(1000);
```
mpolden, line 105: Consider making the sleeper configurable by having an instance of `Consumer<Duration>` in this class, then the unit test doesn't actually have to sleep.

mpolden: Consider extracting a constant for the sleep period that is used both in the log message and sleep parameter.

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
```diff
@@ -86,7 +88,26 @@ private void reconfigure(ZookeeperServerConfig newConfig) {
...
-        zooKeeperReconfigure(connectionSpec(activeConfig), joiningServers, leavingServers);
+
+        String connectionSpec = connectionSpec(activeConfig);
+        boolean reconfigured = false;
+        Instant end = Instant.now().plus(retryReconfigurationPeriod);
+        // Loop reconfiguring since we might need to wait until another reconfiguration is finished before we can succeed
+        while ( ! reconfigured && Instant.now().isBefore(end)) {
+            try {
+                zooKeeperReconfigure(connectionSpec, joiningServers, leavingServers);
+                reconfigured = true;
```
mpolden, line 99: Consider logging how long reconfig took, so that we fine-adjust the timeouts.

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
hmusum: Yes, I have no idea how long this should. But we can try 30 seconds

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
hmusum: Yes, good idea

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
hmusum: Good suggestions, will implement

mpolden (approved):
👍

PR #15684, opened 2020-12-04
Log the total time taken to reconfigure

mpolden:
...is what I meant initially, but missed it in the previous review.

@hmusum


mpolden:
Last commit: A `null` watcher seems to cause a NPE when establishing connections between nodes, this then may cause `ConnectionLoss` for a node doing reconfig.

hmusum (approved):

PR #15686, opened 2020-12-04
Keep internal-redeploy serverside. Remove old restartOnDeploy impleme…

bratseth:
…ntation.


bratseth:
Factory is way too green so I'm merging this, please review it though.

hmusum:
👍

PR #15695, opened 2020-12-05
Only handle wakeup if necessary.

baldersheim:
@havardpe PR

### DIFF
--- a/fnet/src/vespa/fnet/transport_thread.h
+++ b/fnet/src/vespa/fnet/transport_thread.h
@@ -172,8 +172,6 @@ private:
         return _shutdown.load(std::memory_order_relaxed);
     }
 
-    void handle_wakeup_events();
-
 public:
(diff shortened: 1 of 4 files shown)


fnet/src/vespa/fnet/transport_thread.h
```diff
@@ -405,7 +403,7 @@ class FNET_TransportThread : public FastOS_Runnable
 
 
     // Empty selector call-back for selector wakeup
-    void handle_wakeup() { }
+    void handle_wakeup();
```
havardpe, line 406: should change comment as well

havardpe (approved):

fnet/src/vespa/fnet/transport_thread.h
baldersheim: Fixed

PR #15697, opened 2020-12-06
Only maintain capacity for nodes that are active

bjormel:


bratseth:
I think you need to do this at line 39 in CapacityChecker instead.
But perhaps what you want is to remove some of the states from relevantNodeStates?

bjormel:
Thanks @bratseth !

PR #15717, opened 2020-12-07
Revert "Simplify symlink"

hmusum:
Reverts vespa-engine/vespa#15596

Seems like this broke a system test. Haven't looked into it, so could be the test needs an update


hakonhall:
Could be - I verified the access log symlinks were as expected on a config server in CD.  I'll take a look at the systemtests.

PR #15722, opened 2020-12-07
Add a force_insert method to the hash_table. It is faster as it skips…

baldersheim:
… the presence check and avoid equality compare.

It allows duplicates and should hence be used carefully.
At least resize will be faster as it it can safely be used there.

@havardpe PR

### DIFF
--- a/vespalib/src/tests/stllike/hashtable_test.cpp
+++ b/vespalib/src/tests/stllike/hashtable_test.cpp
@@ -7,6 +7,7 @@
 #include <vespa/vespalib/testkit/testapp.h>
 #include <memory>
 #include <vector>
+#include <vespa/vespalib/stllike/hash_map.h>
 
 using vespalib::hashtable;
 using std::vector;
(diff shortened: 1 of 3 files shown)


vespalib/src/tests/stllike/hashtable_test.cpp
```diff
@@ -73,6 +73,48 @@ TEST("require that hashtable<int> can be copied") {
     EXPECT_EQUAL(42, *table2.find(42));
 }
 
+TEST("require that you can insert duplicates") {
+    using Pair = std::pair<int, vespalib::string>;
+    using Map = hashtable<int, Pair, vespalib::hash<int>, std::equal_to<int>, First<int, Pair>>;
```
havardpe, line 78: consider aliasing First to Select1st (use actual code instead of custom code that does the same)

vespalib/src/tests/stllike/hashtable_test.cpp
```diff
@@ -73,6 +73,48 @@ TEST("require that hashtable<int> can be copied") {
...
+    ASSERT_TRUE(found != m.end());
+    EXPECT_EQUAL(found->second, "1.3");
+
+    m.erase(1);
+    EXPECT_EQUAL(1u, m.size());
+    EXPECT_EQUAL(16u, m.capacity());
+    found = m.find(1);
+    ASSERT_TRUE(found != m.end());
+    EXPECT_EQUAL(found->second, "1.2");
+}
```
havardpe, line 116: should also verify that duplicates survive resizing

havardpe (approved):

vespalib/src/tests/stllike/hashtable_test.cpp
baldersheim: Fixed

vespalib/src/tests/stllike/hashtable_test.cpp
baldersheim: Done

PR #15725, opened 2020-12-07
Block ideal state operations towards buckets that are locked

vekterli:
@geirst please review
@jonmv FYI

Prevents ideal state ops from modifying buckets that are being used in a read-for-write context.
Move `OperationSequencer` to main `Distributor` to more easily facilitate sharing of it across components.


geirst (approved):
👍

PR #15727, opened 2020-12-07
Refactor code into module that does not depend on zookeeper

hmusum:


zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/LoggingWatcher.java
```diff
@@ -0,0 +1,19 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.zookeeper;
+
+import org.apache.zookeeper.WatchedEvent;
+import org.apache.zookeeper.Watcher;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+class LoggingWatcher implements Watcher {
```
mpolden, line 10: Consider inlining this as `(event) -> log.log(Level.INFO, event.toString())` so we don't need to have this class in every versioned module.

zookeeper-server/zookeeper-server-common/src/main/java/com/yahoo/vespa/zookeeper/ZkAdmin.java
```diff
@@ -0,0 +1,18 @@
...
+package com.yahoo.vespa.zookeeper;
+
+import java.time.Duration;
+
+/**
+ * Interface for administering a zookeeper cluster. Currently only supports reconfiguring a zookeeper cluster.
+ *
+ * @author hmusum
+ */
+public interface ZkAdmin {
```
mpolden, line 11: Consider naming it `VespaZooKeeperAdmin`, similar to `VespaZooKeeperServer`.

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/LoggingWatcher.java
hmusum: Yes, good point, fixed.

zookeeper-server/zookeeper-server-common/src/main/java/com/yahoo/vespa/zookeeper/ZkAdmin.java
hmusum: Done

mpolden (approved):
Great!

mpolden:
Test failure in unrelated (RpcServerTest.testRebinding failed).

PR #15731, opened 2020-12-08
Revert "Revert "Revert "Always deconstruct in reverse creation order, including Provider objects"""

jonmv:
Reverts vespa-engine/vespa#15724

@bjorncs I think this broke a system-test. It's probably fine to update the test, though, but no one to review right now ... ;p


jonmv:
https://github.com/vespa-engine/system-test/pull/1186

PR #15737, opened 2020-12-08
Remove code that nobody understands the need for

hmusum:
Cannot find any usage of nodes created. This might be some initialization
code to check if operations work, but I cannot see the need for it.
If this breaks something we can at least document why this is needed.

Please review only.


hakonhall (approved):
We used to verify ZooKeeper was up and running if the cluster controller was specified to run with a standalone ZooKeeper.  This was done by trying to erase, then write, a /CLUSTERNAME-ID path to ZK.  If this failed it was retried a number of times until a timeout was reached.

When that feature was removed, a tombstone removal of those paths were left.  They are long gone now...

Btw verifying ZK connection on bootstrap sounds like a great idea IMO.  It was removed when we moved from using ZooKeeper to Curator, for unknown reasons.  We should have something like that to avoid random error in "unrelated" later component.  E.g. when ZK fails to start today it typically shows as something wrong with flags, which is not the root cause...

hmusum:
Yes, agreed, and we have done checking in config server bootstrap for a long time. I thought this was done in `Curator` so that this was in place for all users of that class, but I see now that this is done when constructing `ConfigCurator`, which is only used by config server

PR #15739, opened 2020-12-08
Revert "Bratseth/simplify config take 2"

hmusum:
Reverts vespa-engine/vespa#15734

Build has not completed successfully in 45 minutes, I suspect this PR, but looking at the logs just makes me dizzy, so can't say for sure.

@aressem FYI


bratseth:
Let's see ...

PR #15740, opened 2020-12-08
Use exponential backoff for reconfig retries

mpolden:
Combined with nodes connecting to localhost this should reduce the time it takes
for all nodes to complete reconfig.

@hmusum


hmusum (approved):
Great!

PR #15741, opened 2020-12-08
Remove "use-fast-value-tensor-implementation" flag.

geirst:
Has been default on since 7.319.17 (25-Nov-2020).

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #15742, opened 2020-12-08
Log allocation details of node being moved

mpolden:
...and not for the host we're moving from (which is always allocated to
`hosted-vespa.tenant-host`)

@freva


freva (approved):
Great, you beat me to line 228 :smile:

PR #15751, opened 2020-12-09
Revert "Remove code that nobody understands the need for"

aressem:
Reverts vespa-engine/vespa#15737

Seems to have broken a bunch of system tests.


hmusum:
Thanks

PR #15757, opened 2020-12-09
Bounce reindexing visitor with BUSY if merge is pending for bucket

vekterli:
@geirst please review

Since reindexing visitors take a bucket lock when they arrive and
wait for pending ops to drain before they start, doing so when there's
a pending merge risks starving the bucket for a long time. This is
because merges may linger for a long time in the merge throttling
queues in the cluster. By not starting such visitors if there is a
pending merge, we avoid this edge case. Functionality is already in
place to inhibit merges from starting if there's an active bucket lock
present.


geirst (approved):
👍

PR #15766, opened 2020-12-09
Ignore case on autocomplete in querybuilder

klebadev:

As is, if I write a small word and use autocomplete, I get my word plus choice. Established the ability to use lowercase in querybuilde.


bratseth (approved):

bratseth:
👍

PR #15768, opened 2020-12-09
2 hacks to be able to use ZooKeeper server 3.6

hmusum:
Explicitly depend on io.dropwizard.metrics and org.xerial.snappy since
they are used by metrics provider (even NullMetricsProvider) when
bootstrapping ServerMetrics in zookeeper code and cannot be found
when running in a container

Do some class loading trick to circumvent zookeeper itself loading a class
with the threads current class loader, which does not work well in
a container.

There might be better ways of doing this, but this makes the zookeeper server start and work in a cluster with
3.5 and 3.6 servers. Also tested that generated config works with 3.5 server

### DIFF
--- a/zookeeper-server/zookeeper-server-3.6.2/pom.xml
+++ b/zookeeper-server/zookeeper-server-3.6.2/pom.xml
@@ -31,6 +31,9 @@
       <artifactId>slf4j-log4j12</artifactId>
       <version>1.7.5</version>
     </dependency>
+    <!-- snappy-java and metrics-core are included here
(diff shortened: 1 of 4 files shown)


mpolden (approved):
What if we implemented our own NullMetricsProvider, would we still need these dependencies?

hmusum:
Yes, tried implementing our own, same issue. The problem is in bootstrapping the metrics provider (in the static field `ServerMetrics.DEFAULT_METRICS_FOR_TESTS`), which loads `DefaultMetricsProvider` which drags in all these dependencies.

PR #15774, opened 2020-12-10
Delete dynamic config files if server is not a reconfigurable server

hmusum:
Delete ZooKeeper dynamic config files before starting server. They do not have to be used in cases where we are not going to dynamically reconfigure the ZooKeeper cluster and there are ZooKeeper bugs related to dynamic reconfig, at least on versions 3.5.x


mpolden:
Not needed.

PR #15775, opened 2020-12-10
Inline BucketInfo destructor.

geirst:
@baldersheim please review


baldersheim:
Looking at the code I am not sure if perf is giving the right picture. because deinlining the destructor seems to make sense in this case. I will have another run and look in details at the generated code.

toregge:
With this commit applied, i get a compile failure on Fedora 33 (using gcc 10) due to failure to inline BTreeBuilderMerger destructor.
De-inlining that destructor allows the compilation to succeed.

geirst:
Then I think we close this one for now.

PR #15778, opened 2020-12-10
Retry start of zookeeper server

hmusum:
An already running server may take som time to shut down, starting
a new one might fail in that case so we need to retry.


mpolden (approved):
👍

PR #15780, opened 2020-12-10
Bratseth/demand fewer metric readings

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -159,10 +160,15 @@ static Duration maxScalingWindow() {
...
-        return 4;
+    /** Returns the minimum measurements per node (average) we require to give autoscaling advice.*/
+    private int minimumMeasurementsPerNode(Duration scalingWindow) {
+        // Measurements are ideally taken every minute, but no guarantees
+        // (network, nodes may be down, collecting is single threaded and may take longer than 1 minute to complete.
+        // Since the metric window is 5 minutes, we won't really improve from measuring more often:
+        long minimumMeasurements = scalingWindow.toMinutes() / 5;
+        minimumMeasurements = Math.round(0.8 * minimumMeasurements); // Allow 20% metrics collection blackout
+        if (minimumMeasurements < 1) minimumMeasurements = 1;
+        return (int)minimumMeasurements;
```
mpolden, line 171: Consider simplifying to `Math.max(1, minimumMeasurements)`.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -159,10 +160,15 @@ static Duration maxScalingWindow() {
...
     }
 
-    /** Measurements are currently taken once a minute. See also scalingWindow */
-    static int minimumMeasurementsPerNode(ClusterSpec cluster) {
-        if (cluster.isStateful()) return 60;
-        return 4;
+    /** Returns the minimum measurements per node (average) we require to give autoscaling advice.*/
+    private int minimumMeasurementsPerNode(Duration scalingWindow) {
+        // Measurements are ideally taken every minute, but no guarantees
+        // (network, nodes may be down, collecting is single threaded and may take longer than 1 minute to complete.
```
mpolden, line 166: Unmatched `(`. 😄

mpolden (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
bratseth: Thanks. i'll fix it later

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
bratseth: I always have to think for 1 second when I see max implementing min, so some times I try to do without.

PR #15781, opened 2020-12-10
Change pointer type for LLVM

arnej27959:
* With LLVM "pointer to void" is not a valid type.
  Use char* aka "pointer to int8" instead.
  This worked unintentionally because LLVM was
  compiled with -DNDEBUG, disabling asserts.

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
+++ b/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
@@ -101,7 +101,7 @@ struct FunctionBuilder : public NodeVisitor, public NodeTraverser {
 
     llvm::PointerType *make_eval_forest_funptr_t() {
         std::vector<llvm::Type*> param_types;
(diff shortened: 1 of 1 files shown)


havardpe (commented):
should we also change the corresponding gcc pointer types?

arnej27959:
> should we also change the corresponding gcc pointer types?

I don't know; this change was enough to get the unit tests to pass (LLVM 7.0.1 compiled with assertions).
To be more consistent maybe we should say "char *" instead of "void *" in the gcc pointer types also, my guess is that the generated code will be the same.

@toregge do you have any input on this?

havardpe:
It is probably ok, since not all of the pointers are void* in C++ either. Seems strange for LLVM to not support void*, though.

toregge:
I would suggest not changing gcc pointer types, e.g. keep `void*` in gcc and map it to `char*` or` int8_t*` for llvm until the support for opaque pointers is in place for all versions of llvm used by vespa.

toregge (approved):

PR #15782, opened 2020-12-10
Use StringBuilder

bratseth:


baldersheim (approved):
StringBuffer ... I though that was way in the past :)

bratseth:
_Somebody_ brought it back :satisfied:

PR #15787, opened 2020-12-11
Use staggered delay for all config server maintainers

hmusum:


hakonhall (approved):

jonmv:
I assume there was a reason an initial delay was used here. I was thinking of preserving the delay, but adding that on top of the staggered start that will be used with this PR.

jonmv:
If this is not needed, then this is better :)

hmusum:
I don't think an initial delay is needed, we now inject `ConfigServerBootstrap` into `ConfigServerMaintenance` to make sure some things are setup properly before running a maintainer. At least I can't think of any reason right now.

PR #15795, opened 2020-12-11
Only let reindexing puts through locked bucket if their TaS token matches that of the lock

vekterli:
@geirst please review
@jonmv FYI

Avoids race condition edge case where reindexing puts from an outdated
visitor may pass through a bucket lock intended for a newly created
visitor operation

Tokens are 128-bit values derived from a CSPRNG, so uniqueness is for
all intents and purposes guaranteed.


jonmv:
🥳

geirst (approved):
👍

PR #15797, opened 2020-12-11
Avoid lookup bucket database entries two times when handling fast pat…

geirst:
…h update operation.

@vekterli please review
@baldersheim @toregge FYI


vekterli (approved):
:+1:

PR #15800, opened 2020-12-14
Refactor code and document code to verify that ZooKeeper is working

hmusum:
Based on non-working system tests when removing this code, document why it is needed/useful.

Also add annotation for unused constructor and a non-functional change in `getFleetControllers`

### DIFF
--- a/clustercontroller-apps/src/main/java/com/yahoo/vespa/clustercontroller/apps/clustercontroller/ClusterController.java
+++ b/clustercontroller-apps/src/main/java/com/yahoo/vespa/clustercontroller/apps/clustercontroller/ClusterController.java
@@ -96,16 +96,12 @@ public class ClusterController extends AbstractComponent
     }
 
     /**
-     * Creates a path in zookeeper for this fleetcontroller. Seems like this is meant as a check
-     * that zookeeper server is up and that we are able to do operations against it.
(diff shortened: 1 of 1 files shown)


clustercontroller-apps/src/main/java/com/yahoo/vespa/clustercontroller/apps/clustercontroller/ClusterController.java
```diff
@@ -104,4 +95,18 @@ void shutdownController(FleetController controller) throws Exception {
         controller.shutdown();
     }
 
+    /**
+     * Creates a path in zookeeper for this fleetcontroller. Seems like this is meant as a check
+     * that zookeeper server is up and that we are able to do operations against it.
```
hakonhall, line 100: Consider implementing this by calling CuratorFramework.blockUntilConnected().

hakonhall (approved):

clustercontroller-apps/src/main/java/com/yahoo/vespa/clustercontroller/apps/clustercontroller/ClusterController.java
hmusum: Much nicer, thanks, fixed.

hakonhall (approved):

PR #15806, opened 2020-12-14
Provide some human-readable responses to reindexing commands

jonmv:
@vekterli please reivew

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationHandler.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationHandler.java
@@ -88,7 +88,7 @@ public class ApplicationHandler extends HttpHandler {
 
         if (isReindexingRequest(request)) {
             applicationRepository.modifyReindexing(applicationId, reindexing -> reindexing.enabled(false));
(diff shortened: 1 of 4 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationHandler.java
```diff
@@ -86,7 +88,7 @@ public HttpResponse handleDELETE(HttpRequest request) {
 
         if (isReindexingRequest(request)) {
             applicationRepository.modifyReindexing(applicationId, reindexing -> reindexing.enabled(false));
-            return new JSONResponse(Response.Status.OK);
+            return new JSONResponse(Response.Status.OK) { { object.setString("message", "Reindexing disabled"); } };
```
vekterli, line 91: This feels sufficiently magical that I think it'd be good to encapsulate it in a `createJsonResponseWithMessage`-ish method and only do the wizardry once there

vekterli (approved):
:+1:

configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationHandler.java
jonmv: You're right.

PR #15810, opened 2020-12-14
Gracefully abort outdated reindexing visitors

vekterli:
@geirst please review
@jonmv FYI

Instead of sending `REJECTED` when lock token does not match, send
`TEST_AND_SET_CONDITION_FAILED` from distributors. The reindexing visitor
will detect this and remap the failure code to `ABORTED` while
simultaneously failing the backend visitor.
`ABORTED` will be considered transient by the visiting client, allowing
it to retry towards another distributor without the whole thing
having to fail out.


geirst (approved):
👍

jonmv (approved):
Nice. I'll verify client behaviour!

PR #15814, opened 2020-12-15
Move classloading hack to code that is only run for ZooKeeper 3.6 server

hmusum:


mpolden:
Does it make sense to move this? It's likely that future versions will need the same hack.

hmusum:
Yes, we will probably need it for future versions, good point.

PR #15815, opened 2020-12-15
Shut down ZooKeeper properly

mpolden:
...and kill it with fire if it doesn't.

@hmusum


hmusum (approved):

zookeeper-server/zookeeper-server-3.5.6/src/main/java/com/yahoo/vespa/zookeeper/VespaQuorumPeer.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final Logger LOG = Logger.getLogger(VespaQuorumPeer.class.getName());
+
+    public void start(Path path) {
+        initializeAndRun(new String[]{ path.toFile().getAbsolutePath()});
+    }
+
+    public void shutdown() {
+        if (quorumPeer != null) {
+            try {
+                quorumPeer.shutdown();
```
hakonhall, line 30: I think this should have worked as follows: 
1. quorumPeer.running should be set to false (under a synchronized(quorumPeer)), which means the quorumPeer thread will eventually exit its thread
2. wait until quorumPeer thread exits with quorumPeer.join(),
3. then do the rest of shutdown() tearing down resources.

As it is now (3) happens before (2), tearing down resources that might be used concurrently by the quorum peer thread.

And (2) may happen a long time after we return from this method, meaning it may still run concurrently with trying to start another instance of ZK.

mpolden: > quorumPeer.running should be set to false (under a synchronized(quorumPeer)), which means the quorumPeer thread will eventually exit its thread

`quorumPeer.running` is private though, as far as I can tell we can only initiate shutdown through `quorumPeer.shutdown()`

> wait until quorumPeer thread exits with quorumPeer.join(),

Yes, will fix.

> And (2) may happen a long time after we return from this method, meaning it may still run concurrently with trying to start another instance of ZK.

Yes, there any many reasons that the old server will linger. Component destruction only happens after 60s for example. This is why server start is retried.

PR #15826, opened 2020-12-15
Gjoranv/wait for deconstruct upon shutdown

gjoranv:
This is to prevent uninstalling bundles and ClassNotFound exceptions when the container is shutting down.

FYI: @bjorncs

### DIFF
--- a/container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
+++ b/container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
@@ -46,8 +46,7 @@ public class Deconstructor implements ComponentDeconstructor {
                    // are deconstructed, to prevent shutting down while deconstruct is in progress.
     }
 
-    // TODO: make private again
-    final ScheduledExecutorService executor =
+    private final ScheduledExecutorService executor =
             Executors.newScheduledThreadPool(2, ThreadFactoryFactory.getThreadFactory("component-deconstructor"));
 
     private final Mode mode;
--- a/container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
+++ b/container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
@@ -12,6 +12,7 @@ import java.util.List;
 
 
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import org.osgi.framework.Bundle;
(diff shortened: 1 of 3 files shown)


container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
```diff
@@ -61,9 +81,24 @@ public void deconstruct(List<Object> components, Collection<Bundle> bundles) {
...
         }
-        if (! destructibleComponents.isEmpty() || ! bundles.isEmpty())
-            executor.schedule(new DestructComponentTask(destructibleComponents, bundles),
+        if (!destructibleComponents.isEmpty() || !bundles.isEmpty()) {
+            var task = executor.schedule(new DestructComponentTask(destructibleComponents, bundles),
                               delay.getSeconds(), TimeUnit.SECONDS);
+            if (mode.equals(Mode.SHUTDOWN)) {
+                // Wait for deconstruction to finish
+                try {
+                    log.info("Waiting up to " + SHUTDOWN_DECONSTRUCT_TIMEOUT.toSeconds() + " seconds for all components to deconstruct.");
```
jonmv: Isn't there also a 50s timeout on deconstruct, somewhere further up?

gjoranv: Yes, you're right. That seems a bit harsh. Perhaps we should increase it to at least 90 sec, and set this one to ~80?

container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
```diff
@@ -35,13 +39,29 @@
...
 
+    private static final Duration SHUTDOWN_DECONSTRUCT_TIMEOUT = Duration.ofMinutes(10);
+
+    public enum Mode {
+        RECONFIG,  // Delay deconstruction to allow old components to finish processing in-flight requests.
+        SHUTDOWN   // The container is shutting down. Start deconstructing immediately, and wait until all components
+                   // are deconstructed, to prevent shutting down while deconstruct is in progress.
+    }
+
+    // TODO: make private again
```
jonmv: Yes, was made visible for testing. All tests can use the synchronous shutdown instead :)

jonmv (approved):
Do check the shutdown timeout.

container-disc/src/main/java/com/yahoo/container/jdisc/component/Deconstructor.java
jonmv: I think 45s is good!

jonmv:
I'm too slow, I see :) This is very good!

jonmv (approved):

PR #15829, opened 2020-12-15
Make the feed concurrency computations understandable.

baldersheim:
- Default is 0.5, unless you have at least one streaming cluster. Then it is 1.0.
- If you specify it directly, what you see is what you get. No magic division by 2 in some cases.

@geirst PR


geirst (approved):
👍

PR #15830, opened 2020-12-15
Turn on reindexing by default

jonmv:
@vekterli please review and merge. @bjorncs FYI.


jonmv:
Will not cause any actual reindexing unless this is explicitly triggered.

baldersheim:
How do you trigger it explicit ?

jonmv:
By an explicit API call to the config server, or (for hosted application) by submitting an application package which causes the indexing pipeline to change, with the appropriate validation override.

jonmv:
[Documentation](https://docs.vespa.ai/documentation/reindexing)

vekterli (approved):

PR #15835, opened 2020-12-16
Always force shutdown of non-reconfigurable ZooKeeper server

mpolden:
This basically reverts to the original behaviour, where the container is forced
to exit when a new VespaZooKeeperServer component is setup.

@hmusum


hmusum (approved):
👍

PR #15837, opened 2020-12-16
Use feature flag to decide how to run ZooKeeper server

hmusum:
Add flag to choose between reconfigurable ZooKeeper server or not for
the ZooKeeper server used by cluster controller.


mpolden (approved):
👍

PR #15838, opened 2020-12-16
Track explicitly when we are initializing config

bratseth:


config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
```diff
@@ -155,11 +155,18 @@ protected void subscribeAndHandleErrors(ConfigSubscription<?> sub, ConfigKey<?>
      *
      * If the call times out (timeout 1000 ms), no handle will have the changed flag set. You should not configure anything then.
      *
+     * @param isInitializing true if this the config is needed to create the initial configuration for the caller,
```
jonmv, line 158:
```suggestion
     * @param isInitializing true if this config is needed to create the initial configuration for the caller,
```

config/src/main/java/com/yahoo/config/subscription/ConfigSubscriber.java
```diff
@@ -227,11 +248,18 @@ public boolean nextGeneration() {
      * If the call times out (timeout 1000 ms), no handle will have the changed flag set. You should not configure anything then.
      *
      * @param timeoutMillis timeout in milliseconds
+     * @param isInitializing true if this the next generation is needed to create the initial configuration for the caller,
```
jonmv, line 251:
```suggestion
     * @param isInitializing true if this next generation is needed to create the initial configuration for the caller,
```

jonmv (approved):
Great!

hmusum (commented):
👍

PR #15840, opened 2020-12-17
Hide irrelevant test jobs from deployment API payload

jonmv:
@smorgrav please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
@@ -194,8 +194,14 @@ public class DeploymentStatus {
         return instances.build();
     }
 
-    /** The step status for all steps in the deployment spec of this, in the same order as in the deployment spec. */
-    public List<StepStatus> allSteps() { return allSteps; }
+    /** The step status for all relevant steps in the deployment spec of this, in the same order as in the deployment spec. */
+    public List<StepStatus> allSteps() {
+        List<JobId> firstTestJobs = List.of(firstDeclaredOrElseImplicitTest(systemTest),
+                                            firstDeclaredOrElseImplicitTest(stagingTest));
+        return allSteps.stream()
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
```diff
@@ -194,8 +194,14 @@ public boolean hasFailures() {
         return instances.build();
     }
 
-    /** The step status for all steps in the deployment spec of this, in the same order as in the deployment spec. */
-    public List<StepStatus> allSteps() { return allSteps; }
+    /** The step status for all relevant steps in the deployment spec of this, in the same order as in the deployment spec. */
+    public List<StepStatus> allSteps() {
+        List<JobId> firstTestJobs = List.of(firstDeclaredOrElseImplicitTest(systemTest),
```
jonmv, line 199: Only these test jobs will be triggered automatically. Undeclared jobs may still be triggered by users, though, but that likely won't happen if they aren't shown here :)

smorgrav: When will an implicit job be present here?

smorgrav (approved):
I trust you on this, albeit I'm not sure I understand the implications on when the implicit jobs are included or not

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
jonmv: When there aren't any explicit ones.

jonmv:
If there is a declared job of a specific type, the first of these will be chosen as the test job for instances without a declared job, which need that test run; in this case, no implicit job will run, i.e., not be shown here. 
If there are no declared jobs of a type, the same (first) instance will be used to run that test job whenever needed; in this case, that job will be among these `firstTestJobs`, and the step with that implicit job will not be filtered out.

jonmv:
This is way too complicated.

PR #15843, opened 2020-12-17
Add some smartness to what to build in PR's.

aressem:


travis/travis-build.sh
```diff
@@ -0,0 +1,49 @@
...
+    make -j ${NUM_THREADS}
+    ctest3 --output-on-failure -j ${NUM_THREADS}
+    ccache --show-stats
+    ;;
+  java)
+    env VESPA_MAVEN_EXTRA_OPTS="--no-snapshot-updates --batch-mode --threads ${NUM_THREADS}" sh ./bootstrap.sh java
+    mvn -V install --no-snapshot-updates --batch-mode --threads ${NUM_THREADS}
+    ;;
+  *)
+    env VESPA_MAVEN_EXTRA_OPTS="--no-snapshot-updates --batch-mode --threads ${NUM_THREADS}" sh ./bootstrap.sh java
```
hmusum, line 39: Shouldn't this be `bootstrap.sh full`?

hmusum (changes requested):
One comment that I think should be addressed, otherwise looks OK to me.

travis/travis-build.sh
aressem: When we build all, all the java artifacts will be built with Maven.  Using `full` here will work but the modules will be recompiled (waste) on the next line.

hmusum: Great!

hmusum (approved):

PR #15850, opened 2020-12-19
Make StreamedValueStore store shared_ptr to StreamedValue instances 

arnej27959:
This is scaffolding that must be rewritten, to avoid storing
the ValueType object in every tensor (again).
Note also that with this change StreamedValueStore and
DirectTensorStore have only a few small differences
in behavior, so we probably want to refactor them to
use more common code later.

The unit test change is needed because now this store
uses free-lists also. I'm not sure if we ever hit the
compacting path in the test.


arnej27959:
@havardpe PTAL

havardpe (approved):
thank you

PR #15851, opened 2020-12-20
Construct SessionCounter once

hmusum:
Creating SessionCounter every time we need the next session is wrong,
we might end up getting the same session id for two different
deployments, e.g. if they are happening in two different threads or on
two different config servers at the same time.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -88,6 +88,7 @@ public class SessionRepository {
     private final Path sessionsPath;
     private final TenantName tenantName;
     private final GlobalComponentRegistry componentRegistry;
+    private final SessionCounter sessionCounter;
 
     public SessionRepository(TenantName tenantName,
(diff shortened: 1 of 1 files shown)


jonmv (commented):
What does this change? The data is stored in ZK, and synchronisation taken care of by the Curator library. The (local) initialisation of the `SessionCounter` only does something if the node doesn't already exist — something which shouldn't be the case unless something very bad has happened. The returned handle(s) must work independently; otherwise it would never work across different JVMs either. 

In any case, this is cleaner code, but I don't think it really changes anything other than that, unless you convince me otherwise :)

hmusum:
If application `A` and `B` for the same tenant is deployed simultaneously on 2 different config servers in a cluster getNextSessionId() might be called simultaneously. Say `A` creates counter, initializes it to 1, then creates a new session, so new value should be 2. But `B` might initialize it to 1 (because it finds no existing sessions) although it has already been increased to 2. See `InitializedCounter()`. I guess this can still happen after this PR, but haven't looked into it yet.

We have had unstable system tests with issues that seems to be caused by session id not being incremented correctly. I haven't seen any of them after this went in (but I haven't looked at all tests, so can't say for sure).

PR #15856, opened 2020-12-21
Jonmv/more restricted uri pattern

jonmv:
@bratseth please review.


jonmv:
The requests whose URIs end up here are either HTTP requests from HttpRequestFactory or mbus requests with URLs that are either `mbus://localhost/<session name>` or `mbus://remotehost/<session name>`, so shouldn't break _us_ at least. I think it's very unlikely it breaks others.

jonmv:
@bjorncs FYI.

bratseth (approved):
Great!

PR #15857, opened 2020-12-21
Revert "Jonmv/more restricted uri pattern"

jonmv:
Reverts vespa-engine/vespa#15856

Broke a unit test in a different module.


jonmv:
Or not ...

PR #15859, opened 2020-12-21
Revert treating empty paths as "/"

jonmv:
@bratseth


jonmv:
Broke a unit test in a different module.

bratseth:
👍

PR #15861, opened 2020-12-21
Default to conservative behaviour when config servers down

jonmv:
@bratseth please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
@@ -249,16 +249,13 @@ public class ApplicationController {
 
     public ApplicationStore applicationStore() {  return applicationStore; }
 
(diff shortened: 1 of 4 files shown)


bratseth:
I think both of these changes are potential future bugs (I guess not current since you probably reviewed current usage?).
The methods don't quite do what they advertise any more: contentClustersByZone may just return some of the clusters, and isSuspended may return true when it is not suspended. 

So is there another way you could do this?

jonmv:
Well, the cluster code is supposed to return the set of content clusters you can currently reach, so that's fine. I can rename and re-doc to reflect this. 
I think the same argument can be made for the suspension; if we negate it and rename to "isDefinitelyNotSuspendedIn...", or something like that, it should be fine. Then it's OK to answer `true` only when config server says it's not, or when it's not deployed there.

jonmv:
PTAL :)

bratseth (approved):
Nice

PR #15862, opened 2020-12-21
Find oldest platform from job history when possible

jonmv:
@bratseth please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
@@ -62,7 +62,9 @@ import com.yahoo.vespa.hosted.controller.athenz.impl.AthenzFacade;
 import com.yahoo.vespa.hosted.controller.certificate.EndpointCertificateManager;
 import com.yahoo.vespa.hosted.controller.concurrent.Once;
 import com.yahoo.vespa.hosted.controller.deployment.DeploymentTrigger;
+import com.yahoo.vespa.hosted.controller.deployment.JobStatus;
 import com.yahoo.vespa.hosted.controller.deployment.Run;
+import com.yahoo.vespa.hosted.controller.deployment.RunStatus;
 import com.yahoo.vespa.hosted.controller.deployment.Versions;
 import com.yahoo.vespa.hosted.controller.persistence.CuratorDb;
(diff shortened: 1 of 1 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -256,18 +265,40 @@ public Instance requireInstance(ApplicationId id) {
...
 
+    /** Reads the oldest installed platform for the given application and zone from job history, or a node repo. */
+    private Optional<Version> oldestInstalledPlatform(JobStatus job) {
+        Version oldest = null;
+        for (Run run : job.runs().descendingMap().values()) {
+            Version version = run.versions().targetPlatform();
+            if (oldest == null || version.isBefore(oldest))
+                oldest = version;
+
+            if (run.status() == RunStatus.success)
```
bratseth, line 276: I don't understand this ...

This returns the on encountering the newest (? is that what descendingMap does) job that was a success? But wouldn't you then return that jobs version? And why is it oldest?

jonmv: The assumption is that the last/newest success converged, making all nodes have its target version. The versions of that run, and all runs after it (since they're all non-successes) may have caused at least one node to have their target version, so the versions to consider are all of these. The oldest one is what we're after. Oldest platform version = lowest version number.

jonmv: Yes, the runs are ordered by increasing run number, so the descending map iterator starts with the latest run.

bratseth: Okay, and the 'oldest' thing is in case we're downgrading then. It might be good with some more explanation here but at least I agree this looks not wrong.

bratseth (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
jonmv: Yes, in case we downgraded at one point. It's also not more confusing to explicitly pick the lowest version, than to assume we always upgrade, and pick the earliest installed, imho. I'll add a note.

PR #15867, opened 2020-12-22
Merge StoragePolicy down into ContentPolicy, and always use this

jonmv:
@vekterli please review. 

This will probably enable some simplifications and cleanup elsewhere.


jonmv:
Wow, that's an inflated diff. It's really just replacing the overridden methods from `ContentPolicy` in `StoragePolicy`, and renaming the latter.

vekterli (commented):
:+1:

PR #15871, opened 2020-12-22
Merge StoragePolicy down into ContentPolicy, and always use this

aressem:
This is the same as https://github.com/vespa-engine/vespa/pull/15867 which I accidentally merged and was reverted. Sorry about this.

PR #15873, opened 2020-12-23
Jonmv/remove storage policy

jonmv:
@vekterli


jonmv:
Hmm, could probably just replace the `[Content:cluster=<name>]` with just `<name>` as well ...

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -377,10 +377,6 @@ private DocumentOperationParameters parametersFromRequest(HttpRequest request, S
                                                           .map(parameters::withRoute)
                                                           .orElse(parameters);
                 break;
-            case CONDITION:
-                parameters = getProperty(request, ROUTE).map(parameters::withRoute)
-                                                        .orElse(parameters);
-                break;
```
jonmv, line 383: Oops, not intended to be here, but this is just removal of dead code that shouldn't be there in the first place.

documentapi/src/tests/policies/policies_test.cpp
```diff
@@ -83,10 +83,10 @@ class Test : public vespalib::TestApp {
...
     void requireThatExternPolicySelectsFromExternSlobrok();
     void requireThatExternPolicyMergesOneReplyAsProtocol();
-    void requireThatStoragePolicyWithIllegalParamIsAnErrorPolicy();
-    void requireThatStoragePolicyIsRandomWithoutState();
-    void requireThatStoragePolicyIsTargetedWithState();
-    void requireThatStoragePolicyCombinesSystemAndSlobrokState();
+    void requireThatContentPolicyWithIllegalParamIsAnErrorPolicy();
+    void requireThatContentPolicyIsRandomWithoutState();
+    void requireThatContentPolicyIsTargetedWithState();
+    void requireThatContentPolicyCombinesSystemAndSlobrokState();
```
vekterli, line 89: Should we retain at least one test per language that tests that `Storage` still works?

vekterli (approved):
:+1: I presume this entirely subsumes #15871 / #15867 ?

I've been pondering if there are any scenarios where this can cause issues during rolling upgrades (where what a "storage policy" entails may differ between nodes), but from what I can tell all these policies should in practice be evaluated in the same process as the sender.

documentapi/src/vespa/documentapi/messagebus/documentprotocol.cpp
```diff
@@ -32,14 +32,14 @@ DocumentProtocol::DocumentProtocol(std::shared_ptr<const DocumentTypeRepo> repo,
     // When adding factories to this list, please KEEP THEM ORDERED alphabetically like they are now.
     putRoutingPolicyFactory("AND", std::make_shared<RoutingPolicyFactories::AndPolicyFactory>());
     putRoutingPolicyFactory("Content", std::make_shared<RoutingPolicyFactories::ContentPolicyFactory>());
-    putRoutingPolicyFactory("MessageType", std::make_shared<RoutingPolicyFactories::MessageTypePolicyFactory>());
+    putRoutingPolicyFactory("Storage", std::make_shared<RoutingPolicyFactories::ContentPolicyFactory>()); // TODO: remove
```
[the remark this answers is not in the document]
jonmv:
```suggestion
    putRoutingPolicyFactory("Storage", std::make_shared<RoutingPolicyFactories::ContentPolicyFactory>()); // TODO Vespa 8: remove
```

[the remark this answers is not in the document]
jonmv: Yes. Major version is a good idea, since this kills the "Storage" policy.

documentapi/src/tests/policies/policies_test.cpp
jonmv: Yes.

jonmv: I think there weren't any tests that "Content" worked, before ... 🙀

jonmv: So in Java-land there is no test of this setup — only a mock setup where some random string (`"storage"`) is used as the policy name. 
In C++ land there is a test which involves getting a policy from a semi-real mbus. The simplest thing to do would probably just be to verify that whatever policy was returned for `"Storage"` was the same as what was returned for `"Content"` — do you feel the call?

jonmv: It's really a trivial thing, though. Not absolutely convinced we need it.

PR #15878, opened 2021-01-04
- Do not stop persitence thread until it has been interrupted.

baldersheim:
- fetch messages even if closed to allow draining the Q.
- GC unused code.

@vekterli PR


vekterli (approved):
:+1:

PR #15898, opened 2021-01-04
Always set 'is-master' metric. Remove 'master-change'

olaaun:
Only setting `is-master` on leadership change might hide bad states.

`master-change` is unexposed/unused

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/MetricUpdater.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/MetricUpdater.java
@@ -75,14 +75,8 @@ public class MetricUpdater {
         metricReporter.set("agreed-master-votes", maxCount);
     }
 
-    public void becameMaster() {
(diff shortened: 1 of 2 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/MetricUpdater.java
```diff
@@ -75,14 +75,8 @@ public void updateMasterElectionMetrics(Map<Integer, Integer> data) {
...
 
-    public void becameMaster() {
-        metricReporter.set("is-master", 1);
-        metricReporter.add("master-change", 1);
-    }
-
-    public void noLongerMaster() {
-        metricReporter.set("is-master", 0);
-        metricReporter.add("master-change", 1);
+    public void isMaster(boolean isMaster) {
```
vekterli, line 78: Consider renaming to something like `updateMasterState` to make the method sound less like a getter

vekterli (approved):
:+1:

PR #15903, opened 2021-01-05
Copy to temp dir and move

hmusum:


jonmv (approved):
To achieve atomicity?

PR #15904, opened 2021-01-05
Revert "Reapply "Upgrade to Curator 4""

hmusum:
Reverts vespa-engine/vespa#15400

In case this breaks something, please review only.

@aressem FYI


aressem (approved):

hmusum:
One system failing, need to look into it

PR #15906, opened 2021-01-05
Do not fail if app has already been copied [run-systemtest]

hmusum:


jonmv (commented):
I'm afraid this could lead to a situation where the target directory is created, but copying something else inside it fails, causing an incomplete application package to reside there; but we return happily.

hmusum:
Yeah, you are right, needs to be solved another way.

PR #15907, opened 2021-01-05
Revert "Jonmv/remove storage policy"

jonmv:
Reverts vespa-engine/vespa#15873

Broke tests. I found a problem with streaming clusters, where the <cluster-name>-index route does not exist, which would probably explain a lot of the failures.


vekterli (approved):
:+1:

PR #15909, opened 2021-01-05
Only fetch document IDs in bucket when splitting, not whole documents

vekterli:
@baldersheim please review

Splitting code only needs to look at the document IDs to figure out
the distribution of the target buckets. Remove tracking of document
sizes (which _did_ need the whole documents) since it was only used
in a warning log message.

### DIFF
--- a/storage/src/vespa/storage/persistence/splitbitdetector.cpp
+++ b/storage/src/vespa/storage/persistence/splitbitdetector.cpp
@@ -4,6 +4,7 @@
 #include "bucketprocessor.h"
 #include <vespa/document/bucket/bucketidfactory.h>
 #include <vespa/document/base/documentid.h>
+#include <vespa/document/fieldset/fieldsets.h>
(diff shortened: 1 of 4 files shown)


baldersheim (approved):

storage/src/vespa/storage/persistence/splitbitdetector.cpp
```diff
@@ -211,8 +209,7 @@ SplitBitDetector::detectSplit(spi::PersistenceProvider& provider,
             error << "Could not find differing bit to split bucket contents "
                      "around due to bucket ID collisions. Forcing resulting "
                      "bucket to be 58 bits. Bucket has "
-                  << detector._docCount << " docs totalling "
-                  << detector._docSize << " bytes. ";
```
baldersheim, line 215: All the efforts we go to to get the logging correctly detailed.

PR #15911, opened 2021-01-05
Make clients use orchestratorStatus instead of allowedToBeDown

hakonhall:


hmusum:
Are you sure this works in the transition? How about keeping `allowedToBeDown` in the API responses until transition is finished and use that if `orchestratorStatus` is not available? Or is this not an issue?

hakonhall:
I do not change any responses with this PR - it already returns both.  So this is just updating the client from one to the other.  Once this has rolled out I can remove the old allowedToBeDown in the response.

hmusum (approved):
👍

PR #15918, opened 2021-01-05
Jonmv/reapply remove storage plicy

jonmv:
@vekterli please review. Same as #15873 but without implicitly generated routes for visits.


vekterli (approved):
:+1:

PR #15920, opened 2021-01-05
Fix deletion of tenant

hmusum:
Discovered when trying to use Curator version 4, this should work for all versions of Curator.


jonmv (approved):
What was the issue on 4?

PR #15921, opened 2021-01-05
Allow updates with only remove/clear updates and operations to single…

baldersheim:
… value numeric fields to pass

even when feed rejection is in effect.
The rationale is that these updates will not affect memory or disk footprint negatively,
but it can have a significant positive effect.
Note that this will only be allowed for buckets that are in sync.

@geirst and @toregge and @vekterli PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/feed_reject_helper.h
+++ b/searchcore/src/vespa/searchcore/proton/server/feed_reject_helper.h
@@ -13,6 +13,10 @@ namespace proton {
 class FeedOperation;
 class UpdateOperation;
 
+/**
+ * Tells wether an operation should be blocked when resourcelimits have been reached.
+ * It looks at the operation type and also the content if it is an 'update' operation.
+ */
 class FeedRejectHelper {
 public:
     static bool isRejectableFeedOperation(const FeedOperation & op);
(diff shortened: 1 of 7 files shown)


searchcore/src/vespa/searchcore/proton/server/feed_reject_helper.h
```diff
@@ -0,0 +1,26 @@
...
+    class DocumentUpdate;
+    class ValueUpdate;
+}
+
+namespace proton {
+
+class FeedOperation;
+class UpdateOperation;
+
+class FeedRejectHelper {
```
geirst, line 16: Please add class description.

geirst (approved):
👍  @toregge please have a look as well.

toregge (approved):
Note that the operations now allowed will use space in the transaction log. If flushing does not keep up with feeding (i.e. unable to prune transaction log, compact doc store or perform disk index fusion) then we might still run out of disk space.

vekterli (approved):
:+1:

searchcore/src/vespa/searchcore/proton/server/feed_reject_helper.h
baldersheim: Done

baldersheim:
Since these operations allowed are small that should not be an issue.

PR #15922, opened 2021-01-06
Add support for pruning managed containers

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/ContainerOperationsImpl.java
```diff
@@ -324,6 +325,16 @@ public boolean noManagedContainersRunning() {
         return containerEngine.noManagedContainersRunning(MANAGER_NAME);
     }
 
+    @Override
+    public boolean retainManagedContainers(Set<ContainerName> containerNames) {
+        return containerEngine.listManagedContainers(MANAGER_NAME).stream()
+                .filter(containerName -> ! containerNames.contains(containerName))
+                .peek(containerName -> {
+                    containerEngine.stopContainer(containerName);
```
hakonhall, line 333: will this be a no-op if the container is already stopped?

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/ContainerOperationsImpl.java
freva: Yes;
https://github.com/vespa-engine/vespa/blob/130ac9c430496b5e31098818bff19b869d6c9f44/docker-api/src/main/java/com/yahoo/vespa/hosted/dockerapi/DockerEngine.java#L231

PR #15924, opened 2021-01-06
Make clients use orchestratorStatus instead of allowedToBeDown, try 2 [run-systemtest]

hakonhall:
Reverts vespa-engine/vespa#15923


hakonhall:
The problem in the previous try was that NodeRepositoryNode is used to serialize to a JSON patch request:  A null orchestratorStatus field SHOULD make a patch request with the orchestratorStatus field absent.  Fixed in 25bdcc8.

hakonhall:
The screwdriver failure is unrelated.

hmusum (approved):

PR #15933, opened 2021-01-06
Allow expressions as arguments

lesters:
@bratseth Please review. Practically same as https://github.com/vespa-engine/vespa/pull/12930, with some better testing. I do think this is correct as well, but an additional pair of eyes would be helpful in identifying any unintended consequences.


bratseth (approved):
Very nice!

PR #15947, opened 2021-01-07
Count lock timeout as unsuccessful run

mpolden:
Let it build pls. Didn't bother building all the modules.

@jonmv

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
@@ -34,7 +34,7 @@ public abstract class ConfigServerMaintainer extends Maintainer {
     /** Creates a maintainer where maintainers on different nodes in this cluster run with even delay. */
     ConfigServerMaintainer(ApplicationRepository applicationRepository, Curator curator, FlagSource flagSource,
                            Instant now, Duration interval) {
-        super(null, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
+        super(null, Mode.exclusive, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
               jobMetrics(applicationRepository.metric()), cluster(curator));
         this.applicationRepository = applicationRepository;
     }
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
@@ -34,7 +34,7 @@ public abstract class ConfigServerMaintainer extends Maintainer {
     /** Creates a maintainer where maintainers on different nodes in this cluster run with even delay. */
     ConfigServerMaintainer(ApplicationRepository applicationRepository, Curator curator, FlagSource flagSource,
                            Instant now, Duration interval) {
-        super(null, Mode.exclusive, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
+        super(null, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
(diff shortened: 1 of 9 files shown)


vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
```diff
@@ -92,17 +84,23 @@ public void awaitShutdown() {
...
-                jobMetrics.recordRunOf(name());
-                if (maintain()) jobMetrics.recordSuccessOf(name());
-            } finally {
-                // Always forward metrics
-                jobMetrics.forward(name());
+            if (maintain()) jobMetrics.recordSuccessOf(name());
+        } catch (UncheckedTimeoutException ignored) {
+            if (mode == Mode.shared) {
+                // This is fine as we're colliding with a run on another node
+                jobMetrics.recordSuccessOf(name());
```
jonmv, line 96: It's not a problem, but it's not really a success either.

jonmv: I think it would be better to change this so the `recordRunOf(name())` is called when the lock is acquired _or_ mode is `shared`, and record success only when it is an actual success. Otherwise a — somewhat contrived, perhaps — case where maintainers on all servers fail every time, but run long enough to block each other out, will seem like they all succeed every second time or so.

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
```diff
@@ -34,7 +34,7 @@
     /** Creates a maintainer where maintainers on different nodes in this cluster run with even delay. */
     ConfigServerMaintainer(ApplicationRepository applicationRepository, Curator curator, FlagSource flagSource,
                            Instant now, Duration interval) {
-        super(null, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
+        super(null, Mode.exclusive, interval, now, new JobControl(new JobControlFlags(curator, flagSource)),
```
jonmv, line 37: Some of the configserver maintainers are shared as well. I don't remember which. This is probably fine, in any case.

vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
mpolden: When lock is acquired or mode is `exclusive` you mean?

Maybe we should remove special treatment of lock timeouts altogether instead. With staggered delay, it's unlikely that a run collides. If a maintainer frequently collides its interval is likely too low.

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
mpolden: Any input on this, @hmusum?

vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
jonmv: That's what I mean :)

I agree with the reasoning about intervals.

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ConfigServerMaintainer.java
hmusum: All config server maintainers need to run on all nodes so this looks right to me

jonmv: Perhaps it was just the ReindexingMaintainer that was shared. The others look exclusive, indeed. Just leave it like this.

mpolden:
PTAL

jonmv (approved):
:)

PR #15951, opened 2021-01-07
Disable topk optimisation on dispatch when content distribution is se…

baldersheim:
…verly skewed.

When the skew is too large the assumption that docs are evenly and randomly distributed hold.
The impact and is larger on smaller systems. In large systems the where this optimisation is more important,
the probabilitity of large skew will be less.

@jonmv PR

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
@@ -22,6 +22,8 @@ public class Group {
     private final AtomicBoolean hasFullCoverage = new AtomicBoolean(true);
     private final AtomicLong activeDocuments = new AtomicLong(0);
     private final AtomicBoolean isBlockingWrites = new AtomicBoolean(false);
+    private final AtomicBoolean isContentWellBalanced = new AtomicBoolean(true);
+    private final static double MAX_UNBALANCE = 0.10; // If documents on a node is more than 10% off from the average the group is unbalanced
 
     public Group(int id, List<Node> nodes) {
         this.id = id;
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
```diff
@@ -53,25 +55,30 @@ void setHasSufficientCoverage(boolean sufficientCoverage) {
...
+            long deviation = nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).mapToLong(node -> Math.abs(node.getActiveDocuments() - average)).sum();
+            isContentWellBalanced.set(deviation <= (activeDocs * MAX_UNBALANCE));
+        } else {
+            isContentWellBalanced.set(true);
+        }
+
     }
 
-    /** Returns the active documents on this node. If unknown, 0 is returned. */
+    /** Returns the active documents on this groip. If unknown, 0 is returned. */
```
jonmv, line 76:
```suggestion
    /** Returns the active documents on this group. If unknown, 0 is returned. */
```

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
```diff
@@ -53,25 +55,30 @@ void setHasSufficientCoverage(boolean sufficientCoverage) {
...
 
     public int workingNodes() {
-        int nodesUp = 0;
-        for (Node node : nodes) {
-            if (node.isWorking() == Boolean.TRUE) {
-                nodesUp++;
-            }
-        }
-        return nodesUp;
+        return (int) nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).count();
```
jonmv, line 58:
```suggestion
        return (int) nodes.stream().filter(node -> node.isWorking()).count();
```

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
```diff
@@ -53,25 +55,30 @@ void setHasSufficientCoverage(boolean sufficientCoverage) {
...
     }
 
     void aggregateNodeValues() {
-        activeDocuments.set(nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).mapToLong(Node::getActiveDocuments).sum());
-        isBlockingWrites.set(nodes.stream().anyMatch(node -> node.isBlockingWrites()));
+        long activeDocs = nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).mapToLong(Node::getActiveDocuments).sum();
```
jonmv, line 62:
```suggestion
        long activeDocs = nodes.stream().filter(node -> node.isWorking()).mapToLong(Node::getActiveDocuments).sum();
```

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
```diff
@@ -53,25 +55,30 @@ void setHasSufficientCoverage(boolean sufficientCoverage) {
...
+        activeDocuments.set(activeDocs);
+        isBlockingWrites.set(nodes.stream().anyMatch(Node::isBlockingWrites));
+        int numWorkingNodes = workingNodes();
+        if (numWorkingNodes > 0) {
+            long average = activeDocs / numWorkingNodes;
+            long deviation = nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).mapToLong(node -> Math.abs(node.getActiveDocuments() - average)).sum();
```
jonmv, line 68:
```suggestion
            long deviation = nodes.stream().filter(node -> node.isWorking()).mapToLong(node -> Math.abs(node.getActiveDocuments() - average)).sum();
```

jonmv (approved):
Just some details. Nice! What about nodes with duplicate corpuses?

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
baldersheim: What happens if 'null' is returned ? Wont that give NullPointerException ?

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
baldersheim: See comment above

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
baldersheim: See comment above

baldersheim:
Node.getActiveDocuments() only counts the primary copy which is the only one that is considered.

jonmv:
Ouch, `Boolean` return value! You're right, that would be an NPE. 

How do the primary copies change during redistribution from a retiring node?

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
jonmv: 💥

baldersheim:
Once a bucket has been copied to a new node, the copy which is now primary will be activated and the previous will be deactivated. Only active copies are searched. For a short period of time(milliseconds) there might be duplicates, but that is filtered higher up in the stack based on the 96bit GID.

jonmv:
So if all content nodes in a group are concurrently migrating from a set of old nodes to a set of new ones, and they're all halfway ±5%, then this will enable the optimisation anyway? Only in this scenario? If it _is_ enabled in this scenario, we'll get wrong results, assuming the search considers not just the primaries on a node (which I believe it does?).

baldersheim:
Yes once they are halfway it will be enabled again, and results will be correct.
Only 1 copy of a bucket is searched, except for the small subset of buckets being activated/deactivated.

jonmv:
Ok, perfect :)

PR #15953, opened 2021-01-07
Avoid changing the input string field value by cloning it before doin…

geirst:
…g changes.

This is needed as the input can be used by other expressions as well.
The same cloning is done in ExactExpression and TokenizeExpression.

@bratseth please review


bratseth (approved):
👍

PR #15954, opened 2021-01-07
Synchronize when creating application package

hmusum:
Synchronize to avoid threads trying to create an application package concurrently
(e.g. a maintainer and an external deployment).


jonmv (commented):
Why would it be a problem if they did?

jonmv (approved):

hmusum:
> Why would it be a problem if they did?

`copyApp()` could fail

PR #15960, opened 2021-01-08
Revert "Revert "Disable topk optimisation on dispatch when content distribution is se…""

baldersheim:
Reverts vespa-engine/vespa#15959


baldersheim:
Superceeded by #15973

PR #15973, opened 2021-01-08
Balder/do not allow topk optimisation when distribution is severly skewed take2

baldersheim:
@jonmv PR
Only the last 2 commits are new.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
@@ -370,7 +370,7 @@ public class SearchCluster implements NodeManager<Node> {
 
     public boolean isGroupWellBalanced(OptionalInt groupId) {
         if (groupId.isEmpty()) return false;
-        Group group = groups().get(groupId);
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -370,7 +370,7 @@ private boolean isGroupNodeCoverageSufficient(int workingNodes, int nodesInGroup
 
     public boolean isGroupWellBalanced(OptionalInt groupId) {
         if (groupId.isEmpty()) return false;
-        Group group = groups().get(groupId);
+        Group group = groups().get(groupId.getAsInt());
```
jonmv, line 373: XD

baldersheim: ?

jonmv: Same as 😆

jonmv (commented):
Perfect!

PR #15979, opened 2021-01-10
Reapply "Remove unnecessary component" [run-systemtest]

hmusum:
Now also uses just 1 host registry, with application id as key, whereas we earlier used 2 (one with application id as key and one with tenant name as key).


jonmv (approved):
Great!

PR #15982, opened 2021-01-11
Actually add new plugin goal :facepalm:

jonmv:
@mpolden


vespa-maven-plugin/src/main/java/ai/vespa/hosted/plugin/SuspendMojo.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Sets suspension status for a Vespa application deployment.
+ *
+ * @author jonmv
+ */
+@Mojo(name = "suspend")
+public class SuspendMojo extends AbstractVespaDeploymentMojo {
+
+    @Parameter(property = "suspend", required = true)
+    protected boolean suspend;
```
jonmv, line 16:
```suggestion
    private boolean suspend;
```

mpolden (approved):

PR #15987, opened 2021-01-11
Reapply "Remove unnecessary component" [run-systemtest]"

hmusum:
Please review only


hmusum:
Wait a bit, I want to do some more work here

jonmv (approved):
Hehe, seems like a necessary change :)

PR #15999, opened 2021-01-11
You must use one of the candidates in the set you got.

baldersheim:
It is not legal to change the set of candidates and perform a re-election if you are unhappy.

@jonmv PR


baldersheim:
Or @hmusum PR

hmusum (approved):

jonmv:
D'oh :)

PR #15985, opened 2021-01-11
Add possibility to select a subset of groups and apply random selecti…

baldersheim:
…on among them.

@jonmv PR

PR #16000, opened 2021-01-11
Revert "Add possibility to select a subset of groups and apply random selecti…"

hmusum:
Reverts vespa-engine/vespa#15985

System test tests/search/hierarchic_distribution/search_while_feeding.rb has failed every run after this was merged


baldersheim:
The root cause has been fixed in #15999

PR #16001, opened 2021-01-11
move metaData to recursive tenant api

ldalves:
I checked the console and this is the right place to have this extra information at the moment... 


jonmv (approved):
Yeah, that's right :) Sorry I missed this.

PR #16012, opened 2021-01-12
Reapply "Remove unnecessary component" [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#16003

Please review only


hakonhall:
#16003 was merged because it fixed CD, so why do you want to revert that?

jonmv (approved):

hmusum:
I thought the issue in CD was caused by https://github.com/vespa-engine/vespa/pull/15979, which was merged yesterday, but I think you are right in that this change does not work. Need to look more into it

hakonhall:
CD passed before #15987 was merged today, but broke with it.  I observed some of the same errors in the log from yesterday, perhaps due to #15924, and which may have failed some proxy nodes.

hmusum:
Found an issue where a component was using another host registry than elsewhere

jonmv:
Heh. Sneaky.

hakonhall (approved):

hakonhall (approved):
Interesting - I have been waiting for this kind of issue with the global component.

PR #16014, opened 2021-01-12
Reduce cost as test from 400 to below 100

baldersheim:
@vekterli PR


vekterli (approved):
:+1: what part of this test suite is taking so much time? Could consider pulling that out as its own test binary

baldersheim:
How is that done ?
As far as I can tell it is TEST(DistributionTest, test_distribution) in distribution/distributiontest.cpp

PR #16015, opened 2021-01-12
Replace the use of std::list with std::vector. This avoids a lof of n…

baldersheim:
…ew/delete.

The number of elements that are moved around is small and significantly less than
the gain by more efficient memory management.

@vekterli PR This brought down execution time from 400s to 120s.

### DIFF
--- a/vdslib/src/vespa/vdslib/distribution/distribution.cpp
+++ b/vdslib/src/vespa/vdslib/distribution/distribution.cpp
@@ -221,12 +221,12 @@ namespace {
 
     /** Used to record scored groups during ideal groups calculation. */
     struct ScoredGroup {
+        double       _score;
         const Group* _group;
-        double _score;
 
-        ScoredGroup() : _group(nullptr), _score(0) {}
-        ScoredGroup(const Group* group, double score) noexcept
-            : _group(group), _score(score) {}
+        ScoredGroup() noexcept : _score(0), _group(nullptr) { }
+        ScoredGroup(double score, const Group* group) noexcept
+            : _score(score), _group(group) { }
 
         bool operator<(const ScoredGroup& other) const noexcept {
             return (_score > other._score);
@@ -235,12 +235,13 @@ namespace {
 
     /** Used to record scored nodes during ideal nodes calculation. */
     struct ScoredNode {
+        double _score;
         uint16_t _index;
         uint16_t _reliability;
-        double _score;
 
-        ScoredNode(uint16_t index, uint16_t reliability, double score) noexcept
-            : _index(index), _reliability(reliability), _score(score) {}
+        ScoredNode() noexcept : _score(0), _index(0), _reliability(0) {}
+        ScoredNode(double score, uint16_t index, uint16_t reliability) noexcept
+            : _score(score), _index(index), _reliability(reliability) {}
 
         bool operator<(const ScoredNode& other) const noexcept {
             return (_score < other._score);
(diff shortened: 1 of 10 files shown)


vdslib/src/vespa/vdslib/distribution/distribution.cpp
```diff
@@ -443,7 +429,7 @@ Distribution::getIdealNodes(const NodeType& nodeType,
         // Create temporary place to hold results. Use double linked list
         // for cheap access to back(). Stuff in redundancy fake entries to
         // avoid needing to check size during iteration.
-        std::list<ScoredNode> tmpResults(groupRedundancy, ScoredNode(0, 0, 0));
+        std::vector<ScoredNode> tmpResults(groupRedundancy, ScoredNode(0, 0, 0));
```
vekterli, line 432: Consider updating comment that explicitly mentions linked list usage

vekterli (approved):
:+1:

vdslib/src/vespa/vdslib/distribution/distribution.cpp
baldersheim: Comment updated, and some other related improvements.

baldersheim:
@vekterli PTAL

vdslib/src/vespa/vdslib/distribution/distribution.cpp
```diff
@@ -459,12 +460,12 @@ Distribution::getIdealNodes(const NodeType& nodeType,
...
                 for (; it != tmpResults.end(); ++it) {
                     if (score > it->_score) {
-                        tmpResults.insert(it, ScoredNode(nodes[j], nodeState.getReliability(), score));
+                        tmpResults.insert(it, ScoredNode(score, nodes[j], nodeState.getReliability()));
                         break;
                     }
                 }
                 if (it == tmpResults.end()) {
-                    tmpResults.emplace_back(nodes[j], nodeState.getReliability(), score);
+                    tmpResults.emplace_back(score, nodes[j], nodeState.getReliability());
```
vekterli, line 468: Hmm is it technically possible for `it` to be invalidated by the `insert` above, i.e. comparing a bogus iterator with `end`? It _probably_ won't be since it's pre-reserved, but I'm not sure what guarantees are given here... Maybe check on something else than the iterator to be sure? If this is pulled out into its own function then could return instead of break and avoid the check entirely.

vekterli (commented):
A potential edge case to be pondered first, otherwise LGTM

vdslib/src/vespa/vdslib/distribution/distribution.cpp
baldersheim: Fixed

baldersheim:
@vekterli PTAL I found loads of other inefficient code too.

vdslib/src/vespa/vdslib/state/clusterstate.cpp
```diff
@@ -280,34 +280,40 @@ ClusterState::getNodeCount(const NodeType& type) const
...
-        throw vespalib::IllegalStateException(
-                "Unknown node type " + node.getType().toString(), VESPA_STRLOC);
-    }
-        // If it actually has an entry in map, return that
+    // If it actually has an entry in map, return that
     std::map<Node, NodeState>::const_iterator it = _nodeStates.find(node);
     if (it != _nodeStates.end()) return it->second;
+
+    // If beyond node count, the node is down.
+    if (__builtin_expect(node.getIndex() >= _nodeCount[node.getType()], false)) {
```
vekterli, line 302: Is it cheaper in practice to always do the map lookup than to have this check prior to it?

vekterli (approved):
:+1:

vdslib/src/vespa/vdslib/state/clusterstate.cpp
baldersheim: It layed out the code in a more efficient way. There seemed to be less jumps back and forth.
The most important change was moving the throwing out to separate non inlined method as it eliminated the need for pushing and popping on the stack.
The second one was reorganising to only call throwUnknownType in one place.
And then I assumed it is more likely the node is in the _nodeStates map, than being beyond the node count.

PR #16022, opened 2021-01-12
Lesters/expressions as arguments 3

lesters:
@bratseth Please review.

So, with more comprehensive testing. The rank deriving test in `config-model/src/test/derived/function_arguments/test.sd` is a guard as it tests what was previously parseable (the old `arg()` function in `RankingExpressionParser.jj`) does not change with this PR, e.g. constants (including negative ones), rank features and identifiers.

The test in `config-model/src/test/derived/function_arguments_with_expressions/test.sd` tests that autogenerated features are generated for the rest, e.g. arithmetic, if's, tensor functions and slices (with required `rankingExpression(...)` wrapper)

System test is incoming.


searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/ExpressionFunction.java
```diff
@@ -151,6 +153,15 @@ public Instance expand(SerializationContext context, List<ExpressionNode> argume
         return new Instance(symbol, expressionString);
     }
 
+    private boolean shouldGenerateFeature(ExpressionNode expr) {
+        if (expr instanceof ConstantNode) return false;
+        if (expr instanceof ReferenceNode) return false;
+        if (expr instanceof NameNode) return false;
+        if (expr instanceof FunctionNode) return false;
+        if (expr instanceof NegativeNode && ((NegativeNode) expr).getValue() instanceof ConstantNode) return false;
```
bratseth, line 161: Shouldn't this be 
`if (expr instanceof NegativeNode && ! shouldGenerateFeature((NegativeNode) expr).getValue())) return false;
`
?

bratseth (commented):
Looks good otherwise

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/ExpressionFunction.java
lesters: Not really. This captures the fact that the expression parsing represents a negative constant as a `NegativeNode` containing a `ConstantNode`. So that can be output safely as is. Previous parsing (old `arg`) had a regular `ConstantNode` with negative value.

A `NegativeNode` in general should probably be in an expression. For instance `-attribute(field)` (`NegativeNode` with a `ReferenceNode`).

bratseth (approved):

PR #16024, opened 2021-01-12
- Wait for zero inactive threads before trying to start the one that …

baldersheim:
…shall fail.

- Also ensure that the started threads do not stop until we have completed the test.

@hmusum or @toregge PR

### DIFF
--- a/fastos/src/tests/threadtest.cpp
+++ b/fastos/src/tests/threadtest.cpp
@@ -23,34 +23,29 @@ class ThreadTest : public ThreadTestBase
 
       if (Progress(pool != nullptr, "Allocating ThreadPool")) {
          int i;
-         Job jobs[MAX_THREADS];
+         Job jobs[MAX_THREADS+1];
 
-         for (i=0; i<MAX_THREADS; i++) {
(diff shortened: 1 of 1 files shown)


fastos/src/tests/threadtest.cpp
```diff
@@ -23,34 +23,29 @@ class ThreadTest : public ThreadTestBase
...
-
-         Progress(true, "Verifying result codes...");
-         for (i=0; i<MAX_THREADS; i++) {
-            Progress(jobs[i].result ==
-                     static_cast<int>(strlen(jobs[i].message)),
-                     "Checking result code from thread (%d==%d)",
-                     jobs[i].result, strlen(jobs[i].message));
          }
-
+          for (i=0; i<MAX_THREADS; i++) {
```
hmusum, line 45: Should it be `MAX_THREADS+1` here as well?

baldersheim: No, no this tests that we are not able to start the last thread. So the last one does is invalid.

hmusum (approved):

toregge (approved):

PR #16030, opened 2021-01-13
Ensure minimum window size is never less than 1.0, as that stops the …

jonmv:
…session silently

@vekterli please reivew.


vekterli (approved):
:+1:

PR #16033, opened 2021-01-13
use direct string id for small numbers

havardpe:
add stat for min_free (free address space in most filled part)

fail with assert when address space is exhausted, but not before

@arnej27959 @geirst please review
(this PR might impact tensor feeding performance; making it faster for numeric labels)


arnej27959 (approved):
👍

geirst (commented):
👍

PR #16034, opened 2021-01-13
Allow parenthesis in Yum package name

mpolden:
@hakonhall


hakonhall (approved):
who in their right mind would...

PR #16044, opened 2021-01-14
Remove unused provider yca fields

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/chains/search/DomProviderBuilder.java
```diff
@@ -62,10 +59,6 @@
             connectionPoolTimeout = readConnectionPoolTimeout(providerElement);
             retries = readRetries(providerElement);
             nodes = readNodes(providerElement);
-            certificateApplicationId = readCertificateApplicationId(providerElement);
```
bjorncs, line 65: Can these `readCertificate*` methods be removed as well?

bjorncs (approved):

config-model/src/main/java/com/yahoo/vespa/model/builder/xml/dom/chains/search/DomProviderBuilder.java
bratseth: Yes, done - thanks

PR #16048, opened 2021-01-14
Patch NetUtils so it will use hostname if it exists

hmusum:
In some environments we have Docker containers where we have internal
IP addresses (e.g. 172.x.x.x) that are only valid inside the container.
This change makes sure that we use hostname instead of theses addresses
when possible.

Please review only.


freva (approved):

hakonhall (approved):
Would have been nice if the first commit was a copy of upstream, so we could see the diff and make it easier to rebase later.

hmusum:
Good point @hakonhall , I will make a new PR

hmusum:
Replaced by https://github.com/vespa-engine/vespa/pull/16052

PR #16064, opened 2021-01-15
Bratseth/allocate max half nodes

bratseth:
@yngveaasheim fyi, we already capped autoscaling allocations to max half of hosts,  but using all defined host flavors. This switches to use max of hosts actually present in the zone instead.


freva (commented):
This wont work well in dynamically provisioned zones...

bratseth:
Yes, this code is invoked from the  
`( !clusterSpec.isExclusive() && !nodeRepository.zone().getCloud().dynamicProvisioning())
`
branch.

freva (approved):

PR #16071, opened 2021-01-15
Move logic for indexed document types inside config model api wrapping

jonmv:


jonmv:
Wrap to avoid class cast exception, as vespa models may be loaded with different class loader than the code using it here.

frodelu (approved):

PR #16060, opened 2021-01-15
- Add an BucketExecutor interface taking a BucketTask for executing a…

baldersheim:
… task by the content layer.

- Add method to register this interface.

@geirst @toregge @vekterli PR

PR #16072, opened 2021-01-15
Balder/add execute interface to persistence provider rebased2

baldersheim:
@toregge or @geirst PR
This is a rebase of #16060 with the last fixes according to comments


persistence/src/vespa/persistence/spi/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
geirst, line 1: Was adding this file intentional?

geirst (commented):
👍

persistence/src/vespa/persistence/spi/package.json
baldersheim: No certainly not. I wonder if this is something clion has generated.....
https://github.com/vespa-engine/vespa/pull/16080

PR #16079, opened 2021-01-18
Improved model for reindexing data

jonmv:
@bjorncs please reivew.

Making the users of reindexing data config aware allows getting rid of the
confusing default-for-app-and-cluster status, and the dummy first-time reindexing
in the reindexer.
Since the upgraded config servers will wipe the "common" status, the non-upgraded
config servers will read these as EPOCH, instead of some time last year. This should
not be a problem for any users of these data, as far as I can tell


bjorncs (approved):
👍 🎖️

PR #16082, opened 2021-01-18
Always use permanently down status

hakonhall:


hakonhall:
This has been in production for long time now

vekterli (approved):
:+1:

PR #16094, opened 2021-01-19
Tenant created at

freva:


jonmv (commented):
Any chance of computing a value for already existing tenants, instead of making it optional? Time of creation of earliest application under the tenant, or the ctime of the ZK node, perhaps?

bratseth:
Agree, it would be nice to avoid an optional here for all future if possible. Could also just set it to the launch date of hosted Vespa or something.

freva:
Changed it to zk node ctime or 2021-01-01T00:00:00Z. I assume the tenants are reserialized on controller startup, if so we can remove the supplier after this is released.

jonmv (approved):
Great, thanks!

PR #16095, opened 2021-01-19
Reject feed in distributor when feed in cluster is blocked

geirst:
@vekterli please review
@toregge FYI


vekterli (approved):
:+1: as a follow-up we also need to ensure that trivial updates that get converted into two-phase get/put-updates due to replica inconsistencies get bounced

PR #16103, opened 2021-01-19
Support group maintenance [run-systemtest]

hakonhall:
If at least 2 storage nodes in a hierarchical group is in maintenance with description "Orchestrator", then this PR will:
* set all remaining storage nodes in the group to maintenance
* set all remaining distributor nodes in the group to down


vekterli (approved):
:+1:

PR #16137, opened 2021-01-21
libzstd 1.4.7 needs more bytes for compressed data.

toregge:
@baldersheim : please review


baldersheim (approved):
I assume the format is backwards and backwards compatible....

PR #16142, opened 2021-01-21
Reinstate the check for progress.

baldersheim:
Remove the now defunct maxdocs parameter that was set explicit to 3 in this test.
It is void.

@geirst PR


geirst (commented):
👍

PR #16147, opened 2021-01-21
Wait for resources in prepare

freva:
VESPA-19467: Node-repository maintainers assume that once a `prepare()` goes through, the `activate()` will not fail, but in dynamically provisioned zones `activate()` will throw `TransientException`s until node's hosts are `active`. To avoid that, we should instead wait in `prepare()` until the parent hosts are ready, at least for internal deployments (for external deployments, the transient failures are retried by the controller and each failure provides live feedback to the user in the deployment log about the progress).


bratseth (commented):
If load balancers take a long time to get ready, it looks like we'll end up with a transaction with lots of duplicate node moves. Will that work?

Otherwise it looks fine, I hope we'll do the work later to check readiness explicitly though.

bratseth (approved):

PR #16150, opened 2021-01-21
Allow concurrent safe maintenance [run-systemtest]

hakonhall:
Allows setting a node safely to maintenance in these two new circumstances:

 1.  The node has state MAINTENANCE with (user) wanted state UP.
 2.  There are other nodes in the same hierarchical group that are set in
     MAINTENANCE with the same description.

Also made the following change.

 3. Deny a request for safe MAINTENANCE or DOWN, if the wanted state is already
    set but with a different description.  If the descriptions are the same, it
    is assumed to be the same operator (e.g. Orchestrator) having changed its
    mind.


vekterli (approved):
:+1:

PR #16153, opened 2021-01-21
Make a lidspace job that uses a bucket executor to hold the bucket lock.

baldersheim:
@geirst pr @toregge PR
This refactor the lid space job into a base class and the current based on frozen bucket, and the new that uses the content layer to acquire the proper bucket lock. No new code is wired in.
I will refactor the test in a separate PR.


baldersheim:
I have separated out the decoupling and non-semantic changes in #16154
I will rebase this one afterwards.

PR #16166, opened 2021-01-22
Add zstd as runtime dependency of Vespa

bjorncs:

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -162,6 +162,7 @@ Requires: gdb
 Requires: nc
 Requires: net-tools
 Requires: unzip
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -162,6 +162,7 @@ Requires: gdb
 Requires: nc
 Requires: net-tools
 Requires: unzip
+Requires: zstd # For inspection of compressed jdisc access log files
```
baldersheim, line 165: I think we already have vespa_zstd installed in /opt/vespa-dependencies/bin

bjorncs: Thanks!

PR #16168, opened 2021-01-22
Support delegating content node suspension to cluster controller

hakonhall:
This PR introduces a new flag group-suspension, which if true, enables:
 - Instead of allowing at most one storagenode to suspend at any given time, it
   will now ignore storagenode, searchnode, and distributor service clusters,
   and rely on the cluster controller to allow or deny the request to suspend.

   This will increase the load on the cluster controllers.

Combined with earlier changes to the cluster controller, this new flag
effectively guard the feature of allowing all nodes within a hierarchical group
to suspend concurrently.

I also took the opportunity to tune related policies:

 - Allow at most one config server and controller to be down at any given time.
   This is actually a no-op, since it was effectivelly equal to the older
   policy of 10% down.

 - Allows 20% of all host-admins to be down, not just tenant host-admins.  This
   is effectively equal to the old policy of 10% except that it may allow 2
   proxy host-admins to go down at the same time.  Should be fine.

### DIFF
--- a/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicy.java
+++ b/orchestrator/src/main/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicy.java
@@ -1,22 +1,39 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.orchestrator.policy;
 
+import com.yahoo.vespa.applicationmodel.ServiceType;
+import com.yahoo.vespa.flags.BooleanFlag;
+import com.yahoo.vespa.flags.FetchVector;
+import com.yahoo.vespa.flags.FlagSource;
+import com.yahoo.vespa.flags.Flags;
 import com.yahoo.vespa.orchestrator.model.ClusterApi;
 import com.yahoo.vespa.orchestrator.model.VespaModelUtil;
 
 import java.util.Optional;
+import java.util.Set;
 
 import static com.yahoo.vespa.orchestrator.policy.HostedVespaPolicy.ENOUGH_SERVICES_UP_CONSTRAINT;
 
 public class HostedVespaClusterPolicy implements ClusterPolicy {
 
+    private final BooleanFlag groupSuspensionFlag;
+
+    public HostedVespaClusterPolicy(FlagSource flagSource) {
+        // Note that the "group" in this flag refers to hierarchical groups of a content cluster.
+        this.groupSuspensionFlag = Flags.GROUP_SUSPENSION.bindTo(flagSource);
+    }
+
     @Override
     public SuspensionReasons verifyGroupGoingDownIsFine(ClusterApi clusterApi) throws HostStateChangeDeniedException {
+        boolean enableContentGroupSuspension = groupSuspensionFlag
+                .with(FetchVector.Dimension.APPLICATION_ID, clusterApi.getApplication().applicationId().serializedForm())
+                .value();
+
         if (clusterApi.noServicesOutsideGroupIsDown()) {
             return SuspensionReasons.nothingNoteworthy();
         }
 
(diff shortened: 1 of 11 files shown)


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicy.java
```diff
@@ -62,32 +81,110 @@ public void verifyGroupGoingDownPermanentlyIsFine(ClusterApi clusterApi)
...
+            }
 
-        if (VespaModelUtil.CLUSTER_CONTROLLER_SERVICE_TYPE.equals(clusterApi.serviceType())) {
-            // All nodes have all state and we need to be able to remove the half that are retired on cluster migration
-            return ConcurrentSuspensionLimitForCluster.FIFTY_PERCENT;
-        }
+            if (Set.of(ServiceType.STORAGE, ServiceType.SEARCH, ServiceType.DISTRIBUTOR, ServiceType.TRANSACTION_LOG_SERVER)
+                    .contains(clusterApi.serviceType())) {
+                // Delegate to the cluster controller
+                return ConcurrentSuspensionLimitForCluster.ALL_NODES;
```
vekterli, line 133: Any estimates on how much increased load the cluster controller(s) will be getting after this change?

vekterli (approved):
:+1:

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicy.java
hakonhall: Worst-case, all Docker hosts wants to reboot, and in case each content node may end up sending 2 requests (one probe and a follow-up non-probe) per 30s host-admin tick.  For large applications there are, say, 50 nodes, so QPS is 3.3.  This QPS would fairly flat, i.e. not bursty.  Once all hosts wants to reboot we'd hit this max QPS, and it would drop linearly as hosts were rebooted.

For a more realistic and common scenario, whenever a large application upgrades, they'll send 1 request/tick, i.e. a max QPS of 1.67 that will drop off linearly.

These high QPS MAY also appear today.  I looked at an application with 50 content nodes being upgraded:
 * The first request set node 92 to maintenance, but took longer than the 10s timeout to complete, so that node retried 30s later on the next tick, and when it got permission.
 * In the 20s window in between, there were 30s of accumulated queries: different nodes trying to get permission to suspend.  I counted 49 requests, i.e. QPS of 1.63 and confirming the theory.
 * However in the 10-13s window there was a burst of queries accumulated from 0-13s, and therefore 4x of normal high QPS.
 * The latencies were >10s for the queries that actually induce cluster state changes, which is expected.  All others have sub-0.1s response times, and spent cumulative 2.6s, which means the CC can handle sustained 20 QPS, or a cluster size of 600 for upgrades, or 300 for host suspensions.

orchestrator/src/test/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicyTest.java
```diff
@@ -34,33 +36,41 @@ public void setUp() {
...
-                policy.getConcurrentSuspensionLimit(clusterApi));
+                policy.getConcurrentSuspensionLimit(clusterApi, false));
+    }
+
+    @Test
+    public void testStorageSuspensionLimit_legacy() {
+        when(clusterApi.clusterId()).thenReturn(new ClusterId("some-cluster-id"));
+        when(clusterApi.isStorageCluster()).thenReturn(true);
+        assertEquals(ConcurrentSuspensionLimitForCluster.ONE_NODE,
+                policy.getConcurrentSuspensionLimit(clusterApi, false));
```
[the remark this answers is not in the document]
hakonhall: Thanks, fixed.

hakonhall:
SD failed on unrelated "Error: CircuitBreaker timeout".

hakonhall:
7.347.20

PR #16172, opened 2021-01-22
Write connection log entry directly to output stream [run-systemtest]

bjorncs:


tokle (approved):

bjorncs:
We need to fix the timing sensitive access log tests first. Some test assumes entries are immediately available (which is no longer the case after fixing buffering).

Issue #14845, opened 2020-10-13
Improve ONNX support with BatchNormalization


francoisWeber:
Hi Vespa Team, 
I'm using PyTorch to implement learning-to-rank for Vespa. I use ONNX models as serialization medium between PyTorch and Vespa.
However, it seems that a really useful ONNX operation is not yet supported inside Vespa : [BatchNormalization](https://github.com/onnx/onnx/blob/master/docs/Operators.md#BatchNormalization). It would be really usefull to add support for this operation soon 🙂

Best regards.
François Weber

jobergum:
Thanks a lot for your interest in Vespa.ai, 

We recently integrated ONNX-RT in Vespa. The documentation is not up to date with this change yet but we are fixing that this sprint. You can see an example here https://github.com/vespa-engine/sample-apps/blob/master/dense-passage-retrieval-with-ann/src/main/application/schemas/wiki.sd#L48-L55

Using **onnxModel()** will use the onnx-rt, while **onnx()** uses the Vespa implementation.

francoisWeber:
Thanks !

jobergum:
Thank you @francoisWeber, we will publish more on the onnx-rt integration and how to export PyTorch models to ONNX, how to optimize the ONNX graph (Graph optimizations, quantization etc) to improve run time serving performance. On CPU still, but see https://github.com/vespa-engine/vespa/issues/14406 for GPU. If you like us to implement GPU support feel free to add your cases to mentioned ticket. Thank you again.

francoisWeber:
Hi Vespa team, 
Sorry for re-opening this issue but I still have some trouble using the ONNX-RT feature that @jobergum spoke about recently. With the Vespa image` vespaengine/vespa:7.343.1`, I'm using an ONNX network involving `BatchNormalization` layers within my search-definition:
```
    onnx-model onnx_model {
        file: models/nn_bipedal_cuckoo_9fa0_10feats_7a1e.onnx
        input "input_10feats_7a1e": attribute(adult)
        output "output": output
    }

    rank-profile onnx_profile {
        first-phase {
            expression: onnxModel(onnx_model).output
        }
    }
```
This raises the following exception when deploying the new search-definition:
```
java.lang.IllegalArgumentException: Could not import nn_bipedal_cuckoo_9fa0_10feats_7a1e: Operation 'BatchNormalization' is currently not implemented
[...]

PR #16176, opened 2021-01-22
Skip, don't fail, if model can't be imported when put under models dir

lesters:
@bratseth For discussion. Ref https://github.com/vespa-engine/vespa/issues/14845. Models that are put under the `models` directory are automatically imported, and made available for instance for stateless evaluation. This uses our "old" import code that converts to Vespa expressions. However, a model that can be imported fine with ONNX Runtime, but is not supported by our code, will fail deployment if put in this directory.

This logs a warning instead and skips "global" import of the offending model. The model can still be used for onnx runtime however.

Should the deploy logger be used instead in this case?


bratseth (commented):
Yes, you should use the deploylogger as it will then show up in the users deploy log. Other than that, I agree this is better than the current behavior.

bratseth (approved):

PR #16183, opened 2021-01-23
Wire in the BucketExecutor.

baldersheim:
@toregge and @geirst PR


toregge (approved):

searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
```diff
@@ -44,16 +47,20 @@ CompactionJob::moveDocument(const search::DocumentMetaData & metaThen, std::shar
     if (metaNow.lid != metaThen.lid) return;
```
geirst, line 47: __scanItr->getMetaData()_ is now called by the bucket executor thread. Is that really safe? We are looking into a structure that is traditionally updated by the docdb writer thread. @toregge I would like your opinion here as well.

searchcore/src/vespa/searchcore/proton/server/maintenance_jobs_injector.cpp
```diff
@@ -34,15 +36,27 @@ injectLidSpaceCompactionJobs(MaintenanceController &controller,
...
-                config.getLidSpaceCompactionConfig(),
-                *lidHandler, opStorer, fbHandler,
-                diskMemUsageNotifier,
-                config.getBlockableJobConfig(),
-                clusterStateChangedNotifier,
-                (calc ? calc->nodeRetired() : false));
+
+        IMaintenanceJob::UP job;
+        if (config.getLidSpaceCompactionConfig().useBucketExecutor()) {
+            job = std::make_unique<LidSpaceCompactionJob>(
```
geirst, line 43: _lidspace::CompactionJob_ should be instantiated here, or we must negate the expression in the if statement.

searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
baldersheim: Yes, I came to the same conclusion myself in #16190

searchcore/src/vespa/searchcore/proton/server/maintenance_jobs_injector.cpp
baldersheim: Correct, fixed #16190
Very good observations, both you and factory came to the same conclusion :)

PR #16189, opened 2021-01-24
Try to reduce contention when bootstrapping tenants [run-systemtest]

hmusum:


bratseth (commented):
This may cause multiple threads to execute createAndWriteTenantMetaData(tenant) simultaneously but I think that's ok.

PR #16191, opened 2021-01-25
Depend on zstd tools

bjorncs:
Required for access log inspection (both on host and container in hosted environment).

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #16193, opened 2021-01-25
Revert "Depend on zstd tools"

hmusum:
Reverts vespa-engine/vespa#16191

Can't see exactly what's wrong, but building RPMs fail with:

error: line 242: Dependency tokens must begin with alpha-numeric, '_' or '/': Requires: zstd # Required for zstd tools used for inspection of access log files


aressem (approved):

aressem:
I suspect that the comment on the same line as the Requires is not supported.

PR #16194, opened 2021-01-25
Depend on zstd tools [run-systemtest]

bjorncs:
Required for access log inspection (both on host and container in hosted environment).


hmusum (approved):

bjorncs:
FYI @tokle

baldersheim:
I still do not see why this is necessary.
https://github.com/vespa-engine/vespa/pull/16191#issuecomment-766741821
Note that zstd is explicitly excluded since we uses vespa_zstd.

PR #16206, opened 2021-01-25
Add config to cluster controller on whether it should block feed base…

geirst:
…d on reported resource usage from content nodes.

Each resource category has a specified limit.
This is currently disabled by default.

@vekterli please review
@toregge FYI


vekterli (approved):
:+1:

PR #16214, opened 2021-01-25
Revert "Use synchronized backing for sessions"

hmusum:
Reverts vespa-engine/vespa#16208

Merge in case this breaks something.


hakonhall (approved):

hmusum:
Performance tests show much more time spent activating a session, need to look into it.

PR #16227, opened 2021-01-26
Gjoranv/remove state metric consumer

gjoranv:
FYI: @bjorncs


bratseth (approved):
Great! Let's just try it ... I'll leave merging to you though.

PR #16233, opened 2021-01-26
Remove the syncFeedView upcall as it has no purpose other than forcin…

baldersheim:
…g it more than necessary.

Instead do it in the main documentdb after all subds are reconfigured.

@toregge and @geirst PR


toregge (approved):

geirst (commented):
👍

PR #16250, opened 2021-01-27
Reduce inactiveExpiry to 2 sec in prod

bjormel:


bratseth:
This is a safety mechanism, I'm not sure we should removed it completely, and if so we should just remove the inactive state.

Alternatively, we should keep it for manual changes only.

bratseth:
After some discussion consensus is to do the latter (keep inactive for manual). I'll do that (unless you want to Bjørn).

bjormel:
It is all yours :)

PR #16257, opened 2021-01-27
Add cluster feed block support to cluster controller [run-systemtest]

vekterli:
@geirst please review

Will push out a new cluster state bundle indicating cluster feed blocked
if one or more nodes in the cluster has one or more resources exhausted.
Similarly, a new state will be pushed out once no nodes have resources
exhausted any more.

The feed block description currently contains up to 3 separate exhausted
resources, possibly across multiple nodes.

A cluster-level event is emitted for both the block and unblock edges.

No hysteresis is present yet, so if a node is oscillating around a block-limit,
so will the cluster state.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculator.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculator.java
@@ -1,0 +1,79 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.clustercontroller.core;
+
+import com.yahoo.vdslib.state.Node;
+import com.yahoo.vespa.clustercontroller.core.hostinfo.HostInfo;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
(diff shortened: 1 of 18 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculator.java
```diff
@@ -0,0 +1,79 @@
...
+        var exhaustions = enumerateNodeResourceExhaustionsAcrossAllNodes(nodeInfos);
+        if (exhaustions.isEmpty()) {
+            return null;
+        }
+        int maxDescriptions = 3;
+        String description = exhaustions.stream()
+                .limit(maxDescriptions)
+                .map(n -> String.format("%s on node %s (%.3g > %.3g)",
+                        n.resourceType, n.node.getIndex(),
+                        n.resourceUsage.getUsage(), n.limit))
```
geirst, line 42: Resource types “attribute-enum-store” and “attribute-multi-value” will also contain a name. This should be part of the description message, and I guess added to ResourceUsage.

geirst (approved):
👍

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculator.java
vekterli: Good point, I will add this as part of the follow-up work

PR #16264, opened 2021-01-28
Create module for jdisc aws resources

tokle:


jdisc-cloud-aws/src/main/java/com/yahoo/vespa/aws/AwsParameterStore.java
```diff
@@ -0,0 +1,21 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+package com.yahoo.vespa.aws;
```
oyving, line 3: Make this `com.yahoo.vespa.cloud.aws` to make it mirror the module name?

oyving (approved):

PR #16269, opened 2021-01-28
In order to ensure that lid stats are sampled at the right time ensur…

baldersheim:
…e that you do it the *next* time you are in the

master thread. This is to ensure that the sync call has taken effect.
This also keep the iterator creation logic in one method enhancing readability.

@toregge and @geirst PR


toregge (approved):

geirst (commented):
👍

PR #16275, opened 2021-01-28
Do not use vespamalloc for metrics-proxy. It is pure java.

baldersheim:
@hmusum or @gjoranv PR


hmusum (approved):

config-model/src/main/java/com/yahoo/vespa/model/AbstractService.java
```diff
@@ -53,7 +53,7 @@
 
     /** The optional PRELOAD libraries for this Service. */
     // Please keep non-null, as passed to command line in service startup
```
gjoranv, line 55: This comment seems out of place now ("Please keep non-null")

PR #16281, opened 2021-01-28
We only new a handfull of threads for the httpserver in the clusterco…

baldersheim:
…ntroller and metricsproxy.

@bjorncs PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/http/JettyHttpServer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/http/JettyHttpServer.java
@@ -7,6 +7,7 @@ import com.yahoo.config.model.api.ModelContext;
 import com.yahoo.container.bundle.BundleInstantiationSpecification;
 import com.yahoo.jdisc.http.ServerConfig;
 import com.yahoo.osgi.provider.model.ComponentModel;
+import com.yahoo.vespa.model.container.ApplicationContainerCluster;
 import com.yahoo.vespa.model.container.ContainerCluster;
 import com.yahoo.vespa.model.container.component.SimpleComponent;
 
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/http/JettyHttpServer.java
```diff
@@ -82,10 +75,17 @@ public void getConfig(ServerConfig.Builder builder) {
 
     private void configureJettyThreadpool(ServerConfig.Builder builder) {
         if (cluster == null) return;
+        if (cluster instanceof ApplicationContainerCluster) {
+            configureApplicationClusterJettyThreadPool(builder);
+        } else {
+            builder.minWorkerThreads(2).maxWorkerThreads(4);
```
bjorncs, line 81: Have you tried this? I'm pretty sure 2-4 is too low.

bjorncs (approved):

jdisc_http_service/src/test/java/com/yahoo/container/logging/JsonConnectionLogWriterTest.java
```diff
@@ -1,7 +1,7 @@
 package com.yahoo.container.logging;// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 import com.yahoo.test.json.JsonTestHelper;
-import org.junit.jupiter.api.Test;
+import org.junit.Test;
```
bjorncs, line 4: You are effectively disabling all tests here now

bjorncs (changes requested):

jdisc_http_service/src/test/java/com/yahoo/container/logging/JsonConnectionLogWriterTest.java
bjorncs: FYI `org.junit.jupiter.api` is junit5 while `org.junit` is junit4.

config-model/src/main/java/com/yahoo/vespa/model/container/http/JettyHttpServer.java
baldersheim: No, not tried. Will run a few system tests.

jdisc_http_service/src/test/java/com/yahoo/container/logging/JsonConnectionLogWriterTest.java
baldersheim: This was not intended. That was just a temporary thing to get intellij to run tests in config-model. It did not recognize the jupiter.api.

bjorncs (approved):

PR #16292, opened 2021-01-29
Support optional resource usage name field

vekterli:
@geirst please review

If present, will be reported alongside the resource type in the feed block description string.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResourceUsage.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResourceUsage.java
@@ -11,9 +11,12 @@ import java.util.Objects;
  */
 public class ResourceUsage {
     private final Double usage;
+    private final String name;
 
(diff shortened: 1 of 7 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResourceUsage.java
```diff
@@ -11,26 +11,33 @@
  */
 public class ResourceUsage {
     private final Double usage;
+    private final String name;
```
geirst, line 14: We could consider using Optional here, but I guess that doesn't work well with the @JsonProperty annotations used in the constructor.

geirst (approved):
👍

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResourceUsage.java
vekterli: Will consider this for a later follow-up

vekterli:
CI run seems to have failed before even starting compilation. Wasn't me!

PR #16296, opened 2021-01-31
Non-empty _pendingPrune will shortcut throttleing when doing flushAll.

baldersheim:
That is done on triggerFlush and prepareForRestart. That is incorrect as it causes a memory surge.

@toregge PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/flushengine/flushengine.h
+++ b/searchcore/src/vespa/searchcore/proton/flushengine/flushengine.h
@@ -79,7 +79,7 @@ private:
     uint32_t initFlush(const IFlushHandler::SP &handler, const IFlushTarget::SP &target);
     void flushDone(const FlushContext &ctx, uint32_t taskId);
     bool canFlushMore(const std::unique_lock<std::mutex> &guard) const;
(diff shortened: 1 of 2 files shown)


toregge (approved):

searchcore/src/vespa/searchcore/proton/flushengine/flushengine.h
```diff
@@ -79,7 +79,7 @@ class FlushEngine final : public FastOS_Runnable
     uint32_t initFlush(const IFlushHandler::SP &handler, const IFlushTarget::SP &target);
     void flushDone(const FlushContext &ctx, uint32_t taskId);
     bool canFlushMore(const std::unique_lock<std::mutex> &guard) const;
-    bool wait(vespalib::duration minimumWaitTimeIfReady);
+    bool wait(vespalib::duration minimumWaitTimeIfReady, bool considerPendingPrune);
```
geirst, line 82: This is called _ignorePendingPrune_ in the cpp file.

baldersheim: Fixed in #16298

PR #16301, opened 2021-02-01
keep serialized tensor in TensorFieldValue

arnej27959:
* do lazy deserialization in getTensor() and getTensorType()
* avoid re-serializing when forwarding a TensorFieldValue

note: this means any exceptions during deserializing (wrong tensor type, or garbage input) will be thrown at an unexpected time.

@geirst please review
@baldersheim FYI


geirst (approved):
👍  @baldersheim you might take a look as well.

baldersheim (approved):
LGTM

PR #16302, opened 2021-02-01
Add zstd as runtime dependency of Vespa

bjorncs:
Vespa needs zstd binaries to inspect access logs. Some vespa-zstd binaries that are
installed in vespa-deps binary directory does not work (e.g zstdless assumes zstdcat is on path).


aressem (approved):

bjorncs:
FYI @hmusum

PR #16317, opened 2021-02-01
7.350 is the last model using mergeChunkSize(), but return 32m instea…

baldersheim:
…d of throwing.

Minor fix to major problem in #16314 
@hmusum or @vekterli PR


hmusum (commented):
👍

PR #16324, opened 2021-02-02
Delegate blocking ops to executor in JobRunner

jonmv:
@mpolden please review.


jonmv:
Unused file deletion also staged, thanks to overly helpful IDEA :)

mpolden (approved):
👍

jonmv:
plong

PR #16279, opened 2021-01-28
Remove org.json usage [run-systemtest]

bjorncs:
https://github.com/vespa-engine/vespa/issues/14762

PR #16326, opened 2021-02-02
Remove usage of org.json [run-systemtest]

bjorncs:
@baldersheim  OR @hmusum please review

Same as #16279 , but without the changes to `container-core` and `metrics-proxy`


hmusum (approved):
👍

PR #16330, opened 2021-02-02
Arnej/use genspec in unit tests 2

arnej27959:

@havardpe please review


havardpe (approved):
Some thoughts:
* All layout vectors should be defined as const to avoid inplace modification
* Loops generating values with different cell types should be more explicit
* We should stop doing spec(number), but rather do GenSpec(number).gen()

PR #16331, opened 2021-02-02
Check for availability of attribute before using it.

toregge:
@baldersheim : please review


eval/src/vespa/eval/instruction/sparse_dot_product_function.cpp
```diff
@@ -35,15 +35,19 @@ void my_sparse_dot_product_op(InterpretedFunction::State &state, uint64_t num_ma
     const LCT *lhs_cells = state.peek(1).cells().typify<LCT>().cbegin();
     const RCT *rhs_cells = state.peek(0).cells().typify<RCT>().cbegin();
     if (auto indexes = detect_type<FastValueIndex>(lhs_idx, rhs_idx)) {
+#if __has_cpp_attribute(likely)
         [[likely]];
```
baldersheim, line 39: Is this a gcc 8/9 or a gcc/clang issue ?

baldersheim (approved):

eval/src/vespa/eval/instruction/sparse_dot_product_function.cpp
toregge: Both gcc 8 and clang 11 fails to compile the code due to attribute being unknown unless the test is added.

PR #16338, opened 2021-02-02
Gjoranv/hit metrics histogram

gjoranv:
Please review carefully. Is it necessary to declare the Values explicitly, and is logging of the mean for totalhits desirable?

I have verified in a docker container, that the metrics will be output from the metrics-proxy.

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/statistics/StatisticsSearcher.java
+++ b/container-search/src/main/java/com/yahoo/prelude/statistics/StatisticsSearcher.java
@@ -77,6 +77,7 @@ public class StatisticsSearcher extends Searcher {
     private final Value peakQPS; // peak 1s QPS
     private final Counter emptyResults; // number of results containing no concrete hits
     private final Value hitsPerQuery; // mean number of hits per query
+    private final Value hitsPerQueryBuckets;
 
     private final PeakQpsReporter peakQpsReporter;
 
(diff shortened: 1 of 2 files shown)


baldersheim (approved):
This looks fine to me, but this is unfamiliar territory so I think that @arnej27959 might want to have a look too.

container-search/src/main/java/com/yahoo/prelude/statistics/StatisticsSearcher.java
```diff
@@ -77,6 +77,9 @@
     private final Value peakQPS; // peak 1s QPS
     private final Counter emptyResults; // number of results containing no concrete hits
     private final Value hitsPerQuery; // mean number of hits per query
+    private final Value hitsPerQueryBuckets;
```
arnej27959, line 80: the two new "*Buckets" values doesn't do anything useful here, you can just remove them.

arnej27959 (approved):
try removing the extra Values, otherwise looks good.

container-search/src/main/java/com/yahoo/prelude/statistics/StatisticsSearcher.java
gjoranv: Done. Doesn't this mean that also the `queryLatencyBuckets` value is redundant?

arnej27959: queryLatency logs to "mean_query_latency" while queryLatencyBuckets is the one that logs to "query_latency".

actually, if you just want the new metrics in /state/v1/metrics (and so on) you should not need new Values at all; but they have the benefit of outputting their metric to vespa.log periodically which may be useful.

gjoranv: Thanks for clarifying! There were several points I had missed here.

PR #16339, opened 2021-02-02
Recompute cluster state if set of resource exhaustions changes

vekterli:
@geirst please review. Will follow up with changes to diff calculator so that logged events are emitted when the set changes as well.

Ensures that feed block description pushed to nodes is updated as
further resource exhaustions are recorded (or disappear).

### DIFF
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ClusterFeedBlockTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ClusterFeedBlockTest.java
@@ -23,6 +23,7 @@ import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.mapOf;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.setOf;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.usage;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.createResourceUsageJson;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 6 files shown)


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ClusterFeedBlockTest.java
```diff
@@ -133,4 +134,39 @@ public void cluster_feed_block_state_is_recomputed_when_options_are_updated() th
...
+        initialize(createOptions(mapOf(usage("cheese", 0.7), usage("wine", 0.4))));
+        assertFalse(ctrl.getClusterStateBundle().clusterFeedIsBlocked());
+
+        reportResourceUsageFromNode(1, setOf(usage("cheese", 0.8), usage("wine", 0.3)));
+        var bundle = ctrl.getClusterStateBundle();
+        assertTrue(bundle.clusterFeedIsBlocked());
+        assertEquals("cheese on node 1 [unknown hostname] (0.800 > 0.700)", bundle.getFeedBlock().get().getDescription());
+
+        // 80% -> 90%, should not trigger new state.
+        reportResourceUsageFromNode(1, setOf(usage("cheese", 0.9), usage("wine", 0.4)));
```
geirst, line 166: Nitpick: Should _wine_ be kept at 0.3 here? It doesn't affect the result, but the reader might wonder why it was changed.

geirst (approved):
👍

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ClusterFeedBlockTest.java
vekterli: Will change it back in a follow-up

PR #16345, opened 2021-02-02
Revert "Properly track execution of BucketTasks and provide sync() and order… "

baldersheim:
Reverts vespa-engine/vespa#16323
Still not good @vekterli PR


vekterli (commented):
:+1:

PR #16349, opened 2021-02-02
Restart container on any resource change

mpolden:
@freva and @hakonhall

Hold merge because all applications must upgrade to a version containing #16336 first.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
@@ -323,9 +323,9 @@ public class NodeAgentImpl implements NodeAgent {
         }
 
         ContainerResources wantedContainerResources = getContainerResources(context);
-        if (!wantedContainerResources.equalsMemory(existingContainer.resources)) {
-            reasons.add("Container should be running with different memory allocation, wanted: " +
-                    wantedContainerResources.toStringMemory() + ", actual: " + existingContainer.resources.toStringMemory());
+        if (!wantedContainerResources.equals(existingContainer.resources)) {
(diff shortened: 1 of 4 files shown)


freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -358,25 +358,6 @@ private void removeContainer(NodeAgentContext context, Container existingContain
...
-                .map(clock.instant().minus(warmUpDuration(context.zone()))::isBefore)
-                .orElse(true))
-            return existingContainer;
-
-        if (wantedContainerResources.equalsCpu(existingContainer.resources)) return existingContainer;
-        context.log(logger, "Container should be running with different CPU allocation, wanted: %s, current: %s",
-                wantedContainerResources.toStringCpu(), existingContainer.resources.toStringCpu());
-
-        // Only update CPU resources
-        containerOperations.updateContainer(context, wantedContainerResources.withMemoryBytes(existingContainer.resources.memoryBytes()));
```
hakonhall, line 375: Shoot - this we actually want.  When a container starts, it is given no cap on the CPU until the warmup period is over to give the JVM sufficient resources to JIT the code, and in case the cap is reinstated.  This was implemented not too long ago and I'm guessing we want to continue supporting it, but we can look back to see whether the benefits still makes sense.  Orchestrating such a change defeats the purpose as the JVM is restarted.

@freva , did we discuss removing support for this too?

Didn't we conclude that implementing this would probably be as simple as maintaining the 2 corresponding cgroups values?  We should then also read them properly to ensure we're not getting stale data from podman.

mpolden:
Superseded by https://github.com/vespa-engine/vespa/pull/16404.

PR #16355, opened 2021-02-02
The StateManager is destructed before the FileStorManager leading to …

baldersheim:
…possibly accesing invalid memory. Disconnecting upcalls must happen during close.

@toregge or @vekterli PR


toregge (approved):

vekterli (commented):
:+1:

PR #16357, opened 2021-02-02
Remove the sync interface from Bucket Executor. Due to the nature of …

baldersheim:
…requiring a bucket lock it is very hard to get sync to work in a safe way.

Instead the users must do their own accounting as they know their own threading model.

@vekterli and @toregge PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
+++ b/searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
@@ -11,6 +11,7 @@
 #include <vespa/vespalib/util/destructor_callbacks.h>
 #include <vespa/vespalib/util/lambdatask.h>
 #include <cassert>
+#include <thread>
 
 using search::DocumentMetaData;
 using search::LidUsageStats;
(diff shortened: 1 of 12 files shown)


searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
```diff
@@ -82,20 +84,25 @@ CompactionJob::CompactionJob(const DocumentDBLidSpaceCompactionConfig &config,
...
+CompactionJob::inSync() const {
+    return _executedCount == _startedCount;
 }
 
 void
 CompactionJob::onStop() {
-    sync();
+    _stopped = true;
+    while ( ! inSync() ) {
+        std::this_thread::sleep_for(10us);
```
toregge, line 104: Consider 1ms instead of 10us.

searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
```diff
@@ -34,19 +35,19 @@ CompactionJob::scanDocuments(const LidUsageStats &stats)
...
     }
-    if (!_scanItr->valid()) {
-        sync();
-    }
     return false;
 }
 
 void
 CompactionJob::moveDocument(const search::DocumentMetaData & meta, std::shared_ptr<IDestructorCallback> context) {
+    _executedCount.fetch_add(1, std::memory_order_relaxed);
```
toregge, line 49: `_executedCount` should be increased after queuing a task on the `_master` executor (in the normal case) and in the early returns. Otherwise, `onStop()` might return after a bucket executor has increased `_executedCount` but before it has queued the task, and `MaintenanceController::killJobs()` might queue the task to drop held jobs too early. The task eventually queued by the bucket executor could then access freed memory.

toregge (changes requested):

searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
baldersheim: 1ms it is.

searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job_take2.cpp
baldersheim: Agree, added a guard that should count at the right places.

toregge (approved):

vekterli (approved):
:+1:

PR #16361, opened 2021-02-03
Extend flag lifetime.

baldersheim:
@hmusum or @vekterli PR


hmusum (approved):
If RESPONSE_SEQUENCER_TYPE is a permant flag it could be moved toPermanentFlags

baldersheim:
Not permanent, but there are a set of flags related to feeding that we will be experimenting with probably then a 6-12 months.

hmusum:
👍

vekterli (commented):
:+1:

PR #16362, opened 2021-02-03
Improve the quota error messages

oyving:


kkraune (approved):
better!

bratseth (commented):
"Free up some quota" sounds strange to me. What they can free up are the resources they use.

PR #16363, opened 2021-02-03
Gjoranv/cleanup state handler

gjoranv:
Cleanup after removing the StateMetricsConsumer.

EDIT: I originally forgot to fix the actual TODOs, done in the third commit.


bratseth (approved):

gjoranv:
Thanks, that was faster than lightning..

PR #16365, opened 2021-02-03
Bump intervals of maintainers still colliding

mpolden:
Adaptive interval would've been neat.

@jonmv


jonmv (approved):

jonmv:
With a lower limit ... :)

PR #16376, opened 2021-02-03
Add feature flag for max dead bytes ratio.

toregge:
@baldersheim : please review

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -259,6 +259,12 @@ public class Flags {
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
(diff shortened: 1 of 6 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -259,6 +259,12 @@
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundDoubleFlag MAX_DEAD_BYTES_RATIO = defineDoubleFlag(
+            "max-dead-bytes-ratio", 0.2,
+            List.of("baldersheim", "geirst","toregge"), "2021-02-03", "2022-02-03",
```
baldersheim, line 264: Let it expire at end of Q1 2021-04-01: We should have rolled out a sensible default value by then.

baldersheim (approved):

baldersheim (approved):

PR #16378, opened 2021-02-03
Scale content clusters to minimum 3 nodes

bratseth:
There is no cluster controller redundancy with 2 nodes
and this leads to operational problems.


freva (approved):
What happens if this is this min is larger than nodes limit max, e.g.
```
<nodes count="2">
  <resources vcpu="[2, 4]" ...
</node>
```
This means that limits is not empty, but nodes min = max = 2, but we will get 3 nodes?

bratseth:
Great point,  fixed and merging.

PR #16389, opened 2021-02-04
Reduce NameServiceDispatcher interval

mpolden:
@ldalves


mpolden:
or @jonmv. Increasing it made dispatch too slow.

jonmv (approved):

jonmv:
If it needs to run often, but runs for a long time when it does, we just have to accept some collisions :)

PR #16392, opened 2021-02-04
Add resource usage per (content) node to cluster controller status page

vekterli:
@geirst please review.

Also adds top-level cluster feed block status. Does not yet make
enum store/multivalue limit feed blocks visible per node.

Disclaimer: this HTML 1.0 masterpiece is not likely to win (m)any prestigious design awards.


geirst (approved):
👍

PR #16409, opened 2021-02-05
Ignore maintenance collisions in controller and node-repository

mpolden:
Concluded that this approach is fine. Some maintainers have unpredictable run
durations while also needing to run as often as possible, which causes frequent
timeouts and log noise.

@jonmv


jonmv (approved):

jonmv:
:)

mpolden:
.

PR #16410, opened 2021-02-05
Only block for a short time when draining hold lists. MERGEOK

toregge:
@baldersheim : please review
@geirst : FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/attributevector.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/attributevector.cpp
@@ -774,7 +774,7 @@ void
 AttributeVector::drain_hold(uint64_t hold_limit)
 {
     incGeneration();
-    for (int retry = 0; ; ++retry) {
+    for (int retry = 0; retry < 40; ++retry) {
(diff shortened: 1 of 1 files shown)


searchlib/src/vespa/searchlib/attribute/attributevector.cpp
```diff
@@ -788,11 +788,16 @@ void
 AttributeVector::update_config(const Config& cfg)
 {
     commit(true);
-    drain_hold(1024 * 1024); // Wait until 1MiB or less on hold
+    bool compaction_strategy_changed = (cfg.getCompactionStrategy() != _config.getCompactionStrategy());
```
baldersheim, line 791: I think it would be easier to read if you moved _config.setGrowStrategy(cfg.getGrowStrategy()) first, and then added a guard if (cfg.getCompactionStrategy() == _config.getCompactionStrategy()) return;
Then you do not need the 2 ifs.

baldersheim (approved):

baldersheim (approved):

Issue #16360, opened 2021-02-03
Simple support for replacing OR by WAND


bratseth:
Most that do OR queries over text end up wanting to use WAND for performance, but switching to that is a bit of work.
This can be made easier by creating a searcher which replaces OrItems over text indexes by WeakAndItems.
This could be included as a noop by default (in the 'local' chain) and activated by a query parameter.

jobergum:
Also relates to https://github.com/vespa-engine/vespa/issues/13076

johans1:
Done

PR #16411, opened 2021-02-05
Karowan/replacing or by wand 16360

karowan:

Pull request for #16360. This searcher recursively replaces all instances of OrItems with WeakAndItems if the query property weakand.replace is true and is otherwise a noop searcher. 

If there is a preferable name for the searcher or activating property, let me know and I will change it to whatever is asked.

I am not incredibly familiar with adding a searcher to the local chain so I will wait for some guidance on that aspect. I am assuming that it is just a matter of adding it to the [LocalProviderSpec](https://github.com/vespa-engine/vespa/blob/master/container-search/src/main/java/com/yahoo/search/searchchain/model/federation/LocalProviderSpec.java) but there could be a lot of things that I might be missing so I will refrain from doing so.

Thank you for any feedback.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/querytransform/WeakAndReplacementSearcher.java
+++ b/container-search/src/main/java/com/yahoo/search/querytransform/WeakAndReplacementSearcher.java
@@ -1,0 +1,43 @@
+package com.yahoo.search.querytransform;
+
+import com.yahoo.prelude.query.*;
+import com.yahoo.processing.request.CompoundName;
+import com.yahoo.search.Query;
+import com.yahoo.search.Result;
+import com.yahoo.search.Searcher;
+import com.yahoo.search.searchchain.Execution;
+
+public class WeakAndReplacementSearcher extends Searcher {
+    private static final CompoundName WEAKAND_REPLACE = new CompoundName("weakand.replace");
+
+    @Override public Result search(Query query, Execution execution) {
+        if (!query.properties().getBoolean(WEAKAND_REPLACE)) {
+            return execution.search(query);
+        }
+        replaceOrItems(query);
+        return execution.search(query);
+    }
+
+    private void replaceOrItems(Query query) {
+        Item root = query.getModel().getQueryTree().getRoot();
+        int hits = query.properties().getInteger("wand.hits", WeakAndItem.defaultN);
+        query.getModel().getQueryTree().setRoot(replaceOrItems(root, hits));
+    }
+
+    private Item replaceOrItems(Item item, int hits) {
+        if (!(item instanceof CompositeItem)) {
+            return item;
+        }
+        CompositeItem compositeItem = (CompositeItem) item;
+        if (item instanceof OrItem) {
+            WeakAndItem newItem = new WeakAndItem(hits);
+            newItem.setWeight(item.getWeight());
+            compositeItem.items().forEach(newItem::addItem);
+            compositeItem = newItem;
+        }
+        for (int i = 0; i < compositeItem.getItemCount(); i++) {
+            compositeItem.setItem(i, replaceOrItems(compositeItem.getItem(i), hits));
+        }
+        return compositeItem;
+    }
+}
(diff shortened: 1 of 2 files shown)


container-search/src/main/java/com/yahoo/search/querytransform/WeakAndReplacementSearcher.java
```diff
@@ -0,0 +1,62 @@
...
+import com.yahoo.search.Result;
+import com.yahoo.search.Searcher;
+import com.yahoo.search.searchchain.Execution;
+
+/**
+ * Recursively replaces all instances of OrItems with WeakAndItems if the query property weakand.replace is true.
+ * Otherwise a noop searcher
+ */
+public class WeakAndReplacementSearcher extends Searcher {
+    private static final CompoundName WEAKAND_REPLACE = new CompoundName("weakand.replace");
```
bratseth, line 15:
```suggestion
    private static final CompoundName WEAKAND_REPLACE = new CompoundName("weakAnd.replace");
```

container-search/src/main/java/com/yahoo/search/querytransform/WeakAndReplacementSearcher.java
```diff
@@ -0,0 +1,62 @@
...
+import com.yahoo.prelude.query.*;
+import com.yahoo.processing.request.CompoundName;
+import com.yahoo.search.Query;
+import com.yahoo.search.Result;
+import com.yahoo.search.Searcher;
+import com.yahoo.search.searchchain.Execution;
+
+/**
+ * Recursively replaces all instances of OrItems with WeakAndItems if the query property weakand.replace is true.
+ * Otherwise a noop searcher
```
bratseth, line 12:
```suggestion
 * Otherwise a noop searcher.
 *
 * @author karowan
```

container-search/src/main/java/com/yahoo/search/querytransform/WeakAndReplacementSearcher.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Extracts the queryTree root and the wand.hits property to send to the recursive replacement function
+     * @param query the search query
+     */
+    private void replaceOrItems(Query query) {
+        Item root = query.getModel().getQueryTree().getRoot();
+        int hits = query.properties().getInteger("wand.hits", WeakAndItem.defaultN);
+        query.getModel().getQueryTree().setRoot(replaceOrItems(root, hits));
```
bratseth, line 32:
```suggestion
        query.getModel().getQueryTree().setRoot(replaceOrItems(root, hits));
        if (root != query.getModel().getQueryTree().getRoot())
            query.trace("Replaced OR by WeakAnd", true, 2);
```

bratseth (approved):
Great! 

After some consideration I think it's better to keep this schema independent and thus add it to the "vespa" chain instead, but we can do that in a separate PR.

karowan:
Awesome!

I'm not familiar with all the procedures required for adding a searcher to the vespa chain so I would appreciate some assistance on that to make sure all the required steps are taken when I open the PR. Again, I'm assuming that it is just adding it to the [VespaSearchers.java](https://github.com/vespa-engine/vespa/blob/master/container-search/src/main/java/com/yahoo/search/searchchain/model/VespaSearchers.java#L36) but I'm not sure if there are any config files/tests associated with that which will need updating. If you go ahead and add it yourself I would really appreciate sending that PR to me so that I can learn from it for the future.

bratseth:
Nothing else is needed besides adding it to VespaSearchers.

karowan:
Ok great I'll open up a PR for that then.

jobergum:
Thank you very much @karowan for this contributio. If you have time,  could you also document this new parameter over at the search api guide and in the high level using wand doc ?

- https://docs.vespa.ai/en/reference/query-api-reference.html
- https://docs.vespa.ai/en/using-wand-with-vespa.html

karowan:
@jobergum Yes I can definitely do that. I'll open a PR for that sometime today.

PR #16412, opened 2021-02-05
fix bug in lazy deserialization MERGEOK

arnej27959:
* the API used by concrete documents needs to perform
  lazy deserialization also.


geirst (commented):
👍  But I suggest adding a test case in src/test/java/com/yahoo/document/datatypes/TensorFieldValueTestCase.java as well.

baldersheim (approved):

PR #16417, opened 2021-02-05
Bratseth/less inactive 

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -515,7 +515,13 @@ public void setRemovable(ApplicationId application, List<Node> nodes) {
      * transaction commits.
      */
     public List<Node> deactivate(List<Node> nodes, ApplicationTransaction transaction) {
-        return db.writeTo(State.inactive, nodes, Agent.application, Optional.empty(), transaction.nested());
+        var stateless = NodeList.copyOf(nodes).stateless();
+        var stateful  = NodeList.copyOf(nodes).stateful();
```
mpolden, line 519: Nit: Consider copying once.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -568,7 +564,39 @@ public Node setDirty(Node node, Agent agent, String reason) {
...
 
-        return nodesToDirty.stream().map(node -> setDirty(node, agent, reason)).collect(Collectors.toList());
+        return nodesToDirty.stream().map(node -> deallocate(node, agent, reason)).collect(Collectors.toList());
+    }
+
+    /**
+     * Set a node dirty  or parked, allowed if it is in the provisioned, inactive, failed or parked state.
+     * Use this to clean newly provisioned nodes or to recycle failed nodes which have been repaired or put on hold.
+     *
+     * @throws IllegalArgumentException if the node has hardware failure
```
mpolden, line 574: Where is this thrown?

bratseth: I just moved this ... but you are right it doesn't look like it is thrown anywhere so I removed it.

mpolden (approved):

PR #16427, opened 2021-02-07
To ensure that lid stats is sampled after everything has been complet…

baldersheim:
…ed compaction must happen in 2 stages.

First it must reach inSync, then it must rescheduled once more in the master thread. Then all movement is visible in the master thread.

@geirst or @toregge PR


geirst (approved):
👍

PR #16428, opened 2021-02-08
Bratseth/node repo refactor

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancers.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private LoadBalancerList list(Predicate<LoadBalancerId> predicate) {
+        return LoadBalancerList.copyOf(db.readLoadBalancers(predicate).values());
+    }
+
+    /**
+     * Returns the ACL for the node (trusted nodes, networks and ports)
+     */
+    private NodeAcl getNodeAcl(Node node, NodeList candidates) {
```
mpolden, line 53: This doesn't belong here. It creates a list of addresses and ports a node should trust, which is unrelated to load balancers (load balancer networks are just one of the things to trust).

It should be moved back, or into a separate class (`NodeAcls`?).

bratseth: I was afraid you'd say that because it is true.
But only logically true, since load balancers is the only *state* they depend on, and the tree of singleton objects below the node repo are all about state.
The more correct way is to put this below node and pass in the load balancers ... it's just more work. I'll do that then.

mpolden: > load balancers is the only state they depend on

They also depend on node state. `public List<NodeAcl> getNodeAcls(Node node, boolean children)` reads all nodes.

bratseth:
@mpolden ptal. 

The child acl method should be under Host when we get it, so leaving it for now.

Afaik we always pass children=true,  should we remove the option to request just one childs ACL?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -213,15 +208,30 @@ private void rewriteNodes() {
...
     public int spareCount() { return spareCount; }
 
+    /**
+     * Returns ACLs for the children of the given host.
+     *
+     * @param host node for which to generate ACLs
+     * @return the list of node ACLs
+     */
+    public List<NodeAcl> getChildAcls(Node host) {
+        if ( host.type() != NodeType.host) throw new IllegalArgumentException("Only hosts have children");
```
mpolden, line 228: Should check `isHost()` instead.

mpolden (commented):
> Afaik we always pass children=true, should we remove the option to request just one childs ACL?

Consider adding a TODO, I haven't checked. Looks like the client decides this (REST API accepts a `children` parameter).

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
bratseth: Done -  thanks.  I see we have many other NodeType.host comparisons as well. I guess all should likely use isHost(), or have you looked at that before?

mpolden: Yes, generally we should use `isHost()`, but I think there is a lot of code that only applies to tenant hosts.

I did a quick check on usages in this module, and looks like `NodeType.host` is mostly used in dynamic provisioning, capacity decisions related to node allocation and rebalancing. I.e. cases where you really mean `NodeType.host`.

mpolden (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancers.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Comparator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * The load balancers of this node repo and the information (such asd ACL's) which need load balancer information.
```
mpolden, line 21: This should stop mentioning ACLs now.

bratseth: Fixed, thanks

PR #16429, opened 2021-02-08
Deprecate score-threshold

bratseth:
@arnej27959 I don't think we're going to bring that back?


arnej27959:
I’m not sure if the part about RISE is really true. Some people also wanted a distance threshold for ANN, and said they wanted it to be similar to this threshold, for some reason. @geirst can you check if this score threshold is potentially useful?

geirst:
The score threshold is no longer used by the WeakAnd operator in the backend. RISE is long gone. I think it's fine to remove.

geirst (approved):
👍

arnej27959 (approved):
so the WandItem score threshold is in use, but not the WeakAndItem one? very well..

PR #16433, opened 2021-02-08
Print info when detecting hostname fails and we fallback to localhost

hmusum:
Naive attempt at printing a bit more when detecting hostname fails

### DIFF
--- a/vespalib/src/apps/vespa-detect-hostname/detect_hostname.cpp
+++ b/vespalib/src/apps/vespa-detect-hostname/detect_hostname.cpp
@@ -49,6 +49,10 @@ int main(int, char **) {
     if (check(my_hostname, my_ip_set, my_hostname_error)) {
         fprintf(stdout, "%s\n", my_hostname.c_str());
     } else if (check(localhost, my_ip_set, localhost_error)) {
+        fprintf(stderr, "INFO: hostname detection failed, falling back to local hostname: %s", localhost.c_str());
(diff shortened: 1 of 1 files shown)


vespalib/src/apps/vespa-detect-hostname/detect_hostname.cpp
```diff
@@ -49,6 +49,10 @@ int main(int, char **) {
     if (check(my_hostname, my_ip_set, my_hostname_error)) {
         fprintf(stdout, "%s\n", my_hostname.c_str());
     } else if (check(localhost, my_ip_set, localhost_error)) {
+        fprintf(stderr, "INFO: hostname detection failed, falling back to local hostname: %s", localhost.c_str());
```
mpolden, line 52: Change to `WARN`?

vespalib/src/apps/vespa-detect-hostname/detect_hostname.cpp
```diff
@@ -49,6 +49,10 @@ int main(int, char **) {
...
+        fprintf(stderr, "  INFO: canonical hostname (from gethostname/getaddrinfo): %s\n", my_hostname.c_str());
+        fprintf(stderr, "  INFO: %s\n", my_hostname_error.c_str());
```
mpolden, line 54: And here.

hmusum: Yes, fixed

mpolden (approved):

PR #16438, opened 2021-02-08
exclude temporary deployments when calculating quota

andreer:


bratseth (commented):
What is "QuotaUsage" used for? We should charge for test resources ...

oyving:
`QuotaUsage` is how many resources/hour an application can hypothetically consume given a cluster scaled to max. It's only used to make sure tenants do not consume too many resources. Quota and Metering are two related concepts, but they do not overlap completely. Changing this will not affect metering.

oyving (approved):

PR #16450, opened 2021-02-09
Remove unused code

hmusum:
We will not use container nodes for cluster controllers anymore.


jonmv (approved):
But maybe we _do_ want to use this ... right now? Not in this state, though.

PR #16458, opened 2021-02-10
add flag to sync host logs to s3 bucket

andreer:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -214,6 +214,13 @@
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundStringFlag SYNC_HOST_LOGS_TO_S3_BUCKET = defineStringFlag(
+            "sync-host-logs-to-s3-bucket", "", List.of("andreer", "valerijf"), "2021-02-10", "2021-03-01",
+            "Host-admin should sync host logs to an S3 bucket named by this flag. If left empty, sync is disabled",
+            "Takes effect on next run of S3 log sync task in host-admin",
+            ZONE_ID
```
freva, line 221: Zone ID is not needed since this is a zone level flag.
Add Application ID as dimension?

freva (approved):

andreer:
Merging as build succeeded locally.

PR #16463, opened 2021-02-10
Add flag for controller JVM max heap size for cluster controller

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -218,8 +218,13 @@
             "sync-host-logs-to-s3-bucket", "", List.of("andreer", "valerijf"), "2021-02-10", "2021-03-01",
             "Host-admin should sync host logs to an S3 bucket named by this flag. If left empty, sync is disabled",
             "Takes effect on next run of S3 log sync task in host-admin",
-            APPLICATION_ID
-    );
+            APPLICATION_ID);
+
+    public static final UnboundIntFlag CLUSTER_CONTROLLER_MAX_HEAP_SIZE_IN_MB = defineIntFlag(
+            "cluster-controller-max-heap-size-in-mb", 512,
```
baldersheim, line 224: Is 512 current max ?

hmusum: Yes

baldersheim (approved):

PR #16464, opened 2021-02-10
Use feature flag for max heap size for cluster controller

hmusum:
Feature flag default is 512, so no change of default value

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainerCluster.java
@@ -1,9 +1,11 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.admin.clustercontroller;
 
+import com.yahoo.config.model.api.ModelContext;
 import com.yahoo.config.model.api.Reindexing;
 import com.yahoo.config.model.deploy.DeployState;
 import com.yahoo.config.model.producer.AbstractConfigProducer;
+import com.yahoo.search.config.QrStartConfig;
 import com.yahoo.vespa.model.container.ContainerCluster;
 
 /**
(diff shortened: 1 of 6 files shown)


hmusum:
PR builds does not work right now

baldersheim (approved):

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainerCluster.java
```diff
@@ -28,6 +32,23 @@ protected void doPrepare(DeployState deployState) { }
 
     @Override protected boolean messageBusEnabled() { return false; }
 
+    @Override
+    public void getConfig(QrStartConfig.Builder builder) {
+        int maxHeapSize = featureFlags.clusterControllerMaxHeapSizeInMb();
+        boolean verboseGc = (maxHeapSize < 512);
+        builder.jvm
```
baldersheim, line 39: Where was this produced earlier on ?

hmusum: ContainerCluster

jonmv: Would prefer a call to `super.getConfig(builder)`, followed by modifications, to avoid duplication.

jonmv:
Missing actual binding to a flag?

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainerCluster.java
hmusum: Yup, fixed

hmusum:
The code in ModelContextImpl, I think this is the new way of doing this

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -179,6 +180,7 @@ public FeatureFlags(FlagSource source, ApplicationId appId) {
             this.useBucketExecutorForLidSpaceCompact = flagValue(source, appId, Flags.USE_BUCKET_EXECUTOR_FOR_LID_SPACE_COMPACT);
             this.enableFeedBlockInDistributor = flagValue(source, appId, Flags.ENABLE_FEED_BLOCK_IN_DISTRIBUTOR);
             this.maxDeadBytesRatio = flagValue(source, appId, Flags.MAX_DEAD_BYTES_RATIO);
+            this.clusterControllerMaxHeapSizeInMb = flagValue(source, appId, Flags.CLUSTER_CONTROLLER_MAX_HEAP_SIZE_IN_MB);
```
jonmv, line 183: Yeah, sorry, missed this.

jonmv: But that enum constant I don't see!

hmusum: Added in an earlier PR (https://github.com/vespa-engine/vespa/pull/16463), that @baldersheim merged before I added any reviewers, so forgot to add you afterwards.

PR #16468, opened 2021-02-10
Support configurable feed block hysteresis on the cluster controller

vekterli:
@geirst please review

Adds an absolute number delta that is subtracted from the feed block limit
when a node has a resource already in feed blocked state. This means that
there's a lower watermark threshold that must be crossed before feeding
can be unblocked. Avoids flip-flopping between block states.

Default is currently 0.0, i.e. effectively disabled. To be modified
later for system tests and trial roll-outs.

A couple of caveats with the current implementation:
* The cluster state is not recomputed automatically when just the hysteresis
  threshold is crossed, so the description will be out of date on the
  content nodes. However, if any other feed block event happens (or the
  hysteresis threshold is crossed), the state will be recomputed as expected.
  This does not affect correctness, since the feed is still to be blocked.
* A node event remove/add pair is emitted for feed block status when the
  hysteresis threshold is crossed and there's a cluster state recomputation.

### DIFF
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
@@ -5,8 +5,10 @@ import org.junit.Test;
 
 import static com.yahoo.vespa.clustercontroller.core.ClusterFixture.storageNode;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.createFixtureWithReportedUsages;
+import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.exhaustion;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.forNode;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.mapOf;
+import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.setOf;
 import static com.yahoo.vespa.clustercontroller.core.FeedBlockUtil.usage;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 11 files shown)


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
```diff
@@ -98,4 +100,45 @@ public void no_feed_block_returned_when_feed_block_disabled() {
...
+    public void retain_node_feed_block_status_when_within_hysteresis_window_limit_crossed_edge_case() {
+        var curFeedBlock = ClusterStateBundle.FeedBlock.blockedWith("foo", setOf(exhaustion(1, "memory", 0.51)));
+        var calc = new ResourceExhaustionCalculator(true, mapOf(usage("disk", 0.5), usage("memory", 0.5)), curFeedBlock, 0.1);
+        // Node 1 goes from 0.51 to 0.49, crossing the 0.5 threshold. Should still be blocked.
+        // Node 2 is at 0.49 but was not previously blocked and should not be blocked now either.
+        var cf = createFixtureWithReportedUsages(forNode(1, usage("disk", 0.3), usage("memory", 0.49)),
+                                                 forNode(2, usage("disk", 0.3), usage("memory", 0.49)));
+        var feedBlock = calc.inferContentClusterFeedBlockOrNull(cf.cluster().getNodeInfo());
+        assertNotNull(feedBlock);
+        // TODO should we not change the limits themselves? Explicit mention of hysteresis state?
```
geirst, line 113: Yes, I think that would be nice. It would make the error messages less confusing.

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
```diff
@@ -98,4 +100,45 @@ public void no_feed_block_returned_when_feed_block_disabled() {
...
+        assertEquals("memory on node 1 [storage.1.local] (0.490 > 0.400)",
+                     feedBlock.getDescription());
+    }
+
+    @Test
+    public void retain_node_feed_block_status_when_within_hysteresis_window_under_limit_edge_case() {
+        var curFeedBlock = ClusterStateBundle.FeedBlock.blockedWith("foo", setOf(exhaustion(1, "memory", 0.49)));
+        var calc = new ResourceExhaustionCalculator(true, mapOf(usage("disk", 0.5), usage("memory", 0.5)), curFeedBlock, 0.1);
+        // Node 1 goes from 0.49 to 0.48, NOT crossing the 0.5 threshold. Should still be blocked.
```
geirst, line 122: Is it the adjusted threshold (0.4) we should use in the comment?

geirst (approved):
👍

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
vekterli: In this case, the comment is referring to the original feed limit, not the adjusted hysteresis limit. It tests going from feed blocked above the hysteresis limit (but below the original limit) to feed blocked _slightly less_ above the hysteresis limit.

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/ResourceExhaustionCalculatorTest.java
vekterli: Will do as a separate quality-of-life pass later

PR #16477, opened 2021-02-11
Avoid possible concurrent modification

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -73,9 +73,7 @@ import com.yahoo.vespa.curator.Curator;
 import com.yahoo.vespa.curator.stats.LockStats;
 import com.yahoo.vespa.curator.stats.ThreadLockStats;
 import com.yahoo.vespa.defaults.Defaults;
-import com.yahoo.vespa.flags.BooleanFlag;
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -835,7 +830,9 @@ public long createSession(ApplicationId applicationId, TimeoutBudget timeoutBudg
 
     public void deleteExpiredLocalSessions() {
         Map<Tenant, Collection<LocalSession>> sessionsPerTenant = new HashMap<>();
-        tenantRepository.getAllTenants().forEach(tenant -> sessionsPerTenant.put(tenant, tenant.getSessionRepository().getLocalSessions()));
+        tenantRepository.getAllTenants()
+                        .forEach(tenant -> sessionsPerTenant.put(tenant,
+                                                             Collections.unmodifiableCollection(tenant.getSessionRepository().getLocalSessions())));
```
jonmv, line 835: I think you want a copy—from what I can see, modifications to the underlying collection will still cause `ConcurrentModificationException` when you iterate through the unmodifiable view, as that just forwards to the underlying collection.

hmusum: Fixed

jonmv (approved):

PR #16481, opened 2021-02-11
Exit ZooKeeper server if reconfiguration fails

hmusum:
When going from 2 to 1 zookeeper servers (after going from 1 to 2 nodes
because a node is about to be retired), reconfiguration will fail
if the second node is removed before reconfiguration has happened.
Reconfiguration will never work in these cases, since 2 nodes is
needed to have ZooKeeper quorum. Just exit if this happens, new
config will be used in this case and ZooKeeper should work fine again.

### DIFF
--- a/zookeeper-server/zookeeper-server-common/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
+++ b/zookeeper-server/zookeeper-server-common/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
@@ -5,6 +5,7 @@ import com.google.inject.Inject;
 import com.yahoo.cloud.config.ZookeeperServerConfig;
 import com.yahoo.component.AbstractComponent;
 import com.yahoo.net.HostName;
+import com.yahoo.protect.Process;
 import com.yahoo.yolean.Exceptions;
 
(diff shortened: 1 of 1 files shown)


zookeeper-server/zookeeper-server-common/src/main/java/com/yahoo/vespa/zookeeper/Reconfigurer.java
```diff
@@ -111,6 +115,9 @@ private void reconfigure(ZookeeperServerConfig newConfig) {
                 now = Instant.now();
             }
         }
+
+        // Reconfiguration failed
+        return false;
```
mpolden, line 120: Instead of returning boolean, how about calling

```
server.shutdown();
Process.logAndDie("Reconfiguration did not complete within timeout " + reconfigTimeout(newServers.size()) + ". Forcing shutdown")
```

here? Could extract a variable for reconfigTimeout as it's used earlier in this method.

hmusum: Yes, that sounds, better, thanks

mpolden (approved):

PR #16482, opened 2021-02-11
Reenable boundary checks when mapping lid to lid for referenced document type.

toregge:
@geirst : please review
@arnej27959 , @baldersheim : FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/imported_attribute_vector_read_guard.h
+++ b/searchlib/src/vespa/searchlib/attribute/imported_attribute_vector_read_guard.h
@@ -41,6 +41,7 @@ protected:
 
 protected:
     uint32_t getTargetLid(uint32_t lid) const {
+        // Check range to avoid reading memory beyond end of mapping array
         return lid < _targetLids.size() ? _targetLids[lid] : 0u;
     }
 
--- a/searchlib/src/vespa/searchlib/attribute/imported_attribute_vector_read_guard.h
+++ b/searchlib/src/vespa/searchlib/attribute/imported_attribute_vector_read_guard.h
@@ -41,7 +41,7 @@ protected:
 
 protected:
     uint32_t getTargetLid(uint32_t lid) const {
-        return _targetLids[lid];
+        return lid < _targetLids.size() ? _targetLids[lid] : 0u;
     }
 
 public:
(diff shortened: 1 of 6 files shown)


searchlib/src/vespa/searchlib/attribute/imported_attribute_vector_read_guard.h
```diff
@@ -41,7 +41,7 @@ class ImportedAttributeVectorReadGuard : public IAttributeVector,
 
 protected:
     uint32_t getTargetLid(uint32_t lid) const {
-        return _targetLids[lid];
+        return lid < _targetLids.size() ? _targetLids[lid] : 0u;
```
geirst, line 44: Consider making a unit test for this. E.g in imported_attribute_vector_test.cpp. Would also be nice with a comment on why this is needed.

searchlib/src/vespa/searchlib/attribute/imported_search_context.h
```diff
@@ -40,7 +40,7 @@ class ImportedSearchContext : public ISearchContext {
     SearchContextParams                             _params;
 
     uint32_t getTargetLid(uint32_t lid) const {
-        return _targetLids[lid];
+        return lid < _targetLids.size() ? _targetLids[lid] : 0u;
```
geirst, line 43: Consider making a unit test for this. E.g in imported_search_context_test.cpp. Would also be nice with a comment on why this is needed.

searchlib/src/vespa/searchlib/attribute/reference_mappings.h
```diff
@@ -86,7 +86,9 @@ class ReferenceMappings
         std::atomic_thread_fence(std::memory_order_acquire);
         return TargetLids(&_targetLids[0], committedDocIdLimit);
     }
-    uint32_t getTargetLid(uint32_t doc) const { return _targetLids[doc]; }
+    uint32_t getTargetLid(uint32_t doc) const {
+        return doc < _committedDocIdLimit ? _targetLids[doc] : 0u;
```
geirst, line 90: Consider making a unit test for this. E.g in reference_attribute_test.cpp, as ReferenceAttribute uses ReferenceMappings. Would also be nice with a comment on why this is needed.

geirst (approved):
👍 But I think we should make some unit tests to better track the rationale for why this change is needed.

geirst (approved):
👍

PR #16486, opened 2021-02-11
Increase zstd frame size for compressed access log archive

bjorncs:

### DIFF
--- a/jdisc_http_service/src/main/java/com/yahoo/container/logging/LogFileHandler.java
+++ b/jdisc_http_service/src/main/java/com/yahoo/container/logging/LogFileHandler.java
@@ -401,7 +401,7 @@ class LogFileHandler <LOGTYPE> {
         private static void runCompressionZstd(NativeIO nativeIO, Path oldFile) {
             try {
                 Path compressedFile = Paths.get(oldFile.toString() + ".zst");
-                int bufferSize = 0x400000; // 4M
+                int bufferSize = 48*1024*1024;
                 try (FileOutputStream fileOut = AtomicFileOutputStream.create(compressedFile);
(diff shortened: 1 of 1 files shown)


freva (approved):

jdisc_http_service/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -401,7 +401,7 @@ private static void runCompressionNone(NativeIO nativeIO, Path oldFile) {
         private static void runCompressionZstd(NativeIO nativeIO, Path oldFile) {
             try {
                 Path compressedFile = Paths.get(oldFile.toString() + ".zst");
-                int bufferSize = 0x400000; // 4M
+                int bufferSize = 48*1024*1024;
```
baldersheim, line 404: I think this is on the high side. This means that you will have a 48M buffer on even the smallest container.
Is this really necessary ? What is the gain. 4M sound like plenty.

bjorncs: The motivation was to improve compression ratio compared gzip. The zstandard frames are independent, so bigger frame should reduce the overall file size (for large access logs).

bjorncs: You could change it to config and use smaller value for non-application container cluster.

baldersheim: Have you measured the gain. When I tested earlier on there was not very much gain after 1M. Then it is better to increase compression level.

PR #16487, opened 2021-02-11
Avoid starving global merges with default space bucket deletions

vekterli:
@geirst please review

Introduces a new top-most internal maintenance priority level and
changes bucket activation to use this level (still considered the
most important maintenance operation). Have global merge operations
use the old `VERY_HIGH` level. No other operations share this level.


baldersheim (approved):

geirst (approved):
👍

PR #16491, opened 2021-02-11
Ensure that we do not go above 16G as we then might trigger other int…

baldersheim:
…ernal limits that we do not have

full control of.

@toregge or @geirst PR


toregge (approved):

geirst (commented):
👍

PR #16492, opened 2021-02-12
Use small buffers where size matters more than speed.

baldersheim:
@havardpe or @hmusum PR

### DIFF
--- a/jrt/src/com/yahoo/jrt/Supervisor.java
+++ b/jrt/src/com/yahoo/jrt/Supervisor.java
@@ -39,7 +39,7 @@ public class Supervisor {
      * Will optimize buffers size for small memory footprint
      * Use this when you have many connections with very little traffic.
      **/
-    public Supervisor enableSmallBuffers() {
+    public Supervisor useSmallBuffers() {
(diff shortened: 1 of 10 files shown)


hmusum (approved):
Looks good to me, as max buffer sizes are not absolute limits.

jrt/src/com/yahoo/jrt/Supervisor.java
```diff
@@ -33,6 +35,16 @@ public Supervisor(Transport transport) {
         new MandatoryMethods(this);
     }
 
+    /**
+     * Will optimize buffers size for small memory footprint
+     * Use this when you have many connections with very little traffic.
+     **/
+    public Supervisor enableSmallBuffers() {
```
havardpe, line 42: I think 'useSmallBuffers' would be a better name, since it does not enable the use of small buffers as much as it disables the use of large ones.

hmusum (approved):

jrt/src/com/yahoo/jrt/Supervisor.java
baldersheim: Agree, done

PR #16494, opened 2021-02-12
Also deny maintenance when another node is in maintenance

hakonhall:
The cluster controller today already denies setting a node X safely to
maintenance M, if there is another node Y in another group that has wanted
state M.  Which means that if Y is in M but wanted state is not M, X is allowed
to be set in M.  This is an edge case which is rare.


vekterli (approved):
:+1: consider adding an explicit unit test for this case

vekterli (approved):

PR #16498, opened 2021-02-12
Close the underlying Curator to avoid leaking threads and other resou…

baldersheim:
…rces.

@jonmv @bjorncs PR
@hmusum @vekterli FYI


hmusum (approved):

jonmv:
The curator is injected, and should be shut down by its owner, not by this.

jonmv:
Sorry, no.

jonmv:
Just used to using the injected one. Anyway, this should probably be a shared Curator, owned by the maintainer.

PR #16506, opened 2021-02-14
Jonmv/shut down reincexing curator

jonmv:
@baldersheim @hmusum or @bjorncs please review and merge.


jonmv:
Should also wrap the injected `Curator` in a `Provider<Curator>`, which can shut it down. Good catch @baldersheim

hmusum (approved):

PR #16514, opened 2021-02-15
Provide fallback ConnectionLog instance for all container types

bjorncs:
The ConnectionLog component was previously missing for non-application container clusters (e.g metrics-proxy and cluster-controller).
This caused the JettyHttpServer component to always reconstruct on new config.

FYI @tokle

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
@@ -187,6 +187,7 @@ public abstract class ContainerCluster<CONTAINER extends Container>
         addSimpleComponent("com.yahoo.container.jdisc.ContainerThreadFactory");
         addSimpleComponent("com.yahoo.container.handler.VipStatus");
         addSimpleComponent(com.yahoo.container.handler.ClustersStatus.class.getName());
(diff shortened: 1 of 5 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
```diff
@@ -187,6 +187,7 @@ public ContainerCluster(AbstractConfigProducer<?> parent, String configSubId, St
         addSimpleComponent("com.yahoo.container.jdisc.ContainerThreadFactory");
         addSimpleComponent("com.yahoo.container.handler.VipStatus");
         addSimpleComponent(com.yahoo.container.handler.ClustersStatus.class.getName());
+        addSimpleComponent("com.yahoo.container.jdisc.DisabledConnectionLogProvider");
```
baldersheim, line 190: Won't you get 2 conflicting providers for application clusters then ?

baldersheim (approved):
I am not an expert on components. @tokle or @jonmv should verify this.

config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
bjorncs: No, a provider acts as a fallback if not component implementing `ConnectionLog` is configured.

jonmv (commented):
👍

PR #16516, opened 2021-02-15
Prevent node movers from increasing skew

mpolden:
@bratseth


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/NodeSerializer.java
```diff
@@ -82,6 +82,7 @@
     private static final String nodeTypeKey = "type";
     private static final String wantToRetireKey = "wantToRetire";
     private static final String wantToDeprovisionKey = "wantToDeprovision";
+    private static final String preferToRetire = "preferToRetire";
```
bratseth, line 85: Nit: Field should be preferToRetireKey

bratseth (approved):
Great!

PR #16517, opened 2021-02-15
Use singletons for all fallback components

bjorncs:


baldersheim (approved):
Perhaps @hmusum @jonmv @gv or @bratseth Wants to have a look too.

jdisc_http_service/src/main/java/com/yahoo/container/logging/AccessLog.java
```diff
@@ -13,6 +13,8 @@
  */
 public class AccessLog implements RequestLog {
 
+    public static final AccessLog VOID_INSTANCE = new AccessLog(new ComponentRegistry<>());
```
bratseth, line 16: You could call this "none" ...

bratseth:
LGTM

hmusum (approved):

gjoranv (commented):
👍

PR #16518, opened 2021-02-15
Reserve memory for other processes on the content node the proper way…

geirst:
… by including it in hwinfo memory.

This removes the need to scale the write filter limit used for memory.

@baldersheim and @toregge please review

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
@@ -15,12 +15,15 @@ import static java.lang.Long.min;
 public class NodeResourcesTuning implements ProtonConfig.Producer {
 
     final static long MB = 1024 * 1024;
-    final static long GB = MB * 1024;
+    public final static long GB = MB * 1024;
     private final NodeResources resources;
     private final int redundancy;
     private final int searchableCopies;
     private final int threadsPerSearch;
     private final boolean combined;
+
+    // "Reserve" 1GB of memory for other processes running on the content node (config-proxy, cluster-controller, metrics-proxy).
+    public static final double reservedMemoryGb = 1;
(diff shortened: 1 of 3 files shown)


toregge (approved):

config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
```diff
@@ -121,20 +120,17 @@ private void tuneRequestThreads(ProtonConfig.Builder builder) {
...
-        double defaultMemoryLimit = new ProtonConfig.Writefilter(new ProtonConfig.Writefilter.Builder()).memorylimit();
-        double scaledMemoryLimit = ((usableMemoryGb() - reservedMemoryGb) * defaultMemoryLimit) / usableMemoryGb();
-        builder.memorylimit(scaledMemoryLimit);
-    }
-
     /** Returns the memory we can expect will be available for the content node processes */
     private double usableMemoryGb() {
-        if ( ! combined ) return resources.memoryGb();
+        // "Reserve" 1GB of memory for other processes running on the content node (config-proxy, cluster-controller, metrics-proxy).
+        final double reservedMemoryGb = 1;
```
baldersheim, line 126: Instead of spread '1' around I would make it a constant that could be used in the tests. Or a static method that could be used in the tests compute real memory.

baldersheim (changes requested):
I think a named constant is necessary to understand the reasoning behind  +1 / -1

config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
geirst: Yes, I agree. Fixed. @baldersheim PTAL.

baldersheim (approved):

PR #16525, opened 2021-02-15
Use a static map of locks (by cluster and index) to sync CCs in same …

jonmv:
…JDisc

@hmusum or @hakonhall please review.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
@@ -40,8 +40,11 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


hakonhall (approved):
Please add logging both at lock and unlock, if there is contention.  It could include e.g. the queue depth.

I also think this lock should be fair.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
```diff
@@ -52,6 +55,9 @@
 
     private static final Logger log = Logger.getLogger(FleetController.class.getName());
 
+    // Components of different DI graph generation shan't run() simultaneously.
+    private static final Map<String, Lock> controllerLocks = new ConcurrentHashMap<>();
```
baldersheim, line 59: Note that this will only work as long as the FleetController is part of an installed bundle. If it were in a different bundle, as all user provided code is this approach will not work as you will get one map per classloader/bundle.

I do seem to recall that we have another mechanism for this that will also work across bundles. I think that @gv @bjorncs @hmusum or @bratseth might know the details.

jonmv: Yes, it only works for installed bundles, and within the same JVM—we potentially have this problem when new CC-0 starts up before old CC-0 retires, as well, but that probably shouldn't happen in practice. 

We have `ClassLocking`, which uses class names as keys, and this will work across bundles. It can't be used directly here, however, as the key needs to include cluster name and ID.

jonmv:
Can add logging. I wonder, though, if the code wasn't written with reconfiguration in mind, and perhaps using that instead is a better option.

jonmv:
https://github.com/vespa-engine/vespa/pull/16527 has that alternative.

jonmv:
Prefer alternative.

PR #16526, opened 2021-02-15
Take copy of list before iterating

hmusum:


jonmv (approved):
I guess it could probably fail to copy, as well, although much less likely?

PR #16527, opened 2021-02-16
Avoid recreation of ClusterController when config changes

jonmv:
@bratseth please review. 
@hmusum @hakonhall @baldersheim this is the alternative.


bratseth (approved):
Yes, I think this is the right way to do it.

jonmv:
Yes, this was much simpler than I thought at first—obvious once I realised the original code was written like this 🤦

PR #16537, opened 2021-02-16
Enable group suspension by default [run-systemtest]

hakonhall:
group-suspension has been rolled out to all systems.


vekterli (approved):
:+1:

PR #16541, opened 2021-02-16
Gjoranv/merge chain into processing 

gjoranv:
@aressem, are there any other files that need updating when a module is removed?

FYI: @bjorncs


aressem (approved):
Should be ok with respect to module removal and packaging. No references to this in RPM specs.

bjorncs (approved):

PR #16543, opened 2021-02-16
Jonmv/separate clustercontroller cluster

jonmv:
@hmusum please review. 

I need some unit tests, and an actual check of whether an app is "in sync", using the `/cluster/v2` API on the CC. 
I'm also going to remove support for the `<controllers>` element under `<content>`, which is unused.


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionPreparer.java
```diff
@@ -332,6 +338,8 @@ private void writeStateToZooKeeper(SessionZooKeeperClient zooKeeperClient,
             zooKeeperClient.writeDockerImageRepository(dockerImageRepository);
             zooKeeperClient.writeAthenzDomain(athenzDomain);
             zooKeeperClient.writeQuota(quota);
+            if (dedicatedClusterControllerCluster)
+                zooKeeperClient.writeDedicatedClusterControllerCluster();
```
hmusum, line 342: You need to do the same in `SessionRepository.createSessionFromExisting()`, as per the comment on line 334

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionZooKeeperClient.java
```diff
@@ -52,6 +52,7 @@
     private static final String DOCKER_IMAGE_REPOSITORY_PATH = "dockerImageRepository";
     private static final String ATHENZ_DOMAIN = "athenzDomain";
     private static final String QUOTA_PATH = "quota";
+    private static final String DEDICATED_CLUSTER_CONTROLLER_CLUSTER_PATH = "dedicatedClusterControllerCluster";
```
hmusum, line 55: Consider documenting what this node is for, compared to the one on application path (and perhaps document the other one as well)

hmusum (approved):
Overall LGTM, two comments

jonmv:
Thanks! Nice catch.

hmusum (approved):

jonmv:
Lots of tests, and a little fix in `NodeSpecification`.

jonmv:
4 last commits are new since you reviewed. Conflict, so had to force-push.

config-model/src/test/java/com/yahoo/config/model/provision/ModelProvisioningTest.java
```diff
@@ -1034,6 +1074,50 @@ public void testRequiringMoreNodesThanAreAvailable() {
...
+                "  <content version='1.0' id='foo'>" +
+                "     <redundancy>1</redundancy>" +
+                "     <documents>" +
+                "       <document type='type1' mode='index'/>" +
+                "     </documents>" +
+                "     <nodes count='2' required='true'/>" +
+                "  </content>" +
+                "</services>";
+
+        int numberOfHosts = 4; // needs 2 for foo and 3 for cluster controllers.
```
hmusum, line 1091: 4 is not equal to 5, so what happens here?

hmusum (approved):
Looks good, one minor comment. We need some testing of transitioning from cluster controllers on content nodes to dedicated nodes and vice versa, but might be easier/better to do that in an integration test

config-model/src/test/java/com/yahoo/config/model/provision/ModelProvisioningTest.java
jonmv: That's why the exception is expected.

jonmv:
Yes, we need that. I also think an integration test is the right way, since we're not keeping the state.

config-model/src/test/java/com/yahoo/config/model/provision/ModelProvisioningTest.java
hmusum: D'oh 🤦

PR #16547, opened 2021-02-17
Secret store validation

olaaun:


configserver/src/main/java/com/yahoo/vespa/config/server/http/SecretStoreValidator.java
```diff
@@ -0,0 +1,80 @@
...
+        var hostname = application.getModel().getHosts()
+                .stream()
+                .filter(hostInfo ->
+                    hostInfo.getServices()
+                            .stream()
+                            .filter(service -> CONTAINER.serviceName.equals(service.getServiceType()))
+                            .count() > 0)
+                .map(HostInfo::getHostname)
+                .findFirst().orElseThrow();
+        return URI.create(hostname + AWS_PARAMETER_VALIDATION_HANDLER_POSTFIX);
```
tokle, line 63: is hostname really hostname, or does it include protocol? port? If not include "https://".
If port is included we need to replace the port.

configserver/src/main/java/com/yahoo/vespa/config/server/http/SecretStoreValidator.java
```diff
@@ -0,0 +1,80 @@
...
+    private HttpResponse postRequest(URI uri, Slime slime) {
+        var postRequest = new HttpPost(uri);
+        var data = uncheck(() -> SlimeUtils.toJsonBytes(slime));
+        var entity = new ByteArrayEntity(data);
+        postRequest.setEntity(entity);
+        try (CloseableHttpResponse response = httpClient.execute(postRequest)){
+            return new ProxyResponse(response);
+        } catch (IOException e) {
+            return HttpErrorResponse.internalServerError(
+                    String.format("Failed to post request to %s: %s", uri, Exceptions.toMessageString(e))
```
tokle, line 75: Do we want to return a structured response?

configserver/src/main/java/com/yahoo/vespa/config/server/http/SecretStoreValidator.java
```diff
@@ -26,6 +25,8 @@
  */
 public class SecretStoreValidator {
 
+    // http for easier testing. VespaHttpClient rewrites to https
+    private static final String PROTOCOL = "http://";
```
tokle, line 29: https or will it be automatically changed to https using this client?

tokle (approved):

PR #16554, opened 2021-02-17
Log shutdown complete before shutting down logging

jonmv:
@hmusum


hmusum (approved):

jonmv:
Meh, this is just for the ZK logging filter, it seems.

PR #16555, opened 2021-02-17
Use a list instead of a set, we want to keep order

hmusum:


jonmv (approved):

jonmv:
Perhaps add a `distinct()`?

hmusum:
Good idea, done

PR #16556, opened 2021-02-17
Config for blocking feed to global endpoints

tokle:


config-model/src/main/java/com/yahoo/vespa/model/container/http/BlockFeedGlobalEndpointsFilter.java
```diff
@@ -38,8 +41,8 @@ public void getConfig(RuleBasedFilterConfig.Builder builder) {
                 .collect(Collectors.toSet());
         RuleBasedFilterConfig.Rule.Builder rule = new RuleBasedFilterConfig.Rule.Builder()
                 .hostNames(hostNames)
-                .pathExpressions("/feed")
                 .pathExpressions("/reserved-for-internal-use/feedapi")
```
bjorncs, line 44: Consider reusing constants for `ContainerCluster` etc

bjorncs (approved):

bjorncs (approved):

PR #16559, opened 2021-02-17
Bulk reconfig API is simpler to use, but is it correct?

jonmv:
@hmusum please review.

### DIFF
--- a/zookeeper-server/zookeeper-server-3.6.2/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperAdminImpl.java
+++ b/zookeeper-server/zookeeper-server-3.6.2/src/main/java/com/yahoo/vespa/zookeeper/VespaZooKeeperAdminImpl.java
@@ -17,15 +17,17 @@ public class VespaZooKeeperAdminImpl implements VespaZooKeeperAdmin {
 
     private static final Logger log = java.util.logging.Logger.getLogger(VespaZooKeeperAdminImpl.class.getName());
 
-    @Override
-    public void reconfigure(String connectionSpec, String joiningServers, String leavingServers) throws ReconfigException {
+    private void reconfigure(String connectionSpec, String joiningServers, String leavingServers, String newMembers) throws ReconfigException {
+        if (newMembers != null && (joiningServers != null || leavingServers != null))
+            throw new IllegalArgumentException("Cannot supply both newMembers and joining or leaving servers");
+
         try {
             ZooKeeperAdmin zooKeeperAdmin = new ZooKeeperAdmin(connectionSpec,
                                                                (int) sessionTimeout().toMillis(),
                                                                (event) -> log.log(Level.INFO, event.toString()));
             long fromConfig = -1;
             // Using string parameters because the List variant of reconfigure fails to join empty lists (observed on 3.5.6, fixed in 3.7.0)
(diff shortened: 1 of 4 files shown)


jonmv:
@mpolden do you remember any reason not to use this?

mpolden:
I don't remember the details, but we concluded that incremental mode is likely best fit for how we trigger reconfiguration. Bulk API implies non-incremental mode, which behaves different wrt. concurrent reconfig: https://zookeeper.apache.org/doc/r3.6.2/zookeeperReconfig.html#sc_reconfig_nonincremental

If there's not a good reason to change it (besides cleanup) I suggest keeping it incremental. Reconfig has lots of subtle details and a poor API, none of which are documented well.

jonmv:
So the incremental API is unclear on what to do with servers which change index. The bulk one deals with that for us. It's just easier to use. I couldn't find any reason in the doc to prefer incremental over bulk, in our case.

mpolden:
> incremental API is unclear on what to do with servers which change index

You mean ID? I believe the ID is the only server identifier ZK cares about, so if a server changes it's index/ID from 3 to 5, you would get 3 as leaving and 5 as joining.

jonmv:
Yes, that seems like the correct thing. I don't think the docs were clear :)

hmusum:
As @mpolden says, I think it was the possibility of concurrent reconfig (which _will_ happen the way we reconfigure zookeeper) that made us choose incremental mode.

jonmv:
Hmm ... I read the docs are bulk mode is what we should use:

>Liveness: To better understand the difference between incremental and non-incremental reconfiguration, suppose that client C1 adds server D to the system while a different client C2 adds server E. With the non-incremental mode, each client would first invoke config to find out the current configuration, and then locally create a new list of servers by adding its own suggested server. The new configuration can then be submitted using the non-incremental reconfig command. After both reconfigurations complete, only one of E or D will be added (not both), depending on which client's request arrives second to the leader, overwriting the previous configuration. The other client can repeat the process until its change takes effect. This method guarantees system-wide progress (i.e., for one of the clients), but does not ensure that every client succeeds. To have more control C2 may request to only execute the reconfiguration in case the version of the current configuration hasn't changed, as explained in the section Conditional reconfig. In this way it may avoid blindly overwriting the configuration of C1 if C1's configuration reached the leader first.
>
>With incremental reconfiguration, both changes will take effect as they are simply applied by the leader one after the other to the current configuration, whatever that is (assuming that the second reconfig request reaches the leader after it sends a commit message for the first reconfig request -- currently the leader will refuse to propose a reconfiguration if another one is already pending). Since both clients are guaranteed to make progress, this method guarantees stronger liveness. In practice, multiple concurrent reconfigurations are probably rare. Non-incremental reconfiguration is currently the only way to dynamically change the Quorum System. Incremental configuration is currently only allowed with the Majority Quorum System.

So with bulk reconfig, and the way we do it, all clients will ask for the same `newMembers` to be used. One overwriting another shouldn't be an issue. 
With incremental reconfig, each client will compute the same difference, and everyone will ask for it to be applied, meaning the same servers will be added or removed multiple times. I'm not sure this can actually lead to problems the way we currently do things, but that seems to be just luck.

jonmv:
That is, we have idempotent incremental diffs, and perhaps we'll always have, but it feels iffy.

jonmv:
This seems to work well enough now.

PR #16566, opened 2021-02-17
Track sent and acked bytes to ensure a smooth flow of bytes not excee…

baldersheim:
…ding maxpending by more than

than a small byte buffer.

@hmusum and @bjorncs and @jonmv PR

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
@@ -101,7 +101,9 @@ public class LogHandler extends ThreadedHttpRequestHandler {
         public void send(ByteBuffer src) throws IOException {
             try {
                 stallWhilePendingAbove(maxPending);
-            } catch (InterruptedException e) {}
+            } catch (InterruptedException ignored) {
+                throw new IOException("Interrupted waiting for IO");
+            }
             CompletionHandler pendingTracker = new TrackCompletition(src.remaining());
             try {
                 send(src, pendingTracker);
--- a/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
@@ -126,7 +126,7 @@ public class LogHandler extends ThreadedHttpRequestHandler {
                 stallWhilePendingAbove(0);
             }
             catch (InterruptedException e) {
-                throw new RuntimeException("Interrupted waiting for underlying IO to complete", e);
+                throw new IOException("Interrupted waiting for IO");
             }
         }
 
--- a/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
@@ -15,13 +15,14 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.Optional;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 
 public class LogHandler extends ThreadedHttpRequestHandler {
 
     private final LogReader logReader;
+    private static final long MB = 1024*1024;
 
     @Inject
     public LogHandler(Executor executor, LogHandlerConfig config) {
@@ -45,7 +46,7 @@ public class LogHandler extends ThreadedHttpRequestHandler {
             @Override
             public void render(OutputStream output, ContentChannel networkChannel, CompletionHandler handler) {
                 try {
-                    OutputStream blockingOutput = new BlockingFlushContentChannelOutputStream(networkChannel);
+                    OutputStream blockingOutput = new MaxPendingContentChannelOutputStream(networkChannel, 1*MB);
                     logReader.writeLogs(blockingOutput, from, to, hostname);
                     blockingOutput.close();
                 }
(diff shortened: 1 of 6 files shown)


baldersheim:
Improves #16553

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
...
+        public void send(ByteBuffer src) throws IOException {
+            try {
+                stallWhilePendingAbove(maxPending);
+            } catch (InterruptedException e) {}
+            send(src, new TrackCompletition(src.remaining()));
+        }
+
+        private void stallWhilePendingAbove(long pending) throws InterruptedException {
+            while (pendingBytes() > pending) {
+                Thread.sleep(1);
```
jonmv, line 106: Perhaps a `java.util.concurrent.Condition` would be preferable over `Thread.sleep()`?

jonmv: Hmm, not sure, after all.

jonmv:
Yes, this is nice.

hmusum (approved):
LGTM

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
     }
 
 
-    private static class BlockingFlushContentChannelOutputStream extends ContentChannelOutputStream {
-
+    private static class MaxPendingContentChannelOutputStream extends ContentChannelOutputStream {
         private final ContentChannel channel;
```
jonmv, line 65: Now unused.

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
...
+        private final long maxPending;
+        private AtomicLong sent = new AtomicLong(0);
+        private AtomicLong acked = new AtomicLong(0);
```
jonmv, line 68: Should be `final`.

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
...
+            @Override
+            public void failed(Throwable t) {
+                acked.addAndGet(written);
+            }
+        }
+        @Override
+        public void send(ByteBuffer src) throws IOException {
+            try {
+                stallWhilePendingAbove(maxPending);
+            } catch (InterruptedException e) {}
```
jonmv, line 100:
```suggestion
            } catch (InterruptedException e) {
                throw new IOException("Interrupted waiting for IO");
            }
```

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
...
-                          new CompletionHandler() {
-                              @Override public void completed() { latch.countDown(); }
-                              @Override public void failed(Throwable t) { latch.countDown(); }
-                          });
             try {
-                latch.await();
+                stallWhilePendingAbove(0);
             }
             catch (InterruptedException e) {
                 throw new RuntimeException("Interrupted waiting for underlying IO to complete", e);
```
jonmv, line 117:
```suggestion
                throw new IOException("Interrupted waiting for IO");
```

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
baldersheim: If you use signalling you increase the cost as that is around 3x more costly than an atomic add operation. We do not need the immediate wakeup as the inflight amount is significantly larger than a 1ms window. In these cases I find it preferable to use sleep.

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
baldersheim: Fixed

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
baldersheim: Fixed

baldersheim:
Take a good look at the exception handling as the tracking must be correct, if not unhappiness might occur.
This is significantly simpler in c++ with destructors...

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
jonmv: Yup. Thanks for the numbers.

jonmv:
You'd think an API which lets you submit an `onDone() + onError()` callback object wouldn't throw exceptions—glad you spotted this. Looks correct.

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -60,26 +61,57 @@ public void render(OutputStream output, ContentChannel networkChannel, Completio
...
 
-        public BlockingFlushContentChannelOutputStream(ContentChannel endpoint) {
+        public MaxPendingContentChannelOutputStream(ContentChannel endpoint, long maxPending) {
             super(endpoint);
             this.channel = endpoint;
+            this.maxPending = maxPending;
+        }
+
+        private long pendingBytes() {
+            return sent.get() - acked.get();
```
bjorncs, line 76: Consider replacing with a single atomic that is decremented for `completed()`/`failed()`

bjorncs (approved):

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
baldersheim: I have concluded that most of the time it is better to have separate counters both counting up.
1 - There will be less contention.
2 - You preserve more information both for debugging and if you want monitoring.

PR #16575, opened 2021-02-18
Avoid suggesting node moves to spares

mpolden:
Moves to spares never succeed because spares can only be used in cases where
we're replacing a failed node.

Merge together with internal PR due to first commit.

@bratseth


bratseth (approved):
Great!

PR #16576, opened 2021-02-18
Expose node archive URL in nodes response

freva:
This exposes a remote URL (for now a path in an S3 bucket) where the node should upload their logs (for now, in future maybe core dumps, heaps, etc.) to.

Currently this just backed by a feature flag, but in future the config server will create at least 1 bucket (maybe more if there are many applications in zone), and keep track of which applications (and thereby nodes) should sync to where.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
@@ -11,6 +11,9 @@ import com.yahoo.container.jdisc.HttpRequest;
 import com.yahoo.restapi.SlimeJsonResponse;
 import com.yahoo.slime.Cursor;
 import com.yahoo.vespa.applicationmodel.HostName;
+import com.yahoo.vespa.flags.FetchVector;
+import com.yahoo.vespa.flags.FlagSource;
+import com.yahoo.vespa.flags.Flags;
 import com.yahoo.vespa.hosted.provision.Node;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
 import com.yahoo.vespa.hosted.provision.node.Address;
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
```diff
@@ -246,4 +233,29 @@ private String lastElement(String path) {
         return path.substring(lastSlash+1);
     }
 
+    static Optional<String> nodeArchiveUrl(FlagSource flagSource, Node node) {
```
freva, line 236: This is just temporary, eventually the config servers will create bucket and assign applications to buckets, so this will be stored either on the node itself or in `com.yahoo.vespa.hosted.provision.applications.Application`.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
```diff
@@ -246,4 +233,29 @@ private String lastElement(String path) {
...
+        if (bucket.isBlank()) return Optional.empty();
+
+        StringBuilder sb = new StringBuilder(100).append("s3://").append(bucket).append('/');
+        if (node.type() == NodeType.tenant) {
+            if (node.allocation().isEmpty()) return Optional.empty();
+            ApplicationId app = node.allocation().get().owner();
+
+            sb.append(app.tenant().value()).append('/').append(app.application().value()).append('/').append(app.instance().value()).append('/');
+        } else {
+            sb.append("hosted-vespa/");
```
freva, line 250: Just using `hosted-vespa` for all non-tenants since
1. It is a reserved name, so there will be no real tenant with that name
2. Infrastructure nodes can be without allocation for some time and we still might want to have logs/core dumps from it

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
```diff
@@ -192,6 +178,7 @@ private void toSlime(Node node, boolean allFields, Cursor object) {
         node.reports().toSlime(object, "reports");
         node.modelName().ifPresent(modelName -> object.setString("modelName", modelName));
         node.switchHostname().ifPresent(switchHostname -> object.setString("switchHostname", switchHostname));
+        nodeArchiveUri(nodeRepository.flagSource(), node).ifPresent(url -> object.setString("nodeArchiveUri", url));
```
mpolden, line 181: How about just `archiveUri`?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesResponse.java
mpolden: Consider adding a TODO for this.

mpolden (approved):

PR #16580, opened 2021-02-19
Fail safe maintenance if other nodes are not up

hakonhall:
The test_upgrade_of_downed_node system test fails with group-suspension enabled because the Cluster Controller allows the suspension of a node even if another node is down.  Disabled, the Orchestrator blocks the suspension.

It used to be that:
 1. The CC disallowed suspension if any other node (outside the hierarchical group if any) were in maintenance or user wanted state maintenance.
 2. The CC would disallow suspension if the minStorageNodesUp or minRatiofStorageNodesUp thresholds were already violated.

The PR changes these to:
 - If any other storage node or distributor has a generated or user wanted state unlike UP or RETIRED, then disallow safe maintenance.
 - There is no requirement on other nodes being up when setting a node permanently down.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
@@ -153,11 +153,6 @@ public class NodeStateChangeChecker {
                     + currentState);
         }
 
-        Result allNodesAreUpCheck = checkAllNodesAreUp(clusterState);
-        if (!allNodesAreUpCheck.settingWantedStateIsAllowed()) {
-            return allNodesAreUpCheck;
-        }
-
         HostInfo hostInfo = nodeInfo.getHostInfo();
         Integer hostInfoNodeVersion = hostInfo.getClusterStateVersionOrNull();
         int clusterControllerVersion = clusterState.getVersion();
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
@@ -29,20 +29,14 @@ public class NodeStateChangeChecker {
     public static final String BUCKETS_METRIC_NAME = "vds.datastored.bucket_space.buckets_total";
     public static final Map<String, String> BUCKETS_METRIC_DIMENSIONS = Map.of("bucketSpace", "default");
 
-    private final int minStorageNodesUp;
-    private final double minRatioOfStorageNodesUp;
     private final int requiredRedundancy;
     private final HierarchicalGroupVisiting groupVisiting;
     private final ClusterInfo clusterInfo;
 
     public NodeStateChangeChecker(
-            int minStorageNodesUp,
-            double minRatioOfStorageNodesUp,
             int requiredRedundancy,
             HierarchicalGroupVisiting groupVisiting,
             ClusterInfo clusterInfo) {
-        this.minStorageNodesUp = minStorageNodesUp;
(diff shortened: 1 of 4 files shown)


hakonhall:
There may be one argument for keeping (2):  Disallow setting too many nodes of a group to maintenance, if that would cause the thresholds to be crossed.  The hypothesis is that these thresholds determine cluster availability (will need to double-check this), and it would make sense to avoid this.

On the other hand, it would be a bit strange to configure your cluster to go down during upgrade, and any such change may  have to be rolled out carefully.

If we'd like to do this, we need to make 2 changes to the current/old item (2):
 a. the check has to be moved to before the check on the group
 b. the check has to test whether allowing suspension would cross the threshold, not whether the thresholds are already crossed.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -281,48 +270,38 @@ private static boolean groupContainsNode(Group group, Node node) {
...
-        final int nodeCount = clusterState.getNodeCount(NodeType.STORAGE);
-        int upNodesCount = 0;
-        for (int i = 0; i < nodeCount; ++i) {
-            final Node node = new Node(NodeType.STORAGE, i);
-            final State state = clusterState.getNodeState(node).getState();
-            if (state == State.UP || state == State.RETIRED || state == State.INITIALIZING) {
-                upNodesCount++;
+        for (NodeInfo distributorNodeInfo : clusterInfo.getDistributorNodeInfo()) {
+            State wantedState = distributorNodeInfo.getUserWantedState().getState();
+            if (wantedState != State.UP && wantedState != State.RETIRED) {
```
vekterli, line 293: Side note: as far as I can recall, distributors are not set to `retired` as they do not themselves have any data, so that particular check is not likely to ever trigger. No harm in having it around, though.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -281,48 +270,38 @@ private static boolean groupContainsNode(Group group, Node node) {
...
-                return Result.createDisallowed("Another node is already in maintenance:" + nodeInfo.getNodeIndex());
+    private Result checkAllNodesAreUp(ClusterState clusterState) {
+        // This method verifies both storage nodes and distributors are up (or retired).
+        // The complicated part is making a summary error message.
+
+        for (NodeInfo storageNodeInfo : clusterInfo.getStorageNodeInfo()) {
+            State wantedState = storageNodeInfo.getUserWantedState().getState();
+            if (wantedState != State.UP && wantedState != State.RETIRED) {
+                return Result.createDisallowed("Another storage node wants state " +
+                        wantedState.toString().toUpperCase() + ": " + storageNodeInfo.getNodeIndex());
```
vekterli, line 281: Is it possible for this to cause upgrade "starvation" by nodes tagged as `permanently down`? Or are such nodes kept around for such a brief period of time that it does not matter in practice?

vekterli (approved):
:+1:

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
hakonhall: When a node is permanently set down, the application is redeployed immediately to remove the node.  If such a redeployment fails, e.g. a redeployment requires a new node but there is no spare capacity, then the permanently down node will linger and starve upgrades++.  I haven't witnessed this yet.

Come to think of it -- requiring all other nodes are up when setting a node permanently down seems too stringent.  We verify the node is on the latest cluster state version, and manages no buckets, so that should be sufficient?  Let me remove the check on all nodes being up.

vekterli (approved):
:+1:

PR #16585, opened 2021-02-19
Die if waiting for shutdown takes longer than 1 minute, and log when …

hmusum:
…shutting down server


zookeeper-server/zookeeper-server-3.6.2/src/main/java/com/yahoo/vespa/zookeeper/VespaQuorumPeer.java
```diff
@@ -17,15 +20,19 @@
...
         initializeAndRun(new String[]{ path.toFile().getAbsolutePath()});
     }
 
     public void shutdown() {
         if (quorumPeer != null) {
+            log.log(Level.INFO, "Shutting down ZooKeeper server");
             try {
                 quorumPeer.shutdown();
-                quorumPeer.join(); // Wait for shutdown to complete
+                quorumPeer.join(timeToWaitForShutdown.toMillis()); // Wait for shutdown to complete
```
jonmv, line 35: We should also warn when we failed to join. `Thread.isAlive()` or some such should allow us to check that. But can do that in a separate PR. Let's also consider shutting down the correct peer.

jonmv (approved):

PR #16593, opened 2021-02-19
Show cost in public as well

freva:


bratseth (commented):
Ah, I forgot - we divide by 3 to show (approx) AWS equivalent cost of the on-prem deployment (since our prices are 3x AWS). For public we should not do that.

bratseth (approved):

PR #16613, opened 2021-02-21
Lock first

bratseth:


hmusum (approved):

jonmv:
D’oh

jonmv:
For some reason I read this as “throw if app doesn’t exist”, when changing order.

PR #16620, opened 2021-02-21
Shut down container with ZK on quorum loss, instead of hanging for ever

jonmv:
@hmusum please review. 

This will potentially take down all containers in a ZK container cluster simultaneously, though. I'm not sure we really want that. 
Would it be possible to just shut down ZK instead, and restart it? Is it hard to do that reliably?


jonmv:
Let's rather avoid having CCs shuffle indices.

PR #16628, opened 2021-02-22
Wait 10 minutes after last deployment before migrating to dedicated CCC

jonmv:
@hmusum please review.

No point in not carrying out the deployment, when `isQuiescent` has changed to false, as the flag has already been flipped. Don't want to flip it back, either, in case another deployment has taken place in the meantime, and we only want to migrate once, forward.


hmusum (approved):
Great!

PR #16640, opened 2021-02-23
less broken links

kkraune:
We also need some way to skip link checking, example https://github.com/vespa-engine/vespa/blob/master/config-lib/src/test/java/com/yahoo/config/ConfigInstanceBuilderTest.java#L191 - maybe adding a comment like `// skip-link-check`

the link to developer guide will still fail as I have not renamed another doc to this yet


andreer (approved):

andreer:
We could just add a pattern matching those files to scripts/link_checker_config.json?

kkraune:
&gt; pattern matching those files to scripts/link_checker_config.json?

agreed, willdo

PR #16644, opened 2021-02-23
Read exclusiveTo field

tokle:


freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentContextImpl.java
```diff
@@ -46,12 +47,14 @@
     private final String vespaUserOnHost;
     private final double cpuSpeedup;
     private final Set<NodeAgentTask> disabledNodeAgentTasks;
+    private final Optional<ApplicationId> hostExclusiveTo;
```
freva, line 50: Add to `toString()`

freva (approved):

PR #16645, opened 2021-02-23
Shuffle node moves

mpolden:
Since we switched to soft request to retire (`preferToRetire`), node allocation
may choose to ignore the request. This combined with `NodeMover` suggesting node
moves in the same order on every iteration may result the `NodeMover` getting
stuck.

Example case where this can happen:

1. A cluster has _nodeA_ and _nodeB_ on the same switch. `SwitchRebalancer`
suggests moving `nodeA`, because that is the first suggestion and it does not
consider any other constraints.

2. `NodeAllocation` decides that `nodeA` is the best node on its switch because
of some other constraint, e.g. because its skew is lower than `nodeB`. `nodeA`
is thus not retired.

3. `GOTO 1`

@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMover.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMover.java
@@ -12,6 +12,7 @@ import com.yahoo.vespa.hosted.provision.provisioning.HostCapacity;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Random;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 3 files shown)


bratseth (approved):
Longer term I think we should extract the multi-step planning algorithm from SpareCapacityMaintainer and use it for the other node movers. But then we may also need something to stop them from undoing each other too much.

bratseth (approved):

jonmv:
This made a unit test unstable. Probably a non-deterministic iteration order collection somewhere?

jonmv:
Hehe, read the PR after making that comment 😝

PR #16646, opened 2021-02-23
Jonmv/cc cleanup

jonmv:
@mpolden or @hmusum please review.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
@@ -79,7 +79,7 @@ public class ClusterControllerContainer extends Container implements
 
     @Override
     public boolean requiresWantedPort() {
-        return index() == 0;
+        return false;
     }
 
     @Override
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
@@ -436,9 +436,7 @@ public class ContentCluster extends AbstractConfigProducer implements
                             new ClusterControllerContainer(clusterControllers, index, multitenant, deployState);
                     clusterControllerContainer.setHostResource(host);
                     clusterControllerContainer.initService(deployState.getDeployLogger());
-                    clusterControllerContainer.setProp("clustertype", "admin")
-                            .setProp("clustername", clusterControllers.getName())
-                            .setProp("index", String.valueOf(index));
+                    clusterControllerContainer.setProp("clustertype", "admin");
                     containers.add(clusterControllerContainer);
                     ++index;
                 }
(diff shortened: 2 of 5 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
```diff
@@ -436,9 +436,7 @@ private ClusterControllerContainerCluster createClusterControllers(AbstractConfi
                             new ClusterControllerContainer(clusterControllers, index, multitenant, deployState);
                     clusterControllerContainer.setHostResource(host);
                     clusterControllerContainer.initService(deployState.getDeployLogger());
-                    clusterControllerContainer.setProp("clustertype", "admin")
-                            .setProp("clustername", clusterControllers.getName())
-                            .setProp("index", String.valueOf(index));
```
jonmv, line 441: These are overwritten when container is added to cluster a few lines below.

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
```diff
@@ -79,7 +79,7 @@ public int getWantedPort() {
 
     @Override
     public boolean requiresWantedPort() {
-        return index() == 0;
+        return false;
```
jonmv, line 82: Would lead to collision if multiple CCs on same host, from different clusters. This isn't something we support, anyway. If there are overlapping CCs, they're joined. Otherwise they're disjoint, or there's a single cluster.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DedicatedClusterControllerClusterMigrator.java
```diff
@@ -44,6 +45,7 @@ protected DedicatedClusterControllerClusterMigrator(Deployer deployer, Metric me
 
         ZonedDateTime date = ZonedDateTime.ofInstant(clock().instant(), java.time.ZoneId.of("Europe/Oslo"));
         if (   ! nodeRepository().zone().system().isCd()
+            &&   nodeRepository().zone().environment() != Environment.staging
```
hmusum, line 48: Since we wait 10 minutes after last deployment with migration I think most deployments will not migrate anyway, but I guess this is OK.

hmusum (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DedicatedClusterControllerClusterMigrator.java
jonmv: Yes, most deployments won't. We could also lower that, I guess, since it's just for the purpose of a working integration test.

mpolden: Anything special about staging or should this be `!nodeRepository().zone().environment().isTest()`?

baldersheim:
Build broken
SwitchRebalancerTest.rebalance:85 Node is retired in cluster 'c1' expected:<1> but was:<0>

PR #16647, opened 2021-02-23
Revert "Jonmv/cc cleanup"

baldersheim:
Reverts vespa-engine/vespa#16646
@jonmv or @mpolden or @hmusum PR

Build broken
SwitchRebalancerTest.rebalance:85 Node is retired in cluster 'c1' expected:<1> but was:<0>


jonmv:
I can't see how that would be caused by this.

mpolden:
Yes, this is unrelated. Fixed in #16648.

PR #16649, opened 2021-02-23
Flag service for orchestrated restart when ZK myid changes

jonmv:
@hmusum or @mpolden please review.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
@@ -430,8 +430,11 @@ public class ContentCluster extends AbstractConfigProducer implements
             if (clusterControllers.getContainers().isEmpty()) {
                 int index = 0;
                 for (HostResource host : hosts) {
+                    int ccIndex = deployState.getProperties().dedicatedClusterControllerCluster()
+                                  ? host.spec().membership().map(ClusterMembership::index).orElse(index)
+                                  : index;
                     boolean retired = host.spec().membership().map(ClusterMembership::retired).orElse(false);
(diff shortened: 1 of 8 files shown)


jonmv:
Use node repo index for dedicated cluster controllers, instead of highly unstable count-from-zero index.

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
```diff
@@ -430,8 +430,11 @@ private ClusterControllerContainerCluster createClusterControllers(AbstractConfi
             if (clusterControllers.getContainers().isEmpty()) {
                 int index = 0;
                 for (HostResource host : hosts) {
+                    int ccIndex = deployState.getProperties().dedicatedClusterControllerCluster()
+                                  ? host.spec().membership().map(ClusterMembership::index).orElse(index)
+                                  : index;
                     boolean retired = host.spec().membership().map(ClusterMembership::retired).orElse(false);
-                    var clusterControllerContainer = new ClusterControllerContainer(clusterControllers, index, multitenant, deployState, retired);
+                    var clusterControllerContainer = new ClusterControllerContainer(clusterControllers, ccIndex, multitenant, deployState, retired);
```
mpolden, line 437: Not 100% sure on the impact of this. Does it affect CC leader election?

mpolden (approved):

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
jonmv: The CC index is set below, in `addComponentsForThisCluster`. I'm also not certain this will work, so will keep an eye on CD.

PR #16657, opened 2021-02-24
Jonmv/mind the gaps

jonmv:
@bratseth please review.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
@@ -1,0 +1,57 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.provision.provisioning;
+
+import com.yahoo.config.provision.ClusterSpec;
+import com.yahoo.vespa.hosted.provision.NodeList;
+
+import java.util.List;
+
+import static java.util.Comparator.naturalOrder;
+
+/**
+ * Tracks indices of a node cluster, and proposes the index of the next allocation.
+ *
+ * @author jonmv
+ */
+class NodeIndices {
+
+    private final boolean compact;
+    private final List<Integer> used;
+
+    private int last;
+    private int probe;
+
+    NodeIndices(NodeList nodes, boolean compact) {
+        this.compact = compact;
+        this.used = nodes.mapToList(node -> node.allocation().get().membership().index());
+        this.last = compact ? -1 : used.stream().max(naturalOrder()).orElse(-1);
+        this.probe = last;
+    }
+
+    /** Returns the next available index and commits to using it. Throws if a probe is ongoing. */
+    int next() {
+        if (probe != last)
+            throw new IllegalStateException("Must commit ongoing probe before calling 'next'");
+
+        probeNext();
+        commitProbe();
+        return last;
+    }
+
+    /** Returns the next available index, without committing to using it. May be called multiple times. */
+    int probeNext() {
+        while (used.contains(++probe));
+        return probe;
+    }
+
+    /** Commits to using any indices returned by an ongoing probe. */
+    void commitProbe() {
+        last = probe;
+    }
+
+    /** Resets any probed state to what's currently committed. */
+    void resetProbe() {
+        probe = last;
+    }
+
+}
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
@@ -15,15 +15,14 @@ import static java.util.Comparator.naturalOrder;
  */
 class NodeIndices {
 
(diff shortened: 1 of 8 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -60,13 +61,16 @@ public Preparer(NodeRepository nodeRepository, FlagSource flagSource, Optional<H
     private List<Node> prepareNodes(ApplicationId application, ClusterSpec cluster, NodeSpec requestedNodes, int wantedGroups) {
         List<Node> surplusNodes = findNodesInRemovableGroups(application, cluster, wantedGroups);
 
-        MutableInteger highestIndex = new MutableInteger(findHighestIndex(application, cluster));
+        NodeList nodesInCluster = nodeRepository.nodes().list(Node.State.allocatedStates().toArray(new Node.State[0]))
```
bratseth, line 64: NodeList nodesInCluster = nodeRepository.nodes().list().owner(application).cluster(cluster);

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
```diff
@@ -0,0 +1,57 @@
...
+        if (probe != last)
+            throw new IllegalStateException("Must commit ongoing probe before calling 'next'");
+
+        probeNext();
+        commitProbe();
+        return last;
+    }
+
+    /** Returns the next available index, without committing to using it. May be called multiple times. */
+    int probeNext() {
```
bratseth, line 42: Each successive invocation will increase probe by 1 which is not what we want?

bratseth (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
jonmv: `cluster(cluster.id())` in that case. The code I replaced also checked that `type` was the same—I modified this to be in line with what the actual provisioning code does, which is to use `satisfies(...)`.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
jonmv: Yes, it's what we want. Successive probes should yield increasing values—a commit is for all of them at once.

jonmv: I guess the "May be called multiple times." could be more clear, then.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
jonmv: .. but you're saying that's irrelevant, and that id is the only thing that matters (which I was I was expecting, tbh.)

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeIndices.java
bratseth: Aha, ok. Yes fwiw my expectation upon reading that was it would return the same index each time.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
bratseth: Yes (and also that these filters out unallocated).

jonmv:
PTAL.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
jonmv: Thanks, good point.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -61,10 +61,11 @@ public Preparer(NodeRepository nodeRepository, FlagSource flagSource, Optional<H
     private List<Node> prepareNodes(ApplicationId application, ClusterSpec cluster, NodeSpec requestedNodes, int wantedGroups) {
         List<Node> surplusNodes = findNodesInRemovableGroups(application, cluster, wantedGroups);
 
-        NodeList nodesInCluster = nodeRepository.nodes().list(Node.State.allocatedStates().toArray(new Node.State[0]))
-                                                .owner(application)
-                                                .matching(node -> node.allocation().get().membership().cluster().satisfies(cluster));
-        NodeIndices indices = new NodeIndices(nodesInCluster, ! cluster.type().isContent());
+        List<Integer> usedIndices = nodeRepository.nodes().list(Node.State.allocatedStates().toArray(new Node.State[0]))
```
bratseth: You can just remove "Node.State.allocatedStates().toArray(new Node.State[0])"

PR #16658, opened 2021-02-24
Prefer activating ready replicas with more documents

vekterli:
@geirst please review

This goes contrary to the old behavior where we activated the replica
the distributor thinks is "probably" most up to date, but where the
information needed to make such a decision disappears upon bucket
ownership transfer. In turn, this can cause highly sub-optimal replica
activation if the distributors makes the wrong decision.

Remove using the "trusted" concept entirely from the bucket activation
heuristic in favor of a much more deterministic one based on replica
readiness, document count and ideal state order.

Also fixes an incidentally discovered old bug where an edge case in
the cluster state's handling of node counts had the potential of not
clearing pending maintenance operations if the newly unavailable node
had a distribution key equal to `node count - 1`.


geirst (approved):
👍

PR #16663, opened 2021-02-24
Sync up to 100 files when in dirty, including unrotated ones

freva:


node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfoTest.java
```diff
@@ -31,21 +31,40 @@
...
+        assertForLogFile(connectionLogPath1, "s3://vespa-data-bucket/vespa/music/main/h432a/logs/connection/ConnectionLog.default.20210210.zst", ZSTD, false);
+
+        assertForLogFile(connectionLogPath2, "s3://vespa-data-bucket/vespa/music/main/h432a/logs/connection/ConnectionLog.default.20210212.zst", NONE, true);
+        assertForLogFile(connectionLogPath2, "s3://vespa-data-bucket/vespa/music/main/h432a/logs/connection/ConnectionLog.default.20210212.zst", NONE, false);
+    }
+
+    @Test
+    public void vespa_logs() {
+        assertForLogFile(vespaLogPath1, null, null, true);
+        assertForLogFile(vespaLogPath1, "s3://vespa-data-bucket/vespa/music/main/h432a/logs/vespa/vespa.log.zst", ZSTD, false);
```
andreer, line 60: Hm. I find it a little unfortunate that we will upload this file with that name - in theory, if the host h432 remains, a new h432a could be allocated for the same app, then be retired and produce a new vespa.log file - and being outside the time-based naming scheme makes it more difficult to include this file when processing logs.

Perhaps we should instead always ensure it gets rotated first?

andreer (approved):

node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfoTest.java
freva: That would be great, but I don't see how that can be done, other than node-admin rotating it manually. In that case it'd be easier to just upload it with a different name?

andreer: Yes ... will think about it. Feel free to go ahead and merge this PR, we can change that later if we want to.

PR #16664, opened 2021-02-24
Make TLS protocol version configurable in TLS config file

bjorncs:
Only protocols listed in allowlist can be configured.
TLSv1.2 is the only supported version at the moment, but TLSv1.3 will most likely be included in the future.


vekterli (approved):
:+1:

PR #16666, opened 2021-02-24
Bjorncs/localport stale connector

bjorncs:

### DIFF
--- a/jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpServletRequestUtils.java
+++ b/jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpServletRequestUtils.java
@@ -2,7 +2,6 @@
 package com.yahoo.jdisc.http.server.jetty;
 
 import org.eclipse.jetty.server.HttpConnection;
-import org.eclipse.jetty.server.ServerConnector;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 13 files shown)


jdisc_http_service/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpServletRequestUtils.java
```diff
@@ -21,8 +20,19 @@ public static HttpConnection getConnection(HttpServletRequest request) {
      * @return the actual local port of the underlying Jetty connector
      */
     public static int getConnectorLocalPort(HttpServletRequest request) {
-        ServerConnector jettyConnector = (ServerConnector) getConnection(request).getConnector();
-        return jettyConnector.getLocalPort();
+        JDiscServerConnector connector = (JDiscServerConnector) getConnection(request).getConnector();
```
jonmv, line 23: What's this used for, and will it be possible to trick it somehow, or is it just that we prefer the `getLocalPort()` over the `listenPort()` before it's more often right defined, but both these are always the same if defined?

bjorncs: It's the same unless `listenPort()` is configured to `0` (ephemeral port), where the OS determines the listen port (for unit tests).

PR #16676, opened 2021-02-25
Jonmv/dedicated ccs

jonmv:
@bratseth or @hmusum please review.


jonmv:
And/or @freva :)

jonmv:
Logservers don't use much memory. The 4Gb is due to minimum feasible host size. With shared hosts, we can ignore that limitation :)

freva (approved):

hmusum:
👍

PR #16678, opened 2021-02-25
small vector

havardpe:

@arnej27959 please review
@geirst FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/util/small_vector.h
+++ b/vespalib/src/vespa/vespalib/util/small_vector.h
@@ -1,0 +1,155 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "alloc.h"
+#include "traits.h"
+#include <string.h>
+#include <cstdint>
+#include <cassert>
+#include <memory>
+
+namespace vespalib {
+
+namespace small_vector {
+
+template<typename T, typename... Args>
+void create_at(T *ptr, Args &&...args) {
+    // https://en.cppreference.com/w/cpp/memory/construct_at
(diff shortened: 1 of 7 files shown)


vespalib/src/vespa/vespalib/util/small_vector.h
```diff
@@ -0,0 +1,155 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
```
arnej27959, line 2: missing #pragma once

arnej27959 (approved):
looks good to me.
i think we want some way to construct with size, for example
SmallVector(size_t initial_size, const T& value)

havardpe:
@arnej27959 PTAL

vekterli:
Re: the LLVM version of this, should we also consider having a `SmallVectorImpl`/`SmallVectorBase` concept that type-erases away the `N`?

arnej27959 (approved):
👍

arnej27959:
after trying to use it some random places, I think we rather want conversion to and from ConstArrayRef.

havardpe:
> Re: the LLVM version of this, should we also consider having a `SmallVectorImpl`/`SmallVectorBase` concept that type-erases away the `N`?

Maybe in the future. Erasing the N is only really useful if we want to use it for APIs that modify our vector or we want to use it with different N for the same T. It also complicate things like knowing when the elements are externally allocated and reverting a vector back to its initial state.

havardpe:
> after trying to use it some random places, I think we rather want conversion to and from ConstArrayRef.

Lets talk about the convenient uses and I will make a follow-up PR for those

PR #16686, opened 2021-02-26
Periodically reconfigure config server ZooKeeper cluster

mpolden:
@hmusum

FYI @hakonhall


hmusum (approved):
Great!

hmusum (changes requested):
Unit tests in `node-admin` fail with 

java.lang.NoClassDefFoundError: com/yahoo/vespa/zookeeper/Reconfigurer

hmusum (approved):

PR #16687, opened 2021-02-26
Consider reservedTo before switch priority

mpolden:
@bratseth

FYI @yngveaasheim

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
@@ -133,10 +133,6 @@ public abstract class NodeCandidate implements Nodelike, Comparable<NodeCandidat
         if (!this.isSurplus && other.isSurplus) return -1;
         if (!other.isSurplus && this.isSurplus) return 1;
 
-        // Prefer node on exclusive switch
-        int switchPriority = switchPriority(other);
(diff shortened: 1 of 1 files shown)


bratseth:
Hmm, are we sure we want this? They will just be moved afterwards when switch rebalancing is enabled.

yngveaasheim:
At the moment, this is a problem due to VESPA-19616, making it pretty hard to avoid nodes ending up on another host with bad SSD, eg as experienced in VESPA-19879. Also ref VESPA-19882.

bratseth:
Yes, we're missing a feature for dealing with this problem (we's want something like the reverse of reservedTo), and are trying to make do with what we have.

Fine, but could you add a todo on correcting the order again later @mpolden ?

mpolden:
> They will just be moved afterwards when switch rebalancing is enabled.

Yes, but I think that's fine. This also applies to rebalancing, if skew can be reduced by using a non-reserved host Rebalancer will move away from the reserved host eventually (since `reservedTo` is cleared after first deployment).

Additionally if a non-reserved host is always picked over a reserved one, the latter will never become available to other tenants (`reservedTo` is never cleared).

bratseth:
> Additionally if a non-reserved host is always picked over a reserved one, the latter will never become available to other tenants (reservedTo is never cleared).

Good point. Reminds me why I don't like the reservedTo feature.

bratseth (approved):

PR #16689, opened 2021-02-26
Enable group suspension by default, attempt 2 [run-systemtest]

hakonhall:
The first attempt in #16537 failed because a system test failed because it was allowed to
set a content node to maintenance even though another content node was down.
This was a bug fixed in https://github.com/vespa-engine/vespa/pull/16580.


vekterli (approved):
:+1:

PR #16692, opened 2021-02-26
Store archive URI in node-repo ZK

freva:


vespajlib/src/main/java/com/yahoo/lang/CachedSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+import java.time.Instant;
+import java.util.function.Supplier;
+
+/**
+ * Supplier that caches the value for a given duration with ability to invalidate at demand.
+ * Is thread safe.
+ *
+ * @author freva
+ */
+public class CachedSupplier<T> implements Supplier<T> {
```
freva, line 15: Or `MemoizedSupplier`?
I also considered having 2 factory methods similar to Guavas `Suppliers`: `memoize()` and `memoizeWithExpiration()`, but I'm not sure if we ever need a simple memoized supplier that needs refreshing....

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/ArchiveUris.java
```diff
@@ -0,0 +1,86 @@
...
+            archiveUri.map(ArchiveUris::normalizeUri).ifPresentOrElse(uri -> archiveUris.put(tenant, uri),
+                                                                      () -> archiveUris.remove(tenant));
+            db.writeArchiveUris(archiveUris);
+            this.archiveUris.refresh(); // Throw away current cache
+            log.info("Set archive URI for " + tenant + " to " + archiveUri.orElse(null));
+        }
+    }
+
+    static String normalizeUri(String uri) {
+        if (!uri.endsWith("/")) uri = uri + "/";
```
mpolden, line 81: Any reason to not use `java.net.URI` for the URIs? Then you get (some) validation and normalization for free.

freva: `java.net.URI` is just way to permissive, f.ex. all of the invalid URIs from my test are legal `java.net.URI`. With the regex above we can limit it down just to the format we want it.

vespajlib/src/main/java/com/yahoo/lang/CachedSupplier.java
freva: The main thing is ability to invalidate instead of having to create new Supplier, another bonus is that `CachedSupplier` is a `java.util.function.Supplier`, while Guava returns `com.google.common.base.Supplier`.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/ArchiveUris.java
mpolden: Sure, but I still think using `java.net.URI` for something that is an URI is better.

mpolden (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/ArchiveUris.java
freva: That is true, decided against it since it is only used in REST API, and even before that, the URI needs to be extended, so we would have to do a lot of `URI::create` and `URI::toString` to do anything with it.

PR #16695, opened 2021-02-26
handle any vector-like loop/stride pairs

arnej27959:

@havardpe please review
NOTE: I did run the benchmark on my laptop and all the numbers seemed OK.  Will re-check on real HW.


havardpe (approved):
what about the one having 2 sets of strides?

PR #16696, opened 2021-02-26
Revert "Enable group suspension by default, attempt 2" [run-systemtest]

hakonhall:
Reverts vespa-engine/vespa#16689

This revert PR is in case the original breaks the system tests.


vekterli (approved):

hakonhall:
Systemtests passed: no need for revert

PR #16699, opened 2021-02-26
Only compute feed blocked state from available nodes

vekterli:
@geirst please review

Available nodes here mean nodes that are reported as Up/Initializing
and where the wanted state is Up/Retired.


geirst (approved):
👍

PR #16659, opened 2021-02-24
Balder/put commit params in struct

baldersheim:
@toregge or @geirst PR

PR #16705, opened 2021-02-27
Put all params for commit into a single CommitParam struct

baldersheim:
@geirst or @toregge PR
This is the non semantic changes from #16659
The semantic changes will come in a separate PR

### DIFF
--- a/searchcore/src/tests/proton/attribute/attributeflush_test.cpp
+++ b/searchcore/src/tests/proton/attribute/attributeflush_test.cpp
@@ -161,7 +161,7 @@ UpdaterTask::run()
             }
             ia.update(j, i);
         }
-        ia.commit(CommitParam(i-1, i)); // save i as last sync token
+        ia.commit(CommitParam(i-1, i, CommitParam::UpdateStats::SKIP)); // save i as last sync token
         needFlushToken = i;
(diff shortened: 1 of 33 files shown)


searchcore/src/tests/proton/attribute/attributeflush_test.cpp
```diff
@@ -161,11 +161,11 @@ UpdaterTask::run()
             }
             ia.update(j, i);
         }
-        ia.commit(i-1, i); // save i as last sync token
+        ia.commit(CommitParam(i-1, i)); // save i as last sync token
```
toregge, line 164: i will be interpreted as bool here. Consider adding a third argument `false` to ensure i is considered to be a serial number.

toregge (changes requested):

searchcore/src/tests/proton/attribute/attributeflush_test.cpp
baldersheim: Ouch, then we have some very bad signatures here. Must fix.

baldersheim: Fixed, PTAL

toregge (approved):

geirst (commented):
👍

PR #16706, opened 2021-02-28
If a bucket is cancelled we must wait until all inflight operations h…

baldersheim:
…as been drained, before recomputing correct movement status.

@toregge PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/bucketmovejobv2.cpp
+++ b/searchcore/src/vespa/searchcore/proton/server/bucketmovejobv2.cpp
@@ -18,6 +18,7 @@
 #include <vespa/vespalib/util/destructor_callbacks.h>
 #include <vespa/vespalib/util/lambdatask.h>
 #include <thread>
+#include <vespa/vespalib/stllike/hash_set.hpp>
 
 #include <vespa/log/log.h>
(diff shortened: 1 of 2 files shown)


searchcore/src/vespa/searchcore/proton/server/bucketmovejobv2.cpp
```diff
@@ -260,6 +267,13 @@ BucketMoveJobV2::considerBucket(const bucketdb::Guard & guard, BucketId bucket)
     } else {
         _buckets2Move.erase(bucket);
         cancelMovesForBucket(bucket);
+        if (_movers2Complete.find(bucket) != _movers2Complete.end()) {
+            // This will continue until all inflight operations on this bucket is completed
+            _master.execute(makeLambdaTask([this, bucket]() {
```
toregge, line 272: This looks like the wrong place to schedule the extra check. If a bucket executor has a task for the bucket but waits for the bucket lock then you get a busy loop scheduling considerBucket tasks. Similarly if we end up with a stray leftover _moversToComplete entry.

baldersheim: I agree this is not a good solution.

PR #16710, opened 2021-03-01
Control max heap size for metrics proxy with feature flag

hmusum:


jonmv (commented):
I think we should differentiate between different cluster types. I believe this can be used as a dimension in the flag.

hmusum:
Yes, I was about to look into that

hmusum:
Feature flags in ModelContext does not fit too well with cluster type, so this is not ideal...

jonmv (approved):

PR #16716, opened 2021-03-01
Arnej/small vector in rename

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/generic_rename.cpp
+++ b/eval/src/vespa/eval/instruction/generic_rename.cpp
@@ -69,10 +69,13 @@ generic_rename(const Value &a,
                const ValueType &res_type, const ValueBuilderFactory &factory)
 {
     auto cells = a.cells().typify<CT>();
-    std::vector<string_id> output_address(sparse_plan.mapped_dims);
-    std::vector<string_id*> input_address;
+    SmallVector<string_id> output_address;
+    for (size_t i = 0; i < sparse_plan.mapped_dims; ++i) {
+        output_address.emplace_back(string_id{});
+    }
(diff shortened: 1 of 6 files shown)


eval/src/vespa/eval/instruction/generic_rename.cpp
```diff
@@ -69,10 +69,13 @@ generic_rename(const Value &a,
                const ValueType &res_type, const ValueBuilderFactory &factory)
 {
     auto cells = a.cells().typify<CT>();
-    std::vector<string_id> output_address(sparse_plan.mapped_dims);
-    std::vector<string_id*> input_address;
+    SmallVector<string_id> output_address;
```
havardpe, line 72: there is no need for an explicit loop here, since you can create a small vector with n default constructed objects directly

eval/src/vespa/eval/instruction/generic_rename.cpp
```diff
@@ -172,8 +175,12 @@ DenseRenamePlan::DenseRenamePlan(const ValueType &lhs_type,
     const auto out_dims = output_type.nontrivial_indexed_dimensions();
     size_t num_dense_dims = lhs_dims.size();
     assert(num_dense_dims == out_dims.size());
-    std::vector<size_t> lhs_loopcnt(num_dense_dims);
-    std::vector<size_t> lhs_stride(num_dense_dims, 1);
+    SmallVector<size_t> lhs_loopcnt;
+    SmallVector<size_t> lhs_stride;
+    for (size_t i = 0; i < num_dense_dims; ++i) {
```
havardpe, line 180: these can also be done as part of construction

havardpe (approved):

PR #16718, opened 2021-03-01
Secret store path format changes

olaaun:
FYI @ldalves


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -277,6 +277,7 @@ private HttpResponse handlePUT(Path path, HttpRequest request) {
     private HttpResponse handlePOST(Path path, HttpRequest request) {
         if (path.matches("/application/v4/tenant/{tenant}")) return createTenant(path.get("tenant"), request);
         if (path.matches("/application/v4/tenant/{tenant}/key")) return addDeveloperKey(path.get("tenant"), request);
+        if (path.matches("/application/v4/tenant/{tenant}/secret-store/{name}/validate")) return validateSecretStore(path.get("tenant"), path.get("name"), request);
```
oyving, line 280: We could make this a normal GET that just uses the values from the persisted settings.

oyving (changes requested):

oyving (approved):

PR #16722, opened 2021-03-01
Gjoranv/reduce dependencies part2

gjoranv:
This removes the dependency on `provided-dependencies` from a couple of smaller modules. Note that modules that have compile scoped deps should use it to get the correct Import-Packages. 

I have also moved a couple of provided jakson deps into `provided-dependencies` to simplify dependencies in `container-disc` and any other module using container-core.

### DIFF
--- a/config-bundle/README
+++ b/config-bundle/README
@@ -1,1 +1,4 @@
 This package contains the config library osgi bundle.
+Vespa internal container plugins using the config library can depend on this
+module in scope 'provided' to ensure that all necessary packages are imported
+in the container plugin bundle's OSGi manifest.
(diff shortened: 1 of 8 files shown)


bjorncs (approved):

aressem (approved):

bjorncs:
Looks like `jetty-util` is missing on test classpath for some modules.

bjorncs (approved):

gjoranv:
I put `provided-dependencies` back for container-search, as it needs most of the artifacts for unit tests.

aressem (approved):

PR #16731, opened 2021-03-02
Make ConfigServerApplication and ControllerApplication stateful

mpolden:
@hakonhall


service-monitor/src/main/java/com/yahoo/vespa/service/duper/ConfigServerApplication.java
```diff
@@ -1,15 +1,29 @@
...
+ * @author hakonhall
  */
 public class ConfigServerApplication extends ConfigServerLikeApplication {
+
     public ConfigServerApplication() {
         super("zone-config-servers", NodeType.config, ClusterSpec.Type.admin, ServiceType.CONFIG_SERVER);
     }
+
+    @Override
+    public ClusterSpec getClusterSpecWithVersion(Version version) {
```
hakonhall, line 21: Consider moving this method up into ConfigServerLikeApplication, which allows removing the duplicate method in ControllerApplication.

hakonhall (approved):

PR #16741, opened 2021-03-02
Remove notion of node-specific reliability from C++ distribution code

vekterli:

I have never seen this in use anywhere, and can find no code that ever
sets it (be it in the server config or in the cluster state itself).

Bonus is that node candidate trimming can be vastly simplified with this stuff removed.


baldersheim (approved):

baldersheim:
What didn't we plan for ......

vekterli:
As an added snack, from what I can tell the Java code doesn't even try to do any adjustments based on the reliability. So if anyone _did_ use it, the code bases would likely disagree on the ideal state for buckets. Will tear the Java bits out next...

PR #16752, opened 2021-03-02
Provision config servers dynamically

mpolden:

@hakonhall

FYI @freva

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
@@ -75,4 +75,16 @@ public enum NodeType {
     public boolean canRun(NodeType type) {
         return childNodeTypes.contains(type);
     }
+
+    /** Returns the host type of this */
+    public NodeType hostType() {
+        if (isHost()) return this;
+        for (NodeType nodeType : values()) {
+            if (nodeType.childNodeTypes.size() == 1 && nodeType.canRun(this)) {
+                return nodeType;
+            }
+        }
+        throw new IllegalArgumentException("No host of " + this + " exists");
+    }
+
 }
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -103,9 +103,9 @@ public final class Node implements Nodelike {
         }
 
         if (type != NodeType.host && reservedTo.isPresent())
-            throw new IllegalArgumentException("Only hosts can be reserved to a tenant");
+            throw new IllegalArgumentException("Only tenant hosts can be reserved to a tenant");
 
-        if (type != NodeType.host && exclusiveTo.isPresent())
+        if (!type.isHost() && exclusiveTo.isPresent())
             throw new IllegalArgumentException("Only hosts can be exclusive to an application");
     }
 
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InactiveExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InactiveExpirer.java
@@ -1,6 +1,7 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.maintenance;
 
+import com.yahoo.config.provision.NodeType;
 import com.yahoo.jdisc.Metric;
 import com.yahoo.vespa.hosted.provision.Node;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
@@ -10,6 +11,7 @@ import com.yahoo.vespa.hosted.provision.node.Status;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.Map;
 
 /**
  * Maintenance job which moves inactive nodes to dirty or parked after timeout.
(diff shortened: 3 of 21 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -290,17 +290,47 @@ public boolean fulfilledAndNoChanges() {
...
+            return Optional.empty(); // Requests for these node types never have a deficit
+        }
+        return Optional.of(new FlavorCount(requestedNodes.resources().orElseGet(NodeResources::unspecified),
+                                           requestedNodes.fulfilledDeficitCount(accepted())))
+                       .filter(flavorCount -> flavorCount.getCount() > 0);
+    }
+
+    /** Returns the indices to use when provisioning hosts for this */
+    List<Integer> provisionIndices(int count) {
+        if (count == 0) return List.of();
```
freva, line 309: This should never happen?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -290,17 +290,47 @@ public boolean fulfilledAndNoChanges() {
...
+        NodeType hostType = requestedNodes.type().hostType();
+
+        // Tenant hosts have a continuously increasing index
+        if (hostType == NodeType.host) return nodeRepository.database().readProvisionIndices(count);
+
+        // Infrastructure hosts have fixed indices: Their cluster index + 1
+        int offset = 1;
+        Set<Integer> currentIndices = allNodes.nodeType(hostType)
```
freva, line 317: It does look like the initial index maps to hostname index with the +1 relationship, but in zones that have replaced a cfg, their index is >= 3. We have no started filling index gaps, so I guess we will overtime return to all `config` being in [0-2], but isn't it possible if they get reprovisioned in "wrong" order, this breaks?
E.g.:
Say cfg1 has index 5, cfg2 has 6, cfg3 has 2. If we attempt to reprovision cfg2, it will be given index 1, which will overwrite the original cfg1's DNS records?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
```diff
@@ -103,9 +103,9 @@ public Node(String id, IP.Config ipConfig, String hostname, Optional<String> par
         }
 
         if (type != NodeType.host && reservedTo.isPresent())
-            throw new IllegalArgumentException("Only hosts can be reserved to a tenant");
+            throw new IllegalArgumentException("Only tenant hosts can be reserved to a tenant");
 
-        if (type != NodeType.host && exclusiveTo.isPresent())
+        if (!type.isHost() && exclusiveTo.isPresent())
```
freva, line 108: Any reason to allow this for other hosts?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InactiveExpirer.java
```diff
@@ -45,8 +52,12 @@ protected void expire(List<Node> expired) {
 
     @Override
     protected boolean isExpired(Node node) {
-        return    super.isExpired(node)
-               || node.allocation().get().owner().instance().isTester();
+        return super.isExpired(node, timeout(node)) ||
```
freva, line 55: Need to update `Nodes::deactivate` to move `config` nodes to `inactive`?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
mpolden: Yes, infra applications are `exclusive` which causes the `HostProvisioner` implemention to set `exclusiveTo` for those hosts.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
mpolden: True, it can throw instead.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
mpolden: Good point, we have to ensure current hosts have 0, 1, 2 or find another way to decide indices. Will look into it.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InactiveExpirer.java
mpolden: Not sure what you mean? According to the tests I added, this works as intended.

freva: Interesting, never mind then.
I thought they would be moved straight to `dirty` since they are not stateful (only content nodes are by default), but I see now we explicitly set them to stateful in `ConfigServerLikeApplication`

hakonhall (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
```diff
@@ -75,4 +75,16 @@ public NodeType childNodeType() {
     public boolean canRun(NodeType type) {
         return childNodeTypes.contains(type);
     }
+
+    /** Returns the host type of this */
+    public NodeType hostType() {
+        if (isHost()) return this;
+        for (NodeType nodeType : values()) {
+            if (nodeType.childNodeTypes.size() == 1 && nodeType.canRun(this)) {
```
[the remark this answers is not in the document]
mpolden: No, that's intentional.

PR #16762, opened 2021-03-03
Use 256 Mb as default max heap for cluster controllers

hmusum:


jonmv:
The flag is currently 128Mb everywhere.

hmusum:
Yes, I just want to run with that for a while to see if anything strange shows up until we set it to 128 Mb for everyone (non-hosted as well)

jonmv:
Ah, yes, my bad. Had a brainfart and thought the default flag value in code was equivalent to the default we have, but obviously it's not.

PR #16766, opened 2021-03-03
WIP

bjorncs:

### DIFF
--- a/container-core/src/main/java/com/yahoo/restapi/RestApi.java
+++ b/container-core/src/main/java/com/yahoo/restapi/RestApi.java
@@ -1,0 +1,119 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.restapi;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.yahoo.container.jdisc.HttpRequest;
+import com.yahoo.container.jdisc.HttpResponse;
+import com.yahoo.slime.Slime;
+import com.yahoo.slime.SlimeUtils;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalLong;
+
+/**
+ * Rest API routing and response serialization
+ *
+ * @author bjorncs
+ */
+public interface RestApi {
+
+    static Builder builder() { return new RestApiImpl.BuilderImpl(); }
+    static RouteBuilder route(String pathPattern) { return new RestApiImpl.RouteBuilderImpl(pathPattern); }
+
+    HttpResponse handleRequest(HttpRequest request);
+    ObjectMapper jacksonJsonMapper();
+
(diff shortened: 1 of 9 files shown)


container-core/src/main/java/com/yahoo/restapi/RestApi.java
```diff
@@ -0,0 +1,367 @@
...
+    private final List<ExceptionMapperHolder<?>> exceptionMappers;
+    private final List<ResponseMapperHolder<?>> responseMappers;
+    private final ObjectMapper jacksonJsonMapper;
+
+    private RestApi(Builder builder) {
+        ObjectMapper jacksonJsonMapper = builder.jacksonJsonMapper != null ? builder.jacksonJsonMapper : JacksonJsonMapper.instance;
+        this.defaultRoute = builder.defaultRoute != null ? builder.defaultRoute : createDefaultRoute();
+        this.routes = List.copyOf(builder.routes);
+        this.exceptionMappers = combineWithDefaultExceptionMappers(builder.exceptionMappers);
+        this.responseMappers = combineWithDefaultResponseMappers(builder.responseMappers, jacksonJsonMapper);
```
hakonhall, line 41: combineWithDefaultExceptionMappers() and combineWithDefaultResponseMappers() should probably be part of the builder.  It would allow a client of the builder to override/clear the defaults, if we want that later.  It would allow the builder to document the defaults, and that adding mappers will be in addition to the defaults.

container-core/src/main/java/com/yahoo/restapi/RestApi.java
```diff
@@ -0,0 +1,367 @@
...
+            this.pathPattern = builder.pathPattern;
+            this.handlerPerMethod = Map.copyOf(builder.handlerPerMethod);
+            this.defaultHandler = builder.defaultHandler != null ? builder.defaultHandler : createDefaultMethodHandler();
+        }
+
+        private static MethodHandler<?> createDefaultMethodHandler() {
+            return context -> { throw new MethodNotAllowedException(context.request()); };
+        }
+
+        public static class Builder {
```
hakonhall, line 280: This is the second builder in this class, that's a bit confusing.  More generally RestApi contains a lot of inner classes and interfaces, making e.g. navigation in this file a bit confusing.  Consider moving all inner classes and interfaces of RestApi to the top-level, except the builder of RestApi.

container-core/src/main/java/com/yahoo/restapi/RestApi.java
bjorncs: Overriding is already possible as the default mappers acts as fallbacks, and a user can override a mappers behaviour just by specifying a custom mapper for the same type. Clearing all mappers is not - that could be solved by adding `disableBuiltinExceptionMappers`/`disableBuiltinResponseMappers` to builder.

container-core/src/main/java/com/yahoo/restapi/RestApi.java
bjorncs: Good point! All private types could be moved out to top level. All classes/interfaces that are public must either be modelled as inner classes/interfaces or moved to new source files (as a Java source files cannot contain multiple public classes at top level).

tokle (approved):
🥈

PR #16770, opened 2021-03-03
Guard against processing bucket requests with inconsistent internal state version

vekterli:
@geirst please review
@baldersheim FYI

There's a tiny window of time between when the bucket manager observes a new
state version and when the state version actually is visible in the rest of
the process. We must ensure that we don't end up processing requests when
these two differ, or we might erroneously process requests for version X
using a state only valid for version Y < X.


geirst (approved):
👍

PR #16778, opened 2021-03-04
Bratseth/delay initial maintenance

bratseth:
Delay initial maintenance by 30 seconds.
If we're starting an entire system it's not helpful to start maintenance while things are still coming up.

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
+++ b/vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
@@ -43,7 +43,8 @@ public abstract class Maintainer implements Runnable {
         this.ignoreCollision = ignoreCollision;
         Objects.requireNonNull(startedAt);
         Objects.requireNonNull(clusterHostnames);
-        Duration initialDelay = staggeredDelay(interval, startedAt, HostName.getLocalhost(), clusterHostnames);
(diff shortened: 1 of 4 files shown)


vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
```diff
@@ -43,7 +43,8 @@ public Maintainer(String name, Duration interval, Instant startedAt, JobControl
         this.ignoreCollision = ignoreCollision;
         Objects.requireNonNull(startedAt);
         Objects.requireNonNull(clusterHostnames);
-        Duration initialDelay = staggeredDelay(interval, startedAt, HostName.getLocalhost(), clusterHostnames);
+        Duration initialDelay = staggeredDelay(interval, startedAt, HostName.getLocalhost(), clusterHostnames)
+                                .plus(Duration.ofSeconds(30));
```
mpolden, line 47: Consider extracting a descriptive constant for this fixed delay.

mpolden (approved):

vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
bratseth: Added a description

PR #16783, opened 2021-03-04
Use unique_ptr for regex instead of shared_ptr to keep object small.

baldersheim:
There is not need for it to be copied and it is used a few places where size is important.

@toregge PR


searchlib/src/tests/attribute/stringattribute/stringattribute_test.cpp
```diff
@@ -434,21 +383,29 @@ StringAttributeTest::testSingleValue(Attribute & svsa, Config &cfg)
...
+        SingleValueStringPostingAttribute svsa("svspb", cfg);
+        testSingleValue(svsa, cfg);
 
+        SingleValueStringPostingAttribute svsb("svspb", cfg);
+        testDefaultValueOnAddDoc(svsb);
+    }
 }
 
-TEST_APPHOOK(search::StringAttributeTest);
+TEST_MAIN() { TEST_RUN_ALL(); }
```
toregge, line 411: Consider adding a newline at end of file.

toregge (approved):

searchlib/src/tests/attribute/stringattribute/stringattribute_test.cpp
baldersheim: Fixed

PR #16794, opened 2021-03-04
Arnej/send result type arg to make instruction

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/generic_cell_cast.cpp
+++ b/eval/src/vespa/eval/instruction/generic_cell_cast.cpp
@@ -41,12 +41,13 @@ struct SelectGenericCellCastOp {
 } // namespace <unnamed>
 
 InterpretedFunction::Instruction
-GenericCellCast::make_instruction(const ValueType &input_type,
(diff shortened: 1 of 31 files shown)


eval/src/vespa/eval/instruction/generic_cell_cast.cpp
```diff
@@ -41,12 +41,13 @@ struct SelectGenericCellCastOp {
...
 InterpretedFunction::Instruction
-GenericCellCast::make_instruction(const ValueType &input_type,
+GenericCellCast::make_instruction(const ValueType &result_type,
+                                  const ValueType &input_type,
                                   CellType to_cell_type,
                                   Stash &stash)
 {
     CellType from = input_type.cell_type();
-    auto result_type = input_type.cell_cast(to_cell_type);
+    assert(result_type == input_type.cell_cast(to_cell_type));
```
havardpe, line 50: could assert that dimensions are equal for result_type and input_type and that result_type cell type is equal to to_cell_type

havardpe (approved):
we should also clean up the various Param objects (pass in result type)

PR #16798, opened 2021-03-04
Non-functional changes only

bratseth:


metrics-proxy/src/main/java/ai/vespa/metricsproxy/core/MetricsManager.java
```diff
@@ -75,8 +76,8 @@ public String getMetricsByConfigId(String configId) {
     /**
      * Returns the metrics for the given services. The empty list is returned if no services are given.
      *
-     * @param services The services to retrieve metrics for.
-     * @return Metrics for all matching services.
+     * @param services the services to retrieve metrics for
+     * @return tetrics for all matching services
```
gjoranv, line 80: 'metrics'

config-model/src/main/java/com/yahoo/vespa/model/admin/Admin.java
```diff
@@ -105,6 +105,7 @@ public MetricsProxyContainerCluster getMetricsProxyCluster() {
         return metricsProxyCluster;
     }
 
+    /** Use by model amenders */
```
gjoranv, line 108: Should probably be 'Used'?

gjoranv (approved):

PR #16801, opened 2021-03-04
Remove allocation on expiry from dirty in dynamically provisioned zones

freva:
The condition to remove a dynamically provisioned host is that it has no nodes with allocation. Now that `dirty` does not clear allocation (#16661) broken hosts in dynamically provisioned zones are never removed. 

We want to give the host enough time to clean up and upload node logs to S3, but if the host is unable to do so, we should just give up after dirty expiry time, clear the allocation, and let it be deprovisioned.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -290,6 +290,10 @@ public class Nodes {
         return move(hostname, true, Node.State.failed, agent, Optional.of(reason));
     }
 
+    public Node fail(String hostname, boolean keepAllocation, Agent agent, String reason) {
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -290,6 +290,10 @@ public Node fail(String hostname, Agent agent, String reason) {
         return move(hostname, true, Node.State.failed, agent, Optional.of(reason));
     }
 
+    public Node fail(String hostname, boolean keepAllocation, Agent agent, String reason) {
+        return move(hostname, keepAllocation, Node.State.failed, agent, Optional.of(reason));
```
mpolden, line 294: Consider making the other `fail` method call this.

mpolden (approved):

PR #16804, opened 2021-03-04
Keep only one ucs4 buffer, and create the buffer lazy.

baldersheim:
@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/query/query_term_ucs4.cpp
+++ b/searchlib/src/vespa/searchlib/query/query_term_ucs4.cpp
@@ -3,37 +3,30 @@
 #include "query_term_ucs4.h"
 #include <vespa/vespalib/objects/visit.h>
 #include <vespa/vespalib/text/utf8.h>
+#include <mutex>
 
 namespace search {
 
-QueryTermUCS4::UCS4StringT
-QueryTermUCS4::getUCS4Term() const {
-    UCS4StringT ucs4;
-    const string & term = getTermString();
-    ucs4.reserve(term.size() + 1);
-    vespalib::Utf8Reader r(term);
-    while (r.hasMore()) {
-        ucs4_t u = r.getChar();
-        ucs4.push_back(u);
-    }
-    ucs4.push_back(0);
-    return ucs4;
+namespace {
+    std::mutex _globalMutex;
 }
+
+ucs4_t QueryTermUCS4::ZERO_TERM(0);
 
 QueryTermUCS4::QueryTermUCS4() :
     QueryTermSimple(),
+    _termUCS4(),
     _cachedTermLen(0),
-    _termUCS4()
-{
-    _termUCS4.push_back(0);
-}
+    _filled(true)
+{ }
 
 QueryTermUCS4::~QueryTermUCS4() = default;
 
 QueryTermUCS4::QueryTermUCS4(const string & termS, Type type) :
     QueryTermSimple(termS, type),
+    _termUCS4(),
     _cachedTermLen(0),
-    _termUCS4()
+    _filled(false)
 {
     vespalib::Utf8Reader r(termS);
     while (r.hasMore()) {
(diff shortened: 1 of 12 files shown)


searchlib/src/vespa/searchlib/query/query_term_ucs4.h
```diff
@@ -14,29 +14,30 @@ namespace search {
...
         return _cachedTermLen;
     }
 private:
-    size_t                       _cachedTermLen;
-    UCS4StringT                  _termUCS4;
+    void fillUCS4();
+    static ucs4_t ZERO_TERM;
+    std::unique_ptr<ucs4_t[]>  _termUCS4;
+    uint32_t                   _cachedTermLen;
+    bool                       _filled;
```
toregge, line 40: `std::atomic<bool>` ?

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
```diff
@@ -14,29 +14,30 @@ namespace search {
...
+    uint32_t getTermLen() const { return _cachedTermLen; }
+    uint32_t term(const char * & t)     const { t = getTerm(); return _cachedTermLen; }
     void visitMembers(vespalib::ObjectVisitor &visitor) const override;
-    size_t term(const ucs4_t * & t) {
-        if (_termUCS4.empty()) {
-            _termUCS4 = getUCS4Term();
+    uint32_t term(const ucs4_t * & t) {
+        if (!_filled) {
+            fillUCS4();
         }
```
toregge, line 31: Might need a read barrier ?

searchlib/src/vespa/searchlib/query/query_term_ucs4.cpp
```diff
@@ -43,6 +36,25 @@ QueryTermUCS4::QueryTermUCS4(const string & termS, Type type) :
...
+    std::lock_guard guard(_globalMutex);
+    if (_filled) return;
+    _termUCS4.reset(new ucs4_t[_cachedTermLen + 1]);
+    vespalib::Utf8Reader r(getTermString());
+    uint32_t i(0);
+    while (r.hasMore()) {
+        _termUCS4.get()[i++] = r.getChar();
+    }
+    _termUCS4.get()[_cachedTermLen] = 0;
+    _filled = true;
```
toregge, line 55: Might need a write barrier before setting _filled ?

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
```diff
@@ -14,29 +14,30 @@ namespace search {
...
-        t = &_termUCS4[0];
+        t = (_termUCS4) ? _termUCS4.get() : &ZERO_TERM;
```
toregge, line 32: `ZERO_TERM` should never be returned if `fillUCS4` has set` _termUCS4`.

toregge (approved):

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
baldersheim: Done

searchlib/src/vespa/searchlib/query/query_term_ucs4.cpp
baldersheim: I guess you get a barrier when assigning an atomic bool. I think that is default.

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
baldersheim: I assume the std:.atomic<bool> does the trick.

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
baldersheim: _filled is true if it is default constructed. Then it will return ZERO_TERM

baldersheim:
PTAL

searchlib/src/vespa/searchlib/query/query_term_ucs4.h
```diff
@@ -17,16 +15,16 @@ class QueryTermUCS4 : public QueryTermSimple {
...
+    QueryTermUCS4 & operator = (QueryTermUCS4 &&) = delete;
     QueryTermUCS4();
     QueryTermUCS4(const string & term_, Type type);
     ~QueryTermUCS4();
     uint32_t getTermLen() const { return _cachedTermLen; }
     uint32_t term(const char * & t)     const { t = getTerm(); return _cachedTermLen; }
     void visitMembers(vespalib::ObjectVisitor &visitor) const override;
     uint32_t term(const ucs4_t * & t) {
-        if (!_filled) {
+        if (!_filled.load(std::memory_order_relaxed)) {
```
toregge, line 27: `std::memory_order_relaxed` is not sufficient here. The compiler can generate code to read `_termUCS4` before reading `_filled,` and even if it doesn't, speculative execution might cause the CPU to still load `_termUCS4` before `_filled`. It might be better to just change `_termUCS4` to be `std::atomic<ucs4_t*>` or `std::atomic<const ucs4_t*>`. That way, the need for the member variable `_filled` can be eliminated.

baldersheim: Thanks, I removed the default constructors and simplified with a simple std::atomic<ucs4_t *>

PR #16807, opened 2021-03-04
Consider growth rate vs. scaling time

bratseth:
This adds a query rate timeseries per cluster, finds the max rate of significant (> 30%) growth on it and uses that plus the predicted rescaling time to determine the cpu headroom needed to handle similar growth in the future.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterTimeseries.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterTimeseries.java
@@ -2,11 +2,9 @@
 package com.yahoo.vespa.hosted.provision.autoscale;
 
 import com.yahoo.vespa.hosted.provision.NodeList;
-import com.yahoo.vespa.hosted.provision.NodeRepository;
 import com.yahoo.vespa.hosted.provision.applications.Cluster;
 
 import java.time.Duration;
-import java.time.Instant;
 import java.util.List;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -18,12 +16,14 @@ import java.util.stream.Collectors;
  */
 public class ClusterTimeseries {
 
+    private final Cluster cluster;
     private final NodeList clusterNodes;
 
     /** The measurements for all nodes in this snapshot */
-    private final List<NodeTimeseries> allTimeseries;
+    private final List<NodeTimeseries> timeseries;
 
     public ClusterTimeseries(Duration period, Cluster cluster, NodeList clusterNodes, MetricsDb db) {
+        this.cluster = cluster;
         this.clusterNodes = clusterNodes;
         var timeseries = db.getNodeTimeseries(period, clusterNodes);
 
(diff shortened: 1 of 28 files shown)


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTest.java
```diff
@@ -491,6 +491,44 @@ public void test_autoscaling_considers_read_share() {
...
+        AutoscalingTester tester = new AutoscalingTester(resources.withVcpu(resources.vcpu() * 2));
+
+        ApplicationId application1 = tester.applicationId("application1");
+        ClusterSpec cluster1 = tester.clusterSpec(ClusterSpec.Type.container, "cluster1");
+
+        tester.deploy(application1, cluster1, 5, 1, resources);
+        tester.addCpuMeasurements(0.25f, 1f, 120, application1);
+
+        // (no query rate data)
+        tester.assertResources("Advice to scale up sine we assume we need 2x cpu for growth when no data",
```
freva, line 508:
```suggestion
        tester.assertResources("Advice to scale up since we assume we need 2x cpu for growth when no data",
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterTimeseries.java
```diff
@@ -1,70 +1,103 @@
...
+    /** The max query growth rate we can predict from this time-series as a fraction of the current traffic per minute */
+    public double maxQueryGrowthRate() {
+        if (snapshots.isEmpty()) return 0.1;
+
+        // Find the period having the highest growth rate, where total growth exceeds 30% increase
+        double maxGrowthRate = 0; // In query rate per minute
+        for (int start = 0; start < snapshots.size(); start++) {
+            if (start > 0) { // Optimization: Skip this point when starting from the previous is better relative to the best rate so far
+                Duration duration = durationBetween(start - 1, start);
+                if ( ! duration.isZero()) {
```
freva, line 58: When is `duration.isZero()`? Shouldn't we check if `duration.toMinutes() == 0` instead? Same on line 67

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
```diff
@@ -28,14 +27,21 @@
  */
 public class MetricsResponse {
 
-    private final Collection<Pair<String, MetricSnapshot>> nodeMetrics;
+    /** Node level metrics */
+    private final Collection<Pair<String, NodeMetricSnapshot>> nodeMetrics;
+
+    /**
+     * Cluster level metrics (collect in
```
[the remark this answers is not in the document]
bratseth: Fixed, thanks

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterTimeseries.java
bratseth: Good point; done. (This could happen if we end up with duplicate reads of the same metric window and nothing else.)

PR #16814, opened 2021-03-05
Avoid duplicating VelocityRuntime for each template

jonmv:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/Template.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/Template.java
@@ -3,7 +3,6 @@ package com.yahoo.vespa.hosted.node.admin.task.util.file;
 
 import org.apache.velocity.VelocityContext;
 import org.apache.velocity.app.Velocity;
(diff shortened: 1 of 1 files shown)


jonmv:
They hold 2Mb of byte buffers each.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/Template.java
```diff
@@ -50,7 +49,7 @@ public FileWriter getFileWriterTo(Path destinationPath) {
 
     public String render() {
         StringWriter writer = new StringWriter();
-        velocityEngine.evaluate(velocityContext, writer, "Template", template);
+        Velocity.evaluate(velocityContext, writer, "Template", template);
```
hakonhall, line 52: This appears to use synchronized around a singleton, but we're not that multi-threaded that it would cause a problem.

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/file/Template.java
jonmv: Nope.

PR #16818, opened 2021-03-05
Always use dedicated cluster controllers when hosted (except in some …

jonmv:
…tests)

@hmusum please review, but do _*NOT*_ merge yet :)


jonmv:
The intention here is to always use dedicated cluster controllers when `hosted`, but to continue writing this to data stores used by other config servers, potentially on lower version during upgrade.

hmusum (approved):

hmusum:
All apps in hosted are using dedicated cluster controllers, merging

PR #16822, opened 2021-03-05
Do not wait for reserved nodes to converge, because they will not

jonmv:
@bratseth


bratseth (approved):
This is genius level ...

jonmv:
You make me blush ...

PR #16835, opened 2021-03-08
Only include infra cfg (proxy) nodes of active parents in the cfg (proxy) application

hakonhall:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -72,6 +72,15 @@ public Preparer(NodeRepository nodeRepository, FlagSource flagSource, Optional<H
             List<Node> accepted = groupPreparer.prepare(application, clusterGroup,
                                                         requestedNodes.fraction(wantedGroups), surplusNodes,
                                                         indices, wantedGroups);
+
+            if (requestedNodes.requireActiveParent()) {
```
mpolden, line 76: Shouldn't this be inverted? If we _not_  require active parents, we can ignore non-active parents.

hakonhall: Yes, hold on

hakonhall: Wait, I don't think your statement makes sense

freva: Something needs to be inverted here, either the name or the implementation + statement

hakonhall: Changes the method name to `rejectNonActiveParents()`

mpolden (approved):

PR #16839, opened 2021-03-08
do not assert when receiving invalid data (from network)

arnej27959:

@havardpe please review


eval/src/vespa/eval/eval/value_codec.cpp
```diff
@@ -117,8 +117,10 @@ ValueType decode_type(nbostream &input, const Format &format) {
             dim_list.emplace_back(name, input.getInt1_4Bytes());
         }
     }
-    if (dim_list.empty()) {
-        assert(cell_type == CellType::DOUBLE);
+    if (dim_list.empty() && (cell_type != CellType::DOUBLE)) {
```
havardpe, line 120: I think doing make_type and checking for is_error after would be more robust here

PR #16842, opened 2021-03-08
Better handling of ZK connectivity issues concurrent with elections [run-systemtest]

vekterli:
@geirst @hakonhall please review. These are predominantly racing edge cases dependent on ZK behavior and therefore tricky to test explicitly.

Adds the following safeguards/improvements:
- Do not clear pending (non-persisted) writes over a `connect()` edge.
  Avoids having the controller eternally wait for a doomed pending
  write to be completed when it has no other events that can trigger
  a new write.
- Trigger `lostDatabaseConnection()` whenever ZK is reconfigured to
  ensure we reload the newest state before trying to compute/publish
  any new states.
- Explicitly drop leadership in `lostDatabaseConnection()` to immediately
  prevent controller from trying any funny leader-related business
  since it no longer can depend on ZK watches triggering.
- When falling back to default state/cluster bundle, ensure that any
  subsequent dependent znode write is predicated on the pre-existing
  znode version being 0, i.e. did not previously exist.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
@@ -43,11 +43,15 @@ public class DatabaseHandler {
         ClusterStateBundle clusterStateBundle;
 
         void clear() {
-            masterVote = null;
+            clearNonClusterStateFields();
             lastSystemStateVersion = null;
+            clusterStateBundle = null;
+        }
+
+        void clearNonClusterStateFields() {
+            masterVote = null;
             wantedStates = null;
             startTimestamps = null;
-            clusterStateBundle = null;
         }
     }
     private class DatabaseListener implements Database.DatabaseListener {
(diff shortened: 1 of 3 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
```diff
@@ -126,37 +130,46 @@ public void reset() {
...
         // Preserve who we want to vote for
         Integer currentVote = (pendingStore.masterVote != null ? pendingStore.masterVote : currentlyStored.masterVote);
         currentlyStored.clear();
-        pendingStore.clear();
+        if (clearPendingStateWrites) {
+            pendingStore.clear();
+        } else {
+            // If we have pending cluster state writes we cannot drop these on the floor, as otherwise the
+            // core CC logic may keep thinking it has persisted writes it really has not. Clearing pending
+            // state writes woudl also prevent the controller from detecting itself being out of sync by
```
geirst, line 150: woudl -> would

geirst (approved):
👍

hakonhall (approved):

PR #16850, opened 2021-03-09
Reapply "Jonmv/new hosted apps use dedicated cluster controllers"

hmusum:
`createApplication()` might be called even if application already exists, do nothing in that case (otherwise we might end up modifying reindexing or dedicated cluster controllers setup).

@jonmv FYI


hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/application/ApplicationCuratorDatabase.java
```diff
@@ -77,6 +77,8 @@ public void createApplication(ApplicationId id) {
         if ( ! id.tenant().equals(tenant))
             throw new IllegalArgumentException("Cannot write application id '" + id + "' for tenant '" + tenant + "'");
         try (Lock lock = lock(id)) {
+            if (curator.exists(applicationPath(id))) return;
```
jonmv, line 80: Gee, thanks ... !

PR #16855, opened 2021-03-09
Make controller emit JSON from secret store validation

oyving:
FYI - @ldalves 

This changes the output from the API to be of the following format:
```json
{
  "target": "scoober.albums in prod.us-central-1",
  "result": {
    "settings": {
      "name": "foo",
      "role": "some-role-name-here",
      "awsId": "1234547899",
      "externalId": "*****",
      "region": "us-east-1"
    },
    "status": "ok"
  }
}
```

With an error:
```json
{
  "target": "scoober.albums in prod.us-central-1",
  "result": {
    "settings": {
      "name": "foo",
      "role": "some-role-name-here",
      "awsId": "1234547899",
      "externalId": "*****",
      "region": "us-east-1"
    },
    "status": "error",
    "errors": [
      {
        "type": "SecretNotFoundException",
        "message": "Could not find secret 'nalle' in any configured secret store"
      }
    ]
  }
}
```

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -33,6 +33,7 @@ import com.yahoo.restapi.SlimeJsonResponse;
 import com.yahoo.security.KeyUtils;
 import com.yahoo.slime.Cursor;
 import com.yahoo.slime.Inspector;
+import com.yahoo.slime.JsonParseException;
 import com.yahoo.slime.Slime;
 import com.yahoo.slime.SlimeUtils;
 import com.yahoo.vespa.hosted.controller.Controller;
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -602,7 +603,17 @@ private HttpResponse validateSecretStore(String tenantName, String name, String
...
+        try {
+            var responseRoot = new Slime();
+            var responseCursor = responseRoot.setObject();
+            responseCursor.setString("target", deployment.get().toString());
+            var responseResultCursor = responseCursor.setObject("result");
+            var responseSlime = SlimeUtils.jsonToSlime(response);
+            SlimeUtils.copyObject(responseSlime.get(), responseResultCursor);
+            return new SlimeJsonResponse(responseRoot);
+        } catch (JsonParseException e) {
+            return new MessageResponse(response);
```
olaaun, line 615: We should propagate the status code. This will always return `200`

oyving: Good catch. Changed to `ErrorResponse.internalServerError()`

olaaun (approved):
We should also have `configServer().validateSecretStore` return something more than String, so we can propagate that status code as well

tokle (approved):

PR #16856, opened 2021-03-09
Immediately exit cluster controller if node index config is changed live

vekterli:
@geirst please review
@hmusum FYI

We do not support live reconfigs of CC index, so swiftly exit if we
detect this, allowing the config sentinel to restart the service.


geirst (approved):
👍

baldersheim (approved):

PR #16857, opened 2021-03-09
avoid explicit switch on cell types

arnej27959:
* use CellTypeUtils::mem_size utility which does exactly
  what we need instead.

@geirst please review
@havardpe FYI


geirst (approved):
👍

PR #16858, opened 2021-03-09
Convert comparator from being a simple comparator to using an explici…

baldersheim:
…t less method.

@toregge or @geirst PR


toregge (approved):

geirst (approved):
👍

PR #16864, opened 2021-03-09
Add simple equal to Comparator interface.

baldersheim:
@toregge or @geirst PR


baldersheim:
@toregge and @geirst
I guess a simple optimisation is that for numeric values equal can compare refs directly.
For strings it can be added as an or too that might prevent fetching value.

toregge (approved):

PR #16865, opened 2021-03-09
cell meta with both compile-time and run-time resolving

havardpe:
wired into generic join and reduce as examples of usage

@arnej27959 please review


arnej27959 (approved):
I think I understand this now 👍

PR #16873, opened 2021-03-10
Validate NodeRepository numbers

bratseth:
@baldersheim optimized toString printed NaN as "0.0" :sob:


freva (approved):

baldersheim:
@bratseth Which toString ?

bratseth:
The one in NodeResources. 

```
        long x10 = Math.round(d * 10);
        sb.append(x10 / 10).append('.').append(x10 % 10);
```

Appends "0.0" id d is NaN. This code is ok here since in this context d shouldn't be NaN, it just confused me mightily because I had forgotten we had this special toString.

PR #16890, opened 2021-03-10
Shrink the size of the NodeState object by using float over double fo…

baldersheim:
…r initProgress and capacity. Also gc unused 'reliability' member.

@vekterli PR


vekterli (approved):
:+1: disk states haven't been in use since ye olde VDS days, so could save an alloc per `NodeState` by tearing that out as well.

In Vespa 8 we can remove the init progress stuff as well since I've rewritten how content node init works so that we transition directly from Down to Up on startup.

There is a _theoretical_ possibility that the C++ (which still uses `double`) and Java ideal state algorithms may diverge if capacity isn't exactly expressible in a `float`, but this is a pretty far fetched scenario.

baldersheim:
The was the answer I was hoping for. I will start tearing out the disk states too.

PR #16892, opened 2021-03-11
Wait maximum 60 seconds for other config servers to create session

hmusum:
Creating a session is cheap and should not take long, so give up if it takes more than 60 seconds


mpolden (approved):
👍

bratseth (approved):

PR #16894, opened 2021-03-11
Arnej/map prep celltypes

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/mixed_map_function.h
+++ b/eval/src/vespa/eval/instruction/mixed_map_function.h
@@ -7,7 +7,7 @@
 namespace vespalib::eval {
 
 /**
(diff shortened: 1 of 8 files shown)


eval/src/vespa/eval/instruction/mixed_map_function.h
```diff
@@ -7,7 +7,7 @@
 namespace vespalib::eval {
 
 /**
- * Tensor function optimizing map operations on tensors.
+ * Tensor function optimizing in-place map operations on tensors.
```
havardpe, line 10: class should ideally be renamed to InplaceMapFunction to match its changed functionality, but I suggest doing that in a separate PR.

havardpe (approved):

PR #16895, opened 2021-03-11
Support equiv

bratseth:


arnej27959 (approved):
looks ok to me.  please avoid collapsing unrelated changes in different modules into same commit.

PR #16897, opened 2021-03-11
Lock less

bratseth:
- Only lock the application if a new autoscaling target is produced
  or cluster information should be updated.
- Only create a maintenance deployment if we should redeploy.


mpolden (approved):
👍

PR #16917, opened 2021-03-12
Remove redundant logging (and log text generation)

jonmv:

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/CuratorCompletionWaiter.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/CuratorCompletionWaiter.java
@@ -57,8 +57,10 @@ class CuratorCompletionWaiter implements Curator.CompletionWaiter {
         List<String> respondents;
         do {
             respondents = curator.framework().getChildren().forPath(barrierPath);
-            log.log(Level.FINE, respondents.size() + "/" + curator.zooKeeperEnsembleCount() + " responded: " +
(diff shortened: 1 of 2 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/CuratorCompletionWaiter.java
```diff
@@ -57,8 +57,10 @@ public void awaitCompletion(Duration timeout) {
         List<String> respondents;
         do {
             respondents = curator.framework().getChildren().forPath(barrierPath);
-            log.log(Level.FINE, respondents.size() + "/" + curator.zooKeeperEnsembleCount() + " responded: " +
-                                respondents + ", all participants: " + curator.zooKeeperEnsembleConnectionSpec());
+            if (log.isLoggable(Level.FINE)) {
```
hmusum, line 60: You can use `log.log(Level.FINE, () -> `instead

hmusum (approved):
One minor comment

zkfacade/src/main/java/com/yahoo/vespa/curator/CuratorCompletionWaiter.java
jonmv: Yes, but in this particular case one of the arguments is non-final, and making it final is much more work :)

hmusum: Yes, right, that's a valid case for sure

PR #16918, opened 2021-03-12
Wrap exception to avoid having to export package [run-systemtest]

hmusum:


hmusum:
clustercontroller-core code uses ZooKeeper directly, so this does not work as it is now

baldersheim (approved):

hmusum:
Need to rewrite cluster controller to use Curator to be able to do this. Requires some work, closing.

PR #16930, opened 2021-03-12
Remove active queries metrics that are never set

bratseth:
I noticed we define these metrics but never set them. I guess they should be safe to remove then.


bratseth:
Ping @yngveaasheim

yngveaasheim (approved):
👍

PR #16946, opened 2021-03-15
use CellMeta::join and reduce to compute result cell type

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/eval/cell_type.h
+++ b/eval/src/vespa/eval/eval/cell_type.h
@@ -124,6 +124,10 @@ struct CellMeta {
     static constexpr CellMeta reduce(CellType input_cell_type, bool output_is_scalar) {
         return normalize(input_cell_type, output_is_scalar).decay();
     }
+    constexpr CellMeta reduce(bool output_is_scalar) {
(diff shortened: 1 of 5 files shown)


eval/src/vespa/eval/eval/cell_type.h
```diff
@@ -124,6 +124,10 @@ struct CellMeta {
     static constexpr CellMeta reduce(CellType input_cell_type, bool output_is_scalar) {
         return normalize(input_cell_type, output_is_scalar).decay();
     }
+    constexpr CellMeta reduce(bool output_is_scalar) {
```
havardpe, line 127: function should be const

eval/src/vespa/eval/eval/cell_type.h
```diff
@@ -124,6 +124,10 @@ struct CellMeta {
...
+        assert(!is_scalar);
```
havardpe, line 128: input can be a scalar

havardpe (commented):
we should consider making CellMeta reduce/peek non-static only (in separate PR)

havardpe (approved):

PR #16948, opened 2021-03-15
Require 3 active config servers/controllers when allowing removal

hakonhall:
Makes the following changes to the completion of retirement of a config server or controller node:
 - require all 3 nodes to be active
 - disallow automatic expiry after 4 days

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
@@ -106,7 +106,7 @@ public class RetiredExpirer extends NodeRepositoryMaintainer {
                 // and PERMANENTLY_DOWN for cfg2 is cleaned up.
                 //
                 // If the RetiredExpirer on cfg3 now runs before its InfrastructureProvisioner, then
-                //  a. But the duper model still contains cfg2
+                //  a. The duper model still contains cfg2
                 //  b. The service model still monitors cfg2 for health and it is UP
                 //  c. The Orchestrator has no host status (like PERMANENTLY_DOWN) for cfg2,
                 //     which is equivalent to NO_REMARKS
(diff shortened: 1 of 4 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
```diff
@@ -93,7 +95,32 @@ private boolean canRemove(Node node) {
...
+
+            if (activeNodes.nodeType(node.type()).asSet().size() < NUM_CONFIG_SERVERS) {
+                // Scenario:  All 3 config servers want to retire.
+                //
+                // Say RetiredExpirer runs on cfg1 and gives cfg2 permission to be removed (PERMANENTLY_DOWN in ZK).
+                // The consequent redeployment moves cfg2 to inactive, removing cfg2 from the application,
+                // and PERMANENTLY_DOWN for cfg2 is cleaned up.
+                //
+                // If the RetiredExpirer on cfg3 now runs before its InfrastructureProvisioner, then
+                //  a. But the duper model still contains cfg2
```
mpolden, line 109: s/But //

mpolden (changes requested):
Good, but this needs a unit test.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
hakonhall: Done

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
```diff
@@ -93,7 +95,32 @@ private boolean canRemove(Node node) {
             return false;
         }
 
-        if (node.history().hasEventBefore(History.Event.Type.retired, clock().instant().minus(retiredExpiry))) {
+        if (node.type().isConfigServerLike()) {
+            // Avoid eventual expiry of configserver-like nodes
+
+            if (activeNodes.nodeType(node.type()).asSet().size() < NUM_CONFIG_SERVERS) {
```
mpolden, line 101: Nit: `NodeList` has `size` so need to copy into a set.

mpolden (approved):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
hakonhall: Fixed

mpolden (approved):

PR #16953, opened 2021-03-15
GC unused env variables

hmusum:


baldersheim (approved):
Ancient magic sprinkled around .... :)

PR #16963, opened 2021-03-15
Set wantToDeprovision recursively

mpolden:
When requesting deprovisioning of a host we always want its children to lose
their allocation (via parked), so that the host can be deprovisioned.

@hakonhall


hakonhall (commented):
I think this PR be problematic on-prem:  When NodeDeprovisioner tries to deprovision and delete a host, it will fail because the host has children.  It might be sufficient to delete recursively.  Other options like failing may also need to be made recursive.

mpolden:
`NodeDeprovisioner` won't hit this code path? `Nodes#deprovision` is only called from `RetiringUpgrader`.

Also, I think `NodeDeprovisioner` should behave in the same way as `DynamicProvisioningMaintainer` wrt. to deprovisioning.

hakonhall:
> Nodes#deprovision is only called from RetiringUpgrader

Good point, LGTM

hakonhall (approved):

PR #16965, opened 2021-03-15
Revert "Bjorncs/zk server client port tls only [run-systemtest]"

aressem:
Reverts vespa-engine/vespa#16962

Seems to have broken the tests/cloudconfig/configserver/configserver.rb system test.


aressem:
Seems not to be the cause. Closing.

PR #16966, opened 2021-03-15
Consume content metrics

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
```diff
@@ -56,104 +57,120 @@ private MetricsResponse(Slime response, NodeList applicationNodes, NodeRepositor
...
+                               "content.proton.documentdb.matching.queries.rate");
+            }
         },
         writeRate { // writes per second
-            public String metricResponseName() { return "feed.http-requests.rate"; }
-            double convertValue(double metricValue) { return (float) metricValue; }
-            double defaultValue() { return 0.0; }
+            public List<String> metricResponseNames() {
+                return List.of("feed.http-requests.rate",
+                               "vds.filestor.alldisks.allthreads.put.sum.count.rate",
```
freva, line 154: Should these be added to `AutoscalingMetrics.java`?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
```diff
@@ -56,104 +57,120 @@ private MetricsResponse(Slime response, NodeList applicationNodes, NodeRepositor
...
+            }
         },
         inService {
-            public String metricResponseName() { return "in_service"; }
-            double convertValue(double metricValue) { return (float)metricValue; } // Really a boolean
-            double defaultValue() { return 1.0; }
+            public List<String> metricResponseNames() { return List.of("in_service"); }
+            double computeFinal(List<Double> values) {
+                // Really a boolean. Default true. If any is oos -> oos.
+                return values.stream().anyMatch(v -> v == 0) ? 0 : 1;
```
freva, line 142: Should we return 0 if `values` is empty?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
bratseth: Yes, i forgot - thanks!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
bratseth: Not all services send this (yet) so it needs to default to 1.

PR #16970, opened 2021-03-16
Do not use sudo when running vespa-zk* commands

hmusum:
Try to remove sudo, causes issues when using environment variables.
Can't see why it should be needed by default either. Alternatively do as in https://github.com/vespa-engine/vespa/pull/16972


aressem (approved):

aressem:
This is best if it works.

hmusum:
Agreed, let's try this first

PR #16984, opened 2021-03-16
Retire all wanttoretire nodes with TypeNodeSpec

hakonhall:
When wantToRetire is set on a set of infra nodes (any non-tenant application), at most 1 node is allowed to retire.

This causes a problem when wanting to reprovision all config servers:  Say cfg1 and cfghost2 is allowed to retire.  cfghost2 will not be allowed to exit retirement because cfg2 is not parked, which it cannot be because it isn't even retired yet.

Capping the number of retiring nodes seems unnecessary.  For instance RetiredExpirer ensures only one config server is allowed to complete retirement (be removed from application).

This may have an effect on proxy(host) retirement:  If >1 proxies are wantToRetire, and the orchestrator disallows removal, it used to be that first one proxy node would expire after 4 days, then another after 4 days, etc.  Now, all such proxies would expire after the first 4 day period.

### DIFF
--- a/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirerTest.java
+++ b/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirerTest.java
@@ -42,7 +42,6 @@ import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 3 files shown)


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirerTest.java
```diff
@@ -191,25 +190,24 @@ public void config_server_reprovisioning() throws OrchestrationException {
...
-        // Only 1 config server is allowed to retire at any given point in time
-        List<Node> retiredNodes = tester.nodeRepository().nodes().list(() -> {}).stream()
-                .filter(node -> node.allocation().map(allocation -> allocation.membership().retired()).orElse(false))
-                .collect(Collectors.toList());
-        assertEquals(1, retiredNodes.size());
-        Node retiredNode = retiredNodes.get(0);
-        String retiredNodeHostname = retiredNode.hostname();
-
-        // Allow retiredNodeHostname to be removed
+        List<Node> retiredNodes = tester.nodeRepository()
```
freva, line 193: FYI: There is a `NodeList::retired` method that does all this

freva (approved):

mpolden (approved):
👍

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirerTest.java
hakonhall: Changed accordingly

PR #16985, opened 2021-03-16
Remove the admin from quota usage we calculate in controller

oyving:


andreer (approved):

andreer:
In theory someone could now make an admin cluster with 100 expensive nodes, on a trial account. Consider if we want to put in some kind of limit, or including it in quota/billing if cost exceeds some large sum

oyving:
Agreed, but I think that's a different issue. These services were already removed from the validation done in the configuration server. They should become managed services that are kept outside of regular scaling.

PR #17003, opened 2021-03-17
Use ZK node version as cache, instead of reading all bytes and comput…

jonmv:
…ing hash

@mpolden please review. This is already done for deployment job data :)

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
@@ -106,6 +106,9 @@ public class CuratorDb {
     private final Curator curator;
     private final Duration tryLockTimeout;
 
+    // For each application id (path), store the ZK node version and its deserialised data - update when version changes.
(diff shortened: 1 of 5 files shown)


jonmv:
This cache also keeps exactly one copy of each application, so multiple writes and reads to one app won't evict the others.

mpolden (approved):
LGTM.

This cache will grow forever, which is not an issue in practice with our current number of applications, but probably worth a comment.

Not sure if the second commits belongs here? It LGTM though.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -106,6 +106,9 @@
     private final Curator curator;
     private final Duration tryLockTimeout;
 
+    // For each application id (path), store the ZK node version and its deserialised data - update when version changes.
```
jonmv, line 109:
```suggestion
    // For each application id (path), store the ZK node version and its deserialised data - update when version changes.
    // This will grow to keep all applications in memory, but this should be OK
     ```

jonmv:
Second commit was just a minor thing, so added it here >_<

PR #17029, opened 2021-03-18
Inhibit ZooKeeper connections until our local Slobrok mirror is ready.

vekterli:
@hakonhall please review
@baldersheim @geirst FYI

Otherwise, if there are transient Slobrok issues during CC startup and
we end up winning the election, we risk publishing a cluster state where
the entire cluster appears down (since we do not have any knowledge of
Slobrok node mapping state). This will adversely affect availability for
all the obvious reasons.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
@@ -91,6 +91,7 @@ public class DatabaseHandler {
     private long lastZooKeeperConnectionAttempt = 0;
     private static final int minimumWaitBetweenFailedConnectionAttempts = 10000;
     private boolean lostZooKeeperConnectionEvent = false;
+    private boolean connectionEstablishmentIsAllowed = false;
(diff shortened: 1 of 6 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
```diff
@@ -267,6 +268,14 @@ public boolean doNextZooKeeperTask(Context context) throws InterruptedException
         return didWork;
     }
 
+    private boolean connectionEstablishmentIsAllowed() {
```
hakonhall, line 271: This method seems unnecessary.

hakonhall (approved):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/database/DatabaseHandler.java
vekterli: Removed

vekterli:
@hakonhall PTAL

hakonhall (approved):

PR #17036, opened 2021-03-18
Don't write to standard err

bratseth:
@geirst I don't think we should allow this?


geirst (approved):
👍  Yes, zero ready copies makes no sense. @baldersheim FYI

PR #17044, opened 2021-03-18
Only expire tenant hosts from provisioned

mpolden:
@hakonhall or @freva


freva (approved):
So what's going to happen to them? Could be "expire" them by setting `wantToRetire` & `wantToDeprovision` instead in dynamically provisioned zones?

mpolden:
You mean if the host is stuck? Nothing.

I think for now we want them to remain in that state for inspection, but if it becomes a frequent occurrence we should probably deprovision and try again.

PR #17045, opened 2021-03-18
Set & expose archive access role in /application/v4

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tenant/CloudTenant.java
```diff
@@ -36,6 +39,8 @@ public CloudTenant(TenantName name, Instant createdAt, LastLoginInfo lastLoginIn
         this.info = Objects.requireNonNull(info);
         this.tenantSecretStores = tenantSecretStores;
         this.archiveAccessRole = archiveAccessRole;
+        if (!archiveAccessRole.map(role -> VALID_ARCHIVE_ACCESS_ROLE_PATTERN.matcher(role).matches()).orElse(true))
+            throw new IllegalArgumentException("Invalid archive access role name: " + archiveAccessRole.get());
```
andreer, line 43: Great with additional validation - maybe add the pattern string to the error message so it's easier to find the problem?

andreer (approved):

andreer (approved):
Error messages like this tend to save me a step or two in the debugging process 👍

PR #17047, opened 2021-03-18
Allow dash in ident as it is legal in source names

bratseth:
@arnej27959 please reviev


arnej27959 (approved):
this seems like it might cause expressions with minus to be lexed as ID sometimes, is that a possible problem?

bratseth:
I considered that, and I don't think so (it must be preceded by a letter character).

PR #17056, opened 2021-03-19
update expected test results

kkraune:


kkraune:
@jonmv  maybe the proper fix is changing the output, keeping the test, I leave it to you to decide

PR #17061, opened 2021-03-19
Require primary IP addresses for ready hosts

mpolden:
This disallows a host from moving to ready until it has IP addresses. A scenario
where this can happen:

- A host is dynamically provisioned
- The host sets itself ready before
  `DynamicProvisioningMaintainer#resumeProvisioning` writes the assigned IPs
- The host ends up in reserved without IP addresses, unable to activate (because `active` requires IPs)
- `DynamicProvisioningMaintainer#resumeProvisioning` no longer considers the
  host because it's not in `provisioned`
  
@hakonhall

FYI @freva

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -102,6 +102,11 @@ public final class Node implements Nodelike {
             if (switchHostname.isPresent()) throw new IllegalArgumentException("A child node cannot have switch hostname set");
         }
 
+        if (type.isHost() && state == State.ready) {
+            requireNonEmpty(ipConfig.primary(), "A " + type + " must have at least one primary IP address in state " + state);
+            if (type == NodeType.host) requireNonEmpty(ipConfig.pool().asSet(), "A " + type + " must have a non-empty IP address pool in state " + state);
+        }
+
         if (type != NodeType.host && reservedTo.isPresent())
             throw new IllegalArgumentException("Only tenant hosts can be reserved to a tenant");
 
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -97,7 +97,7 @@ public final class Node implements Nodelike {
             requireNonEmpty(ipConfig.primary(), "Active node " + hostname + " must have at least one valid IP address");
 
         if (parentHostname.isPresent()) {
-            if (!ipConfig.pool().getIpSet().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
+            if (!ipConfig.pool().ipSet().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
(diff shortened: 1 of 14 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
```diff
@@ -97,11 +97,16 @@ public Node(String id, IP.Config ipConfig, String hostname, Optional<String> par
...
         if (parentHostname.isPresent()) {
-            if (!ipConfig.pool().getIpSet().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
+            if (!ipConfig.pool().isEmpty()) throw new IllegalArgumentException("A child node cannot have an IP address pool");
             if (modelName.isPresent()) throw new IllegalArgumentException("A child node cannot have model name set");
             if (switchHostname.isPresent()) throw new IllegalArgumentException("A child node cannot have switch hostname set");
         }
 
+        if (type.isHost() && state == State.ready) {
+            requireNonEmpty(ipConfig.primary(), "A " + type + " must have at least one primary IP address in state " + state);
+            if (type == NodeType.host) requireNonEmpty(ipConfig.pool().asSet(), "A " + type + " must have a non-empty IP address pool in state " + state);
```
hakonhall, line 107: I think all host types require non-empty IP address pool.

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/IP.java
```diff
@@ -325,10 +325,14 @@ public int eventuallyUnusedAddressCount(NodeList nodes) {
             return ipAddresses.protocol;
         }
 
-        public Set<String> getIpSet() {
+        public Set<String> asSet() {
```
hakonhall, line 328: A Pool is a pool of IP addresses and/or hostnames, which is why this method should be named getIpSet().

freva:
Oops, I'm just FYI, didn't mean to approve. Leaving this for @hakonhall

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/IP.java
```diff
@@ -325,10 +325,14 @@ public int eventuallyUnusedAddressCount(NodeList nodes) {
...
             return ipAddresses.asSet();
         }
 
+        public boolean isEmpty() {
```
hakonhall, line 332: The method name is misleading as it indicates there are no addresses in this pool either.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
mpolden: Agree, but many tests violate this. Ok, I'll fix it...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/IP.java
mpolden: Ok, I'll revert. It's very confusing that an _IP address pool_ contains something that's not an IP address, which is also called "Address".

mpolden:
PTAL

hakonhall (approved):

PR #17068, opened 2021-03-19
Start the metrics manager after everything has been setup to avoid th…

baldersheim:
…e expensive metrics altered handling. This causes valgrind runs to be way too costly.

@vekterli PR


vekterli (approved):
:+1:

PR #17071, opened 2021-03-19
Arnej/rewrite mixed simple join function test

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
+++ b/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
@@ -127,6 +127,7 @@ TEST("require that unit join is optimized") {
 
 TEST("require that trivial dimensions do not affect overlap calculation") {
     TEST_DO(verify_optimized("c5d1+b1c5", Primary::RHS, Overlap::FULL, 1));
+    TEST_DO(verify_simple("@c5d1+@b1c5", Primary::RHS, Overlap::FULL, 1, true, true, true));
 }
 
 TEST("require that outer nesting is preferred to inner nesting") {
--- a/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
+++ b/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
@@ -173,17 +173,6 @@ struct LhsRhs {
     }
 };
 
-vespalib::string adjust_param(const vespalib::string &str, bool mut_cells, bool is_rhs) {
-    vespalib::string result = str;
-    if (mut_cells) {
-        result = "@" + result;
-    }
-    if (is_rhs) {
-        result += "$2";
-    }
-    return result;
-}
-
 CellType join_ct(CellType lct, CellType rct) {
     if (lct == CellType::DOUBLE || rct == CellType::DOUBLE) {
         return CellType::DOUBLE;
--- a/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
+++ b/eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
@@ -162,14 +162,6 @@ struct LhsRhs {
     }
 };
 
-CellType join_ct(CellType lct, CellType rct) {
-    if (lct == CellType::DOUBLE || rct == CellType::DOUBLE) {
-        return CellType::DOUBLE;
-    } else {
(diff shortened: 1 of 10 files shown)


eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
```diff
@@ -41,78 +41,95 @@ std::ostream &operator<<(std::ostream &os, Overlap overlap)
...
         }
+        if (require_inplace) {
+            EXPECT_TRUE(fun.inplace());
+        }
+        if (fun.inplace()) {
+            EXPECT_TRUE(fun.primary_is_mutable());
+            size_t idx = (fun.primary() == Primary::LHS) ? 0 : 1;
+            EXPECT_EQUAL(fixture.result_value().cells().data,
+                         fixture.param_value(idx).cells().data);
+        }
```
havardpe, line 75: consider also verifying that we are not inplace if we claim not to be

eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
```diff
@@ -41,78 +41,95 @@ std::ostream &operator<<(std::ostream &os, Overlap overlap)
...
-            if (!fixture.result().cells().empty()) {
-                EXPECT_NOT_EQUAL(fixture.get_param(i), fixture.result());
+struct FunInfo {
+    using LookFor = MixedSimpleJoinFunction;
+    Overlap overlap;
+    size_t factor;
+    std::optional<Primary> primary;
+    bool l_mut;
+    bool r_mut;
+    bool require_inplace;
```
havardpe, line 51: should probably be optional to model potential knowledge of either true or false

eval/src/vespa/eval/instruction/mixed_simple_join_function.cpp
```diff
@@ -176,14 +176,24 @@ MixedSimpleJoinFunction::MixedSimpleJoinFunction(const ValueType &result_type,
...
-    } else {
-        return rhs().result_is_mutable();
-    }
+    return primary_child().result_is_mutable();
+}
+
+bool
+MixedSimpleJoinFunction::inplace() const
+{
+    return primary_is_mutable() &&
```
havardpe, line 195: I think we also might be inplace if the only difference is trivial dimensions

eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
```diff
@@ -144,108 +161,85 @@ struct LhsRhs {
...
+                                 LhsRhs("y5", "x3y5", 5, 15, Overlap::INNER),
+                                 LhsRhs("y5", "y5z3", 5, 15, Overlap::OUTER),
+                                 LhsRhs("x3y5", "y5", 15, 5, Overlap::INNER),
+                                 LhsRhs("y5z3", "y5", 15, 5, Overlap::OUTER)})
+                {
+                    vespalib::string left = adjust_param(params.lhs, left_mut, false);
+                    vespalib::string right = adjust_param(params.rhs, right_mut, true);
+                    vespalib::string expr = fmt(op_pattern, left.c_str(), right.c_str());
+                    TEST_STATE(expr.c_str());
+                    verify_optimized(expr, params.overlap, params.factor,
```
havardpe, line 190: it might not be a good match to use the EvalFixture::verify function for this, since we already map out various combinations for the parameters and would benefit from having knowledge of the cell types up front for each case. We might also want to use a fixed expression with 'a' and 'b' and only modify the input parameters as well...

arnej27959:

havardpe (commented):
maybe we need fewer versions on verify_optimized now that we known things?

eval/src/tests/instruction/mixed_simple_join_function/mixed_simple_join_function_test.cpp
```diff
@@ -144,55 +173,66 @@ struct LhsRhs {
...
+                            } else {
+                                param_repo.add("a", a_spec);
+                            }
+                            if (right_mut) {
+                                param_repo.add_mutable("b", b_spec);
+                            } else {
+                                param_repo.add("b", b_spec);
+                            }
+                            TEST_STATE(expr);
+                            CellType result_ct = join_ct(lct, rct);
```
havardpe, line 211: consider using CellMeta::join here

havardpe (approved):

PR #17073, opened 2021-03-19
archive bucket db

andreer:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * This class decides which tenant goes in what bucket, and creates new buckets when required.
+ *
+ * @author andreer
+ */
+public class CuratorArchiveBucketDb extends AbstractComponent implements ArchiveBucketDb {
```
freva, line 28: This doesnt need to be in `ServiceRegistry` or be an `AbstractComponent`, I think you can just `new` it in `Controller` and have a getter for it there.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
```diff
@@ -0,0 +1,112 @@
...
+        String bucketArn = bucketNameFlag
+                .with(FetchVector.Dimension.ZONE_ID, zoneId.value())
+                .with(FetchVector.Dimension.TENANT_ID, tenant.value())
+                .value();
+
+        if (bucketArn.isBlank()) return Optional.empty();
+
+        if ("auto".equals(bucketArn)) bucketArn = findOrAssignBucket(zoneId, tenant);
+
+        return Optional.of(URI.create(String.format("s3://%s/%s/", bucketArn, tenant.value())));
```
freva, line 59: The URI should have the bucket name, not the bucket ARN.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/archive/ArchiveService.java
```diff
@@ -4,20 +4,18 @@
...
+ * @author andreer
  */
 public interface ArchiveService {
 
-    Optional<URI> archiveUriFor(ZoneId zoneId, TenantName tenant);
-
-    // TODO: Method to configure archive permissions/access for a tenant
-
-    // TODO: Method to revoke permission/access for a tenant
+    ArchiveBucket createArchiveBucketFor(ZoneId zoneId, Set<TenantName> tenantName);
```
freva, line 18: The tenant names doesn't really belong in this API, with the `.withTenant()` method (see other comment) the user should just call `createArchiveBucketFor(zoneId).withTenant(tenantName)`

freva (approved):

PR #17076, opened 2021-03-19
Expire deployment based on time of deployment attempt

mpolden:
@jonmv

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
@@ -1,13 +1,18 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.maintenance;
 
+import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.vespa.hosted.controller.Application;
 import com.yahoo.vespa.hosted.controller.Controller;
 import com.yahoo.vespa.hosted.controller.Instance;
+import com.yahoo.vespa.hosted.controller.api.integration.deployment.JobId;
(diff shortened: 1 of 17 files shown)


jonmv (approved):
Nice. Instead of the last commit, we could also just look at when the last deployment job to the zone started. That would also catch failures which happened before the actual deployment attempt against the config server, and we'd avoid adding an additional field to `Deployment`. 
That is available as something like 
```java
jobController.last(applicationId, jobType).map(run -> run.start());
```

mpolden:
Good idea. I can look into that.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
```diff
@@ -45,10 +50,19 @@ protected boolean maintain() {
...
-                           .map(timeToLive -> deployment.at().plus(timeToLive).isBefore(controller().clock().instant()))
+
+        Optional<Duration> ttl = controller().zoneRegistry().getDeploymentTimeToLive(deployment.zone());
+        if (ttl.isEmpty()) return false;
+
+        Optional<JobId> jobId = JobType.from(controller().system(), deployment.zone())
+                                       .map(type -> new JobId(instance, type));
+        if (jobId.isEmpty()) return false;
+
+        return controller().jobController().last(jobId.get())
```
jonmv, line 63: Could consider a fallback to the deployment's `at` when there is no record of a last job, or when the job-type is unknown. Otherwise LGTM :)

mpolden: Sure, but can that happen?

jonmv: Not now.

mpolden: Ok, if it happens it sounds like a bug somewhere else and I don't think we should work around that here.

PR #17078, opened 2021-03-19
Revert "Don't write to standard err"

aressem:
Reverts vespa-engine/vespa#17036

Broke : https://github.com/vespa-engine/system-test/blob/master/tests/search/hierarchic_distribution/dispatch.rb


geirst (approved):
👍

PR #17082, opened 2021-03-19
More consistent cluster controller ZooKeeper and election handling [run-systemtest]

vekterli:
@hakonhall please review. This contains the changes in #17029 as well as some additional goodies:
* If we're running with ZooKeeper configured, do not allow short-circuiting election phase if only one node is configured. This strictly orders leadership edges to happen after ZK is connected.
* Be consistent in what election state the core `FleetController` logic consults. Instead of (mostly) checking the election handler, check our local state instead. The latter depends on the former, but will only be set _after_ we have fully processed the leadership edge. This strictly orders any state publishes to happen after we've read back from ZK, transitively after ZK connection establishment, transitively after Slobrok reports itself as ready. Note that there's still some auxiliary code sites related to node state change requests that consult the election handler; these should probably be changed to check the core controller state as well.

@geirst FYI

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
@@ -521,6 +521,7 @@ public class FleetController implements NodeStateOrHostInfoChangeHandler, NodeAd
 
         masterElectionHandler.setFleetControllerCount(options.fleetControllerCount);
         masterElectionHandler.setMasterZooKeeperCooldownPeriod(options.masterZooKeeperCooldownPeriod);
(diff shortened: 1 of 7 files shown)


hakonhall (commented):
I'm not sure we want to do the proposed changes to guard ZK connection on Slobrok.  The reported state is set from both getnodestate and slobrok, so we should probably handle both in a similar way.  But getting slobrok and getnodestate is currently guarded on having zk connection (in resyncLocallyCachedState), which is the opposite strategy of that employed here.

The other changes LGTM.

vekterli:
@hakonhall PTAL, I've reverted the deferred ZooKeeper connectivity but left the other changes in places.

hakonhall (approved):
Some isReady() methods are now unused.

vekterli:
I decided to leave them in due to a hunch we might want them again soon 🦄✨ If that's not the case they can be removed very quickly

hakonhall:
👍

PR #17083, opened 2021-03-19
First steps of splitting distributor logic up into disjoint stripes

vekterli:
@geirst let's give this one a whirl on Factory. We'll revert if it blows up.

For now only a single stripe may be used.


geirst (approved):
👍

PR #17085, opened 2021-03-19
Avoid safe mutations in master moratorium and increase first cluster state broadcast deadline [run-systemtest]

hakonhall:
This PR will
 - Designate the master CC as in "moratorium" until it has broadcast its first cluster state
 - Deny the safe setting of node states in moratorium
 - Increase the time to the first cluster state broadcast as described below.

The first cluster state broadcast by a new CC will increase from 5 to 120 seconds, by default.  The time from start to the first broadcast can be increased with config, but not decreased below either of the following two config settings:
 - `get_node_state_request_timeout` 
 - `max_slobrok_disconnect_grace_period`

The time to the first broadcast may happen before 120 seconds if all nodes have reported their getnodestate and Slobrok (they are not down).  In an upgrade scenario, all nodes will likely report within a reasonable short time period, except with hierarchical groups where >1 node is allowed to be suspended.  The increase from 5 to 120 seconds will then appear twice: Once when CC 0 goes down and CC 1 becomes master, and once when CC 0 resumes mastership.


hakonhall:
This may not be sufficient:  the master CC should not make any decisions, like processing safe maintenance requests, until about the same time -- when it has waited long enough on Slobrok and getnodestate data.

Also as an optimization, if the master CC was already part of firstN, then there is typically no reason to wait an additional time since it has monitored Slobrok and getnodestate for a long time already.

vekterli (approved):
:+1:

vekterli (approved):
:+1:

PR #17089, opened 2021-03-22
Get lock after deleting application

hmusum:
This is an attempt at waiting for other config servers to pick up
that the application has been deleted and let them finish before
returning.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -74,7 +74,6 @@ import com.yahoo.vespa.config.server.tenant.EndpointCertificateMetadataStore;
 import com.yahoo.vespa.config.server.tenant.Tenant;
 import com.yahoo.vespa.config.server.tenant.TenantMetaData;
 import com.yahoo.vespa.config.server.tenant.TenantRepository;
-import com.yahoo.config.model.api.TenantSecretStore;
 import com.yahoo.vespa.curator.Curator;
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -533,10 +532,17 @@ public boolean delete(ApplicationId applicationId) {
...
-            return true;
         } finally {
             applicationTransaction.ifPresent(ApplicationTransaction::close);
         }
+
+        // Get the lock to remove the application. Other config servers should have attempted to
+        // get the lock (and lock is fair), so this will essentially let us wait for the others to finish.
+        // If some of them are down they will not be taking the lock, if something else is wrong, getting the
+        // lock times out after 1 minute
+        try (var lock = tenantApplications.lock(applicationId)) {
```
jonmv, line 543: Should we consider a barrier here?

jonmv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
hmusum: We are deleting an application and all nodes related to it in ZooKeeper, so there is nowhere to store it. We could store it somewhere else or add application state as we do for sessions. Makes it even more complicated, though. Can't say this is an elegant solution, either, though...

hmusum:
Nah, this won't work, we might get the lock before the others (seen when trying this out)

PR #17095, opened 2021-03-22
Revert "Revert "Don't write to standard err""

bratseth:
Reverts vespa-engine/vespa#17078

Should be ok after https://github.com/vespa-engine/system-test/pull/1435

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/IndexedHierarchicDistributionValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/IndexedHierarchicDistributionValidator.java
@@ -97,10 +97,10 @@ public class IndexedHierarchicDistributionValidator {
         }
         if (totalReadyCopies % groupCount != 0) {
             throw new IllegalArgumentException(getErrorMsgPrefix(clusterName) + "Expected equal amount of ready copies per group, but " +
-                                totalReadyCopies + " ready copies is specified with " + groupCount + " groups");
+                                               totalReadyCopies + " ready copies is specified with " + groupCount + " groups");
         }
         if (totalReadyCopies == 0) {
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/IndexedHierarchicDistributionValidator.java
```diff
@@ -97,10 +97,10 @@ static public void validateThatReadyCopiesIsCompatibleWithRedundancy(String clus
         }
         if (totalReadyCopies % groupCount != 0) {
             throw new IllegalArgumentException(getErrorMsgPrefix(clusterName) + "Expected equal amount of ready copies per group, but " +
-                                totalReadyCopies + " ready copies is specified with " + groupCount + " groups");
+                                               totalReadyCopies + " ready copies is specified with " + groupCount + " groups");
         }
         if (totalReadyCopies == 0) {
-            System.err.println(getErrorMsgPrefix(clusterName) + "Warning. No ready copies configured. At least one is recommended.");
+            throw new IllegalArgumentException(getErrorMsgPrefix(clusterName) + "Warning. No ready copies configured. At least one is recommended.");
```
geirst, line 103: Perhaps we should re-phrase to "At least one is required"?

geirst (approved):
👍

config-model/src/main/java/com/yahoo/vespa/model/content/IndexedHierarchicDistributionValidator.java
bratseth: done

hmusum:
HierarchicDistributionTest::test_no_ready_copies started failing after this went in. Consider reverting or annotating/fixing test

PR #17099, opened 2021-03-22
Wait until application has been removed before responding

hmusum:
On the config server where the request for deleting an application
happens, wait until applications has been removed. We still don't know
anything about the other servers, but hopefully they have also gotten the
ZooKeeper event and removed the application.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
@@ -74,7 +74,6 @@ import com.yahoo.vespa.config.server.tenant.EndpointCertificateMetadataStore;
 import com.yahoo.vespa.config.server.tenant.Tenant;
 import com.yahoo.vespa.config.server.tenant.TenantMetaData;
 import com.yahoo.vespa.config.server.tenant.TenantRepository;
-import com.yahoo.config.model.api.TenantSecretStore;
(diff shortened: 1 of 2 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -591,6 +591,21 @@ public HttpResponse filedistributionStatus(ApplicationId applicationId, Duration
...
     }
 
+    private void waitForApplicationRemoved(TenantApplications applications, ApplicationId applicationId) {
+        log.log(Level.INFO, "Waiting for " + applicationId + " to be deleted");
+        Duration duration = Duration.ofSeconds(5);
+        Instant end = Instant.now().plus(duration);
+        do {
+            if ( ! (applications.hasApplication(applicationId)))
+                return;
+            log.log(Level.INFO, "Application " + applicationId + " not deleted yet, will retry");
```
jonmv, line 601: A bit spammy, but OK for now :)

jonmv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
hmusum: Right, thanks, I'll add a TODO or just remove it tomorrow.

PR #17120, opened 2021-03-23
Require 3 config server (and controller) hosts

hakonhall:
We already require 3 config server (and controller) nodes, but it is not
sufficient to protect the hosts from being left with only 1 healthy host: Say
the config server host application contains 2 nodes.  An upgrade of host-admin
on one of those nodes is allowed, since only the host is suspended and none of
the 2 nodes are down.  This is fixed by handling config server hosts similar to
config servers: assume 3 nodes.


mpolden (approved):
Great!

PR #17124, opened 2021-03-23
Compute node repo utilization

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepoStats.java
```diff
@@ -0,0 +1,73 @@
...
+                             divide(memory, totalHostResources.memoryGb()),
+                             divide(disk, totalHostResources.diskGb()));
+        Load activeLoad = new Load(divide(cpu, totalActiveResources.vcpu()),
+                                   divide(memory, totalActiveResources.memoryGb()),
+                                   divide(disk, totalActiveResources.diskGb()));
+        return new NodeRepoStats(load, activeLoad);
+    }
+
+    private static double divide(double a, double b) {
+        if (a == 0 && b == 0) return 0;
```
freva, line 69: Why not check just `b == 0`?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepoStats.java
bratseth: If a is != 0 and b is 0 I want it to throw as something is wrong.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepoStats.java
```diff
@@ -0,0 +1,73 @@
...
+        NodeResources totalActiveResources = new NodeResources(0, 0, 0, 0);
+        double cpu = 0, memory = 0, disk = 0;
+        for (var nodeTimeseries : nodeRepository.metricsDb().getNodeTimeseries(Duration.ofHours(1), Set.of())) {
+            Optional<Node> node = allNodes.node(nodeTimeseries.hostname());
+            if (node.isEmpty() || node.get().state() != Node.State.active) continue;
+
+            Optional<NodeMetricSnapshot> snapshot = nodeTimeseries.last();
+            if (snapshot.isEmpty()) continue;
+
+            cpu += snapshot.get().cpu() * node.get().resources().vcpu();
```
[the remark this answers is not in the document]
bratseth: That would be a measure of "waste", not "utilization". That could be useful too. I didn't really consider it, but now I'm going to.
Utilization is also useful though, (more interpretable, and it could be that the 0.4 number is wrong), so I suggest this can be merged.

PR #17125, opened 2021-03-23
Expose system as zone through ZoneRegistry

mpolden:

@freva

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
@@ -31,17 +31,18 @@ import java.util.stream.Stream;
 @SuppressWarnings("unused")
 public class ConfigServerApiHandler extends AuditLoggingRequestHandler {
 
-    private static final ZoneId CONTROLLER_ZONE = ZoneId.from("prod", "controller");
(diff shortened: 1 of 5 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
```diff
@@ -102,7 +103,7 @@ private HttpResponse root(HttpRequest request) {
         ZoneList zoneList = zoneRegistry.zones().reachable();
 
         Cursor zones = root.setArray("zones");
-        Stream.concat(Stream.of(CONTROLLER_ZONE), zoneRegistry.zones().reachable().ids().stream())
+        Stream.concat(Stream.of(controllerZone), zoneRegistry.zones().reachable().ids().stream())
```
freva, line 106: We should consider adding a method to list zones including the controller zone so we avoid having to do this concat everywhere

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/configserver/ConfigServerApiHandler.java
mpolden: Yes, I thought about, but it wasn't immediately obvious how to best add this to `ZoneList`. Will do it in a future PR.

PR #17136, opened 2021-03-23
Arnej/add more cell types 32 

arnej27959:

### DIFF
--- a/eval/src/tests/instruction/join_with_number/join_with_number_function_test.cpp
+++ b/eval/src/tests/instruction/join_with_number/join_with_number_function_test.cpp
@@ -51,6 +51,8 @@ void verify_optimized(const vespalib::string &expr, Primary primary, bool inplac
     const auto stable_types = CellTypeSpace({CellType::FLOAT, CellType::DOUBLE}, 2);
     FunInfo stable_details{primary, inplace};
     TEST_DO(EvalFixture::verify<FunInfo>(expr, {stable_details}, stable_types));
+    const auto unstable_types = CellTypeSpace({CellType::BFLOAT16, CellType::INT8}, 2);
+    TEST_DO(EvalFixture::verify<FunInfo>(expr, {}, unstable_types));
 }
 
 void verify_not_optimized(const vespalib::string &expr) {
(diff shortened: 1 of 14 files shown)


eval/src/tests/instruction/join_with_number/join_with_number_function_test.cpp
```diff
@@ -51,6 +51,8 @@ void verify_optimized(const vespalib::string &expr, Primary primary, bool inplac
     const auto stable_types = CellTypeSpace({CellType::FLOAT, CellType::DOUBLE}, 2);
     FunInfo stable_details{primary, inplace};
     TEST_DO(EvalFixture::verify<FunInfo>(expr, {stable_details}, stable_types));
+    const auto unstable_types = CellTypeSpace({CellType::BFLOAT16, CellType::INT8}, 2);
+    TEST_DO(EvalFixture::verify<FunInfo>(expr, {}, unstable_types));
```
havardpe, line 55: We probably want this optimizer to also apply to unstable cells, since it is one of the usual ways to handle quantized values (multiply with a constant). I guess the issue is with template instantiation due to sharing code between inplace and non-inplace cases...

eval/src/vespa/eval/eval/cell_type.cpp
```diff
@@ -33,7 +33,7 @@ CellTypeUtils::mem_size(CellType cell_type, size_t sz)
 std::vector<CellType>
 CellTypeUtils::list_types()
 {
-    return {CellType::FLOAT, CellType::DOUBLE};
+    return {CellType::INT8, CellType::BFLOAT16, CellType::FLOAT, CellType::DOUBLE};
```
havardpe, line 36: I suggest listing types in their natural order (same as enum). Also consider adding separate utility functions for stable/unstable cell types.

eval/src/vespa/eval/eval/cell_type.h
```diff
@@ -6,15 +6,19 @@
 #include <vector>
 #include <cstdint>
 #include <cassert>
+#include "int8float.h"
+#include <vespa/vespalib/util/bfloat16.h>
 
 namespace vespalib::eval {
 
-enum class CellType : char { FLOAT, DOUBLE };
+enum class CellType : char { FLOAT, DOUBLE, BFLOAT16, INT8 };
```
havardpe, line 14: I am leaning towards switching the ordering of FLOAT and DOUBLE, to make everything have the types in the same order (This would not mean that enum value is tied to the binary format cell type id, even if it happens to be the same).

havardpe (approved):
I like this a lot

havardpe (approved):
consider using list_stable/unstable_types in more tests (specifically those changed in this PR to test with stable vs unstable types)

havardpe (approved):

PR #17122, opened 2021-03-23
Balder/orderting 2 type

baldersheim:
@geirst and @toregge

PR #17137, opened 2021-03-23
Allow hash, btree and dictionary as identifiers.

baldersheim:
@hmusum or @geirst or @bratseth PR


bratseth (commented):
Could you add them to the list of identifiers as tokens as well (near the end I think).

baldersheim:
I think that was what I just did.....
The PR for adding the dictionary handling was merged earlier on in #17122

bratseth:
Right ... 🤓 👍

PR #17138, opened 2021-03-23
Cluster controller master moratorium 

hakonhall:


hakonhall:
@vekterli The first attempt at separating out a state, before the first broadcast is allowed, that is as-if the fleet controller is not a master in some ways, but is a master in other ways.

This causes at least 22 unit test failures in MasterElectionTest and StateChangeTest...  The one I looked at in default_space_nodes_not_marked_as_maintenance_when_cluster_has_no_global_document_types hangs in waitForStableSystem(), waiting for what appears to be acks from distributors on the current version, and I'm not entirely certain where/how that's supposed to work.

hakonhall:
Superseded by #17163

PR #17140, opened 2021-03-23
Gjoranv/merge http into core (rebased) 

gjoranv:
This removes the container-core-config and jdisc_http_service modules. Verified with a simple system test in a centos container.

FYI: @aressem


bjorncs (approved):
🥇 🍾

PR #17152, opened 2021-03-24
Start admin nodes with uncapped CPU

freva:


hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -617,10 +617,8 @@ private Duration warmUpDuration(NodeAgentContext context) {
         Optional<NodeMembership> membership = context.node().membership();
         return zone.getSystemName().isCd()
                || zone.getEnvironment().isTest()
-               || (context.nodeType() != NodeType.tenant)
-               || membership.map(mem -> ! (mem.type().isContainer() ||
-                                           mem.type().isCombined()))
-                                  .orElse(false)
+               || context.nodeType() != NodeType.tenant
+               || membership.map(mem -> ! (mem.type().isContainer() || mem.type().isAdmin())).orElse(false)
```
hmusum, line 621: What about combined nodes? They have containers as well

hmusum: Ah, isContainer() includes combined. That's OK then

PR #17163, opened 2021-03-24
Avoid safe mutations in master moratorium and increase first cluster state broadcast deadline, try 2 [run-systemtest] 

hakonhall:
This PR will
 - Impose a master CC moratorium until it has broadcast its first cluster state
 - Deny the safe setting of node states in moratorium
 - Increase the default time to the first cluster state broadcast from 5s to 120s.

First commit is the double-revert of #17085 .


vekterli (approved):
:+1:

PR #17165, opened 2021-03-24
Add has... methods to ClusterType 

hmusum:
Using `has...` methods and making `is...` methods for equality is more precise and sometimes avoids confusion about what we really check for IMHO.


hakonhall (approved):

freva (approved):
There's also a few users in host-admin

hmusum:
Thanks, will create a PR for usage in host-admin

hmusum:
.

PR #17171, opened 2021-03-25
- Avoid inlining exchange_location_nodes to be able to see its execut…

baldersheim:
…ion cost with.

- Add isIntermediate, and isLocation virtual methods to avoid very expensive dynamic_cast.
- Use search::query::LocationTerm, instead of ProtonLocationTerm when possible.

@havardpe PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/query.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/query.cpp
@@ -36,6 +36,7 @@ using search::queryeval::IntermediateBlueprint;
 using search::queryeval::Blueprint;
 using search::queryeval::IRequestContext;
 using search::queryeval::SearchIterator;
+using search::query::LocationTerm;
 using vespalib::string;
 using std::vector;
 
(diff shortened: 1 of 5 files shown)


havardpe (approved):

searchcore/src/vespa/searchcore/proton/matching/query.cpp
```diff
@@ -59,16 +60,17 @@ inject(Node::UP query, Node::UP to_inject) {
...
+std::vector<LocationTerm *>
 find_location_terms(Node *tree) {
-    std::vector<ProtonLocationTerm *> retval;
+    std::vector<LocationTerm *> retval;
     std::vector<Node *> nodes;
     nodes.push_back(tree);
-    for (size_t i = 0; i < nodes.size(); ++i) {
-        if (auto loc = dynamic_cast<ProtonLocationTerm *>(nodes[i])) {
-            retval.push_back(loc);
+    for (Node * node : nodes) {
```
arnej27959, line 68: this is just wrong, please revert

baldersheim: hmm, did not notice modification of the vector being iterated. Perhaps a unit test for this case is needed.

PR #17175, opened 2021-03-25
add 'is_search' flag when creating match tools factory

havardpe:
drop global filter handling for non-search setups

@arnej27959 maybe something like this?


arnej27959 (approved):
this works well, but do not merge yet.
we need to figure out the other bug which is observed via this one.

PR #17187, opened 2021-03-25
Avoid frequent dynamic_cast to check if a node is an intermediate.

baldersheim:
@havardpe or @arnej27959 PR
I will remove the cast one at a time as I suspect the remaining one is more expensive than the other.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/query.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/query.cpp
@@ -64,11 +64,14 @@ find_location_terms(Node *tree) {
     std::vector<ProtonLocationTerm *> retval;
     std::vector<Node *> nodes;
     nodes.push_back(tree);
+    // Note the nodes vector being iterated over is appended in the loop
     for (size_t i = 0; i < nodes.size(); ++i) {
-        if (auto loc = dynamic_cast<ProtonLocationTerm *>(nodes[i])) {
+        Node * node = nodes[i];
+        if (auto loc = dynamic_cast<ProtonLocationTerm *>(node)) {
             retval.push_back(loc);
         }
-        if (auto parent = dynamic_cast<const search::query::Intermediate *>(nodes[i])) {
+        if (node->isIntermediate()) {
+            auto parent = static_cast<const search::query::Intermediate *>(node);
             for (Node * child : parent->getChildren()) {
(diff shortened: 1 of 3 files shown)


arnej27959 (approved):
👍

searchcore/src/vespa/searchcore/proton/matching/query.cpp
```diff
@@ -64,11 +64,14 @@ find_location_terms(Node *tree) {
     std::vector<ProtonLocationTerm *> retval;
     std::vector<Node *> nodes;
     nodes.push_back(tree);
+    // Note the nodes vector being iterated over is appended in the loop
     for (size_t i = 0; i < nodes.size(); ++i) {
-        if (auto loc = dynamic_cast<ProtonLocationTerm *>(nodes[i])) {
+        Node * node = nodes[i];
+        if (auto loc = dynamic_cast<ProtonLocationTerm *>(node)) {
```
havardpe, line 70: if you are looking for only a single specific class (ensure no sub-classes by making it final), using std::type_index(typeid(...)) is much faster than dynamic cast (and even faster than virtual predicates combined with static cast)

searchcore/src/vespa/searchcore/proton/matching/query.cpp
```diff
@@ -64,11 +64,14 @@ find_location_terms(Node *tree) {
...
             retval.push_back(loc);
         }
-        if (auto parent = dynamic_cast<const search::query::Intermediate *>(nodes[i])) {
+        if (node->isIntermediate()) {
```
havardpe, line 73: a node cannot be both intermediate and a location term at a same time, consider using 'else' to model exclusivity.

baldersheim: That is fixed in an upcoming PR

searchcore/src/vespa/searchcore/proton/matching/query.cpp
baldersheim: hmm, that one is new to me. Will take a look.

havardpe: various strategies are tested in 'vespa/vespalib/src/tests/detect_type_benchmark'

PR #17189, opened 2021-03-25
Revert "Bjorncs/test"

baldersheim:
Reverts vespa-engine/vespa#17186
@bjorncs or @jonmv PR
Broke the build
HttpServerTest.requireThatRequestIsTrackedInAccessLog:879 » ArrayIndexOutOfBounds


jonmv:
Guess it's racy. Can provide a fix instead.

jonmv:
So seems the blocking log that was removed here was a workaround foro asynchronous shutdown of jetty? The test which used that has also failed.

PR #17190, opened 2021-03-26
CopyOnWriteArrayList makes no guarantee about read ordering

jonmv:


jonmv:
Nevermind, the underlying array is `volatile`.

PR #17193, opened 2021-03-26
Add feature flag that can be used to disable OmitStackTraceInFastThro…

hmusum:
…w JVM option.

A bit cumbersome code in `ModelContextImpl`, would have liked a ToBooleanFunction to exist...

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainerCluster.java
@@ -9,6 +9,8 @@ import com.yahoo.vespa.model.container.ContainerCluster;
 import com.yahoo.vespa.model.container.component.Handler;
 import com.yahoo.vespa.model.container.component.SystemBindingPattern;
 
+import java.util.Optional;
+
 /**
  * @author hmusum
  */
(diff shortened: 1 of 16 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainerCluster.java
```diff
@@ -21,7 +19,7 @@ public LogserverContainerCluster(AbstractConfigProducer<?> parent, String name,
 
         addDefaultHandlersWithVip();
         addLogHandler();
-        setJvmGCOptions(deployState.getProperties().jvmGCOptions(Optional.of(ClusterSpec.Type.admin)));
+        setJvmGCOptions(deployState.featureFlags().jvmOmitStackTraceInFastThrowOption(ClusterSpec.Type.admin));
```
baldersheim, line 22: Intentional to remove jvmGCOptions ? Not append instead ?

baldersheim (approved):

config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainerCluster.java
hmusum: Not intentional, thanks, will fix

PR #17195, opened 2021-03-26
Remove call to remove unsued applications for every child event

hmusum:
There is already a background job to remove unused applications,
there should be no need to do it here. It looks like deleting many
applications for the same tenant in a short time frame takes a long
time, this might help.


jonmv:
Hmm, this will make it less responsive when we "have" lost events, though?

jonmv (approved):

hmusum:
Yes, but we have never seen that we have actually lost an event, that code was just a band-aid hoping to fix occasional failures to delete an application. I think we now have enough logging enabled for system test runs that we can figure out if we actually lost events.

PR #17202, opened 2021-03-26
Add groupId metric dimension.

gjoranv:
This should show up as `groupId: group 1` etc. for all metrics on content nodes.


yngveaasheim (approved):
👍

PR #17204, opened 2021-03-26
Manually allocating a stack and growing is more expensive than

baldersheim:
recursing as that is allocation free will have a smaller L1 footprint.

@havardpe or @arnej27959 PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/query.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/query.cpp
@@ -60,25 +60,23 @@ inject(Node::UP query, Node::UP to_inject) {
     return query;
 }
 
-std::vector<LocationTerm *>
-find_location_terms(Node *tree) {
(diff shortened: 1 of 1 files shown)


arnej27959:
we know from bitter experience that this risks using a lot of stack space (potentially more than 256 KB or whatever our current limit is).  If we want to do this generally we should just use 8MB stacks, removing all manual tuning of stack sizes for threads.

baldersheim:
Yes, in general recursion must be used with care. However in this case it is only the maximum depth of the tree
that matters and we already
- Use recursion in other parts of query setup
- Collapses the depth of the query tree before leaving the container.

So I think this should be perfectly safe to use here. As far as I can tell each recursion will use 24 or 32 bytes per iteration so depth should be on the safe side.

However as a orthogonal thing we might consider using larger stacks.
It is more than 10 years since those were set.

havardpe (approved):

PR #17226, opened 2021-03-29
- Let DotProduct,Wand and WeightedSet be Term nodes in the query tree…

baldersheim:
… as they really are.

  That restricts the nodes to what they can really do and makes them significantly cheaper.
- In addition type conversion of numeric terms is delayed to when it is necessary.
   And as next step they can be avoided completely.

@havardpe PR

### DIFF
--- a/searchcore/src/tests/proton/matching/query_test.cpp
+++ b/searchcore/src/tests/proton/matching/query_test.cpp
@@ -164,6 +164,17 @@ fef_test::IndexEnvironment plain_index_env;
 fef_test::IndexEnvironment resolved_index_env;
 fef_test::IndexEnvironment attribute_index_env;
 
+vespalib::string
+termAsString(const search::query::Range &term) {
+    vespalib::asciistream os;
+    return (os << term).str();
+}
+
+const vespalib::string &
+termAsString(const vespalib::string & term) {
+    return term;
+}
+
 void setupIndexEnvironments()
 {
     FieldInfo field_info(FieldType::INDEX, CollectionType::SINGLE, field, field_id);
--- a/searchcore/src/tests/proton/matching/query_test.cpp
+++ b/searchcore/src/tests/proton/matching/query_test.cpp
@@ -898,9 +898,9 @@ void Test::requireThatWeakAndBlueprintsAreCreatedCorrectly() {
 void Test::requireThatParallelWandBlueprintsAreCreatedCorrectly() {
     using search::queryeval::WeakAndBlueprint;
 
-    ProtonWandTerm wand(field, 42, Weight(100), 123, 9000, 1.25);
-    wand.append(Node::UP(new ProtonStringTerm("foo", field, 0, Weight(3))));
-    wand.append(Node::UP(new ProtonStringTerm("bar", field, 0, Weight(7))));
+    ProtonWandTerm wand(2, field, 42, Weight(100), 123, 9000, 1.25);
+    wand.addTerm("foo", Weight(3));
+    wand.addTerm("bar", Weight(7));
 
     ViewResolver viewResolver;
     ResolveViewVisitor resolve_visitor(viewResolver, attribute_index_env);
--- a/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
@@ -13,7 +13,7 @@ public:
 
     vespalib::stringref asString() const override { return _key; }
 private:
-    vespalib::string _key;
+    vespalib::stringref _key;
 };
 }
 
--- a/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
@@ -1,3 +1,32 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "i_document_weight_attribute.h"
+#include <charconv>
+
+namespace search {
+namespace {
+class StringAsKey final : public IDocumentWeightAttribute::LookupKey {
+public:
+    StringAsKey(vespalib::stringref key)
+            : _key(key)
+    { }
+
+    vespalib::stringref asString() const override { return _key; }
+private:
+    vespalib::string _key;
+};
+}
+
+bool
+IDocumentWeightAttribute::LookupKey::asInteger(int64_t &value) const {
+    vespalib::stringref str = asString();
+    const char *end = str.data() + str.size();
+    auto res = std::from_chars(str.data(), end, value);
+    return res.ptr == end;
+}
+
+IDocumentWeightAttribute::LookupResult
+IDocumentWeightAttribute::lookup(vespalib::stringref term, vespalib::datastore::EntryRef dictionary_snapshot) const {
+    return lookup(StringAsKey(term), dictionary_snapshot);
+}
+}
--- a/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
+++ b/searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
@@ -3,17 +3,20 @@
 #pragma once
 
 #include "postinglisttraits.h"
-
 #include <functional>
 
 namespace search {
-
-namespace query { class Node; }
 
 using DocumentWeightIterator = attribute::PostingListTraits<int32_t>::const_iterator;
 
 struct IDocumentWeightAttribute
 {
+    struct LookupKey {
+        virtual ~LookupKey() = default;
+        virtual vespalib::stringref asString() const = 0;
+        virtual bool asInteger(int64_t &value) const;
+    };
+
     struct LookupResult {
         const vespalib::datastore::EntryRef posting_idx;
         const uint32_t posting_size;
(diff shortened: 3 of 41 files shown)


searchlib/src/vespa/searchlib/query/tree/queryreplicator.h
```diff
@@ -78,26 +78,32 @@ class QueryReplicator : private QueryVisitor {
         visitNodes(node.getChildren());
     }
 
+    void replicateMultiTerm(const MultiTerm &original, MultiTerm & replica) {
+        for (uint32_t i(0); i < original.getNumTerms(); i++) {
+            auto v = original.getAsString(i);
+            replica.addTerm(v.first, v.second);
```
havardpe, line 84: this makes the replica different from the original in some cases

searchlib/src/vespa/searchlib/query/tree/stackdumpquerycreator.h
```diff
@@ -45,8 +47,26 @@ class StackDumpQueryCreator {
         return builder.build();
     }
 
-private: 
-    static Term * createQueryTerm(search::SimpleQueryStackDumpIterator &queryStack, QueryBuilder<NodeTypes> & builder, vespalib::stringref & pureTermView) {
+private:
+    static void populateMultiTerm(search::SimpleQueryStackDumpIterator &queryStack, QueryBuilderBase & builder, MultiTerm & mt) {
+        for (uint32_t i(0); i < mt.getNumTerms(); i++) {
+            queryStack.next();
```
havardpe, line 53: need to check for errors and end-of-stack (also, we might want to roll back to a deterministic error state in the next() function in case of en exception, just in case someone does something like this in the future...)

searchcore/src/tests/proton/matching/query_test.cpp
```diff
@@ -164,6 +164,17 @@ fef_test::IndexEnvironment plain_index_env;
 fef_test::IndexEnvironment resolved_index_env;
 fef_test::IndexEnvironment attribute_index_env;
 
+vespalib::string
+termAsString(const search::query::Range &term) {
```
havardpe, line 168: Why are these here? (to avoid updating the test to use the new API?)

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
```diff
@@ -1,3 +1,32 @@
...
+namespace {
+class StringAsKey final : public IDocumentWeightAttribute::LookupKey {
+public:
+    StringAsKey(vespalib::stringref key)
+            : _key(key)
+    { }
+
+    vespalib::stringref asString() const override { return _key; }
+private:
+    vespalib::string _key;
```
havardpe, line 16: you probably do not need a copy here

searchlib/src/vespa/searchlib/query/tree/termnodes.cpp
```diff
@@ -1,26 +1,111 @@
...
+}
+
+class StringTermVector final : public MultiTerm::TermVector {
+public:
+    StringTermVector(uint32_t sz) : _terms() { _terms.reserve(sz); }
+    void addTerm(stringref term, Weight weight) override {
+        _terms.emplace_back(term, weight);
+    }
+    void addTerm(int64_t, Weight) override {
+        badType("string", "int64_t");
```
havardpe, line 42: this is a very well hidden restriction of the query stack format... how is this modeled in Java? is the term data type annotated on the parent multi-term node?

searchlib/src/vespa/searchlib/queryeval/create_blueprint_visitor_helper.cpp
```diff
@@ -66,12 +66,12 @@ template <typename WS, typename NODE>
...
 CreateBlueprintVisitorHelper::createWeightedSet(std::unique_ptr<WS> bp, NODE &n) {
     FieldSpecList fields;
-    for (size_t i = 0; i < n.getChildren().size(); ++i) {
+    for (size_t i = 0; i < n.getNumTerms(); ++i) {
         fields.clear();
         fields.add(bp->getNextChildField(_field));
-        const query::Node &node = *n.getChildren()[i];
-        uint32_t weight = getWeightFromNode(node).percent();
-        bp->addTerm(_searchable.createBlueprint(_requestContext, fields, node), weight);
+        const auto & term = n.getAsString(i);
```
havardpe, line 72: when you have a const reference to a temporary object that contains a string reference to a mutable scratch buffer things become a bit more complicated than they seem...

baldersheim: Fixed

searchcore/src/tests/proton/matching/query_test.cpp
baldersheim: Only used in test.

searchlib/src/vespa/searchlib/query/tree/stackdumpquerycreator.h
baldersheim: Fixed

searchlib/src/vespa/searchlib/query/tree/queryreplicator.h
baldersheim: Fixed by introducing a type.

searchlib/src/vespa/searchlib/attribute/multistringpostattribute.hpp
```diff
@@ -108,10 +108,10 @@ MultiValueStringPostingAttributeT<B, T>::DocumentWeightAttributeAdapter::get_dic
 
 template <typename B, typename T>
 IDocumentWeightAttribute::LookupResult
-MultiValueStringPostingAttributeT<B, T>::DocumentWeightAttributeAdapter::lookup(const vespalib::string &term, vespalib::datastore::EntryRef dictionary_snapshot) const
+MultiValueStringPostingAttributeT<B, T>::DocumentWeightAttributeAdapter::lookup(const LookupKey & key, vespalib::datastore::EntryRef dictionary_snapshot) const
 {
     const IEnumStoreDictionary& dictionary = self._enumStore.get_dictionary();
-    auto comp = self._enumStore.make_folded_comparator(term.c_str());
+    auto comp = self._enumStore.make_folded_comparator(key.asString().data());
```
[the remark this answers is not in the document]
baldersheim: Ensured it is zero terminated at the source.
However the zero termination restriction should be lifted on the comparators in the enumstore.

searchlib/src/vespa/searchlib/query/tree/termnodes.cpp
baldersheim: Restriction lifted

baldersheim:
I have added some basic testing of MultiTerm too.

searchlib/src/vespa/searchlib/attribute/multistringpostattribute.hpp
havardpe, line 114: I still think this is dangerous, since it makes the LookupKey API dishonest and using it in a natural way could cause bad results.

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
```diff
@@ -3,17 +3,20 @@
...
 
-namespace query { class Node; }
-
 using DocumentWeightIterator = attribute::PostingListTraits<int32_t>::const_iterator;
 
 struct IDocumentWeightAttribute
 {
+    struct LookupKey {
+        virtual ~LookupKey() = default;
+        virtual vespalib::stringref asString() const = 0;
```
havardpe, line 16: this should probably have a warning sign about the return value being 0-terminated

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
```diff
@@ -25,7 +28,8 @@ struct IDocumentWeightAttribute
             : posting_idx(posting_idx_in), posting_size(posting_size_in), min_weight(min_weight_in), max_weight(max_weight_in), enum_idx(enum_idx_in) {}
     };
     virtual vespalib::datastore::EntryRef get_dictionary_snapshot() const = 0;
-    virtual LookupResult lookup(const vespalib::string &term, vespalib::datastore::EntryRef dictionary_snapshot) const = 0;
+    virtual LookupResult lookup(const LookupKey & key, vespalib::datastore::EntryRef dictionary_snapshot) const = 0;
+    LookupResult lookup(vespalib::stringref term, vespalib::datastore::EntryRef dictionary_snapshot) const;
```
havardpe, line 32: this should probably have a warning sign about the term needing to be 0-terminated

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
```diff
@@ -13,7 +13,7 @@ class StringAsKey final : public IDocumentWeightAttribute::LookupKey {
 
     vespalib::stringref asString() const override { return _key; }
 private:
-    vespalib::string _key;
+    vespalib::stringref _key;
```
havardpe, line 16: it might actually be better to keep the string, to avoid the 0-terminated stringref requirement to escape the LookupKey API and into the attribute API...

searchlib/src/vespa/searchlib/attribute/multistringpostattribute.hpp
havardpe: I suggest adding an assert that verifies the 0-termination to make sure it fails in a deterministic way

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
baldersheim: Note added.

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.h
baldersheim: Adde a note being only used in tests. Zero termination handled on the inside.

searchlib/src/vespa/searchlib/attribute/multistringpostattribute.hpp
baldersheim: Assert added

searchlib/src/vespa/searchlib/attribute/i_document_weight_attribute.cpp
baldersheim: done

baldersheim:
Comments addressed in https://github.com/vespa-engine/vespa/pull/17251

PR #17228, opened 2021-03-29
Store resized event for Node

freva:
I think this fixes VESPA-19337? I wasn't at that ticket meeting and it doesn't say what was decided, but my conclusion in original ticket was at least that it is harmless and impossible to prevent as long we allocate cluster at the time, but having the event makes it easier to debug at least.


bratseth (approved):
👍

PR #17230, opened 2021-03-29
Gjoranv/merge mbus modules

gjoranv:
This merges jdisc_messagebus_service into container-messagebus, and removes the empty module messagebus-disc. The latter was used to build a messagebus bundle that was not deployed anywhere.


freva (approved):

bjorncs (commented):
🥇

PR #17231, opened 2021-03-29
Remove duplicate test deps.

gjoranv:
I probably forgot to remove these when rearranging deps in an earlier PR.


gjoranv:
Adding this to another PR instead.

PR #17236, opened 2021-03-30
add wiring to allow NNS for imported tensor

arnej27959:

@havardpe please review
@geirst FYI


havardpe (commented):
should probably still limit to only allowing dense tensors (checking the tensor value type, not the attribute class type)

arnej27959:
> should probably still limit to only allowing dense tensors (checking the tensor value type, not the attribute class type)

good catch, done

PR #17237, opened 2021-03-30
Cap number of log entries per fetch

jonmv:


freva (approved):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/LogEntry.java
```diff
@@ -69,6 +69,7 @@ public String message() {
                                                     parts[6].replaceAll("\\\\n", "\n")
                                                             .replaceAll("\\\\t", "\t")))
                          .filter(entry -> entry.at().isAfter(from))
+                         .limit(1_000_000)
```
baldersheim, line 72: 1 million lines, that sounds a bit on the high side.

jonmv: Can probably set it lower, too. OOM at 40M lines, so it should be enough to fix _that_ issue, at least.

PR #17251, opened 2021-03-31
Balder/ensure and check zero termination

baldersheim:
@havardpe PR


havardpe (approved):
looks good. Consider also doing assert/throw in the virtual function adding a string token to an int multi-term, since that is no longer possible.

PR #17256, opened 2021-04-01
Resolve container-cpu-cap based on cluster type and ID

hakonhall:
The goal is to enable higher cap for the cluster type "admin" and cluster ID "cluster-controllers".  If the cap is applied to cluster type "admin" it would also include "logserver" (which would probably be OK).

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
@@ -1,6 +1,7 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.node.admin.nodeagent;
 
+import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.DockerImage;
 import com.yahoo.config.provision.Environment;
 import com.yahoo.config.provision.NodeType;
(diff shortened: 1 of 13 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -379,13 +379,25 @@ private Container updateContainerIfNeeded(NodeAgentContext context, Container ex
...
+            DoubleFlag containerCpuCapFlag = containerCpuCap.with(FetchVector.Dimension.HOSTNAME, context.node().hostname());;
+
+            if (context.node().owner().isPresent()) {
+                containerCpuCapFlag = containerCpuCapFlag.with(FetchVector.Dimension.APPLICATION_ID,
+                                                               context.node().owner().get().serializedForm());
+            }
+
+            if (context.node().membership().isPresent()) {
+                containerCpuCapFlag = containerCpuCapFlag
+                        .with(FetchVector.Dimension.CLUSTER_TYPE, context.node().membership().get().type().value())
```
freva, line 395: I wish there was a `with(FetchVector.Dimension, Optional<String>)` method also so one could just do
```
.with(FetchVector.Dimension.HOSTNAME, context.node().hostname())
.with(FetchVector.Dimension.APPLICATION_ID, context.node().owner().map(ApplicationId::serializedForm))
.with(...)
```

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
hakonhall: Good suggestion, fixed.

PR #17269, opened 2021-04-05
Use separate executors for ZooKeeper watchers for applications and se…

hmusum:
…ssions

Session events might involve doing expensive operations (e.g. creating
config models for an application). Having just one executor has been
observed to lead to events for applications take so long that some
operations (e.g. deleting an application) times out and prevents holding
state in sync across config servers.


jonmv (approved):
Yes, yes ...

PR #17270, opened 2021-04-05
Actually throw exception

hmusum:


hmusum:
Hm, the code seems to do the right thing, it's just that the IDE complains about exception not being thrown and usage for method says `Usage: throw throwUnchecked(e);`

bjorncs (approved):

PR #17274, opened 2021-04-06
Update junit dependency to unit-vintage to combine 4 and 5

jonmv:
@bjorncs please review and merge. @arnej27959 FYI :)

### DIFF
--- a/vespajlib/pom.xml
+++ b/vespajlib/pom.xml
@@ -68,11 +68,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


arnej27959:
what about all the other modules using jupiter?

jonmv:
Mhm. Was thinking we could just replace all Junit4 dependencies with jupiter-vintage-engine, but need to verify first if that works standalone.

jonmv:
... because this is a real gotcha >_<

bjorncs (approved):

PR #17282, opened 2021-04-06
add optimization for join between bfloat16 tensor and number

arnej27959:

@havardpe please review

### DIFF
--- a/eval/src/vespa/eval/instruction/join_with_number_function.cpp
+++ b/eval/src/vespa/eval/instruction/join_with_number_function.cpp
@@ -93,7 +93,7 @@ JoinWithNumberFunction::compile_self(const ValueBuilderFactory &, Stash &stash) 
 {
     const auto &param = stash.create<JoinWithNumberParam>(result_type(), _function);
     auto input_type = (_primary == Primary::LHS) ? lhs().result_type() : rhs().result_type();
-    assert(result_type() == input_type.map());
+    assert(result_type() == ValueType::join(input_type, ValueType::double_type()));
     auto op = typify_invoke<4,MyTypify,SelectJoinWithNumberOp>(input_type.cell_meta(),
                                                                _function,
                                                                primary_is_mutable(),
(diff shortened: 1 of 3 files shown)


eval/src/vespa/eval/instruction/join_with_number_function.cpp
```diff
@@ -98,14 +119,12 @@ JoinWithNumberFunction::optimize(const TensorFunction &expr, Stash &stash)
             const ValueType &result_type = join->result_type();
             const TensorFunction &lhs = join->lhs();
             const TensorFunction &rhs = join->rhs();
-            if (lhs.result_type().is_double() &&
-                (result_type == rhs.result_type()))
-            {
+            if (lhs.result_type().is_double()) {
+                assert(result_type.dimensions() == rhs.result_type().dimensions());
```
havardpe, line 123: these asserts are just weaker forms of the input_type.map() -> output_type assert in compile_self. consider removing them (or use the same assert for consistency).

eval/src/vespa/eval/instruction/join_with_number_function.cpp
```diff
@@ -64,22 +80,27 @@ JoinWithNumberFunction::~JoinWithNumberFunction() = default;
...
+using MyTypify = TypifyValue<TypifyCellMeta,vespalib::TypifyBool,operation::TypifyOp2>;
 
 InterpretedFunction::Instruction
-JoinWithNumberFunction::compile_self(const ValueBuilderFactory &, Stash &) const
+JoinWithNumberFunction::compile_self(const ValueBuilderFactory &, Stash &stash) const
 {
-    auto op = typify_invoke<4,MyTypify,SelectJoinWithNumberOp>(result_type().cell_type(),
+    const auto &param = stash.create<JoinWithNumberParam>(result_type(), _function);
+    auto input_type = (_primary == Primary::LHS) ? lhs().result_type() : rhs().result_type();
+    assert(result_type() == input_type.map());
```
havardpe, line 98: this is a convenient truth, but not really for us to decide. prefer checking join with ValueType::double_type() instead.

eval/src/vespa/eval/instruction/join_with_number_function.cpp
```diff
@@ -14,39 +14,55 @@ namespace vespalib::eval {
...
+    template<typename CM, typename Fun,
+             typename Inplace, typename NumberWasLeft>
     static auto invoke() {
-        return my_number_join_op<CT, Fun, InputIsMutable::value, NumberWasLeft::value>;
+        constexpr CellMeta icm = CM::value;
+        constexpr CellMeta num(CellType::DOUBLE, true);
+        constexpr CellMeta ocm = CellMeta::join(icm, num); 
+        using ICT = CellValueType<icm.cell_type>;
+        using OCT = CellValueType<ocm.cell_type>;
+        constexpr bool inplace = (Inplace::value && std::is_same_v<ICT,OCT>);
```
havardpe, line 64: this makes it look like Inplace::value is not authorative. I suggest adding a non-static assert checking that Inplace::value implies ICT and OCT being the same type; to trap the generated code that will never be called instead of handling parameter inconsistencies silently.

havardpe: an alternative is to downgrade the flag from 'inplace' to 'pri_mut' and infer inplace as done here. That will reduce unused code generation.

havardpe (approved):
looks good, some minor comments to consider

havardpe (approved):

PR #17304, opened 2021-04-07
Use async http the simple way.

baldersheim:
@bjorncs PR
This is an intermediate step on the way to fully streaming.


baldersheim:
I will coordinate merge

bjorncs (approved):

PR #17307, opened 2021-04-08
Add bfloat16 and int8 tensor cell types in Java

lesters:
@arnej27959 Please review.
@havardpe FYI.
@bratseth Anything I might have forgotten here?

This is primarily for serialization and deserialization. There might be some gaps in evaluation and tensor cell value type resolving, I'll get to that in a follow-up.


vespajlib/src/main/java/com/yahoo/tensor/IndexedTensor.java
```diff
@@ -225,6 +225,10 @@ static void indexedBlockToString(IndexedTensor tensor, Indexes indexes, StringBu
                 b.append(tensor.get(index));
             else if (tensor.type().valueType() == TensorType.Value.FLOAT)
                 b.append(tensor.getFloat(index));
+            else if (tensor.type().valueType() == TensorType.Value.BFLOAT16)
+                b.append(tensor.getFloat(index));
```
arnej27959, line 229: it's probably better to use a switch instead of if-else-if-else here and similar places

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
```diff
@@ -59,6 +59,9 @@ public static Value largestOf(List<Value> values) {
 
         public static Value largestOf(Value value1, Value value2) {
             if (value1 == DOUBLE || value2 == DOUBLE) return DOUBLE;
+            if (value1 == FLOAT || value2 == FLOAT) return FLOAT;
```
arnej27959, line 62: while this looks like it will promote int8->float, it won't work correctly in many cases, and now it's not longer doing "largestOf" anymore.  It's probably best to make this actually "largestOf" and do something else when type resolving the various operations.

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
```diff
@@ -69,8 +72,10 @@ public static Value fromId(String valueTypeString) {
             switch (valueTypeString) {
                 case "double" : return Value.DOUBLE;
                 case "float" : return Value.FLOAT;
-                default : throw new IllegalArgumentException("Value type must be either 'double' or 'float'" +
-                                                             " but was '" + valueTypeString + "'");
+                case "bfloat16" : return Value.BFLOAT16;
```
arnej27959, line 75: consider looping over Value.values() and comparing with their "id" instead of repeating the names here.

bratseth:
LGTM!

vespajlib/src/main/java/com/yahoo/tensor/serialization/DenseBinaryFormat.java
```diff
@@ -66,6 +68,16 @@ private void encodeFloatCells(IndexedTensor tensor, GrowableByteBuffer buffer) {
             buffer.putFloat(tensor.getFloat(i));
     }
 
+    private void encodeBFloat16Cells(IndexedTensor tensor, GrowableByteBuffer buffer) {
+        for (int i = 0; i < tensor.size(); i++)
+            buffer.putShort((short)(Float.floatToRawIntBits(tensor.getFloat(i)) >>> 16));
```
arnej27959, line 73: consider adding some utility methods (in TypedBinaryFormat maybe) for the low-level bit shuffling:
```
    static short bitsFromBFloat16(float bf16) {
        int bits = Float.floatToRawIntBits(bf16);
        return (short) (bits >>> 16);
    }

    static float bFloat16FromBits(short bits) {
        return Float.intBitsToFloat(bits << 16);
    }
 ```

arnej27959 (approved):
looks good - only minor comments from me, mostly cosmetic.

vespajlib/src/main/java/com/yahoo/tensor/IndexedTensor.java
lesters: Thanks, fixed.

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
lesters: Thanks, fixed.

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
lesters: Thanks, fixed.

vespajlib/src/main/java/com/yahoo/tensor/serialization/DenseBinaryFormat.java
lesters: Thanks, fixed.

arnej27959 (approved):
👍

PR #17310, opened 2021-04-08
Move nodes to 'failed' during activate

bratseth:
Node failing was based on a quick hack where we made a change to active nodes outside deployment.
This fixes that by introducing a wantToFail setting.

That avoids the spurious scalingEvents entries @ldalves and @freva.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
@@ -91,9 +91,9 @@ public class NodeList extends AbstractFilteringList<Node, NodeList> {
                                 !node.status().vespaVersion().get().equals(node.allocation().get().membership().cluster().vespaVersion()));
     }
 
-    /** Returns the subset of nodes with want to fail status matching the given value */
-    public NodeList wantToFail(boolean wantToFail) {
-        return matching(node -> node.status().wantToFail() == wantToFail);
+    /** Returns the subset of nodes with want to fail set to true */
+    public NodeList failing() {
+        return matching(node -> node.status().wantToFail());
     }
 
     /** Returns the subset of nodes that are currently changing their OS version to given version */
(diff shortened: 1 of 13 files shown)


freva:
What about nodes failed manually by operator?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
```diff
@@ -91,6 +91,11 @@ public NodeList changingVersion() {
                                 !node.status().vespaVersion().get().equals(node.allocation().get().membership().cluster().vespaVersion()));
     }
 
+    /** Returns the subset of nodes with want to fail status matching the given value */
+    public NodeList wantToFail(boolean wantToFail) {
```
mpolden, line 95: This should not take a `boolean` as the parent class of this provides inverting through `not()`. E.g. you can do `list.wantToFail()` and `list.not().wantToFail()` to invert the condition.

Consider naming this method `failing`, similar to the existing `deprovisioning` and `rebuilding` methods.

mpolden:
> What about nodes failed manually by operator?

Won't `OperatorChangeApplicationMaintainer` take care of that?

freva:
`OperatorChangeApplicationMaintainer` will eventually allocate a new node to replace the failed node, yes. My comment was to 
> That avoids the spurious scalingEvents entries

If an operator fails out a node, it is immediately moved to `failed` and on the next deployment configserver will see that the application has gone form `x` nodes to `x+1` nodes, which is a diff, so it creates a bogus scaling event.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
bratseth: Yes, the only drawback with not passing a boolean is the cases where it is a parameter in the caller. Anyway, I'll follow the pattern, I didn't find one, but failing etc. is fine.

mpolden (approved):

bratseth:
Yes, nodes failed by operators should absolutely use this. I'll do that next.

PR #17334, opened 2021-04-09
add archive bucket count metric

andreer:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArchiveAccessMaintainer.java
```diff
@@ -19,19 +22,32 @@
...
     @Override
     protected boolean maintain() {
+
+        // Count buckets - so we can alert if we get close to the account limit of 1000
+        metric.add(bucketCountMetricName,
+                zoneRegistry.zones().all().ids().stream()
+                        .map(archiveBucketDb::buckets)
+                        .mapToLong(Collection::size)
+                        .sum(),
+                metric.createContext(Map.of()));
```
freva, line 49: We want at least the `system` dimension here. But ideally i think we should report per zone with `zone` dimension and we can aggregate in yamas for entire system if we want.

freva (approved):

PR #17341, opened 2021-04-09
Havardpe/integrate new cell types with onnx

havardpe:

@arnej27959 FYI
we need a newer version of onnxruntime to be able to evaluate models with bfloat16/int8
the test uses opset 12, but Add needs to be opset 14 to support bfloat16/int8

### DIFF
--- a/eval/src/tests/tensor/onnx_wrapper/dynamic.py
+++ b/eval/src/tests/tensor/onnx_wrapper/dynamic.py
@@ -35,5 +35,5 @@ graph_def = helper.make_graph(
     ],
     [OUTPUT],
 )
-model_def = helper.make_model(graph_def, producer_name='dynamic.py')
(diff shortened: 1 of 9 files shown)


arnej27959:
even with onnxruntime 1.7.1 the test cannot work, it only has opset 13 and Add doesn't work with int8 or bfloat16.  But it's possible to add some temporary variables and use the Cast operator and then the test works.

I think we can just skip the zero-copy test, since using int8 and bfloat16 inside onnx seems to be for the future only; we should instead test that we can use vespa int8 and bfloat16 on the vespa side and use the "simple.onnx" model and that conversion works correctly.

havardpe:
I know that we need opset 14 for Add to work with int8, but I thought Cast was opset 14 as well...?

If we are to drop testing the zero-copy path we should also remove the support for int8 inputs/outputs in our onnx integration... I think we want these tests as soon as possible.

havardpe:
superseeded by https://github.com/vespa-engine/vespa/pull/17411

PR #17346, opened 2021-04-09
Jonmv/apache hc5 client

jonmv:
@bjorncs please review.
@hakonhall FYI.


configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
```diff
@@ -0,0 +1,263 @@
...
+            return this;
+        }
+
+        @Override
+        public ConfigServerClient.RequestBuilder body(byte[] json) {
+            return body(HttpEntities.create(json, ContentType.APPLICATION_JSON));
+        }
+
+        @Override
+        public RequestBuilder body(HttpEntity entity) {
```
bjorncs, line 147: Consider adding body overloads for Slime object and Jackson type as well.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+    }
+
+    /** What host(s) to try for a request, in what order. A host may be specified multiple times, for retries.  */
+    @FunctionalInterface
+    interface HostStrategy extends Iterable<URI> {
+
+        /** Attempts each request once against each listed host. */
+        static HostStrategy ordered(List<URI> hosts) {
```
bjorncs, line 93: Is it only the hostname that is different among the supplied URIs? Consider taking a list of hostnames and the other fixed URI parts as additional parameters

bjorncs (approved):
Nice!

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
jonmv: Yes, I considered this back and forth, and so far I think that's better placed in the users of this, but I may still change my mind. 
```java
    .map(bytes -> mapper.readValue(bytes, Some.class)
```
and 
```java
    .map(someSerializer::fromJson)
```
seem like the best approach right now, imho.

jonmv: I guess it was called `read(...)` now, not `map(...)`, but whatever.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
jonmv: Hmm ... I guess that's true right now. Will consider it.

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+    }
+
+    @SuppressWarnings("unchecked")
+    private static <T extends Throwable> void sneakyThrow(Throwable t) throws T {
+        throw (T) t;
+    }
+
+    private static ConfigServerException readException(byte[] serialised) {
```
hakonhall, line 253: This mechanism hides the HTTP response code, and the body unless it is in a specific format, and both are very useful.  E.g. the Orchestrator contains a detailed response on 409 Conflict.  I'm guessing Jetty may also return useful error responses (?).

Also, deserializing HTTP error responses in this class seems inconsistent with requiring clients to deserialize successful responses.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
```diff
@@ -0,0 +1,148 @@
...
+        /** Sets the request body. */
+        RequestBuilder body(HttpEntity entity);
+
+        /** Sets the parameter key/values for the request. Number of arguments must be even. */
+        RequestBuilder parameters(String... pairs);
+
+        /** Overrides the default socket read timeout of the request. {@code Duration.ZERO} gives infinite timeout. */
+        RequestBuilder timeout(Duration timeout);
+
+        /** Overrides the default socket read timeout of the request. {@code null} allows infinite timeout. */
```
hakonhall, line 50: Description not matching method.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
```diff
@@ -0,0 +1,148 @@
...
+        /**
+         * Sets custom retry/failure logic for this.
+         * <p>
+         * Exactly one of the arguments (response, exception) are non-null.
+         * Return a value to have that returned to the caller;
+         * throw a {@link RetryException} to have the request retried; or
+         * throw any other unchecked exception to have this propagate out to the caller.
+         * The caller must close the provided response, if any.
+         */
+        <T> T handle(BiFunction<ClassicHttpResponse, IOException, T> handler) throws UncheckedIOException;
```
hakonhall, line 62: How about defining two handlers, one to deal with ClassicHttpResponse and the other to deal with IOException?  Otherwise all handlers would be similar to

```
if (exception == null)
  return handleResponse(response);
else
  return handleException(exception);
```

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
```diff
@@ -0,0 +1,148 @@
...
+
+        /** Returns the raw input stream of the response, if successful. The caller must close the returned stream. */
+        InputStream stream() throws UncheckedIOException, ConfigServerException;
+
+    }
+
+    /** Exception wrapper that signals retries should be attempted. */
+    final class RetryException extends RuntimeException {
+
+        public RetryException(IOException cause) {
```
hakonhall, line 78: Note: I would be OK with a checked exception for this, as this exception is not supposed to escape from the client, and would cause compilation error if it does.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
jonmv: Thanks.

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
```diff
@@ -0,0 +1,263 @@
...
+
+            for (int i = 0; i < pairs.length; )
+                uriBuilder.setParameter(pairs[i++], pairs[i++]);
+
+            return this;
+        }
+
+        @Override
+        public RequestBuilder timeout(Duration timeout) {
+            return config(RequestConfig.copy(defaultRequestConfig)
```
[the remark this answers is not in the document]
jonmv: Thanks.

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
jonmv: This is what's currently done in the controller, and covers all uses there. I'll see whether it makes sense to change the orchestrator responses to match this, or not. 

So, this is what you have the `handle(...)` for. But you want both a different exception deserialisation _and_ easy response mapping at the same time, I guess. Fair enough. I'll get back to this.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
jonmv: I considered that, but limitations in Java generics then causes you to have to always call one of them before the other, which I think is much more confusing.

jonmv: Uhm, I guess you ocould pass two `Function`s ... yes.

hakonhall: Yes, I was thinking of the latter.

hakonhall: Perhaps you were thinking of having two of these handle() methods (say `handle` and `handleException`).  That would allow you to define a default behavior with IOException which may be useful?

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
jonmv: So there are three things: 
1. Handle successful response.
2. Handle non-2XX.
3. Handle exception.

Perhaps `handle` should accept a function for each of these, and overloads could let you specify only a subset. There's some code to handle exceptions and re-throw that the user shouldn't have to reimplement.

configserver-client/src/main/java/ai/vespa/hosted/client/ConfigServerClient.java
hakonhall: Presumably a RuntimeException could also be caught, and ought to be forwarded to a handle method?

jonmv: The apache client API doesn't list runtime exceptions as part of its API, but states problems and failures are wrapped in an IOException.

configserver-client/src/main/java/ai/vespa/hosted/client/AbstractConfigServerClient.java
hakonhall: Perhaps it should be possible to provide a function that overrides the default handling of (2), and:
 - if it throws an exception propagate that
 - if it returns true, fall back to default handling of (2)
 - if it returns false, fall back to handling as-if (1)

Similarly with (3).

An Orchestrator client could set custom (2) and (3).

jonmv: I think I'll rather expose a default handler for each, and make it easy to compose with those. But I started doubting whether a good default for 2. exists.

hakonhall: We could just defer any such refactoring then until we know more about the config server clients.

jonmv: Yes, that's the plan :)

PR #17358, opened 2021-04-12
Bratseth/want to fail on operator fail

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
```diff
@@ -277,6 +280,12 @@ private void ensureSingleCluster() {
     /** Returns the nodes of this as a stream */
     public Stream<Node> stream() { return asList().stream(); }
 
+    public static NodeList of(Node ... nodes) {
+        List<Node> nodeList = new ArrayList<>();
+        Collections.addAll(nodeList, nodes);
+        return copyOf(nodeList);
```
mpolden, line 286: Can be simplified to `return copyOf(List.of(nodes))`.

mpolden (approved):
LGTM, but build failed. I can't restart it without granting Screwdriver full access to my GH account.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeList.java
bratseth: Thanks, done

bratseth:
Build error is unrelated - merging.

PR #17371, opened 2021-04-12
add TypeResolver

arnej27959:
* with unit tests mostly cribbed from C++ version

@havardpe please review
@lesters please review

### DIFF
--- a/vespajlib/abi-spec.json
+++ b/vespajlib/abi-spec.json
@@ -1459,6 +1459,25 @@
     ],
     "fields": []
   },
+  "com.yahoo.tensor.TypeResolver": {
+    "superClass": "java.lang.Object",
+    "interfaces": [],
+    "attributes": [
+      "public"
+    ],
+    "methods": [
+      "public void <init>()",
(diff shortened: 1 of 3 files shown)


havardpe (approved):
I have had a quick look and it seems ok (concat a bit  complex)
I will let @lesters decide how it fits with existing code

lesters (approved):
So, we already have type resolving associated with each of the primitive Tensor classes, as each `TensorFunction` must implement the `type` method, so this PR introduces some duplication. However, tensor cell value type is not sufficiently handled in existing code. The point of this PR is that it is difficult to split the type resolving in terms of tensor dimensions etc from that of the resulting cell value types in some cases.

I think the code here is preferable to the way most of the type resolving is handled in existing code as that is spread a bit around. Also, this code closely resembles the logic in the C++ version. I am unsure if it is more logical to move the actual type resolving code (which is centralized in this class) to the actual classes implementing the functions. However, this PR does not wire in the code, so we have to address that in a follow-up anyway.

So let's pull in this PR now and then decide how to wire it in. LGTM!

PR #17376, opened 2021-04-12
Balder/allow cased uncased

baldersheim:
@bratseth and @geirst PR
This only propagates correct attributes config for the backend.
Lowercasing control is up next.

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
@@ -1,0 +1,9 @@
+package com.yahoo.searchdefinition.document;
+
+public enum Case {
+    CASED("cased"),
+    UNCASED("uncased");
+    private String name;
+    Case(String name) { this.name = name; }
+    public String getName() { return name;}
+}
--- a/config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
@@ -538,9 +538,12 @@ public class SDField extends Field implements TypedKey, FieldOperationContainer,
      * Returns Dictionary settings.
      */
     public Dictionary getDictionary() { return dictionary; }
-
-
-    public void setDictionary(Dictionary dictionary) { this.dictionary=dictionary; }
+    public Dictionary getOrSetDictionary() {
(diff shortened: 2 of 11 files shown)


bratseth (approved):
👍

config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
```diff
@@ -0,0 +1,9 @@
+package com.yahoo.searchdefinition.document;
```
geirst, line 1: Please add copyright.

config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
```diff
@@ -0,0 +1,9 @@
...
+
+public enum Case {
```
geirst, line 3: Consider adding comment.

config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
```diff
@@ -554,6 +557,17 @@ public void setMatchingType(Matching.Type type) {
         }
     }
 
+    /**
+     * Set the matching type for this field and all subfields.
+     */
+    // TODO: When this is not the same as getMatching().setthis we have a potential for inconsistency. Find the right
```
geirst, line 563: Typo: 'setthis'

geirst (commented):
👍

config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
baldersheim: Fixed

config-model/src/main/java/com/yahoo/searchdefinition/document/Case.java
baldersheim: Fixed

config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
baldersheim: I just copied the comment from setMatchingType and setMatchingAlgorithm just to ensure that they all got clenaed up when the original author of the comment decided to clean this up. Not sure what he really meant. Hope it was not me......

PR #17377, opened 2021-04-12
Non-functional changes only

bratseth:


container-search/src/main/java/com/yahoo/search/dispatch/rpc/ProtobufSerialization.java
```diff
@@ -196,7 +196,7 @@ static InvokerResult convertToResult(Query query, SearchProtocol.SearchReply pro
         result.getResult().setTotalHitCount(protobuf.getTotalHitCount());
         result.getResult().setCoverage(convertToCoverage(protobuf));
 
-        var haveGrouping = protobuf.getGroupingBlob() != null && !protobuf.getGroupingBlob().isEmpty();
+        var haveGrouping = ! protobuf.getGroupingBlob().isEmpty();
```
baldersheim, line 199: Are you sure it can not be null ?

container-search/src/main/java/com/yahoo/search/dispatch/rpc/ProtobufSerialization.java
```diff
@@ -219,7 +219,7 @@ static InvokerResult convertToResult(Query query, SearchProtocol.SearchReply pro
         }
 
         var slimeTrace = protobuf.getSlimeTrace();
-        if (slimeTrace != null && !slimeTrace.isEmpty()) {
+        if ( ! slimeTrace.isEmpty()) {
```
baldersheim, line 222: Are you sure it can not be null ?

baldersheim (approved):
As long as you have verified that the null check can be removed this should be fine.

container-search/src/main/java/com/yahoo/search/dispatch/rpc/ProtobufSerialization.java
bratseth: IntelliJ is sure.

container-search/src/main/java/com/yahoo/search/dispatch/rpc/ProtobufSerialization.java
bratseth: IntelliJ is sure.

PR #17401, opened 2021-04-13
Avoid to have a period of a day since that is a normal period in cron…

baldersheim:
…tab like job control.

@geirst or @toregge PR
@yngveaasheim @hakonhall @vekterli FYI

### DIFF
--- a/searchcore/src/vespa/searchcore/config/proton.def
+++ b/searchcore/src/vespa/searchcore/config/proton.def
@@ -71,8 +71,10 @@ flush.memory.each.maxmemory long default=1073741824
 flush.memory.each.diskbloatfactor double default=0.2
 
 ## Age of unflushed content before forcing age prioritization.
-## Unit is seconds with 1 day being the default.
-flush.memory.maxage.time double default=86400.0
+## Unit is seconds with 31 hours being the default.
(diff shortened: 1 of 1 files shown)


searchcore/src/vespa/searchcore/config/proton.def
```diff
@@ -71,8 +71,10 @@ flush.memory.each.maxmemory long default=1073741824
 flush.memory.each.diskbloatfactor double default=0.2
 
 ## Age of unflushed content before forcing age prioritization.
-## Unit is seconds with 1 day being the default.
-flush.memory.maxage.time double default=86400.0
+## Unit is seconds with 31 hours being the default.
+## 31 is selected as it is both a prime and (31-24=7) is a prime and hence it will
+## not be closer to a multiple of a day for a month, and it will be at least one hour away.
+flush.memory.maxage.time double default=111600.0
```
hakonhall, line 77: 111599 is a prime

baldersheim: Yes, but that will only bring it 31 sec away on the 31st day, and since the duration is measured in minutes I think it is better  to stick to hours.

geirst (approved):
👍

PR #17405, opened 2021-04-13
allow newer onnx runtime

havardpe:

@aressem please review


havardpe:
@arnej27959 FYI

toregge:
This might trigger issues with older version of vespa needing libonnxruntime.so.1.4.0
while libonnxruntime.so.1.7.1 is installed instead. The library version is encoded in the
run time link dependy (cf. `objdump -p vespa-proton-bin  | grep libonnx`).

havardpe:
so, it is better to just bump directly to exactly 1.7.1 and allow some temporary breakage?
(the plan was to update spec references, then update to 1.7.1, then update references again, possibly with = instead of >=...)

toregge:
Yes. We have a similar issue with protobuf version.

PR #17408, opened 2021-04-13
Test concat

bratseth:


lesters (approved):
This wasn't tested previously? I'm surprised. 🥇

PR #17413, opened 2021-04-13
Collapse nesten compatible weakands

bratseth:
@arnej27959 questions:
- Does it make any sense to set indexName on the WeakAndItem itself?
- Shouldn't it be output as WeakAnd not WAND in the (debug) string?

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/query/QueryCanonicalizer.java
+++ b/container-search/src/main/java/com/yahoo/prelude/query/QueryCanonicalizer.java
@@ -94,7 +94,7 @@ public class QueryCanonicalizer {
         if (composite instanceof RankItem || composite instanceof NotItem) {
             collapseLevels(composite, composite.getItemIterator()); // collapse the first item only
         }
-        else if (composite instanceof AndItem || composite instanceof OrItem) {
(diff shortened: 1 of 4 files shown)


arnej27959 (approved):
looks good to me. after some digging I have some guesses on your questions:

> Does it make any sense to set indexName on the WeakAndItem itself?

the backend looks like it's prepared to get "pure" terms (without index name) as children of the WeakAndItem.  But I think that for this operator the children are allowed to have different indexes, so the possibility isn't actually used as far as I can see.

Maybe @baldersheim can answer whether it's possible to use it at all.

> Shouldn't it be output as WeakAnd not WAND in the (debug) string?

yes, the WandItem (which is a variant of WeightedSetItem) calls itself WAND so it's better to be able to differentiate.

baldersheim:
Agree that it should be named WeakAnd as not WAND.

I am not sure what happens if you set the index name on the weakand and just send down the items without index.
It is the oldest, and it can be used across different indexes. I guess that must be tested.

baldersheim (approved):

bratseth:
Thanks. Leaving the index for now then, and I'll create another pr to rename.

PR #17415, opened 2021-04-13
- Depending on sync of an executor that is shared is not a good idea …

baldersheim:
…inside a documentdb component.

- Just wait for what you are interested in.

@toregge or @geirst PR


toregge (approved):

toregge (approved):

geirst (commented):
👍

PR #17432, opened 2021-04-14
Log components from fallback injector as info

bjorncs:
FYI @geirst


baldersheim (approved):
Good, time for some cleanup.

PR #17448, opened 2021-04-15
Update model info from HostInfoUpdater

bjormel:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
@@ -19,18 +19,18 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 /**
- * Ensures that the switch information for all hosts is up to date.
+ * Ensures that the OpsDB information for all hosts is up to date.
  *
- * @author mpolden
+ * @author mpolden, bjormel
  */
-public class HostSwitchUpdater extends ControllerMaintainer {
+public class OpsDbInfoUpdater extends ControllerMaintainer {
 
-    private static final Logger LOG = Logger.getLogger(HostSwitchUpdater.class.getName());
+    private static final Logger LOG = Logger.getLogger(OpsDbInfoUpdater.class.getName());
     private static final Pattern HOST_PATTERN = Pattern.compile("^(proxy|cfg|controller)host(.+)$");
 
     private final NodeRepository nodeRepository;
 
-    public HostSwitchUpdater(Controller controller, Duration interval) {
+    public OpsDbInfoUpdater(Controller controller, Duration interval) {
         super(controller, interval, null, EnumSet.of(SystemName.cd, SystemName.main));
         this.nodeRepository = controller.serviceRegistry().configServer().nodeRepository();
     }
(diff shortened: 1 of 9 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
```diff
@@ -19,18 +19,18 @@
...
 
 /**
- * Ensures that the switch information for all hosts is up to date.
+ * Ensures that the OpsDB information for all hosts is up to date.
  *
- * @author mpolden
+ * @author mpolden, bjormel
  */
-public class HostSwitchUpdater extends ControllerMaintainer {
+public class OpsDbInfoUpdater extends ControllerMaintainer {
```
mpolden, line 26: Rename this to `HostInfoUpdater` to avoid leaking internal service names.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
```diff
@@ -19,18 +19,18 @@
 import java.util.stream.Collectors;
 
 /**
- * Ensures that the switch information for all hosts is up to date.
+ * Ensures that the OpsDB information for all hosts is up to date.
  *
- * @author mpolden
+ * @author mpolden, bjormel
```
mpolden, line 24: Multiple authors are typically listed on separate lines:
```
@author A
@author B
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
```diff
@@ -73,8 +74,8 @@ private static String registeredHostnameOf(Node host) {
 
     private static boolean shouldUpdate(Node node, NodeEntity nodeEntity) {
         if (nodeEntity == null) return false;
-        if (nodeEntity.switchHostname().isEmpty()) return false;
-        return !node.switchHostname().equals(nodeEntity.switchHostname());
+        if (nodeEntity.switchHostname().isEmpty() || nodeEntity.model().isEmpty())  return false;
```
mpolden, line 77: This looks wrong to me. If switch is non-empty, but model is empty we still may want to update the switch name and vice versa. 

You can create two separate methods: `shouldUpdateSwitch` and `shouldUpdateModel`. Patching can be skipped if both are `false`.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdaterTest.java
```diff
@@ -18,9 +18,9 @@
 import static org.junit.Assert.assertTrue;
 
 /**
- * @author mpolden
+ * @author mpolden, bjormel
```
mpolden, line 21: Same as above.

mpolden (changes requested):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OpsDbInfoUpdater.java
bjormel: 👍

mpolden (approved):
Great!

PR #17455, opened 2021-04-16
Remove implicit reboot of nodes moving to dirty

mpolden:
As far as I can tell this has no effect as node admin simply removes any
containers that are in dirty. The comment about "lingering processes" is from
the bare metal days.

@freva

FYI @hakonhall


freva (approved):
:100:

PR #17463, opened 2021-04-16
Disallow >1 group to suspend

hakonhall:
* If there is more than one group, disallow suspending a node if there is a node in another group that has a user wanted state != UP.
* If there is 1 group, disallow suspending more than 1 node.

In addition, we no longer allow suspension if a node is already in maintenance, as maintenance is used with grace period exactly to defer downing of node.

### DIFF
--- a/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
+++ b/clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
@@ -4,6 +4,7 @@ package com.yahoo.vespa.clustercontroller.core;
 import com.yahoo.vdslib.distribution.ConfiguredNode;
 import com.yahoo.vdslib.distribution.Distribution;
 import com.yahoo.vdslib.distribution.Group;
+import com.yahoo.vdslib.distribution.GroupVisitor;
 import com.yahoo.vdslib.state.ClusterState;
 import com.yahoo.vdslib.state.Node;
 import com.yahoo.vdslib.state.NodeState;
@@ -11,6 +12,7 @@ import com.yahoo.vdslib.state.NodeType;
 import com.yahoo.vdslib.state.State;
 import com.yahoo.vespa.clustercontroller.core.hostinfo.HostInfo;
 import com.yahoo.vespa.clustercontroller.utils.staterestapi.requests.SetUnitStateRequest;
+import com.yahoo.vespa.config.content.StorDistributionConfig;
 import org.junit.Test;
 
 import java.text.ParseException;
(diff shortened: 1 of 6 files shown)


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
```diff
@@ -139,6 +146,158 @@ public void testUnknownStorageNode() {
...
+        assertThat(result.getReason(), is("At most one node can have a wanted state when #groups = 1: " +
+                "Other storage node 0 has wanted state Maintenance"));
+    }
+
+    @Test
+    public void testWhenOtherDistributorIsDown() {
+        // Nodes 0-3, storage node 0 being in maintenance with "Orchestrator" description.
+        ContentCluster cluster = createCluster(createNodes(4));
+        cluster.clusterInfo().getDistributorNodeInfo(0)
+                .setWantedState(new NodeState(NodeType.DISTRIBUTOR, State.DOWN).setDescription("Orchestrator"));
```
vekterli, line 174: Should we test without the explicit `Orchestrator` reason for at least one test? I seem to recall we have some special logic for that particular string.

vekterli (approved):
:+1:

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
hakonhall: "Orchestrator" isn't hard-coded in the cluster controller, but the rule "allow suspending if another node in the group is suspended" finds Maintenance nodes with the same description, hence would find "Orchestrator" Maintenance storage nodes.  This match on the description is tested in testDifferentDescriptionImpliesDenied.

The description is set to Orchestrator here mostly for documentation purposes - I could remove it or use a random description.

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
```diff
@@ -139,6 +146,158 @@ public void testUnknownStorageNode() {
         assertThat(result.getReason(), is("Unknown node storage.10"));
     }
 
+    @Test
+    public void testWhenOtherStorageNodeIsSuspended() {
```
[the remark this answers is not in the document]
hakonhall: Test names updated.

vekterli (approved):
:+1:

hmusum:
Looks like system test https://github.com/vespa-engine/system-test/blob/master/tests/vds/teststatetools.rb started failing after this PR, might be that test just needs to be updated, haven't looked closely at it.

hakonhall:
Message needs to be updated, fix in https://github.com/vespa-engine/system-test/pull/1496.

PR #17465, opened 2021-04-16
Improve coverage message

bratseth:


baldersheim (approved):

hmusum:
System test in https://github.com/vespa-engine/system-test/blob/master/tests/search/uselocalnode/uselocalnode.rb needs to be updated, since it checks log message

PR #17469, opened 2021-04-16
Cache nodes based on zknode version

mpolden:
Reduces reads, avoids potential hash collisions and reduces evictions. Same
pattern is already used in controller-server (see #17003).

@jonmv


jonmv (approved):
Nice :)

PR #17471, opened 2021-04-16
Balder/reduce netty default buffer

baldersheim:
@jonmv or @hmusum or @bjorncs or @geirst PR


hmusum (approved):

jonmv:
The best

bjorncs (commented):
Should be do the same for all Zookeeper instances (anything using zkfacade or zookeeper-server artifacts)?

baldersheim:
Yes, probably. Where would that be ?

jonmv:
`16Mb` isn't usually as significant as here. What are our other clients? The `vespa-zkcli` family, and the config servers?

baldersheim:
16 is not significant, but 16*N might be. It seems like additional chunks will get allocated with the same size in a thread local cache, which might give you #NioEventLoops * 16

jonmv:
It's surprising  if many such threads are created, though?

jonmv:
It could also be that 16Mb is more suitable for our other clients. The config server and controller zk usage is much heftier.

baldersheim:
CC has at least 6 for zookeeper and 3 for zkfacade

bjorncs:
ZK client config is created by `ZkClientConfigBuilder`. There is an similar class for server-side in zkfacade.

PR #17474, opened 2021-04-16
Revert "Cache nodes based on zknode version"

jonmv:
Reverts vespa-engine/vespa#17469

@mpolden close but no cigar. I wonder if it's the difference in liveness of the stat vs. the actual data?


jonmv:
Hmm, yes: 
1. read session at session counter S1
2. some node which is cached in the session changes, and stat-version and session counter are both increased
3. an updated stat for the node is read, but the data is old, belonging to S1—this old data is then cached with the new stat
4. read session at session counter S2—this invalidates the session cache
5. the same stat is read for the node, and the cached, now outdated, node is used

PR #17477, opened 2021-04-17
Cache nodes and children based on ZNode versions

mpolden:
This replaces both the current object cache and the byte-level cache.

Compared to the current object cache this:

* avoids hashing and thus hash collisions
* reduces reads as we don't need to read the znode for hashing purposes
* reduces redundant entries as we don't keep multiple versions of the same entry
  in the cache

Compared to the byte-level cache this:

* does not throw away the entire cache on writes
* only caches Node and path children, but these types dominate reads. Additional
  caches can be added per type, if necessary.

Rationale:

* a single cache mechanism
* the hit rate for the current object cache is high (typically 95-99%) as node objects don't
  change that often
* reduces reads because the object cache doesn't need to read data to determine
  if a cached entry is stale and invalidates data per path, not globally


mpolden:
Will revisit this somewhere down the line. Need a good way to measure it, and probably a feature flag.

PR #17480, opened 2021-04-19
add more logging

arnej27959:
* make it possible to track which location broker
  various components are registered with or
  mirroring information from.
* most of these log messages should now look the same
  in the Java and C++ implementations

@hakonhall please review
@vekterli please review
@havardpe FYI

### DIFF
--- a/jrt/src/com/yahoo/jrt/slobrok/api/Register.java
+++ b/jrt/src/com/yahoo/jrt/slobrok/api/Register.java
@@ -181,6 +181,7 @@ public class Register {
             reqDone = false;
 
             boolean logOnSuccess = false;
+            boolean logOnFailure = true;
             synchronized (this) {
                 if (req.methodName().equals(UNREGISTER_METHOD_NAME)) {
                     logOnSuccess = true;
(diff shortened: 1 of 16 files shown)


hakonhall (approved):

slobrok/src/vespa/slobrok/sbmirror.cpp
```diff
@@ -256,10 +263,12 @@ MirrorAPI::handleReconnect()
             }
             double delay = _backOff.get();
             reSched(delay);
+            std::string cps = _slobrokSpecs.logString();
+            static const char * const msgfmt = "no location brokers available, retrying: %s (in %.1f seconds)";
```
vekterli, line 267: What is the rationale behind having this as `static` versus just being a regular const var? (also applies to usage in `sbregister.cpp`)

hakonhall (approved):

jrt/src/com/yahoo/jrt/slobrok/api/Register.java
```diff
@@ -191,16 +192,21 @@ private void handleUpdate() {
...
                     final Boolean lastSucceeded = lastRegisterSucceeded.get(name);
-                    if (lastSucceeded == null || lastSucceeded != !req.isError()) {
+                    if (lastSucceeded == null) {
+                        logOnSuccess = true;
+                        logOnFailure = false;
+                    } else if (lastSucceeded != !req.isError()) {
                         logOnSuccess = true;
-                        lastRegisterSucceeded.put(name, !req.isError());
                     }
+                    lastRegisterSucceeded.put(name, !req.isError());
```
vekterli, line 201: This method is now called unconditionally wrt. `lastSucceeded`, can this have any unintended side effects?

vekterli (approved):
:+1: just a couple of minor questions

slobrok/src/vespa/slobrok/sbmirror.cpp
arnej27959: > What is the rationale behind having this as static versus just being a regular const var?
I was worried about the compiler wanting to check the format string, but you're right it doesn't need to be static.

jrt/src/com/yahoo/jrt/slobrok/api/Register.java
arnej27959: > This method is now called unconditionally wrt. lastSucceeded, can this have any unintended side effects?

no - if it already had the value from before, we will overwrite with same value, to avoid duplicating this line.

PR #17482, opened 2021-04-19
Consider casing when setting up enum store.

baldersheim:
@toregge PR


toregge (approved):
Note that there are places during loading, updating and search that uses make_folded_comparator() or is_folded_change(). That comparator needs corresponding adjustments.

baldersheim:
I will close this one now as it need to mature a bit.
Smaller preparations will follow.

PR #17490, opened 2021-04-19
A tensor field that specifies hnsw index parameters must also specify…

geirst:
… 'index' in 'indexing'.

@bratseth please review
@jobergum FYI


bratseth (approved):
👍

hmusum:
You're sure there are no existing applications with this problem? We would fail to deploy an existing app with this issue when upgrading Vespa in such a system and the upgrade would fail.

geirst:
No, we are not sure about that. But without this validation, you can deploy an application thinking you are using an hnsw index when you are not. So it is better to fail at deploy time so the application can be fixed.

kkraune:
I will have a go at existing apps

hmusum:
Yes, I agree that this needs to be validated at deploy time. The problem is that if there are existing apps with this issue the rollout to production will be blocked and we cannot get new releases out until the customer has fixed the application or we have rolled back infrastructure to previous version manually. 

There is an internal ticket 'Allow for "warn-only" config validations' that is related to this.

PR #17492, opened 2021-04-19
Double timeout waiting for activation on config servers to 120 seconds

hmusum:
Seeing long activation times in one zone, double it to see if it helps.


jonmv (approved):
What would make activation this slow, though?

hmusum:
Well, it's to hard to see why, that's part of the problem, need to look into that as well

PR #17513, opened 2021-04-21
turn up the logging level

arnej27959:
* these tools are also in the category where users
  will consider any INFO level message noise.
  Since they don't use the vespa LogSetup, just
  increase the threshold programmatically.

@vekterli please review
@baldersheim @bjorncs FYI


baldersheim (approved):

vekterli (approved):
:+1:

bjorncs (approved):

PR #17518, opened 2021-04-21
Just return not found if application not loaded

hmusum:


hmusum:
ping @jonmv

jonmv:
Ah, yes, forgot about this one. So this will cause immediate exceptions until someone else calls `ensureApplicationLoaded()`, which _shouldn't_ be a problem, right?

jonmv (approved):

hmusum:
Right. We'll have to see how this works out in practice, though.

PR #17521, opened 2021-04-21
Add heart beat age metric.

toregge:
@baldersheim : please review

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/metrics/documentdb_tagged_metrics.cpp
+++ b/searchcore/src/vespa/searchcore/proton/metrics/documentdb_tagged_metrics.cpp
@@ -272,6 +272,7 @@ DocumentDBTaggedMetrics::DocumentDBTaggedMetrics(const vespalib::string &docType
       bucketMove(this),
       totalMemoryUsage(this),
       totalDiskUsage("disk_usage", {}, "The total disk usage (in bytes) for this document db", this),
+      heart_beat_age("heart_beat_age", {}, "How long ago heart beat maintenace job was run", this),
       maxNumThreads(maxNumThreads_)
 {
 }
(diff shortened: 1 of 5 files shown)


searchcore/src/vespa/searchcore/proton/server/feedhandler.h
```diff
@@ -96,6 +96,8 @@ class FeedHandler: private search::transactionlog::client::Callback,
     std::mutex                             _syncLock;
     SerialNum                              _syncedSerialNum; 
     bool                                   _allowSync; // Sanity check
+    mutable std::mutex                     _heart_beat_time_lock;
+    vespalib::steady_time                  _heart_beat_time;
```
baldersheim, line 100: Can't you make this an atomic to avoid the separate extra lock ?

baldersheim (changes requested):
Consider using an atomic instead

searchcore/src/vespa/searchcore/proton/metrics/documentdb_tagged_metrics.cpp
```diff
@@ -272,6 +272,7 @@ DocumentDBTaggedMetrics::DocumentDBTaggedMetrics(const vespalib::string &docType
       bucketMove(this),
       totalMemoryUsage(this),
       totalDiskUsage("disk_usage", {}, "The total disk usage (in bytes) for this document db", this),
+      heart_beat_age("heart_beat_age", {}, "How long ago heart beat maintenace job was run", this),
```
geirst, line 275: Please add info on unit used: "ago (in seconds) heart"

baldersheim (approved):
So steady_time can not be atomic then ?

searchcore/src/vespa/searchcore/proton/server/feedhandler.h
```diff
@@ -96,6 +96,7 @@ class FeedHandler: private search::transactionlog::client::Callback,
     std::mutex                             _syncLock;
     SerialNum                              _syncedSerialNum; 
     bool                                   _allowSync; // Sanity check
+    std::atomic<uint64_t>                  _heart_beat_time;
```
baldersheim, line 100: I do see that we have test for atomic time.
static_assert(std::atomic<steady_time>::is_always_lock_free, "std::atomic<steady_time> should be atomic");
Then I suggest that you use std::atomic<steady_time>. Then you avoid all the casting.

baldersheim (approved):
Perfect

PR #17524, opened 2021-04-21
More generic error response exception

jonmv:
@bjorncs please review.


jonmv:
Or @hakonhall 
The idea is that users of this interface will defined their own error handling, and wrap calls to this client in helper code that sets a default error response mapper, where applicable.

bjorncs (approved):

jonmv:
Hmm, there are some missing building blocks here, though.

hakonhall (approved):

jonmv:
Please take a new look. I think this is better. Gives separation between the different concerns, at the cost of being able to return normally from an IOException, which I think is a niche use case anyway.

hakonhall (approved):

jonmv:
New PR ready, but may as well merge this, as most of these changes are still live ;)

PR #17525, opened 2021-04-21
Log output of nodectl subcommands

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/ContainerOperationsImpl.java
```diff
@@ -243,28 +243,28 @@ public CommandResult executeCommandInNetworkNamespace(NodeAgentContext context,
     }
 
     @Override
-    public void resumeNode(NodeAgentContext context) {
-        executeNodeCtlInContainer(context, "resume");
+    public String resumeNode(NodeAgentContext context) {
+        return executeNodeCtlInContainer(context, "resume").getOutput();
```
freva, line 247: Return `Optional<String>` from here and `executeNodeCtlInContainer()` so we dont need to check `isBlank()` everywhere?

freva (approved):
Why not just log in `ContainerOperationsImpl`/`Podman`? They already have the `context` and it is there that you know which script is actually being executed.

hakonhall:
I thought that was at a too low level. Much logging is done at the NodeAgentImpl level.  I think it makes it a bit easier to verify correct and consistent logging, and makes it easier to combine log messages or add additional context that isn't known at a lower level.  For instance the restart services log message is able to add the reason.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/docker/ContainerOperationsImpl.java
hakonhall: The only way to avoid testing for isBlank() everywhere would be to add a method that returns Optional<String> from e.g. ProcessResult.  I don't know -- the output of a program is best modelled as String, not Optional<String> IMO.  It's the invoker of the particular program that knows how to interpret the output, and in these cases and for now I'm thinking isBlank => ignore, but that may change once this rolls out.

PR #17528, opened 2021-04-21
Remove support for non-dedicated cc's on hosted

bratseth:


config-model/src/main/java/com/yahoo/config/model/provision/InMemoryProvisioner.java
```diff
@@ -238,6 +239,14 @@ private HostSpec retire(HostSpec host) {
...
     }
 
+    private void dumpFreeNodes(String description) {
+        System.out.println("  Free nodes " + description + ":");
+        for (var entry : freeNodes.entrySet()) {
+            for (var node : entry.getValue())
+                System.out.println("    " + node.hostname());
+        }
+    }
+
```
jonmv, line 249: Leftover?

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
```diff
@@ -293,26 +293,8 @@ private void addClusterControllers(ConfigModelContext context,
             if (contentCluster.getPersistence() == null) return;
 
             ClusterControllerContainerCluster clusterControllers;
-
-            ContentCluster overlappingCluster = findOverlappingCluster(context.getParentProducer().getRoot(), contentCluster);
-
-            if (overlappingCluster != null && overlappingCluster.getClusterControllers() != null) {
-                // Borrow the cluster controllers of the other cluster in this case.
-                // This condition only occurs on non-hosted systems with a shared config server,
-                // a combination which only exists in system tests.
```
jonmv, line 302: You dare challenge this comment?

jonmv: Ah, never mind, I'm just bad at reading.

jonmv (approved):
Great!

config-model/src/main/java/com/yahoo/config/model/provision/InMemoryProvisioner.java
jonmv, line 249:
```suggestion
```

bratseth: Thought I'm going to need it later, but ok :-)

PR #17529, opened 2021-04-21
Initial implementation and wiring of cross-stripe state and DB handling

vekterli:
@geirst please review

Introduces the concept of stripe access guards, that ensure safe and
non-concurrent access to the underlying state of all running distributor
stripes. Also bring back a top-level `BucketDBUpdater` component responsible
for managing cluster state/distribution config and all related bucket info
fetching for the entire node as a whole. This component abstracts away all
stripe-specific operations via the new guard interface.

For now, only a single stripe can be used via the new code path, and by
default the legacy code path (single stripe acts as an entire distirbutor)
is used. New path may be enabled via (non-live) config, but is not yet
production ready.


geirst (approved):
👍

PR #17531, opened 2021-04-21
No restart annotation here, as it does not work for attributes, and i…

baldersheim:
…t is already handled in the model.

@toregge or @geirst PR


toregge (approved):

geirst (commented):
👍

PR #17538, opened 2021-04-22
Wire in tensor cell type resolving for concat in Java

lesters:
@arnej27959 Please review. Like with `join` - I modified the rules a bit in `TypeResolver` to be conservative and fit existing behavior regarding unbound and mapped dimension.

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/functions/Concat.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/functions/Concat.java
@@ -7,6 +7,7 @@ import com.yahoo.tensor.IndexedTensor;
 import com.yahoo.tensor.Tensor;
 import com.yahoo.tensor.TensorAddress;
 import com.yahoo.tensor.TensorType;
+import com.yahoo.tensor.TypeResolver;
 import com.yahoo.tensor.evaluation.EvaluationContext;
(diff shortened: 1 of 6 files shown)


vespajlib/src/main/java/com/yahoo/tensor/functions/Concat.java
```diff
@@ -97,7 +75,7 @@ public Tensor evaluate(EvaluationContext<NAMETYPE> context) {
         IndexedTensor aIndexed = (IndexedTensor) a; // If you get an exception here you have implemented a mixed tensor
         IndexedTensor bIndexed = (IndexedTensor) b;
 
-        TensorType concatType = type(a.type(), b.type());
+        TensorType concatType = TypeResolver.concat(a.type(), b.type(), dimension);
```
arnej27959, line 78: it looks like this should be some lines earlier, and then compute `combinedValueType` from `concatType`

arnej27959 (approved):
one minor worry from me, otherwise looks good

vespajlib/src/main/java/com/yahoo/tensor/functions/Concat.java
lesters: Thanks!

PR #17542, opened 2021-04-22
Arnej/resolve bindings via parent context

arnej27959:

it looks like we have to handle resolving of arguments to rank features, like `foo` in `attribute(foo)` from the "outside".  I'm not  100% sure if the fix here is complete.

@bratseth please review
@lesters FYI


bratseth (approved):
👍

PR #17546, opened 2021-04-22
Revert "Log components from fallback injector as info"

bjorncs:
Reverts vespa-engine/vespa#17432


baldersheim (changes requested):
I think we should just leave it as it is. If there is anything we should do it would be to silence our own, which are actually correct. This is the only way the users can figure out that there is something wrong.

PR #17549, opened 2021-04-22
Provide standard mapping

tokle:


container-core/src/main/java/com/yahoo/container/jdisc/HttpMethodAclMapping.java
```diff
@@ -0,0 +1,71 @@
...
+import static com.yahoo.jdisc.http.HttpRequest.Method.OPTIONS;
+import static com.yahoo.jdisc.http.HttpRequest.Method.PATCH;
+import static com.yahoo.jdisc.http.HttpRequest.Method.POST;
+import static com.yahoo.jdisc.http.HttpRequest.Method.PUT;
+import static com.yahoo.jdisc.http.HttpRequest.Method.TRACE;
+
+/**
+ * Acl Mapping based on http method.
+ * Defaults to:
+ * {GET, HEAD, OPTIONS} -> READ
```
bjorncs, line 25: Should `{` and `}` be escaped?

container-core/src/main/java/com/yahoo/container/jdisc/RequestHandlerSpec.java
```diff
@@ -31,7 +31,7 @@ public static Builder builder(){
 
     public static class Builder {
 
-        private AclMapping aclMapping = new DefaultAclMapping();
+        private AclMapping aclMapping = HttpMethodAclMapping.standard().build();
```
bjorncs, line 34: Consider having a shorthand for `.standard().build()`.

bjorncs (approved):

PR #17551, opened 2021-04-22
- Prevent the rewrite that happens with equiv and phrase to handle fl…

baldersheim:
…oating point queries in string fields, from happening in under a same-element node.

- Add raw stackdump from previously failing query.

@geirst PR


geirst (approved):
👍

PR #17556, opened 2021-04-23
Make sure we are able to output context for validation errors

hmusum:
Not ideal to read the content twice, but these files are usually very small, so I think this is OK


frodelu (approved):
Looks good - will be really helpful with the (post-processed) line numbers as part of context!

PR #17540, opened 2021-04-22
Support both case sensitive, and case-insensitive search in non-fast-…

baldersheim:
…search attributes.

@toregge or @geirst PR

PR #17558, opened 2021-04-23
Add testing of too short strings.

baldersheim:
@geirst PR Follow up from #17540

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/stringbase.h
+++ b/searchlib/src/vespa/searchlib/attribute/stringbase.h
@@ -16,6 +16,10 @@
 
 namespace search {
 
+/**
+ * helper class for searchcontext when doing scanning of string fields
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/attribute/stringbase.h
```diff
@@ -16,6 +16,10 @@
 
 namespace search {
 
+/**
+ * helper class for searchcontext when doing scanning of string fields
+ * It handles differend search settings like prefix and cased/uncased and regex.
```
geirst, line 21: differend -> different

geirst (approved):
👍

searchlib/src/vespa/searchlib/attribute/stringbase.h
baldersheim: Fixed

PR #17562, opened 2021-04-23
Make DistributorStripe aware of whether it uses legacy mode or not an…

geirst:
…d add asserts.

@vekterli please review


vekterli (approved):
:+1:

PR #17566, opened 2021-04-23
Flag to allow disabling mtls

tokle:


configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -392,6 +396,10 @@ public String athenzDnsSuffix() {
             return flagValueForClusterType(jvmGCOptionsFlag, clusterType);
         }
 
+        @Override
+        public boolean allowDisableMtls() {
+            return allowDisableMtls.value();
```
bjorncs, line 401: I suspect you should extract the flag value in the constructor to ensure a consistent value (in case flag is flipped while retrieving config).

bjorncs (approved):

bjorncs (approved):

PR #17569, opened 2021-04-23
Propagate distributor config via internal snapshots [run-systemtest]

vekterli:
@geirst please review.

Add functionality for updating config for stripes via accessor guard
interface. Stripes will now keep an explicit, immutable config snapshot
internally (and get explicit knowledge of when this changes) instead
of accessing config via magical component interface.

Introduce notion of internal config generation (not directly related
to generations received from underlying config system) to be able to
cheaply know if new config should be propagated. Since a lot of unit
tests use dirty "modify in place" tricks to alter config for tests,
have to do some extra work per tick to ensure non-generational config
changes are applied to the distributor code. This should be fixed as
a later follow-up.

Also remove `TickingThread` inheritance from `DistributorStripe` to begin
preparing for redesign of stripe threading model.

### DIFF
--- a/storage/src/vespa/storage/distributor/distributor_stripe.h
+++ b/storage/src/vespa/storage/distributor/distributor_stripe.h
@@ -48,7 +48,6 @@ class DistributorStripe final
     : public DistributorStripeInterface,
       public StatusDelegator,
       public framework::StatusReporter,
-      public framework::TickingThread,
       public MinReplicaProvider,
       public BucketSpacesStatsProvider,
       public NonTrackingMessageSender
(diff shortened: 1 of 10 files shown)


storage/src/vespa/storage/distributor/distributor_stripe.h
```diff
@@ -151,7 +150,7 @@ class DistributorStripe final
     }
 
     const DistributorConfiguration& getConfig() const override {
-        return _component.getTotalDistributorConfig();
+        return *_total_config;
```
geirst, line 153: Have you verified that no of the sub components of the stripe calls _component.getTotalDistributorConfig()_?

geirst (approved):
👍

PR #17586, opened 2021-04-25
- Add non-folded prefix.

baldersheim:
- Make methods in stateless FoldedStringCompare static.

@toregge PR


toregge (approved):

toregge:
Note that a dictionary containing [ "AB", "ab", "AC", "ac" ] will fail a lookup for "a" as a case sensitive prefix. The matching entries
do not represent a range in the dictionary.  For case sensitive prefix to work, a different sort order is needed, e.g. [ "AB", "AC", "ab", "ac" ].

PR #17600, opened 2021-04-26
Expose feeding blocked metric to controller

freva:

This metric will be used to create feeding blocked notification in next PR.


jonmv (approved):

configserver/src/test/resources/metrics/content_metrics.json
```diff
@@ -0,0 +1,36 @@
...
+    },
+    {
+      "name": "vespa.searchnode",
+      "timestamp": 1557306075,
+      "metrics": [
+        {
+          "values": {
+            "content.proton.resource_usage.disk.average": 0.6716298061329,
+            "content.proton.resource_usage.memory.average": 0.7172795340244,
+            "content.proton.resource_usage.feeding_blocked.last": 0
```
geirst, line 26: Are you sure this is the only metric we want to expose? This represents the hard blocking limit inside proton, while external feed is blocked earlier by the cluster controller. When the following metric is larger than 0 external feed is blocked: "vespa.container-clustercontroller.cluster-controller.resource_usage.nodes_above_limit.last". More details in https://docs.vespa.ai/en/reads-and-writes.html#feed-block.

freva: Thanks, I'll change to use that metric instead in next PR

PR #17605, opened 2021-04-26
Remove processing of single bucket info replies.

geirst:
These are per stripe and handled by StripeBucketDBUpdater instead.

@vekterli please review


vekterli (approved):
:+1:

PR #17606, opened 2021-04-26
Use a single-threaded executor that delegates to another for ZK sessi…

hmusum:
…on work

Use a single-threaded executor to make sure that events are handled in order,
but delegate to a multi-threaded executor for the real work.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantRepository.java
@@ -645,6 +645,12 @@ public class TenantRepository {
             return executorService.shutdownNow();
         }
 
+        @Override
+        public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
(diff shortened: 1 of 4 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantRepository.java
```diff
@@ -612,4 +615,36 @@ public static Path getBarriersPath() {
...
+        public void shutdown() {
+            super.shutdown();
+            executorService.shutdown();
+        }
+
+        @Override
+        public List<Runnable> shutdownNow() {
+            super.shutdownNow();
+            return executorService.shutdownNow();
+        }
```
jonmv, line 646: Could also delegate `awaitTermination` to both parties.

jonmv (approved):
Exciting times :)

configserver/src/main/java/com/yahoo/vespa/config/server/tenant/TenantRepository.java
hmusum: Yup, will add

PR #17608, opened 2021-04-26
Remove deployment failure notification on running

freva:
Looks like the `success` status is set first after `report`, so we should just clear the notifications while in `running`


jonmv (commented):
Yes, this is correct.

PR #17609, opened 2021-04-26
Set a max scaling duration

bratseth:
If the autoscaling maintainer is disabled it won't record completion.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterModel.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterModel.java
@@ -150,7 +150,8 @@ public class ClusterModel {
         for (ScalingEvent event : cluster.scalingEvents()) {
             if (event.duration().isEmpty()) continue;
             completedEventCount++;
-            totalDuration = totalDuration.plus(event.duration().get());
+            // Assume we have missed timely recording completion if it is longer than 4 days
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterModel.java
```diff
@@ -150,7 +150,8 @@ private static Duration computeScalingDuration(Cluster cluster, ClusterSpec clus
         for (ScalingEvent event : cluster.scalingEvents()) {
             if (event.duration().isEmpty()) continue;
             completedEventCount++;
-            totalDuration = totalDuration.plus(event.duration().get());
+            // Assume we have missed timely recording completion if it is longer than 4 days
+            totalDuration = totalDuration.plus(maximum(Duration.ofDays(4), event.duration().get()));
```
freva, line 154: This should be minimum?

bratseth: The duration shall be maximum 4 days. Isn't that clear?

freva: Uhh, I guess. I thought this worked same as `Math.max()`, i.e. it returned the max of the 2 args, but this depends on the position of args and duration-minus-isNegative isn't the easiest thing to read.

freva (approved):

PR #17610, opened 2021-04-26
In the case that the first item in a rank item evaluated to false it …

baldersheim:
…would be removed, amking the next the first.

That is correct behaviour for all, but the first item. If the first is false, the whole rank will be false.

@bratseth PR


bratseth (approved):
👍  should've called it AndRankItem

PR #17612, opened 2021-04-27
Change the way we switch connection when the current one is not working

hmusum:
Consider only healthy sources (if there are any) when switching to
a new one due to failures.
Also log when connecting to a source.

### DIFF
--- a/config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
+++ b/config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
@@ -19,7 +19,7 @@ import java.util.stream.Collectors;
  * The current connection is chosen randomly when calling {#link {@link #switchConnection()}}
  * (it will continue to use the same connection if there is only one source).
  * The current connection is available with {@link #getCurrent()}.
- * When calling {@link #setError(Connection, int)}, {#link {@link #switchConnection()}} will always be called.
+ * When calling {@link #setError(Connection, int)}, {@link #switchConnection()} will always be called.
  *
  * @author Gunnar Gauslaa Bergem
  * @author hmusum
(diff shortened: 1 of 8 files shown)


config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
```diff
@@ -67,20 +65,34 @@ public synchronized JRTConnection getCurrent() {
...
-     * @return a JRTConnection
-     */
-    public synchronized JRTConnection setNewCurrentConnection() {
+    @Override
+    public synchronized JRTConnection switchConnection() {
         List<JRTConnection> sources = getSources();
-        currentConnection = sources.get(ThreadLocalRandom.current().nextInt(0, sources.size()));
+        if (sources.size() <= 1) return currentConnection;
+
+        List<JRTConnection> healthySources = sources.stream()
```
hakonhall, line 73: This selects the healthy connections.  But isn't the healthiness of a connection only updated when it is "current"?  And so what may happen is:
* Say there are 3 members A, B, and C.  They start out as healthy, and A is chosen to be current.
* Say A becomes unhealhty, and then a random of B and C is chosen, say B.
* While B is the current connection, A has no chance of changing its healthiness so it stays unhealthy.
* When B becomes unhealthy, C is chosen.
* When C becomes unhealthy, C is kept per below test on `size() == 0`

hmusum: Yes, you are right. We could either start doing ping out-of-band or try an unhealthy connection after some time (if there is not any healthy ones to choose from). Will look into it.

hakonhall: setNewCurrentConnection() is only called on creation or when exactly 1 connection is bad.  So a simple method might be to select one randomly, or select randomly except that one connection, respectively?

hmusum: Thanks, I have implemented a suggestion with some test

hakonhall (approved):

PR #17619, opened 2021-04-27
Control use of fsync in services.xml.

baldersheim:
@bratseth and @geirst PR and @vekterli PR
@jobergum FYI

Note that this introduces a new tag in services.xml 'sync-transactionlog' under content.engine.proton.
Perhaps it needs an api review.


bratseth (approved):
Yes, we should talk about it on Thursday. LGTM to merge before then though (we can doc it after approval).

PR #17622, opened 2021-04-27
Double check unfolded if search is cased.

baldersheim:
@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
+++ b/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
@@ -180,18 +180,11 @@ class StringPostingSearchContext
     : public PostingSearchContext<BaseSC, PostingListFoldedSearchContextT<DataT>, AttrT>
 {
 private:
-    using AggregationTraits = PostingListTraits<DataT>;
-    using PostingList = typename AggregationTraits::PostingList;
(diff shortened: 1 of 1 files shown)


searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
```diff
@@ -300,6 +290,18 @@ StringPostingSearchContext(QueryTermSimpleUP qTerm, bool useBitVector, const Att
     }
 }
 
+template <typename BaseSC, typename AttrT, typename DataT>
+bool
+StringPostingSearchContext<BaseSC, AttrT, DataT>::useThis(const PostingListSearchContext::DictionaryConstIterator & it) const {
+    if ( this->isCased() ) {
+        this->isMatch(_enumStore.get_value(it.getKey()));
```
toregge, line 297: What about cased regex ?
Also, the return value from this->isMatch() is ignored.

toregge (changes requested):

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
baldersheim: Planned to take regex in separate round, but after checking code regex case sensitivity is handler construction time.
So reordering the conditions should take care of this.

toregge (approved):

PR #17625, opened 2021-04-27
Allow 128 connections per config server and 1024 total

jonmv:
@freva please review. @bjorncs FYI. 8 was not enough in CD.


jonmv:
Rather liberal now, and adjust down later.

freva (approved):

PR #17626, opened 2021-04-28
Switch connection only when the current one is failing [run-systemtest]

hmusum:
Subscribers set error on a connection if it fails, so when having
many subscribers we should not switch unless the current connection
is having errors.

### DIFF
--- a/config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
+++ b/config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
@@ -12,7 +12,6 @@ import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 /**
  * A pool of JRT connections to a config source (either a config server or a config proxy).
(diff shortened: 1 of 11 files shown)


config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
```diff
@@ -66,19 +66,31 @@ public synchronized JRTConnection getCurrent() {
...
-    public synchronized JRTConnection switchConnection() {
+    public synchronized JRTConnection switchConnection(Connection failingConnection) {
         List<JRTConnection> sources = getSources();
         if (sources.size() <= 1) return currentConnection;
 
-        List<JRTConnection> sourceCandidates = sources.stream()
-                                                    .filter(JRTConnection::isHealthy)
-                                                    .collect(Collectors.toList());
-        JRTConnection newConnection;
+        if ( ! currentConnection.equals(failingConnection)) return currentConnection;
```
hakonhall, line 73: Override `equals()` in JRTConnection or use `==` ?

config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
```diff
@@ -66,19 +66,31 @@ public synchronized JRTConnection getCurrent() {
     }
 
     @Override
-    public synchronized JRTConnection switchConnection() {
+    public synchronized JRTConnection switchConnection(Connection failingConnection) {
```
hakonhall, line 69: An alternative strategy here may be to:
* Return current connection if unequal to failingConnection, or
* return a random connection != failingConnection.

There would then be no need for isHealthy, or the 3 `last` timestamps.

hakonhall (approved):

config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
hmusum: Good suggestion, will implement

config/src/main/java/com/yahoo/vespa/config/JRTConnectionPool.java
hmusum: Yup, thanks, will do

hakonhall (approved):

PR #17627, opened 2021-04-28
Serialise LATER document updates only when they will be logged

jonmv:
@bratseth please review and merge :)


bratseth (approved):
Holy fsck

PR #17629, opened 2021-04-28
Prefer NodeList operations where possible

mpolden:
@bratseth


bratseth (approved):
:ok_hand:

Let's merge tomorrow though

PR #17637, opened 2021-04-28
Use transactions when moving nodes

mpolden:
@bratseth


bratseth (approved):
Great!

An aside regarding all those overloaded state transition methods - there are so many combinations of arguments we want to pass, or not, for each state change that I think it's better to move to an API that is only moveTo(State, ...) with overloads for all the argument combinations, consolidates to one argument set, and then switches into transition-specific logic.

mpolden:
>  there are so many combinations of arguments we want to pass, or not, for each state change

Not sure I understand. Any examples? Right now it's only `keepAllocation`.

bratseth:
Check out e.g fail(...) or deallocate(...).
I want to remove those from the API and only provide move(...).

PR #17641, opened 2021-04-28
Use owner's dimensions in BackedOverridableQueryProfile

bratseth:
Use owner's dimensions rather than the backed's in
BackedObverridableQueryProfile. This matters when a profile
references a profile with different but overlapping dimensions
and both the owner and referred profile assigns values for
the same paths, within their respective dimension spaces.


arnej27959 (approved):
looks good to me

PR #17645, opened 2021-04-28
Use yield instead of pause on arm.

toregge:
@baldersheim : please review


baldersheim (changes requested):
Use std::this_thread::yield instead of all the ifdefs.
https://en.cppreference.com/w/cpp/thread/yield

baldersheim (approved):

PR #17646, opened 2021-04-28
Add Slobroks to admin cluster controllers

bratseth:


hmusum (approved):
Great! This got pretty simple and I can't see any obvious pitfalls regarding topology or service indexes either...

PR #17657, opened 2021-04-29
allow a string (with a hex dump of binary representation) as cell values

arnej27959:

Here is a proposal for new allowed input format for tensor values, which could be particularly useful for int8 and bfloat16 types. Instead of  `[16,17,127,-128,-1,0,...]`  this allows `"10117f80ff00..."`

@lesters @bratseth @havardpe what do you think?

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
@@ -365,6 +365,8 @@ public interface Tensor {
     }
 
     static boolean approxEquals(double x, double y, double tolerance) {
+        if (x == y) return true;
+        if (Double.isNaN(x) && Double.isNaN(y)) return true;
         return Math.abs(x-y) < tolerance;
     }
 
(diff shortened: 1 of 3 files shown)


havardpe (commented):
we need something like this for int8, the question is whether this is also a good thing for other cell types or not... 
should this be allowed for other things than 'int8'?
should the hexdump be called something else than 'values'?

bratseth:
I think it's good like this, and doing it for all types should solve @jobergum complaints as well, if we have a documented way to generate this for all value types.

jobergum:
Nice, it looks like the PR has support for both double, float, bfloat16 and int8 for dense dimensions and hex is probably the most compact representation we can use with JSON?

I'm wondering for really large tensors like embeddings from bert large with 1024 floats if we can also compress (eg zstd) before hexdump serialization for JSON.

havardpe:
why not compress the json? isn't that supported by HTTP already?
base64 (6 bits per byte) would be more compact than hexdump (4 bits per byte)
another option would be to tunnel serialized (and possibly compressed) tensors, but that would be almost like using JNI in Java, opening a whole new world of worms.

jobergum:
yes, feed clients supports compression (gzip) also across more than 1 document for asynch java http based client.

PR #17658, opened 2021-04-29
Support loadbalancer/v1 and application/v2/tenant without trailing slash

jonmv:
@bjorncs please review.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/TenantHandler.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/TenantHandler.java
@@ -94,6 +94,7 @@ public class TenantHandler extends HttpHandler {
     private static BindingMatch<?> getBindingMatch(HttpRequest request) {
         return HttpConfigRequests.getBindingMatch(request,
                 "http://*/application/v2/tenant",
+                "http://*/application/v2/tenant/",
                 "http://*/application/v2/tenant/*");
     }
 
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/TenantHandler.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/TenantHandler.java
@@ -93,7 +93,7 @@ public class TenantHandler extends HttpHandler {
 
     private static BindingMatch<?> getBindingMatch(HttpRequest request) {
         return HttpConfigRequests.getBindingMatch(request,
-                "http://*/application/v2/tenant/",
+                "http://*/application/v2/tenant",
                 "http://*/application/v2/tenant/*");
     }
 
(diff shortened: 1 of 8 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/TenantHandler.java
```diff
@@ -103,7 +104,7 @@ private static boolean isGetTenantRequest(HttpRequest request) {
 
     private static boolean isListTenantsRequest(HttpRequest request) {
         return getBindingMatch(request).groupCount() == 2 &&
-                request.getUri().getPath().endsWith("/tenant/");
+                request.getUri().getPath().matches("/application/v2/tenant/?");
```
bjorncs, line 107: I'm not fond of manual routing through regular expression...

bjorncs (approved):

configserver/src/main/resources/configserver-app/services.xml
```diff
@@ -102,8 +102,7 @@
       <binding>http://*/status</binding>
     </handler>
     <handler id='com.yahoo.vespa.config.server.http.v2.TenantHandler' bundle='configserver'>
-      <binding>http://*/application/v2/tenant</binding>
-      <binding>http://*/application/v2/tenant/*</binding>
+      <binding>http://*/application/v2/tenant*</binding>
```
bjorncs, line 105: I'm surprised this pattern has not been commonly used before (instead of having the two previous bindings)

jonmv: Yes. It doesn't work well for extracting the matched groups, as those now contain the `/` in sub paths, but these bindings aren't used for that. I suspect that's the reason, though.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/LoadBalancersV1ApiHandler.java
```diff
@@ -1,53 +1,37 @@
...
-        catch (IllegalArgumentException e) {
-            return ErrorResponse.badRequest(Exceptions.toMessageString(e));
-        }
-        catch (RuntimeException e) {
-            log.log(Level.WARNING, "Unexpected error handling '" + request.getUri() + "'", e);
-            return ErrorResponse.internalServerError(Exceptions.toMessageString(e));
-        }
+    private static RestApi createRestApiDefinition(LoadBalancersV1ApiHandler self) {
+        return RestApi.builder()
+                      .addRoute(RestApi.route("/loadbalancers/v1")
```
bjorncs, line 28: You need to declare explicit exception mapping  for all exception types that are not `RestApiException` (or subclass). They will otherwise result in the default 500 response from jdisc.

bjorncs (changes requested):

bjorncs (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/LoadBalancersV1ApiHandler.java
jonmv: Thanks, will keep in mind.

PR #17660, opened 2021-04-29
Simplify calculating delay after config failures

hmusum:
When there are several subscribers to a config and 1 JRTConfiguRequester
the counter for failures might quickly increase, so back-off based
on these numbers will not be useful. Simplify by using boolean
for failures instead. Max delay for configured subscriptions will
now be 30 seconds instead of 150 seconds (which is way too much)

### DIFF
--- a/config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
+++ b/config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
@@ -44,8 +44,8 @@ public class JRTConfigRequester implements RequestWaiter {
     private static final JRTManagedConnectionPools managedPool = new JRTManagedConnectionPools();
     private static final int TRACELEVEL = 6;
     private final TimingValues timingValues;
-    private int fatalFailures = 0; // independent of transientFailures
-    private int transientFailures = 0;  // independent of fatalFailures
+    private boolean fatalFailures = false;
+    private boolean transientFailures = false;
     private final ScheduledThreadPoolExecutor scheduler;
     private Instant noApplicationWarningLogged = Instant.MIN;
     private static final Duration delayBetweenWarnings = Duration.ofSeconds(60);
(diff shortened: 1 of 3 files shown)


hmusum:
More to be done here, but I think this fixes the important issue.

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
```diff
@@ -194,7 +196,7 @@ private void handleTransientlyFailed(JRTClientConfigRequest jrtReq,
                                          JRTConfigSubscription<ConfigInstance> sub,
                                          long delay,
                                          Connection connection) {
-        transientFailures++;
+        transientFailures = true;
```
hakonhall, line 199: Should we also set fatalFailures to false here?

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
```diff
@@ -217,7 +219,7 @@ private long calculateErrorTimeout() {
     private void handleFatallyFailed(JRTClientConfigRequest jrtReq,
                                      JRTConfigSubscription<ConfigInstance> sub, long delay) {
         if (sub.getState() != ConfigSubscription.State.OPEN) return;
-        fatalFailures++;
+        fatalFailures = true;
```
hakonhall, line 222: ... and transientFailures to false here?

hakonhall (approved):

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
hmusum: Yeah, I think that makes sense, I'll do it in the next PR.

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
hmusum: Yeah, I think that makes sense, I'll do it in the next PR.

PR #17662, opened 2021-04-29
Remove explicit dependency between bucket db updater and single distributor stripe

geirst:
@vekterli please review


vekterli (approved):
:+1:

PR #17669, opened 2021-04-30
Reapply "Use exact Protobuf version as dependency"

aressem:
Reverts vespa-engine/vespa#17663


vekterli (approved):
:+1:

aressem:
Verified that it built. Added the same for the runtime deps now. Should be good to go.

vekterli (approved):

PR #17673, opened 2021-04-30
Disable vespamalloc unless architecture is x86_64

toregge:
@baldersheim : please review


baldersheim (approved):
Is there X86_64 code there ?

PR #17675, opened 2021-04-30
Avoid SSE2 built-in functions when CPU arch is not x86_64

toregge:
@baldersheim : please review


baldersheim (approved):
perhaps we should benchmark this again, to see if there still is any gain.

vekterli:
I think most implementations of `memchr` use SIMD these days. [glibc does for x86-64, at least](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/x86_64/memchr.S;h=beff2708de6a1e40de4141f94ff6fe763f041164;hb=HEAD)

PR #17681, opened 2021-04-30
Avoid intel asm when not using x86_64 platform.

toregge:
@baldersheim : please review


baldersheim (changes requested):
Here we can remove all ifdefs and just use __builtin I think.

baldersheim (approved):
Perfect :)

PR #17688, opened 2021-04-30
Do not process metrics we dont care about

freva:
This is a bit ugly, but the issue is that `Inspector::asDouble` returns 0 even if it's not valid. What happened before was that for each metrics block we would aggregate a bunch of 0s, so the end result didn't change. But with my feed blocked metric we now get metric that cluster-controller (which does not report this metric for itself, only on behalf of content nodes) has feed limit 0 and util 0.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/metrics/ClusterDeploymentMetricsRetriever.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/metrics/ClusterDeploymentMetricsRetriever.java
@@ -165,7 +165,7 @@ public class ClusterDeploymentMetricsRetriever {
         }
     }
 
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/metrics/ClusterDeploymentMetricsRetriever.java
```diff
@@ -162,4 +164,8 @@ private static Slime doMetricsRequest(URI hostURI) {
             return new Slime();
         }
     }
+
+    public static OptionalDouble optionalDouble(Inspector field) {
```
mpolden, line 168: Make it private?

mpolden (approved):

PR #17691, opened 2021-04-30
Make more Distributor internals only available to friended tests

vekterli:
@geirst please review

Also add more assertions that such functions are only called in legacy mode.


geirst (approved):
👍

PR #17692, opened 2021-04-30
Revert "Jonmv/reindexing when new extra field is added"

jonmv:
Reverts vespa-engine/vespa#17670

@freva please merge.


jonmv:
Broke an assumption that some optional is always present, I guess.

freva (approved):

PR #17693, opened 2021-04-30
add api for detecting cell existence

arnej27959:
* new API "has(TensorAddress)" detects if a Tensor has a cell with the given address.
* use new API in join and merge.  This will give different results for
  cells that are present but contain NaN versus cells that aren't present at all.
* use new API in slice. This gives a different default (0, not NaN) when trying to
  access cells that aren't present.

@lesters please review
@bratseth FYI
@havardpe FYI


lesters (approved):
👍

PR #17695, opened 2021-04-30
Avoid SSE2 built-in functions when not using x86_64 arch.

toregge:
@baldersheim : please review


baldersheim (approved):
This must be cleaned up later on. Too many ifdef spread around.

PR #17697, opened 2021-04-30
Jonmv/reapply rest apis without trailing slash requirement

jonmv:
@bjorncs please review.

### DIFF
--- a/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
+++ b/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
@@ -144,7 +144,7 @@ class RestApiImpl implements RestApi {
     private static Route createDefaultRoute() {
         RouteBuilder routeBuilder = new RouteBuilderImpl("{*}")
                 .defaultHandler(context -> {
-                    throw new RestApiException.NotFound();
+                    throw new RestApiException.NotFound(context.request());
                 });
         return ((RouteBuilderImpl)routeBuilder).build();
     }
--- a/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
+++ b/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
@@ -151,10 +151,11 @@ class RestApiImpl implements RestApi {
 
     private static List<ExceptionMapperHolder<?>> combineWithDefaultExceptionMappers(
             List<ExceptionMapperHolder<?>> configuredExceptionMappers, boolean disableDefaultMappers) {
-        List<ExceptionMapperHolder<?>> exceptionMappers = new ArrayList<>(configuredExceptionMappers);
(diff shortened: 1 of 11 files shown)


container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
```diff
@@ -151,10 +151,11 @@ private static Route createDefaultRoute() {
 
     private static List<ExceptionMapperHolder<?>> combineWithDefaultExceptionMappers(
             List<ExceptionMapperHolder<?>> configuredExceptionMappers, boolean disableDefaultMappers) {
-        List<ExceptionMapperHolder<?>> exceptionMappers = new ArrayList<>(configuredExceptionMappers);
+        List<ExceptionMapperHolder<?>> exceptionMappers = new ArrayList<>();
```
bjorncs, line 154: This will no longer make is possible to override a single builtin handlers (as they will always have priority).

jonmv:
PTaAL

container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
```diff
@@ -151,11 +152,12 @@ private static Route createDefaultRoute() {
...
     private static List<ExceptionMapperHolder<?>> combineWithDefaultExceptionMappers(
             List<ExceptionMapperHolder<?>> configuredExceptionMappers, boolean disableDefaultMappers) {
-        List<ExceptionMapperHolder<?>> exceptionMappers = new ArrayList<>();
+        List<ExceptionMapperHolder<?>> exceptionMappers = new ArrayList<>(configuredExceptionMappers);
         if (!disableDefaultMappers){
             exceptionMappers.add(new ExceptionMapperHolder<>(RestApiException.class, (context, exception) -> exception.response()));
         }
-        exceptionMappers.addAll(configuredExceptionMappers);
+        // Topologically sort children before superclasses, so most the specific match is found by iterating through mappers in order.
+        exceptionMappers.sort((a, b) -> a.type.isAssignableFrom(b.type) ? 1 : b.type.isAssignableFrom(a.type) ? -1 : 0);
```
bjorncs, line 160: Is comparator correct for case when a.type == b.type ?

bjorncs (approved):

container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
jonmv: No, but perhaps that should be detected construction time as a user error? Or perhaps ee should Allie overprising the default ones, but only once?

jonmv:
```suggestion
        exceptionMappers.sort((a, b) -> (a.type.isAssignableFrom(b.type) ? 1 : 0) + (b.type.isAssignableFrom(a.type) ? -1 : 0));
```

PR #17702, opened 2021-05-01
Exclude jakarta artifacts that conflict or overlap

gjoranv:
Verified system test in docker.


bjorncs (approved):

bjorncs:
There are also enforcers in the internal repository that needs to be updated.

PR #17705, opened 2021-05-02
- Use the enum batch updater directly. This avoids a temporary map an…

baldersheim:
…d you have everything already in the L1 cache.

- Remeber the enum in the scratchPad. This avoid a lookup later on during commit.

@toregge PR


toregge (approved):
@geirst : FYI

baldersheim:
Build failure is unrelated.

PR #17708, opened 2021-05-03
Pick the same slobroks each time

bratseth:
... when there are only container clusters. Review but do not merge.


hmusum:
This might change slobrok ids, which cannot be avoided, but if that might lead to issues, perhaps guard with a feature flag?

bratseth:
Ids are by index so it won't really change, but location may change.
That should be ok, but to be sure I want to just wait until this is only used for apps without content clusters.

hmusum (approved):
👍

PR #17713, opened 2021-05-03
Make more distributor internals private 2; the much anticipated sequel [run-systemtest]

vekterli:
@geirst please review. Same as #17691 but with added commit a4e1bdc to avoid triggering any assertions when running under the legacy code path.


geirst (approved):
👍

PR #17718, opened 2021-05-03
Remove notification serialization migration

freva:


mpolden (commented):
`NotificationsDb` doesn't appear rewrite all data on startup. Are you sure this is safe to merge?

freva:
Yeah, I had to manually clear all notifications because some notifications were stuck forever since their deployment was expired before the code to clean notifications on expiry was merged (#17689)

mpolden (approved):

PR #17720, opened 2021-05-04
Store application package warning notification for direct deployments with deployment ID as source

freva:
Currently application package warnings have 2 different source types;
* For prod deployments they have tenant+application as source, the idea is that the warnings will be the same across the different instances and zones, since it is the same application package.
* For manual deployments, the source is the full application ID. This PR changes that to be the deployment ID because:
  - This simplifies the code in figuring out which notifications are related to dev/perf deployments so that we can hide those from the main views in the console
  - Allows for having different warnings if user deploys different application package, under the same application ID, to f.ex. dev vs perf.
  - Automatically cleans up notifications on deployment expiry: https://github.com/vespa-engine/vespa/blob/705b5fb3b553bebd279ad27838267d407533f0bb/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java#L706


mpolden (approved):
👍

PR #17725, opened 2021-05-04
Remove unnecessary interface and package exports

mpolden:
`CuratorArchiveBucketDb` is not a component (maybe it was in the past?).

@andreer or @freva


freva (approved):
:trophy:

PR #17731, opened 2021-05-04
Remove unnecessary interface and package export

mpolden:
Same as #17725, but keep exporting
`com.yahoo.vespa.hosted.controller.api.integration.archive` as classes in that
package are used by the `ServiceRegistry` component.

@hmusum (merge this tomorrow)


hmusum (approved):

andreer:
Thanks @mpolden !

PR #17732, opened 2021-05-04
Maintain iam roles for tenants with prod deployment

tokle:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/TenantRoleMaintainer.java
```diff
@@ -26,10 +27,15 @@ protected boolean maintain() {
...
+                .filter(this::hasProductionDeployment)
                 .collect(Collectors.toList());
         roleService.maintainRoles(tenantsWithRoles);
         return true;
     }
+
+    private boolean hasProductionDeployment(TenantName tenant) {
+        return controller().applications().asList(tenant).stream()
+                .map(Application::productionInstances)
+                .anyMatch(prodInstances -> prodInstances.size()>0);
```
oyving: `isEmpty`?

oyving (approved):

PR #17734, opened 2021-05-04
Revert "Revert "- GC unused deab-bytes-ratio.""

baldersheim:
Reverts vespa-engine/vespa#17729


baldersheim:
@hmusum Now I have GCed the test. Too many branches on multiple container and the same colour on the terminals can cause tests to be run in the wrong terminal...

hmusum:
👍

PR #17742, opened 2021-05-05
skip adding logforwarder on hosts if they have membership with cluste…

arnej27959:
…r type admin

@bratseth please review
@jonmv FYI

### DIFF
--- a/config-model/src/main/resources/schema/admin.rnc
+++ b/config-model/src/main/resources/schema/admin.rnc
@@ -105,7 +105,7 @@ ClusterControllers = element cluster-controllers {
 }
 
 LogForwarding = element logforwarding {
-    attribute unconditional { xsd:boolean }? &
+    attribute include-admin { xsd:boolean }? &
     element splunk {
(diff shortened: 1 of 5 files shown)


config-model/src/main/resources/schema/admin.rnc
```diff
@@ -105,6 +105,7 @@ ClusterControllers = element cluster-controllers {
 }
 
 LogForwarding = element logforwarding {
+    attribute unconditional { xsd:boolean }? &
```
bratseth, line 108: "all"? "include-admin"?

bratseth (commented):
Not sure if "unconditional" is a good name, will we expose this?
Otherwise LGTM

arnej27959:

config-model/src/main/resources/schema/admin.rnc
arnej27959: include-admin sounds good.  it was mostly added in case it was needed for some unit test :-)

bratseth (approved):

PR #17744, opened 2021-05-05
Run single stripe in its own thread when not using legacy mode [run-systemtest]

vekterli:
@geirst please review

The (currently single) stripe is now run as part of the distributor
stripe pool instead of being transitively invoked by the main thread.

Introduce an explicit message mutex per stripe that is used for
external messages and status requests when not using legacy mode.
Use per-stripe wakeup mechanisms instead of the framework-global
mutex used in the legacy code path.

Additional work remains to bring back a dedicated message run-queue
for the top-level distributor, so this is not yet thread safe for
operations to the main `BucketDBUpdater`.


geirst (approved):
👍

PR #17751, opened 2021-05-05
Dispatch messages to be handled by BucketDBUpdater to main distributor thread [run-systemtest]

vekterli:
@geirst please review

Required to ensure no race conditions can happen from processing such
messages from arbitrary RPC/CommunicationManager threads.


geirst (approved):
👍

PR #17752, opened 2021-05-05
Make status reporting from distributor and bucket db updater work in new stripe mode [run-systemtest]

geirst:
@vekterli please review


vekterli (approved):
👍🎉

PR #17753, opened 2021-05-05
Now use of bucket executor is default also for pruning of removed doc…

baldersheim:
…uments.

@geirst or @toregge PR
Just review, merge will happen once burn in has completed.


geirst (approved):
👍

PR #17761, opened 2021-05-06
Upgrade athenz to 1.10.14

tokle:


bjorncs (approved):

jonmv:
Is this the cause for the reduced performance of the Athenz performance tests?

jonmv:
Hmm, probably not. We have the same behaviour with some other pure http tests.

PR #17762, opened 2021-05-06
Create metric for nr of tenants on each plan

oyving:
Not sure if the metric name is great, so please advise

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
@@ -45,7 +45,7 @@ import java.util.stream.Collectors;
  */
 public class MetricsReporter extends ControllerMaintainer {
 
-    public static final String TENANT_METRIC = "tenant";
+    public static final String TENANT_METRIC = "billing.tenants";
     public static final String DEPLOYMENT_FAIL_METRIC = "deployment.failurePercentage";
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -193,6 +197,23 @@ private void reportChangeDurations(Map<NodeVersion, Duration> changeDurations, S
...
 
+    private void reportTenantMetrics() {
+        if (! controller().system().isPublic()) return;
+
+        var planCounter = new TreeMap<String, Integer>();
+
+        controller().tenants().asList().forEach(tenant -> {
+            var planId = controller().serviceRegistry().billingController().getPlan(tenant.name());
+            planCounter.computeIfPresent(planId.value(), (k, v) -> v + 1);
+            planCounter.putIfAbsent(planId.value(), 1);
```
mpolden, line 208: Can be simplified to `planCounter.merge(planId.value(), 1, Integer::sum)`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporter.java
```diff
@@ -43,6 +45,7 @@
  */
 public class MetricsReporter extends ControllerMaintainer {
 
+    public static final String TENANT_METRIC = "tenant";
```
mpolden, line 48: Should have a namespace like the other metrics. Maybe `billing.tenants`?

mpolden (approved):

PR #17763, opened 2021-05-06
Remove assert for a scenario that might occur in wait_until_unparked()

geirst:
A stripe thread is parked as part of another thread calling DistributorStripePool::park_all_threads().
The stripe thread will then be inside DistributorStripePool::park_thread_until_released(),
just waiting to call DistributorStripeThread::wait_until_unparked().
Before this is called, the other thread can call DistributorStripePool::unpark_all_threads(),
and the _should_park variable in DistributorStripeThread is set to false again.
When the stripe thread calls DistributorStripeThread::wait_until_unparked(), it is already unparked.
This is a scenario that might occur when the parking / unparking loop is short.

@vekterli please review


vekterli (approved):
🦄 👍

PR #17768, opened 2021-05-06
Only include non-retired nodes in redundancy computation

bratseth:
Usually redundancy is down-adjusted to match the nodes per group,
but not if the node count requested is high enough but downscaled
by capacity policies. The redundancy is also adjusted down if there
are not enough nodes in the actual allocation, which normally
handles this case, but not if there are retired nodes.
That is fixed in this PR by not counting retired nodes
when making that adjustment.


mpolden (approved):
Great!

PR #17780, opened 2021-05-07
Check for query timeout before dispatching streaming search query

bjorncs:


container-search/src/main/java/com/yahoo/vespa/streamingvisitors/VdsStreamingSearcher.java
```diff
@@ -136,6 +136,10 @@ private int inferEffectiveQueryTraceLevel(Query query) {
 
     @Override
     public Result doSearch2(Query query, Execution execution) {
+        if (query.getTimeLeft() < 0) {
```
baldersheim, line 139: You should probably use <= 0 in case there is some capping to positive number. As Hjallis said. "Me kan aldri gå på 0"

baldersheim (changes requested):
<=

PR #17784, opened 2021-05-07
Balder/make final and constructor public

baldersheim:
@hmusum or @bjorncs PR


hmusum (approved):
Overall looks good to me, but there are some details in the low level stuff that I haven't looked to close at

baldersheim:
There should not be anything special here. This is the first of 3.
I wanted to split them out as there as some details later on that are more important to verify.

PR #17785, opened 2021-05-07
Reduce the lifetime and scope of ConfigInstance.Builder and the Confi…

baldersheim:
…gPayload

by using a separate method.

@hmusum PR


hmusum (approved):
Great!

PR #17786, opened 2021-05-07
Use AbstractUtf8Array in interface and a more flexible implementation…

baldersheim:
… to reduce amount of copying.

@hmusum PR


hmusum (changes requested):
Looks good to me, but some unit tests fail

baldersheim:
I think it is OK now, but I want to refactor the ByteWriter test and also use that for the Utf8ByteWriter first. o will hold for now.

hmusum:
Fails one unit test (where config is empty)

hmusum (approved):

PR #17798, opened 2021-05-10
Don't consider number of working nodes in coverage

bratseth:
Trying to figure out the right groups to send queries to based
on the number of nodes in the group has many potential issues
at times of topology changes.

Since we could the number of documents available in each group
by summing documents in working nodes, we do not need to also
separately consider the number of working nodes in the group for
correctness.

Since we use adaptive dispatching by default we also do not need
to consider it to avoid overloading groups with less resources
available but enough documents.


baldersheim (approved):
👍

PR #17800, opened 2021-05-10
Use just one executor

hmusum:


baldersheim (approved):
@hmusum I suggest you gate the merge of this one.

PR #17802, opened 2021-05-10
Add timed batching of explicit host info sends triggered by stripes [run-systemtest]

vekterli:
@geirst please review

Since distributor stripes may independently reach a conclusion that
a `GetNodeState` reply containing new host info should be sent back to
the cluster controller, implement basic rate limiting/batching of
concurrent sends.

Batching has two separate modes of operation:
- If the node is initializing, host info will be sent immediately after
  _all_ stripes have reported in (they will always do this post-init).
  This is not timed, in order to minimize latency of bucket info being
  visible to the cluster controller.
- If the node has already initialized, have a grace period of up to 1
  second from the time the first stripe signals its intent to send
  host info until it's actually sent. This allows several stripes
  to complete their recovery mode and signal host info intents during
  this second.

Batch time period is currently not configurable, may be done later if
deemed useful or necessary.

### DIFF
--- a/storage/src/vespa/storage/distributor/distributor.cpp
+++ b/storage/src/vespa/storage/distributor/distributor.cpp
@@ -606,7 +606,7 @@ Distributor::notify_stripe_wants_to_send_host_info(uint16_t stripe_index)
 }
 
 bool
-Distributor::may_send_host_info_on_behalf_of_stripes() noexcept
+Distributor::may_send_host_info_on_behalf_of_stripes([[maybe_unused]] std::lock_guard<std::mutex>& held_lock) noexcept
 {
     bool any_stripe_wants_to_send = false;
     for (const auto& stats : _stripe_scan_stats) {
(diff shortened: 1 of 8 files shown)


storage/src/vespa/storage/distributor/distributor.cpp
```diff
@@ -575,6 +590,56 @@ Distributor::enableNextConfig() // TODO STRIPE rename to enable_next_config_if_c
...
+    assert(stripe_index < _stripe_scan_stats.size());
+    auto& stats = _stripe_scan_stats[stripe_index];
+    stats.wants_to_send_host_info = true;
+    stats.has_reported_in_at_least_once = true;
+    // TODO STRIPE consider if we want to wake up distributor thread here. Will be rechecked
+    //  every nth millisecond anyway. Not really an issue for out-of-band CC notifications.
+}
+
+bool
+Distributor::may_send_host_info_on_behalf_of_stripes() noexcept
```
geirst, line 609: Consider taking lock_guard reference as parameter to signal that _stripe_scan_notify_mutex must be held.

geirst (approved):
👍

storage/src/vespa/storage/distributor/distributor.cpp
vekterli: Done

PR #17803, opened 2021-05-10
vespa-base-libs requires epel-release (for centos), openblas and re2.

toregge:
@baldersheim : please review


toregge:
@aressem : FYI

baldersheim (approved):

PR #17813, opened 2021-05-10
Include legacy endpoint in certificate in public

mpolden:
Some applications, both in public CD and public systems, have dev endpoints in
the certificate issued for production instances.

Due to the check (now) removed in 8d64ba3dd4, any attempts to change DNS names
for these deployments will fail because we no longer include dev endpoints when
deploying a production instance. We probably did at some point?

In any case I don't see why we need this check. Both unit and integration tests
verify that we request certificates for the correct names.

Same as #17801 + two last commits.

@andreer


andreer (approved):

mpolden:
@andreer or @tokle: PTAL, only last commit is new.

andreer (approved):
LGTM, good to use a flag for this

PR #17815, opened 2021-05-11
Revert "Revert "Don't consider number of working nodes in coverage""

bratseth:
Reverts vespa-engine/vespa#17807


bratseth:
Must go with https://github.com/vespa-engine/system-test/pull/1559

baldersheim (approved):

PR #17819, opened 2021-05-11
Extend TickableStripe interface to avoid direct access to DistributorStripe internals

vekterli:
@geirst please review

Also lets us test guard functionality much more easily since its target is now fully  mockable.


geirst (approved):
👍

PR #17822, opened 2021-05-11
Arnej/wire constant types

arnej27959:

NOTE: this feature (tensor constants directly in the rank profile) isn't recognised by the back-end, but
after discussion with @havardpe we think this is easy to handle from the config produced by config-model.

The alternative is to deprecate this completely and remove various unit tests (and .sd parsing).

@bratseth please review
@lesters FYI


bratseth (approved):
👍

PR #17824, opened 2021-05-11
Stop all stripe threads before starting shutdown (and closing) of the…

geirst:
… storage link chain.

This is required to avoid stripe threads being able to send up messages
while the communication manager is being closed.
Such messages will fail at the RPC layer (already closed)
and an error reply is sent down from the communication manager.
This triggers an assert in StorageLink::sendDown() which is already CLOSED.

@vekterli please review


vekterli (approved):
:+1:

PR #17826, opened 2021-05-11
support functional constants

havardpe:
be more specific about const number values (GDBT trees):

is_const -> is_const_double
get_const_value -> get_const_double_value

Add more generic 'get_const_value' that can also be used for tensor
values. Allow it to be called even in the case of parse error, in
which case it does not produce a value (same as for non-const
sub-expressions).

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #17829, opened 2021-05-12
added config for file-distributed ranking expressions

havardpe:

@baldersheim please review


searchcore/src/vespa/searchcore/config/ranking-expressions.def
```diff
@@ -0,0 +1,5 @@
+# Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
```
geirst, line 1: Please update copyright.

searchcore/src/vespa/searchcore/config/ranking-expressions.def
```diff
@@ -0,0 +1,5 @@
...
+namespace=vespa.config.search.core
+
+constant[].name string
```
baldersheim, line 4: Is 'constant' the correct name, or should it be 'expression' ?

searchcore/src/vespa/searchcore/config/ranking-expressions.def
```diff
@@ -0,0 +1,5 @@
+# Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+namespace=vespa.config.search.core
+
```
baldersheim, line 3: I wonder if it should be put in configdefinitions together with rank-profiles.def. I guess it might be need in java too.

baldersheim, line 3: Is it so tightly coupled with rank-profiles config that it should be added to rank-profiles.def  ?
Or is it better to keep them as separate configs.

baldersheim (changes requested):

searchcore/src/vespa/searchcore/config/ranking-expressions.def
havardpe: done

searchcore/src/vespa/searchcore/config/ranking-expressions.def
havardpe: fixed

searchcore/src/vespa/searchcore/config/ranking-expressions.def
havardpe: I put it where the ranking constants already are located

havardpe: I did it the same way as constants and onnx models were done. Putting it into rank-profiles.def would make it a separate top-level array that needs to be handled separately form the rest of the config. We do need some extra wiring to add a new config though. Maybe we should combine config about distributed files somehow... @geirst any thoughts?

baldersheim: I noticed afterwards. Agree Better keep them close as they are the same.

baldersheim: I suggest we add this one now, and then we consider consolidation afterwards.

baldersheim (approved):

PR #17831, opened 2021-05-12
Specify Janitor as component in container cluster

bjorncs:


config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
```diff
@@ -188,6 +188,7 @@ public ContainerCluster(AbstractConfigProducer<?> parent, String configSubId, St
         addSimpleComponent("com.yahoo.container.handler.VipStatus");
         addSimpleComponent(com.yahoo.container.handler.ClustersStatus.class.getName());
         addSimpleComponent("com.yahoo.container.jdisc.DisabledConnectionLogProvider");
+        addSimpleComponent(com.yahoo.jdisc.http.server.jetty.Janitor.class);
```
baldersheim, line 191: not .getName() as for ClusterStatus ?

baldersheim (approved):

config-model/src/main/java/com/yahoo/vespa/model/container/ContainerCluster.java
bjorncs:
```
private void addSimpleComponent(Class<?> clazz) {
        addSimpleComponent(clazz.getName());
    }
```

PR #17834, opened 2021-05-12
Implement class that provides mapping from bucket space to state for …

geirst:
…that space.

@vekterli please review

### DIFF
--- a/storage/src/vespa/storage/distributor/bucket_space_state_map.h
+++ b/storage/src/vespa/storage/distributor/bucket_space_state_map.h
@@ -1,0 +1,74 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include <vespa/document/bucket/bucketspace.h>
+#include <vespa/persistence/spi/clusterstate.h>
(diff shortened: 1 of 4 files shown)


storage/src/vespa/storage/distributor/bucket_space_state_map.h
```diff
@@ -0,0 +1,74 @@
...
+        return *_distribution;
+    }
+};
+
+/**
+ * Provides mapping from bucket space to state for that space.
+ */
+class BucketSpaceStateMap {
+private:
+    using StateMap = std::unordered_map<document::BucketSpace, std::unique_ptr<BucketSpaceState>, document::BucketSpace::hash>;
```
vekterli, line 52: I'm wondering if we have any hidden assumptions that bucket spaces should be ordered on their `BucketSpace` identifier... Have we used unordered maps for this elsewhere?

vekterli (approved):
:+1:

storage/src/vespa/storage/distributor/bucket_space_state_map.h
geirst: Yes, DistributorBucketSpaceRepo also uses an unordered map.

Issue #17825, opened 2021-05-11
Vespa not able to start the content node on Azure File mount in Kubernetes even though the drives are mounted properly.


labriedion:
I've been able to run apps in Azure Kubernetes Services but Vespa fails to start when I mount /opt/vespa/var to a persistent volume. The Azure file mounts work since logs and files are written inside them when deploying the pod. Basically everything seems fine when deploying the pods, and I get no error messages when I run:

`kubectl exec vespa-0 -n vespa -- bash -c "/opt/vespa/bin/vespa-deploy prepare /vespa-app && /opt/vespa/bin/vespa-deploy activate"`

But we I try to load data or reach documents at http:/myip:8080/document/v1/{mynamespace}/{mydocumenttype}/docid/1, the connection times out. 

I think I located the error in the logs, it seems like content cluster can't even start.

[...]

PR #17835, opened 2021-05-12
Sync is not a mandatory operation. Assuming the fs that fails sync ar…

baldersheim:
…e safe anyway.

So let us just log a warning for now.

@vekterli or @toregge PR
See if this helps #17825

### DIFF
--- a/vespalib/src/vespa/vespalib/io/fileutil.cpp
+++ b/vespalib/src/vespa/vespalib/io/fileutil.cpp
@@ -398,18 +398,13 @@ File::sync()
 {
     if (_fd != -1) {
         if (::fsync(_fd) == 0) {
-            LOG(debug, "sync(%s): File synchronized with disk.",
-                _filename.c_str());
+            LOG(debug, "sync(%s): File synchronized with disk.",_filename.c_str());
         } else {
-            asciistream ost;
-            ost << "sync(" << _filename << "): Failed, errno(" << errno << "): "
-                << safeStrerror(errno);
-            throw IoException(ost.str(), IoException::getErrorType(errno),
-                              VESPA_STRLOC);
+            LOG(warning, "fsync(%s): Failed to sync file. errno(%d): %s",
+                _filename.c_str(), errno, safeStrerror(errno).c_str());
         }
     } else {
-        LOG(debug, "sync(%s): Called on closed file.",
-            _filename.c_str());
+        LOG(debug, "sync(%s): Called on closed file.", _filename.c_str());
     }
 }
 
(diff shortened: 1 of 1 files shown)


vespalib/src/vespa/vespalib/io/fileutil.cpp
```diff
@@ -398,18 +398,13 @@ File::sync()
...
-                _filename.c_str());
+            LOG(debug, "sync(%s): File synchronized with disk.",_filename.c_str());
         } else {
-            asciistream ost;
-            ost << "sync(" << _filename << "): Failed, errno(" << errno << "): "
-                << safeStrerror(errno);
-            throw IoException(ost.str(), IoException::getErrorType(errno),
-                              VESPA_STRLOC);
+            LOG(warning, "fsync(%s): Failed to sync file. errno(%d): %s",
+                _filename.c_str(), errno, safeStrerror(errno).c_str());
```
vekterli, line 404: Should we use a buffering logger here (`LOGBP` and similar) to avoid spamming the log? If `fsync` fails once due to the underlying fs not supporting it, it's likely to fail _every_ time

vespalib/src/vespa/vespalib/io/fileutil.cpp
```diff
@@ -398,18 +398,13 @@ File::sync()
 {
     if (_fd != -1) {
         if (::fsync(_fd) == 0) {
-            LOG(debug, "sync(%s): File synchronized with disk.",
-                _filename.c_str());
+            LOG(debug, "sync(%s): File synchronized with disk.",_filename.c_str());
```
vekterli, line 401: Micro-nit: missing space after comma

vekterli (approved):
:+1: it would be nice to have some notion of whether fsync _is_ supposed to work on a particular file system, in which case it would be considered a mandatory operation

toregge:
I don't think fsync is supposed to always fail on smb shares, but if you have unwritten data and writing it causes some quota to be exceeded then it will likely fail.

vespalib/src/vespa/vespalib/io/fileutil.cpp
baldersheim: Fixed

vespalib/src/vespa/vespalib/io/fileutil.cpp
baldersheim: I doubt that fsync is the root cause. It reports "#define EACCES          13      /* Permission denied */" which is not listed as a possible errno for fsync.
I hope that this might cause it to progress and perhaps shed some light on what is the root cause.

What is LOGBP using for duplicate detection ? Call site or actual message ?

vekterli: From what I can recall `LOGBP` uses the call site (`__FILE__` + `__LINE__` combo), so no need to provide an explicit dedupe token

PR #17836, opened 2021-05-12
Enable more fine grained control of linker options for shared libraries.

toregge:
@baldersheim : please review
@aressem : FYI


baldersheim (approved):

baldersheim:
BTW, what is the usecase ?

toregge:
Use case is a shared library containing a python extension.

PR #17870, opened 2021-05-17
Balder/use thread safe remove

baldersheim:
@jonmv @ @vekterli or @bratseth PR

### DIFF
--- a/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/ContentPolicy.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/ContentPolicy.java
@@ -108,13 +108,13 @@ public class ContentPolicy extends SlobrokPolicy {
             while ((targets.total != 0) &&
                    (100 * targets.list.size() / targets.total >= requiredUpPercentageToSendToKnownGoodNodes))
             {
-                int randIndex = randomizer.nextInt(targets.list.size());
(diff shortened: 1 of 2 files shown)


bratseth (approved):

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/ContentPolicy.java
```diff
@@ -108,13 +108,13 @@ String getRandomTargetSpec(RoutingContext context) {
             while ((targets.total != 0) &&
                    (100 * targets.list.size() / targets.total >= requiredUpPercentageToSendToKnownGoodNodes))
             {
-                int randIndex = randomizer.nextInt(targets.list.size());
-                String targetSpec = getTargetSpec(targets.list.get(randIndex), context);
+                Integer distributor = targets.list.get(randomizer.nextInt(targets.list.size()));
```
vekterli, line 111: Is it possible to race with a concurrent removal from `targets.list` so that `randomizer.nextInt(targets.list.size())` returns a value that is greater than the max value accepted by `targets.list.get()`?

baldersheim: Yes, it is... Will ponder a fix

jonmv: Just use synchronisation instead? Perhaps with round-robin.

PR #17871, opened 2021-05-17
Support nearest neighbor query in client's Field API

eltscheminov:
Context:
While implementing transformers we noticed the java-client lacks an API to query nearest neighbors as described in https://docs.vespa.ai/en/reference/query-language-reference.html#nearestneighbor. This PR aims to extend the Field API to support it.

Pinging @bratseth ( the upward `OWNER` file is the root and it points to you )


bratseth (approved):
Nice

bratseth:
This should be released tomorrow - you can follow it here https://factory.vespa.oath.cloud/

PR #17878, opened 2021-05-18
Synchronise when getting arbitrary distributor in ContentPolicy

jonmv:
@baldersheim please review.


baldersheim (changes requested):
I do not think we want syncronisation in this frequently called code.

jonmv:
It's not that frequently called, though, is it? Only called when the cached state is wrong (outdated, once in a while, or when a distributor is down).

vekterli:
Hmm... It's not immediately obvious (well, not to me at least) what removing randomized selection in favor of round-robin does for the expected convergence time to find a distributor that has valid view of the world. If we assume a majority of distributor nodes are valid, the probability of hitting a valid distributor using randomized selection should increase exponentially with the number of retries.

PR #17883, opened 2021-05-18
Make distributor timestamp generation thread safe [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

New behavior:
- Only allow time to travel forwards within a given distributor process'
  lifetime. This is a change from the old behavior, which would emit a
  warning to the logs and happily continue from a previously used second,
  possibly causing the distributor to reuse timestamps.
- Try to detect cases where the wall clock has been transiently set far
  into the future--only to bounce back--by aborting the process if the
  current observed time is more than 120 seconds older than the highest
  observed wall clock time. This is an attempt to avoid generating _too_
  many bogus future timestamps, as the distributor would otherwise continue
  generating timestamps within the highest observed second.


geirst (approved):
👍

PR #17890, opened 2021-05-19
allow adding un-configured peer slobrok

arnej27959:

@bratseth please review
NOTE: while I tested this manually, I'm not sure which scenarios it will really help with.
also, it makes it even more likely that we'll trigger problems if there is partial network connectivity.


bratseth (approved):

arnej27959:
> Upgrades (or any config change which takes a long time to propagate) changing the set of Slobroks.

yes - it should work if you have two config generations where one is a subset of the other.  If they have just a partial overlap you'll only get full information in the intersection, I think.

> Could you explain the scenario do you have in mind here?

if a slobrok that is outside the cluster (removed, but starts with old config) has some communications open
but can't communicate with any content nodes (removed from ACLs), when it starts it will mark "everything" as down.

bratseth:
> it should work if you have two config generations where one is a subset of the other. 

Yes, agreed.

PR #17895, opened 2021-05-19
Add ONNX-RT evaluator to model-integration module

lesters:
@bratseth Please review.

This is only for exposing an interface (`OnnxEvaluator`) to ONNX Runtime with dependencies inside the `model-integration`.


bratseth (approved):
Great

PR #17897, opened 2021-05-19
Bjorncs/fix

bjorncs:


container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
```diff
@@ -108,6 +113,14 @@ public void onOpened(Connection connection) {
         });
     }
 
+    private void addReferenceToSslEngine(SocketChannelEndPoint endpoint, ConnectionInfo info, SSLEngine sslEngine) {
+        if (sslEngine != null) {
```
jonmv, line 117: Can this be?

jonmv (approved):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
bjorncs: At least for ALPN, but probably not for SSL connection.

PR #17901, opened 2021-05-19
Limit the number of nodes retired per stateful cluster

mpolden:
Prioritizing by flavor didn't work well enough.

@bratseth

FYI @bjormel


bratseth (approved):
Very well done to get this readable

mpolden:
Thanks!

PR #17907, opened 2021-05-19
Balder/raw as base64

### LINKED ISSUE #17668: Raw document fields "decoded" on retrieval
When base64 encoded data is fed to a `raw` type field of a document (for example `"KGxwMApTJ2EnCnAxCmFJMgphKGRwMgpJMQpJMApzYS4="`), the data is base64 decoded before being retrieved (producing in this example `"(lp0\nS'a'\np1\naI2\na(dp2\nI1\nI0\nsa."`). I have verified via the V1 Document API that the encoded data is stored as-is, and is only decoded during retrieval. Importantly, I have found in my own application that the decoding sometimes converts to escaped unicode (for instance `"\\x80\\x03]q\\x00X\\x07\\x00\\x00\\x00contextq\\x01a."` instead of the original string `"\x80\x03]q\x00X\x07\x00\x00\x00contextq\x01a."`), making it a challenge to recover the original data. To my understanding, the data fed to any document field should be returned in the exact format in which it was uploaded.

This issue can be minimally reproduced via updating any of the sample app document schemas with a raw field and updating the corresponding sample data feed accordingly, and then running the sample app with any query. I have made [a PR to the sample-apps repository](https://github.com/vespa-engine/sample-apps/pull/525) modifying the blog-search sample as described above to reproduce the issue. This PR is not meant to be merged, but is there so that the changes can be checked out for ease of reproducibility.

baldersheim:
@bratseth PR
Fixes #17668

Note `raw-as-base64-in-summary` as suggested under `schema` as trigger.

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/Search.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/Search.java
@@ -61,6 +61,8 @@ public class Search implements ImmutableSearch {
     /** True if this doesn't define a search, just a document type */
     private boolean documentsOnly = false;
 
+    private boolean rawAsBase64 = false;
+
     /** The stemming setting of this search definition. Default is BEST. */
     private Stemming stemming = Stemming.BEST;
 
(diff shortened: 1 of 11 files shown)


baldersheim:
Note the first commit is just readability in test. The second is the one that matters.

bratseth (approved):
LGTM!

PR #17909, opened 2021-05-20
Multipart deploy

tokle:
@oyving @andreer fyi

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationApiHandler.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationApiHandler.java
@@ -8,14 +8,25 @@ import com.yahoo.config.provision.Zone;
 import com.yahoo.container.jdisc.HttpRequest;
 import com.yahoo.container.jdisc.HttpResponse;
 import com.yahoo.jdisc.application.BindingMatch;
+import com.yahoo.jdisc.http.HttpHeaders;
 import com.yahoo.vespa.config.server.ApplicationRepository;
 import com.yahoo.vespa.config.server.application.CompressedApplicationInputStream;
+import com.yahoo.vespa.config.server.http.BadRequestException;
 import com.yahoo.vespa.config.server.http.SessionHandler;
 import com.yahoo.vespa.config.server.http.Utils;
 import com.yahoo.vespa.config.server.session.PrepareParams;
 import com.yahoo.vespa.config.server.tenant.TenantRepository;
+import org.eclipse.jetty.http.MultiPartFormInputStream;
 
+import javax.servlet.http.Part;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.logging.Level;
+import java.util.stream.Collectors;
 
 import static com.yahoo.vespa.config.server.application.CompressedApplicationInputStream.createFromCompressedStream;
 import static com.yahoo.vespa.config.server.http.Utils.checkThatTenantExists;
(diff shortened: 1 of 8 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/ApplicationApiHandler.java
```diff
@@ -50,10 +64,34 @@ public ApplicationApiHandler(Context ctx,
...
+                .orElse(false);
+        if(multipartRequest) {
+            try {
+                MultiPartFormInputStream multiPartFormInputStream = new MultiPartFormInputStream(request.getData(), request.getHeader(CONTENT_TYPE), /* config */null, /* contextTmpDir */null);
+                Map<String, Part> parts = multiPartFormInputStream.getParts().stream()
+                        .collect(Collectors.toMap(Part::getName, p -> p));
+
+                byte[] params = parts.get(MULTIPART_PARAMS).getInputStream().readAllBytes();
+                log.log(Level.FINE, "Deploy parameters: [{}]", new String(params, StandardCharsets.UTF_8));
+                prepareParams = PrepareParams.fromJson(parts.get(MULTIPART_PARAMS).getInputStream().readAllBytes(), tenantName, zookeeperBarrierTimeout);
```
hmusum, line 83: Use `params` from above instead of reading all bytes again?

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
```diff
@@ -253,6 +272,53 @@ public static PrepareParams fromHttpRequest(HttpRequest request, TenantName tena
                             .build();
     }
 
+    public static PrepareParams fromJson(byte[] json, TenantName tenant, Duration barrierTimeout) {
+        Slime slime = SlimeUtils.jsonToSlime(json);
+        Inspector params = slime.get();
+
+        return new Builder().ignoreValidationErrors(params.field(IGNORE_VALIDATION_PARAM_NAME).asBool())
```
hmusum, line 279: Won't this fail if there is no field `IGNORE_VALIDATION_PARAM_NAME`?   Check with `valid()` first?

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
```diff
@@ -253,6 +272,53 @@ public static PrepareParams fromHttpRequest(HttpRequest request, TenantName tena
                             .build();
     }
 
+    public static PrepareParams fromJson(byte[] json, TenantName tenant, Duration barrierTimeout) {
+        Slime slime = SlimeUtils.jsonToSlime(json);
```
jonmv, line 276: Prefer `jsonToSlimeOrThrow`.

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
hmusum: Sure, that was just an example, though, same issue with other fields that are not checked explicitly

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
```diff
@@ -253,6 +272,53 @@ public static PrepareParams fromHttpRequest(HttpRequest request, TenantName tena
...
+        Inspector params = slime.get();
+
+        return new Builder().ignoreValidationErrors(params.field(IGNORE_VALIDATION_PARAM_NAME).asBool())
+                .dryRun(params.field(DRY_RUN_PARAM_NAME).asBool())
+                .verbose(params.field(VERBOSE_PARAM_NAME).asBool())
+                .timeoutBudget(SessionHandler.getTimeoutBudget(getTimeout(params, barrierTimeout)))
+                .applicationId(createApplicationId(params, tenant))
+                .vespaVersion(params.field(VESPA_VERSION_PARAM_NAME).asString())
```
jonmv, line 284: This will be `""` if not set. I would think `null` is a better choice?

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
jonmv: Yes. Looks to me like all bool fields are OK, but I didn't check the ones that have separate serialisers.

hmusum (approved):

hmusum (approved):

jonmv (approved):

jonmv:
```
09:56:55   [ERROR] Errors: 
09:56:55   [ERROR]   PrepareParamsTest.compareEmptyUrlparamsVsJson:167 » JsonParse Expected start o...```

PR #17914, opened 2021-05-20
Lesters/stateless onnx rt in model eval

lesters:
@bratseth Please review.

Changes are mainly within `model-evaluation` to evaluate ONNX models using the `OnnxEvaluator` in `model-integration`. The `ModelsEvaluator`  now needs a `OnnxModelConfig`, so the changes in `config-model` injects this. For now, no ONNX models are populated in the global rank profiles (coming soon to a PR near you), so this is not fully wired in yet. 

There are tests both for REST API and the injected `ModelsEvaluator`. Conceptually, if a ranking expression contains `onnxModel(modelname)` and `modelname` exists as a ONNX model, the `FunctionEvaluator` evaluates this first using ONNX runtime, and adds the result to the context before evaluating the rest of the expression.


bratseth (approved):
Great. If we ever get more of these we should do something to the design to localize this functionality more ...

PR #17916, opened 2021-05-20
Allow update of MutableBoolean by AND-ing and OR-ing with another boolean.

hakonhall:


hakonhall:
Hm, I'm unable to work around this anymore and am blocked on it.  Let me add other reviewers to unblock.

hmusum (approved):

PR #17919, opened 2021-05-20
Do not log warning when exact-terminator is not specified

hmusum:
'@@' is default value according to doc, so no need to warn


freva (approved):
:100:

PR #17917, opened 2021-05-20
Add CLI interface of vespa-feed-client

bjorncs:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #17921, opened 2021-05-20
Revert "Add CLI interface of vespa-feed-client"

hmusum:
Reverts vespa-engine/vespa#17917

Broke the build


hmusum:
@jonmv will fix

PR #17923, opened 2021-05-20
Write VCMR reports to node repo

olaaun:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VCMRReport.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public static String getReportId() {
+        return REPORT_ID;
+    }
+
+    /**
+     * Serialization functions - mapped to {@link Node#reports()}
+     */
+    public static VCMRReport fromReports(Map<String, JsonNode> reports) {
```
smorgrav, line 71: Could this be null?

smorgrav (approved):
I assume the null question is covered

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VCMRReport.java
olaaun: No, the map is always initialized

PR #17925, opened 2021-05-20
handle hex strings as tensor cell values in document also

arnej27959:

so, we have two implementations of parsing the same JSON format...

@lesters please review
@geirst @havardpe FYI


lesters (approved):
LGTM!

PR #17931, opened 2021-05-20
Warn when specifying resource limits in hosted Vespa

hmusum:
Want to throw and fail deployments, but need to make sure nobody
is doing this first, so warn for now.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/DomResourceLimitsBuilder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/DomResourceLimitsBuilder.java
@@ -1,8 +1,11 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.content.cluster;
 
+import com.yahoo.config.application.api.DeployLogger;
 import com.yahoo.vespa.model.builder.xml.dom.ModelElement;
 import com.yahoo.vespa.model.content.ResourceLimits;
+
+import java.util.logging.Level;
 
 /**
  * Builder for feed block resource limits.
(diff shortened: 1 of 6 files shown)


config-model/src/test/java/com/yahoo/vespa/model/content/ClusterResourceLimitsTest.java
```diff
@@ -114,6 +119,29 @@ public void default_resource_limits_when_feed_block_is_enabled_in_distributor()
                 new Fixture(true));
     }
 
+    @Test
+    // TODO: Change to expect exception being thrown when noone uses this in hosted
```
geirst, line 123:
```suggestion
    // TODO: Change to expect exception being thrown when no one uses this in hosted
```

geirst (approved):
👍

baldersheim:
Are we doing the same for both proton resources and cluster resources ?

geirst:
Yes, as DomResourceLimitsBuilder (with the check for hosted flag) is used for both cluster wide and proton limits in ClusterResourceLimits.

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/DomResourceLimitsBuilder.java
```diff
@@ -1,20 +1,28 @@
...
 
-    public static ResourceLimits.Builder createBuilder(ModelElement contentXml) {
+    public static ResourceLimits.Builder createBuilder(ModelElement contentXml, boolean hostedVespa, DeployLogger deployLogger) {
         ResourceLimits.Builder builder = new ResourceLimits.Builder();
         ModelElement resourceLimits = contentXml.child("resource-limits");
-        if (resourceLimits == null) {
+        if (resourceLimits == null) { return builder; }
+
+        if (hostedVespa) {
+            deployLogger.log(Level.WARNING, "Element " + resourceLimits + " is not allowed, default limits will be used");
```
freva, line 23: Log with `.logApplicationPackage()` so the tenant gets a notification in console?

hmusum: Yes, thanks, done in https://github.com/vespa-engine/vespa/pull/17948

PR #17934, opened 2021-05-21
Wire in stateless ONNX runtime evaluation

lesters:
@bratseth Please review.

Main change is is `VespaModel.createGlobalRankProfiles` which adds the required onnx model info which is sent to model evaluation. The `StatelessOnnxEvaluationTest` is the main test for this PR. There are quite a bit of small changes to continue to support the conversion to native vespa rank expressions. I think this will have to be cleaned up a bit more going forward.


lesters:
Needs to be merged with https://github.com/vespa-engine/system-test/pull/1610

bratseth (approved):
👍

PR #17941, opened 2021-05-21
Jonmv/cache badges

jonmv:
@freva please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/BadgeApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/BadgeApiHandler.java
@@ -18,7 +18,12 @@ import com.yahoo.yolean.Exceptions;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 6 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/Badges.java
```diff
@@ -111,13 +111,12 @@ static String historyBadge(ApplicationId id, JobStatus status, int length) {
 
         Run lastTriggered = status.lastTriggered().get();
         List<Run> runs = status.runs().descendingMap().values().stream()
-                .filter(Run::hasEnded)
-                .limit(length + (lastTriggered.hasEnded() ? 0 : 1))
-                .collect(toList());
+                               .filter(Run::hasEnded)
+                               .skip(1)
+                               .limit(length + (lastTriggered.hasEnded() ? 0 : 1))
```
jonmv, line 116: Hmm ...

jonmv: Yes.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/BadgeApiHandler.java
```diff
@@ -35,6 +40,7 @@
     private final static Logger log = Logger.getLogger(BadgeApiHandler.class.getName());
 
     private final Controller controller;
+    private final Map<Key, Value> badgeCache = new ConcurrentHashMap<>();
```
mpolden, line 43: Since we're adding this deter abuse we should probably enforce a maximum size?

freva (commented):
LGTM, +1 with @mpolden on max size

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/deployment/BadgeApiHandler.java
jonmv: So each badge is ~5kb. If someone wants to play nasty, that's ~200Kb per job they know about, say ~1Mb per application. We don't have enough applications for this to be an issue. Also, if we were to reach the maximum size, someone could rotate the entries they were asking for, thus invalidating the whole cache anyway. So I'm not sure ... Perhaps we should actively evict entries expired entries, though, in case we get lots of badge serving in the future.

freva (approved):

PR #17946, opened 2021-05-21
Add extra constructor argument to HNSW fixture to normalize vectors.

toregge:
This allows inner dot product distance metric to be used instead of
angular distance metric.

@geirst : please review

### DIFF
--- a/ann_benchmark/src/vespa/ann_benchmark/vespa_ann_benchmark.cpp
+++ b/ann_benchmark/src/vespa/ann_benchmark/vespa_ann_benchmark.cpp
@@ -65,11 +65,13 @@ class HnswIndex
     TensorAttribute*                 _tensor_attribute;
     const NearestNeighborIndex*      _nearest_neighbor_index;
     size_t                           _dim_size;
+    bool                             _normalize_vectors;
 
     bool check_lid(uint32_t lid);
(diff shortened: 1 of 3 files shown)


ann_benchmark/src/vespa/ann_benchmark/vespa_ann_benchmark.cpp
```diff
@@ -180,7 +203,8 @@ HnswIndex::find_top_k(uint32_t k, const std::vector<float>& value, uint32_t expl
      * read guard is not taken here.
      */
     TopKResult result;
-    TypedCells typed_cells(&value[0], CellType::FLOAT, value.size());
+    std::vector<float> normalized_value;
```
geirst, line 206: I think we should keep this vector as a member variable to avoid creating it per search.

geirst (approved):
👍

ann_benchmark/src/vespa/ann_benchmark/vespa_ann_benchmark.cpp
toregge: That would cause incorrect behavior for batch mode.

geirst: Good point.

PR #17954, opened 2021-05-24
Use external rank expressions, gated by feature flag.

baldersheim:
@hmusum or @bratseth or @lesters PR
@havardpe FYI


bratseth (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -222,6 +224,7 @@ public FeatureFlags(FlagSource source, ApplicationId appId) {
         }
         @Override public boolean enableCustomAclMapping() { return enableCustomAclMapping; }
         @Override public int numDistributorStripes() { return numDistributorStripes; }
+        @Override public boolean isUseExternalRankExpression() { return useExternalRankExpression; }
```
hmusum, line 227:
```suggestion
        @Override public boolean useExternalRankExpressions() { return useExternalRankExpression; }
```

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -224,6 +224,13 @@
             "Takes effect after distributor restart",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundBooleanFlag USE_EXTERNAL_RANK_EXPRESSION = defineFeatureFlag(
+            "use-external-rank-expression", false,
+            List.of("baldersheim"), "2021-05-24", "2021-07-01",
+            "Wheter to use distributed external rank expression or inline in rankproperties",
```
hmusum, line 230:
```suggestion
            "Whether to use distributed external rank expression or inline in rankproperties",
```

hmusum (changes requested):
Two suggestions, one is necessary to implement interface correctly. Otherwise LGTM

hmusum (approved):

hmusum (approved):

PR #17957, opened 2021-05-25
Avoid coredump on bad/empty config.

baldersheim:
@havardpe or @arnej27959 PR

### DIFF
--- a/searchcore/src/apps/vespa-proton-cmd/vespa-proton-cmd.cpp
+++ b/searchcore/src/apps/vespa-proton-cmd/vespa-proton-cmd.cpp
@@ -3,6 +3,7 @@
 #include <vespa/slobrok/sbmirror.h>
 #include <vespa/config/common/configsystem.h>
 #include <vespa/config/common/exceptions.h>
+#include <vespa/vespalib/util/exceptions.h>
 #include <vespa/fnet/frt/supervisor.h>
 #include <vespa/fnet/frt/target.h>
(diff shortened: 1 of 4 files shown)


arnej27959 (approved):
this is OK in itself, but we really should figure out why we get empty config - that should never happen, right?

searchcore/src/apps/vespa-proton-cmd/vespa-proton-cmd.cpp
```diff
@@ -158,6 +159,9 @@ class App : public FastOS_Application
         } catch (config::InvalidConfigException& e) {
             fprintf(stderr, "ERROR: failed to get service location broker configuration\n");
             std::_Exit(1);
+        } catch (vespalib::IllegalStateException& e) {
```
havardpe, line 162: consider using a more specific exception, this could be caused by 'anything'

havardpe (approved):

searchcore/src/apps/vespa-proton-cmd/vespa-proton-cmd.cpp
baldersheim: I concluded that this would be fine for now as it is an improvement of something that should not really happen.
This is just to avoid the core dump and provide some hint at why it exited.
I added printing of exact exception info too.

baldersheim:
@arnej27959 Yes, this is just a stop gap to avoid the core dump.

PR #17958, opened 2021-05-25
Detect and work around null records

bratseth:
I'm asking the QuestDb devs about this but in the meantime this should work around the issue.


bratseth:
... hang on I'll do one additional check.

freva (approved):

PR #17978, opened 2021-05-26
Add a couple of stripe work TODOs

vekterli:
@geirst please review

- Ideal state ops cannot look at null-bucket messages for determining
  if full bucket checks are pending when running in striped mode, as
  these are not handled by stripes when not in legacy mode.
- State checker context should use ideal state cache instead of recomputing
  for every checked bucket (observed via `perf` in production).


geirst (approved):
👍

PR #17986, opened 2021-05-26
Set status failed when we do not get a deployment when bootstrapping …

hmusum:
…applications

If we for some reason `deployFromLocalActive` does not return a Deployment we should retry.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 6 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
```diff
@@ -226,13 +227,22 @@ private boolean redeployAllApplications() throws InterruptedException {
...
-                                 .ifPresent(Deployment::activate);
-            log.log(Level.INFO, () -> appId + " redeployed");
+            Optional<Deployment> deployment = applicationRepository.deployFromLocalActive(appId, true /* bootstrap */);
+            if (deployment.isPresent()) {
+                deployment.get().activate();
+                log.log(Level.INFO, () -> appId + " redeployed");
+                return DeploymentStatus.done;
+            } else {
+                // For some reason a deployment is not present, which should not be possible when bootstrapping
+                log.log(Level.INFO, () -> "Deployment failed for " + appId +
```
baldersheim, line 241: Should this be warning ?

baldersheim (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
hmusum: Definitely, thanks, fixed.

PR #17997, opened 2021-05-26
Expose number of retiring nodes in deployment log

freva:
Needed for VESPA-19930: Want to give some indicator to the tenant that the deployment may take a while if there is an ongoing retirement.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
@@ -95,7 +95,8 @@ class RunSerializer {
     private static final String lastTestRecordField = "lastTestRecord";
     private static final String lastVespaLogTimestampField = "lastVespaLogTimestamp";
     private static final String noNodesDownSinceField = "noNodesDownSince";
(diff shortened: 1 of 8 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
```diff
@@ -178,11 +180,10 @@ private ApplicationVersion applicationVersionFrom(Inspector versionObject) {
...
     // Don't change this — introduce a separate array instead.
     private Optional<ConvergenceSummary> convergenceSummaryFrom(Inspector summaryArray) {
-        if ( ! summaryArray.valid() || summaryArray.entries() == 11) // TODO jonmv: fix
-            return Optional.empty();
+        if ( ! summaryArray.valid()) return Optional.empty();
 
-        if (summaryArray.entries() != 12)
-            throw new IllegalArgumentException("Convergence summary must have 12 entries");
+        if (summaryArray.entries() != 12 && summaryArray.entries() != 13)
+            throw new IllegalArgumentException("Convergence summary must have 13 entries");
```
jonmv, line 186: XD

jonmv (approved):
:)

PR #18005, opened 2021-05-27
Load of tensor attribute with nearest neighbor index enabled but no

toregge:
saved index will index documents during the load. Add occasional calls
to commit() to prevent excessive growth of hold lists.

@geirst : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
@@ -207,7 +207,7 @@ DenseTensorAttribute::onLoad()
                 // This ensures that get_vector() (via getTensor()) is able to find the newly added tensor.
                 setCommittedDocIdLimit(lid + 1);
                 _index->add_document(lid);
-                if ((lid % (1u << 8)) == 0) {
+                if ((lid % 256) == 0) {
                     commit();
                 }
             }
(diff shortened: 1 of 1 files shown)


baldersheim (approved):

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -207,6 +207,9 @@ DenseTensorAttribute::onLoad()
                 // This ensures that get_vector() (via getTensor()) is able to find the newly added tensor.
                 setCommittedDocIdLimit(lid + 1);
                 _index->add_document(lid);
+                if ((lid & ((1u << 8) - 1)) == 0) {
```
geirst, line 210: Consider using modulo instead so it's easier to read.

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -207,7 +207,7 @@ DenseTensorAttribute::onLoad()
                 // This ensures that get_vector() (via getTensor()) is able to find the newly added tensor.
                 setCommittedDocIdLimit(lid + 1);
                 _index->add_document(lid);
-                if ((lid & ((1u << 8) - 1)) == 0) {
+                if ((lid % (1u << 8)) == 0) {
```
geirst, line 210:
```suggestion
                if ((lid % 256) == 0) {
```

geirst (approved):
👍  Let's simplify even more.

PR #18007, opened 2021-05-27
Log model build errors to the deploy log

bratseth:
This is to achieve two things:
- Log IllegalArgumentExceptions from model building to the deploy log.
- Return them as application. package errors in the REST API. (Which I hope will cause no retry to happen.)

I also cleaned up the error classification a bit to reflect the common practice that all exceptions probably caused by incorrect input should be IllegalArgumentException instances.


hmusum (approved):
Great!

PR #18010, opened 2021-05-27
Log instead of throwing on identical redefinition of rank-expressions.

baldersheim:
Redefinition to something else is still enforced.

@bratseth or @hmusum PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/RankExpressionFiles.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/RankExpressionFiles.java
@@ -12,7 +12,7 @@ import java.util.logging.Level;
 public class RankExpressionFiles {
     private final Map<String, RankExpressionFile> expressions = new HashMap<>();
 
-    //TODO Deploy logger should not be necessary, as redefinition is illegal, but legacy prevents enforment starting now.
+    //TODO Deploy logger should not be necessary, as redefinition is illegal, but legacy prevents enforcement starting now.
     public void add(RankExpressionFile expression, DeployLogger deployLogger) {
         expression.validate();
         String name = expression.getName();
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/searchdefinition/RankExpressionFiles.java
```diff
@@ -1,20 +1,32 @@
...
+    public void add(RankExpressionFile expression, DeployLogger deployLogger) {
         expression.validate();
         String name = expression.getName();
-        if (expressions.containsKey(name))
-            throw new IllegalArgumentException("Rank expression file '" + name + "' defined twice");
+        if (expressions.containsKey(name)) {
+            if ( expressions.get(name).getFileName().equals(expression.getFileName()) ) {
+                //TODO Throw instead, No later than Vespa 8
+                deployLogger.logApplicationPackage(Level.WARNING, "Rank expression file '" + name +
+                        "' defined twice with identical expression (illegal and will be enfoced soon) '" + expression.getFileName() + "'.");
```
hmusum, line 23:
```suggestion
                        "' defined twice with identical expression (illegal and will be enforced soon) '" + expression.getFileName() + "'.");
```

config-model/src/main/java/com/yahoo/searchdefinition/RankExpressionFiles.java
```diff
@@ -1,20 +1,32 @@
...
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.logging.Level;
 
 public class RankExpressionFiles {
     private final Map<String, RankExpressionFile> expressions = new HashMap<>();
 
-    public void add(RankExpressionFile expression) {
+    //TODO Deploy logger should not be necessary, as redefinition is illegal, but legacy prevents enforment starting now.
```
hmusum, line 15:
```suggestion
    //TODO Deploy logger should not be necessary, as redefinition is illegal, but legacy prevents enforcement starting now.
```

hmusum (approved):

PR #18011, opened 2021-05-27
Expose current cluster resource usage

freva:


bratseth (approved):
👍   I want to restructure this to use Load everywhere including in the JSON formats, but this is no worse than it was.

PR #18035, opened 2021-05-31
Defer encryption if host is upgrading Vespa

mpolden:
@hakonhall


hakonhall (approved):
This is to try to avoid blocking the convergence of a rollout?

mpolden:
Yes, the same mechanism is used for OS upgrades.

PR #18036, opened 2021-05-31
Arnej/prepare for more generic rpc

arnej27959:

@havardpe please review

note that there's no actual RPC call for checkConnectivity yet, this is only (most of) the underlying implementation


arnej27959:
@havardpe please take another look now

havardpe (approved):

PR #18040, opened 2021-05-31
Revert "Arnej/also subscribe on model"

arnej27959:
Reverts vespa-engine/vespa#18025

broke a system test


arnej27959:
@havardpe @hmusum FYI

PR #18047, opened 2021-06-01
add fetch of model config as a separate step

arnej27959:

@havardpe please review

### DIFF
--- a/configd/src/apps/sentinel/env.cpp
+++ b/configd/src/apps/sentinel/env.cpp
@@ -9,9 +9,12 @@
 
 LOG_SETUP(".env");
 
+using namespace std::chrono_literals;
+
 namespace config::sentinel {
 
-constexpr std::chrono::milliseconds CONFIG_TIMEOUT_MS(3 * 60 * 1000);
+constexpr std::chrono::milliseconds CONFIG_TIMEOUT_MS = 3min;
+constexpr std::chrono::milliseconds MODEL_TIMEOUT_MS = 1500ms;
 
 Env::Env()
   : _cfgOwner(),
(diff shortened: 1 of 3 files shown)


configd/src/apps/sentinel/env.cpp
```diff
@@ -9,9 +9,12 @@
 
 LOG_SETUP(".env");
 
+using namespace std::chrono_literals;
+
 namespace config::sentinel {
 
-constexpr std::chrono::milliseconds CONFIG_TIMEOUT_MS(3 * 60 * 1000);
+constexpr std::chrono::milliseconds CONFIG_TIMEOUT_MS = 3min;
+constexpr std::chrono::milliseconds MODEL_TIMEOUT_MS = 1500ms;
```
havardpe, line 17: consider using a bit more timeout

configd/src/apps/sentinel/env.cpp
```diff
@@ -38,6 +41,23 @@ void Env::boot(const std::string &configId) {
...
+        const ModelConfig &model = *up;
+        for (const auto & h : model.hosts) {
+            LOG(info, "- Model for host %s with %zd services", h.name.c_str(), h.services.size());
+            for (const auto & s : h.services) {       
+                if (s.name == "config-sentinel") { 
+                    LOG(info, "  - Model for service %s type %s configid %s with %zd ports",
+                        s.name.c_str(), s.type.c_str(), s.configid.c_str(), s.ports.size());
+                    for (const auto & p : s.ports) {
+                        if (p.tags.find("rpc") != p.tags.npos) {
+                            LOG(info, "    - Model for port %d has tags %s", p.number, p.tags.c_str());
```
havardpe, line 54: will spamming the log have any negative impact on production?

havardpe (approved):

configd/src/apps/sentinel/env.cpp
arnej27959: already did that :-)
I really wanted < 500ms to ensure this doesn't impact anything if the model config is unavailable.  Since this happens after the first config subscription is OK it should be fast anyway.

configd/src/apps/sentinel/env.cpp
arnej27959: considering the logging from other things that happen at startup (just metricsproxy logs > 40 lines) I'm not worried.  But this logging will be removed when we actually use the model config later, I guess - it's mostly a placeholder.

PR #18048, opened 2021-06-01
Do not block global merges to nodes tagged as busy [run-systemtest]

vekterli:
@geirst please review

To avoid starvation of high priority global bucket merges, we do not consider
these for blocking due to a node being "busy" (usually caused by a full merge
throttler queue).

This is for two reasons:
 1. When an ideal state op is blocked, it is still removed from the internal
    maintenance priority queue. This means a blocked high pri operation will
    not be retried until the next DB pass (at which point the node is likely
    to still be marked as busy when there's heavy merge traffic).
 2. Global bucket merges have high priority and will most likely be allowed
    to enter the merge throttler queues, displacing lower priority merges.

We may consider adding a separate busy-dimension for global merges at
some point if it's deemed worth the effort, but for now I think we'll go with
the simple solution.

### DIFF
--- a/storage/src/tests/distributor/mergeoperationtest.cpp
+++ b/storage/src/tests/distributor/mergeoperationtest.cpp
@@ -315,7 +315,7 @@ TEST_F(MergeOperationTest, allow_deleting_active_source_only_replica) {
               _sender.getLastCommand(true));
 }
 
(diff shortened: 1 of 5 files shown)


geirst (approved):
👍

vekterli:
Unrelated build failure:
```
[ERROR] Failures: 
[ERROR]   RoutingTestCase.requireThatRetryCanBeDisabled:357
```

PR #18050, opened 2021-06-01
Add validation of the number of distributor stripes from config and a…

geirst:
…dd more asserts.

This ensures the number of stripes is a power of 2 and within MaxStripes boundary.

@vekterli please review
@toregge FYI

### DIFF
--- a/storage/src/vespa/storage/common/bucket_stripe_utils.h
+++ b/storage/src/vespa/storage/common/bucket_stripe_utils.h
@@ -9,7 +9,7 @@
 namespace storage {
 
 constexpr static uint8_t MaxStripeBits = spi::BucketLimits::MinUsedBits;
-constexpr static size_t MaxStripes = (1ULL << MaxStripeBits);
+constexpr static uint32_t MaxStripes = (1ULL << MaxStripeBits);
 
 /**
  * Returns the stripe in which the given bucket key belongs,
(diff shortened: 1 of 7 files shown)


storage/src/vespa/storage/common/bucket_stripe_utils.h
```diff
@@ -22,7 +22,14 @@ size_t stripe_of_bucket_key(uint64_t key, uint8_t n_stripe_bits) noexcept;
...
  */
-uint8_t calc_num_stripe_bits(size_t n_stripes) noexcept;
+uint8_t calc_num_stripe_bits(uint32_t n_stripes) noexcept;
+
+/**
+ * Validates the number of stripes and returns the (potentially adjusted) value.
+ *
+ * This ensures the number of stripes is a power of 2 and within MaxStripes boundary.
+ */
+uint32_t validate_num_stripes(uint32_t n_stripes) noexcept;
```
vekterli, line 32: Consider adding `[[nodiscard]]` attribute

vekterli, line 32: To me, the prefix "validate" tends to imply that the function performs some explicit action if the validation fails (such as throwing an exception). This function communicates that this is not the case by being tagged `noexcept`, but I wonder if there's a more fitting name for it still. Perhaps just `adjusted_num_stripes` ref. the comment?

storageserver/src/vespa/storageserver/app/distributorprocess.cpp
```diff
@@ -32,6 +33,21 @@ DistributorProcess::shutdown()
     _node.reset();
 }
 
+namespace {
+
+uint32_t
+validate_num_distributor_stripes(uint32_t cfg_n_stripes)
```
vekterli, line 39: Consider if function should be slightly renamed here as well

vekterli (approved):
:+1:

storage/src/vespa/storage/common/bucket_stripe_utils.h
geirst: Fixed

geirst: I agree. Fixed.

storageserver/src/vespa/storageserver/app/distributorprocess.cpp
geirst: Fixed

geirst:
@vekterli PTAL

vekterli (approved):
🦄👍

PR #18051, opened 2021-06-01
Add support for ignorable exceptions when building config model

hmusum:
It's now possible to throw an exception that leads to a warning
being logged and visible for the customer, while the model will
be built again without throwing any exception.


bratseth:
I thought we agreed this isn't needed?

hmusum:
Well, in that case I am just closing this, as this is more complicated than wiring deploy logger to the place where we want to log the warning and just do that.

PR #18055, opened 2021-06-01
GC unused methods from the api.

baldersheim:
@hmusum PR


hmusum (approved):
As long as they aren't used by config models still in use

baldersheim:
Long gone

PR #18058, opened 2021-06-01
Don't fail nodes/v2 requests if QuestDb fails

bratseth:
Deployment jobs pools the application path, but does
not need information from QuestDb, so just return
without it if QuestDb fails.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/ApplicationSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/ApplicationSerializer.java
@@ -15,6 +15,7 @@ import com.yahoo.vespa.hosted.provision.autoscale.MetricsDb;
 
 import java.net.URI;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * Serializes application information for nodes/v2/application responses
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/ApplicationSerializer.java
```diff
@@ -71,12 +71,12 @@ private static void toSlime(Application application,
...
             cluster.suggestedResources().ifPresent(suggested -> toSlime(suggested.resources(), clusterObject.setObject("suggested")));
         cluster.targetResources().ifPresent(target -> toSlime(target, clusterObject.setObject("target")));
-        clusterUtilizationToSlime(clusterModel, clusterObject.setObject("utilization"));
+        clusterModel.ifPresent(model -> clusterUtilizationToSlime(model, clusterObject.setObject("utilization")));
         scalingEventsToSlime(cluster.scalingEvents(), clusterObject.setArray("scalingEvents"));
         clusterObject.setString("autoscalingStatus", cluster.autoscalingStatus());
-        clusterObject.setLong("scalingDuration", clusterModel.scalingDuration().toMillis());
-        clusterObject.setDouble("maxQueryGrowthRate", clusterModel.maxQueryGrowthRate());
-        clusterObject.setDouble("currentQueryFractionOfMax", clusterModel.queryFractionOfMax());
+        clusterModel.ifPresent(model -> clusterObject.setLong("scalingDuration", model.scalingDuration().toMillis()));
```
freva, line 77: This + the 2 field below will cause NPE in controller-api/ClusterData if not set

bratseth: thanks!

freva (approved):

PR #18060, opened 2021-06-01
Revert "Cluster controllers are not tied to content clusters anymore"

aressem:
Reverts vespa-engine/vespa#18056


hmusum:
Fixed in https://git.vzbuilders.com/vespa/vespa-yahoo/pull/6744

PR #18063, opened 2021-06-01
Deprovision children when encrypting host

mpolden:
@hakonhall


mpolden:
Not relevant after offline discussion.

PR #18068, opened 2021-06-01
Revert "Try to reduce test runtime for RoutingTestCase"

hmusum:
Reverts vespa-engine/vespa#18028

Got unstable, fails with timeout, ticket created


hmusum:
@aressem FYI

jonmv (approved):

jonmv:
I wonder if we could try this again with https://github.com/vespa-engine/vespa/pull/19292 which fixes a potential issue where you could get a dead service mirror from a thread-local cache, when networks have been shut down previously ...

jonmv:
Hmm, still getting timeouts locally, on that branch, there must be something more. Will investigate later.

PR #18070, opened 2021-06-01
drop empty buffers

havardpe:

@baldersheim please review
@arnej27959 FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/net/tls/maybe_tls_crypto_socket.cpp
+++ b/vespalib/src/vespa/vespalib/net/tls/maybe_tls_crypto_socket.cpp
@@ -77,12 +77,14 @@ public:
         if (frame > 0) {
             memcpy(buf, src.data, frame);
             _buffer.evict(frame);
+            _buffer.drop_if_empty();
         }
         return frame;
     }
     ssize_t write(const char *buf, size_t len) override { return _socket.write(buf, len); }
     ssize_t flush() override { return 0; }
     ssize_t half_close() override { return _socket.half_close(); }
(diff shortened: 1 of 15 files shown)


vespalib/src/vespa/vespalib/net/tls/maybe_tls_crypto_socket.cpp
```diff
@@ -77,12 +77,14 @@ class MyCryptoSocket : public CryptoSocket
         if (frame > 0) {
             memcpy(buf, src.data, frame);
             _buffer.evict(frame);
+            _buffer.drop_if_empty();
```
arnej27959, line 80: should not this `drop_if_empty` be in `drop_empty_buffers()` below?

baldersheim (approved):
Looks good so far. Note the comment from @arnej27959

vespalib/src/vespa/vespalib/net/tls/maybe_tls_crypto_socket.cpp
havardpe: Actually, no. This is to discard the TLS snoop buffer after it has been drained in the non-tls case. We do direct read/write to the socket in this mode, and do not need the 4_Ki buffer used for snooping during handshaking.

baldersheim:
Wrong button ? close != merge

havardpe:
I got a merge conflict when I tried to merge, then I rebased on top of master and did a force push, which gave me a 0/0 branch since it was already merged. Github API has signed the merge commit, making it 'verified', but will not take responsibility for it and claims it was merged in some other way or commited directly to the master branch...

PR #18074, opened 2021-06-01
Only use tensor store memory usage when considering compaction.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #18076, opened 2021-06-01
Add proof of concept support for multiple distributor stripes.

geirst:
The most basic functionality is now supported using multiple distributor stripes (and threads).

Note that the following is (at least) still missing:
* Stripe-separate metrics with top-level aggregation.
* Aggregation over all stripes in misc functions in Distributor that currently is using the first stripe.
* Handling of messages without bucket id in the top-level Distributor instead of using the first stripe.

@vekterli please review
@toregge @baldersheim FYI

### DIFF
--- a/storage/src/vespa/storage/distributor/distributor.cpp
+++ b/storage/src/vespa/storage/distributor/distributor.cpp
@@ -23,6 +23,7 @@
 #include <vespa/storage/common/nodestateupdater.h>
 #include <vespa/storage/config/distributorconfiguration.h>
 #include <vespa/storage/distributor/maintenance/simplebucketprioritydatabase.h>
+#include <vespa/storageapi/message/persistence.h>
 #include <vespa/storageframework/generic/status/xmlstatusreporter.h>
 #include <vespa/vdslib/distribution/distribution.h>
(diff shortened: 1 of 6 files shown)


storage/src/vespa/storage/distributor/distributor.cpp
```diff
@@ -321,6 +330,43 @@ bool should_be_handled_by_top_level_bucket_db_updater(const api::StorageMessage&
...
+                if (!reply.getBucketInfo().empty()) {
+                    // Note: All bucket ids in this reply belong to the same distributor stripe, so we just use the first entry.
+                    return reply.getBucketInfo()[0]._bucketId;
+                } else {
+                    return reply.getBucketId();
+                }
+            }
+            default:
+                return msg.getBucketId();
+        }
```
vekterli, line 355: Handling of visitor-related requests will probably have to be added here, since e.g. `CreateVisitorCommand` does not have an explicit single bucket ID, and they must be routed to the stripe that owns the superbucket.

vekterli (approved):
:+1: a couple of comments that can be addressed in subsequent work

PR #18080, opened 2021-06-01
Balder/do not use file as trigger

baldersheim:
@bratseth or @hmusum or @lesters


config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
```diff
@@ -54,36 +54,54 @@
 
     public final static String FIRST_PHASE = "firstphase";
     public final static String SECOND_PHASE = "secondphase";
-    /** The search definition-unique name of this rank profile */
+    /**
```
bratseth, line 57: You ran some kind of automated reformatting here? Please don't do that ...

bratseth (approved):
This seems better as far as I can tell

config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
baldersheim: that was not intended. Will undo.

PR #18096, opened 2021-06-02
Use a hash map for specs. If the request is a point lookup then just …

baldersheim:
…use a hash lookup.

If it is a wildcard lookup iterate as earlier on.
Also use vespalib::stringref in interface to avoid conversion.
Use vespalib:string in the hash map to locate string in object aswe are still on old abi.

@havardpe and @arnej27959 PR

### DIFF
--- a/slobrok/src/vespa/slobrok/sbmirror.cpp
+++ b/slobrok/src/vespa/slobrok/sbmirror.cpp
@@ -4,6 +4,7 @@
 #include <vespa/fnet/frt/supervisor.h>
 #include <vespa/fnet/frt/target.h>
 #include <vespa/vespalib/util/exceptions.h>
+#include <vespa/vespalib/stllike/hash_map.hpp>
 
 #include <vespa/log/log.h>
 LOG_SETUP(".slobrok.mirror");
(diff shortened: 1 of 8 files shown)


arnej27959:
I don't see any gain from not using std::string and std::unordered_map here.
Note that the return value from lookup() still uses std::string so there will need to be conversion anyway.
I'm not sure what old ABI you're talking about, we stopped using reference-counted std::string years ago?

baldersheim:
Yes, we did as long as long as we build our own toolchain. But that stopped on RHEL7/CentOS7.
So we are still using refcounted strings. We must wait for RHEL8 to get usable strings.

arnej27959:
> We must wait for RHEL8 to get usable strings.

NOOOOOO !!!!
bleh.  In that case we should replace all std::string with vespalib::string in all of slobrok,
I will do that in a separate PR.

havardpe:
> Yes, we did as long as long as we build our own toolchain. But that stopped on RHEL7/CentOS7.
> So we are still using refcounted strings. We must wait for RHEL8 to get usable strings.

This would also mean that search performance is suffering (again) due to protobuf using std::string for things like document id...

slobrok/src/vespa/slobrok/sbmirror.cpp
```diff
@@ -167,34 +174,26 @@ MirrorAPI::handleIncrementalFetch()
...
+        updateTo(std::move(specs), diff_to);
     } else if (_specsGen == diff_from) {
         // incremental update
-        SpecList specs;
-        SpecList::const_iterator end = _specs.end();
-        for (SpecList::const_iterator it = _specs.begin();
-             it != end;
-             ++it)
-        {
+        SpecMap specs;
```
havardpe, line 184: might be faster to just copy and modify based on diff now that we have a map...

PR #18097, opened 2021-06-02
Use proton metrics for memory and disk utilization

bratseth:
Since these metrics are use for block decisins they should also be
used for autoscaling.

### DIFF
--- a/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsV2MetricsFetcherTest.java
+++ b/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsV2MetricsFetcherTest.java
@@ -161,8 +161,8 @@ public class MetricsV2MetricsFetcherTest {
             "            },\n" +
             "            {\n" +
             "              \"values\": {\n" +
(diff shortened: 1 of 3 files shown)


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsV2MetricsFetcherTest.java
```diff
@@ -161,6 +161,14 @@ public void close() { }
             "            },\n" +
             "            {\n" +
             "              \"values\": {\n" +
+            "                \"content.proton.resource_usage.memory.average\": 35,\n" +
```
freva, line 164: These metrics values are between 0 and 1

freva (approved):
Need to rescale the content node metrics

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsV2MetricsFetcherTest.java
bratseth: thanks!

PR #18101, opened 2021-06-03
Use default node resources when none are specified

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Limits.java
```diff
@@ -56,6 +59,16 @@ public NodeResources cap(NodeResources resources) {
...
     }
 
+    public Limits fullySpecified(ClusterSpec.Type type, NodeRepository nodeRepository) {
+        if (this.isEmpty()) return this;
+        CapacityPolicies capacityPolicies = new CapacityPolicies(nodeRepository);
+        var specifiedMin = min.with(min.nodeResources().isUnspecified() ?
+                                    capacityPolicies.defaultNodeResources(type) : min.nodeResources());
+        var specifiedMax = max.with(max.nodeResources().isUnspecified() ?
+                                    capacityPolicies.defaultNodeResources(type) : max.nodeResources());
+        return new Limits(specifiedMin, specifiedMax);
```
freva, line 69: Consider simplifying by checking if min is specified first
```suggestion
        if (isEmpty() || !min.nodeResources().isUnspecified()) return this;
        NodeResources defaultResources = new CapacityPolicies(nodeRepository).defaultNodeResources(type);
        return new Limits(min.with(defaultResources), max.with(defaultResources));
```

freva (approved):

PR #18108, opened 2021-06-03
Bratseth/unspecified resources

bratseth:
Thanks for the suggestion, didn't see it before merge.
(I want to keep this symmetric though)


freva (approved):
FWIW, we already assume that if min is unspecified, then so is max (in the method above)

PR #18125, opened 2021-06-04
Add benchmark for non POD objects.

baldersheim:
@vekterli PR

### DIFF
--- a/vespalib/src/tests/stllike/hashtable_test.cpp
+++ b/vespalib/src/tests/stllike/hashtable_test.cpp
@@ -5,9 +5,9 @@
 #include <vespa/vespalib/stllike/hash_fun.h>
 #include <vespa/vespalib/stllike/identity.h>
 #include <vespa/vespalib/testkit/testapp.h>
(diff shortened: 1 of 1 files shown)


vespalib/src/tests/stllike/hashtable_test.cpp
```diff
@@ -155,6 +155,32 @@ TEST("benchmark clear") {
...
+    ~NonPOD() { _v = 0; }
+private:
+    uint32_t _v;
+};
+
+/**
+ * Performance is identical for NonPOD objects as with POD object.
+ * Object are are only constructed on insert, and destructed on erase/clear.
+ */
+TEST("benchmark hash table reconstruction with POD objects") {
```
vekterli, line 170: This should probably be "with _non_-POD objects"

vekterli (approved):
:+1:

vespalib/src/tests/stllike/hashtable_test.cpp
baldersheim: Fixed

PR #18131, opened 2021-06-04
More autoscaling status information

bratseth:


freva (approved):
:100:

freva:
Need to update https://github.com/vespa-engine/vespa/blob/b8ebb1b462b2743ffe9d0b2266eb7009b577ab92/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/integration/ConfigServerMock.java#L125

PR #18135, opened 2021-06-04
Bjorncs/vespa feed client in vespa hadoop

bjorncs:

### DIFF
--- a/vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
+++ b/vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
@@ -1,9 +1,13 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hadoop.mapreduce;
 
+import ai.vespa.feed.client.DocumentId;
+import ai.vespa.feed.client.DryrunResult;
 import ai.vespa.feed.client.FeedClient;
 import ai.vespa.feed.client.FeedClientBuilder;
 import ai.vespa.feed.client.JsonStreamFeeder;
+import ai.vespa.feed.client.OperationParameters;
+import ai.vespa.feed.client.Result;
 import com.yahoo.vespa.hadoop.mapreduce.util.VespaConfiguration;
 import com.yahoo.vespa.hadoop.mapreduce.util.VespaCounters;
 import com.yahoo.vespa.http.client.config.FeedParams;
@@ -17,6 +21,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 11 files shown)


vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
```diff
@@ -0,0 +1,146 @@
...
+        feeder = createJsonStreamFeeder();
+        initialized = true;
+    }
+
+    private void validateConfig() {
+        if (!config.useSSL()) {
+            throw new IllegalArgumentException("SSL is required for this feed client implementation");
+        }
+        if (config.dryrun()) {
+            throw new IllegalArgumentException("Dryrun is not supported for this feed client implementation");
```
jonmv: We should probably add a dry-run option.

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        if (config.dataFormat() != FeedParams.DataFormat.JSON_UTF8) {
+            throw new IllegalArgumentException("Only JSON is support by this feed client implementation");
+        }
+        if (config.proxyHost() != null) {
+            log.warning(String.format("Ignoring proxy config (host='%s', port=%d)", config.proxyHost(), config.proxyPort()));
```
jonmv, line 102: Hmm, do we need to support this?

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
```diff
@@ -0,0 +1,146 @@
...
+
+    protected VespaRecordWriter(VespaConfiguration config, VespaCounters counters) {
+        this.counters = counters;
+        this.config = config;
+    }
+
+    @Override
+    public void write(Object key, Object data) throws IOException {
+        initializeOnFirstWrite();
+        String json = "[" + data.toString().trim() + "]";
```
jonmv, line 53: We could probably let the JSON parser accept a series of JSON objects, too, in addition to a JSON array.

jonmv: (Alternatively, `SequenceInputStream` can be used here :))

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
```diff
@@ -143,4 +154,28 @@ private static int streamsPerConnection(VespaConfiguration config) {
                 .map(hostname -> URI.create(String.format("https://%s:%d/", hostname, config.defaultPort())))
                 .collect(toList());
     }
+
+    private static class DryrunClient implements FeedClient {
```
jonmv, line 158: Let's move this to the json parser package. I think we should offer some dry-run option for just verifying the JSON, counting documents, etc., there.

jonmv:
Nice :)

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
bjorncs: I'll add this once we are settled on a http client implementation

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
bjorncs: We are not sure yet - current theory is that it's not necessary. We'll have to try it out in CI.

vespa-hadoop/src/main/java/com/yahoo/vespa/hadoop/mapreduce/VespaRecordWriter.java
bjorncs: Added in b525d31

bjorncs:
Replaced by https://github.com/vespa-engine/vespa/pull/18181

PR #18138, opened 2021-06-04
Revert "Searchdefinition/schemas cleanup"

aressem:
Reverts vespa-engine/vespa#18129

This seems to have introduced some sd validation errors , but only on RHEL 8 ? : 
https://factory.vespa.oath.cloud/testrun/29784417


hmusum:
Just some changes to error message, I'll revert this and fix system tests later

PR #18145, opened 2021-06-06
Balder/add merge throttler metrics

baldersheim:
@vekterli or @geirst PR

### DIFF
--- a/storage/src/vespa/storage/storageserver/mergethrottler.cpp
+++ b/storage/src/vespa/storage/storageserver/mergethrottler.cpp
@@ -416,7 +416,7 @@ MergeThrottler::enqueueMerge(
     if (!validateNewMerge(mergeCmd, nodeSeq, msgGuard)) {
         return;
     }
-    _metrics->queueSize.addValue(_queue.size());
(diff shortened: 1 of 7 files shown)


storage/src/vespa/storage/storageserver/mergethrottler.cpp
```diff
@@ -415,6 +416,7 @@ MergeThrottler::enqueueMerge(
     if (!validateNewMerge(mergeCmd, nodeSeq, msgGuard)) {
         return;
     }
+    _metrics->queueSize.addValue(_queue.size());
```
vekterli, line 419: Since queue size is conceptually a gauge-style metric, it feels more intuitive to me if we call `set()` instead of `addValue()`. The former is just an alias for the latter, so the end result should be the same

vekterli (approved):
:+1:

storage/src/vespa/storage/storageserver/mergethrottler.cpp
baldersheim: K, fixed

geirst (commented):
👍

PR #18152, opened 2021-06-07
Add redundancy increase validation override

bratseth:


hmusum (approved):

hmusum (changes requested):
Need to update abi spec:
 Failed to execute goal com.yahoo.vespa:abi-check-plugin:7-SNAPSHOT:abicheck (default) on project config-model-api: ABI spec mismatch. To update run 'mvn package -Dabicheck.writeSpec'

hmusum (approved):

PR #18154, opened 2021-06-07
Validate redundancy=1 on first deployment

bratseth:


hmusum (changes requested):
I think this will break system tests (default environment is 'production`) and integration tests (most tests are run in production environment). Consider doing this for hosted only, as well, I am not sure how this will affect non-hosted customers and their testing, roullout etc.

bratseth:
Thanks! PTAL again @hmusum

hmusum (approved):

PR #18155, opened 2021-06-07
Revert "Add redundancy increase validation override"

hmusum:
Reverts vespa-engine/vespa#18152

Breaks some system tests

NB! config-model-api changes cannot be reverted, so some more work in this PR is needed before it can be merged


bratseth:
I can just fix those tests ...

bratseth:
Allow redundancy increase in these tests: https://github.com/vespa-engine/system-test/pull/1667

hmusum:
Fixing system tests worked, closing

PR #18163, opened 2021-06-08
Warn instead of failing on validation errors in manual zones

bratseth:
The purpose of validation overrides is to protect production from
destructive changes by requiring a confirmation for such changes.
This is unnecessary in manuall deployed environments.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
@@ -177,6 +177,7 @@ public class ValidationOverrides {
 
     }
 
+    // TODO: Remove this class after June 2021
     public static class AllowAllValidationOverrides extends ValidationOverrides {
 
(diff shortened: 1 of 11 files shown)


hmusum:
Seems like we can remove `ValidationOverrides.AllowAllValidationOverrides` and the code in `DeployState` which uses `AllowAllValidationOverrides` for manually deployed zones with this?

bratseth:
Aha! I vaguely recalled we were supposed to have this but couldn't find it where I expected. This one is fine, so just closing then.

hmusum:
I like this way of doing it much better, though, exactly because it's impossible to find the code that does this (or understand that something like this is actually happening) with the way it is done today.

bratseth:
Fair point. Let me look at it once more ...

bratseth:
Now I've removed (step 1 of 2) the old way of doing it.

hmusum (approved):
💯

hmusum (approved):

PR #18164, opened 2021-06-08
Fix: set idle timeout based on config

bjorncs:
FYI @jonmv


tokle (approved):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/ConnectorFactory.java
```diff
@@ -75,7 +75,7 @@ public ServerConnector createConnector(final Metric metric, final Server server,
         connector.setName(connectorConfig.name());
         connector.setAcceptQueueSize(connectorConfig.acceptQueueSize());
         connector.setReuseAddress(connectorConfig.reuseAddress());
-        connector.setIdleTimeout(toMillis(connector.getIdleTimeout()));
+        connector.setIdleTimeout(toMillis(connectorConfig.idleTimeout()));
```
baldersheim, line 78: Perhaps a unit test to verify that config is actually propagated correctly ?

bjorncs: https://github.com/vespa-engine/vespa/pull/18182

PR #18167, opened 2021-06-08
Move files from searchdefinitions/ to schemas/ when deploying

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -36,7 +36,6 @@ import com.yahoo.vespa.config.server.modelfactory.ModelFactoryRegistry;
 import com.yahoo.vespa.config.server.monitoring.MetricUpdater;
 import com.yahoo.vespa.config.server.monitoring.Metrics;
 import com.yahoo.vespa.config.server.provision.HostProvisionerProvider;
-import com.yahoo.vespa.config.server.tenant.TenantListener;
 import com.yahoo.vespa.config.server.tenant.TenantRepository;
 import com.yahoo.vespa.config.server.zookeeper.ConfigCurator;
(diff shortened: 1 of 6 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -685,6 +687,24 @@ private void copyApp(File sourceDir, File destinationDir) throws IOException {
         }
     }
 
+    // TODO: Remove in Vespa 8 (when we don't allow files in SEARCH_DEFINITIONS_DIR)
+    // Copies schemas from searchdefinitions/ to schemas/ if searchdefinitions/ exists
+    private void copyToSchemasDir(java.nio.file.Path applicationDir) throws IOException {
```
baldersheim, line 692: Is this code run in the prepare stage, or in activate ?
As long as it does not change set of files between the phases is good.

baldersheim (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: It's done in the prepare stage for an external deploy, so after that nothing should change that deployment. I guess it could happen while upgrading, and for old applications that are distributed between config servers... Will look into it

hmusum:
@baldersheim PTAL again

baldersheim (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
baldersheim: Are we giving a warning when people are using searchdefinitions ?

hmusum: Yes, we are giving a warning

PR #18169, opened 2021-06-08
Preserve wantToRebuild until host is readied

mpolden:
This prevents starting too many rebuilds if hosts never transition out of
provisioned after rebuild.

This also ensures that we retry rebuilds until they succeed (host becomes
ready).

@hakonhall

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -167,7 +167,7 @@ public class Nodes {
                     if (rebuilding) {
                         node = node.with(node.status().withWantToRetire(existing.get().status().wantToRetire(),
                                                                         false,
-                                                                        existing.get().status().wantToRebuild()));
+                                                                        rebuilding));
                     }
                     nodesToRemove.add(existing.get());
                 }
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -161,6 +161,14 @@ public boolean isWorking() {
                     node = node.with(node.status().withFailCount(existing.get().status().failCount()));
                     if (existing.get().status().firmwareVerifiedAt().isPresent())
                         node = node.with(node.status().withFirmwareVerifiedAt(existing.get().status().firmwareVerifiedAt().get()));
+                    // Preserve wantToRebuild/wantToRetire when rebuilding as the fields shouldn't be cleared until the
+                    // host is readied (i.e. we know it is up and rebuild completed)
+                    boolean rebuilding = existing.get().status().wantToRebuild();
+                    if (rebuilding) {
+                        node = node.with(node.status().withWantToRetire(existing.get().status().wantToRetire(),
+                                                                        false,
+                                                                        existing.get().status().wantToRebuild()));
```
hakonhall, line 170: could use `rebuilding` here

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
hakonhall: which is `true`

PR #18174, opened 2021-06-09
Route CreateVisitorCommand with too few used bits in the super bucket…

geirst:
… id to a random distributor stripe.

Such commands will eventually be bounced with WRONG_DISTRIBUTION when handled by the stripe.

@vekterli please review
@toregge FYI


vekterli (approved):
:+1:

PR #18178, opened 2021-06-09
Allow -Xmx to be specified as an argument to vespa-visit

bratseth:


vespaclient-java/src/main/sh/vespa-visit.sh
```diff
@@ -74,16 +74,30 @@ findhost
...
+fi
+
 if [ "${VESPA_LOG_LEVEL}" = "" ]; then
     export VESPA_LOG_LEVEL=error,warning
 fi
 
-export MALLOC_ARENA_MAX=1 #Does not need fast allocation
-exec java \
+export MALLOC_ARENA_MAX=1 # Does not need fast allocation
+echo exec java \
```
arnej27959, line 96: looks like you forgot to remove echo here

bratseth: Thanks, fixed.

arnej27959 (approved):
taking a new look I saw some accidentally duplicated lines, so I pushed a commit with slight simplification.

bratseth:
Great!

PR #18179, opened 2021-06-09
Block ideal state ops when a pending cluster state is present [run-systemtest]

vekterli:
@geirst please review

Since distributor stripes no longer have access to the top-level
pending message tracking info, it's no longer possible to infer if
a pending cluster state is happening by looking at the sent messages.
Instead, do this more generally (and efficiently) by looking at the
potential pending cluster state directly.

Rewire the `isBlocked` logic to take in an operation context instead
of just a `PendingMessageTracker`, giving it access to a lot more
relevant information.


geirst (approved):
👍

geirst:
@toregge FYI

PR #18184, opened 2021-06-09
Aggregate metrics from distributor stripes.

toregge:
@geirst : please review
@vekterli : FYI

### DIFF
--- a/storage/src/vespa/storage/distributor/distributor.cpp
+++ b/storage/src/vespa/storage/distributor/distributor.cpp
@@ -38,6 +38,30 @@ using namespace std::chrono_literals;
 
 namespace storage::distributor {
 
+namespace {
+
+std::shared_ptr<DistributorMetricSet>
+make_metrics_sum(uint32_t num_distributor_stripes)
+{
+    if (num_distributor_stripes == 0) {
+        return {};
+    }
+    return std::make_shared<DistributorMetricSet>();
+}
+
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/distributor/distributor.cpp
```diff
@@ -558,7 +584,12 @@ Distributor::propagateInternalScanMetricsToExternal()
     if (_use_legacy_mode) {
         _stripe->propagateInternalScanMetricsToExternal();
     } else {
-        first_stripe().propagateInternalScanMetricsToExternal();
+        _metrics_sum->reset();
```
geirst, line 587: @vekterli should comment on how this interacts with __metrics_sum_ also being the metric set that is registered with the component (ref L120).

geirst (approved):
👍  @vekterli please review as well.

storage/src/vespa/storage/distributor/distributor.cpp
vekterli: Resetting here (i.e. effectively recursively zeroing all child metrics) for `_metrics_sum` should be fine since this function is called in the context of a metric update callback, so there shouldn't be any concurrent meddling from the metric manager.

However, the biggest head-scratcher is how the underlying stripe metrics will interact with periodic metric snapshotting, since I believe it needs to do implicit zeroing of registered metrics to handle deltas. We might need an offline discussion on this. 🎷🦖

PR #18187, opened 2021-06-09
Use stats aggregated by client

jonmv:
@bjorncs please review and merge. Current implementation misses the call to `BenchmarkAggregator.start()`, so gives random time spent XD

### DIFF
--- a/vespa-feed-client/src/main/java/ai/vespa/feed/client/HttpRequest.java
+++ b/vespa-feed-client/src/main/java/ai/vespa/feed/client/HttpRequest.java
@@ -1,3 +1,42 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
-package ai.vespa.feed.client;public class HttpRequest {
+package ai.vespa.feed.client;
+
+import java.util.Map;
+import java.util.function.Supplier;
+
+class HttpRequest {
+
+    private final String method;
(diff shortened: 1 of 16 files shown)


vespa-feed-client/src/main/java/ai/vespa/feed/client/HttpRequest.java
```diff
@@ -0,0 +1,3 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
```
bjorncs, line 1: ????

bjorncs (approved):

vespa-feed-client/src/main/java/ai/vespa/feed/client/HttpRequest.java
jonmv: Hmm, what's wrong with this one? Thought it was the most recent? >_< IntelliJ uses that as a template for me, so I don't really notice when these appear.

bjorncs: I was wondering if this class was added intentionally or not

bjorncs: I probably looked at the code prior to force push.

PR #18193, opened 2021-06-10
Rewrite TSDB query and proxy requests to Horizon

freva:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/horizon/HorizonClient.java
```diff
@@ -9,7 +9,7 @@
  */
 public interface HorizonClient {
 
-    InputStream getMetrics(Slime query);
+    InputStream getMetrics(byte[] query);
```
freva, line 12: Not possible to overwrite/replace/remove object with Slime, so had to do it with Jackson...

olaaun (approved):

PR #18195, opened 2021-06-10
Add ParkedExpirer

olaaun:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
@@ -1,0 +1,62 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.provision.maintenance;
+
+import com.yahoo.config.provision.NodeType;
+import com.yahoo.jdisc.Metric;
+import com.yahoo.vespa.hosted.provision.Node;
+import com.yahoo.vespa.hosted.provision.NodeRepository;
+import com.yahoo.vespa.hosted.provision.node.Agent;
+import com.yahoo.vespa.hosted.provision.node.History;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.logging.Logger;
+
+/**
+ *
+ * Expires parked nodes in dynamically provisioned zones.
+ * If number of parked hosts exceed MAX_ALLOWED_PARKED_HOSTS, recycle in a queue order
+ *
+ * @author olaa
+ */
+public class ParkedExpirer extends NodeRepositoryMaintainer {
+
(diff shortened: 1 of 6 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,62 @@
...
+    protected double maintain() {
+        if (!nodeRepository.zone().getCloud().dynamicProvisioning())
+            return 1.0;
+
+        var parkedHosts = new ArrayList<>(nodeRepository.nodes().list(Node.State.parked)
+                                                        .nodeType(NodeType.host)
+                                                        .asList());
+
+        parkedHosts.sort(Comparator.comparing(this::getParkedTime));
+        var hostsToExpire = parkedHosts.size() - MAX_ALLOWED_PARKED_HOSTS;
```
mpolden, line 47: This should be `Math.max(0, parkedHosts.size() - MAX_ALLOWED_PARKED_HOSTS)` in case there are less than 20 hosts parked.

Also, using `var` for `int` just reduces readability imo.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    protected double maintain() {
+        if (!nodeRepository.zone().getCloud().dynamicProvisioning())
+            return 1.0;
+
+        var parkedHosts = new ArrayList<>(nodeRepository.nodes().list(Node.State.parked)
+                                                        .nodeType(NodeType.host)
+                                                        .asList());
```
mpolden, line 44: NodeList has `sortedBy` so you could simplify this to something like:
```
nodeRepository.nodes().list(Node.State.parked).nodeType(NodeType.host)
                      .sortedBy(this::getParkedTime)
                      .first(hostsToExpire)
                      .forEach(host -> { ... })
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Agent.java
```diff
@@ -21,6 +21,7 @@
     InactiveExpirer,
     ProvisionedExpirer,
     ReservationExpirer,
+    ParkedExpirer,
```
mpolden, line 24: Must be added to `com.yahoo.vespa.hosted.controller.api.integration.noderepository.NodeHistory` as well.

mpolden (changes requested):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Agent.java
mpolden: You can disregard this as #18201 should take care of it.

olaaun: It was already there

mpolden (approved):

PR #18197, opened 2021-06-10
add rpc method for connectivity report

arnej27959:

@havardpe another for discuss / review

### DIFF
--- a/configd/src/apps/sentinel/env.cpp
+++ b/configd/src/apps/sentinel/env.cpp
@@ -36,7 +36,7 @@ constexpr int maxConnectivityRetries = 100;
 
 Env::Env()
   : _cfgOwner(),
-    _modelSubscriber("admin/model"),
+    _modelOwner("admin/model"),
     _rpcCommandQueue(),
     _rpcServer(),
     _stateApi(),
@@ -53,7 +53,7 @@ Env::~Env() = default;
 void Env::boot(const std::string &configId) {
     LOG(debug, "Reading configuration for ID: %s", configId.c_str());
     _cfgOwner.subscribe(configId, CONFIG_TIMEOUT_MS);
-    _modelSubscriber.start(CONFIG_TIMEOUT_MS);
+    _modelOwner.start(CONFIG_TIMEOUT_MS);
     // subscribe() should throw if something is not OK
     Connectivity checker;
     for (int retry = 0; retry < maxConnectivityRetries; ++retry) {
@@ -66,7 +66,7 @@ void Env::boot(const std::string &configId) {
                 configId.c_str(), cfg.port.telnet, cfg.port.rpc);
             rpcPort(cfg.port.rpc);
             statePort(cfg.port.telnet);
-            auto model = _modelSubscriber.getModelConfig();
+            auto model = _modelOwner.getModelConfig();
             if (model.has_value()) {
                 checker.configure(cfg.connectivity, model.value());
             }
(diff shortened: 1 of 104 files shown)


configd/src/apps/sentinel/rpchooks.cpp
```diff
@@ -53,6 +54,12 @@ RPCHooks::initRPC(FRT_Supervisor *supervisor)
     rb.ParamDesc("timeout", "Timeout for check in milliseconds");
     rb.ReturnDesc("status", "Status (ok, bad, or unknown) for peer");
     //-------------------------------------------------------------------------
+    rb.DefineMethod("sentinel.report.connectivity", "", "SS",
+                    FRT_METHOD(RPCHooks::rpc_reportConnectivity), this);
+    rb.MethodDesc("report connectivity for peer sentinels");
```
havardpe, line 59: consider passing ping timeout here as well

configd/src/apps/sentinel/report-connectivity.h
```diff
@@ -0,0 +1,63 @@
...
+        status("unknown"),
+        check(nullptr)
+    {}
+
+    SinglePing(SinglePing &&) = default;
+    SinglePing(const SinglePing &) = default;
+
+    virtual ~SinglePing();
+    void startCheck(FRT_Supervisor &orb);
+    void returnStatus(bool ok) override;
```
havardpe, line 41: consider extending callback with hostname to avoid needing multiple ping classes

configd/src/apps/sentinel/report-connectivity.cpp
```diff
@@ -0,0 +1,86 @@
...
+  : _parentRequest(req),
+    _orb(orb),
+    _result(),
+    _configFetcher()
+{
+    try {
+        _configFetcher.subscribe<ModelConfig>("admin/model", this, 2000ms);
+        _configFetcher.start();
+        return;
+    } catch (ConfigTimeoutException & ex) {
```
havardpe, line 38: we should not block the network thread; consider a separate service/thread that keeps the model config updated that we can get a snapshot from.

havardpe (commented):
I think model config is the pain point

configd/src/apps/sentinel/env.cpp
```diff
@@ -116,7 +121,7 @@ void Env::statePort(int port) {
 void Env::notifyConfigUpdated() {
     vespalib::ComponentConfigProducer::Config current("sentinel", _cfgOwner.getGeneration(), "ok");
     _stateApi.myComponents.addConfig(current);
-
+    _modelOwner.checkForUpdates();
```
havardpe, line 124: consider moving this call to main event loop

configd/src/apps/sentinel/model-owner.h
```diff
@@ -0,0 +1,30 @@
...
+ **/
+class ModelOwner {
+private:
+    std::string _configId;
+    config::ConfigSubscriber _subscriber;
+    config::ConfigHandle<ModelConfig>::UP _modelHandle;
+    std::unique_ptr<ModelConfig> _modelConfig;
+public:
+    ModelOwner(const std::string &configId);
+    virtual ~ModelOwner();
```
havardpe, line 24: why is this virtual?

configd/src/apps/sentinel/model-owner.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <chrono>
+#include <vespa/log/log.h>
+
+LOG_SETUP(".sentinel.model-owner");
+
+using namespace std::chrono_literals;
+
+namespace config::sentinel {
+
+std::optional<ModelConfig> ModelOwner::getModelConfig() {
```
havardpe, line 16: do we need some locking to make sure we do not race with transport thread(s)?

configd/src/apps/sentinel/model-owner.cpp
```diff
@@ -0,0 +1,59 @@
...
+    checkForUpdates();
```
havardpe, line 17: consider just doing this in the main event loop, and also having a lock around the snapshot itself to synchronize between readers and writer

havardpe (commented):
some minor thread issues, otherwise it looks good

havardpe (approved):

PR #18199, opened 2021-06-10
Implement aggregation across distributor stripes for min replica, buc…

geirst:
…ket spaces, and pending maintenance stats.

@vekterli and @toregge please review

### DIFF
--- a/storage/src/vespa/storage/distributor/min_replica_provider.cpp
+++ b/storage/src/vespa/storage/distributor/min_replica_provider.cpp
@@ -1,0 +1,17 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "min_replica_provider.h"
+
+namespace storage::distributor {
+
+void
+merge_min_replica_stats(std::unordered_map<uint16_t, uint32_t>& dest,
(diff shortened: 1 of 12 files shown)


toregge (approved):

storage/src/vespa/storage/distributor/min_replica_provider.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+namespace storage::distributor {
+
+void
+merge_min_replica_stats(std::unordered_map<uint16_t, uint32_t>& dest,
+                        const std::unordered_map<uint16_t, uint32_t>& src)
+{
+    for (const auto& entry : src) {
+        auto node_index = entry.first;
+        dest[node_index] += entry.second;
```
vekterli, line 13: MinReplica counts the minimum bucket replication factor across all buckets for a given node, so this should probably be `std::min` across all stripes instead of the sum

vekterli (commented):
One comment on min replica semantics, otherwise LGTM

storage/src/vespa/storage/distributor/min_replica_provider.cpp
geirst: Thanks, fixed.

PR #18023, opened 2021-05-28
Throw exception if resource-limits is specified

hmusum:
Please review only, cannot be merged until no customers in hosted specify `resource-limits` in services.xml

PR #18205, opened 2021-06-11
Add back logging when resource-limits is specified and we do not throw

hmusum:
Add back logging so that apps that still have this setting get a
notification

Removing logging in https://github.com/vespa-engine/vespa/pull/18023 was premature.

@geirst please review


geirst (approved):
👍

PR #18212, opened 2021-06-11
Balder/expose and fix predicate commit bug

### LINKED ISSUE #17481: Predicate queries exclude documents influenced by other documents predicate value
Hello,

Sometimes when executing a query filtering by a predicate a document which should match isn't included in the result, removing one field from the sent query which isn't present on the predicate condition or changing its value results in the document appearing.

This has happened several times randomly and for different documents, let me explain one case:

Given the query:
```
select * from doc where siteId = 0 and predicate(show_condition,{\"visitorFrequency\":\"5\",
\"section\": [\"28961\"],"}, {});
```

A document `A` with a predicate field `"show_condition": "section in [28961]"` and a field `"siteId": 0`

Another document `B` with the predicate field `"show_condition": "visitorFrequency in [4,5] and country in [US]"` and a field `"siteId": 1`

The document `A` would not show in the results, unless the `visitorFrequency` field in the query predicate was removed or its value wasn't 4 or 5.

There were several ways that somehow fixed the issue:
- Making a partial update on document `A` assigning the same value or another to the `show_condition` field
- Making a whole update on document `A` sending the same document or modifying it
- Removing the `visitorFrequency` value from document `B` predicate, effectively leaving it as ``"show_condition": "country in [US]"``
- Changing the `visitorFrequency` value from document `B` predicate

Other relevant information:
- Query coverage was always 100
- Container cluster has 3 nodes
- Data distribution is `flat` 
- Arity on the problematic predicate field is set to 2
- Couldn't find a way to replicate the issue

The problematic predicate in doc schema:
```
 field show_condition type predicate {
    indexing: summary | attribute
    index {
        arity: 2 # mandatory
    }
}
```

baldersheim:
@geirst or @toregge or @bjorncs 
This fixes #17481

The first commit is some cleanup and a test to expose the bug.
The last is the fix.

### DIFF
--- a/searchlib/src/tests/predicate/predicate_index_test.cpp
+++ b/searchlib/src/tests/predicate/predicate_index_test.cpp
@@ -113,7 +113,6 @@ TEST("require that PredicateIndex can index document") {
     EXPECT_FALSE(index.getIntervalIndex().lookup(hash).valid());
     indexFeature(index, doc_id, min_feature, {{hash, interval}}, {});
     index.commit();
-
     auto posting_it = lookupPosting(index, hash);
     EXPECT_EQUAL(doc_id, posting_it.getKey());
     uint32_t size;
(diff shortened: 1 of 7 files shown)


toregge (approved):

bjorncs (commented):
👍

PR #18221, opened 2021-06-11
Only support legacy mode in getActiveIdealStateOperations().

geirst:
This function is only used by idealstatemanagertest in the context of testing a single stripe.

@vekterli please review
@toregge FYI


vekterli (approved):
:+1:

PR #18233, opened 2021-06-14
Start OS upgrades during work hours

mpolden:
@freva

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OsUpgradeScheduler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OsUpgradeScheduler.java
@@ -43,8 +43,10 @@ public class OsUpgradeScheduler extends ControllerMaintainer {
 
     @Override
     protected double maintain() {
+        Instant now = controller().clock().instant();
+        if (!canTriggerAt(now)) return 1.0;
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/OsUpgradeScheduler.java
```diff
@@ -88,6 +88,14 @@ private boolean hasExpired(Version version) {
                            .collect(Collectors.toUnmodifiableSet());
     }
 
+    private boolean canTriggerAt(Instant instant) {
+        int hourOfDay = instant.atZone(ZoneOffset.UTC).getHour();
+        int dayOfWeek = instant.atZone(ZoneOffset.UTC).getDayOfWeek().getValue();
+        // Upgrade can only be scheduled between 07:00 and 12:59 UTC, Monday-Thursday
+        return hourOfDay >= 7 && hourOfDay <= 13 &&
```
freva, line 95: This is between 07:00 and 13:59? (Ref. comment above)

mpolden: Thanks, fixed.

freva (approved):

Issue #13193, opened 2020-05-08
document processing: double invocation - update documentation


kkraune:
Using https://github.com/vespa-engine/sample-apps/tree/master/vespa-cloud/album-recommendation-docproc

1588936248.407807	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/700	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	In process
1588936248.422296	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/700	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Added to requests pending: 1
1588936248.422867	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/700	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Request pending ID: 1, Progress.LATER
1588936248.453697	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/703	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	In process
1588936248.466975	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/703	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Request pending ID: 1, Progress.LATER
1588936248.476168	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/704	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	In handleResponse
1588936248.477377	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/704	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Async response to put or get, requestID: 1
1588936248.477641	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/704	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Found lyrics for : document 'id:mynamespace:lyrics::1' of type 'lyrics'
1588936248.487219	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/700	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	In process
1588936248.487786	h624a.dev.aws-us-east-1c.vespa-external.aws.oath.cloud	883/700	container	Container.ai.vespa.example.album.LyricsDocumentProcessor	info	  Set lyrics, Progress.DONE
[...]

PR #18237, opened 2021-06-14
Disallow 'default' as indexing chain

bratseth:
@arnej27959 or @kkraune

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/Content.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/Content.java
@@ -111,13 +111,15 @@ public class Content extends ConfigModel {
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


kkraune:
Thanks! ref https://github.com/vespa-engine/vespa/issues/13193

I take it examples in doc should rather use a chain name "foo" and inherit indexing, instead of naming the chain default? (the latter is a bit magic)

kkraune (approved):

bratseth:
Examples of general document processing chains in the container? 
Or examples of indexing chains referenced in a content cluster?

The former should be named "default" as that avoids the need to deal with explicit routes.
The latter should not be named "default" and should inherit "indexing", both of which are now validated.

arnej27959 (approved):
👍

PR #18240, opened 2021-06-14
Proxy filtered metadata request

olaaun:


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/horizon/TsdbQueryRewriterTest.java
```diff
@@ -35,6 +35,10 @@ public void rewrites_query() throws IOException {
         assertRewrite("no-filters.json",
                 "no-filters.expected.json",
                 Role.reader(TenantName.from("tenant2")), Role.athenzTenantAdmin(TenantName.from("tenant3")));
+
+        assertRewrite("meta-query.json",
+                "filters-meta-query.json",
```
freva, line 40: Follow the same filename pattern as the others so they are grouped together, i.e. `meta-query.expected.json`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/horizon/TsdbQueryRewriter.java
```diff
@@ -66,7 +68,7 @@ private static void rewriteFilter(ObjectNode parent, Set<TenantName> tenantNames
         ObjectNode prev = ((ObjectNode) parent.get("filter"));
         ArrayNode filters;
         // If we dont already have a filter object, or the object that we have is not an AND filter
-        if (prev == null || !"Chain".equals(prev.get("type").asText()) || !"AND".equals(prev.get("op").asText())) {
+        if (prev == null || !"Chain".equals(prev.get("type").asText()) || prev.get("op") == null || !"AND".equals(prev.get("op").asText())) {
```
freva, line 71: I think `AND` is default, the documentation is... inconsistent: http://opentsdb.net/docs/3x/build/html/user_guide/semanticquery/filters.html#chain
> Combines one or more filters with a logical AND (the default) or a logical OR. 

But the table says default is `OR`

Logically, we want `AND` when doing suggestions, so I guess `AND` is default then since they are not specifying. In that case, we should change this to `!= null` and let the additional filters be appended to this chain.

freva (approved):

PR #18250, opened 2021-06-14
Jonmv/add missing test job type MERGEOK

jonmv:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/JobType.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/JobType.java
@@ -91,18 +91,22 @@ public enum JobType {
     productionAwsApNortheast1a  ("production-aws-ap-northeast-1a",
                             Map.of(Public, ZoneId.from("prod", "aws-ap-northeast-1a"))),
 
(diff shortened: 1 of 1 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/JobType.java
```diff
@@ -155,7 +162,7 @@
                             Map.of(main, ZoneId.from("perf"   , "us-east-3")));
 
     private final String jobName;
-    private final Map<SystemName, ZoneId> zones;
+    final Map<SystemName, ZoneId> zones;
```
freva, line 165: Why?

freva (approved):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/JobType.java
jonmv: Package private is OK. Access it in test.

oyving:
Nothing against this PR - but shouldn't this be calculated from the ZoneEnum instead of explicitly being encoded?

jonmv:
Yes, we should change this now.

PR #18252, opened 2021-06-14
allow multiple tensor types for same name

arnej27959:
* with several document types you can have fields with the
  same name but different tensor types that all should be
  allowed as target for nearestNeighbor operator.

@geirst please review


geirst (approved):
👍

PR #18256, opened 2021-06-14
Content clusters need enough cpu to manage their content

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
```diff
@@ -64,6 +64,7 @@ public NodeResources enlargeToLegal(NodeResources requested, ClusterSpec.Type cl
 
     private double minAdvertisedVcpu(ClusterSpec.Type clusterType) {
         if (zone().environment() == Environment.dev && !zone().getCloud().dynamicProvisioning()) return 0.1;
+        if (clusterType == ClusterSpec.Type.content && zone().environment().isProduction()) return 1.0;
```
freva, line 67: What about `combined`? Use `clusterType.isContent()`?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
bratseth: 👍 done

PR #18259, opened 2021-06-15
Revert "Disallow 'default' as indexing chain"

hmusum:
Reverts vespa-engine/vespa#18237

We cannot have stricter checks for features that are in use. Broke an app in production. Consider using a feature flag

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/Content.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/Content.java
@@ -111,15 +111,13 @@ public class Content extends ConfigModel {
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


bratseth:
In my opinion, this is a configuration error we didn't catch until now, and we can add validation without further complicated steps.

hmusum:
If the configuration is used it will either break infrastructure upgrade or the app (in the case that infrastructure upgrade is not broken, even though it should have been).

Right now this has been fixed in the app, though, so closing.

bratseth:
Yes, this wasn't done in a validation code path, which is wrong, but now it doesn't mater as you say.
I think I'll move both the new and existing validation to run on validation only to reduce the chance of this happening again.

PR #18269, opened 2021-06-15
Do not recycle deprovisioning hosts

mpolden:
@olaaun or @freva

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
@@ -4,13 +4,13 @@ package com.yahoo.vespa.hosted.provision.maintenance;
 import com.yahoo.config.provision.NodeType;
 import com.yahoo.jdisc.Metric;
 import com.yahoo.vespa.hosted.provision.Node;
+import com.yahoo.vespa.hosted.provision.NodeList;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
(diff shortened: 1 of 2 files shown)


freva (approved):
:+1: 


Not related to this PR, but I'm not sure if this maintainer works as a whole: 
For a host to be deprovisioned, it needs to have no allocated children, and either not be `parked` or have `wantToDeprovision` set. This maintainer will only move the host to `dirty` (doesn't use the recursive method), but even if it did, `dirty` doesn't clear allocation anyway, so the child(ren) will still have allocation?

mpolden:
Agree. I'm not entirely sure of why it's recycling hosts at all. Couldn't this just deprovision excess hosts? Also, it probably shouldn't touch hosts that are parked by an operator.

freva:
Yes, deprovisioning directly is probably easiest and shouldn't touch `parked` by operator unless `wantToDeprovision`

olaaun:
I thought

1. Dirtying parked nodes in dynamically provisioned zones deprovisions them
2. Parked app nodes have no allocation

Both of these are untrue?

mpolden:
Merging this. The other things mentioned can be addressed in a separate PR.

freva:
1. The condition is:
a. Host has no children with allocation
b. Host is not in `parked` _or_ has `wantToDeprovision` set
So if the host already had no allocated children and `wantToDeprovision` wasn't set, then moving it to `dirty` would deprovision it, yes.
2. In general, that is not true, but I do see that hosts expired from `provisioned` do get their allocation removed while moved to `parked`

PR #18272, opened 2021-06-15
Add per stripe handling of ideal state metrics with aggregation on top.

geirst:
This is handled similarly to per stripe distributor metrics.

@vekterli or @toregge please review


vekterli (approved):
🦄👍✨

PR #18278, opened 2021-06-15
Gc unused config options: min-group-coverage and max nodes per group

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/content/cluster/DomTuningDispatchBuilder.java
```diff
@@ -25,12 +25,11 @@ public static DispatchTuning build(ModelElement contentXml, DeployLogger logger)
         builder.setMaxHitsPerPartition(dispatchElement.childAsInteger("max-hits-per-partition"));
         builder.setTopKProbability(dispatchElement.childAsDouble("top-k-probability"));
         builder.setDispatchPolicy(dispatchElement.childAsString("dispatch-policy"));
-        builder.setMinGroupCoverage(dispatchElement.childAsDouble("min-group-coverage"));
         builder.setMinActiveDocsCoverage(dispatchElement.childAsDouble("min-active-docs-coverage"));
 
         if (dispatchElement.child("use-local-node") != null)
```
baldersheim, line 30: You should add a similar warning if 'min-group-coverage' has been configured.

baldersheim (changes requested):
Add a warning and you are good to go.

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/DomTuningDispatchBuilder.java
bratseth: Good point; done

baldersheim (approved):

PR #18279, opened 2021-06-15
Deprovision nodes if not parked by operator

olaaun:
@mpolden or @freva


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -41,13 +41,14 @@ protected double maintain() {
...
                                              .nodeType(NodeType.host)
+                                             .not().matching(this::parkedByOperator)
                                              .not().deprovisioning();
         int hostsToExpire = Math.max(0, parkedHosts.size() - MAX_ALLOWED_PARKED_HOSTS);
         parkedHosts.sortedBy(Comparator.comparing(this::parkedAt))
                    .first(hostsToExpire)
                    .forEach(host -> {
                        log.info("Allowed number of parked nodes exceeded. Recycling " + host.hostname());
-                       nodeRepository.nodes().deallocate(host, Agent.ParkedExpirer, "Expired by ParkedExpirer");
+                       nodeRepository.nodes().deprovision(host.hostname(), Agent.ParkedExpirer, Instant.now());
```
freva, line 51: This just changes the node state, need to actually make a call to AWS to remove the instance, see https://github.com/vespa-engine/vespa/blob/8ef499e16e9fb5daede071d36cb523f4d30538c0/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DynamicProvisioningMaintainer.java#L134:L135

olaaun: Won't `DynamicProvisioningMaintainer` make the call to AWS, as the hosts are set to `wantToDeprovision` here?

freva: Seems to me like this is equal to setting the host to `dirty`, both of these just modify the second condition for DPM to deprovision and neither actually helps if there are allocated children?

mpolden: Do we want to expire non-empty hosts at all though? If I remember correctly we're adding this to handle hosts expired by ProvisionedExpirer (which will always be empty).

freva: In that case, do we even need this maintainer? Cant ProvisionedExpirer just delete them directly?

mpolden: True, ProvisionedExpirer could do it instead. Park as it does today, but mark excess hosts wantToDeprovision could work.

olaaun: The reason it became a separate maintainer is we wanted to remove the oldest parked hosts, but I suppose that's not really required. If anything, it could be an annoyance that we continuously rotate the set of parked hosts.

mpolden: Yes. Let's just keep the oldest ones for now. So basically:
1. ProvisionedExpirer parks hosts that have been too long in provisioned (same as today)
2. ProvisionedExpirer sets wantToDeprovision on the host to be parked if we already have >=20 hosts parked by Agent.ProvisionedExpirer

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ProvisionedExpirer.java
```diff
@@ -22,17 +25,40 @@
...
     protected void expire(List<Node> expired) {
+        int previouslyExpired = numberOfPreviouslyExpired();
         for (Node expiredNode : expired) {
-            if (expiredNode.type() == NodeType.host) {
-                nodeRepository().nodes().parkRecursively(expiredNode.hostname(), Agent.ProvisionedExpirer, "Node is stuck in provisioned");
+            if (expiredNode.type() != NodeType.host)
+                continue;
+            nodeRepository().nodes().parkRecursively(expiredNode.hostname(), Agent.ProvisionedExpirer, "Node is stuck in provisioned");
+            if (MAXIMUM_ALLOWED_EXPIRED_HOSTS < ++previouslyExpired) {
+                nodeRepository.nodes().deprovision(expiredNode.hostname(), Agent.ProvisionedExpirer, Instant.now());
```
mpolden, line 44: Should use `nodeRepository.clock().instant()` instead of `Instant.now()`.

mpolden (approved):

mpolden (approved):
👍

PR #18303, opened 2021-06-17
Internal structs take precedence over external documents

bratseth:


config-model/src/main/java/com/yahoo/searchdefinition/derived/VsmFields.java
```diff
@@ -178,6 +178,7 @@ private StreamingField(String name,DataType sourceType, Matching  matching, bool
 
         /** Converts to the right index type from a field datatype */
         private static Type convertType(DataType fieldType) {
+            System.out.println("Converting field type " + fieldType + " which is a " + fieldType.getClass());
```
baldersheim, line 181: Forgotten debugging I presume.

baldersheim (approved):

config-model/src/main/java/com/yahoo/searchdefinition/derived/VsmFields.java
bratseth: yes, thanks

baldersheim (approved):

PR #18304, opened 2021-06-17
Remove notifications for instances no longer referenced in deployment.xml

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -445,6 +445,11 @@ public LockedApplication storeWithUpdatedConfig(LockedApplication application, A
         // Validate new deployment spec thoroughly before storing it.
         controller.jobController().deploymentStatus(application.get());
 
+        // Clear notifications for instances that are no longer declared
+        for (var name : existingInstances)
+            if ( ! declaredInstances.contains(name))
+                controller.notificationsDb().removeNotifications(NotificationSource.from(application.get().id().instance(name)));
```
jonmv, line 451: We should probably also do the other cleanup, like removing DNS entries?

jonmv (approved):
Nice, more! :p

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
freva: Yes, I'm not sure what happens to those now and if its safe to do it here? FYI: @mpolden

mpolden: `deleteInstance(ApplicationId)` already removes DNS records.

freva: Yes... The problem here is that when an instance is removed from `deployment.xml`, `deleteInstance(ApplicationId)` is never called.

mpolden: 😞 we should make it one shared code path.

PR #18308, opened 2021-06-17
Set the correct path for GDB on RHEL8.

aressem:


baldersheim (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoreCollector.java
```diff
@@ -29,7 +30,9 @@
     private static final Pattern CORE_GENERATOR_PATH_PATTERN = Pattern.compile("^Core was generated by `(?<path>.*?)'.$");
     private static final Pattern EXECFN_PATH_PATTERN = Pattern.compile("^.* execfn: '(?<path>.*?)'");
     private static final Pattern FROM_PATH_PATTERN = Pattern.compile("^.* from '(?<path>.*?)'");
-    static final String GDB_PATH = "/opt/rh/devtoolset-10/root/bin/gdb";
+    static final String GDB_PATH = Files.exists(Path.of("/opt/rh/gcc-toolset-10/root/bin/gdb")) ?
```
toregge, line 33: This checks if the file is present on the host, but it should probably check if the file is present in the container.

baldersheim: yes, you might be right about that. I guess @freva can tell.

freva: Yes, this need to exist in the container.

aressem: Fixed by probing the container.

toregge: Note that old containers having devtoolset-9 or gcc-toolset-9 versions of gdb will be considered as having gcc-toolset-10 version of gdb.

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoreCollector.java
aressem: So this would fail without this change as well when we have a mix of devtoolset-9 and devtoolset-10. I will implement a check that makes this work in both cases.

freva:
Why don't we just have a link to the right `gdb` from `/usr/bin` or something? This is probably too late for devtools 9, but we will get rid of those in a week or so?

aressem:
`/usr/bin/gdb` is the standard GDB that comes with RHEL. We could create a hack where that points to the devtoolset instance. However, it would be better to source something that alters the PATH so we can just execute `gdb`. This could be generalized to work for more than just gdb.

freva (approved):

PR #18314, opened 2021-06-18
Log which tenants we expire

oyving:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CloudTrialExpirer.java
```diff
@@ -38,6 +41,12 @@ protected double maintain() {
                 .filter(this::tenantHasNoDeployments)        // no running deployments active
                 .collect(Collectors.toList());
 
+        var expiredTenantNames = expiredTenants.stream()
+                .map(Tenant::name)
+                .map(TenantName::value)
+                .collect(Collectors.joining(", "));
+
+        log.info("Moving expired tenants to 'none' plan: " + expiredTenantNames);
```
mpolden, line 49: Nit: In general it's better to use `log.log(<level>, () -> "string")` to create the message lazily. You could also put the list collection inside that supplier.

mpolden (approved):

PR #18316, opened 2021-06-18
Actually test the Junit 4 tests in flags

hakonhall:


mpolden (approved):
😞

baldersheim:
Why not just convert to junit5 when you are at it

PR #18320, opened 2021-06-18
Add retrigger queue to avoid errors when job already running

tokle:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/RetriggerMaintainer.java
```diff
@@ -0,0 +1,66 @@
...
+                .filter(run -> run.id().type().equals(jobId.type()))
+                .findFirst();
+        return existingRun.isEmpty();
+    }
+
+    /*
+    Returns true of job needs triggering. I.e the job has not run since the queue item was last run.
+     */
+    private boolean needsTrigger(RetriggerEntry entry) {
+        return controller().jobController().lastSuccess(entry.jobId())
```
jonmv, line 62: Using `lastSuccess` here means this will trigger back to back whenever a job is failing. Is that intentional? I guess there are pros and cons, and this is only meant for manual intervention?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -997,6 +997,7 @@ private HttpResponse disallowSupportAccess(String tenantName, String application
         DeploymentId deployment = new DeploymentId(ApplicationId.from(tenantName, applicationName, instanceName), requireZone(environment, region));
         Principal principal = requireUserPrincipal(request);
         SupportAccess disallowed = controller.supportAccess().disallow(deployment, principal.getName());
+        controller.applications().deploymentTrigger().reTriggerOrAddToQueue(deployment);
```
jonmv, line 1000: Should this also happen on `allowSupportAccess`?

jonmv: NVM, I see this is run below, in a different API.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/RetriggerMaintainerTest.java
```diff
@@ -0,0 +1,70 @@
...
+        // Run outstanding jobs
+        devApp.runJob(JobType.devUsEast1);
+        assertEquals(0, tester.jobs().active(applicationId).size());
+
+        // Run maintainer again, should find that the job has already run successfully and will remove the entry.
+        maintainer.maintain();
+        retriggerEntries = tester.controller().curator().readRetriggerEntries();
+        assertEquals(0, retriggerEntries.size());
+        assertEquals(0, tester.jobs().active(applicationId).size());
+    }
```
jonmv, line 69: Very ncie!

jonmv (approved):
One question, but looks great!

PR #18322, opened 2021-06-18
add 'bit(a,b)' math function

havardpe:

bit(a,b) is similar to (a & (1 << b))
'a' should be an integer in the range [-128,127]
'b' should be an integer in the range [0,7]
conformance testing of bit(a,b) is disabled until we have an implementation in Java as well.

@arnej27959 please review
@lesters @bratseth FYI

### DIFF
--- a/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
+++ b/eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
@@ -4,6 +4,7 @@
 #include "llvm_wrapper.h"
 #include <vespa/eval/eval/node_visitor.h>
 #include <vespa/eval/eval/node_traverser.h>
+#include <vespa/eval/eval/extract_bit.h>
 #include <llvm/IR/Verifier.h>
 #include <llvm/Support/TargetSelect.h>
 #include <llvm/IR/IRBuilder.h>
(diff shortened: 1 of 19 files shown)


eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
```diff
@@ -29,6 +29,12 @@ double vespalib_eval_approx(double a, double b) { return (vespalib::approx_equal
 double vespalib_eval_relu(double a) { return std::max(a, 0.0); }
 double vespalib_eval_sigmoid(double a) { return 1.0 / (1.0 + std::exp(-1.0 * a)); }
 double vespalib_eval_elu(double a) { return (a < 0) ? std::exp(a) - 1.0 : a; }
+double vespalib_eval_bit(double a, double b) {
+    // must match Bit::f
```
arnej27959, line 33: it would be easier to just call Bit::f here

arnej27959 (approved):
looks good to me

eval/src/vespa/eval/eval/llvm/llvm_wrapper.cpp
havardpe: I considered it, but did not think gcc would be able to inline it (feels bad to get 2 function calls per bit). Maybe we should have a common inline implementation that we call from both places.

havardpe:
@arnej27959 please take another look

havardpe:
` 10:54:33 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.5.2:install (default-install) on project bundle-plugin: Failed to install metadata com.yahoo.vespa/maven-metadata.xml: Error updating group repository metadata: in epilog non whitespace content is not allowed but got / (position: END_TAG seen ...</metadata>\n/... @16:2) -> [Help 1] `

arnej27959 (approved):
👍

PR #18346, opened 2021-06-21
Revert "Revert "Compute percentage of zero hit queries.""

baldersheim:
Reverts vespa-engine/vespa#18344


baldersheim:
@havardpe or @hmusum PR
Only the last commit matters.

hmusum (approved):

Issue #6152, opened 2018-06-11
Fix searchnode tuning doc options for index 


jobergum:
http://docs.vespa.ai/documentation/content/setup-proton-tuning.html#index

Incorrectly lists mlock as an valid option for io.search. If you try 'mlock' it will fallback to 'normal' but pass xml schema validation. Also mmap is the default and the others like directio etc has limited test coverage so I'm suggesting to simply remove the io.search options and only allow mmap as io.search option. Instead add a flag for the POPULATE mmap option which is more common in use.

baldersheim:
Fixed in 7.423.20

PR #18347, opened 2021-06-21
remove mlock as option as it is long gone.

baldersheim:
@bratseth or @geirst or @jobergum PR


baldersheim:
This is step 2 for #6152

jobergum (approved):

PR #18294, opened 2021-06-16
Package Apache http client libraries as a single osgi bundle [run-systemtest]

bjorncs:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #18357, opened 2021-06-22
Revert "Package Apache http client libraries as a single osgi bundle [run-systemtest]" MERGEOK

aressem:
Reverts vespa-engine/vespa#18294


baldersheim:
Hold, as untangling is more than this one

PR #18368, opened 2021-06-22
Use more robust throughput calculation with less overhead

jonmv:


vespa-feed-client/src/test/java/ai/vespa/feed/client/HttpRequestStrategyTest.java
```diff
@@ -33,7 +33,7 @@
 
     @Test
     void testConcurrency() {
-        int documents = 1 << 16;
+        int documents = 1 << 24;
```
jonmv, line 36:
```suggestion
        int documents = 1 << 16;
```
Not intentional.

bjorncs (approved):

PR #18370, opened 2021-06-22
Bratseth/lock tables

bratseth:


freva:
27 test failures in node-repository

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -236,122 +199,76 @@ public void close() {
...
+        if (clusterTable.exists())
+            ensureClusterTableIsUpdated();
         else
-            createClusterTable(context);
+            createClusterTable();
+    }
+
+    private void ensureNodeTableIsUpdated() {
+        try {
+            if (0 == engine.getStatus(newContext().getCairoSecurityContext(), new Path(), nodeTable.name)) {
```
freva, line 216: This is just `nodeTable.exists()`, but that is already checked before calling this the method?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
bratseth: Yes, it does nothing now, but I need it each time a column is added.

freva: Yeah, I get that, it's just that this `if` is redundant, could leave the `try` block empty until you need it

bratseth: Good point, done

PR #18376, opened 2021-06-23
Add feature falg for controlling onnx dryrun verification.

baldersheim:
@havardpe PR

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
@@ -86,6 +86,7 @@ public interface ModelContext {
         @ModelFeatureFlag(owners = {"baldersheim"}) default boolean distributeExternalRankExpressions() { return false; }
         @ModelFeatureFlag(owners = {"baldersheim"}) default int maxConcurrentMergesPerNode() { throw new UnsupportedOperationException("TODO specify default value"); }
         @ModelFeatureFlag(owners = {"baldersheim"}) default int maxMergeQueueSize() { throw new UnsupportedOperationException("TODO specify default value"); }
+        @ModelFeatureFlag(owners = {"baldersheim"}) default boolean dryRunOnnxOnSetup() { throw new UnsupportedOperationException("TODO specify default value"); }
         @ModelFeatureFlag(owners = {"geirst"}) default boolean enableFeedBlockInDistributor() { return true; }
         @ModelFeatureFlag(owners = {"hmusum"}, removeAfter = "7.406") default int clusterControllerMaxHeapSizeInMb() { return 128; }
         @ModelFeatureFlag(owners = {"hmusum"}, removeAfter = "7.422") default int metricsProxyMaxHeapSizeInMb(ClusterSpec.Type type) { return 256; }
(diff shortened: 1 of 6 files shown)


config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
```diff
@@ -86,6 +86,7 @@
         @ModelFeatureFlag(owners = {"baldersheim"}) default boolean distributeExternalRankExpressions() { return false; }
         @ModelFeatureFlag(owners = {"baldersheim"}) default int maxConcurrentMergesPerNode() { throw new UnsupportedOperationException("TODO specify default value"); }
         @ModelFeatureFlag(owners = {"baldersheim"}) default int maxMergeQueueSize() { throw new UnsupportedOperationException("TODO specify default value"); }
+        @ModelFeatureFlag(owners = {"baldersheim"}) default boolean dryRunOnnxOnSetup() { throw new UnsupportedOperationException("TODO specify default value"); }
```
havardpe, line 89: why do flags have a default implementation if the best (?) way to implement it is to make sure it is not called?

config-model/src/main/java/com/yahoo/searchdefinition/derived/RankProfileList.java
```diff
@@ -35,13 +35,15 @@
...
     private final OnnxModels onnxModels;
+    private final boolean dryRunOnnxOnSetup;
 
     public static RankProfileList empty = new RankProfileList();
 
     private RankProfileList() {
         rankingConstants = new RankingConstants();
         largeRankExpressions = new LargeRankExpressions();
         onnxModels = new OnnxModels();
+        dryRunOnnxOnSetup = true;
```
havardpe, line 46: This seems a bit eager?

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -286,6 +286,13 @@
             "Whether to load local sessions when bootstrapping config server",
             "Takes effect on restart of config server");
 
+    public static final UnboundBooleanFlag DRY_RUN_ONNX_ON_SETUP = defineFeatureFlag(
+            "dryrun-onnx-on-setup", false,
```
havardpe, line 290: dry-run-onnx-on-setup ?

config-model/src/main/java/com/yahoo/searchdefinition/derived/RankProfileList.java
baldersheim: This is just test code. fail-fast is my preferred apparoach.

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
baldersheim: Correct :) Fixed

config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
baldersheim: Because this is an api between the configmodels and the configserver.
Once use of the method is gone there still must be a default implemenation until the last version is gone.

PR #18379, opened 2021-06-23
Revert "Bjorncs/tls cipher feature flag"

hmusum:
Reverts vespa-engine/vespa#18374

`enableJdiscHttp2()` cannot be removed, still in use:
  Caused by: java.lang.NoSuchMethodError: 'boolean com.yahoo.config.model.api.ModelContext$FeatureFlags.enableJdiscHttp2()'
        at com.yahoo.vespa.model.container.xml.ContainerModelBuilder.addAdditionalHostedConnector(ContainerModelBuilder.java:436)

Well, I guess this revert is not good either (since we cannot remove stuff from API), but `enableJdiscHttp2()` should be added back.


hmusum:
https://github.com/vespa-engine/vespa/pull/18380 adds back a method to API

hmusum:
Fixed by #18380

PR #18380, opened 2021-06-23
Add back method, still in use

hmusum:


hmusum:
@aressem FYI

bjorncs (approved):
Thanks, forgot that the dangling variable is actual usage 😄

PR #18400, opened 2021-06-24
More CLI print options

jonmv:
@bjorncs please review.

### DIFF
--- a/vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
+++ b/vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
@@ -1,6 +1,7 @@
 // Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package ai.vespa.feed.client;
 
+import ai.vespa.feed.client.JsonFeeder.ResultCallback;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 
@@ -12,8 +13,10 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.nio.file.Files;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Main method for CLI interface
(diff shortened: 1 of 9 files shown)


vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
```diff
@@ -60,22 +65,55 @@ private int run(String[] rawArgs) {
             try (InputStream in = createFeedInputStream(cliArgs);
                  FeedClient feedClient = createFeedClient(cliArgs);
                  JsonFeeder feeder = createJsonFeeder(feedClient, cliArgs)) {
+                CountDownLatch latch = new CountDownLatch(1);
+                AtomicReference<FeedException> fatal = new AtomicReference<>();
                 long startNanos = System.nanoTime();
-                feeder.feedMany(in).join();
+                feeder.feedMany(in, new ResultCallback() {
+                    @Override public void onNextResult(Result result, FeedException error) { handleResult(result, error, cliArgs); }
+                    @Override public void onError(FeedException error) { fatal.set(error); }
```
bjorncs, line 73: Should latch count down be invoked `onError`? `onComplete` will not be invoked in this case.

vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
```diff
@@ -60,22 +65,55 @@ private int run(String[] rawArgs) {
...
+                    @Override public void onComplete() { latch.countDown(); }
+                });
+                if (cliArgs.showProgress()) {
+                    new Thread(() -> {
```
bjorncs, line 77: Free threads are ok since this is a CLI, otherwise this would be a potential resource leak. Make sure it's tagged as daemon.

bjorncs (approved):
See PR build failure. The help page expected output must also be updated.

vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
jonmv: Oh, thne it definiteely should :)

vespa-feed-client-cli/src/main/java/ai/vespa/feed/client/CliClient.java
jonmv: Will do.

bjorncs (approved):

PR #18414, opened 2021-06-25
Enlarge remote disks to smallest lergal size if smaller is wanted

bratseth:


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTester.java
```diff
@@ -152,6 +152,37 @@ public void addCpuMeasurements(float value, float otherResourcesLoad,
...
+     */
+    public void addDiskMeasurements(float value, float otherResourcesLoad,
+                                    int count, ApplicationId applicationId) {
+        NodeList nodes = nodeRepository().nodes().list(Node.State.active).owner(applicationId);
+        float oneExtraNodeFactor = (float)(nodes.size() - 1.0) / (nodes.size());
+        for (int i = 0; i < count; i++) {
+            clock().advance(Duration.ofMinutes(5));
+            for (Node node : nodes) {
+                Load load = new Load(ClusterModel.idealQueryCpuLoad * otherResourcesLoad,
+                                     value,
```
freva, line 173: This adds `value` measurement for memory, not disk?

freva (approved):

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/AutoscalingTester.java
bratseth: yes, thanks!

PR #18426, opened 2021-06-28
Include empty applications in recursive tenant

freva:


freva:
Or @mpolden, this is needed to list applications after they are created (and are still empty) in the console

jonmv (approved):
Nice!

PR #18429, opened 2021-06-28
Validate validation-overrides.xml when validating model

hmusum:

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
@@ -49,7 +49,6 @@ public class ValidationOverrides {
     private ValidationOverrides(List<Allow> overrides, String xmlForm) {
         this.overrides = ImmutableList.copyOf(overrides);
         this.xmlForm = xmlForm;
-        validate(Instant.now());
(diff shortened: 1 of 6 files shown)


config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
```diff
@@ -73,15 +73,24 @@ public boolean allows(String validationIdString, Instant now) {
 
     /** Returns whether the given (assumed invalid) change is allowed by this at the moment */
     public boolean allows(ValidationId validationId, Instant now) {
+        validate(now);
```
bratseth, line 76: Shouldn't this line be removed?

bratseth (changes requested):

config-model-api/src/main/java/com/yahoo/config/application/api/ValidationOverrides.java
hmusum: Oops, yes, that should not be there

hmusum: Fixed

bratseth (approved):
👍

PR #18437, opened 2021-06-28
unpack bits function

havardpe:

@arnej27959 please review
@lesters FYI

### DIFF
--- a/eval/src/vespa/eval/instruction/unpack_bits_function.h
+++ b/eval/src/vespa/eval/instruction/unpack_bits_function.h
@@ -1,0 +1,37 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include <vespa/eval/eval/tensor_function.h>
+
+namespace vespalib::eval {
+
+/**
+ * Tensor function unpacking bits into separate values.
+ *
+ * Both the tensor containing the packed bits and the result tensor
+ * must have cell type 'int8'. The bits must be unpacked in canonical
+ * order; bytes are unpacked with increasing index, bits within a byte
+ * are unpacked from most to least significant.
+ *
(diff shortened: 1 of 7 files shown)


eval/src/vespa/eval/instruction/unpack_bits_function.h
```diff
@@ -0,0 +1,37 @@
...
+ * Tensor function unpacking bits into separate values.
+ *
+ * Both the tensor containing the packed bits and the result tensor
+ * must have cell type 'int8'. The bits must be unpacked in canonical
+ * order; bytes are unpacked with increasing index, bits within a byte
+ * are unpacked from most to least significant.
+ *
+ * The baseline expression looks like this:
+ *
+ * tensor<int8>(x[64])(bit(packed{x:(x/8)},7-(x%8)))
```
arnej27959, line 19: My guess is that users will write
`tensor<int8>(x[64])(bit(packed{x:(x/8)},x%8))`
instead...

arnej27959 (approved):
looks good, I just hope we can write some really good documentation on how to use this.

eval/src/vespa/eval/instruction/unpack_bits_function.h
arnej27959: On second thought they will probably write
`tensor(x[64])(bit(packed{x:(x/8)},x%8))`

havardpe: I have considered adding support for 'little endian' as well (top alternative form), but wanted to do the basic expression first. I have checked the unpack_bits function from python, and that one defaults to 'big endian'. The last alternative form could also be recognized if we go for 'any cell type as output', which would actually by quite reasonable. A possible solution is to add an unpack_bits second order function in Java that will expand to the appropriate expression, I am not sure how complicated that would be. @lesters any thoughts?

PR #18440, opened 2021-06-29
US locale

kkraune:


freva (approved):

arnej27959:
I'm guessing this was just for the sake of unit tests?  if that is the case, we should probably override environment in parent/pom.xml instead of sprinkling Locale.US and/or Locale.ENGLISH around.  If we really need to use a specific locale for String.format, we should probably make a small convenience function `Text.fmt()` and use that instead of String.format everywhere - that's easier to keep consistent.

kkraune:
> I'm guessing this was just for the sake of unit tests?

correct - unit test failed on my non-US laptop. I run a full build now and then, once a quarter we need to add this, so not a big problem - but I am find with either

bratseth:
We should always pass locale in the code itself.

PR #18442, opened 2021-06-29
flip feature flag

arnej27959:

@geirst please review
@hmusum FYI

note: this flag is now "true" in all production zones, so I want to flip the default, then change default settings to match what we now want, then later remove the flag entirely.  is that how we usually do things?


baldersheim (approved):
Yes, this is correct approach.

PR #18446, opened 2021-06-29
Remove docker-api

mpolden:
Merge with internal PR.

@hakonhall or @freva


hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -307,8 +301,6 @@ public void suspend(NodeAgentContext context) {
             if (!output.isBlank()) {
                 context.log(logger, "Suspend services output: " + output);
             }
-        } catch (ContainerNotFoundException e) {
```
freva, line 310: Would be nice if we had this with podman client as well

PR #18470, opened 2021-06-29
Arnej/simpler format with locale

arnej27959:

@bratseth here is a suggested small convenience wrapper for String.format with deterministic Locale,
with usage from one module (using automated refactoring).  Please take a look.


bratseth (commented):
Nice! But can we please call it "format".

bratseth (approved):

PR #18474, opened 2021-06-30
Throw instead of returning null for getBytes()/getData()

hmusum:
Update callers to handle exception or check path existence before
calling methods.


bratseth (approved):
👍

PR #18476, opened 2021-06-30
Allow deviation of at least 1 document

hmusum:
Let content be well-balanced when there are few docs in a cluster

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
@@ -68,7 +68,7 @@ public class Group {
         if (numWorkingNodes > 0) {
             long average = activeDocs / numWorkingNodes;
             long deviation = nodes.stream().filter(node -> node.isWorking() == Boolean.TRUE).mapToLong(node -> Math.abs(node.getActiveDocuments() - average)).sum();
-            boolean isDeviationSmall = deviation <= (activeDocs * MAX_UNBALANCE);
+            boolean isDeviationSmall = deviation <= maxUnbalance(activeDocs);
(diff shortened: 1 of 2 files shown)


baldersheim (approved):

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/Group.java
```diff
@@ -79,6 +79,10 @@ void aggregateNodeValues() {
         }
     }
 
+    double maxUnbalance(long activeDocs) {
+        return Math.max(1, activeDocs * MAX_UNBALANCE);
```
bratseth, line 83: If we have just a few documents we can easily end up with a deviation which is more than 1 and 10%. For example, with 2 nodes and 2 documents we have a 50% chance of both documents landing on the same node, and with that isDeviationSmall will still be false. 

There is some stats formula that will do this right for us, but since we just use it to decide whether to turn on the hits per node reduction optimization, which shouldn't be used unless there's a lot of data anyway, we could also just set isDeviationSmall to false unless the total document count is smaller than some number?

hmusum: Yes, I thought about that, not knowing the code and the impact of these changes I didn't change it much. Your suggestions sounds reasonable, though. Any suggestion about what that "small number" should be?

bratseth: A hundred maybe? But with that option we should probably also change the name of this ... I take this one if you like.

hmusum: Yes, please take this one if you have time

PR #18479, opened 2021-06-30
Lookup the symbols outside of the loop.

baldersheim:
@geirst PR
I will coordinate merge once we have som runs with new performance test.


geirst (approved):
👍

PR #18485, opened 2021-06-30
- Ensure that the engine is not accessible after it has been closed.

baldersheim:
- Also protect against multiple close.

@bratseth PR

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
@@ -201,8 +201,7 @@ public class QuestMetricsDb extends AbstractComponent implements MetricsDb {
 
     @Override
     public void close() {
-        if (closed.get()) return;
-        closed.set(true);
+        if (closed.getAndSet(true)) return;
         synchronized (nodeTable.writeLock) {
             synchronized (clusterTable.writeLock) {
                 for (SqlCompiler sqlCompiler : sqlCompilerPool)
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
@@ -10,7 +10,6 @@ import com.yahoo.config.provision.ClusterSpec;
 import com.yahoo.io.IOUtils;
 import com.yahoo.vespa.defaults.Defaults;
 import com.yahoo.yolean.concurrent.ConcurrentResourcePool;
-import com.yahoo.yolean.concurrent.ResourceFactory;
 import io.questdb.cairo.CairoEngine;
 import io.questdb.cairo.CairoException;
 import io.questdb.cairo.DefaultCairoConfiguration;
@@ -37,8 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
@@ -59,9 +57,10 @@ public class QuestMetricsDb extends AbstractComponent implements MetricsDb {
 
     private final Clock clock;
     private final String dataDir;
-    private final AtomicReference<CairoEngine> engine = new AtomicReference<>();
+    private final CairoEngine engine;
     private final ConcurrentResourcePool<SqlCompiler> sqlCompilerPool;
-    private final AtomicInteger nullRecords = new AtomicInteger();
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+
 
     @Inject
     public QuestMetricsDb() {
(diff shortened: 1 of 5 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -55,10 +57,11 @@
 
     private final Clock clock;
     private final String dataDir;
-    private final CairoEngine engine;
+    private final AtomicReference<CairoEngine> engine = new AtomicReference<>();
+    // TODO Is this safe, or even possible. ThreadLocal members should be static
```
bratseth, line 61: There's only ever one instance of this class so it doesn't matter, but having a threadlocal per instance just means you'll have one threadlocal item per thread+instance. Here the threadlocal is just to create fewer instances, there's no constraint that there must be max one per thread.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -222,7 +229,7 @@ private void ensureNodeTableIsUpdated() {
 
     private void ensureClusterTableIsUpdated() {
         try {
-            if (0 == engine.getStatus(newContext().getCairoSecurityContext(), new Path(), clusterTable.name)) {
+            if (0 == engine.get().getStatus(newContext().getCairoSecurityContext(), new Path(), clusterTable.name)) {
```
bratseth, line 232: Now we need to check for null here too.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -196,8 +199,12 @@ public void gc() {
 
     @Override
     public void close() {
-        if (engine != null)
-            engine.close();
+        synchronized (clusterTable.writeLock) {
+            CairoEngine myEngine = engine.getAndSet(null);
```
bratseth, line 203: This seems a little too complicated to me, why not just have an atomic boolean "closed"?
But we still need to check for it everywhere - do we actually have any problems with this?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
baldersheim: Ok, I concluded that was the case too. However I do not think that you want to use thread locals for this. The SqlCompilers needs to be closed. I assume prior to the engine too, but that you are not able to with thread locals.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -196,8 +199,12 @@ public void gc() {
 
     @Override
     public void close() {
-        if (engine != null)
-            engine.close();
+        synchronized (clusterTable.writeLock) {
```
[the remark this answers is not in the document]
baldersheim: I noticed, just removed it.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
baldersheim: Alternatively, just wait for the nullptr exception and fix the incorrect usage. Nothing should be using the engine after it been closed.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
bratseth: I was advised to do it by one of the QuestDb authors, but you are right they have a close() though. But have we seen any problems related to closing?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
baldersheim: We have a issue that shows up in JNI or very lowlevel(UnSafe) java code that we can not explain.
Then it is at least good to verify that we are not accessing objects after they are invalidated.

Using an AtomicBoolean is fine and add checks. This was just a suggestion. There might be better ways to do this.
I do not know the threading model here so I can not say what is the best approach.
But detecting use after close, and closing down the SqlCompilers are minimum I think.
I see that the SqlExecutionContext is Closable too, but calling close() on it seems dangerous as it seems to close a shared resource(MessageBus).

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
baldersheim: I have not looked in detail at the core dumps so I can not say. I guess we need to dive in.

bratseth:
Can I work a bit on this PR now? Seems the simplest is to just be extremely conservative with how we invoke it.

baldersheim:
Go ahead.

bratseth:
@baldersheim please take a look at the last commit. 

This still doesn't protect against concurrent *read* and close. If we need that too it's starting to become cumbersome ...

bratseth:
@baldersheim PTAL

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/QuestMetricsDb.java
```diff
@@ -196,8 +201,15 @@ public void gc() {
 
     @Override
     public void close() {
-        if (engine != null)
-            engine.close();
+        if (closed.get()) return;
+        closed.set(true);
```
baldersheim, line 205: you need to use if (closed.getAndSet(true)) return to make it atomic.
If not there is a race betwene get() and set(true).

baldersheim (commented):
Apart from getAndSet you should be fine.

PR #18488, opened 2021-07-01
Install model evaluation via configmodel [run-systemtest]

gjoranv:
FYI: @lesters

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
@@ -187,6 +187,7 @@ public class ContainerModelBuilder extends ConfigModelBuilder<ContainerModel> {
         throwUponRestApi(spec);  // TODO: remove
         addServlets(deployState, spec, cluster);
         addModelEvaluation(spec, cluster, context);
+        addModelEvaluationBundles(cluster);
 
         addProcessing(deployState, spec, cluster);
         addSearch(deployState, spec, cluster);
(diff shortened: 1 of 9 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -565,6 +566,14 @@ private void addModelEvaluation(Element spec, ApplicationContainerCluster cluste
         cluster.setModelEvaluation(new ContainerModelEvaluation(cluster, profiles));
     }
 
+    protected void addModelEvaluationBundles(ApplicationContainerCluster cluster) {
+        /* These bundles are added to all application container clusters, even if they haven't
+         * declared 'model-evaluation' in services.xml, because there are many public API packages
+         * in the model-evaluation bundle that could be used by customer code. */
+        cluster.addPlatformBundle(ContainerModelEvaluation.MODEL_EVALUATION_BUNDLE_FILE);
+        cluster.addPlatformBundle(ContainerModelEvaluation.MODEL_INTEGRATION_BUNDLE_FILE);
```
bratseth, line 574: I think we should only include model-evaluation by default?

gjoranv: Discussed this with @lesters. `model-evaluation` has many publicApi packages, and the code requires model-integration to run. So it seems unsafe to only install the smaller bundle.

gjoranv: BTW, there is a plan to remove tensorflow from model-integration, which will reduce its size considerably.

bjorncs (approved):

PR #18493, opened 2021-07-01
add some new interfaces

arnej27959:
* add MapListener interface
* add MapSource interface
* add ProxyMapSource for simple forwarding

@havardpe please review (tomorrow)

### DIFF
--- a/slobrok/src/vespa/slobrok/server/proxy_map_source.h
+++ b/slobrok/src/vespa/slobrok/server/proxy_map_source.h
@@ -7,6 +7,9 @@
 
 namespace slobrok {
 
+/**
+ * Proof-of-concept implementation of MapSource broadcasting.
+ **/
 class ProxyMapSource : public MapSource, public MapListener {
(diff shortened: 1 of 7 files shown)


slobrok/src/vespa/slobrok/server/proxy_map_source.h
```diff
@@ -0,0 +1,26 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include "map_source.h"
+#include <set>
+
+namespace slobrok {
+
+class ProxyMapSource : public MapSource, public MapListener {
```
havardpe, line 10: why do we need this? (why not just let the leaf map listeners listen directly to the original map source?)

havardpe (approved):

geirst (commented):
Would also be nice with some class comments on the new interfaces.

PR #18498, opened 2021-07-01
Deploy tester and real apps in parallell

jonmv:
@freva or @mpolden please review.


mpolden (approved):
👍

freva:
Discussed this with @jonmv in a thread in #vespa-devs; this wont work because `deployTester` is not retrying and configserver does not block until provisioning succeeds for external deployments, so this will deploy the tester app once and then fail because the hosts are not ready yet.

To be able to redeploy tester app multiple times (as we do with the real), we need to either store the key/cert somewhere or be able to regenerate it across the different controllers (VESPA-21067)

PR #18502, opened 2021-07-01
Print which tenant and session ID failed to load

hakonhall:
Getting a lot of 

```
[2021-07-01 15:25:02.904] configserver     Unable to read session status, assuming it was deleted
```

which doesn't say much.


hmusum (approved):

hmusum:
The reason this shows up now are the changes in https://github.com/vespa-engine/vespa/pull/18474
Earlier we handled this silently, since `getData()` returned null and we would then return Status.NEW when parsing.

I'll look more into this and see if it can be handled better or of this is some bug or we don't set status correctly when a session is created.

PR #18503, opened 2021-07-01
Revert "Use injected config in SessionCache"

jonmv:
Reverts vespa-engine/vespa#18499

@bjorncs broke clustercontrollers it seems


jonmv:
Ah rather it was containers with a feedhandler, but no docproc, that broke.

PR #18513, opened 2021-07-02
Implement use of multiple rpc targets in mbus

baldersheim:
@geirst PR
This follows the pattern from the java implementation.

### DIFF
--- a/messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
+++ b/messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
@@ -4,23 +4,41 @@
 
 namespace mbus {
 
-RPCTargetPool::Entry::Entry(RPCTarget::SP target, uint64_t lastUse) :
-    _target(target),
-    _lastUse(lastUse)
+RPCTargetPool::Entry::Entry(std::vector<RPCTarget::SP> targets, uint64_t lastUse)
+    : _targets(std::move(targets)),
+      _lastUse(lastUse),
+      _next(0)
 { }
 
-RPCTargetPool::RPCTargetPool(double expireSecs) :
-    _lock(),
-    _targets(),
-    _timer(new SteadyTimer()),
-    _expireMillis(static_cast<uint64_t>(expireSecs * 1000))
+RPCTarget::SP
+RPCTargetPool::Entry::getTarget(const LockGuard &, uint64_t now) {
+    if (_next >= _targets.size()) {
+        _next = 0;
+    }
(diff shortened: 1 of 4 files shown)


messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
```diff
@@ -4,23 +4,41 @@
...
+    RPCTarget::SP target =  _targets[_next++];
+    if ( ! target->isValid()) return RPCTarget::SP();
+    _lastUse = now;
+    return target;
+}
+
+bool
+RPCTargetPool::Entry::inUse(const LockGuard &) const {
+    for (const auto & target : _targets) {
+        if (target.use_count() > 1) return true;
```
geirst, line 27: Should this be "_> 0_" instead ?

messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
```diff
@@ -4,23 +4,41 @@
...
-    _targets(),
-    _timer(new SteadyTimer()),
-    _expireMillis(static_cast<uint64_t>(expireSecs * 1000))
+RPCTarget::SP
+RPCTargetPool::Entry::getTarget(const LockGuard &, uint64_t now) {
+    if (_next >= _targets.size()) {
+        _next = 0;
+    }
+    RPCTarget::SP target =  _targets[_next++];
+    if ( ! target->isValid()) return RPCTarget::SP();
```
geirst, line 19: Consider using braces.

messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
geirst: Consider using braces.

geirst (commented):
Consider adding a basic unit test for RPCTargetPool (in targetpool.cpp) that verifies we get different targets for the same spec/address.

messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
baldersheim: Braces done.
target is a shared_ptr so use_count will never be 0. If it is larger than 1 it means there is another reference than the one in the pool active.

messagebus/src/vespa/messagebus/network/rpctargetpool.cpp
baldersheim: Fixed.

geirst (commented):
👍  We should also add a unit test for this.

PR #18520, opened 2021-07-02
Wire in sending large rankprofiles as files controlled by feature flag

baldersheim:
@hmusum PR
This is guarded by a feature flag so it should a be a no-op for now


hmusum (approved):
LGTM, 

Not related to this change, but code would have been easier to read with a class for properties instead of `List<Pair<String, String>>`

PR #18525, opened 2021-07-05
Session cache with injected config and network [run-systemtest]

jonmv:
This reapplies session cache with injected config, fixing a problem with chain client setup, and separating out the lowest level representation of "the world" in Slobrok to a reusable component, to handle several graph generations being simultaneously registered as destinations for the same session ID.

### DIFF
--- a/container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/test/RemoteClient.java
+++ b/container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/test/RemoteClient.java
@@ -1,9 +1,17 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.messagebus.jdisc.test;
 
+import com.yahoo.cloud.config.SlobroksConfig;
 import com.yahoo.jrt.ListenFailedException;
 import com.yahoo.jrt.slobrok.server.Slobrok;
-import com.yahoo.messagebus.*;
+import com.yahoo.messagebus.Message;
+import com.yahoo.messagebus.MessageBus;
+import com.yahoo.messagebus.MessageBusParams;
+import com.yahoo.messagebus.Protocol;
+import com.yahoo.messagebus.Reply;
+import com.yahoo.messagebus.Result;
+import com.yahoo.messagebus.SourceSession;
+import com.yahoo.messagebus.SourceSessionParams;
 import com.yahoo.messagebus.network.local.LocalNetwork;
 import com.yahoo.messagebus.network.rpc.RPCNetwork;
 import com.yahoo.messagebus.network.rpc.RPCNetworkParams;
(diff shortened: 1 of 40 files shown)


jonmv:
Opted for a multiplexer between message bus and its network. This is set up in an injectable component, and a holder makes sure it is persistent across component graph generations. PTAL @bjorncs .

jonmv:
Hmm, seems there's still something left to do when switching graphs, probably related to the JDisc `Server` setup.

jonmv:
Fixed a bug where non-broadcast sessions wouldn't get messages. Also made sure sessions are deregistered on server-close, which is, AFAICS, the intention. Finally, the last available document protocol is used for decoding incoming messages, which is the same as the current, where self-subscription means the protocol is immediately updated. In fact, this change is better, as the new protocol won't be available very much too early for restart-on-deploy users.

container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/test/RemoteClient.java
```diff
@@ -40,28 +46,22 @@ public Reply awaitReply(int timeout, TimeUnit unit) throws InterruptedException
         return queue.awaitReply(timeout, unit);
     }
 
-    public String slobrokId() {
-        return slobrokId;
+    public SlobroksConfig slobroksConfig() {
+        return TestUtils.configFor(slobrok);
```
bjorncs, line 50: Should this method be used "network==true" (production/non-test?) code path in constructor?

container-messagebus/src/test/java/com/yahoo/messagebus/jdisc/MbusServerTestCase.java
```diff
@@ -355,15 +351,13 @@ public String name() {
...
         @Override
         public ResourceReference refer() {
             ++refCount;
-            return new ResourceReference() {
-                @Override
-                public void close() {
-                    --refCount;
-                }
-            };
+            return () -> --refCount;
```
bjorncs, line 360: 😄

messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
```diff
@@ -0,0 +1,143 @@
...
+ * time, with all copies potential users of the network interface, but where each message bus-registered session
+ * should belong to a single message bus. This class solves these problems by tracking which sessions are
+ * active in which message bus instance, and by (de)registering only when a session is registered to (no) message
+ * bus instances.
+ *
+ * In time, this should allow us to get rid of the shared-this-and-that in the container, too ...
+ *
+ * @author jonmv
+ */
+public class NetworkMultiplexer implements NetworkOwner {
```
bjorncs, line 27: Nice! 🥇

bjorncs (approved):
👍

container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/test/RemoteClient.java
bjorncs: Did not notice the package name...

jonmv:
Unrelated test failure, as far as I can see.

PR #18535, opened 2021-07-05
Upgrade dev at dep-local night

jonmv:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
@@ -62,9 +62,11 @@ public class DeploymentExpirer extends ControllerMaintainer {
                                        .map(type -> new JobId(instance, type));
         if (jobId.isEmpty()) return false;
 
-        return controller().jobController().last(jobId.get())
-                           .flatMap(Run::end)
-                           .map(end -> end.plus(ttl.get()).isBefore(controller().clock().instant()))
+        return controller().jobController().jobStatus(jobId.get())
+                           .runs().descendingMap().values().stream()
+                           .filter(run -> ! run.isRedeployment())
+                           .findFirst()
+                           .map(run -> run.start().plus(ttl.get()).isBefore(controller().clock().instant()))
                            .orElse(false);
     }
 
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
@@ -12,6 +12,7 @@ import com.yahoo.vespa.hosted.controller.deployment.Run;
 import com.yahoo.yolean.Exceptions;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.util.Optional;
 import java.util.logging.Level;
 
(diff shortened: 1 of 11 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgraderTest.java
```diff
@@ -0,0 +1,101 @@
...
+        // After the upgrade, the dev app is mostly (re)deployed to at night, but this doesn't affect what is likely the night.
+        tester.controllerTester().upgradeSystem(new Version(7, 9, 11));
+        tester.clock().advance(Duration.ofHours(48));
+        upgrader.maintain();
+        assertEquals(tester.clock().instant().truncatedTo(MILLIS), tester.jobs().last(devApp.instanceId(), devUsEast1).get().start());
+    }
+
+    @Test
+    public void testNight() {
+        assertEquals(14, mostLikelyWeeHour(new int[]{ 0, 1, 2, 3, 4, 5, 6 }));
```
freva, line 95: Not that it's very important, but I think its easier and you get better result if you just count deployment frequencies by hour and then minimize product sum of `numDeployments(hour + offset) / (offset + 1)`, i.e.:
```
static int mostLikelyWeeHour(int[] starts) {
    int[] buckets = new int[24];
    for (int start : starts)
        buckets[start]++;

    int best = -1;
    double min = Double.MAX_VALUE;
    for (int i = 12; i < 36; i++) {
        double sum = buckets[i % 24];
        for (int j = 1; j < 13; j++)
            sum += (buckets[(i + j) % 24] + buckets[(i - j) % 24]) / (j + 1d);

        if (sum < min) {
            min = sum;
            best = i;
        }
    }
    return best % 24;
}
```
The expectation for test cases then becomes 15, 13, 15 and 18. So 0, 1, 2, ..., 6 gives the same hour as 6, 5, 4, ..., 0 and 0, 12 alternating gives hour furthest apart from them - 18.

freva (approved):

freva:
```
16:26:04 [ERROR] Failures: 
16:26:04 [ERROR]   ControllerApiTest.testControllerApi:53 maintenance.json expected:<...intainer"},{"name":"[]EndpointCertificateM...> but was:<...intainer"},{"name":"[DeploymentUpgrader"},{"name":"]EndpointCertificateM...>
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentExpirer.java
```diff
@@ -62,9 +62,11 @@ private boolean isExpired(Deployment deployment, ApplicationId instance) {
                                        .map(type -> new JobId(instance, type));
         if (jobId.isEmpty()) return false;
 
-        return controller().jobController().last(jobId.get())
-                           .flatMap(Run::end)
-                           .map(end -> end.plus(ttl.get()).isBefore(controller().clock().instant()))
+        return controller().jobController().jobStatus(jobId.get())
+                           .runs().descendingMap().values().stream()
+                           .filter(run -> ! run.isRedeployment())
```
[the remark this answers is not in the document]
jonmv: Yes, I missed that one. Thanks!

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgraderTest.java
jonmv: I like the objective function, but I'm keeping the weights :)

jonmv:
PTAL :)

freva (approved):
:star:

PR #18556, opened 2021-07-07
Verify status path

mpolden:
@freva


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1680,6 +1680,7 @@ private HttpResponse service(String tenantName, String applicationName, String i
 
         if ("container-clustercontroller".equals((serviceName)) && restPath.contains("/status/")) {
             String[] parts = restPath.split("/status/");
+            if (parts.length < 2) throw new IllegalArgumentException("Invalid path '" + restPath + "'");
```
freva, line 1683: This should always produce at least 2 since we know that `restPath.contains("/status/")`, but java returns only 1 result when the match is at the end of the string, fixed in https://github.com/vespa-engine/vespa/pull/18554

PR #18563, opened 2021-07-07
Adjust reporting of disk and memory usage in proton.

toregge:
@geirst : please review


geirst (approved):
👍

geirst:
With this change we track the max resource usage (disk or memory) sampled in a metric reporting interval, and use the max value when reporting metrics.

PR #18565, opened 2021-07-07
Return partial response if getQuota fails

olaaun:


andreer (approved):

andreer:
Code looks OK but why are we doing this? :)

olaaun:
>Code looks OK but why are we doing this? :)

To prevent the console from being unusable if the billing controller is down

PR #18573, opened 2021-07-08
Avoid checking and creating node

hmusum:
If node is created between exists() and create() this will fail,
just catch exception if node already exists and ignore it


jonmv (approved):
Already done by `com.yahoo.vespa.curator.Curator`, but that one doesn't use `String` for paths. It would be great if we could migrate to use this helper everywhere, instead of manipulating the `CuratorFramework` directly.

hmusum:
Yes, that's exactly the plan, working on it

PR #18575, opened 2021-07-08
Detect exclusive host when calculating real resources

mpolden:
The `HostProvisioner` implementation detects an exclusive host correctly and
calculates a lower overhead. However, this code assumed that the host was
not exclusive and calculated a higher overhead.

This mismatch then resulted in `GroupPreparer` rejecting the offered host.

@freva


freva (approved):
Great! :1st_place_medal:

PR #18570, opened 2021-07-08
Don't deploy or install container-jersey2 [run-systemtest]

gjoranv:
I will verify with a system test locally before merging.

PR #18578, opened 2021-07-08
Revert "Don't deploy or install container-jersey2 [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#18570

Seems like a jar is installed but not packaged:
 19:18:33 error: Installed (but unpackaged) file(s) found:
19:18:33    /opt/vespa/lib/jars/container-jersey2-jar-with-dependencies.jar


gjoranv:
Thanks, I didn't know that it also must be removed from install.

gjoranv (commented):
👍

PR #18584, opened 2021-07-09
Cleanup Curator and ConfigCurator usage [run-systemtest]

hmusum:
Lots of cleanup in use of Curator and ConfigCurator and eventually the removal of ConfigCurator

~We lose the check for max bytes in a ZooKeeper node, I'll look into reintroducing that.~ Fixed.


hmusum:
@mpolden maybe you want to have a look

mpolden (approved):
Great!

PR #18587, opened 2021-07-10
Increase session timeout

hmusum:
30 seconds seem to be too little with heavily loaded servers and/or
slow disks.


bratseth (approved):

mpolden (approved):
Let's try it.

PR #18599, opened 2021-07-13
try to trap spurious failure

arnej27959:
* we have seen spurious failures when verifying output from accent dropping;
  so far nothing reproducible, so add some extra logging and retry once
  if it happens (in case it's some kind of race-condition glitch).

@geirst please review


geirst (approved):
👍

PR #18608, opened 2021-07-14
Optimize assign updates to tensor attribute with hnsw index by handli…

geirst:
…ng them as two phase put operations.

This ensures that the costly part of inserting into the hnsw index is done in parallel in the prepare put step.

@vekterli please review


vekterli (approved):
:+1:

PR #18619, opened 2021-07-15
Generalize how address space usage is reported from attribute vectors.

geirst:
This prepares for reporting address space usage for more attribute vector components,
not only enum store and multi-value mapping.

@vekterli please review
@toregge FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/address_space_components.h
+++ b/searchlib/src/vespa/searchlib/attribute/address_space_components.h
@@ -1,0 +1,21 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include <vespa/vespalib/stllike/string.h>
+#include <vespa/vespalib/util/address_space.h>
+
+namespace search {
+
+/**
+ * A set of components in an attribute vector that use address space.
+ */
+class AddressSpaceComponents {
+public:
+    static vespalib::AddressSpace default_enum_store_usage();
+    static vespalib::AddressSpace default_multi_value_usage();
+    static vespalib::string enum_store;
+    static vespalib::string multi_value;
+};
+
+}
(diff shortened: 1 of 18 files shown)


searchlib/src/vespa/searchlib/attribute/address_space_components.h
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A set of components in an attribute vector that use address space.
+ */
+class AddressSpaceComponents {
+public:
+    static vespalib::AddressSpace default_enum_store_usage();
+    static vespalib::AddressSpace default_multi_value_usage();
+    static vespalib::string enum_store;
+    static vespalib::string multi_value;
```
vekterli, line 18: Should these string instances be explicitly marked `const` to prevent any accidental meddling with data that presumably should be considered immutable?

searchlib/src/vespa/searchlib/attribute/address_space_usage.cpp
```diff
@@ -1,33 +1,52 @@
...
+{
+    // TODO: Remove this constructor and instead add usage for each relevant component explicit.
+    add(AddressSpaceComponents::enum_store, enum_store_usage);
+    add(AddressSpaceComponents::multi_value, multi_value_usage);
+}
+
+void
+AddressSpaceUsage::add(const vespalib::string& component, const vespalib::AddressSpace& usage)
+{
+    _map[component] = usage;
```
vekterli, line 27: Is there a use case where `usage` should be _added_ to any existing mapping of `component` instead of overwriting it?

vekterli (approved):
:+1: a couple of minor questions/comments

searchlib/src/vespa/searchlib/attribute/address_space_usage.cpp
geirst: No. I will rename the function to _set()_ to better reflect the intent.

searchlib/src/vespa/searchlib/attribute/address_space_components.h
geirst: Fixed

searchlib/src/vespa/searchlib/attribute/address_space_usage.cpp
geirst: Fixed

PR #18636, opened 2021-07-19
Stop using enums in NodeRepositoryNode fields

mpolden:
This class is only used for serialization and should therefore not use explicit
types. We can now introduce new values for these fields on the server side
without having to update this.

@olaaun


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
```diff
@@ -433,6 +424,19 @@ public void setSwitchHostname(String switchHostname) {
...
+    public boolean hasType(NodeType type) {
+        return type.name().equals(getType());
+    }
+
+    public boolean hasState(NodeState state) {
+        return state.name().equals(getState());
+    }
+
+    // --- end
+
```
olaaun, line 439: Are these used anywhere?

olaaun (approved):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
mpolden: Yes, some internal code still uses this directly.

PR #18638, opened 2021-07-20
Improve exception for package with multiple installed versions

mpolden:
@olaaun or @aressem


aressem:
This one will still fail here if multiple packages are found:
https://git.vzbuilders.com/vespa/vespa-yahoo/blob/d48a699ff651afbfc4ea4c763ffe95b64cff0c9d/hosted/host-admin/src/main/java/com/yahoo/vespa/hosted/hostadmin/task/main/InstallYumPackages.java#L67

mpolden:
Yes, you should have gotten a PR in internal repo.

aressem:
Right, thanks.

aressem (approved):

PR #18647, opened 2021-07-27
Include tenant iam roles with perf deployments

tokle:
@jobergum fyi


freva (approved):
Why is it just prod and perf?

tokle:
Should be everywhere there is a chance we have exclusive hosts (=everywhere), but at the same time need to make sure the policy document does not grow outside the limits.

PR #18648, opened 2021-07-27
Support reserving hosts to nodes of given cluster type

freva:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -189,10 +189,13 @@ private boolean offeredNodeHasParentHostnameAlreadyAccepted(NodeCandidate candid
...
         if (candidate.parentHostname().isEmpty()) return false;
 
-        // In dynamic provisioned zones a node requiring exclusivity must be on a host that has exclusiveTo equal to its owner
-        if (nodeRepository.zone().getCloud().dynamicProvisioning())
-            return requestedNodes.isExclusive() &&
-                    ! candidate.parent.flatMap(Node::exclusiveTo).map(application::equals).orElse(false);
+        // In dynamic provisioned zones, exclusivity is violated if...
+        if (nodeRepository.zone().getCloud().dynamicProvisioning()) {
+                    // If either the parent is dedicated to a cluster type different from this cluster
+            return  ! candidate.parent.flatMap(Node::exclusiveToClusterType).map(cluster.type()::equals).orElse(true) ||
```
hakonhall, line 195: This first part of the || condition will never actually be true given the other changes in this PR, agree?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -189,10 +189,13 @@ private boolean offeredNodeHasParentHostnameAlreadyAccepted(NodeCandidate candid
     private boolean violatesExclusivity(NodeCandidate candidate) {
         if (candidate.parentHostname().isEmpty()) return false;
 
-        // In dynamic provisioned zones a node requiring exclusivity must be on a host that has exclusiveTo equal to its owner
-        if (nodeRepository.zone().getCloud().dynamicProvisioning())
-            return requestedNodes.isExclusive() &&
-                    ! candidate.parent.flatMap(Node::exclusiveTo).map(application::equals).orElse(false);
+        // In dynamic provisioned zones, exclusivity is violated if...
```
hakonhall, line 192: 3rd edge case once/if both admin and generic shared hosts needs to be supported within a single zone:

If the shared-host flag defines an admin shared host, and the candidate is an admin node on a shared host w/o cluster type, then this method ought to return true, but will actually returns false.

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
freva: Yes, this should only happen if `exclusiveToClusterType` somehow changed later, with this PR this is only possible if operator patches the node via /nodes/v2

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
freva: Maybe, I wouldn't say that scenario violates exclusivity, it's just preferable.

Either way, I think we'd prefer to solve this through a dedicated maintainer that would slowly migrate off non `exclusiveToClusterType` hosts instead of starting migration of all nodes of a cluster type within a redeployment interval (30 mins) after a sudden flag change.

hakonhall: otherwise full admin shared hosts would not induce provisioning of additional shared hosts, until the non-admin shared hosts are filled too

freva: Good point, yes, need to handle that somehow without also causing immediate mass-migration for already allocated nodes.

hakonhall: Right. shouldRetire() should probably return false for this case. So it's a "soft" exclusivity violation: It's OK to keep such an allocation, but not make a new one.

PR #18651, opened 2021-07-28
check extra field for change for reindexing

bjormel:

This is an attempt to address the re-indexing issues seen by an application.

Issue was caused by comparing two SimpleLinguistics object, in testing at least, in the class NormalizeExpression

Changed to
`        if (linguistics.getClass() != other.linguistics.getClass()) return false;
`

Not sure if this is the best way to address this.


bjormel:
@jonmv

bjormel:
@arnej27959 did this in #18682

PR #18655, opened 2021-07-29
Metric for last successful update of IP tables

bjormel:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/AclMaintainer.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/AclMaintainer.java
@@ -2,7 +2,10 @@
 package com.yahoo.vespa.hosted.node.admin.maintenance.acl;
 
 import com.google.common.net.InetAddresses;
+import com.yahoo.vespa.hosted.node.admin.configserver.noderepository.NodeSpec;
 import com.yahoo.vespa.hosted.node.admin.container.ContainerOperations;
+import com.yahoo.vespa.hosted.node.admin.container.metrics.Dimensions;
+import com.yahoo.vespa.hosted.node.admin.container.metrics.Metrics;
 import com.yahoo.vespa.hosted.node.admin.nodeagent.NodeAgentContext;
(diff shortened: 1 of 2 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/AclMaintainer.java
```diff
@@ -43,10 +48,19 @@
...
+    private final Metrics metrics;
+    private final Map<String, Long> lastSuccess;
+    private static final String METRIC_NAME_POSTFIX = ".acl.age";
 
-    public AclMaintainer(ContainerOperations containerOperations, IPAddresses ipAddresses) {
+    public AclMaintainer(ContainerOperations containerOperations, IPAddresses ipAddresses, Metrics metrics) {
         this.containerOperations = containerOperations;
         this.ipAddresses = ipAddresses;
+        this.metrics = metrics;
+        this.lastSuccess = new HashMap<>(){{
```
mpolden, line 59: This creates an anonymous subclass of `HashMap`, which is not ideal. Consider doing `new HashMap<>(Map.of(...))` instead.

Consider also extracting a variable for `System.currentTimeMillis` so that both versions have the same initial timestamp.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/acl/AclMaintainer.java
```diff
@@ -115,6 +131,31 @@ private boolean edit(NodeAgentContext context, String table, IPVersion ipVersion
         };
     }
 
+    void updateMetric(NodeAgentContext context, boolean updated, String ipVersion) {
+        Dimensions dimensions = generateDimensions(context);
+        if (!updated) {
+            metrics.declareGauge(Metrics.APPLICATION_NODE, ipVersion + METRIC_NAME_POSTFIX, dimensions, Metrics.DimensionType.PRETAGGED)
+                    .sample((System.currentTimeMillis() - lastSuccess.get(ipVersion)) / 1000);
```
mpolden, line 138: Consider introducing a variable for the metric value, and then simplify to one `declareGauge` call which avoids repeating all the other unchanged arguments.

mpolden (approved):
👍

PR #18668, opened 2021-08-02
Revert "Bit vector are allocated query time and can not take the cost of usin…"

baldersheim:
Reverts vespa-engine/vespa#18666
@jobergum PR
Needs another fix too.


jobergum:
👍

PR #18669, opened 2021-08-02
Revert "Revert "Bit vector are allocated query time and can not take the cost of usin…""

baldersheim:
Reverts vespa-engine/vespa#18668


baldersheim:
@jobergum PR

jobergum (approved):

PR #18671, opened 2021-08-02
Use std::quick_exit instead of abort to avoid coredump when it will n…

baldersheim:
…ot provide any more information.

@arnej27959 or @jobergum or @bjormel or @kkraune PR
@vekterli and @geirst FYI


bjormel (approved):

arnej27959:
the rest of our code base uses std::_Exit for this.
it looks like the only difference is that you can use std::at_quick_exit to register some
cleanup methods with quick_exit().

PR #18673, opened 2021-08-03
make runnable not only from cwd

kkraune:
I am not sure if this is the best fix - when unzipping the artifact I get

````
$ ll
-rw-r--r--  1 xxx  staff      154 Aug  1 11:58 logging.properties
-rwxr-xr-x@ 1 xxx  staff      288 Aug  3 10:54 vespa-feed-client
-rw-r--r--  1 xxx  staff  8784885 Aug  1 12:00 vespa-feed-client-cli-jar-with-dependencies.jar
````

FYI @bjorncs / @jonmv


arnej27959 (approved):
it looks like this should mostly work.

bjorncs (commented):
👍

PR #18676, opened 2021-08-03
Arnej/avoid normalize triggers reindexing

arnej27959:

@baldersheim please review
@bjormel FYI
@jonmv FYI
@bratseth FYI

### DIFF
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/NormalizeExpression.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/NormalizeExpression.java
@@ -81,7 +81,7 @@ public final class NormalizeExpression extends Expression {
     public boolean equals(Object o) {
         if (!(o instanceof NormalizeExpression)) return false;
         NormalizeExpression other = (NormalizeExpression)o;
-        if (linguistics != other.linguistics) return false;
+        if (linguistics.getClass() != other.linguistics.getClass()) return false;
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/NormalizeExpression.java
```diff
@@ -81,7 +81,7 @@ public String toString() {
     public boolean equals(Object o) {
         if (!(o instanceof NormalizeExpression)) return false;
         NormalizeExpression other = (NormalizeExpression)o;
-        if (linguistics != other.linguistics) return false;
+        if (linguistics.getClass() != other.linguistics.getClass()) return false;
```
baldersheim, line 84: Any reason why this differs from the one in NGramExpression ?
Are the classes from the same bundle, or will might they be from different model versions ?

baldersheim (approved):
If my comments have a reasonable answer you are good to go.

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/NormalizeExpression.java
arnej27959: > Any reason why this differs from the one in NGramExpression ?

here linguistics must be present, but in ngram it can be null.

> Are the classes from the same bundle, or will might they be from different model versions ?

after more thinking I'm leaning towards just removing the check for equal linguistics; I think we have (or are planning on) some other mechanism for triggering reindexing when we upgrade Yell or make other changes to linguistics config.  I'm guessing we'll get the same class when they are from different model versions, but I'm not 100% sure.

PR #18681, opened 2021-08-04
Balder/also perform topk without global filter

baldersheim:
@arnej27959 PR


arnej27959 (approved):
this looks good to me

PR #18684, opened 2021-08-05
Specify number of proxies, try 2 [run-systemtest]

hakonhall:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -122,6 +120,12 @@
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundBooleanFlag ORCHESTRATE_MISSING_PROXIES = defineFeatureFlag(
+            "orchestrate-missing-proxies", false,
+            List.of("hakonhall"), "2020-08-05", "2020-10-05",
+            "Whether the Orchestrator can assume any missing proxy services are down.",
+            "Takes effect immediately");
```
freva, line 127: Why is it immediately? Isn't in on construction of `OrchestratorImpl`, i.e. the next cfg restart?

freva (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hakonhall: Thanks, I was hoping to implement this, but it got too intrusive so I reverted to picking up the flag value on config server start.  Doc updated.

freva (approved):

PR #18691, opened 2021-08-09
add LocalRpcMonitorMap serving local history

arnej27959:

@havardpe please review

### DIFF
--- a/slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
+++ b/slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
@@ -28,7 +28,7 @@ LocalRpcMonitorMap::lookup(const ServiceMapping &mapping) {
     return nullptr;
 }
 
-ServiceMapHistory & LocalRpcMonitorMap::localHistory() {
+ServiceMapHistory & LocalRpcMonitorMap::history() {
     return _history;
 }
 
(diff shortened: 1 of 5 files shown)


slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
```diff
@@ -0,0 +1,111 @@
...
+}
+
+LocalRpcMonitorMap::~LocalRpcMonitorMap() = default;
+
+LocalRpcMonitorMap::PerService *
+LocalRpcMonitorMap::lookup(const ServiceMapping &mapping) {
+    auto iter = _map.find(mapping.name);
+    if (iter != _map.end()) {
+        PerService & psd = iter->second;
+        if (psd.mapping.spec == mapping.spec) {
```
havardpe, line 24: should this be allowed to be false? (assert?)

slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.h
```diff
@@ -0,0 +1,61 @@
...
+    FRT_Supervisor &_supervisor;
+    vespalib::GenCnt _genFromGlobal;
+
+    PerService * lookup(const ServiceMapping &mapping);
+    
+public:
+    LocalRpcMonitorMap(FRT_Supervisor &_supervisor);
+    ~LocalRpcMonitorMap();
+
+    ServiceMapHistory & localHistory();
```
havardpe, line 48: consider just calling this 'history', since the 'local' part relates to the class itself

havardpe (approved):

havardpe:
The map might not need to keep an extra service mapping object (the managed rpc server also has this information, but would need to be always present)

arnej27959:
@havardpe Please take another look now

havardpe (approved):

PR #18697, opened 2021-08-09
Respect thresholds for max requests/time for HTTP/2 connections

bjorncs:
FYI @bjormel

### DIFF
--- a/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpRequestDispatch.java
+++ b/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpRequestDispatch.java
@@ -13,8 +13,13 @@ import com.yahoo.jdisc.handler.RequestHandler;
 import com.yahoo.jdisc.http.ConnectorConfig;
 import com.yahoo.jdisc.http.HttpHeaders;
 import com.yahoo.jdisc.http.HttpRequest;
+import org.eclipse.jetty.http2.ErrorCode;
+import org.eclipse.jetty.http2.server.HTTP2ServerConnection;
+import org.eclipse.jetty.io.Connection;
 import org.eclipse.jetty.io.EofException;
+import org.eclipse.jetty.server.HttpConnection;
 import org.eclipse.jetty.server.Request;
+import org.eclipse.jetty.util.Callback;
 
 import javax.servlet.AsyncContext;
 import javax.servlet.ServletInputStream;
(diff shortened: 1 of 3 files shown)


tokle (approved):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/HttpRequestDispatch.java
```diff
@@ -143,24 +147,34 @@ public void dispatch() throws IOException {
...
+    }
+
+    private static void gracefulShutdown(Connection connection, String reason) {
+        if (connection instanceof HttpConnection) {
+            HttpConnection http1 = (HttpConnection) connection;
+            http1.getGenerator().setPersistent(false);
+        } else if (connection instanceof HTTP2ServerConnection) {
+            HTTP2ServerConnection http2 = (HTTP2ServerConnection) connection;
+            // Signal Jetty to do a graceful connection shutdown with GOAWAY frame
+            http2.getSession().close(ErrorCode.NO_ERROR.code, reason, Callback.NOOP);
```
jonmv, line 177: Not the cleanest API, but I take it you're right.

container-core/src/test/java/com/yahoo/jdisc/http/server/jetty/HttpServerTest.java
```diff
@@ -456,18 +458,51 @@ public void requireThatNoConnectionHeaderMeansKeepAliveInHttp11KeepAliveDisabled
...
+            String uri = "https://localhost:" + driver.server().getListenPort() + "/status.html";
+            for (int i = 0; i < MAX_REQUESTS - 1; i++) {
+                SimpleHttpResponse response = client.execute(SimpleRequestBuilder.get(uri).build(), null).get();
+                assertEquals(OK, response.getCode());
+            }
+            try {
+                client.execute(SimpleRequestBuilder.get(uri).build(), null).get();
+                fail();
+            } catch (ExecutionException e) {
+                // Note: this is a weakness with Apache Http Client 5; the failed stream/request will not be retried on a new connection
```
jonmv, line 502: ... !? So what effect will closing the stream have, then? Or do you mean that it results in an exception, rather than an automatic retry?

jonmv (approved):

container-core/src/test/java/com/yahoo/jdisc/http/server/jetty/HttpServerTest.java
bjorncs: Yes, this is the sad thing with Apache http client 5 and the reason why this is "do not merge".

bjorncs: Agreed to merge in offline discussion

PR #18712, opened 2021-08-11
Try using slobrok.internal.fetchLocalView also

arnej27959:
* add a MapView class holding a mirror of remote slobrok
* perform incremental fetch to fill the MapView when possible.

@havardpe please review

### DIFF
--- a/slobrok/src/vespa/slobrok/server/map_view.cpp
+++ b/slobrok/src/vespa/slobrok/server/map_view.cpp
@@ -1,0 +1,89 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "map_view.h"
+#include <vespa/log/log.h>
+LOG_SETUP(".slobrok.map_view");
+
+namespace slobrok {
+
+MapView::MapView()
+  : _map(),
+    _currGen(0),
+    _lock()
+{}
+
+MapView::~MapView() {
+    clear();
+}
+
+void MapView::apply(const MapDiff &diff) {
+    std::lock_guard guard(_lock);
+    LOG(debug, "Applying diff from gen %u", diff.fromGen.getAsInt());
+    for (const auto & name : diff.removed) {
+        auto iter = _map.find(name);
+        if (iter != _map.end()) {
+            LOG(debug, "Apply remove %s->%s", name.c_str(), iter->second.c_str());
+            ServiceMapping mapping(name, iter->second);
+            for (auto * listener : _listeners) {
+                listener->remove(mapping);
+            }
+            _map.erase(iter);
+        } else {
+            LOG(debug, "Apply remove %s [already removed]", name.c_str());
+        }
+    }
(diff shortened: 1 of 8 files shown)


slobrok/src/vespa/slobrok/server/map_view.cpp
```diff
@@ -0,0 +1,89 @@
...
+namespace slobrok {
+
+MapView::MapView()
+  : _map(),
+    _currGen(0),
+    _lock()
+{}
+
+MapView::~MapView() {
+    clear();
```
havardpe, line 16: This implies that we are ok with adding listeners that are never removed. Should we be ok with that?

slobrok/src/vespa/slobrok/server/map_view.cpp
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void MapView::apply(const MapDiff &diff) {
+    std::lock_guard guard(_lock);
+    LOG(debug, "Applying diff from gen %u", diff.fromGen.getAsInt());
```
havardpe, line 21: consider adding assert that from gen matches (map vs diff) to indicate that this needs to be handled externally.

slobrok/src/vespa/slobrok/server/map_view.h
```diff
@@ -0,0 +1,51 @@
...
+#include <mutex>
+#include <set>
+
+namespace slobrok {
+
+/**
+ * @class MapView
+ * @brief Holds a name->spec map which can be incrementally updated
+ **/
+class MapView : public MapSource
```
havardpe, line 19: MapView sounds a bit generic. Other options could be RemoteMapSource (we do it for the events) or ServiceMapMirror or similar (to match it with its partner; ServiceMapHistory)

slobrok/src/vespa/slobrok/server/remote_slobrok.h
```diff
@@ -67,9 +71,11 @@ class RemoteSlobrok: public IRpcServerManager,
     bool isConnected() const { return (_remote != nullptr); }
     void tryConnect();
     void maybePushMine();
+    void maybeStartFetch();
     void invokeAsync(FRT_RPCRequest *req, double timeout, FRT_IRequestWait *rwaiter);
     const std::string & getName() const { return _rpcserver.getName(); }
     const std::string & getSpec() const { return _rpcserver.getSpec(); }
+    const MapView &remoteView() const { return _mapView; }
```
havardpe, line 78: we should probably return MapSource here, which raises the question about mutable event listener lists...

havardpe (approved):
looks reasonable, but also touches on code that really does not want this to run with multiple threads (which makes the locking only half-way useful).

slobrok/src/vespa/slobrok/server/map_view.cpp
arnej27959: > This implies that we are ok with adding listeners that are never removed. Should we be ok with that?

it's an option to assert that listener list is empty instead, will revisit when we get it wired in.

PR #18720, opened 2021-08-12
Prepare all applications before activating when bootstrapping config …

hmusum:
…server

When upgrading to a new version when bootstrapping config server:

If we fail to prepare an application it might be because there is something
wrong with the release. In that case we don't want to activate any application.
This change makes sure we prepare all applications successfully before we
activate any application..

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
@@ -309,7 +309,7 @@ public class ConfigServerBootstrap extends AbstractComponent implements Runnable
                 logProgress(applicationCount, applications);
                 lastLogged = Instant.now();
             }
-        } while (applications.failed().size() + applications.successCount() < applicationCount);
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
```diff
@@ -221,85 +221,179 @@ private boolean redeployAllApplications() throws InterruptedException {
...
                         throw new IllegalArgumentException("Unknown deployment status " + status);
                 }
             });
             if ( ! Duration.between(lastLogged, Instant.now()).minus(Duration.ofSeconds(10)).isNegative()) {
-                logProgress(applicationCount, failedDeployments, finishedDeployments);
+                logProgress(applicationCount, applications);
                 lastLogged = Instant.now();
             }
-        } while (failedDeployments.size() + finishedDeployments.size() < applicationCount);
+        } while (applications.failed().size() + applications.successCount() < applicationCount);
```
baldersheim, line 312: failCount of failed().size().

baldersheim (approved):
See comment on failCount()

PR #18731, opened 2021-08-12
some fixups for mirror of remote slobrok map:

arnej27959:
* require symmetrical register/unregister calls
* modernize a bit
* add unit test for ServiceMapMirror
* use common mock listener in unit tests
* add an explicit shutdown() when removing a RemoteSlobrok instance,
  and perform clear() after canceling all RPC requests and tasks

@havardpe please review

### DIFF
--- a/slobrok/src/tests/service_map_mirror/service_map_mirror_test.cpp
+++ b/slobrok/src/tests/service_map_mirror/service_map_mirror_test.cpp
@@ -1,0 +1,101 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/slobrok/server/mock_map_listener.h>
+#include <vespa/slobrok/server/service_map_mirror.h>
+#include <vespa/vespalib/gtest/gtest.h>
+#include <vespa/vespalib/util/stringfmt.h>
+#include <map>
+
+using namespace vespalib;
+using namespace slobrok;
+using vespalib::make_string_short::fmt;
+
+using Map = std::map<vespalib::string, vespalib::string>;
+
+Map dump(const ServiceMapMirror &history) {
+    Map result;
+    for (const auto & entry : history.allMappings()) {
+        result[entry.name] = entry.spec;
+    }
+    return result;
+}
+
+
+void addTo(ServiceMapMirror &target, const ServiceMapping &mapping) {
+    auto cur = target.currentGeneration();
+    std::vector<vespalib::string> removes = {};
+    ServiceMappingList updates = { mapping };
+    auto nxt = cur;
+    nxt.add();
+    MapDiff diff{cur, removes, updates, nxt};
+    target.apply(diff);
+}
+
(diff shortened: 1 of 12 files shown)


slobrok/src/tests/service_map_mirror/service_map_mirror_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <vespa/vespalib/util/stringfmt.h>
+#include <map>
+
+using namespace vespalib;
+using namespace slobrok;
+using vespalib::make_string_short::fmt;
+
+using Map = std::map<vespalib::string, vespalib::string>;
+
+Map dump(const ServiceMapMirror &history) {
```
havardpe, line 15: I think 'mirror' is better than 'history' here

slobrok/src/tests/service_map_mirror/service_map_mirror_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+void removeFrom(ServiceMapMirror &target, const ServiceMapping &mapping) {
+    auto cur = target.currentGeneration();
+    std::vector<vespalib::string> removes = { mapping.name };
+    ServiceMappingList updates = { };
+    auto nxt = cur;
+    nxt.add();
+    MapDiff diff{cur, removes, updates, nxt};
+    target.apply(diff);
+}
+
```
havardpe, line 43: consider also testing diffs that contain multiple things, like remove/update of the same name etc.

slobrok/src/vespa/slobrok/server/exchange_manager.cpp
```diff
@@ -53,6 +50,7 @@ ExchangeManager::removePartner(const std::string & name)
     auto oldremote = std::move(_partners[name]);
     LOG_ASSERT(oldremote);
     _partners.erase(name);
+    oldremote->shutdown();
```
havardpe, line 53: I was confused about not having an unsubscribe after this one, but realized that we do not have a consensus map just yet. Consider adding a comment.

slobrok/src/vespa/slobrok/server/remote_slobrok.cpp
```diff
@@ -33,8 +33,7 @@ RemoteSlobrok::RemoteSlobrok(const std::string &name, const std::string &spec,
     _rpcserver.healthCheck();
 }
 
-RemoteSlobrok::~RemoteSlobrok()
-{
+void RemoteSlobrok::shutdown() {
```
havardpe, line 36: (not really a comment, just some random thoughts)

If shutdown is called from any other thread than the network thread; the RPC cleanup does not seem safe. If we remove a peer as a result of re-config, that also need to be done in the network thread. Currently, we pretty much need to use a single thread for everything. We should consider just embracing it and not try to be thread-safe 'some places', since we already have identified possible self-deadlock scenarios with 2-way event listening that would probably be unproblematic if we just avoid locking.

havardpe (approved):

PR #18736, opened 2021-08-13
Use the executor for the part that can be parallell when rebuilding i…

baldersheim:
…ndex on load.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
@@ -229,7 +229,6 @@ public:
                 _cond.wait(guard);
             }
         }
-        while (_pending > 1000) { std::this_thread::yield(); }
         ++_pending;
         _shared_executor.execute(vespalib::makeLambdaTask([this, ref, lid]() {
             auto prepared = _attr._index->prepare_add_document(lid, _attr._denseTensorStore.get_typed_cells(ref), _attr.getGenerationHandler().takeGuard());
@@ -238,7 +237,8 @@ public:
                 _attr._index->complete_add_document(lid, std::move(result));
                 {
                     std::unique_lock guard(_mutex);
-                    if (--_pending == MAX_PENDING/2) {
+                    --_pending;
+                    if ((_pending == MAX_PENDING/2) || (_pending == 0)) {
                         _cond.notify_all();
                     }
                 }
@@ -249,7 +249,10 @@ public:
         }));
     }
     void wait_complete() override {
-        while (_pending > 0) { std::this_thread::sleep_for(1ms); }
+        std::unique_lock guard(_mutex);
+        while (_pending > 0) {
+            _cond.wait(guard);
+        }
     }
 private:
     static constexpr uint32_t MAX_PENDING = 1000;
--- a/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
@@ -12,6 +12,9 @@
 #include <vespa/vespalib/data/slime/inserter.h>
 #include <vespa/vespalib/util/memory_allocator.h>
 #include <vespa/vespalib/util/mmap_file_allocator_factory.h>
+#include <vespa/vespalib/util/threadstackexecutor.h>
+#include <vespa/vespalib/util/lambdatask.h>
+#include <thread>
 
 #include <vespa/log/log.h>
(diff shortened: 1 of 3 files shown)


searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -220,23 +223,45 @@ DenseTensorAttribute::onLoad(vespalib::Executor *executor)
...
             tensorReader.readTensor(raw.data, _denseTensorStore.getBufSize());
             _refVector.push_back(raw.ref);
             if (_index && !use_index_file) {
-                // This ensures that get_vector() (via getTensor()) is able to find the newly added tensor.
-                setCommittedDocIdLimit(lid + 1);
-                _index->add_document(lid);
-                if ((lid % 256) == 0) {
-                    commit();
+                if (executor != nullptr) {
+                    while (pending > 1000) { std::this_thread::yield(); }
```
toregge, line 237: Just yield causes wasted cpu usage on systems with many CPU cores. Consider using a condition variable.

toregge (approved):

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
baldersheim: Refactored and fixed. PTAL

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -204,11 +204,84 @@ DenseTensorAttribute::extract_cells_ref(DocId docId) const
...
+          _pending(0)
+    {}
+    void load(uint32_t lid, vespalib::datastore::EntryRef ref) override {
+        {
+            std::unique_lock guard(_mutex);
+            while (_pending > MAX_PENDING) {
+                _cond.wait(guard);
+            }
+        }
+        while (_pending > 1000) { std::this_thread::yield(); }
```
toregge, line 232: Leftover code that should be removed ?

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -204,11 +204,84 @@ DenseTensorAttribute::extract_cells_ref(DocId docId) const
...
+        ++_pending;
+        _shared_executor.execute(vespalib::makeLambdaTask([this, ref, lid]() {
+            auto prepared = _attr._index->prepare_add_document(lid, _attr._denseTensorStore.get_typed_cells(ref), _attr.getGenerationHandler().takeGuard());
+            _complete_executor.execute(vespalib::makeLambdaTask([this, lid, result=std::move(prepared)]() mutable {
```
toregge, line 236: Note that this allows "out of order" completions. If we want "in order" completions (as during feeding) then the shared executor should populate a promise and the complete executor task should take a future as argument and be scheduled in the foreground (not by the shared executor task).

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
baldersheim: Fixed

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
baldersheim: Out of order is not a problem, as we do not have any ordering guarantees in vespa either. As ANN is in nature approximative that should be even less of an issue.
However I rewrote to use a queue and let the foreground thread do the completion too.
That avoids creating an extra thread and any issues with 2 threads modifying the attribute, even if they did seem to be disjunct.

baldersheim:
@toregge PTAL

searchlib/src/vespa/searchlib/tensor/dense_tensor_attribute.cpp
```diff
@@ -213,57 +213,96 @@ class Loader {
...
-            }));
-        }));
     }
-    void wait_complete() override {
-        std::unique_lock guard(_mutex);
-        while (_pending > 0) {
-            _cond.wait(guard);
+
+    void complete(uint32_t lid, std::unique_ptr<PrepareResult> prepared) {
+        _attr.setCommittedDocIdLimit(std::max(_attr.getCommittedDocIdLimit(), lid + 1));
```
toregge, line 257: This assumes in-order completion. The documents with highest lid might not be visible after load due to a later
complete of a document with a lower lid when using out-of-order completions.

toregge (approved):

PR #18739, opened 2021-08-13
let RemoteSlobrok do its own initial health check

arnej27959:

@havardpe please review

### DIFF
--- a/slobrok/src/vespa/slobrok/server/remote_slobrok.cpp
+++ b/slobrok/src/vespa/slobrok/server/remote_slobrok.cpp
@@ -36,7 +36,7 @@ RemoteSlobrok::RemoteSlobrok(const vespalib::string &name, const vespalib::strin
 }
 
 void RemoteSlobrok::shutdown() {
-    _reconnecter.disable();
+    _reconnecter.Kill();
 
     _pending.clear();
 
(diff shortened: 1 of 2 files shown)


havardpe:
/sd/workspace/src/github.com/vespa-engine/vespa/slobrok/src/vespa/slobrok/server/remote_slobrok.cpp:336: Failed assertion: '_checkServerReq == nullptr'

slobrok/src/vespa/slobrok/server/remote_slobrok.cpp
```diff
@@ -336,42 +328,60 @@ RemoteSlobrok::maybePushMine()
...
-    // when _remAddPeerReq is returned, our managed list is added via doAdd()
 }
 
-void
-RemoteSlobrok::tryConnect()
-{
-    _rpcserver.healthCheck();
+void RemoteSlobrok::handleCheckServerResult() {
+    LOG_ASSERT(_checkServerReq != nullptr);
+    auto & req = *_checkServerReq;
```
havardpe, line 346: consider using guard to subref request automatically on return

havardpe (approved):
seems reasonable

some suggestions:
- shutdown should Kill the reconnecter
- fail might want to assert that there is no outstanding check server rpc request (it was a red herring when trying to figure out how we ended up with a request but no connection)

PR #18751, opened 2021-08-16
Revert "Arnej/wire consensus map"

hmusum:
Reverts vespa-engine/vespa#18743

Seeing core dumps in system tests, not sure of this is related to this PR or https://github.com/vespa-engine/vespa/pull/18744


arnej27959 (approved):
this is probably the right one, missing cleanup produces core dumps on shutdown.

PR #18756, opened 2021-08-16
Arnej/wire consensus 2

arnej27959:

@havardpe please review


slobrok/src/vespa/slobrok/server/sbenv.cpp
```diff
@@ -116,15 +116,23 @@ SBEnv::SBEnv(const ConfigShim &shim)
       _rpcsrvmap()
 {
     srandom(time(nullptr) ^ getpid());
-    _rpcsrvmap.proxy().registerListener(_globalVisibleHistory);
-    _rpcsrvmap.proxy().registerListener(_localRpcMonitorMap);
+    _localMonitorSubscription = MapSubscription::subscribe(_rpcsrvmap.proxy(), _globalVisibleHistory);
+    _consensusSubscription = MapSubscription::subscribe(_rpcsrvmap.proxy(), _localRpcMonitorMap);
+    _globalHistorySubscription = MapSubscription::subscribe(_localRpcMonitorMap.dispatcher(), _consensusMap);
```
havardpe, line 121: have the _globalHistorySubscription and _localMonitorSubscription subscriptions been swapped? (got a bit confused by the names...)

slobrok/src/vespa/slobrok/server/sbenv.cpp
```diff
@@ -116,15 +116,23 @@ SBEnv::SBEnv(const ConfigShim &shim)
...
+    auto pList = _exchanger.getPartnerList();
+    for (const auto & partner : pList) {
+        _exchanger.removePartner(partner);
+    }
+}
 
 SBEnv::~SBEnv()
 {
     getTransport()->WaitFinished();
+    cleanup();
```
havardpe, line 135: is the double cleanup needed? (I would expect that WaitFinished blocks forever in the cases where the other cleanup was not called...)
consider doing cleanup only here, but removing the WaitFinished call.

havardpe (approved):

slobrok/src/vespa/slobrok/server/sbenv.cpp
arnej27959: > have the _globalHistorySubscription and _localMonitorSubscription subscriptions been swapped?

not sure how I managed to swap these, should be fixed now.

slobrok/src/vespa/slobrok/server/sbenv.cpp
arnej27959: we expect the transport thread(s) to be finished already when we get here, yes - can we assert on that?
cleanup() is removed again, it was added to ensure that MapListeners were unregistered before destructors.

PR #18769, opened 2021-08-17
Arnej/add add local api

arnej27959:

@havardpe please review


havardpe (approved):
consider also notifying the inflight operation when a mapping gets removed due to a new conflicting global mapping.

arnej27959:

havardpe (commented):
when there is feedback between the local rpc monitor map and the consensus map, we probably need to be more careful about when we call add/remove/update. Specifically, I think we should erase entries from the map before we generate the event for it.

PR #18770, opened 2021-08-17
Prepare all applications before activating when bootstrapping config server, take 2 [run-systemtest]

hmusum:
Same changes as in https://github.com/vespa-engine/vespa/pull/18720, but add back old code and use feature flag to switch between old and new way of redeploying applications. 

https://github.com/vespa-engine/vespa/pull/18737 will hopefull help with the issues seen when https://github.com/vespa-engine/vespa/pull/18720 was tried.

Please review only.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
@@ -20,10 +20,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -32,6 +31,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import static com.yahoo.vespa.config.server.ConfigServerBootstrap.Mode.BOOTSTRAP_IN_CONSTRUCTOR;
 import static com.yahoo.vespa.config.server.ConfigServerBootstrap.Mode.FOR_TESTING_NO_BOOTSTRAP_OF_APPS;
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
```diff
@@ -100,6 +129,8 @@ private ConfigServerBootstrap(ApplicationRepository applicationRepository, RpcSe
         this.sleepTimeWhenRedeployingFails = Duration.ofSeconds(configserverConfig.sleepTimeWhenRedeployingFails());
         this.exitIfRedeployingApplicationsFails = exitIfRedeployingApplicationsFails;
         rpcServerExecutor = Executors.newSingleThreadExecutor(new DaemonThreadFactory("config server RPC server"));
+        prepareAllAppsBeforeActivating = Flags.PREPARE_ALL_APPS_BEFORE_ACTIVATING_AT_BOOTSTRAP
+                .bindTo(flagSource).with(FetchVector.Dimension.ZONE_ID, applicationRepository.zone().toString());
```
hakonhall, line 133: There is seldom any need to specify the zone dimension as all flags can be overridden per zone.  (The only reason I can think of is in the controller, and if it wants to do things differently per zone.)

hakonhall (commented):
How would prepare-at-bootstrap work when bootstrapping 3 downed config server?  Each of the cfg would wait on the prepare (and activate) barriers, and is that OK before the config servers have completed the bootstrap?  

And in particular one would be faced with cfg1 trying to prepare application A that cfg2 has not prepared yet, and would cfg2 then be able to complete the prepare barrier for A?

hmusum:
Redeplopying applications only happens when there is a version upgrade (version is persisted and read when server is started). So bootstrapping with the version that was last used should work as before.

configserver/src/main/java/com/yahoo/vespa/config/server/ConfigServerBootstrap.java
hmusum: Good point, will remove zone dimension.

hakonhall (approved):
What about the zone bootstrap case?

hakonhall:
In the zone bootstrap case, there are no apps, so that's fine.

hakonhall:
So what if all cfgs are down due to some bug or other, and to get it up we need to upgrade?

baldersheim:
@hmusum and @hakonhall Is this PR still relevant ?

hmusum:
Closing, bootstrap seems to work fine now and there are issues when it takes a long time from prepare to activate (new deployments might happen, then this needs to be retried etc.).

PR #18774, opened 2021-08-17
ensure that sending event to dispatch happens after updating internal…

arnej27959:
… state

@havardpe please review


slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
```diff
@@ -120,14 +121,14 @@ void LocalRpcMonitorMap::notifyFailedRpcSrv(ManagedRpcServer *rpcsrv, std::strin
...
     }
-    if (psd.up) {
-        psd.up = false;
-        _dispatcher.remove(psd.mapping());
-    }
     if (psd.localOnly) {
         auto iter = _map.find(psd.name());
         _map.erase(iter);
     }
+    if (psd.up) {
```
havardpe, line 128: should not assume that we are not deleted by the erase above (even if it probably will be impossible)

arnej27959:
there were more dragons here...

PR #18775, opened 2021-08-17
Arnej/use consensus map 1

arnej27959:

@havardpe please review


havardpe (commented):
stacking both implementations on top of each other may add some extra latency during register calls. Not sure if that is an issue (visitors?)

arnej27959:
wait with this until later

PR #18779, opened 2021-08-17
ensure that sending event to dispatch happens after updating internal…

arnej27959:
… state

* map.erase() invalidates reference to contents, move it to a local variable first
* cannot delete a ManagedRpcServer while we are in its callback

@havardpe please review

this seems much more complicated than I hoped...


havardpe (commented):
looks good

seems the only thing missing is the ability to ignore events from already scheduled to be deleted rpc servers (lookup should probably be allowed to not find anything)

PR #18796, opened 2021-08-18
Since the automatic generation of RankProfiles from global standalone… MERGEOK

baldersheim:
… onnx models does

not ensure globally unique naming, there will be models that will not resolve well in a global namespace.
Hence they must be in separate namespaces (OnnxModels).

@lesters or @hmusum PR


lesters (approved):
LGTM!

PR #18802, opened 2021-08-19
Always account for available IP addresses when checking capacity

mpolden:
This triggered unnecessary node moves in SwitchRebalancer.

@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/HostCapacity.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/HostCapacity.java
@@ -62,7 +62,7 @@ public class HostCapacity {
     }
 
     private int compareWithoutInactive(Node a, Node b) {
-        int result = compare(freeCapacityOf(b, true), freeCapacityOf(a, true));
+        int result = compare(availableCapacityOf(b, true, false), availableCapacityOf(a, true, false));
         if (result != 0) return result;
 
         // If resources are equal we want to assign to the one with the most IP addresses free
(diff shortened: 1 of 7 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/HostCapacity.java
```diff
@@ -99,7 +98,8 @@ public NodeResources freeCapacityOf(Node host) {
 
     NodeResources freeCapacityOf(Node host, boolean excludeInactive) {
         // Only hosts have free capacity
-        if ( ! host.type().canRun(NodeType.tenant)) return new NodeResources(0, 0, 0, 0);
+        if ( ! host.type().canRun(NodeType.tenant)) return NodeResources.zero();
+        if (   freeIps(host) == 0) return NodeResources.zero();
```
bratseth, line 102: This looks a bit like a latent bug to me. 
I suggest you rename this method to availableCapacityOf(...) and document that is returns capacity that is both free and usable.

One usage (through HostCapacity.freeCapacityOf(node)) is computing skew, where I'm not that the consequences of using this are quite what we want. Perhaps we need to keep freeCapacityOf with the old behavior in addition?

Btw, the reason we model ip capacity in this half-assed way is that we're not supposed to run out of them. Could we get back to that state?

mpolden: > I suggest you rename this method to availableCapacityOf(...) and document that is returns capacity that is both free and usable.

Agreed.

> Perhaps we need to keep freeCapacityOf with the old behavior in addition?

I considered it, but I can take another look. Maybe if we make it clear that the method with the old behavior does not consider IPs.

> Btw, the reason we model ip capacity in this half-assed way is that we're not supposed to run out of them. Could we get back to that state?

Yes, ideally each host would have enough IPs to satisfy the maximum number of nodes it can fit. For on-premise the answer is still no because of a limitation in the rack switches.

bratseth: Yes, it's just the skew stuff I'm not sure of.

mpolden:
PTAL

bratseth (approved):
Great!

PR #18812, opened 2021-08-20
Bratseth/cli 2

bratseth:


client/go/src/cmd/query.go
```diff
@@ -29,28 +29,36 @@ var queryCmd = &cobra.Command{
...
+            return
+        }
+        if i > 0 {
+            argBuilder.WriteString("&")
+        }
+        argBuilder.WriteString(argument)
     }
 
-    path := "/search/?" + argument
+    path := "/search/?" + argBuilder.String()
```
mpolden, line 56: Consider constructing a `net.URL` from the arguments instead, so that the end result becomes a properly encoded URL.

client/go/src/cmd/deploy.go
```diff
@@ -21,28 +21,59 @@ import (
...
+            deploy(false, "")
+        } else {
+            deploy(false, args[0])
+        }
+    },
+}
+
+func deploy(prepare bool, application string) {
+    // TODO: Support no application (activate)
     if ! strings.HasSuffix(application, ".zip") {
```
mpolden, line 77: `filepath.Ext` is better suited for this.

client/go/src/cmd/query.go
bratseth: Better - thanks

PR #18816, opened 2021-08-21
Revert "fix variable name typo"

aressem:
Reverts vespa-engine/vespa#18815

Will revert https://github.com/vespa-engine/vespa/pull/18808 after this to go back to old behaviour. Need to debug why the services are not starting with this change.


bratseth:
👍

PR #18817, opened 2021-08-21
Revert "Reapply "Fail Vespa startup if no numactl .""

aressem:
Reverts vespa-engine/vespa#18808

Go back to old behaviour. Need to debug why the services are not starting with this change.


bratseth:
👍

PR #18818, opened 2021-08-22
Try to run steps again if we bail out due to status change

jonmv:
If someone completes a step we are trying to run, we will bail out. Meanwhile, that other thread will attempt the next step, but fails to acquire the lock, since we hold it (for the prerequisite), so that thread gives up. When it is our turn again, and we bail out, we should try again to run any ready steps, otherwise we may lose the fact that _someone_ ought to do _something_. This is absolutely not a practical issue, but makes one unit test slightly unstable :)

@bjorncs please review.


bjorncs (approved):

jonmv:
Argh, broke some rest-api-teest ...

jonmv:
Will fix.

jonmv:
https://github.com/vespa-engine/vespa/pull/18830

PR #18836, opened 2021-08-23
Bratseth/cli 3

bratseth:


client/go/src/utils/print.go
```diff
@@ -5,6 +5,7 @@
 package utils
```
mpolden, line 5: Nit: Package names should be singular.

Also, as I explained in https://github.com/vespa-engine/vespa/pull/18354#discussion_r692470370 - `src/` subdir is not necessary.

client/go/src/utils/print.go
```diff
@@ -37,6 +38,18 @@ func Detail(messages ...string) {
     print("\033[33m", messages)
 }
 
+// Prints all the text of the given reader
+func PrintReader(reader io.ReadCloser) {
```
mpolden, line 42: Nit: Change to `io.Reader` to signal that this will never close the reader.

client/go/src/cmd/deploy.go
```diff
@@ -121,17 +121,19 @@ func deploy(prepare bool, application string) {
     }
     serviceDescription := "Deploy service"
     response := utils.HttpDo(request, time.Minute * 10, serviceDescription)
-    defer response.Body.Close()
     if (response == nil) {
```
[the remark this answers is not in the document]
bratseth: Well in this case it makes sense to handle them in the layer below, pulling it into each usage would lead to code duplication.

client/go/src/cmd/deploy_test.go
```diff
@@ -53,4 +61,18 @@ func assertDeployRequestMade(target string, client *mockHttpClient, t *testing.T
     buf := make([]byte, 7) // Just check the first few bytes
     body.Read(buf)
     assert.Equal(t, "PK\x03\x04\x14\x00\b", string(buf))
-}
\ No newline at end of file
+}
+
+func assertApplicationPackageError(t *testing.T, status int, errorMessage string) {
+    client := &mockHttpClient{ nextStatus: status, nextBody: errorMessage, }
+	assert.Equal(t,
```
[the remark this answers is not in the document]
bratseth: okk; fine. But in a separate PR.

client/go/src/utils/print.go
bratseth: Done, but it's not very nice ...

client/go/src/utils/print.go
bratseth: better; thanks

client/go/src/cmd/deploy.go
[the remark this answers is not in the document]
mpolden: I would make `HttpDo` return `(*http.Response, error)` and then add context to the returned error, e.g. `fmt.Errorf("Could not connect to %s at %s: %w", strings.ToLower(description), request.URL.Host, err)`.

With the above change you can do
```
if (err != nil) {
  util.Error(err)
  return
}
```
This wouldn't add much duplication and makes it clear we're handling an error.

client/go/build.sh
```diff
@@ -1,7 +1,14 @@
 # Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 # Execute from this directory to build the command-line client to bin/vespa
-export GOPATH=`pwd`
-cd "$GOPATH/src/"
-go test ./...
+cd util
```
mpolden, line 4: Just do `go test ./...` to run all tests recursively.

bratseth: zsh$ go test ./...
pattern ./...: directory pkg/mod/github.com/davecgh/go-spew@v1.1.1/spew outside available modules

mpolden: This is likely because you have `pkg` in current directory as a result of using `GOPATH` in the past, and mixing it with the project path. Try remove the `pkg`, make sure GOPATH isn't set and run go test again.

mpolden:
I didn't notice it before, but `go.sum` shouldn't be ignored. It contains hashes used for verifying the contents of the dependency tree, to ensure that any third-party hasn't been tampered with. https://golang.org/ref/mod#go-sum-files

client/go/build.sh
bratseth: A+ support, that worked.

mpolden (approved):

PR #18845, opened 2021-08-24
GC unused filereferences.def

baldersheim:
@hmusum PR


hmusum:
I think the .def file is needed until all config models using it has been removed

baldersheim:
hmm, that can not be. They must be in the fat per version bundle. If not changes would be even more cumbersome.
But we might hold until 7.455 is the last.

PR #18858, opened 2021-08-25
Application source heuristics

bratseth:


mpolden (approved):
I'm doing something similar to this as I need it for `vespa cert` as well, but this is good for now.

PR #18866, opened 2021-08-26
Look for pom when resolving application package

bratseth:
Check for existence of a pom when deciding whether or not to use target or source.
I reverted to the previous form of the function to make this work (the thing with candidates wasn't quite right anyway, because some were not candidates but evidence of where to look).

Could we please try to avoid encoding test cases as data you loop over as in deploy_test? It makes it harder to understand what is failing. For example, which test case fails here:
```
--- FAIL: TestFindApplicationPackage (0.00s)
    deploy_test.go:36: #3: FindApplicationPackage("/var/folders/97/srk32nhn027bscz9v43sxbrc0000gn/T/TestFindApplicationPackage2668587650/001") = ("/var/folders/97/srk32nhn027bscz9v43sxbrc0000gn/T/TestFindApplicationPackage2668587650/001/src/main/application", %!s(<nil>)), want ("/var/folders/97/srk32nhn027bscz9v43sxbrc0000gn/T/TestFindApplicationPackage2668587650/001/target/application.zip", nil)
```

I know you do it because it is a pattern in Go, but that pattern is just mistaken ... You can make it even shorter while avoiding the drawback of making test cases illegible by creating your own custom assert function.

### DIFF
--- a/client/go/cmd/deploy_test.go
+++ b/client/go/cmd/deploy_test.go
@@ -62,7 +62,7 @@ func TestDeployApplicationDirectoryWithTarget(t *testing.T) {
 func TestDeployApplicationDirectoryWithEmptyTarget(t *testing.T) {
 	client := &mockHttpClient{}
 	assert.Equal(t,
-		"\x1b[31mno application package found in testdata/applications/withEmptyTarget\x1b[0m\n",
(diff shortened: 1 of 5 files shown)


mpolden (approved):
> Could we please try to avoid encoding test cases as data you loop over as in deploy_test? It makes it harder to understand what is failing. For example, which test case fails here:

Sure, but it says which case it is though: `#3`.

PR #18871, opened 2021-08-26
Improve messages and colorization

mpolden:
- Colorization only happens when output is TTY
- Colorization should work on Windows (didn't test it though)
- Improved use of colors
- Isolates message printing to cmd package

@bratseth


bratseth (approved):
Great

PR #18877, opened 2021-08-26
Use self-subscription for slobroks config in container messagebus

jonmv:
@baldersheim please review. This reverts the move from subscription to fixed config for the RPCNetwork used by the container messagebus.


baldersheim (approved):
You also have at least the DocumentAccessProvider

baldersheim (changes requested):
Build failures must be resolved.

baldersheim (approved):

PR #18879, opened 2021-08-27
vespa config get/set

mpolden:
Basic support for persisting config to file. Only supports setting `target` for now.

@bratseth


bratseth (approved):
Great!

PR #18880, opened 2021-08-27
use getgrouplist and also change supplementary groups

arnej27959:

@toregge please review

### DIFF
--- a/configd/src/apps/su/main.cpp
+++ b/configd/src/apps/su/main.cpp
@@ -34,8 +34,12 @@ int main(int argc, char** argv)
     int mac_gid = g;
     int mac_groups[256];
     int ggl = getgrouplist(username, mac_gid, mac_groups, &group_arr_sz);
-    for (int i = 0; i < group_arr_sz; ++i) {
-        grouplist[i] = (gid_t) mac_groups[i];
+    if (ggl < 0) {
+        group_arr_sz = 0;
+    } else {
+        for (int i = 0; i < group_arr_sz; ++i) {
+            grouplist[i] = (gid_t) mac_groups[i];
+        }
     }
 #else
     int ggl = getgrouplist(username, g, grouplist, &group_arr_sz);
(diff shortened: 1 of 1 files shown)


configd/src/apps/su/main.cpp
```diff
@@ -28,6 +28,10 @@ int main(int argc, char** argv)
     gid_t g = p->pw_gid;
     uid_t u = p->pw_uid;
 
+    gid_t grouplist[256];
+    int group_arr_sz = 256;
+    int ggl = getgrouplist(username, g, grouplist, &group_arr_sz);
```
toregge, line 33: This might trigger a warning when compiling on darwin (gid_t is not int, thus we have a pointer type mismatch).

configd/src/apps/su/main.cpp
```diff
@@ -36,7 +40,11 @@ int main(int argc, char** argv)
         return 1;
     }
     size_t listsize = 1;
-    gid_t grouplist[1] = { g };
+    if (ggl > 0) {
```
toregge, line 43: You might want to use`ggl  >= 0` here, since getgrouplist returns 0 when successful on darwin.

toregge (approved):

arnej27959:
tried rewriting for Darwin but I haven't tested that myself, please take a look.

configd/src/apps/su/main.cpp
```diff
@@ -28,6 +28,19 @@ int main(int argc, char** argv)
     gid_t g = p->pw_gid;
     uid_t u = p->pw_uid;
 
+    gid_t grouplist[256];
+    int group_arr_sz = 256;
+#ifdef __APPLE__
+    int mac_gid = g;
+    int mac_groups[256];
+    int ggl = getgrouplist(username, mac_gid, mac_groups, &group_arr_sz);
+    for (int i = 0; i < group_arr_sz; ++i) {
```
toregge, line 37: This loop should only be called if `ggl == 0` since `getgrouplist` returns 0 on success and -1 on failure on darwin.
`mac_groups` should only be read If `getgrouplist` succeeded.

PR #18883, opened 2021-08-27
Reinstate correct default CC flavour for older config models

jonmv:
@baldersheim please review.

This will cause all older models, which still expect a feature flag, to use the smaller flavour.


baldersheim (approved):
But this should not roll out until most apps have gotten new version and then chosen correct flavour already.

PR #18884, opened 2021-08-27
Add test of top-level distributor functionality

vekterli:
@geirst please review. This is the first batch of adapted tests for the top level distributor code

This is a subset of the legacy distributor tests, adapted to
explicitly test cross-stripe functionality. Once all relevant
tests have been ported to be cross-stripe, the legacy test
code will be removed.


geirst (approved):
👍

PR #18888, opened 2021-08-29
Expect ids on the standard Vespa format

bratseth:


client/go/vespa/id.go
```diff
@@ -0,0 +1,46 @@
...
+// Author: bratseth
+
+package vespa
+
+import (
+	"errors"
+	"strings"
+)
+
+func IdToUrlPath(documentId string) (string, error) {
```
mpolden, line 12: Nit: The Go convention is to use normal capitalization for abbreviations (e.g. `func IDToURLPath(documentID string)`), unlike the Java convention.

mpolden (approved):

PR #18891, opened 2021-08-30
Refactor deploy command to support cloud

mpolden:
Was hoping to fully implement it, but signing requests for Vespa Cloud is not
trivial and my diff was large enough already.

Will do the request signing next.

@bratseth


bratseth (approved):
👍

PR #18893, opened 2021-08-30
As doSeek is called alot more frequent than doUnpack just use locking…

baldersheim:
… of the heap in unpack.

In addition to adjusting the priority Q also update the distance_threshold with a relaxed store to an atomic variable.
On read the distance threshold can be read cheaply with a relaxed load.

@geirst PR
@jobergum FYI


geirst (approved):
👍

geirst:
@arnej27959 FYI

PR #18898, opened 2021-08-30
The global filter is only needed when having a nearest neighbor index…

geirst:
… (hnsw) and doing approximate calculation.

This avoids costly calculation of the global filter in cases it is not needed.

@baldersheim and @arnej27959 please review


baldersheim (approved):

arnej27959 (approved):
👍

PR #18902, opened 2021-08-30
Tweak colors and messages

bratseth:
- Keep the red text short
- Mostly always indicate success or error by green or red
- Use brown for technical detail you don't need to look at most of the time

### DIFF
--- a/client/go/cmd/clone.go
+++ b/client/go/cmd/clone.go
@@ -54,17 +54,10 @@ func cloneApplication(source string, name string) {
 		defer os.Remove(zipFile.Name())
 	}
 
-	createErr := os.Mkdir(name, 0755)
-	if createErr != nil {
-		log.Print("Could not create directory '", color.Cyan(name), "'")
-		log.Print(createErr)
-		return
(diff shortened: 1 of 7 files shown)


mpolden (approved):
👍 

This is starting to look a lot like log levels though.

bratseth:
Well it's not a log but a user interface so we need finer control over details of the rendering.
Anyway, the things I did here would be simpler in the API we replaced by this one as it had functions for emitting errors, detail info etc.

mpolden:
Sure, we should certainly add helper functions for the most obvious patterns. I was just against doing all output printing through our own "log" API, because that forced us to replicate most of stdlib log to cover all formatting use cases.

Example: I'll add something like this in my next PR, to provide some hints if the user e.g. forgets to run `vespa cert`:
```
func errorWithHint(err error, hints ...string) {
	log.Print(color.Red("Error: "), err)
	for _, hint := range hints {
		log.Print(color.Cyan("Hint: "), hint)
	}
}
```

PR #18903, opened 2021-08-30
Avoid copying data just to compress them when it is not necessary.

baldersheim:
@bratseth or @havardpe PR


baldersheim:
Or @hmusum PR

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/rpc/SlimeClusterStateBundleCodec.java
```diff
@@ -48,7 +48,7 @@ public EncodedClusterStateBundle encode(ClusterStateBundle stateBundle) {
         }
 
         byte[] serialized = BinaryFormat.encode(slime);
```
hmusum, line 50:
```suggestion
```

hmusum: Unused, can be removed

hmusum (approved):
LGTM

PR #18904, opened 2021-08-30
Prepare for alternate visited nodes trackers for HNSW index.

toregge:
@geirst : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
+++ b/searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
@@ -8,6 +8,10 @@ namespace search::tensor {
 
 class HnswIndex;
 
+/*
+ * Tracker for visited nodes based on search::AllocatedBitVector. Best when
+ * many nodes are visited.
+ */
 class BitVectorVisitedTracker
 {
     search::AllocatedBitVector _visited;
--- a/searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
+++ b/searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
@@ -1,0 +1,28 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#pragma once
+
+#include <vespa/searchlib/common/allocatedbitvector.h>
+
+namespace search::tensor {
+
+class HnswIndex;
+
+class BitVectorVisitedTracker
+{
+    search::AllocatedBitVector _visited;
+public:
+    BitVectorVisitedTracker(const HnswIndex&, uint32_t doc_id_limit, uint32_t);
+    ~BitVectorVisitedTracker();
+    void mark(uint32_t doc_id) { _visited.setBit(doc_id); }
+    bool try_mark(uint32_t doc_id) {
+        if (_visited.testBit(doc_id)) {
+            return false;
+        } else {
+            _visited.setBit(doc_id);
+            return true;
+        }
+    }
+};
+
+}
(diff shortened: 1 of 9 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
```diff
@@ -206,6 +211,29 @@ HnswIndex::calc_distance(const TypedCells& lhs, uint32_t rhs_docid) const
     return _distance_func->calc(lhs, rhs);
 }
 
+uint32_t
+HnswIndex::estimate_visited_nodes(uint32_t level, uint32_t doc_id_limit, uint32_t neighbors_to_find, const search::BitVector* filter) const
+{
+    uint32_t m_for_level = max_links_for_level(level);
+    uint64_t base_estimate = uint64_t(m_for_level) * neighbors_to_find + 100;
```
baldersheim, line 218: Since you are multiplying with neighbors_to_find, will it not be better to use a number closer to average rather than max ?

baldersheim (approved):

searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include <vespa/searchlib/common/allocatedbitvector.h>
+
+namespace search::tensor {
+
+class HnswIndex;
+
+class BitVectorVisitedTracker
```
geirst, line 11: Please add class comment.

searchlib/src/vespa/searchlib/tensor/hash_set_visited_tracker.h
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+#include <vespa/vespalib/stllike/hash_set.h>
+
+namespace search::tensor {
+
+class HnswIndex;
+
+class HashSetVisitedTracker
```
geirst, line 11: Please add class comment.

searchlib/src/vespa/searchlib/tensor/reusable_set_visited_tracker.h
```diff
@@ -0,0 +1,28 @@
...
+
+#pragma once
+
+#include <vespa/vespalib/util/reusable_set_handle.h>
+
+namespace search::tensor {
+
+class HnswIndex;
+
+class ReusableSetVisitedTracker
```
geirst, line 11: Please add class comment.

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
geirst: _max_links_for_level()_ returns the configured max for that level. We could divide this by two to get an average (assuming the number of links is evenly distributed at that level).

geirst (approved):
👍

searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
baldersheim: Then I suggest we just keep it as is. Only if we maintained an average, or a number that reflected real world.

searchlib/src/vespa/searchlib/tensor/bitvector_visited_tracker.h
toregge: Done.

searchlib/src/vespa/searchlib/tensor/hash_set_visited_tracker.h
toregge: Done.

searchlib/src/vespa/searchlib/tensor/reusable_set_visited_tracker.h
toregge: Done.

baldersheim (approved):

geirst (approved):
👍

PR #18905, opened 2021-08-30
Revert "Avoid copying data just to compress them when it is not necessary."

baldersheim:
Reverts vespa-engine/vespa#18903
@hmusum PR
build broke


hmusum:
👍

PR #18906, opened 2021-08-30
Revert "Revert "Avoid copying data just to compress them when it is not necessary.""

baldersheim:
Reverts vespa-engine/vespa#18905


baldersheim:
@hmusum PR

hmusum (approved):

PR #18907, opened 2021-08-30
File distribution cleanup [run-systemtest]

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDistributionFactory.java
```diff
@@ -17,15 +18,26 @@
 public class FileDistributionFactory {
 
     protected final ConfigserverConfig configserverConfig;
-    private final Supervisor supervisor = new Supervisor(new Transport("filedistribution"));
```
baldersheim, line 20: Any reason why the supervisor/transport was moved ?
now you will have many instead of 1.
Independently it should have been closed.

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDistributionImpl.java
```diff
@@ -10,26 +9,26 @@
...
 /**
  * @author baldersheim
  */
 public class FileDistributionImpl implements FileDistribution, RequestWaiter {
+
     private final static Logger log = Logger.getLogger(FileDistributionImpl.class.getName());
     private final static double rpcTimeout = 1.0;
 
-    private final Supervisor supervisor;
+    private final Supervisor supervisor = new Supervisor(new Transport("filedistribution"));
```
baldersheim, line 27: You need to close/stop the supervisor too, when you create it. It contains threads.

baldersheim (changes requested):
It looks good apart from the ownership of the supervisor/transport used by FileDistributionImpl.

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDistributionFactory.java
hmusum: No, this should have been kept, will fix

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDistributionImpl.java
hmusum: FileDistributionFactory is a singleton and lives for as long as the process is running, I guess that's why it is not closed.

baldersheim (approved):

PR #18919, opened 2021-08-31
Using resource value that is compared to the resource limit

bjormel:


bjormel:
@bratseth, please disregard if this is not inline with the assumptions in the AutoScaler implementation

bratseth:
Yes, it's not. Autoscaling has these hardcoded targets:
    static final double idealMemoryLoad = 0.65;
    static final double idealDiskLoad = 0.6;

PR #18927, opened 2021-09-01
Bratseth/document get

bratseth:


client/go/util/operationResult.go
```diff
@@ -0,0 +1,32 @@
+// Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// A struct containing the result of an operation
+// Author: bratseth
+
+package util
```
mpolden, line 5: Filenames should use snake case: operationResult.go -> operation_result.go.

client/go/vespa/document.go
```diff
@@ -0,0 +1,101 @@
...
+	} else {
+		return util.FailureWithPayload(serviceDescription+" at "+request.URL.Host+": "+response.Status, util.ReaderToJSON(response.Body))
+	}
+}
+
+func Post(documentId string, jsonFile string, target string) *util.OperationResult {
+	header := http.Header{}
+	header.Add("Content-Type", "application/json")
+
+	fileReader, fileError := os.Open(jsonFile)
```
mpolden, line 54: This should be closed: `defer fileReader.Close()` after checking the error.

client/go/util/operationResult.go
```diff
@@ -0,0 +1,32 @@
...
+
+type OperationResult struct {
+	Success bool
+	Message string // Mandatory message
+	Detail  string // Optional detail message
+	Payload string // Optional payload - may be present whether or not the operation was success
+}
+
+func Success(message string) *OperationResult {
```
mpolden, line 14: I think it's better to return a copy here (and the other constructors), seeing as this a light-weight struct.

client/go/util/operationResult.go
bratseth: I couldn't find any consensus on this but ok.

client/go/vespa/document.go
bratseth: Thanks!

PR #18935, opened 2021-09-01
Ensure that only one thread will compute the missing object in the ca…

baldersheim:
…che. The others will wait.

@hmusum PR


hmusum (approved):
Nice

PR #18936, opened 2021-09-01
Improve documentation

bratseth:
First half of fixing documentation

### DIFF
--- a/client/go/cmd/config.go
+++ b/client/go/cmd/config.go
@@ -41,7 +41,7 @@ var configCmd = &cobra.Command{
 }
 
 var setConfigCmd = &cobra.Command{
-	Use:     "set option value",
+	Use:     "set <option> <value>",
 	Short:   "Set a configuration option.",
 	Example: "$ vespa config set target cloud",
 	Args:    cobra.ExactArgs(2),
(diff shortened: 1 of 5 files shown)


client/go/cmd/config.go
```diff
@@ -55,7 +55,7 @@ var setConfigCmd = &cobra.Command{
 }
 
 var getConfigCmd = &cobra.Command{
-	Use:     "get option",
+	Use:     "get <option>",
```
mpolden, line 58: Should be `<[option]>` as omitting it prints all options.

client/go/cmd/deploy.go
```diff
@@ -24,13 +24,13 @@ var (
 
 func init() {
 	rootCmd.AddCommand(deployCmd)
-	rootCmd.AddCommand(prepareCmd)
-	rootCmd.AddCommand(activateCmd)
+	// rootCmd.AddCommand(prepareCmd)
+	// rootCmd.AddCommand(activateCmd)
```
mpolden, line 28: Implemented in #18937.

bratseth: Right, fixed

client/go/cmd/config.go
bratseth: done

mpolden:
👍

PR #18937, opened 2021-09-01
vespa prepare/activate

mpolden:
@bratseth


bratseth (approved):
Great!

Although I don't know how yellow will look with my day mode terminal ... :-)

mpolden:
> Although I don't know how yellow will look with my day mode terminal

It's still the same color, they just changed the name.

PR #18938, opened 2021-09-01
Add short form output option to model-evaluation REST API

lesters:
@bratseth Please review. Adds option to output short format for indexed tensors to the model-evaluation REST API when the parameter `?format=short` is given. The name (and possibly output format) of this parameter should probably be discussed during an architect meeting.

As an example, the output for `tensor(x[3]):[1.0, 2.0, 3.0]` is

```
{
    "type":"tensor(x[3])",
    "value": [1.0, 2.0, 3.0]
}
```


bratseth (approved):
Great! Before we publish this I think we should also implement the other short forms we support on input - for mixed tensors etc., so that we don't end up with even more variants later.

PR #18939, opened 2021-09-01
The metrics proxy is very memory hungry.

baldersheim:
This will avoid creating many identical DimensionIds, ServiceIds.
It will also prefer using cheap singleton lists where possible.
In addition the code relies less on generated Config code.

@hmusum PR


baldersheim:
@gjoranv PR

hmusum (approved):
LGTM

gjoranv:
Great!

PR #18953, opened 2021-09-02
Expose rotation status in application v4 API

oyving:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -1350,12 +1350,13 @@ public class ApplicationApiHandler extends AuditLoggingRequestHandler {
               .ifPresent(version -> toSlime(version, object.setObject("revision")));
     }
 
-    private void toSlime(Endpoint endpoint, Cursor object) {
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1376,16 +1377,18 @@ private void toSlime(Cursor response, DeploymentId deploymentId, Deployment depl
             zoneEndpoints = zoneEndpoints.not().legacy();
         }
         for (var endpoint : controller.routing().directEndpoints(zoneEndpoints, deploymentId.applicationId())) {
-            toSlime(endpoint, endpointArray.addObject());
+            toSlime(endpoint, Optional.empty(), endpointArray.addObject());
         }
         // Add global endpoints
         EndpointList globalEndpoints = controller.routing().endpointsOf(application, deploymentId.applicationId().instance())
                                                  .targets(deploymentId.zoneId());
+        Map<Endpoint, EndpointStatus> globalEndpointsStatus = controller.routing().globalRotationStatus(deploymentId);
```
mpolden, line 1385: This will only give you global endpoint status of endpoints backed by rotations.

`/routing/v1/` supports inspecting global status, independent of the global routing method used, so I think that logic has to be refactored/moved to `RoutingController` to allow it to be used here.

However, it's not ideal that every request to this API may end up passing a request to the config server for each deployment having a global endpoint backed by a rotatation. So maybe it's best to just use `/routing/v1/` for this feature...

PR #18955, opened 2021-09-02
Add controller API to request service dump

bjorncs:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2048,6 +2052,71 @@ private HttpResponse testConfig(ApplicationId id, JobType type) {
...
+        Node node;
+        try {
+            node = nodeRepository.getNode(zone, hostname);
+        } catch (IllegalArgumentException e) {
+            throw new NotExistsException(new Hostname(hostname));
+        }
+        ApplicationId app = ApplicationId.from(tenant, application, instance);
+        ApplicationId owner = node.owner().orElseThrow(() -> new IllegalArgumentException("Node has no owner"));
+        if (!app.equals(owner)) {
+            throw new IllegalArgumentException("Node is owned by " + owner.toFullString());
```
freva, line 2113: Probably don't want to leak this info? If the node doesn't exist, doesn't have an owner, or has a different owner should return same generic error.

oyving (approved):

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
bjorncs: Agree

PR #18957, opened 2021-09-02
Prefer 'Host' header when constructing uri builder

bjorncs:

### DIFF
--- a/container-core/src/main/java/com/yahoo/restapi/RestApi.java
+++ b/container-core/src/main/java/com/yahoo/restapi/RestApi.java
@@ -128,6 +128,11 @@ public interface RestApi {
         Optional<RequestContent> requestContent();
         RequestContent requestContentOrThrow();
         ObjectMapper jacksonJsonMapper();
+        /**
+         * Creates a URI builder pre-initialized with scheme, host and port.
+         * Intended for response generation (e.g for interactive REST APIs).
+         * DO NOT USE FOR CUSTOM ROUTING.
+         */
         UriBuilder uriBuilder();
         AclMapping.Action aclAction();
         Optional<Principal> userPrincipal();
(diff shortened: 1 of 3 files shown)


jonmv:
This feels iffy. It indicates we've somehow set the wrong URI when creating the request, after receiving it from Jetty. I suspect that's not the case, but that this rather has to do with test code? Can't we fix that code instead, if this is the case?

jonmv:
I mean, the URI we have set for the request should be based on the Host header. The URI we've created has been validated, and we ignore that by jumping back to the Host header here. This is possibly OK right now, but ...

bjorncs:
The URI contains the local listen port (for uri matching purposes), while the uri builder is intended for constructing URIs that are part of the response (like interactive rest APIs). The uri builder is not used for routing, in that case I would agree with you.

jonmv:
Hmm ... so the host header will still be wrong when the request is served from behind a proxy? We have several of these responses where we rewrite the response across each proxy. This is horrible. Perhaps a dedicated header with a response-host-port-override could solve this in a much better way? That would also be suitable for this ...

jonmv (approved):
Approved, but let's consider a more thorough solution?

jonmv:
Perhaps make it more clear that this is the intended use, btw..

bjorncs:
We have multiple layer 4 load balancers where port exposed from load balancer and local listen port differs. E.g controllers in public.

jonmv:
> We have multiple layer 4 load balancers where port exposed from load balancer and local listen port differs. E.g controllers in public.

Yes, which warrants a dedicated override for creating responses, something like:
1. if `X-Original-Host-Port` header is set, use its value; or
2. if `Host` header is set, use its value; or
3. use value from request. 

If used throughout, this would let us kill all response rewriting.

jonmv:
(Then we could set the `X`-header when proxying.

bjorncs:
We can add support for https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host.

jonmv:
That's a perfect fit!

jonmv (approved):
Nicee

PR #18959, opened 2021-09-02
Add status for manual deployments in application v4

oyving:
@jonmv - please check that what I'm doing does what I think it does 😄 
@ldalves - this is what you need for the new page design, right?


jonmv (approved):
I think it does what I think you think it does ... XD

ldalves (approved):

Issue #18909, opened 2021-08-31
Empty dense tensor compact representation [] is expanded to 0-valued tensor


jobergum:
Consider a field like this:

```
 field embedding type tensor<float>(x[512]) {
      indexing: attribute|index 
      attribute {
        distance-metric:euclidean
      }
      index {
      hnsw {
      max-links-per-node: 16
      neighbors-to-explore-at-insert: 500
     }
}
```
If you feed an empty [] compact tensor representation like the below

```
{
    "id": "id:n:image::99999999",
    "fields": {
        "id": 99999999,
        "embedding": {
            "values": []
        }
    }
}
```
It gets expanded to a tensor with 512 cells with value 0.0 

```
fields: {
  sddocname: "image",
  id: 99999999,
  embedding: {
   cells: [
   {
   address: {
    x: "0" 
   },
  value: 0
   },
  {
  address: {
    x: "1"
  },
 value: 0

``` 

The above could create many duplicate vectors which the HNSW algorithm can struggle with (bad insert performance)

jobergum:
Disallow/Reject empty [] tensor was the suggestion

lesters:
The above PR should reject empty tensors during feed.

PR #18961, opened 2021-09-02
Disallow feeding empty indexed tensors

lesters:
@bratseth Please review. Ref: https://github.com/vespa-engine/vespa/issues/18909

I guess this can break existing feeds currently doing this?


bratseth (approved):
We'll see, but I think it's more likely that if people do this they have a bug.

PR #18962, opened 2021-09-02
Make vespa status work for cloud target

mpolden:
(sorry for the large diff)

Had to refactor the target stuff to support service discovery and waiting.

- Making `vespa document/query` work for cloud should be easy now.
- I didn't add `deploy --wait` yet, but that should be easy too.

@bratseth


bratseth (approved):
Great!

PR #18963, opened 2021-09-02
Port remaining legacy distributor tests to top-level test suite

vekterli:
@geirst please review

Also fix a minor regression caused by the stripe cluster state change
code path consulting a now unused part of the `StripeBucketDBUpdater`
on whether a cluster state change implies bucket ownership change.
This was used for adding a configurable safe period for client mutations
to ensure distributors can't step on each others toes. The responsibility
for telling stripes that state changes imply ownership changes has
now been moved to the top-level DB updater and happens through the
stripe guard interface instead.


geirst (approved):
👍

PR #18964, opened 2021-09-02
- Add test for large expressions and inheritance.

baldersheim:
- Allow multiple definition as long as they are identical.

@hmusum PR


bratseth (approved):
It's not, like, the most impressively large one I've ever seen, but ok I guess.

baldersheim:
As with many things, small or large, thick or thin, good or bad, ugly or not, it is all in the eye of the beholder.

PR #18965, opened 2021-09-02
Bratseth/tweaks

bratseth:
Minor improvement in output + commenting out prepare+activate:
Prepare doesn't work, and I can't figure out why. I suggest we shelve it for now and focus on what's needed to get the MVP out the door, it's just so much better than the old way already.

### DIFF
--- a/client/go/cmd/deploy.go
+++ b/client/go/cmd/deploy.go
@@ -26,8 +26,8 @@ var (
 
 func init() {
 	rootCmd.AddCommand(deployCmd)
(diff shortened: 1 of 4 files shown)


mpolden (approved):
> Prepare doesn't work, and I can't figure out why. 

Strange, it worked yesterday. Maybe I broke it...

bratseth:
It was broken for me before the latest PR from today as well.
Did you test with a broken application package? That just returns 200 OK, so I don't think the files are really added to the session.

PR #18971, opened 2021-09-05
Add a 5x faster handcoded detection of legal feature names that does …

baldersheim:
…not require quoting.

@bratseth or @lesters PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/FeatureNames.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/FeatureNames.java
@@ -4,7 +4,6 @@ package com.yahoo.searchdefinition;
 import com.yahoo.searchlib.rankingexpression.Reference;
 
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 /**
  * Utility methods for query, document and constant rank feature names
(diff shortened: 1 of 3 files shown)


config-model/src/main/java/com/yahoo/searchdefinition/FeatureNames.java
```diff
@@ -56,10 +52,28 @@ public static boolean isConstantFeature(Reference reference) {
...
+        }
+        return true;
+    }
+    private static boolean isValidFirst(char c) {
+        // [A-Za-z0-9_]
+        return (c == '_') || ((c >= 'a') && (c <= 'z')) || ((c >= 'A') && (c <= 'Z')) || ((c >= '0') && (c <= '9'));
+    }
+    private static boolean isValidAny(char c) {
+        // [A-Za-z0-9_-]*
+        return (c == '_') || (c == '-') || ((c >= 'a') && (c <= 'z')) || ((c >= 'A') && (c <= 'Z')) || ((c >= '0') && (c <= '9'));
```
bratseth, line 76:
```suggestion
        return c == '-' || isValidFirst(c);
```

bratseth (approved):
👍

PR #18973, opened 2021-09-06
Naive deploy --wait

mpolden:
Currently does not check if the deployment you just did has completed, so it's
mostly useful for waiting for a first-time deployment.

@bratseth


bratseth (approved):
👍

PR #18977, opened 2021-09-06
Improve some text

bratseth:


client/go/cmd/api_key.go
```diff
@@ -79,6 +79,7 @@ func printPublicKey(apiKeyFile, tenant string) {
 	}
 	log.Printf("\nThis is your public key:\n%s", color.Green(pemPublicKey))
 	log.Printf("Its fingerprint is:\n%s\n", color.Cyan(fingerprint))
-	log.Print("\nTo use this key in Vespa Cloud it must be added to your tenant here:")
+	log.Print("\nTo use this key in Vespa Cloud click 'Add custom key' at")
```
mpolden, line 82: Ok, but in general I think that if the CLI needs to over-explain how to do something in the console, it's a sign that the console UI itself must be improved.

mpolden (approved):

client/go/cmd/api_key.go
bratseth: Yes, we could improve this, but probably better to do after the cli is released.

PR #18985, opened 2021-09-07
Revert "Revert "Balder/resolve inhertitance early""

baldersheim:
Reverts vespa-engine/vespa#18982


baldersheim:
Better solution in #18996

PR #18988, opened 2021-09-07
Hmusum/use xxhash64 part 1 [run-systemtest]

hmusum:


config/src/main/java/com/yahoo/config/subscription/impl/PayloadChecksum.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Checksums of config definition payload or config payload,
+ * md5 and xxhash64 are the supported types at the moment.
+ *
+ * @author hmusum
+ */
+public class PayloadChecksum {
+
+    private static final Pattern hexChecksumPattern = Pattern.compile("[0-9a-zA-Z]+");
```
baldersheim, line 15: hex ? Perhaps a-fA-F ?

baldersheim (approved):

config/src/main/java/com/yahoo/config/subscription/impl/PayloadChecksum.java
hmusum: I just used the same pattern as before and we use `HexDump.toHexString()`. I'll change the pattern in next PR.

PR #18994, opened 2021-09-07
Expose real node resources

freva:
We may get effectively exclusive host even if the application does not specify `exclusive="true"`: if shared hosts are not configured in the zone, so this calculation is not perfect. I think we should remove `boolean exclusive` from `HostCalculator::realResourcesOf` and instead treat it as exclusive if host resources == node resources.

For hosts, `resources` and `realResources` are now always equal, and both are real resources. I guess we should transition to changing `resources` to return advertised resources once all the clients have migrated to `realResources` for consistency?


bratseth:
>  treat it as exclusive if host resources == node resources

Yes, in that method this should work fine, although not in the others so it may be fine to keep it symmetric.

> I guess we should transition to changing resources to return advertised resources once all the clients have migrated to realResources for consistency?

That would be nice, although not sure its worth the effort. Ideally I also want to split into two classes ...

bratseth (approved):

PR #18996, opened 2021-09-07
Verify that rank profile inheritance is correct and sound. The rank p…

baldersheim:
…rofile must exist and be visible in the inheritance tree of the searchdefinition.

@bratseth or @hmusum PR


config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
```diff
@@ -199,20 +200,25 @@ public void setInherited(String inheritedName) {
...
+                        ((getSearch() != null) ? getSearch().getName() : " global rank profiles") + "'.");
             }
         }
-        return null;
+        return inherited;
+    }
+
+    private RankProfile resolveInherited() {
+        if (inheritedName == null) return null;
+        return (getSearch() != null)
```
bratseth, line 217: Nested flat ternatry :neutral_face:

bratseth (approved):
If anybody is doing this today we'll fail the release. I guess not, so probably fine, but not 100% sure.

baldersheim:
Given that I just made a test case doing this and that there were 1 old test doing the same error, I think guarding it with a feature flag is probably wise.

bratseth (approved):
👍

bratseth:
```
15:46:13 [INFO] Compiling 7 source files to /sd/workspace/src/github.com/vespa-engine/vespa/clustercontroller-apps/target/classes
15:46:13 [ERROR] Tests run: 4, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.996 s <<< FAILURE! - in com.yahoo.vespa.DocumentGenTest
15:46:13 [ERROR] testMusic(com.yahoo.vespa.DocumentGenTest)  Time elapsed: 0.012 s  <<< ERROR!
15:46:13 java.lang.IllegalArgumentException: Rank profile 'default' inherits self, which is illegal.

```
:joy:

config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
baldersheim: Not flat anymore.

baldersheim:
I suspect that we allow inheritance of rankprofile of same name if it exists in a parent document type......
An a faint memory from times far gone says it is to allow overriding of the "default" rank profile. I wonder if the default rankprofile is an implicit unmentioned inheritance.
Why make it simple when it can be made hard.

```
  rank-profile default inherits default {
    first-phase {
      expression: nativeRank(title,artist) + attribute(popularity)
    }
  }
```

bratseth:
Yep, I think we need to make an exception for "default".

PR #19001, opened 2021-09-07
Add metrics for chained merge failures (total and node busy)

vekterli:
@geirst please review
@baldersheim FYI

Improves visibility into how many merges chains are unwound due to hitting queue limits in nodes further down the chain.


geirst (approved):
👍

PR #19002, opened 2021-09-07
Hmusum/use xxhash64 part 2 [run-systemtest]

hmusum:


config/src/main/java/com/yahoo/vespa/config/LZ4PayloadCompressor.java
```diff
@@ -14,25 +13,20 @@
...
 
-    private static final String VESPA_CONFIG_PROTOCOL_COMPRESSION_LEVEL = "VESPA_CONFIG_PROTOCOL_COMPRESSION_LEVEL";
-    private static final Compressor compressor = new Compressor(CompressionType.LZ4, getCompressionLevel());
-
-    private static int getCompressionLevel() {
-        return Integer.parseInt(ConfigUtils.getEnvValue("0",
-                System.getenv(VESPA_CONFIG_PROTOCOL_COMPRESSION_LEVEL),
-                System.getProperty(VESPA_CONFIG_PROTOCOL_COMPRESSION_LEVEL)));
-    }
+    private static final Compressor compressor = new Compressor(CompressionType.LZ4, 0);
```
baldersheim, line 16: Compression level 0 ?
I thought range was 1....15 or something.

baldersheim (approved):

config/src/main/java/com/yahoo/vespa/config/LZ4PayloadCompressor.java
hmusum: Looks like the level isn't used explicitly (from Compressor.java):

    private LZ4Compressor getCompressor() {
        return level < 7 ? factory.fastCompressor() : factory.highCompressor();
    }

PR #19003, opened 2021-09-08
correct heuristic for -(N)

takamabe:
interpret "-(N)" as a negative term searching a positive number.


bratseth (approved):
Ok, yes that makes sense.

takamabe:
Thank you for checking.

PR #19009, opened 2021-09-08
Wait for deployment to complete when discovering services

mpolden:
Also links directly to the deployment run when deploying.

@bratseth


bratseth (approved):
Great!

PR #19015, opened 2021-09-08
Use distributor stripe index when setting up reporter for PendingMess…

geirst:
…ageTracker.

This ensures we can access each individual reporter, instead of just one of them.

@vekterli please review


vekterli (approved):
:+1:

PR #19018, opened 2021-09-08
Bjorncs/service dump

bjorncs:
FYI @kkraune

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/servicedump/JvmDumpProducer.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/servicedump/JvmDumpProducer.java
@@ -1,0 +1,43 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.node.admin.maintenance.servicedump;
+
+import com.yahoo.vespa.hosted.node.admin.container.ContainerOperations;
+import com.yahoo.vespa.hosted.node.admin.nodeagent.NodeAgentContext;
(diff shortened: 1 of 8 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/servicedump/JvmDumpProducer.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override public String name() { return NAME; }
+
+    @Override
+    public void produceArtifact(NodeAgentContext context, String configId, UnixPath resultDirectoryInNode) throws IOException {
+        UnixPath vespaJvmDumper = new UnixPath(context.pathInNodeUnderVespaHome("bin/vespa-jvm-dumper"));
+        context.log(log, Level.INFO,
+                "Executing '" + vespaJvmDumper + "' with arguments '" + configId + "' and '" + resultDirectoryInNode + "'");
+        CommandResult result = container.executeCommandInContainerAsRoot(
+                context, vespaJvmDumper.toString(), configId, resultDirectoryInNode.toString());
+        context.log(log, Level.INFO,
```
hakonhall, line 37: Consider folding the above context.log into this.  Yes, you won't get a warning before execution, but you likely don't need it:  If this statement isn't reached, you'll get an exception and stacktrace pointing to this code. (and hopefully with sufficient details on configId etc?)

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/servicedump/JvmDumpProducer.java
bjorncs: I will do this in future PR.

PR #19022, opened 2021-09-08
Increase tick wait duration for main distributor thread when running …

geirst:
…with multiple stripes.

This because it will no longer be running background maintenance jobs
(non-event tick will instead be used primarily for resending full bucket fetches etc).

@vekterli please review


vekterli (approved):
:+1:

PR #19023, opened 2021-09-08
Port additional DB updater tests and fix delayed sending regression

vekterli:
@geirst please review

Addresses a missing piece of functionality in the new code path where
queued bucket rechecks during a pending cluster state time window would
not be sent as expected when the pending state has been completed and
activated.


geirst (approved):
👍

PR #19024, opened 2021-09-08
Allow inheriting any rankprofile until enforcing is enabled explicit.

baldersheim:
@bratseth or @hmusum PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
@@ -3,6 +3,7 @@ package com.yahoo.searchdefinition;
 
 import ai.vespa.rankingexpression.importer.configmodelview.ImportedMlModels;
 import com.yahoo.config.application.api.ApplicationPackage;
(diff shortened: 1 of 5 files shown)


config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
```diff
@@ -214,7 +217,8 @@ private RankProfile getInherited() {
                 if (search.getDeployProperties().featureFlags().enforceRankProfileInheritance()) {
                     throw new IllegalArgumentException(msg);
                 } else {
-                    log.warning(msg);
+                    deployLogger.log(Level.WARNING, msg);
```
freva, line 220:
```suggestion
                    deployLogger.logApplicationPackage(Level.WARNING, msg);
```
To show this as a notification in console, e.g.:
Screenshot from 2021-09-08 18-25-41

baldersheim: Thanks, will do.

PR #19026, opened 2021-09-08
add extra method for use by unregister RPC api

arnej27959:

@havardpe please review

### DIFF
--- a/slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
+++ b/slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
@@ -93,6 +93,7 @@ void LocalRpcMonitorMap::addLocal(const ServiceMapping &mapping,
         if (exists.spec == mapping.spec) {
             LOG(debug, "added mapping %s->%s was already present",
                 mapping.name.c_str(), mapping.spec.c_str());
+            // clear exists.ignoreFirstOk ?
             inflight->doneHandler(OkState(0, "already registered"));
             return;
         }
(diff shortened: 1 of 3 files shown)


slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.h
```diff
@@ -126,6 +129,9 @@ class LocalRpcMonitorMap : public MapListener,
     void addLocal(const ServiceMapping &mapping,
                   std::unique_ptr<ScriptCommand> inflight);
 
+    /** for use by unregister API, returns true on success */
+    bool removeLocal(const ServiceMapping &mapping);
```
havardpe, line 133: consider using wouldConflict in the consensus map instead to get this bool (is this a 'clean' remove)

havardpe (commented):
might need to clear ignoreFirstOk when reusing the per service object.

arnej27959:

slobrok/src/vespa/slobrok/server/local_rpc_monitor_map.cpp
```diff
@@ -93,6 +93,7 @@ void LocalRpcMonitorMap::addLocal(const ServiceMapping &mapping,
         if (exists.spec == mapping.spec) {
             LOG(debug, "added mapping %s->%s was already present",
                 mapping.name.c_str(), mapping.spec.c_str());
+            // clear exists.ignoreFirstOk ?
```
havardpe, line 96: another question; should we really return 'ok' when we known the service is in the process of being removed?

havardpe (approved):

PR #19028, opened 2021-09-08
As there are very few unique dimensions, normally in the rannge 5-30,…

baldersheim:
… a dictionary will significantly reduce memory bloat.

One large instance running OOM had 25 unique, but 500k DimensionId objects.

@gjoranv PR

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/DimensionId.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/DimensionId.java
@@ -1,6 +1,9 @@
 // Copyright 2020 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package ai.vespa.metricsproxy.metric.model;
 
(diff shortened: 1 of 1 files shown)


baldersheim:
@hmusum PR

hmusum (approved):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/DimensionId.java
```diff
@@ -1,17 +1,23 @@
 // Copyright 2020 Oath Inc. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package ai.vespa.metricsproxy.metric.model;
 
+import com.yahoo.yolean.concurrent.CopyOnWriteHashMap;
```
gjoranv, line 4: Seems like there is a copy of the same class in vespajlib that could be used here. I would like to remove the yolean module and fold its unique contents into vespajlib.

gjoranv (commented):
👍

PR #19029, opened 2021-09-08
Use a dictionary for metrid ids. There are few unique ones (less than…

baldersheim:
… 1k), but many instances.

@gjoranv PR


baldersheim:
Or @hmusum

hmusum (approved):

PR #19030, opened 2021-09-08
Instead of creatin a new Set of consumers every time, just keep an un…

baldersheim:
…midifiable set that can be used directly. Saves both memory and space.

@hmusum or @gjoranv PR


hmusum (approved):

gjoranv:
👍

PR #19031, opened 2021-09-09
Use both xxhash64 and md5 for config payload in requests and responses [run-systemtest]

hmusum:
Chacking of equal config with hasEqualConfig() still uses only md5.


baldersheim (approved):
Good solution with multiple checksums.

PR #19034, opened 2021-09-09
Add guard to avoid NullPointerException

hmusum:
Happened in a PR build...


hmusum:
Fixed in https://github.com/vespa-engine/vespa/pull/19032

PR #19047, opened 2021-09-09
Avoid stack overflow for function inlining case

lesters:
@bratseth Please review.

The issue at hand is something like the following rank profile:

```
rank-profile test {
    first-phase {
        expression: foo
    }
    function foo(x) {
        expression: x + x
    }
    function inline foo() {
        expression: foo(2)
    }
}
```

The expectation of the user writing this is that `inline foo()` is different than `foo(x)`. However, Vespa does not support multiple functions with the same name - even with different signatures. So the last `foo` here replaces the first. Since this is probably not expected by the user, we add a deploy time warning.

More problematically, loop invocation is not detected when calling `foo(2)` here since this is not recognized as a function invocation (number of arguments differ). The code in `FunctionInliner` did not differentiate between arguments, and this would then cause an infinite recursion stack overflow error. A guard was put in there.


bratseth (approved):
👍

PR #19053, opened 2021-09-09
After the first try it is no longer the first try.

baldersheim:
@jonmv or @bjorncs PR


baldersheim:
Or @hmusum PR

bratseth (approved):

baldersheim:
@vekterli FYI

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/SlobrokPolicy.java
```diff
@@ -33,7 +33,7 @@
                         }
                     } catch (InterruptedException e) {
                     }
-                    firstTry = true;
+                    firstTry = false;
```
jonmv, line 36: The best

PR #19054, opened 2021-09-09
Reapply "Use both xxhash64 and md5 for config payload in requests and responses [run-systemtest]"

hmusum:
Fixes an issue with checksum validation that made some tests fail


baldersheim (commented):
👍

PR #19057, opened 2021-09-10
Application package diff

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
```diff
@@ -32,16 +33,20 @@ public DeploymentUpgrader(Controller controller, Duration interval) {
...
                     try {
                         attempts.incrementAndGet();
                         JobId job = new JobId(instance.id(), JobType.from(controller().system(), deployment.zone()).get());
                         if ( ! deployment.zone().environment().isManuallyDeployed()) continue;
+
+                        Run last = controller().jobController().last(job).get();
+                        Versions target = new Versions(systemVersion, last.versions().targetApplication(), Optional.empty(), Optional.empty());
                         if ( ! deployment.version().isBefore(target.targetPlatform())) continue;
-                        if (   controller().clock().instant().isBefore(controller().jobController().last(job).get().start().plus(Duration.ofDays(1)))) continue;
+                        if (   controller().clock().instant().isBefore(last.start().plus(Duration.ofDays(1)))) continue;
```
jonmv, line 49: Hmm, perhaps we should remove this, for those who deploy every day to dev, so their deployments can be upgraded when they sleep, too?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private static Optional<String> diff(Optional<ZipEntryWithContent> left, Optional<ZipEntryWithContent> right, int maxDiffSizePerFile) {
+        Optional<byte[]> leftContent = left.flatMap(ZipEntryWithContent::content);
+        Optional<byte[]> rightContent = right.flatMap(ZipEntryWithContent::content);
+        if (leftContent.isPresent() && rightContent.isPresent() && Arrays.equals(leftContent.get(), rightContent.get()))
+            return Optional.empty();
+
+        if (left.map(entry -> entry.content().isEmpty()).orElse(false) || right.map(entry -> entry.content().isEmpty()).orElse(false))
+            return Optional.of(String.format("Diff skipped: File too large (%s -> %s)\n",
```
jonmv, line 64: Why too large when new/deleted?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
```diff
@@ -0,0 +1,114 @@
...
+                    left.map(e -> e.size() + "B").orElse("new file"), right.map(e -> e.size() + "B").orElse("file deleted")));
+
+        if (leftContent.map(c -> isBinary(c)).or(() -> rightContent.map(c -> isBinary(c))).orElse(false))
```
jonmv, line 67: If left is non-binary, and present, but right is binary, this should still trigger, but it won't.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
```diff
@@ -0,0 +1,114 @@
...
+            return Optional.of(String.format("Diff skipped: File is binary (%s -> %s)\n",
+                    left.map(e -> e.size() + "B").orElse("new file"), right.map(e -> e.size() + "B").orElse("file deleted")));
```
jonmv, line 69: Wouldn't it be better to detect new/deleted files first, and then do all the more involved comparisons aftterwards?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
```diff
@@ -0,0 +1,114 @@
...
+    private static Map<String, ZipEntryWithContent> readContents(ApplicationPackage app, int maxFileSizeToDiff) {
+        return new ZipStreamReader(new ByteArrayInputStream(app.zippedContent()), entry -> true, maxFileSizeToDiff, false).entries().stream()
+                .collect(Collectors.toMap(entry -> entry.zipEntry().getName(), e -> e));
+    }
+
+    private static List<String> lines(byte[] data) {
+        List<String> lines = new ArrayList<>(Math.min(16, data.length / 100));
+        try (ByteArrayInputStream stream = new ByteArrayInputStream(data);
+             InputStreamReader streamReader = new InputStreamReader(stream, StandardCharsets.UTF_8);
+             BufferedReader bufferedReader = new BufferedReader(streamReader)) {
```
jonmv, line 86: It's fine to collapse this to a single declaration, as close will propagate.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
```diff
@@ -0,0 +1,114 @@
...
+        for (int i = 0; i < lengthToCheck; i++) {
+            byte b = data[i];
+            if (b < 0x9) return true;
+
+            // TAB, newline/line feed, carriage return
+            if (b == 0x9 || b == 0xA || b == 0xD) ascii++;
+            else if (b >= 0x20 && b <= 0x7E) ascii++;
+        }
+
+        return (double) ascii / lengthToCheck < 0.95;
```
jonmv, line 112: It's science!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
jonmv: Hmm, I guess you wan to print the diff if it's not large and one file is new.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
jonmv: Ok so this is fine.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -424,11 +425,18 @@ public ApplicationVersion submit(TenantAndApplicationId id, Optional<SourceRevis
                                                 sourceUrl,
                                                 revision.map(SourceRevision::commit),
                                                 false));
+            String diff = application.get().latestVersion()
+                    .map(ApplicationVersion::buildNumber)
+                    .flatMap(build -> build.isPresent() ? Optional.of(build.getAsLong()) : Optional.empty())
```
jonmv, line 445: This should always be present.

jonmv: Why not justt use getApplicationPackage from ApplicationController, with latestVersion and app id as arguments? Ah, that's the one you removedd in the first commit ... :>

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
freva: If file is added/deleted, then then diff is the entire file with lines prefixed with +/-, so the diff will be slightly bigger than the file.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
```diff
@@ -757,6 +757,12 @@ public void testApplicationApi() {
                                       .data(createApplicationSubmissionData(packageWithService, 123)),
                               "{\"message\":\"Application package version: 1.0.2-commit1, source revision of repository 'repository1', branch 'master' with commit 'commit1', by a@b, built against 6.1 at 1970-01-01T00:00:01Z\"}");
 
+        tester.assertResponse(request("/application/v4/tenant/tenant1/application/application1/diff/2", GET).userIdentity(HOSTED_VESPA_OPERATOR),
```
jonmv, line 760: Please also do this for the dev path :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
freva: Not sure if that would be any easier? I still want to do diff as long as the file is non-binary (in that case the diff is effectively the entire file prefixed with +/-)

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiTest.java
freva: Added

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
freva: Simplified.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageDiff.java
freva: Thanks, fixed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
freva: Too scared to introduce more changes in this PR :scream:

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/integration/ApplicationStoreMock.java
```diff
@@ -62,6 +62,14 @@ private static ApplicationId testerId(TenantName tenant, ApplicationName applica
         return Optional.ofNullable(diffs.get(appId(tenantName, applicationName))).map(map -> map.get(buildNumber));
     }
 
+    @Override
+    public void pruneDiffs(TenantName tenantName, ApplicationName applicationName, long beforeBuildNumber) {
+        Optional.ofNullable(diffs.get(appId(tenantName, applicationName)))
+                .ifPresent(map -> map.keySet().stream()
```
jonmv, line 68: `map.keySet().removeIf(...)`

jonmv (approved):

jonmv:
Great, this will be very nice when helping out our cloud users!

PR #19061, opened 2021-09-10
Expose all endpoints for a given instance

oyving:
FYI @ldalves

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
@@ -354,8 +354,8 @@ public class RoutingApiHandler extends AuditLoggingRequestHandler {
     }
 
     private static void endpointToSlime(Cursor object, Endpoint endpoint) {
-        object.setString("endpoint", endpoint.name());
-        object.setString("dns", endpoint.dnsName());
+        object.setString("name", endpoint.name());
+        object.setString("dnsName", endpoint.dnsName());
         object.setString("routingMethod", endpoint.routingMethod().name());
         object.setString("cluster", endpoint.cluster().value());
         object.setString("scope", endpoint.scope().name());
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
@@ -21,6 +21,7 @@ import com.yahoo.vespa.hosted.controller.Controller;
 import com.yahoo.vespa.hosted.controller.Instance;
 import com.yahoo.vespa.hosted.controller.api.application.v4.model.EndpointStatus;
 import com.yahoo.vespa.hosted.controller.api.identifiers.DeploymentId;
+import com.yahoo.vespa.hosted.controller.application.Endpoint;
 import com.yahoo.vespa.hosted.controller.application.TenantAndApplicationId;
 import com.yahoo.vespa.hosted.controller.auditlog.AuditLoggingRequestHandler;
 import com.yahoo.vespa.hosted.controller.routing.GlobalRouting;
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
```diff
@@ -85,11 +86,38 @@ private HttpResponse get(Path path, HttpRequest request) {
...
 
+    private HttpResponse endpoints(Path path) {
+        var instanceId = instanceFrom(path);
+        var endpoints = controller.routing().endpointsOf(instanceId);
+        var zoneStatus = endpoints.asList().stream()
+                .flatMap(e -> e.zones().stream())
+                .distinct()
+                .collect(Collectors.toMap(
+                        zone -> zone,
+                        zone -> controller.routing().globalRotationStatus(new DeploymentId(instanceId, zone))
```
mpolden, line 103: This will only get status for endpoints having rotations. Have a look at this method: https://github.com/vespa-engine/vespa/blob/a8c616b18803e2e3a959b9ba987deb64fabf252a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java#L258-L306

Try extracting a method from this and reuse it here.

oyving:
Alright, take two. Trying to use same methods as the old `toSlime()` method to calculate the status. Not sure if there are some underlying constraints here that I'm ignoring and that could create confusing and contradictory output. E.g. for a given endpoint+zone, can there be two statuses that contradict each other?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
```diff
@@ -369,4 +426,7 @@ private static String asString(RoutingMethod method) {
         }
     }
 
+    private static class RoutingStatus {
```
mpolden, line 429: Unused?

mpolden (commented):
> E.g. for a given endpoint+zone, can there be two statuses that contradict each other?

No, because a deployment/zone either supports shared routing (always rotation) or exclusive routing (always status from routing policy).

However, since we supported both methods at some point this constraint isn't enforced in code.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/routing/RoutingApiHandler.java
oyving: Indeed. Amended and removed

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/routing/responses/endpoint/endpoints.json
```diff
@@ -0,0 +1,31 @@
+{
+  "endpoints": [
+    {
+      "endpoint": "default",
+      "dns": "a1--t1.global.vespa.oath.cloud",
```
mpolden, line 5: How is this going to be used? Perhaps better to include the URL? Otherwise console won't know about scheme or port.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/routing/responses/endpoint/endpoints.json
```diff
@@ -0,0 +1,31 @@
+{
+  "endpoints": [
+    {
+      "endpoint": "default",
```
mpolden, line 4: Consider calling this `name` or `endpointName`.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/routing/responses/endpoint/endpoints.json
oyving: We aren't really going to use those values much from the console. We only care about the name and the status for the zone for now. I just added the other values for completeness.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/routing/responses/endpoint/endpoints.json
oyving: `name`, agreed.

mpolden (approved):

ldalves:
Thanks!

PR #19063, opened 2021-09-10
Add node to system info

gjoranv:
The new user api is `systemInfo.node().id()`, which is meant as a discriminator between Jdisc containers.
Node can be extended with other properties as we find appropriate.

### DIFF
--- a/container-disc/src/main/java/com/yahoo/container/jdisc/SystemInfoProvider.java
+++ b/container-disc/src/main/java/com/yahoo/container/jdisc/SystemInfoProvider.java
@@ -2,11 +2,13 @@
 package com.yahoo.container.jdisc;
 
 import ai.vespa.cloud.Environment;
+import ai.vespa.cloud.Node;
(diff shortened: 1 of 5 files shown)


bratseth (commented):
Would it make sense to just use the integer node index here? That would make this easier to use.

Could we also pass the count of nodes in the cluster?

gjoranv:
Omitting the cluster id and the 'container'/'qrserver' prefix is a good idea. After all, you don't need to keep the id/index unique between clusters. Should we change the name to 'index'?

I'll look into the node count as well.

bjorncs:
`Node` can expose index and full config id in separate methods, or alternatively expose cluster name in addition to index?

bratseth:
> Should we change the name to 'index'?

Yes? And document it is guaranteed to be unique but not contiguous from 0.

gjoranv:
Starting fresh in a new PR.

PR #19067, opened 2021-09-10
Arnej/use new logic

arnej27959:

@havardpe for discussion


havardpe (commented):
looks good.
New world logic should probably ignore doAdd

havardpe (approved):
minor note: the new doAdd is counted as a wantAdd in cnt.

PR #19083, opened 2021-09-12
Revert "Revert "Balder/wire executor to ml model importing""

baldersheim:
Reverts vespa-engine/vespa#19081


baldersheim:
@hmusum or @lesters PR
Note that the last commit should not have been necessary. I think that it should have been in the config-model-fat bundle.

hmusum (approved):

PR #19085, opened 2021-09-13
Add node to system info take2

gjoranv:
New apis:
`systemInfo.node().index()`
`systemInfo.cluster().size()`

I thought about putting the Node inside Cluster, but the actual node for the runtime seems more natural to keep in SystemInfo. However, the cluster could be extended to contain  e.g. a list of node indices, cluster id etc.

### DIFF
--- a/container-core/src/main/resources/configdefinitions/container.qr.def
+++ b/container-core/src/main/resources/configdefinitions/container.qr.def
@@ -23,5 +23,10 @@ rpc.slobrokId string default="" restart
 ## this string will be unique for every QRS in a Vespa application.
 discriminator string default="qrserver.0" restart
 
+## Index of this container inside the cluster. Guaranteed to be non-negative
+## and unique for every container in a cluster, but not necessarily contiguous
+## or starting from zero.
+nodeIndex int default=0
+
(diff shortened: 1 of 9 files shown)


container-core/src/main/resources/configdefinitions/container.qr.def
```diff
@@ -23,5 +23,10 @@ rpc.slobrokId string default="" restart
 ## this string will be unique for every QRS in a Vespa application.
 discriminator string default="qrserver.0" restart
 
+## Index of this container inside the cluster. Guaranteed to be non-negative
+## and unique for every container in a cluster, but not necessarily contiguous
+## or starting from zero.
+nodeIndex int default=0
```
bjorncs, line 29: Consider rename to `clusterIndex`

bjorncs (approved):

container-core/src/main/resources/configdefinitions/container.qr.def
gjoranv: To me, that would sound like the index of a cluster. Anyway, this is purely internal, so we can change it later.

bjorncs (approved):

bratseth (approved):
Great!

PR #19086, opened 2021-09-13
Ensure internal node-admin state is in sync with orchestrator state

freva:
Sometimes the host is suspended/resumed from outside by operator or node-admin converges before switching to real, then node-admin wont ever re-converge again because of the internal state. This PR makes sure we only take the short circuit if the orchestrator state from `NodeSpec` is equal to what we expect with the converged internal state.

Fixes VESPA-21249

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminStateUpdater.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminStateUpdater.java
@@ -102,6 +102,8 @@ public class NodeAdminStateUpdater {
      * with respect to: freeze, Orchestrator, and services running.
      */
     public void converge(State wantedState) {
+        NodeSpec node = nodeRepository.getNode(hostHostname);
+        boolean hostIsActiveInNR = node.state() == NodeState.active;
         if (wantedState == RESUMED) {
             adjustNodeAgentsToRunFromNodeRepository();
         } else if (currentState == TRANSITIONING && nodeAdmin.subsystemFreezeDuration().compareTo(FREEZE_CONVERGENCE_TIMEOUT) > 0) {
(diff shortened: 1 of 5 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminStateUpdater.java
```diff
@@ -110,21 +112,18 @@ public void converge(State wantedState) {
...
-            NodeState currentNodeState = nodeRepository.getNode(hostHostname).state();
-            if (currentNodeState == NodeState.active) orchestrator.resume(hostHostname);
+            if (hostIsActiveInNR) orchestrator.resume(hostHostname);
 
             throw new ConvergenceException("Timed out trying to freeze all nodes: will force an unfrozen tick");
         }
 
-        if (currentState == wantedState) return;
+        boolean wantFrozen = wantedState != RESUMED;
+        if (currentState == wantedState && wantFrozen == node.orchestratorStatus().isSuspended()) return;
```
hakonhall, line 121: This will return even if wantedState is SUSPENDED, the host is suspended, and a child is resumed.  Which is one of the use-case you're trying to fix?

Case B: If the wanted state is RESUMED and !hostIsActiveInNR, then I believe isSuspended() is true since it is UNKNOWN and != NO_REMARKS.  As it is now it would mean printing "State changed from ... to ..." every tick.  Instead we'd like to return here.

freva: The main use-case is when HA converges before switching to real, typically it is resuming after updating HA. After it switches, it wont resume against real and the upgrade halts.

Case B: As far as I can see from the code, `orchestratorStatus` is not set in node response if the node is not allocated, that is also the case for `NO_REMARKS`, so it should be `NO_REMARKS` in `NodeSpec::orchestratorStatus`?

hakonhall: Right, but you also mentioned operators: So if an operator forces resume of a child.  Not a big deal.  But the logic is a bit harder to follow since it doesn't match the below.

Case B: Agree.

hakonhall (approved):

PR #19087, opened 2021-09-13
Ensure unique NodeAgent ID for reallocations

freva:
Fixes VESPA-21439

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/RealNodeRepository.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/RealNodeRepository.java
@@ -178,6 +178,7 @@ public class RealNodeRepository implements NodeRepository {
                 node.ipAddresses,
                 node.additionalIpAddresses,
                 reports,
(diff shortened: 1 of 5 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/RealNodeRepository.java
```diff
@@ -178,6 +178,7 @@ private static NodeSpec createNodeSpec(NodeRepositoryNode node) {
                 node.ipAddresses,
                 node.additionalIpAddresses,
                 reports,
+                node.history.stream().map(event -> new Event(event.agent, event.event, Instant.ofEpochMilli(event.at))).collect(Collectors.toUnmodifiableList()),
```
hakonhall, line 181: handle null event.at ?

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/RealNodeRepository.java
freva: Cant be null, it is `long`

hakonhall: OK, handle that `at` is absent then?

PR #19088, opened 2021-09-13
Add runtime dependency to vespa-gtest.

toregge:
@aressem : please review


toregge:
This is related to vespa-engine/vespa-3rdparty-deps#85

aressem (approved):

baldersheim (commented):
Why does vespa need gtest as runtime dependency at all ?

toregge:
vespa-tensor-instructions-benchmark is part of vespa and links with gtest.

baldersheim:
Why is that part of runtime install ?
If is benchmarking it does not sound right to depend on gtest.
I noticed that it seemed to be used in a system test. If that is the only use it should be in a different package.

baldersheim (approved):
Since this PR does not introduce any dependency, approve. Fix root issue later on.

PR #19089, opened 2021-09-13
Require all event fields

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/bindings/NodeRepositoryNode.java
```diff
@@ -203,12 +203,17 @@ public String toString() {
...
-        public String agent;
-        @JsonProperty
-        public String event;
-        @JsonProperty
-        public long at;
+        public final String agent;
+        public final String event;
+        public final long at;
+
+        public Event(@JsonProperty(value = "agent", required = true) String agent,
```
hakonhall, line 210: This will throw an exception if not present, and would require a patch on host-admin or configserver to fix!?  How about ignoring av event if it cannot be fully deserialized (w/all fields present)?

freva: :shrug: I don't see the point handling this either way. This is not gonna happen since all event fields in NR are non-optional anyway.
The current handling will default to 0 which is `Instant.EPOCH` and isn't that (in general) better than not including the event entirely?

hakonhall: The rest of this file (except 2 properties in Membership) follows the strategy that everything can be missing, and use Long, Boolean, and Integer instead of the primitives. Thus deserialization never fails, but the code accessing the deserialized jackson instance may choose to interpret null however it sees fit: fail, replace it with Epoch, etc. Example: `Optional.ofNullable(node.wantedFirmwareCheck).map(Instant::ofEpochMilli)`.  I agree that interpreting a null Event.at as Epoch is better than filtering the event.  So my concern ends upas  just as a nit: instead of `long` here, do something like `orElse(Instant.Epoch())` where `.at()` is invoked. LGTM either way.

But the class should be annotated with `@JsonIgnoreProperties(ignoreUnknown = true)` to allow the server to add fields without breaking deserialization. `@JsonInclude(JsonInclude.Include.NON_NULL)` is the other annotation typically added.  It's only needed for serialization to JSON, and I don't see the need for us to ever need to serialize the history, but then it doesn't hurt to be consistent and add it either.

hakonhall (approved):

PR #19110, opened 2021-09-14
Remove double mapping in cache [run-systemtest]

hmusum:
If configs have the same checksum, they are the same, no need for
double mapping.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/ServerCache.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/ServerCache.java
@@ -1,4 +1,4 @@
-// Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.config.server;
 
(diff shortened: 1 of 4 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/ServerCache.java
```diff
@@ -21,9 +21,7 @@
     private final ConfigDefinitionRepo builtinConfigDefinitions;
     private final ConfigDefinitionRepo userConfigDefinitions;
 
-    // NOTE: The reason we do a double mapping here is to de-dupe configs that have the same md5.
-    private final Map<ConfigCacheKey, String> md5Sums = new ConcurrentHashMap<>();
-    private final Map<String, ConfigResponse> md5ToConfig = new ConcurrentHashMap<>();
+    private final Map<ConfigCacheKey, ConfigResponse> configChecksums = new ConcurrentHashMap<>();
```
baldersheim, line 24: Before doing this change you should tae a heapdump in production and see if we have a benefit from the indirection or not. I seem to recall that was a significant difference in the number of elements in md5sums and md5ToConfig.

hmusum: Yes, you are right, looking at a heap dump from production shows that there are about 4 times as many entries in `md5sums` as in `md5ToConfig`. Will close this PR

PR #19119, opened 2021-09-14
Add feature flag to enable tenant s3 archive uploading

bjorncs:
@hakonhall or @freva please review

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
@@ -5,6 +5,9 @@ import com.yahoo.config.provision.SystemName;
 import com.yahoo.config.provision.TenantName;
 import com.yahoo.config.provision.zone.ZoneId;
 import com.yahoo.text.Text;
+import com.yahoo.vespa.flags.FetchVector;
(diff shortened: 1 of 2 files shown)


freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
```diff
@@ -41,23 +44,30 @@
...
-        if (enabled) {
+        if (enabled(zoneId)) {
             return Optional.of(URI.create(Text.format("s3://%s/%s/", findOrAssignBucket(zoneId, tenant), tenant.value())));
         } else {
             return Optional.empty();
         }
     }
 
+    private boolean enabled(ZoneId zone) {
+        return system.isPublic() || Flags.ENABLE_ONPREM_TENANT_S3_ARCHIVE.bindTo(flagSource)
```
hakonhall, line 66: This will disable it for CD?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/archive/CuratorArchiveBucketDb.java
bjorncs: It's already effectively disabled as the archive maintainer is disabled through feature flag in cd.

PR #19127, opened 2021-09-14
Move initializing handling to top-level distributor

vekterli:
@geirst and @toregge please review

Add a listener interface that lets the top-level distributor intercept
cluster state activations and use this for triggering the node init edge.
This happens when all stripes are paused so this is safe from data races.
Legacy code in the `DistributorStripe` remains for now.

Note: the `TopLevelBucketDBUpdater` component only exists when legacy
mode is _not_ active, so the new code path cannot be invoked in this case.


toregge (approved):

geirst (approved):
👍

PR #19133, opened 2021-09-14
Create persistence engine before creating document db in BmNode.

toregge:
This order is needed to pass a proper bucket executor to the document db.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #19149, opened 2021-09-15
Tenant removal refinements

freva:
Fixes VESPA-21368 by creating a new type of tenant: `deleted`. This type tenant is by default ignored except in a few places: Tenant creation, tenant deletion, and in REST API if `includeDeleted` query property is set to `true`.

Regular users can only mark a tenant as deleted, operators have the option to forget the tenant completely by doing the same REST DELETE, but with query property `forget` set to `true`.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/TenantController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/TenantController.java
@@ -4,10 +4,7 @@ package com.yahoo.vespa.hosted.controller;
 import com.yahoo.config.provision.TenantName;
 import com.yahoo.text.Text;
 import com.yahoo.vespa.curator.Lock;
-import com.yahoo.vespa.flags.BooleanFlag;
-import com.yahoo.vespa.flags.FetchVector;
 import com.yahoo.vespa.flags.FlagSource;
-import com.yahoo.vespa.flags.Flags;
 import com.yahoo.vespa.hosted.controller.api.identifiers.TenantId;
 import com.yahoo.vespa.hosted.controller.application.SystemApplication;
 import com.yahoo.vespa.hosted.controller.concurrent.Once;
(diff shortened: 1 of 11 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/TenantController.java
```diff
@@ -153,22 +163,28 @@ public void updateLastLogin(TenantName tenantName, List<LastLoginInfo.UserLevel>
...
-            controller.notificationsDb().removeNotifications(NotificationSource.from(tenant));
+            Tenant oldTenant = get(tenant, true)
+                    .orElseThrow(() -> new NotExistsException("Could not delete tenant '" + tenant + "': Tenant not found"));
+
+            if (oldTenant.type() != Tenant.Type.deleted) {
+                if (!controller.applications().asList(tenant).isEmpty())
+                    throw new IllegalArgumentException("Could not delete tenant '" + tenant.value()
+                            + "': This tenant has active applications");
+
+                accessControl.deleteTenant(tenant, credentials.get());
```
jonmv, line 176: So the tenant owner will not be able to re-create a tenant once it is deleted, without operators first clearing the tombstone? I guess that's OK.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -321,23 +336,20 @@ private void writeContact(Contact contact, Cursor contactCursor) {
...
         }
     }
 
     private static String valueOf(Tenant.Type type) {
         switch (type) {
-            case athenz: return "athenz";
-            case cloud:  return "cloud";
+            case athenz:  return "athenz";
+            case cloud:   return "cloud";
+            case deleted: return "deleted";
```
jonmv, line 352: Just don't delete any tenants before all controllers have upgraded!

jonmv (approved):
Nice :)

PR #19151, opened 2021-09-15
Dirty all nodes on application removal

freva:
Fixes VESPA-21576. With this, we will only get `inactive` nodes when new application deployment is unable to reuse some of the previous stateful nodes, e.g. due to flavor change or cluster size reduction. 

Application removal will move nodes directly to `dirty`. This is OK because application removal already requires `deployment-removal` validation override, or going through a confirmation menu in the console.


bratseth (approved):
Simple!

PR #19162, opened 2021-09-16
Flip to always use the new distributor stripe code path.

geirst:
If the number of stripes is not configured, we tune it based on the sampled number of CPU cores.

@vekterli please review
@baldersheim @toregge FYI


vekterli (approved):
😎👍

PR #19168, opened 2021-09-16
low-level testing of fnet enabled by

havardpe:
1. Manipulating how much time is spent waiting for events while also
   manipulating how the passing of time is observed by the transport
   threads.
   (fnet::TimeTools)

2. Periodic call-backs from FNET_Transport. All transport threads are
   blocked during these call-backs. Between each call-back, all
   transport threads perform a single event loop iteration and any
   async operations not performed in transport threads are flushed.
   (FNET_Transport::attach_capture_hook)

3. Combining 1 and 2 in a way that lets the test code run all
   transport threads one iteration at a time, letting the transport
   threads run concurrently with each other, but no transport thread
   is allowed to run concurrently with the test code.
   (fnet::TransportDebugger)

@arnej27959 please review

### DIFF
--- a/fnet/CMakeLists.txt
+++ b/fnet/CMakeLists.txt
@@ -29,4 +29,5 @@ vespa_define_module(
     src/tests/sync_execute
     src/tests/thread_selection
     src/tests/time
+    src/tests/transport_debugger
 )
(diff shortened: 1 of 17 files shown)


arnej27959 (approved):
looks like we should have a tech talk about this 👍

PR #19193, opened 2021-09-17
Avoid extra context switch to another thread not necessary.

baldersheim:
@vekterli PR


baldersheim:
This will be void once mutating bucket tree operations cause feed to be busy bounced.
Too risky....

PR #19195, opened 2021-09-17
We do not need the stack trace for expected warnings.

baldersheim:
@bjorncs or @hmusum PR


bjorncs (approved):
Consider changing to level fine instead of removing after some track time

baldersheim:
Will do

PR #19202, opened 2021-09-20
Remove most traces of distributor legacy mode [run-systemtest]

vekterli:
@geirst please review

Some assorted legacy bits and pieces still remain on the factory floor,
these will be cleaned up in follow-ups.

Added some unrelated FIXMEs on the use of Initializing mode for distributors in cluster
state in some old unit tests, as these should never end up being used in practice
(the distributor jumps instantly from Down -> Up without an intermediate Init state)


geirst (approved):
🥇

PR #19207, opened 2021-09-20
Remove code for only logging when using resource-limits in hosted

hmusum:
No apps are using this now, so we will always throw exception if
'resource-limits' is used in services.xml in hosted. Prepare for
removing feature flag.

@geirst please review only, will merge after change in feature flag has been pushed.


geirst (approved):
👍

PR #19212, opened 2021-09-20
Rename functions to more consistent names.

toregge:
Adjust calculation of moved docs ratio for replace mode.

@geirst : please review


geirst (approved):
👍

PR #19213, opened 2021-09-20
Encode tensors passed as encode(text)

bratseth:
This lets you pass tensors in queries as "encode(some text)" instead of a tensor value, to create a tensor of the type required by this variable created by passing the text enclosed in "encode()" through the configured encoder.

https://github.com/vespa-engine/vespa/pull/19198 lets you write tensor fields in documents by passing a text through the "encode" il function, which will again derive the target type from the target field.

Do you think this makes sense as a possible way to use encoders for simple use cases without requiring Java code?

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
+++ b/container-core/src/main/java/com/yahoo/container/core/config/testutil/HandlersConfigurerTestWrapper.java
@@ -17,6 +17,7 @@ import com.yahoo.jdisc.Metric;
 import com.yahoo.jdisc.handler.RequestHandler;
 import com.yahoo.jdisc.test.MockMetric;
 import com.yahoo.language.Linguistics;
+import com.yahoo.language.process.Encoder;
 import com.yahoo.language.simple.SimpleLinguistics;
 
 import java.io.File;
@@ -140,6 +141,7 @@ public class HandlersConfigurerTestWrapper {
             protected void configure() {
                 // Needed by e.g. SearchHandler
                 bind(Linguistics.class).to(SimpleLinguistics.class).in(Scopes.SINGLETON);
+                bind(Encoder.class).to(Encoder.FailingEncoder.class).in(Scopes.SINGLETON);
                 bind(ContainerThreadPool.class).to(SimpleContainerThreadpool.class);
                 bind(Metric.class).to(MockMetric.class);
             }
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -1786,6 +1786,27 @@
     ],
     "fields": []
   },
+  "com.yahoo.search.Query$Builder": {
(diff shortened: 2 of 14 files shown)


jobergum:
I think there is a lot of details left to be mapped out still but this is a great step forward. 

For example, getting into the details one needs the ability to specify the padding value when there are fewer tokens in the encoded string than the dimensionality of the target tensor. For regular BERT subword tokenizer/encoder it uses PAD = 0, but using MASK=104 is also used in some cases, e.g ColBERT. I'm sure there are similar behaviours for the sentencepiece tokenizer. 

Also should it really be encode and not tokenize? Since the encode does not really prepare the input to a model I think it should be tokenize and not encode. 

e.g

```
query(query_token_ids)=subword_tokenizer("user query", padding_token=0)
query(m_query_token_ids)=subpiece_tokenizer("user query")
```

And encode used to create tensor input to a given model (e.g create the input_ids, attention_mask and token_type_ids inputs to a Transformer model) 

```
query(embedding)=encode(query_token_ids), model="embedding",output="output_0") 
```

jobergum (approved):

bratseth:
Good feedback - thanks. I think padding value etc. should probably be configuration to the encoder. I'm not sure we won't need encoding-time arguments as well but it could be added when needed (with some more parsing effort), without breaking compatibility.

Terminology: We already have 
- tokenizers: Turn a text into a list of regularized (text) tokens.
- segmenters: Split a contiguous text into meaningful/useful chunks.

So imo we need another word for something that turns a contiguous text into a tensor.

bratseth:
What do you mean by this? "Since the encode does not really prepare the input to a model"

jobergum:
>What do you mean by this? "Since the encode does not really prepare the input to a model"

```
encode("This is the input text")
[2023, 2003, 1996, 7953, 3793]
```
Versus 
```
encode("this is the input text")
{'input_ids': [101, 2023, 2003, 1996, 7953, 3793, 102], 'token_type_ids': [0, 0, 0, 0, 0, 0, 0], 'attention_mask': [1, 1, 1, 1, 1, 1, 1]}
```
The latter encode example prepares the input to a model. 

>So imo we need another word for something that turns a contiguous text into a tensor.

Yes, encode is probably the right word, but we need variants which prepares input to a model

bratseth:
Ok, yes if the model needs padding and additional tensor args I think that belongs on the model side.
If the model runs in the container nodes we can already do this with tensor math. Might be we need a little bit additional support to do that with stateless models, but I see that as separate.

jobergum:
Yes, agree that it's tied to the model. Your encode will work well with the tensor model input helper functions in backend which takes the query input tensor and the document tensor. 

https://github.com/vespa-engine/sample-apps/blob/master/msmarco-ranking/src/main/application/schemas/passage.sd#L122-L132

PR #19219, opened 2021-09-21
Make flag volatile

hmusum:


metrics-proxy/src/main/java/ai/vespa/metricsproxy/http/application/ApplicationMetricsRetriever.java
```diff
@@ -52,7 +52,7 @@
     private final Thread pollThread;
     private final Set<ConsumerId> consumerSet;
     private long pollCount = 0;
-    private boolean stopped;
+    private volatile boolean stopped;
```
baldersheim, line 55: Is this necessary ?
Except for the initialisation the access is synchronized on the pollThread object.

hmusum: Synchronization is OK, but I don't think you guarantee visibility of a write to other threads that read without `volatile`? 

I see that tests still fail when deconstructing metricsproxy-caontiner, but I am not sure this is the cause...

baldersheim: But the read is also within synchronized(pollThread), and that should be safe.

I think the reason for deconstruction issue is something else. I think that this change to poll regulary just made it visible. Earlier on it would only fetch metrics if someone asked and TTL was expired. Now it happens every 30s.
I guess that requests did not come in during shutdown, but ttl can now expire during shutdown triggering error handling that is not perfect.

hmusum: OK, closing this PR

PR #19232, opened 2021-09-21
YQL programCompileExceptions are IllegalInputExceptions

bratseth:
This will avoid returning stack traces on illegal YQL input.


arnej27959 (approved):
looks good to me

PR #19249, opened 2021-09-22
Expose user feature flags in /user/v1/user response

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserFlagsSerializer.java
```diff
@@ -0,0 +1,86 @@
...
+        byte[] bytes = FlagData.serializeListToUtf8Json(filteredFlagData);
+        SlimeUtils.copyObject(SlimeUtils.jsonToSlime(bytes).get(), cursor);
+    }
+
+    private static <T> FlagData filteredFlagData(UnboundFlag<T, ?, ?> definition, Optional<FlagData> original,
+                                                 Set<TenantName> authorizedForTenantNames, boolean isOperator, FetchVector resolveVector) {
+        MutableBoolean encounteredEmpty = new MutableBoolean(false);
+        Optional<RawFlag> defaultValue = Optional.of(definition.serializer().serialize(definition.defaultValue()));
+        // Include the original rules from flag DB and the default value from code if there is no default rule in DB
+        List<Rule> rules = Stream.concat(original.stream().flatMap(fd -> fd.rules().stream()), Stream.of(new Rule(defaultValue)))
```
hakonhall, line 50: Why is it necessary to add the default value rule at the end?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserFlagsSerializer.java
freva: Flag DB could consist only of values with conditions, if none of them satisfy the fetch vector, the user would not have a value to use. This is easy in java since we have access to the flag definition, but for console we need to expose this explicitly. By putting the default value without any conditions at the end, we are guaranteed that there will always be a value that satisfies the fetch vector.

PR #19259, opened 2021-09-23
Start manual deployments on existing deployment version

jonmv:
@freva please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
@@ -46,7 +46,6 @@ public class DeploymentUpgrader extends ControllerMaintainer {
                         Run last = controller().jobController().last(job).get();
                         Versions target = new Versions(systemVersion, last.versions().targetApplication(), Optional.of(last.versions().targetPlatform()), Optional.of(last.versions().targetApplication()));
                         if ( ! deployment.version().isBefore(target.targetPlatform())) continue;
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
```diff
@@ -46,7 +46,6 @@ protected double maintain() {
                         Run last = controller().jobController().last(job).get();
                         Versions target = new Versions(systemVersion, last.versions().targetApplication(), Optional.of(last.versions().targetPlatform()), Optional.of(last.versions().targetApplication()));
                         if ( ! deployment.version().isBefore(target.targetPlatform())) continue;
-                        if (   controller().clock().instant().isBefore(last.start().plus(Duration.ofDays(1)))) continue;
```
jonmv, line 49: Can't let upgrade be blocked indefinitely by deploying once a day.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/DeploymentUpgrader.java
```diff
@@ -46,7 +46,6 @@ protected double maintain() {
...
                         if ( ! isLikelyNightFor(job)) continue;
```
jonmv, line 49: I guess someone could theoretically block upgrade by redeploying once an hour, or so, but until someone actually does this, let's leave this. The remedy for that would be to always upgrade if the version is more than, say, 7 days old. Then again, that someone could deploy with a particular runtime version once an hour ... and this is already possible today, so I don't think it's actually a problem.

freva (approved):
Great!

PR #19260, opened 2021-09-23
Arnej/java add hamming

arnej27959:

@bratseth FYI
@havardpe please review


havardpe (commented):
looks ok, but since this changes the public API spec, we probably need to agree on the function name before merging.

havardpe (approved):

PR #19268, opened 2021-09-23
Disallow cfg suspension based solely on being down

hakonhall:


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/policy/HostedVespaClusterPolicy.java
```diff
@@ -32,46 +42,28 @@ public SuspensionReasons verifyGroupGoingDownIsFine(ClusterApi clusterApi) throw
             return SuspensionReasons.nothingNoteworthy();
         }
 
-        Optional<SuspensionReasons> suspensionReasons = clusterApi.reasonsForNoServicesInGroupIsUp();
-        if (suspensionReasons.isPresent()) {
-            return suspensionReasons.get();
+        // Be a bit more cautious when removing nodes permanently
+        if (!permanent) {
+            // Disallow suspending a 2nd and downed config server to avoid losing ZK quorum.
+            if (!clusterApi.isConfigServerLike()) {
```
hakonhall, line 48: This is the only functional change in this PR.  The remaining changes are improvements to error messages and minor refactoring.

hmusum (approved):

PR #19272, opened 2021-09-24
- Shorten grace period from 30m to 5m.

baldersheim:
- Check to remove every second to get rid of deactivated instances sooner.

@hmusum or @bjorncs PR


hmusum (approved):

baldersheim:
@geirst FYI

PR #19273, opened 2021-09-24
Stopping is server is a one way street.

baldersheim:
Returning BUSY as a temporary error will cause retries, instead of fast-fail.

@bjorncs or @vekterli or @gjoranv PR

### DIFF
--- a/container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/MbusClient.java
+++ b/container-messagebus/src/main/java/com/yahoo/messagebus/jdisc/MbusClient.java
@@ -9,6 +9,8 @@ import com.yahoo.jdisc.handler.ContentChannel;
 import com.yahoo.jdisc.handler.RequestDeniedException;
 import com.yahoo.jdisc.handler.ResponseHandler;
 import com.yahoo.jdisc.service.ClientProvider;
+
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import com.yahoo.messagebus.EmptyReply;
 import com.yahoo.messagebus.Error;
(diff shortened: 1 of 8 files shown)


baldersheim:
@geirst FYI

bjorncs (approved):

jonmv:
And why don't we want to retry?

jonmv:
The whole wiring of these servers is wonky, tbh.. We should be able to stop a server from receiving new requests when signalling shutdown, let it drain its requests, and then shut it down completely. We don't do that, because this code is a mix of manual reference counting and graph generations replacing each other, where servers are first stopped, and then started, so there will be a gap where a server may be stopped, but still receive messages. These shouldn't fail permanently, but be retried against a different container, or against this container when the new graph is ready. This is my understanding, anyway.

baldersheim:
Because once a server is stopped it is never started again, as far as I can tell.
So any messages scheduled for resending will never complete.

jonmv:
This is for messages received by the container, from messagebus, not messages sent by the container.

jonmv:
This is a bridge for receiving messagebus messages, routing them through the container to a `RequestHandler` (only the docproc handler is an actual recipient, AFAIK), which in turn uses messagebus to obtain a reply from further down the chain, converts this to a response which this `MbusServer`converts to a reply again, and sends back over messagebus to the original client.

baldersheim:
Yes, and that is 99%+ of all mbus traffic

jonmv:
So the `DocumentProcessingHandler` seems to abandon any feed operations with `Progress.LATER` on destruction, on purpose. The `MbusClient` used by this handler uses either an `SharedIntermediateSession` or a `SharedSourceSession` to send its message. It looks like a closed source session will return an error, but a closed intermediate session will silently discard messages. Perhaps the intermediate session is prematurely shut down, then?

jonmv:
My point is that the original client doesn't need to get a fatal error, but _should_ get a transient error. It definitely shouldn't get no response at all.

jonmv:
We're seeing issues for all apps, but much more for some than for others. Perhaps we have both of these problems? Silently discarding `LATER` tasks, _and_ messages sent from an intermediate session?

baldersheim:
Yes, I agree that the original sender should not FATAL, if it is on a different node.
I will hold that part as I have found other issues that might cause this. However I think this might still be a root cause as I suspect we have an issue with messages dispatched to same node.

jonmv:
https://github.com/vespa-engine/vespa/pull/19285 should hopefully let the container finish its in-flight messages before shutting down the network, and solve that.

PR #19275, opened 2021-09-24
Support overriding Vespa Cloud system

mpolden:
@oyving

### DIFF
--- a/client/go/cmd/helpers.go
+++ b/client/go/cmd/helpers.go
@@ -16,8 +16,6 @@ import (
 	"github.com/vespa-engine/vespa/client/go/vespa"
 )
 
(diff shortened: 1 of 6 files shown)


client/go/cmd/helpers.go
```diff
@@ -114,6 +112,23 @@ func getService(service string, sessionOrRunID int64) *vespa.Service {
 	return s
 }
 
+func getConsoleURL() string {
+	system := os.Getenv("VESPA_CLI_CLOUD_SYSTEM")
+	if system == "cd" {
```
oyving, line 117: Use `publiccd` for consistency, even though we don't expose it?

oyving (approved):

PR #19279, opened 2021-09-24
Add high-level test that maintenance is inhibited during pending state transitions

vekterli:
@geirst please review

Maintenance inhibition is already present, but it happens at a much lower level.
Add a high-level test to ensure that the wiring works as expected.


geirst (approved):
👍

PR #19281, opened 2021-09-24
Tolerate missing file references

bratseth:
If Vespa contains a config definition with a path, and this is
set by the application, this will cause a hash file reference,
which we should not attemt to resolve server side.


hmusum (approved):
I am not sure what else we can do when this happens, so let's try.

PR #19285, opened 2021-09-24
Shut down network when all owners also have detached

jonmv:
@baldersheim please review.

### DIFF
--- a/messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
+++ b/messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
@@ -10,7 +10,6 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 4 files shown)


baldersheim (approved):

messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
```diff
@@ -111,19 +112,21 @@ public void detach(NetworkOwner owner) {
...
+                                            "this automatically shuts down when detached from--or " +
+                                            "called multiple times on a shared multiplexer");
 
-        if ( ! owners.isEmpty())
-            log.warning("NetworkMultiplexer destroyed before all owners detached: " + this);
+        countDown();
+    }
 
-        net.shutdown();
+    private void countDown() {
```
gjoranv, line 127: Without knowing much about messagebus, this method does not seem to be counting down anything, so perhaps e.g. 'shutdownIfOwnerless' would be more descriptive?

jonmv: Yes, it would. Used a counter at first, but that wasn't expressive enough. PTAL.

gjoranv (commented):
👍

PR #19288, opened 2021-09-24
Use network in RPC service cache key to avoid getting dead services

jonmv:
@baldersheim please review.


baldersheim:
Solution without thread local in #19292

jonmv:
Great!

PR #19289, opened 2021-09-24
Create application on deploy in all systems

oyving:
Create an application object implicitly on deployment. Not sure if the instance level is really necessary, though. If not, I'll remove it before merge.

FYI - @ldalves


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2718,5 +2723,21 @@ private static boolean isOperator(HttpRequest request) {
...
 
+    private void ensureApplicationExistsForPublic(TenantAndApplicationId id, HttpRequest request) {
+        if (controller.system().isPublic() && controller.applications().getApplication(id).isEmpty()) {
+            log.fine("Application does not exist in public, creating: " + id);
+            var credentials = accessControlRequests.credentials(id.tenant(), null /* not used on public */ , request.getJDiscRequest());
+            controller.applications().createApplication(id, credentials);
+        }
+    }
+
+    private void ensureInstanceExistsForPublic(ApplicationId id, HttpRequest request) {
```
freva, line 2734: Not necessary, instance is created if it doesn't exist in `JobController::deploy`. Verified this already works by deploying to a new application without any instances.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2718,5 +2723,21 @@ private static boolean isOperator(HttpRequest request) {
                               .anyMatch(definition -> definition == RoleDefinition.hostedOperator);
     }
 
+    private void ensureApplicationExistsForPublic(TenantAndApplicationId id, HttpRequest request) {
+        if (controller.system().isPublic() && controller.applications().getApplication(id).isEmpty()) {
```
freva, line 2727: I think we can enable this for non-public as well. We do not send any body when creating an application control and `AthenzAccessControlRequests::credentials` only checks the body if a domain is not set on tenant

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
oyving: Alright, will remove.

freva (approved):
:100:

freva:
FYI: @mpolden CLI can now deploy without needing the user to create application in console first.

ldalves:
@bratseth ^^

bratseth:
Great! We should update the getting started then.

PR #19295, opened 2021-09-26
Balder/gc enforce rank profile inheritance

baldersheim:
@hmusum PR

### DIFF
--- a/container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
+++ b/container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
@@ -1,6 +1,7 @@
 // Copyright 2017 Yahoo Holdings. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.container.jdisc;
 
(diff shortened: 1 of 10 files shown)


container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
```diff
@@ -235,6 +239,11 @@ private void watchPortChange() {
         }
     }
 
+    void recofigure(QrConfig qrConfig) {
```
hmusum, line 242: Typo in method name

hmusum (approved):

container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
baldersheim: Thanks, fixed in #19293

baldersheim: Have to redo this one. The first part should not be there.

baldersheim:
This is superseded by #19296

PR #19304, opened 2021-09-27
Stateless REST API: short forms for sparse and mixed tensors 

lesters:
@bratseth Please review.


bratseth (changes requested):
I think the format here should be the same as in JsonFormat. And also be implemented there, with a parameter indicating what method to use.
I see some differences to what's there:
- This uses "values" regardless of the format used, while the jsonformat uses "values", "cells" or "blocks".
- This does not support the "blocks" short form with multiple sparse dimensions.
Perhaps we should change the JSON format (with backwards compatibility), but then it should be done also for all reading. Once we start to output this we're stuck so we should be sure we mean it.

I want this parameter to also work for result rendering (fields and summary features), and documents. Then it can't be named "format". How about "format.tensors"?

I suggest renaming "literal" to "string".

We should also run this by arch review.

lesters:
@bratseth Excellent points. I've added the values/cells/blocks format and support for multiple mapped dimensions for "blocks". I'll look into result rendering and documents.

`format.tensors` sounds good.

bratseth:
Great, I think it's fine to do results and documents later/separately if you prefer. As long as the parameter is free in all cases we're good.

bratseth (approved):

lesters:
Some restructuring is required to use `JsonFormat` from the `JsonRenderer` in container search. I'll start a separate PR for that.

PR #19308, opened 2021-09-27
Add grace period inhibiting maintenance after state transitions with bucket ownership transfer [run-systemtest]

vekterli:

Avoids the case where different distributors can start merges with a max timestamp
that is lower than timestamps generated intra-second by other distributors used for
feed bound to the same bucket.

This is analogous to the existing "safe time period" functionality used for
handling external feed, and uses the same max clock skew config as this.
Correctness of this grace period is therefore inherently dependent on actual
cluster clock skew being less than this configured number.

Bucket activations are still allowed to take place during the grace period time
window, as these do not mutate the bucket contents and are therefore safe.


toregge (approved):

geirst (approved):
👍

PR #19309, opened 2021-09-27
Revert "Use default value of jdisc.debug.resources."

jonmv:
Reverts vespa-engine/vespa#19303

@bjorncs @baldersheim this turned on debug references for all of CD 😝

### DIFF
--- a/configserver/src/main/sh/start-configserver
+++ b/configserver/src/main/sh/start-configserver
@@ -188,6 +188,7 @@ vespa-run-as-vespa-user vespa-runserver -s configserver -r 30 -p $pidfile -- \
 	-Djdisc.config.file=$cfpfile \
 	-Djdisc.export.packages= \
 	-Djdisc.cache.path=$bundlecachedir \
+	-Djdisc.debug.resources=false \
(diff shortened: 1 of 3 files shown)


hakonhall (approved):

bjorncs:
Are you sure it was not because of the related feature flag change?

jonmv:
Hmm, this wasn't _quite_ right, but the logservers were set to debugging mode, and they got so slow everything failed :)

bjorncs:
Yes, the feature flag affected all admin type of clusters. We could not restrict the flag to metricsproxy only.

hakonhall:
Couldn't you fetch with the CLUSTER_ID dimension set to the cluster name, i..e "metrics" for the metrics proxy?

It's a slight contract violation of CLUSTER_ID, as it is supposed to be `com.yahoo.config.provision.ClusterSpec.Id::value`, but as long as the serialized feature flag overrides matches whatever is done in code it should be fine.  We could update the CLUSTER_ID doc about this since the model and service monitor cares more about the cluster name (aka cluster id in service monitor) than the node's cluster spec ID.

Alternatively we could add a CLUSTER_NAME.

PR #19312, opened 2021-09-27
Only ask for stack trace if necessary

bjorncs:
Creating a stack trace is expensive and Throwable.getStackTrace()
should therefore only be invoked when required.


bjorncs:
Actually, we want to keep the current behaviour since we need the filled stack trace to be easily available in a heap dump

PR #19326, opened 2021-09-28
Expose aggregated low-level data movement statistics as metrics [run-systemtest]

vekterli:
@geirst please review.

Adds metrics for the following:

 * Bucket replicas that should be moved out, e.g. retirement case or
   node added to cluster that has higher ideal state priority.
 * Bucket replicas that should be copied out, e.g. node is in ideal state
   but might have to provide data other nodes in a merge.
 * Bucket replicas that should be copied in, e.g. node does not have a
   replica for a bucket that it is in ideal state for
 * Bucket replicas that need syncing due to mismatching metadata.

These are aggregates across all bucket replicas, buckets and bucket spaces.
Should aid in visibility for data movement during node retirements when there
are concurrent replicas out of sync events.


geirst (approved):
👍

PR #19341, opened 2021-09-29
Do not activate application when this is a dry run

hmusum:


jonmv (approved):
Do you also need to not reserve the nodes?

hmusum:
I am not sure if we can avoid that. I think we want to do this as close to possible as a real deployment, to test all aspects of deploying, but without requiring (much) more HW in cd.

PR #19345, opened 2021-09-29
Precompute host -> children mapping as it is frequently used and expe…

baldersheim:
…nsive to compute individually.

@bratseth or @mpolden or @hmusum PR

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodesAndHosts.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodesAndHosts.java
@@ -1,0 +1,58 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.provision;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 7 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodesAndHosts.java
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Return the NodeList used for construction
+    public NL nodes() { return nodes; }
+
+    public NodeList childrenOf(Node host) {
+        HostAndNodes hostAndNodes = host2Nodes.get(host.hostname());
+        return hostAndNodes != null ? NodeList.copyOf(hostAndNodes.children) : NodeList.of();
+    }
+
+    public Node parentOf(Node node) {
```
mpolden, line 41: Can we make this return `Optional<Node>`? So it's obvious to callers that it can be empty.

mpolden (commented):
Good idea!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodesAndHosts.java
baldersheim: Agree, done.

PR #19352, opened 2021-09-29
Add tensor short form format option for result rendering

lesters:
For field and summary rendering. Requires `?format.tensors=short` option to activate. 

@bratseth Please review.


bratseth (approved):
Nice

PR #19366, opened 2021-09-30
Introduce new job profile for dev dry run

hmusum:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobProfile.java
```diff
@@ -68,7 +68,10 @@
 
     development(EnumSet.of(deployReal,
                            installReal,
-                           copyVespaLogs));
+                           copyVespaLogs)),
+
+    developmentDryRun(EnumSet.of(deployReal,
+                                 copyVespaLogs));
```
jonmv, line 74: Won't be any logss to copy either, since there will never be a logserver.

jonmv (commented):
Yess, this is a good iea :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobProfile.java
hmusum: Ah, good point, will fix

PR #19376, opened 2021-09-30
Bjorncs/scale threadpools locally

bjorncs:
FYI @bratseth

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/SearchHandler.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/SearchHandler.java
@@ -51,7 +51,7 @@ class SearchHandler extends ProcessingHandler<SearchChains> {
             if (cluster.isHostedVespa()) {
                 builder.maxThreads(-2).minThreads(-2).queueSize(-40);
             } else {
-                builder.maxThreads(500).minThreads(500).queueSize(0);
+                builder.maxThreads(-4).minThreads(-4).queueSize(-40);
             }
         }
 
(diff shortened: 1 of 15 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/xml/SearchHandler.java
```diff
@@ -48,22 +48,10 @@ public void getConfig(ContainerThreadpoolConfig.Builder builder) {
...
-                // Controls max number of concurrent requests per container
-                int workerThreads = Math.max(8, (int)Math.ceil(vcpu * 2.0));
-                builder.maxThreads(workerThreads);
-                builder.minThreads(workerThreads);
-
-                // This controls your burst handling capability.
-                // 0 => No extra burst handling beyond you max concurrent requests (maxthreads).
-                // N => N times max concurrent requests as a buffer for handling bursts
-                builder.queueSize((int)(workerThreads * 40.0));
+                builder.maxThreads(500).minThreads(500).queueSize(0);
```
baldersheim, line 54: Use -4 for non hosted, but same approach.

container-core/src/main/java/com/yahoo/container/handler/threadpool/DefaultContainerThreadpool.java
```diff
@@ -74,24 +86,26 @@ private void closeInternal() {
...
+        else threads = Math.abs(config.minThreads()) * cpus;
+        return Math.min(threads, max);
     }
 
-    private static int computeCoreThreadPoolSize(int corePoolSize, int maxNumThreads) {
-        return Math.min(
-                corePoolSize <= 0 ? Runtime.getRuntime().availableProcessors() * 2 : corePoolSize,
-                maxNumThreads);
+    private int queueSize(ContainerThreadpoolConfig config, int maxThreads) {
+        return config.queueSize() >= 0 ? config.queueSize() : Math.max(650, Math.abs(config.queueSize()) * maxThreads);
```
baldersheim, line 108: Use a named constant that gives a hint to purpose.

baldersheim (changes requested):
Fix the comments and you should be good.

bjorncs:
@baldersheim PTAL

baldersheim (approved):

PR #19379, opened 2021-09-30
Don't let a blocked maintenance operation inhibit remaining maintenance queue [run-systemtest]

vekterli:
@geirst please review
@toregge FYI

The old maintenance scheduler behavior is to only remove a bucket from the
priority DB if its maintenance operation was successfully started. Failing
to start an operation could happen from both max pending throttling as well
as operation/bucket-specific blocking behavior. Since the scheduler would
encounter the same bucket as the one previously blocked upon its next tick
invocation, a single blocked bucket would run the risk of head-of-line
stalling the rest of the remaining maintenance queue (assuming the ongoing
DB scan did not encounter any higher priority buckets).

This commit changes the following aspects of maintenance scheduling:
 * Always clear entries from the priority DB before trying to start an
   operation. A blocked operation will be retried the next time the
   regular bucket DB scan encounters the bucket.
 * Avoid trying to start (and clear) inherently doomed operations by
   _not_ trying to schedule any operations if it would be blocked due
   to too many pending maintenance operations anyway. Introduces a
   new `PendingWindowChecker` interface for this purpose.
 * Explicitly inhibit all maintenance scheduling if a pending cluster
   state is present. Operations are already _implicitly_ blocked from
   starting if there's a pending cluster state, but this would cause
   the priority DB from being pointlessly cleared.

### DIFF
--- a/storage/src/vespa/storage/distributor/maintenance/pending_window_checker.h
+++ b/storage/src/vespa/storage/distributor/maintenance/pending_window_checker.h
@@ -1,0 +1,19 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include <vespa/storage/distributor/operationstarter.h>
+
+namespace storage::distributor {
+
+/**
+ * Allows for easily and cheaply checking if an operation with a given internal maintenance
+ * priority could possibly be started "downstream" due to there being available capacity
+ * in the maintenance pending window.
+ */
+class PendingWindowChecker {
+public:
+    virtual ~PendingWindowChecker() = default;
+    [[nodiscard]] virtual bool may_allow_operation_with_priority(OperationStarter::Priority) const noexcept = 0;
+};
+
+}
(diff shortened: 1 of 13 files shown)


storage/src/vespa/storage/distributor/maintenance/pending_window_checker.h
```diff
@@ -0,0 +1,14 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include <vespa/storage/distributor/operationstarter.h>
+
+namespace storage::distributor {
+
+class PendingWindowChecker {
```
geirst, line 8: Consider adding comment.

geirst (approved):
👍  @toregge please review as well

storage/src/vespa/storage/distributor/maintenance/pending_window_checker.h
vekterli: Will do this in a follow-up

toregge (approved):

toregge (approved):

geirst (approved):
👍

PR #19398, opened 2021-10-01
use saturated sum as hit estimate for OR

havardpe:
The upper bound for the hit estimate is the docid limit. The lower
bound for the docid limit is the maximum child estimate.

In the future we probably want a solution where everyone knows the
docid limit from the start, which might also require us to have a
separate way of estimating the document frequency if we want to avoid
ranking score regression.

@baldersheim @arnej please review


arnej27959 (approved):
👍

baldersheim (approved):

PR #19400, opened 2021-10-01
Fix importing complex fields from parents using inheritance

geirst:
Do not replace existing struct fields when populating a field.

This fixes a problem where a type inheriting another type would replace (and remove)
the struct field attributes (of complex fields) of the base type.
This made it impossible to import the base complex field (using parent/child) as the struct field attributes no longer existed.

@baldersheim please review

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
@@ -293,7 +293,7 @@ public class SDField extends Field implements TypedKey, FieldOperationContainer,
                 for (Field field : subType.fieldSet()) {
                     SDField subField = new SDField(sdoc, name.concat(".").concat(field.getName()), field.getDataType(),
                                                    subType, new Matching(), true, recursion + 1);
-                    structFields.put(field.getName(), subField);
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
```diff
@@ -293,7 +293,9 @@ public void populateWithStructFields(SDDocumentType sdoc, String name, DataType
                 for (Field field : subType.fieldSet()) {
                     SDField subField = new SDField(sdoc, name.concat(".").concat(field.getName()), field.getDataType(),
                                                    subType, new Matching(), true, recursion + 1);
-                    structFields.put(field.getName(), subField);
+                    if (!structFields.containsKey(field.getName())) {
```
baldersheim, line 296: Use putIfAbsent instead.

baldersheim (changes requested):
put -> putIfAbsent is better.

baldersheim (approved):

config-model/src/main/java/com/yahoo/searchdefinition/document/SDField.java
geirst: Thanks, fixed.

PR #19424, opened 2021-10-05
Do not busy-bounce merges forwarded from other nodes

vekterli:
@geirst please review
@baldersheim FYI

Let any merge through that has already passed through at least one
other node's merge window, as that has already taken up a logical
resource slot on all those nodes. Busy-bouncing a merge at that point
would undo a great amount of time already expended.

The max number of enqueued merges is bounded by the number of nodes
in the system, as each node can still only accept a configurable number
of merges from distributors, and each distributor throttles their
maintenance operations based on priority-relative max pending limits.

This feature can be toggled with a live reconfig. Currently defaults to false,
i.e. queue limits are still enforced.

### DIFF
--- a/storage/src/vespa/storage/config/stor-server.def
+++ b/storage/src/vespa/storage/config/stor-server.def
@@ -45,6 +45,14 @@ max_merge_queue_size int default=1024
 ## "source only", as merges do not cause mutations on such nodes.
 resource_exhaustion_merge_back_pressure_duration_secs double default=30.0
 
+## If true, received merges that have already been accepted into the pending
+## merge window on at least one node will not be restricted by the configured
+## max_merge_queue_size limit. They will be allowed to enqueue regardless of
+## the current queue size. This avoids wasting the time spent being accepted
+## into merge windows, which would happen if the merge were to be bounced with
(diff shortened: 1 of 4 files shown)


storage/src/vespa/storage/config/stor-server.def
```diff
@@ -45,6 +45,14 @@ max_merge_queue_size int default=1024
 ## "source only", as merges do not cause mutations on such nodes.
 resource_exhaustion_merge_back_pressure_duration_secs double default=30.0
 
+## If true, received merges that have already been accepted into the pending
+## merge window on at least one node will not be restricted by the configured
+## max_merge_queue_size limit. They will be allowed to enqueue regardless of
+## the current queue size. This avoids wasting the time spent being accepted
+## into merge windows, which would happen if the merge were to be bounced with
+## a busy-reply that would subsequently be unwound throuh the entire merge chain.
```
geirst, line 53:
```suggestion
## a busy-reply that would subsequently be unwound through the entire merge chain.
```

geirst (approved):
👍

geirst:
@toregge FYI

PR #19430, opened 2021-10-06
Add non-primitive tensor expand function

lesters:
Adds the non-primitive `expand` function to add a dimension to a tensor. Today, to add the dimension `x` to a tensor `T` you need to do the following:

```
T * tensor(x[1])(1)
```

With this change:

```
expand(T, x)
```

@bratseth Please review.
@jobergum FYI.


bratseth (approved):
nice!

PR #19439, opened 2021-10-06
Arnej/wire errors from backend

arnej27959:

for discussion: do we want new "codes" (extending the enum in container-core)?
do we want to use this instead of getting errors inside slime container (for docsum reply, currently both)?

@havardpe
@bratseth


havardpe (approved):
I have no idea what the semantics of RESULT_HAS_ERRORS is.
We could consider a back-end evaluation/execution error to match the back-end communication error we already have.

other than that; looks good (and simple)

bratseth (approved):
This looks good to me. I don't think those error codes have ever paid their rent, so just reusing somewhat generic ones like this seems optimal.

PR #19460, opened 2021-10-07
If execution is rejected, fall back to doing it yourself.

baldersheim:
@bratseth or @bjorncs PR


bratseth (commented):
When do you think we'll need this?

bratseth:
So, you are saying you want to use a bounded thread pool here now?

bratseth:
Thanks.

PR #19468, opened 2021-10-07
- Use the default threadpool for federation.

baldersheim:
- Move test classes into the same package as the code to avoid making test only methods public.
- Deprecate old AsyncExecution.search/fill methods and let them run in its own thread.
  There should be no external usage of this class.

@bratseth PR
The essential changes are in AsyncExecution/FederationSearcher. The rest are cleanup and api changes in the tests.


bratseth:
Not reviewed yet (in a meeting), and not sure what you meant above, but AsyncExecution is definitely a public API with valid use cases. 

For example, construct two different queries doing different things and run them in parallel. All this can be done using federation but it's more to learn and if you don't need the composability it may not be worth the effort.

bjorncs (approved):

baldersheim (commented):
Yes, much better.

PR #19472, opened 2021-10-07
Do not embed artifacts provided by Jdisc container [run-systemtest]

gjoranv:
- Reduces bundle size from 13.3 to 3.4 MB

FYI: @freva


baldersheim (approved):

bjorncs (commented):
🥇

PR #19476, opened 2021-10-08
Control disk-bloat-factor and compression level with feature flag.

baldersheim:
@geirst PR
@yngveaasheim @jobergum FYI


geirst (approved):
👍

PR #19481, opened 2021-10-08
extend element size to at least contain all matches

havardpe:
Increase element size rather than reduce the number of matches. This
is to avoid getting scores over 1 for cases where we have matches
outside the element. We do not know if or when this might happen, but
reversing the guard makes it more robust.

@arnej27959 please review
@bratseth FYI


arnej27959 (approved):
👍

PR #19486, opened 2021-10-08
Balder/resolve executor

baldersheim:
@bratseth or @bjorncs or @hmusum PR


hmusum (approved):

baldersheim:
Yes, and no. We lost one threadpool on the way.
Fixed size threadpool for the SearhHandler as before.
Federation and ClusterSearcher now use the same "default" threadpool with similar settings as
the one use by federation. And they are all controlable by settings in services in overrides are necessary.
Which I hope there will be no need for.

PR #19489, opened 2021-10-09
No need to create a temporary execution, that is discarded right afte…

baldersheim:
…rwards.

The execution creates and empty context with a single threaded executor. That is not free.

@bratseth or @bjorncs


bratseth (commented):
Yes, but now the API this can't guarantee it is not returning a null execution. I'll try to fix it without the cost somehow ...

baldersheim:
What about using the InThreadExecutorService ? That has no other cost than the object.

bratseth:
Yes, I did that first (or really just Runnable::run), but some tests hung. I'll look into it more.

bjorncs (commented):
👍

PR #19504, opened 2021-10-11
report issues instead of logging warnings in some features

arnej27959:

@havardpe please review - not sure if we can just merge this yet?


havardpe (commented):
I found one issue that should probably stay a log, otherwise this looks good.
I am not sure how many system tests this will break, though...

PR #19507, opened 2021-10-11
Expose more content node merge metrics.

geirst:
@baldersheim please review and merge
@toregge @vekterli FYI


baldersheim (changes requested):
I think you should expose sum too so that average can be extracted too from sum/count.

geirst:
Done. @baldersheim PTAL

baldersheim (approved):

PR #19511, opened 2021-10-12
We should have same default for both final chunk compression, and tem…

baldersheim:
…porary compact compression.

@geirst PR


geirst (approved):
👍

PR #19520, opened 2021-10-12
A group hit is not meta since it contains result data

bratseth:
The only functional change here is in the Group constructor.


kkraune (approved):

arnej27959 (commented):
👍

PR #19522, opened 2021-10-13
contributing vespaclient in go by <Get.It>

renevall:

## Vespa Go Client Proposal

For the past 8-12 months we have been building a Go Client for Vespa that we planned to share/donate/open source to Vespa and its community. We believe this is only a fraction of the work to be done, but it has served us well in production to date and we believe it is something we can build on top of.  We are proud to say that despite any potential shortcoming, anyone could use it in the current state to work with vespa's `/api/v1/document` api. 

The current implementation was inspired by [olivere/elastic](https://github.com/olivere/elastic) library, where a key element of the API is method chaining to describe a request.  This does conflict with the regular way go does error checking and bubbling technique, but in exchange, the API to our eyes seemed straightforward to follow. The errors are still properly exposed and can be handled by the end-user. 

The most basic use-case for search would look like:

```go
var opts []vespa.ClientOptionFunc
connStr := "http://localhost:8080"
opts = append(opts,
  vespa.SetURL(connStr),
)

client, err := vespa.NewClient(ctx, opts...)
if err != nil {
  panic(err)
}

res, err := vespa.NewSearchService(client).
  Scheme("docs").
  Query("Trucking Class B").
  Where("userQuery() AND status contains \"active\" AND geoLocation(location, 33.64, -84.33, \"25mi\") AND hidden = false").
  Timeout("5s").
  Ranking(vespa.Ranking{Profile: "bm25"}).
  Do(ctx)
```

The current implementation supports the following:
- Create
- Read
- Update
- Partial Update
- Delete
- Search
- Batch Create, Update. 
- HTTP2 support through official http2 implementation

One of the known shortcomings is that the Batch process sends one request per document, essentially making it a fake batch operation. What it does provide instead is an abstraction to a batch operation, where each request launches its own routine and they are synchronized again with the use of a wait-group. Good enough for anyone to use and get up to seen 3000qps for feed operations.

As stated, we don't believe this implementation is perfect by any means, but it has done the job for us for a while. We hope this becomes either something to build on or inspiration for the official library in the future. 

Thanks all,

Get.it Team

### DIFF
--- a/client/go/Makefile
+++ b/client/go/Makefile
@@ -12,7 +12,7 @@ BIN ?= $(CURDIR)/bin
 SHARE ?= $(CURDIR)/share
 DIST ?= $(CURDIR)/dist
 
-GO_FLAGS := -ldflags "-X github.com/vespa-engine/vespa/client/go/build.Version=$(VERSION)"
+GO_FLAGS := -ldflags "-X github.com/vespa-engine/client/go/api/build.Version=$(VERSION)"
 GIT_ROOT := $(shell git rev-parse --show-toplevel)
 DIST_TARGETS := dist-mac dist-mac-arm64 dist-linux dist-win32 dist-win64
 
--- a/client/go/api/.gitignore
+++ b/client/go/api/.gitignore
@@ -1,0 +1,2 @@
+client.crt
+client.key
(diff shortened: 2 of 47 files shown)


renevall:
thanks @mpolden. I'm not sure how I missed that. I recall amending the commit. Could you check again, I left only client/go/api folder.

aressem:
@bratseth This can probably be closed?

bratseth:
@mpolden ?

PR #19556, opened 2021-10-14
Add metric for max time since bucket GC was last run

vekterli:
@geirst please review
@yngveaasheim FYI

Max time is aggregated across all buckets. If this metric value grows
substantially larger than the configured GC period it indicates that
GC is being starved.

Currently tracked as a single global value across all bucket spaces.


geirst (approved):
👍

PR #19581, opened 2021-10-15
Add class representing async state for applying bucket diff to local node.

toregge:
@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #19582, opened 2021-10-15
Container config improvements take2

gjoranv:
The first attempt failed because of a simplification around component subscribers that was unrelated to the main idea. The fix is in the second commit. The last commit is a cherry-pick of your "throw upon bad config", which this is built around, so we might as well take it here.

Depends on https://github.com/vespa-engine/system-test/pull/1927

FYI: @jonmv


hmusum (approved):
👍

PR #19583, opened 2021-10-15
Validate summary inheritance

bratseth:


bratseth:
Or @hmusum

baldersheim (changes requested):
Remove debug statement, and it should be good to go.

PR #19594, opened 2021-10-15
Only block bucket remove operation if overlapping with pending message target node [run-systemtest]

vekterli:
@geirst please review
@toregge FYI

Previously we'd block if there were pending messages to _any_ replica
node for the bucket, which is a massive pessimization. This behavior
also made it likely that a merge operation with one or more source-only
replicas would end up being tagged as failed due to concurrent client
load to the other replicas.


geirst (approved):
👍

PR #19595, opened 2021-10-15
the new protobuf protocol should have good performance now

arnej27959:

For discussion:

We want to get rid of the "extra" docsum fetching protocol.  From the code it looks like the default is to use the new protobuf protocol always, but I don't know if some customers override these defaults, and (if so) if there's any performance gain to get from it.

@baldersheim please review
@havardpe


baldersheim (approved):
I think it is fine to remove this one now.

arnej27959:
Merging now to see if any performance tests pick it up.

PR #19602, opened 2021-10-18
Warn upon undeclared components

gjoranv:
This will most likely generate log warnings for some user applications, on startup and every redeploy until the issue is fixed on their side.


baldersheim (approved):

gjoranv:
FYI: @hmusum

PR #19605, opened 2021-10-18
Add guard against lockstep wait in ApplyBucketDiffState::wait().

toregge:
Add guard against throwing exception out of ApplyBucketDiffState destructor.

@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #19613, opened 2021-10-18
Move MonitoredRefCount and RetainGuard to vespalib.

toregge:
@baldersheim : please review
@geirst , @vekterli : FYI


baldersheim (approved):
Nice, these I were not aware of....

PR #19614, opened 2021-10-18
Set permissions afterwards so it's not affected by umask

bjorncs:


freva (approved):

hakonhall (approved):

freva:
Does it need to be this open? Could we limit this to group?

PR #19622, opened 2021-10-18
add ITEM_TRUE and ITEM_FALSE query nodes

arnej27959:
@havardpe flushing today's work, not sure if it is really ready for review...


havardpe (approved):
consider using same name for abstract and custom templated node types.

havardpe (approved):

PR #19625, opened 2021-10-18
Only keep async variant to simplify what to implement and what fallba…

baldersheim:
…ck there are.

@vekterli PR


vekterli (approved):
:+1:

PR #19636, opened 2021-10-19
Change editMergeStatus to return a shared pointer to merge status

toregge:
to ensure it remains live while being used in merge handler.

@geirst : please review


geirst (approved):
👍

PR #19652, opened 2021-10-20
empty hostname means "localhost"

arnej27959:
* avoid core-dump with short RPC spec

@havardpe please review
FYI: I tried `vespa-rpc-invoke tcp/19099 ...` and got a crypto exception because of empty hostname.


havardpe (commented):
I think we want to solve this one in a different way (PR coming soon)

PR #19654, opened 2021-10-20
add forward_issues config option

arnej27959:
* default true, can be set to false to can disable sending
  issues from backend to QRS
* issues not send to QRS are silently discarded.

@havardpe for review/discussion


havardpe (approved):
consider logging dropped issues and making the forward flags atomic

havardpe (approved):

arnej27959:
@geirst please take a look also

geirst (approved):
👍

PR #19656, opened 2021-10-20
Remove config generation at :19050/clustercontroller-status/v1/<clustername>/state

hakonhall:
On an example node:

```
vespa-curl-wrapper https://localhost:19050/clustercontroller-status/v1/<clustername>/state
{
  "status" : {
    "code" : "up"
  },
  "config" : {
    "component" : {
      "generation" : -1
    }
  }
}
```

From the code, the generation may be either -1 or 0.  At no point does it show the config generation of any component.  It's unlikely anyone depends on this behavior.  So this PR removes it and we'll see how it goes...  If a client wants the config generation they should use the standard /state/v1/config.


vekterli (approved):
:+1:

PR #19658, opened 2021-10-20
WeakAnd has no field

bratseth:


arnej27959 (approved):
if possible, remove fieldname in the constructor also

PR #19664, opened 2021-10-20
Jonmv/streaming doc v1 visit

jonmv:
@bjorncs please review. Exact query parameters etc. to be decided.


jonmv:
Hang on ... let me ignore wantedDocumentCount when streaming mode is on.

bjorncs (approved):
See comment regarding executor

bjorncs (approved):

PR #19671, opened 2021-10-21
Delay replies for async apply bucket diff.

toregge:
@geirst : please review


geirst (approved):
👍  But @vekterli should review this as well.

vekterli (approved):
:+1:

PR #19689, opened 2021-10-21
Revert "Get a truly unique prefix per cluster searcher globally."

aressem:
Reverts vespa-engine/vespa#19684

Broke system test https://github.com/vespa-engine/system-test/blob/master/tests/search/explain/explain.rb


baldersheim:
Tests needs an update. Will fix.

PR #19690, opened 2021-10-21
Revert "If the uncommitted changes pass 128k for an attribute force a commit."

aressem:
Reverts vespa-engine/vespa#19675

Several performance tests outside their limits


baldersheim:
Changes expected and intentional.

PR #19692, opened 2021-10-22
Add a metric for how many times a worker in a thread pool has woken up.

baldersheim:
Also track the idle time a worker has and add metric for the utilization.

@havardpe and @geirst PR
Note the TODO in threadstackexecutorbase.cpp


geirst (commented):
👍 for the new metrics. @havardpe must review the rest.

PR #19709, opened 2021-10-22
Fix MergeStatus::check_delayed_error: Empty fail message is not an error.

toregge:
@geirst : please review


geirst (approved):
👍

PR #19712, opened 2021-10-25
Log version to complete remote task

hakonhall:
Normally, if a SetNodeStateRequest changes the state of a node,
scheduleVersionDependentTasksForFutureCompletion(FleetController.java:1003)
will ensure that the request waits for the successful publication of the next
cluster state version before returning 200.

There are reasons to believe there is an edge case, likely triggered by losing
the ZooKeeper connection just prior to trying to set the new wanted state in
ZK, that makes scheduleVersionDependentTasksForFutureCompletion() complete the
request at the current version.

This PR will make it possible to prove or disprove the theory.


vekterli (approved):
:+1:

PR #19714, opened 2021-10-25
Add metering metrics

olaaun:


oyving (approved):
LGTM. I think maybe we want to add metrics for the cost as well, but we can do that in a separate PR. Can we make sure these metrics are only sent to the `vespa-cloud` namespace?

PR #19725, opened 2021-10-25
Prioritize forwarded merges in MergeThrottler queue

vekterli:
@geirst please review
@baldersheim @toregge FYI

Rationale: merges that already are part of an active merge window
are taking up logical resources on one or more nodes in the cluster
and we should prefer completing these before starting new merges
queued from distributors.


baldersheim (approved):

geirst (commented):
👍

PR #19735, opened 2021-10-26
Handover tracker to ApplyBucketDiffState on exceptions.

toregge:
@geirst : please review


vekterli (approved):
:+1: the implicit data flow is getting a bit hard to track, but to make things a lot better I think we're going to need a more fundamental redesign of the merge handler (and probably the persistence layer in general)

geirst (approved):
👍

PR #19747, opened 2021-10-27
Bratset/items equality

bratseth:


arnej27959 (approved):
I think this looks good

PR #19753, opened 2021-10-27
Adjust dummy persistence spi semantics towards proton spi semantics when

toregge:
bucket doesn't exist:
  setActiveState(), put(), remove() creates bucket if it doesn't already exist.

@baldersheim : please review


baldersheim (approved):

vekterli (approved):
:+1:

PR #19792, opened 2021-10-29
Bratseth/stable query profile ids

bratseth:
This is to avoid changing the query profiles config unnecessarily on redeploy.


bratseth:
Or @baldersheim

baldersheim (approved):

PR #19796, opened 2021-10-29
JDK17: Depend on latest groovy and use latest gmavenplus-plugin.

gjoranv:


mpolden (approved):

mpolden:
Groovy?! We should rewrite this.

PR #19801, opened 2021-10-31
Update vespa version in pom files as part of preparing for build.

toregge:
@aressem : please review

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -493,7 +493,7 @@ case '%{version}' in
 	:
 	;;
     *)
-	sed -i -e 's,<version>7-SNAPSHOT</version>,<version>%{version}</version>,' $(find . -name pom.xml -print)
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -487,6 +487,15 @@ nearest neighbor search used for low-level benchmarking.
 %setup -c -D -T
 %else
 %setup -q
+echo '%{version}' > VERSION
+case '%{version}' in
+    *.0)
+	:
+	;;
+    *)
+	sed -i -e 's,<version>7-SNAPSHOT</version>,<version>%{version}</version>,' $(find . -name pom.xml -print)
```
aressem, line 496: Maybe match [0-9].*-SNAPSHOT here as no versions should be snapshot anyway ?

aressem (approved):

aressem (approved):

PR #19810, opened 2021-11-01
Require that you have at least 1G per thread as each of them can requ…

baldersheim:
…ire quite a lot during expression parsing.

@hmusum PR


hmusum (changes requested):
Seems like unit tests fail due to numThreads being zero. Make sure it always is >= 1?

hmusum (approved):

PR #19824, opened 2021-11-02
added new rank feature "onnx(model_name)"

havardpe:
Works the same as "onnxModel(model_name)", but is not treated as the
exact same feature (features are currently not allowed to have
multiple base names). If both variants are used at the same time, the
model may be calculated twice, but the model cache will still make
sure that the model itself is only loaded once.

The plan is to deprecate and possibly remove the
"onnxModel(model_name)" variant at some point in the future.

@lesters please review


lesters (approved):
LGTM!

PR #19834, opened 2021-11-02
Avoid keeping string around in the leaf nodes. It is not worth the sp…

baldersheim:
…ace.

@bratseth PR


bratseth (approved):

baldersheim:
Superceeded by #19850

PR #19850, opened 2021-11-03
Avoid keeping string around in the leaf nodes. It is not worth the sp…

baldersheim:
…ace.

@bratseth or @hmusum PR


bratseth (approved):

hmusum (changes requested):
LGTM, but need to fix some unit tests (`toString()` uses `sourceString()`)

hmusum (approved):

PR #19855, opened 2021-11-03
Avoid intermediate NegativeNode in the leaf nodes, adding approximate…

baldersheim:
…ly 15-20% extra nodes.

@bratseth PR


bratseth (commented):
👍

PR #19858, opened 2021-11-04
I think that the active initial push of files are gone, and should no… [run-systemtest]

baldersheim:
…t be needed anymore.

@hmusum PR


hmusum (commented):
`FileSender.send()` is not a good name, it does not send files anymore, so that part looks OK. But that method also adds files to FileReferencesRepository. So users of that will probably need to be updated/rewritten as well?

baldersheim:
Ah, so this associates reference with service. Must be rethought then.
Nothing broke though...

PR #19861, opened 2021-11-04
Run some config server maintainers without maintanence job lock

hmusum:
Some maintainers only operate on local data and can be run
simultaneously across servers (ApplicationPackageMaintainer and
FileDistributionMaintainer changed to do this in this PR) without
distributed locking


hmusum:
ping @baldersheim

baldersheim (approved):

PR #19865, opened 2021-11-04
add "match-features" syntax

arnej27959:
* based on the way "summary-features" is handled
* also add "rank-features" some places where it seems to be missing

@bratseth please review (and test in your IntelliJ)
@shahariel FYI


bratseth (approved):
👍

PR #19892, opened 2021-11-05
Just fetch file references directly from fileregistry as there is no …

baldersheim:
…need to assiciate them with services.

@hmusum PR


hmusum (approved):
Great!

PR #19903, opened 2021-11-08
let match-features inherit also

arnej27959:
* a rank-profile can now inherit its parent's match-features and
  add its own, exactly the same way as summary-features.

@bratseth please review
@havardpe FYI


bratseth (approved):

config-model/src/main/java/com/yahoo/searchdefinition/RankProfile.java
```diff
@@ -519,8 +521,30 @@ public void setInheritedSummaryFeatures(String parentProfile) {
         this.inheritedSummaryFeatures = parentProfile;
     }
 
+    /**
+     * Sets the name of a profile this should inherit the match features of.
+     * Without setting this, this will either have the match features of the parent,
+     * or if match features are set in this, only have the match features in this.
+     * With this set the resulting match features of this will be the superset of those defined in this and
+     * the final (with inheritance included) summary features of the given parent.
```
havardpe, line 529: should be 'match' instead of 'summary' here as well

PR #19919, opened 2021-11-08
Use alternate executor id for push stage when sharing sequenced task …

toregge:
…executor with invert stage.

@geirst : please review
@baldersheim : FYI


baldersheim (approved):

geirst (commented):
👍

PR #19922, opened 2021-11-09
Use last access time to find file references in use [run-systemtest]

hmusum:
File references being used by active applications will never be deleted.
Use last access time instead of last modified time to find which ones to
delete or keep for the rest.


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -688,15 +688,15 @@ private Application getApplication(ApplicationId applicationId, Optional<Version
                 .collect(Collectors.toList());
     }
 
-    private boolean isFileLastModifiedBefore(File fileReference, Instant instant) {
-        return lastModified(fileReference).isBefore(instant);
+    private boolean isFileAccessedBefore(File fileReference, Instant instant) {
```
baldersheim, line 691: isLastFileAccessBefore is better.

hmusum: Agreed, fixed

baldersheim (approved):

PR #19927, opened 2021-11-09
Let default watermark be at 50% instead of 10%.

baldersheim:
That will favour more frequent wakeups, and should give more stable flow.

@geirst PR


geirst (approved):
👍

PR #19928, opened 2021-11-09
Validate system can set up cloud secret store

tokle:


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -281,6 +281,8 @@ private void addSecretStore(ApplicationContainerCluster cluster, Element spec, D
 
     private void addCloudSecretStore(ApplicationContainerCluster cluster, Element secretStoreElement, DeployState deployState) {
         if ( ! deployState.isHosted()) return;
+        if ( ! cluster.getZone().system().isPublic())
+            throw new RuntimeException("cloud secret store is not available in this zone");
```
oyving, line 285: The error message should be something about using ckms instead for non-public?

oyving (changes requested):

oyving (approved):

PR #19933, opened 2021-11-09
Revert "Extract field value in attribute writer thread."

toregge:
Reverts vespa-engine/vespa#19926

It caused some system tests to generate core dumps.


geirst (approved):
👍

PR #19934, opened 2021-11-09
Prepare for extracting field value in attribute writer thread.

toregge:
This contains most of the refactoring from #19926, but still extracts field values in document db main thread.


geirst (approved):
👍

PR #19935, opened 2021-11-09
Bratseth/schema summary inheritance

bratseth:


config-model/src/main/java/com/yahoo/searchdefinition/processing/ImplicitSummaries.java
```diff
@@ -29,17 +29,16 @@ public ImplicitSummaries(Schema schema, DeployLogger deployLogger, RankProfileRe
 
     @Override
     public void process(boolean validate, boolean documentsOnly) {
-        DocumentSummary defaultSummary = schema.getSummary("default");
+        DocumentSummary defaultSummary = schema.getSummariesInThis().get("default");
         if (defaultSummary == null) {
             defaultSummary = new DocumentSummary("default", schema);
-            defaultSummary.setFromDisk(true);
+            defaultSummary.setFromDisk(true); // TODO: Not necessarily
```
baldersheim, line 35: It is necessary I think as you get the document id with default. So it will access disk.

config-model/src/main/java/com/yahoo/searchdefinition/processing/ImplicitSummaries.java
```diff
@@ -114,7 +113,7 @@ private void collectSummaries(SDField field , Schema schema, boolean validate) {
     }
 
     private DocumentSummary getOrCreateAttributePrefetchSummary(Schema schema) {
-        DocumentSummary summary = schema.getSummary("attributeprefetch");
+        DocumentSummary summary = schema.getSummariesInThis().get("attributeprefetch");
```
baldersheim, line 116: attributeprefetch not dead yet.....

vespajlib/src/test/java/com/yahoo/tensor/TensorTestCase.java
```diff
@@ -41,6 +41,11 @@ public void testStringForm() {
                      Tensor.from("{ {d1:'[[\":\"]]'}: 5, {d1:\"'''\"}:6.0 }").toString());
     }
 
+    @Test
+    public void testTmp() {
+        System.out.println(Tensor.from("tensor(x[3]):[1, 2, 3]").join(Tensor.from("tensor(y[3]):[4, 5, 6]"), (x,y) -> x*y).map(v -> v*3));
```
baldersheim, line 46: Intentional ?

baldersheim (changes requested):

vespajlib/src/test/java/com/yahoo/tensor/TensorTestCase.java
bratseth: no, thanks

config-model/src/main/java/com/yahoo/searchdefinition/processing/ImplicitSummaries.java
bratseth: Somebody needs to work on that and the "default" nonsense if we're to make it in time for Vespa 8 ...

config-model/src/main/java/com/yahoo/searchdefinition/processing/ImplicitSummaries.java
bratseth: True

PR #19937, opened 2021-11-09
Compression is not supported for a very long time.

baldersheim:
So drop lazy decompression support in favor of thread safe access to field values.

@vekterli or @toregge PR
@geirst FYI


toregge (approved):

vekterli (approved):
:+1: the more mutability and complexity we can remove from the document module, the better

PR #19938, opened 2021-11-09
Revert "Prepare for extracting field value in attribute writer thread."

bratseth:
Reverts vespa-engine/vespa#19934

By the elimination method I think this may have broken a number of struct system tests.


freva (approved):

toregge:
Thanks.

PR #19947, opened 2021-11-10
remove entire "statistics" java module since it is now useless

arnej27959:
* all useful functionality is removed, so code using this module
  will just silently stop working
* we no longer use it in our own code
* these classes predate "jdisc" and are probably not used by
  anyone else either; but if there is some old code out there
  using it, that code must now be rewritten to use new "metrics"
  jdisc framework instead.

@bratseth please review
@hmusum @baldersheim FYI


baldersheim:
The Statistics object is still part of a few public interfaces (injection) so I doubt this will go in just like this.

baldersheim:
Superceeded by #22320

PR #19952, opened 2021-11-10
Remove implicit write to tmp as part of cluster controller health ping

vekterli:
@geirst please review

This was a feature tracing its lineage back to ye olde VDS days where we
had a dusty root drive holding the OS and Vespa and a dozen spinning rust
drives storing the actual data. The health ping would confirm that the
root drive was still functioning properly by triggering a write to it,
as the node would swiftly bail if any of the I/O operations failed.

Today's access patterns are wildly different so we'll detect disk problems
quickly anyway, not to mention Vespa Cloud has disk health monitoring built in.


geirst (approved):
👍

PR #19954, opened 2021-11-10
Revert "Always update config when generation has changed [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#19951

In case this breaks something, please review only


gjoranv (approved):

hmusum:
Need another solution that works without always setting `configChanged` to `true`

PR #19955, opened 2021-11-10
Generate routing endpoints in config model

tokle:


config-model-api/src/main/java/com/yahoo/config/model/api/ApplicationClusterEndpoint.java
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Represents one endpoint for an application cluster
+ *
+ * @author mortent
+ */
+public class ApplicationClusterEndpoint {
+    public enum Scope {APPLICATION, GLOBAL, ZONE}
+
+    public enum RoutingMethod {SHARED, SHAREDLAYER4}
```
mpolden, line 22: Can we use the same case as elsewhere? `application`, `sharedLayer4` etc.

config-model-api/src/main/java/com/yahoo/config/model/api/ApplicationClusterEndpoint.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private final DnsName dnsName;
+    private final Scope scope;
+    private final RoutingMethod routingMethod;
+    private final int weight;
+    private final List<String> hosts;
```
mpolden, line 28: Is this hostnames? Consider renaming.

config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
```diff
@@ -184,6 +194,60 @@ public void addServlet(Servlet servlet) {
...
+            endpoints.add(ApplicationClusterEndpoint.builder()
+                                  .zoneScope()
+                                  .sharedRouting()
+                                  .dnsName(l7Name)
+                                  .hosts(hosts)
+                                  .build());
+        }
+
+        // Then get all endpoints provided by controller. Can be created with L4 routing only
+        Set<ContainerEndpoint> endpointsFromController = deployState.getEndpoints();
```
mpolden, line 237: This will also include application-level endpoints now (if declared), but we haven't modeled that on the config server side yet.

configdefinitions/src/vespa/lb-services.def
```diff
@@ -6,17 +6,24 @@ namespace=cloud.config
...
+tenants{}.applications{}.hosts{}.services{}.index                       int default=0
+tenants{}.applications{}.hosts{}.services{}.ports[].number              int default=-1
+tenants{}.applications{}.hosts{}.services{}.ports[].tags                string default="(notags)"
+tenants{}.applications{}.hosts{}.services{}.servicealiases[]            string
+tenants{}.applications{}.hosts{}.services{}.endpointaliases[]           string
+
+
+tenants{}.applications{}.endpoints[].dnsName         string
+tenants{}.applications{}.endpoints[].scope           enum {APPLICATION, GLOBAL, ZONE}
+tenants{}.applications{}.endpoints[].routingMethod   enum {SHARED, SHAREDLAYER4}
```
mpolden, line 27: Same as above.

configdefinitions/src/vespa/lb-services.def
```diff
@@ -6,17 +6,24 @@ namespace=cloud.config
...
+tenants{}.applications{}.endpoints[].weight          int default=0
```
mpolden, line 28: Isn't 1 a better default? Seeing as 0 has special meaning (`backup`)  in Nginx.

mpolden (approved):

PR #19960, opened 2021-11-10
Revert "Always update config when generation has changed [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#19957

Please review only, merge if this breaks something


baldersheim (approved):

bjorncs (approved):

hmusum:
This will fail because config proxy has not implemented `setNewConfigAndGeneration()`.  Might be more changes needed in config proxy as well

PR #19961, opened 2021-11-10
Test the distribution we get with 8 attributes and 4/8 threads.

baldersheim:
@geirst PR

Perfect with 4 executors,
With 8 1 executor gets 2 and then he will do twice as much and become the bottleneck....
Which explains what we see.


geirst (approved):
👍

PR #19968, opened 2021-11-11
Extend test to cover case where we get new generation, but same config

hmusum:
With this we would have caught the issue that resulted in https://github.com/vespa-engine/vespa/pull/19960


gjoranv (approved):
👍

PR #19971, opened 2021-11-11
Revert "Generate routing endpoints in config model"

baldersheim:
Reverts vespa-engine/vespa#19955
@tokle or or @mpolden @hmusum PR I think his one broke the build


tokle:
Will fix the issue

PR #19980, opened 2021-11-11
Use given messagebus config also when slobroks config is self-sub

jonmv:

Both these were either present or absent. With the move back to self-subscription for slobroks config, this was no longer true. We still don't want self-subscription for messagebus config, though.


bjorncs (approved):
👍

PR #19981, opened 2021-11-11
Remove double book keeping of count.

baldersheim:
@geirst or @toregge PR


toregge (approved):
`AllocatedBitVector::grow()` calls `clearInterval()` which invalidates `_numTrueBits` when capacity is unchanged and size is reduced. Consider calling `clearIntervalNoInvalidation()` instead and then call `updateCount()` after `setSize()`.

baldersheim:
Comments addressed.

PR #19988, opened 2021-11-12
When node is retired we can relax compaction strategy as we have peaked

baldersheim:
in memory usage and will go reduce from here.

@geirst or @toregge PR
@vekterli FYI

### DIFF
--- a/searchcore/src/tests/proton/documentdb/document_subdbs/document_subdbs_test.cpp
+++ b/searchcore/src/tests/proton/documentdb/document_subdbs/document_subdbs_test.cpp
@@ -580,6 +580,37 @@ TEST_F("require that subdb reflect retirement", FastAccessFixture)
     EXPECT_TRUE(cfg == unretired_cfg);
 }
 
+TEST_F("require that attribute compaction config reflect retirement", FastAccessFixture) {
+    search::CompactionStrategy default_cfg(0.05, 0.2);
+    search::CompactionStrategy retired_cfg(0.5, 0.5);
+
+    auto guard = f._subDb.getAttributeManager()->getAttribute("attr1");
+    EXPECT_EQUAL(default_cfg, (*guard)->getConfig().getCompactionStrategy());
+    EXPECT_EQUAL(default_cfg, dynamic_cast<const proton::DocumentMetaStore &>(f._subDb.getDocumentMetaStoreContext().get()).getConfig().getCompactionStrategy());
+
+    auto calc = std::make_shared<proton::test::BucketStateCalculator>();
+    calc->setNodeRetired(true);
+    f._subDb.setBucketStateCalculator(calc);
+    f._writeService.sync();
+    guard = f._subDb.getAttributeManager()->getAttribute("attr1");
+    EXPECT_EQUAL(retired_cfg, (*guard)->getConfig().getCompactionStrategy());
+    EXPECT_EQUAL(retired_cfg, dynamic_cast<const proton::DocumentMetaStore &>(f._subDb.getDocumentMetaStoreContext().get()).getConfig().getCompactionStrategy());
+
+    f.basicReconfig(10);
+    f._writeService.sync();
+    guard = f._subDb.getAttributeManager()->getAttribute("attr1");
+    EXPECT_EQUAL(retired_cfg, (*guard)->getConfig().getCompactionStrategy());
+    EXPECT_EQUAL(retired_cfg, dynamic_cast<const proton::DocumentMetaStore &>(f._subDb.getDocumentMetaStoreContext().get()).getConfig().getCompactionStrategy());
+
+    calc->setNodeRetired(false);
+    f._subDb.setBucketStateCalculator(calc);
+    f._writeService.sync();
+    guard = f._subDb.getAttributeManager()->getAttribute("attr1");
+    EXPECT_EQUAL(default_cfg, (*guard)->getConfig().getCompactionStrategy());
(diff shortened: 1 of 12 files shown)


searchcore/src/vespa/searchcore/proton/attribute/attributemanager.cpp
```diff
@@ -583,6 +584,16 @@ AttributeManager::asyncForEachAttribute(std::shared_ptr<IConstAttributeFunctor>
     }
 }
 
+void
+AttributeManager::asyncForEachAttribute(std::shared_ptr<IAttributeFunctor> func) const
+{
+    for (const auto &attr : _attributes) {
+        AttributeVector::SP attrsp = attr.second.getAttribute();
```
toregge, line 591: Missing check and skip if `attr.second.isExtra()` is true.

searchcore/src/vespa/searchcore/proton/server/storeonlydocsubdb.cpp
```diff
@@ -426,9 +428,37 @@ StoreOnlyDocSubDB::reconfigure(const search::LogDocumentStore::Config & config,
...
+namespace {
+
+constexpr double RETIRED_DEAD_RATIO = 0.5;
+
+struct UpdateConfig : public search::attribute::IAttributeFunctor {
+    void operator()(search::attribute::IAttributeVector &iAttributeVector) override {
+        auto attributeVector = dynamic_cast<search::AttributeVector *>(&iAttributeVector);
+        if (attributeVector != nullptr) {
+            auto cfg = attributeVector->getConfig();
+            cfg.setCompactionStrategy(search::CompactionStrategy(RETIRED_DEAD_RATIO, RETIRED_DEAD_RATIO));
```
toregge, line 440: This has two issues:

1. normal config is not reapplied if node is moved from retired to up state.
2. new config without cluster state changes will set normal compaction strategy on the node even if it is still retiring.

searchcore/src/vespa/searchcore/proton/attribute/attributemanager.cpp
baldersheim: Why should it be different for the extra attributes ?

searchcore/src/vespa/searchcore/proton/server/storeonlydocsubdb.cpp
baldersheim: Yes, will address that.

toregge:
It might be better to treat node being (un)retired as a config change and amend the config. We already have AttributeAspectDelayer to delay attribute aspect changes until next node restart.

searchcore/src/vespa/searchcore/proton/attribute/attributemanager.cpp
toregge: Because the extra attribute (document meta store) uses the document db main executor.

baldersheim: Yes, that was the only extra one.

searchcore/src/vespa/searchcore/proton/server/storeonlydocsubdb.cpp
baldersheim: Now both issues should be addressed.

geirst (approved):
👍  But I think we should add a unit test for this. This might be a good place: searchcore/src/tests/proton/documentdb/document_subdbs/document_subdbs_test.cpp

searchcore/src/vespa/searchcore/proton/server/storeonlydocsubdb.cpp
toregge: The first issue is only addressed for document meta store and not for normal attribute vectors.

baldersheim: Now it should address for both

searchcore/src/vespa/searchcore/proton/server/fast_access_doc_subdb.cpp
```diff
@@ -275,6 +273,9 @@ FastAccessDocSubDB::applyConfig(const DocumentDBConfig &newConfigSnapshot, const
             reconfigureAttributeMetrics(*newMgr, *oldMgr);
         }
         _iFeedView.set(_fastAccessFeedView.get());
+        if (isNodeRetired()) {
+            reconfigureAttributesConsideringNodeState();
```
toregge, line 277: Attribute vectors on retired node will get a config glitch that is fixed up here, but might have spent some time performing compaction due to the compaction strategy temporarily being set as if the node was not retired.

searchcore/src/tests/proton/documentdb/document_subdbs/document_subdbs_test.cpp
```diff
@@ -557,6 +557,29 @@ TEST_F("require that attribute manager can be reconfigured", SearchableFixture)
     requireThatAttributeManagerCanBeReconfigured(f);
 }
 
+TEST_F("require that subdb reflect retirement", FastAccessFixture)
```
geirst, line 560: We should also add a test that verifies that the new compaction config is applied to an attribute vector.

baldersheim: Done, both documentmetastore and attributes.

searchcore/src/vespa/searchcore/proton/server/fast_access_doc_subdb.cpp
baldersheim: Then the modified config will be persisted and un retirement will not work.

geirst (commented):
👍

PR #20001, opened 2021-11-13
Add a fixed size table of 8 * num_exutors with 16 bit entries. Use th…

baldersheim:
…is for mapping the first components exact.

With more components than 8x we fall abck to to using shrunk id of 8 bits as before.
This enables perfect distribution for the first 8x and then 'good enough' for the rest.
The more there are the less impact of imperfect distribution will be.

@geirst and @toregge PR

### DIFF
--- a/staging_vespalib/src/tests/sequencedtaskexecutor/sequencedtaskexecutor_test.cpp
+++ b/staging_vespalib/src/tests/sequencedtaskexecutor/sequencedtaskexecutor_test.cpp
@@ -242,11 +242,16 @@ TEST("require that you get correct number of executors") {
 TEST("require that you distribute well") {
     auto seven = SequencedTaskExecutor::create(sequenced_executor, 7);
     const SequencedTaskExecutor & seq = dynamic_cast<const SequencedTaskExecutor &>(*seven);
+    const uint32_t NUM_EXACT = 8 * seven->getNumExecutors();
     EXPECT_EQUAL(7u, seven->getNumExecutors());
     EXPECT_EQUAL(97u, seq.getComponentHashSize());
     EXPECT_EQUAL(0u, seq.getComponentEffectiveHashSize());
     for (uint32_t id=0; id < 1000; id++) {
-        EXPECT_EQUAL((id%97)%7, seven->getExecutorId(id).getId());
+        if (id < NUM_EXACT) {
+            EXPECT_EQUAL(id % seven->getNumExecutors(), seven->getExecutorId(id).getId());
+        } else {
+            EXPECT_EQUAL(((id - NUM_EXACT) % 97) % seven->getNumExecutors(), seven->getExecutorId(id).getId());
+        }
     }
     EXPECT_EQUAL(97u, seq.getComponentHashSize());
     EXPECT_EQUAL(97u, seq.getComponentEffectiveHashSize());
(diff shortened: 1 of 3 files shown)


toregge (approved):

staging_vespalib/src/vespa/vespalib/util/sequencedtaskexecutor.cpp
```diff
@@ -103,34 +105,59 @@ ExecutorStats
...
+    }
+}
+
+ISequencedTaskExecutor::ExecutorId
+SequencedTaskExecutor::getExecutorIdPerfect(uint64_t componentId) const {
+    PerfectKeyT key = componentId;
+    std::lock_guard guard(_mutex);
+    auto found = std::find(_component2IdPerfect.begin(), _component2IdPerfect.end(), key);
+    if (found == _component2IdPerfect.end()) {
+        _component2IdPerfect.push_back(key);
```
toregge, line 133: Might need some protection against pushing too many elements onto `_component2IdPerfect`. If the old size is the same as `getNumExecutors() * 8 - 1` and to threads tries to get executor id for two new unique component ids, then
both might extend the vector, causing an unsafe resize (considering that it's read without lock).

Another issue is that `std::find()` in `getExecutorId()` might access uninitialized memory if `end()` returns new end before new value is written. If that uninitialized memory matches the key used for the search then we get an invalid result.

toregge (changes requested):

baldersheim:
@toregge PTAL
I hope both issues are now addressed

staging_vespalib/src/vespa/vespalib/util/sequencedtaskexecutor.h
```diff
@@ -43,17 +43,27 @@ class SequencedTaskExecutor final : public ISequencedTaskExecutor
     const vespalib::SyncableThreadExecutor* first_executor() const;
 
 private:
+    class ValidId {
```
toregge, line 46: consider using `std::optional<ExecutorId>` instead.

toregge (approved):

staging_vespalib/src/tests/sequencedtaskexecutor/sequencedtaskexecutor_test.cpp
```diff
@@ -272,8 +277,8 @@ TEST("require that similar names gets 7/8 unique ids with 8 executors") {
     EXPECT_EQUAL(3u, four->getExecutorIdFromName("f4").getId());
     EXPECT_EQUAL(4u, four->getExecutorIdFromName("f5").getId());
     EXPECT_EQUAL(5u, four->getExecutorIdFromName("f6").getId());
-    EXPECT_EQUAL(2u, four->getExecutorIdFromName("f7").getId());
-    EXPECT_EQUAL(6u, four->getExecutorIdFromName("f8").getId());
+    EXPECT_EQUAL(6u, four->getExecutorIdFromName("f7").getId());
```
geirst, line 280: Please update the name of the test to reflect the new reality.

staging_vespalib/src/vespa/vespalib/util/sequencedtaskexecutor.h
baldersheim: Done

staging_vespalib/src/tests/sequencedtaskexecutor/sequencedtaskexecutor_test.cpp
baldersheim: Fixed in #20029

PR #20009, opened 2021-11-15
Optimize proton::documentmetastore::LidAllocator::unregister_lids().

toregge:
@baldersheim : please review
@geirst : FYI


baldersheim (approved):
what was the gain ?

toregge:
Testing on a desktop machine indicated  a gain from about 100-120 Mlids/s (calling unregisterLid in loop) to about 180-200 Mlids/s (using the optimized version of unregister_lids).

PR #20013, opened 2021-11-15
Use a custom property for setting relase version for clients.

gjoranv:
Other naming suggestions for the maven property?

- Always set release version via maven-compiler-plugin, instead
  of maven property which is overridden by compiler-plugin config.
- Using a custom property with self-explanatory name makes
  comments redundant.
- Remove explicit jdkToolchain config, as these modules no longer
  compile with jdk pre 9, due to the --release flag.

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -16,12 +16,6 @@
   <packaging>jar</packaging>
   <version>7-SNAPSHOT</version>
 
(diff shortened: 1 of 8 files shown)


hmusum (approved):
LGTM 👍

aressem (approved):

PR #20016, opened 2021-11-15
Remove old devtools paths

freva:


aressem (approved):

aressem:
Note that devtoolset-10 is still installed in open source CentOS 7 hosts/containers. But node-admin is not used there, so this is probably ok.

PR #20018, opened 2021-11-15
ignore requested stack sizes

arnej27959:
* for C++ threads only
* triggered by feature flag, by default use old behavior

@havardpe please review
@baldersheim @toregge FYI


havardpe (commented):
Could we end up with an issue due to restarting everything at the same time?

Darwin seems to be an odd case with a C++ stack size of only 512k, which can cause problems with highly nested rank feature executor setups (@toregge brought this up).

toregge:
When running unit tests on Darwin with `env VESPA_IGNORE_REQUESTED_STACK_SIZES=true ctest`, I get 
```
The following tests FAILED:
	547 - searchlib_ranksetup_test_app (Bus error)
	548 - searchlib_verify_feature_test_app (Bus error)
```
This is due to blueprint resolver needing more stack than the default (512kiB) on Darwin (cf. PR #19301 ).

If thread stack size is removed as argument from `vespalib::ThreadStackExecutor` constructor then it would be nice to have it hardcoded to 8 MiB in `FastOS_UNIX_Thread::Initialize()` if __APPLE__ is defined.

Programs with larger stack size will need more memory on systems without memory overcommit, e.g. 1024 threads will use 8 GiB memory when each thread has 8 MiB stack, while only using 256 MiB memory when each thread has 256 kiB stack.

havardpe:
I suspect both failures are caused by the 'chain' executor that explicitly tests the maximum nesting depth, which was increased from 64 to 256 when we started converting onnx bert models into (very many, highly nested) native vespa ranking expressions. The thing that makes it hard to re-write the compiler as non-recursive is that the feature blueprint setup function expects to be able to identify the types of requested inputs, which it cannot know until they are set up.

havardpe (approved):
I suggest testing it out on factory as a first step.

PR #20025, opened 2021-11-16
Use std::atomics for all, but x86_64.

baldersheim:
@toregge @vekterli PR
@aressem FYI

### DIFF
--- a/vespamalloc/src/vespamalloc/malloc/allocchunk.h
+++ b/vespamalloc/src/vespamalloc/malloc/allocchunk.h
@@ -16,9 +16,9 @@ namespace vespamalloc {
  * but requires the double-word compare-and-swap instruction.
  * Very early Amd K7/8 CPUs are lacking this and will fail (Illegal Instruction).
  **/
-struct TaggedPtr {
-    TaggedPtr() noexcept : _ptr(nullptr), _tag(0) { }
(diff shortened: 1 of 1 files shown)


toregge (approved):

vespamalloc/src/vespamalloc/malloc/allocchunk.h
```diff
@@ -61,16 +61,20 @@ struct AtomicTaggedPtr {
...
 #endif
 
 
 class AFListBase
 {
 public:
-    using HeadPtr = std::conditional<std::atomic<TaggedPtr>::is_always_lock_free, TaggedPtr, AtomicTaggedPtr>::type;
-    using AtomicHeadPtr = std::conditional<std::atomic<TaggedPtr>::is_always_lock_free, std::atomic<TaggedPtr>, AtomicTaggedPtr>::type;
+    using HeadPtr = TaggedPtr;
+    using AtomicHeadPtr = std::atomic<TaggedPtr>;
```
vekterli, line 77: Did this use to be `std::atomic` on x64? I was under the impression that `is_always_lock_free` is `false` for 128-bit x64 atomics due to the whole "can't use 128-bit CaS since it doesn't work with read-only memory" debacle. If so, can this cause any performance regressions due to an introduction of implicit spinlocking?

PR #20032, opened 2021-11-16
Require graviton2 for aarch64i as default.

baldersheim:
@toregge and @aressem PR


default_build_settings.cmake
```diff
@@ -239,6 +239,8 @@ function(vespa_use_default_build_settings)
       else()
         set(DEFAULT_VESPA_CPU_ARCH_FLAGS "-mtune=intel")
       endif()
+    elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64")
+      set(DEFAULT_VESPA_CPU_ARCH_FLAGS "-march=armv8.2-a+fp16+rcpc+dotprod+crypto -mtune=neoverse-n1")
```
toregge, line 243: I believe this will break compilation on Raspberry pi 4, and possibly also compilation on M1.

toregge (approved):

default_build_settings.cmake
baldersheim: Correct, we should revisit before making aarch64 officially supported.
But for now we let it be optimal for graviton2 instances on AWS so we can get some comparable benchmarks.

PR #20033, opened 2021-11-16
Default is to use atomic library explicit.

baldersheim:
@toregge PR


toregge (commented):
I believe the atomic library is now needed for RHEL 8 too. `dist/vespa.spec` also need an update (more BuildRequires)

toregge (approved):

PR #20036, opened 2021-11-16
Let developers run prod tests against non-default instances

jonmv:
@hakonhall please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -2042,7 +2042,7 @@ public class ApplicationApiHandler extends AuditLoggingRequestHandler {
         return new MessageResponse("Deactivated " + id);
     }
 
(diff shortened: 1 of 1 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2044,17 +2044,18 @@ private HttpResponse deactivate(String tenantName, String applicationName, Strin
 
     /** Returns test config for indicated job, with production deployments of the default instance. */
```
hakonhall, line 2045: update this too?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
jonmv:
```suggestion
    /** Returns test config for indicated job, with production deployments of the default instance if the given is not in deployment spec. */
```

PR #20038, opened 2021-11-16
Include declared application endpoints in certificate

mpolden:
@tokle

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/certificate/EndpointCertificatesTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/certificate/EndpointCertificatesTest.java
@@ -5,6 +5,8 @@ import com.yahoo.config.application.api.DeploymentSpec;
 import com.yahoo.config.application.api.xml.DeploymentSpecXmlReader;
 import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.Environment;
+import com.yahoo.config.provision.InstanceName;
+import com.yahoo.config.provision.RegionName;
 import com.yahoo.config.provision.SystemName;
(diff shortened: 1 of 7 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/certificate/EndpointCertificatesTest.java
```diff
@@ -215,12 +219,57 @@ public void includes_zones_in_deployment_spec_when_deploying_to_staging() {
...
+                                     Map.of(InstanceName.from("beta"), 1,
+                                            InstanceName.from("main"), 1))
+                .applicationEndpoint("c", "qrs", zone1.region().value(),
+                                     Map.of(InstanceName.from("beta"), 4,
+                                            InstanceName.from("main"), 6))
+                .build();
+        ControllerTester tester = new ControllerTester(SystemName.Public);
+        EndpointCertificateValidatorImpl endpointCertificateValidator = new EndpointCertificateValidatorImpl(secretStore, clock);
+        EndpointCertificates endpointCertificates = new EndpointCertificates(tester.controller(), endpointCertificateMock, endpointCertificateValidator);
+        List<String> expectedSans = List.of(
```
tokle, line 252: It is strange that this does not include san dns name for zone aws-us-west-2a ... But this is not really what you test here ...

tokle (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/certificate/EndpointCertificatesTest.java
mpolden: Agree, these tests should be rewritten to use DeploymentTester.

PR #20052, opened 2021-11-16
We only the docid for remove and stat.

baldersheim:
@vekterli PR


vekterli (approved):
:+1: I'm a bit rusty on the `createIterator` semantics, but I presume the `FieldSet` passed to it does _not_ affect the set of fields actually fetched to process the document selection? I.e. you can still select on any field in the document even with a `DocIdOnly` `FieldSet`.

baldersheim:
Yes, as far as I could tell that is fine. I would hope that some unit tests would fail if that was not the case :)

PR #20072, opened 2021-11-17
Move removeLocation over to Asynchandler and issue all removes for on…

baldersheim:
…e bucket before waiting for the replies.

Prepare RemoveResult to contain more replies.

@vekterli PR

### DIFF
--- a/storage/src/tests/persistence/processalltest.cpp
+++ b/storage/src/tests/persistence/processalltest.cpp
@@ -13,7 +13,8 @@ using namespace ::testing;
 
 namespace storage {
 
-class ProcessAllHandlerTest : public SingleDiskPersistenceTestUtils {
+struct ProcessAllHandlerTest : public PersistenceTestUtils {
+    document::BucketIdFactory _bucketIdFactory;
 };
 
 TEST_F(ProcessAllHandlerTest, change_of_repos_is_reflected) {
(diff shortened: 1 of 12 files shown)


storage/src/tests/persistence/processalltest.cpp
```diff
@@ -101,7 +105,8 @@ TEST_F(ProcessAllHandlerTest, remove_location_throws_exception_on_bogus_selectio
     document::Bucket bucket = makeDocumentBucket(bucketId);
     auto cmd = std::make_shared<api::RemoveLocationCommand>("id.bogus != badgers", bucket);
 
-    ProcessAllHandler handler(getEnv(), getPersistenceProvider());
+    document::BucketIdFactory bucketIdFactory;
```
vekterli, line 108: Consider moving the `BucketIdFactory` up into the `ProcessAllHandlerClass` test fixture class

storage/src/vespa/storage/persistence/asynchandler.cpp
```diff
@@ -378,4 +397,38 @@ AsyncHandler::checkProviderBucketInfoMatches(const spi::Bucket& bucket, const ap
...
+                                processor, spi::NEWEST_DOCUMENT_ONLY,tracker->context());
+
+    std::vector<std::future<std::unique_ptr<spi::Result>>> results;
+    results.reserve(to_remove.size());
+    for (auto & entry : to_remove) {
+        auto catcher = std::make_unique<spi::CatchResult>();
+        results.push_back(catcher->future_result());
+        _spi.removeAsync(bucket, entry.first, entry.second, tracker->context(), std::move(catcher));
+    }
+    for (auto & future : results) {
```
vekterli, line 423: Do we want to do a `results.back().wait()` before this (iff `results` is non-empty of course) to effectively guarantee we only put the calling thread to sleep once (assuming removes have their completion callbacks invoked in FIFO order)? This also ensures we don't throw any exceptions below before all pending ops to the bucket have completed.

vekterli (approved):
:+1:

storage/src/tests/persistence/processalltest.cpp
baldersheim: Fixed

storage/src/vespa/storage/persistence/asynchandler.cpp
baldersheim: I do not think fifo order is guaranteed.
Anyway this is shortlived as I have another PR in the making that moves the asyncness further down.

PR #20076, opened 2021-11-18
Revert "Add vespa program to vespa-clients RPM"

aressem:
Reverts vespa-engine/vespa#20069

RPM building failed:
```
07:56:07 Processing files: vespa-7.504.2-1.el8.x86_64
07:56:07 error: Missing build-id in /vespa/vespa-rpmbuild/BUILDROOT/vespa-7.504.2-1.el8.x86_64/opt/vespa/bin/vespa
07:56:07 error: Generating build-id links failed
07:56:07 
07:56:07 
07:56:07 RPM build errors:
07:56:07     Missing build-id in /vespa/vespa-rpmbuild/BUILDROOT/vespa-7.504.2-1.el8.x86_64/opt/vespa/bin/vespa
07:56:07     Generating build-id links failed
```

PR #20077, opened 2021-11-18
Add vespa program to vespa-clients RPM, try 2

hakonhall:
Reverts vespa-engine/vespa#20076


dist/vespa.spec
```diff
@@ -6,6 +6,10 @@
 # Only strip debug info
 %global _find_debuginfo_opts -g
 
+# Go binaries' build-ids are not recognized by RPMs yet, see
+# https://github.com/rpm-software-management/rpm/issues/367.
+%global _missing_build_ids_terminate_build 0
```
hakonhall, line 11: This ignores build-ids for all packages.  I actually only need it for the clients subpackage, but don't know if, say, putting the %global statement in `%package clients` would work.

aressem (approved):

PR #20085, opened 2021-11-18
Jonmv/vespa cli test runner

jonmv:


hakonhall (approved):

vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
```diff
@@ -43,7 +43,7 @@
 
     @Inject
     public VespaCliTestRunner(VespaCliTestRunnerConfig config) {
-        this(config.artifactsPath());
+        this(config.artifactsPath().resolve("artifacts"));
```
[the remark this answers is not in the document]
jonmv: We can't change that.

jonmv:
PTAL.

vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
```diff
@@ -133,8 +137,7 @@ private void log(Level level, String message, Throwable thrown) {
         }
     }
 
-    static String toEndpointsConfig(byte[] testConfig) throws IOException {
-        TestConfig config = TestConfig.fromJson(testConfig);
+    static String toEndpointsConfig(TestConfig config) throws IOException {
         Cursor root = new Slime().setObject();
         Cursor endpointsArray = root.setArray("endpoints");
         config.deployments().get(config.zone()).forEach((cluster, url) -> {
```
hakonhall, line 143: and scope

hakonhall (approved):

vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
jonmv: ?

hakonhall: You should add scope "zone" to each endpoint

jonmv: Ah. So we probably end up with custom parser code for this anyway, so we don't need those. See following PR.

hakonhall: No, my bad.  If the intention is anyway to just end up with the map, then no scope is needed.  I was thinking we needed to reuse the code that filters on scope, but there is no need for that.

PR #20087, opened 2021-11-18
vespa-cli: Build endpoints map

hakonhall:
This PR is supposed to be a no-op.  But the idea is to hold on to the full map from cluster name to URL endpoints in cloudTarget, and resue the target across multiple requests using possibly different clusters.

### DIFF
--- a/client/go/vespa/target.go
+++ b/client/go/vespa/target.go
@@ -208,9 +208,32 @@ type cloudTarget struct {
 	tlsOptions TLSOptions
 	logOptions LogOptions
 
-	queryURL       string
-	documentURL    string
+	urlsByCluster  map[string]string
 	authConfigPath string
+}
+
+func (t *cloudTarget) resolveEndpoint(cluster string) (string, error) {
+	if cluster == "" {
+		for _, u := range t.urlsByCluster {
+			if len(t.urlsByCluster) == 1 {
+				return u, nil
+			} else {
+				return "", fmt.Errorf("multiple clusters, none chosen: %v", t.urlsByCluster)
+			}
+		}
+	} else {
+		u := t.urlsByCluster[cluster]
+		if u == "" {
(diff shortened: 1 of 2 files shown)


hakonhall:
FYI @jonmv

client/go/vespa/target.go
```diff
@@ -208,11 +208,34 @@ type cloudTarget struct {
...
 
-	queryURL       string
-	documentURL    string
+	urlsByCluster  map[string]string
 	authConfigPath string
 }
 
+func (t *cloudTarget) resolveEndpoint(cluster string) (string, error) {
+	if cluster == "" {
+		for _, u := range t.urlsByCluster {
```
mpolden, line 217: Isn't this loop unnecessary?

client/go/vespa/target.go
```diff
@@ -208,11 +208,34 @@ type cloudTarget struct {
...
+			}
+		}
+	} else {
+		u := t.urlsByCluster[cluster]
+		if u == "" {
+			clusters := make([]string, len(t.urlsByCluster))
+			for c := range t.urlsByCluster {
+				clusters = append(clusters, c)
+			}
+			return "", fmt.Errorf("unknown cluster '%s': must be one of %v, ", cluster, clusters)
```
mpolden, line 231: Any reason for the trailing `, ` in the message?

client/go/vespa/target.go
hakonhall: How would I get the value of the entry, if the map contains exactly one entry?  I found this which seems ugly:

```
t.urlsByCluster[reflect.ValueOf(t.urlsByCluster).MapKeys()[0]]
```

client/go/vespa/target.go
hakonhall: Thanks, fixed

mpolden (approved):

PR #20093, opened 2021-11-18
Revert "Use live gid to lid map when processing partial update."

toregge:
Reverts vespa-engine/vespa#20083

Need a slightly different fix.


toregge:
Will just remove unneeded assert in inspectExisting instead.

PR #20098, opened 2021-11-18
Allow overriding creds paths, and add dummy test command

jonmv:
@mpolden please review. This is just a draft for now.


jonmv:
@hakonhall FYI

jonmv:
The plan is to pre-populate the `urlsByCluster` of the `cloudTarget` with given `--endpoints`, and if this is done, we skip the wait-for-endpoints when getting a `Service` later down the road.

client/go/cmd/test.go
```diff
@@ -0,0 +1,100 @@
...
+
+var (
+	endpointsArg     string
+	dataPlaneCertArg string
+	dataPlaneKeyArg  string
+)
+
+func init() {
+	rootCmd.AddCommand(testCmd)
+	testCmd.PersistentFlags().StringVar(&endpointsArg, endpointsFlag, "", "The endpoints to use, for each container cluster")
```
mpolden, line 30: Not sure I understand why this is an option. Shouldn't this discover relevant endpoints automatically?

jonmv: When run by a developer, yes.
When run in the tester-container, endpoints are provided as config, and we can't ask the controller. 

I wonder, though, if these overrides all should be ENV instead?

mpolden: > When run in the tester-container, endpoints are provided as config, and we can't ask the controller.

Right.

> I wonder, though, if these overrides all should be ENV instead?

Good idea. That what we use for overriding system today, i.e. things regular users don't typically care about.

jonmv:
Code seems to work, but will make some adjustments to output, etc.. Will also add a unit test for the new command, but I think it's ready for review, @mpolden :)

@hakonhall FYI.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+	Args:              cobra.MaximumNArgs(1),
+	DisableAutoGenTag: true,
+	Run: func(cmd *cobra.Command, args []string) {
+		target := getTarget()
+		testPath := "."
+		if len(args) > 0 {
+			testPath = args[0]
+		}
+		if count, failed := runTests(testPath, target); len(failed) != 0 {
+			_, _ = fmt.Fprintf(stdout, "\nFailed %d of %d tests:\n", len(failed), count)
```
mpolden, line 49: Remove the unnecessary `_, _ =`. Same for all the other occurrences below.

client/go/vespa/target.go
```diff
@@ -129,7 +128,7 @@ func (s *Service) Description() string {
 
 func (t *customTarget) Type() string { return t.targetType }
 
-func (t *customTarget) Service(name string, timeout time.Duration, sessionID int64) (*Service, error) {
+func (t *customTarget) Service(name string, timeout time.Duration, _ int64, _ string) (*Service, error) {
```
mpolden, line 131: I think it's better to keep the parameter names the same as the interface declaration, even if they're unused.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+		} else {
+			_, _ = fmt.Fprintf(stdout, "%d tests completed successfully\n", count)
+		}
+	},
+}
+
+func runTests(rootPath string, target vespa.Target) (int, []string) {
+	count := 0
+	failed := make([]string, 0)
+	if stat, err := os.Stat(rootPath); err != nil {
```
mpolden, line 66: Not needed? I'm sure `os.ReadDir` fails if `rootPath` is not a directory.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+func init() {
+	rootCmd.AddCommand(testCmd)
+}
+
+var testCmd = &cobra.Command{
+	Use:   "test [tests directory or test file]",
+	Short: "Run a test suite, or a single test",
+	Long: `Run a test suite, or a single test
+
+Runs all JSON test files in the specified directory, or the single JSON
```
mpolden, line 34: I assume we'll eventually have a documentation page for this feature. We should link to that here when it's added.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+
+func getParameters(parametersRaw []byte, testsPath string) (map[string]string, error) {
+	if parametersRaw != nil {
+		var parametersPath string
+		if err := json.Unmarshal(parametersRaw, &parametersPath); err == nil {
+			parametersRaw = readFile(path.Join(testsPath, parametersPath))
+		}
+		var parameters map[string]string
+		if err := json.Unmarshal(parametersRaw, &parameters); err != nil {
+			return nil, fmt.Errorf("request parameters must be JSON with only string values: %v", err)
```
mpolden, line 298: Nit: Prefer using `%w` as the format specifier for errors as this allows unwrapping.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+
+func getBody(bodyRaw []byte, testsPath string) ([]byte, error) {
+	var bodyPath string
+	if err := json.Unmarshal(bodyRaw, &bodyPath); err == nil {
+		return readFile(path.Join(testsPath, bodyPath)), nil
+	}
+	return bodyRaw, nil
+}
+
+func readFile(filePath string) []byte {
```
mpolden, line 313: Use `ioutil.ReadFile` instead, which does the same as this.

client/go/cmd/test.go
```diff
@@ -0,0 +1,357 @@
...
+	failed := make([]string, 0)
+	if stat, err := os.Stat(rootPath); err != nil {
+		fatalErr(err, "Failed reading specified test path")
+	} else if stat.IsDir() {
+		tests, err := os.ReadDir(rootPath)
+		if err != nil {
+			fatalErr(err, "Failed reading specified test directory")
+		}
+		for _, test := range tests {
+			if !test.IsDir() && strings.HasSuffix(test.Name(), ".json") {
```
mpolden, line 74: Nit: Consider `filepath.Ext(test.Name()) == ".json"`.

client/go/cmd/test.go
mpolden: Never mind, I saw the other usage of `stat`.

client/go/cmd/test.go
jonmv: IntelliJ complains when these aren't there, but ok :)

client/go/cmd/test.go
jonmv: Good idea, will add a TODO for this.

client/go/vespa/target.go
jonmv: OK.

client/go/cmd/test.go
jonmv: Thanks!

client/go/cmd/test.go
jonmv: Thanks!

client/go/cmd/test.go
jonmv: Nice.

mpolden (commented):
LGTM, but build fails on formatting (can be fixed with `make fmt`)

jonmv:
Yeah, there's a unit tests coming too :)

jonmv:
There we go. PTAL :)

jonmv:
@hakonhall we should probably have another `go` at formatting, names, etc.. But having this merged for testing purposes would be great :)

mpolden (approved):
👍 build still fails though, but not on Go code.

PR #20109, opened 2021-11-19
Avoid sync_all by using an explicit gate.

baldersheim:
@toregge PR


toregge (approved):
Note that this will fail if the executor has been shut down.

baldersheim:
ISequencedTaskExecutor has no shutdown. That can only happen in destructor.

PR #20110, opened 2021-11-19
Avoid sync_all in AttributeWriter destructor.

baldersheim:
@toregge PR


toregge (approved):
Note that this will fail if the executor has been shut down.

baldersheim:
Superceeded by rebase in #20122

PR #20117, opened 2021-11-19
No routing policies to for tester applications

jonmv:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
@@ -517,7 +517,8 @@ public class ApplicationController {
         } finally {
             // Even if prepare fails, a load balancer may have been provisioned. Always refresh routing policies so that
             // any DNS updates can be propagated as early as possible.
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -517,7 +517,8 @@ private ActivateResult deploy(ApplicationId application, ApplicationPackage appl
         } finally {
             // Even if prepare fails, a load balancer may have been provisioned. Always refresh routing policies so that
             // any DNS updates can be propagated as early as possible.
-            controller.routing().policies().refresh(application, applicationPackage.deploymentSpec(), zone);
+            if ( ! application.instance().isTester())
```
mpolden, line 520: This is not necessary as LoadBalancerProvisioner never provision load balancers testers.

mpolden (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
jonmv: Great.

PR #20127, opened 2021-11-19
Freeze gid to lid map tree before compacting it.

toregge:
@baldersheim : please review


baldersheim (approved):

toregge:
This fixes the issue observed with #20106.

PR #20136, opened 2021-11-21
GC some unused code and less sync_all_executors.

baldersheim:
@toregge PR

PR #20137, opened 2021-11-22
Follow renaming

baldersheim:
@toregge PR
Belongs to #20136


baldersheim:
Merging as this should have been part of #20136

PR #20156, opened 2021-11-22
Continue serving search queries when in Maintenance node state [run-systemtest]

vekterli:
@geirst @baldersheim @toregge please review.

Previously, entering maintenance state would implicitly deactivate
all buckets on the searchnode and cause empty responses to be returned
for searches.

However, container query dispatch uses async health pings to decide
which nodes to route queries to, so it would be possible for a node to
still be used for queries for a few seconds until the ping discovered
that the node should not be used. In the case of multiple groups without
multiple ready replicas within the group, this would cause transient
coverage loss since the dispatcher would not realize it should route
queries to other groups instead.

With this commit, maintenance edge behavior is changed as follows:

 - Buckets are _not_ deactivated when going from an available state
   to the maintenance state. However, they _are_ deactivate when going
   from maintenance state to an available state in order to avoid transient
   query duplicates immediately after the change.
 - Searches are executed as normal instead of returning empty replies
   when the node is in maintenance state.

The following behavior is intentionally _not_ changed:

 - The search interface is still marked as offline when in maintenance
   state, as this signals that the node should be taken out of rotation.
   In particular, it's critical that the RPC health ping response is
   explicitly tagged as having zero active docs when the search interface
   is offline, even though many buckets may now actually be active.
   Otherwise, queries would not be gracefully drained from the node.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/buckethandler.cpp
+++ b/searchcore/src/vespa/searchcore/proton/server/buckethandler.cpp
@@ -144,6 +144,12 @@ BucketHandler::handlePopulateActiveBuckets(document::BucketId::List &buckets,
     }));
 }
 
+namespace {
+constexpr const char* bool_str(bool v) noexcept {
+    return v ? "true" : "false";
+}
+}
+
 void
 BucketHandler::notifyClusterStateChanged(const std::shared_ptr<IBucketStateCalculator> & newCalc)
 {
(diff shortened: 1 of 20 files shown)


baldersheim (approved):

searchcore/src/vespa/searchcore/proton/server/buckethandler.cpp
```diff
@@ -147,9 +148,20 @@ void
...
+    if (_nodeMaintenance) {
+        return; // Don't deactivate buckets in maintenance mode; let query traffic drain away naturally.
+    }
+    // We implicitly deactivate buckets in two edge cases:
+    //  - Up -> Down (not maintenance; handled above), since the node can not be expected to offer
+    //    any graceful query draining when set Down.
+    //  - Maintenance -> Up, since we'd otherwise introduce a bunch of transient duplicate results
+    //    into queries. The assumption is that the system has already activated buckets on other
+    //    nodes in such a scenario.
+    if ((oldNodeUp && !_nodeUp) || (oldNodeMaintenance && _nodeUp)) {
```
toregge, line 164: The transitions up -> maintenance -> down -> up seems to not trigger deactivation of buckets.
If old state is maintenance and new state is not maintenance then all buckets should be deactivated.

i.e. instead of `(oldNodeMaintenance && _nodeUp)` consider using `(oldNodeMaintenance && !_nodeMaintenance)` or `(!oldNodeUp && _nodeUp)`

toregge (approved):

searchcore/src/vespa/searchcore/proton/server/buckethandler.cpp
vekterli: Good point. I think `oldNodeMaintenance && !_nodeMaintenance` might be the best, since this particular edge case isn't expected to happen except from manual administrative tasks (i.e. `vespa-set-node-state`). `(!oldNodeUp && _nodeUp)` would presumably trigger on every normal startup, when everything is already implicitly marked as inactive.

However, I think I need to ponder these changes a bit more since the current interaction with having multiple bucket spaces with differing maintenance state (i.e. global document merging) isn't as well-defined as I would like it to be.

vekterli:
I believe I have addressed the intricacies of multi bucket space node states. @geirst @baldersheim @toregge PTAL

baldersheim (approved):

storage/src/tests/persistence/filestorage/deactivatebucketstest.cpp
```diff
@@ -14,43 +14,100 @@ using namespace ::testing;
...
+    _node->getStateUpdater().setClusterState(state_of("storage:2 .1.s:m distributor:2"));
+    _node->getStateUpdater().setClusterState(state_of("storage:2 distributor:2"));
+    EXPECT_FALSE(is_active(test_bucket()));
 }
 
+TEST_F(DeactivateBucketsTest, buckets_deactivated_when_node_goes_from_maintenance_to_down)
+{
+    EXPECT_TRUE(is_active(test_bucket()));
+    _node->getStateUpdater().setClusterState(state_of("storage:2 .1.s:m distributor:2"));
+    _node->getStateUpdater().setClusterState(state_of("storage:2 distributor:2"));
```
geirst, line 91: Is this state down? It is equal to line 83 ("storage:2 distributor:2") which seems to be up.

geirst (commented):
👍  With one comment

storage/src/tests/persistence/filestorage/deactivatebucketstest.cpp
vekterli: Whoops, that's definitely an oversight. Will fix as part of a new PR which also has a fix for a missing member initialization

PR #20171, opened 2021-11-23
Bjorncs/status code metrics

bjorncs:
FYI @baldersheim @yngveaasheim @oyving


bjorncs:
We need to run this through performance tests to evaluate the penalty of the `ConcurrentHashMap` lookup

bjormel:
@bjorncs looks good to me, but would like @baldersheim to take a look too

bjormel (approved):

PR #20173, opened 2021-11-23
Revert "Continue serving search queries when in Maintenance node state [run-systemtest]"

baldersheim:
Reverts vespa-engine/vespa#20156
@vekterli PR
valgrind was not happy.


vekterli (commented):
👀 👍

PR #20177, opened 2021-11-24
Remove CGroups v2, map user NS and mount readonly feature flags

freva:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -45,26 +45,6 @@
 
     private static volatile TreeMap<FlagId, FlagDefinition> flags = new TreeMap<>();
 
-    public static final UnboundBooleanFlag MAP_USER_NAMESPACE = defineFeatureFlag(
-            "map-user-namespace", false,
```
hakonhall, line 49: Removing these will break hosted-feature-flags since they have overrides.

freva: Removed commit, PTAL.

hakonhall (approved):

PR #20183, opened 2021-11-24
Allow searches when node is in maintenance mode (déjà vu)

vekterli:
@baldersheim @geirst @baldersheim same as #20156 but rebased on master and two added commits on top of the cake:
* 4e3265effeadd40efa70251402b5f6edcc3f2371 adds the missing fragrant and delicious memory initialization that computers love so very much
* 651d28fe93864e569a76be72fbb7219766d71d0f ensures we actually test a Maintenance -> Down node state transition


geirst (approved):
👍

toregge (approved):

baldersheim (approved):

PR #20188, opened 2021-11-24
Hmusum/config tests refactoring

hmusum:


config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
```diff
@@ -53,8 +53,9 @@
     private final ConnectionPool connectionPool;
     private final ConfigSourceSet configSourceSet;
 
-    private Instant noApplicationWarningLogged = Instant.MIN;
+    private Instant timeForLastLogWarning;
     private int failures = 0;
+    private boolean closed = false;
```
gjoranv, line 58: Should this be atomic or volatile? It's only used for logging and testing now, but maybe more later?

gjoranv (approved):

config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
hmusum: Yes, good point, will fix

PR #20190, opened 2021-11-24
Format changes, and rename to steps array

jonmv:
@hakonhall please review and merge. Check the `expeced-suite.json` for sample output with lots happening.

### DIFF
--- a/client/go/cmd/testdata/tests/expected-suite.out
+++ b/client/go/cmd/testdata/tests/expected-suite.out
@@ -1,229 +1,9 @@
-Running testdata/tests/system-test/test.json: .... OK!
-Running testdata/tests/system-test/wrong-bool-value.json:
-Failed verifying assertion 0:
-Expected JSON at /root/coverage/full (false) does not match actual (true). Response body:
-{
-  "root": {
-    "children": [
-      {
-        "fields": {
-          "artist": "Foo Fighters",
-          "documentid": "id:test:music::doc",
-          "sddocname": "music"
-        },
-        "id": "id:test:music::doc",
-        "relevance": 0.38186238359951247,
-        "source": "music"
-      }
-    ],
-    "coverage": {
-      "coverage": 100,
-      "documents": 1,
-      "full": true,
-      "nodes": 1,
-      "results": 1,
-      "resultsFull": 1
-    },
-    "fields": {
-      "totalCount": 1
-    },
-    "id": "toplevel",
-    "relevance": 1
-  },
-  "timing": {
-    "querytime": 0.003,
-    "searchtime": 0.004,
-    "summaryfetchtime": 0
-  }
-}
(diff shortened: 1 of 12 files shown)


client/go/cmd/testdata/tests/expected-suite.out
```diff
@@ -259,11 +271,52 @@ Expected JSON at /root/children/0/fields/artist ("Boo Fighters") does not match
...
+  }
+}
+
+
+Failed 8 of 9 tests:
+testdata/tests/system-test/wrong-bool-value.json: step 1: Unexpected value at /root/coverage/full: true
+testdata/tests/system-test/wrong-element-count.json: step 1: Unexpected number of elements at /root/children: 1
+testdata/tests/system-test/wrong-field-name.json: step 1: Missing expected field at /root/fields/totalCountDracula
+testdata/tests/system-test/wrong-float-value.json: step 1: Unexpected value at /root/children/0/relevance: 0.38186238359951247
+testdata/tests/system-test/wrong-int-value.json: step 1: Unexpected value at /root/fields/totalCount: 1
```
hakonhall, line 319: "Unexpected value" -> "Unexpected response value" may make it slightly clearer? Up to you.

client/go/cmd/testdata/tests/expected-suite.out
```diff
@@ -1,7 +1,9 @@
-Running testdata/tests/system-test/test.json: .... OK!
-Running testdata/tests/system-test/wrong-bool-value.json:
-Failed verifying assertion 0:
-Expected JSON at /root/coverage/full (false) does not match actual (true). Response body:
+Running my test: .... OK
```
hakonhall, line 1: Could you add another successful test right after this?  I just want to verify there is no empty line between two successful tests.

hakonhall (approved):

client/go/cmd/testdata/tests/expected-suite.out
jonmv: Hmm, there is an empty line between all tests. Would you rather it weren't?

client/go/cmd/testdata/tests/expected-suite.out
jonmv: I think it's better without.

client/go/cmd/testdata/tests/expected-suite.out
hakonhall: It's nice with an empty line after an unsuccessful test, because the error message for the test is itself multiline, so it would be difficult to see where the next test begins.

hakonhall: But between successful tests there should be no empty lines? You want

```
Running foo.json: ....ok
Running bar.json: ....ok
Running zoo.json: ....ok
```

PR #20191, opened 2021-11-24
Make and embed vespa cli in vespa-clients RPM

hakonhall:


hakonhall:
The go code is currently explicitly invoked outside of cmake.

PR #20200, opened 2021-11-24
Fix spinner for the 'deploy' command

ldalves:
ref: https://jira.ouryahoo.com/browse/VESPA-22318

### DIFF
--- a/client/go/vespa/deploy.go
+++ b/client/go/vespa/deploy.go
@@ -370,7 +370,12 @@ func uploadApplicationPackage(url *url.URL, opts DeploymentOpts) (int64, error) 
 	if err := opts.Target.PrepareApiRequest(request, sigKeyId); err != nil {
 		return 0, err
 	}
-	response, err := util.HttpDo(request, time.Minute*10, serviceDescription)
+
+	var response *http.Response
+	util.Spinner("Uploading application package ...", func() error {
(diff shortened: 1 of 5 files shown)


client/go/vespa/deploy.go
```diff
@@ -370,7 +370,12 @@ func uploadApplicationPackage(url *url.URL, opts DeploymentOpts) (int64, error)
 	if err := opts.Target.PrepareApiRequest(request, sigKeyId); err != nil {
 		return 0, err
 	}
-	response, err := util.HttpDo(request, time.Minute*10, serviceDescription)
+
+	var response *http.Response
+	util.Spinner("Uploading application package ...", func() error {
```
mpolden, line 375: You must assign the returned error here. `err = util.Spinner ...`

ldalves: `util.Spinner` doesn't return any type:

https://github.com/vespa-engine/vespa/blob/master/client/go/util/spinner.go#L22

If we change that, we'll need to change all other places we use the `Spinner`. 
Do you think that would be a better approach?

ldalves: Ok. I think it is a better approach. Fixed. :)

Tested and everything is working like charm.

mpolden (approved):

PR #20206, opened 2021-11-25
Use git archive to create vespa source tar file.

toregge:
@aressem : please review
@hakonhall : FYI


hakonhall (approved):

aressem (approved):
This should work even on Copr as the vespa/.copr/Makefile installs git.

PR #20207, opened 2021-11-25
add experimental HW-optimized code using intel intrinsics

arnej27959:
* NB: needs -march=westmere or similar

@baldersheim something for you :-)


baldersheim:
Closing as the compiler was able to optimize it itself given the right datatypes.

PR #20226, opened 2021-11-25
- Add optimisation of int8_t squared euclidian distance.

baldersheim:
- Also add a benchmark.

@arnej27959 or @havardpe PR
@vekterli @geirst FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
+++ b/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
@@ -49,9 +49,7 @@ public:
     static const int8_t *cast(const vespalib::eval::Int8Float * p) { return reinterpret_cast<const int8_t *>(p); }
     double calc(const vespalib::eval::TypedCells& lhs, const vespalib::eval::TypedCells& rhs) const override {
         constexpr vespalib::eval::CellType expected = vespalib::eval::get_cell_type<FloatType>();
-        if ((lhs.type != expected) || (rhs.type != expected)) {
-            return SquaredEuclideanDistance::calc(lhs, rhs);
-        }
+        assert(lhs.type == expected && rhs.type == expected);
         auto lhs_vector = lhs.typify<FloatType>();
         auto rhs_vector = rhs.typify<FloatType>();
         size_t sz = lhs_vector.size();
--- a/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
+++ b/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
@@ -44,14 +44,19 @@ public:
         assert(expected_cell_type() == vespalib::eval::get_cell_type<FloatType>());
     }
 
+    static const double *cast(const double * p) { return p; }
+    static const float *cast(const float * p) { return p; }
+    static const int8_t *cast(const vespalib::eval::Int8Float * p) { return reinterpret_cast<const int8_t *>(p); }
     double calc(const vespalib::eval::TypedCells& lhs, const vespalib::eval::TypedCells& rhs) const override {
         constexpr vespalib::eval::CellType expected = vespalib::eval::get_cell_type<FloatType>();
-        assert(lhs.type == expected && rhs.type == expected);
+        if ((lhs.type != expected) || (rhs.type == expected)) {
+            return SquaredEuclideanDistance::calc(lhs, rhs);
+        }
         auto lhs_vector = lhs.typify<FloatType>();
         auto rhs_vector = rhs.typify<FloatType>();
         size_t sz = lhs_vector.size();
         assert(sz == rhs_vector.size());
-        return _computer.squaredEuclideanDistance(&lhs_vector[0], &rhs_vector[0], sz);
+        return _computer.squaredEuclideanDistance(cast(&lhs_vector[0]), cast(&rhs_vector[0]), sz);
     }
 
     double calc_with_limit(const vespalib::eval::TypedCells& lhs,
--- a/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
+++ b/searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
@@ -49,7 +49,7 @@ public:
     static const int8_t *cast(const vespalib::eval::Int8Float * p) { return reinterpret_cast<const int8_t *>(p); }
     double calc(const vespalib::eval::TypedCells& lhs, const vespalib::eval::TypedCells& rhs) const override {
         constexpr vespalib::eval::CellType expected = vespalib::eval::get_cell_type<FloatType>();
-        if ((lhs.type != expected) || (rhs.type == expected)) {
+        if ((lhs.type != expected) || (rhs.type != expected)) {
             return SquaredEuclideanDistance::calc(lhs, rhs);
         }
         auto lhs_vector = lhs.typify<FloatType>();
--- a/vespalib/src/tests/hwaccelrated/hwaccelrated_test.cpp
+++ b/vespalib/src/tests/hwaccelrated/hwaccelrated_test.cpp
@@ -3,6 +3,8 @@
 #include <vespa/vespalib/testkit/test_kit.h>
 #include <vespa/vespalib/hwaccelrated/iaccelrated.h>
 #include <vespa/vespalib/hwaccelrated/generic.h>
+#include <vespa/log/log.h>
+LOG_SETUP("hwaccelrated_test");
 
 using namespace vespalib;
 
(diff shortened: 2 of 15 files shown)


searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
```diff
@@ -44,14 +44,19 @@ class SquaredEuclideanDistanceHW : public SquaredEuclideanDistance {
         assert(expected_cell_type() == vespalib::eval::get_cell_type<FloatType>());
     }
 
+    static const double *cast(const double * p) { return p; }
+    static const float *cast(const float * p) { return p; }
+    static const int8_t *cast(const vespalib::eval::Int8Float * p) { return reinterpret_cast<const int8_t *>(p); }
     double calc(const vespalib::eval::TypedCells& lhs, const vespalib::eval::TypedCells& rhs) const override {
         constexpr vespalib::eval::CellType expected = vespalib::eval::get_cell_type<FloatType>();
-        assert(lhs.type == expected && rhs.type == expected);
+        if ((lhs.type != expected) || (rhs.type == expected)) {
```
havardpe, line 52: without context: The rhs type should be tested for inequality, not equality. with context: This used to be an invariant that always holds because external code would make sure to cast the query vector to the appropriate (expected) type. Is this true also for int8? if so, is that correct? (we should at least perform a saturated conversion, if at all).

vespalib/src/tests/hwaccelrated/hwaccelrated_test.cpp
```diff
@@ -15,26 +17,33 @@ std::vector<T> createAndFill(size_t sz) {
...
+    verifyEuclideanDistance<float, float>(accelrator, testLength);
+    verifyEuclideanDistance<double, float>(accelrator, testLength);
+}
+
 TEST("test euclidean distance") {
     hwaccelrated::GenericAccelrator genericAccelrator;
-    verifyEuclideanDistance<float>(genericAccelrator);
-    verifyEuclideanDistance<double >(genericAccelrator);
+    verifyEuclideanDistance(hwaccelrated::GenericAccelrator(),255);
+    verifyEuclideanDistance(hwaccelrated::IAccelrated::getAccelerator(),255);
```
havardpe, line 46: We should test all supported accelerators, not just the generic and 'best' ones. Ideally, we should also make sure that the build CPUs support 'all' accelerators, to get maximum test coverage.

vespalib/src/vespa/vespalib/hwaccelrated/private_helpers.hpp
```diff
@@ -74,5 +74,31 @@ orChunks(size_t offset, const std::vector<std::pair<const void *, bool>> & src,
...
+    TemporaryT sum = 0;
+    for (size_t i(0); i < sz; i++) {
+        int16_t d = int16_t(a[i]) - int16_t(b[i]);
+        sum += d * d;
+    }
+    return sum;
+}
+
+inline double
+euclideanDistance(const int8_t * a, const int8_t * b, size_t sz) {
```
havardpe, line 92: should use 'squared' in the name here as well

vespalib/src/vespa/vespalib/hwaccelrated/private_helpers.hpp
```diff
@@ -74,5 +74,31 @@ orChunks(size_t offset, const std::vector<std::pair<const void *, bool>> & src,
...
+    constexpr size_t LOOP_COUNT = 0x10000;
+    double sum(0);
+    size_t i=0;
+    for (; i + LOOP_COUNT <= sz; i += LOOP_COUNT) {
+        sum += euclideanDistanceT<int32_t>(a+i, b+i, LOOP_COUNT);
```
havardpe, line 97: consider testing with enough input to also trigger this loop

havardpe (commented):
direct conversion of input tensors to int8 might not be safe (bad things will happen if the values are not in the [-128,127] range).

searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
baldersheim: I only added this one due to a single test only done for Int8Float
at https://github.com/vespa-engine/vespa/blob/c1f3078567a9553bfb9ba98449937cc8850ccc35/searchlib/src/tests/tensor/distance_functions/distance_functions_test.cpp#L74 where it tests euclidian distance between **Int8Float** tensor and a **double** tensor.

vespalib/src/tests/hwaccelrated/hwaccelrated_test.cpp
baldersheim: When an application starts it tests the accelrator chosen for the cpu against the generic one.
It is suboptimal, but the best I can think of. We will never be able to test all anyway.

searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
baldersheim: Logic flaw with rhs fixed.

vespalib/src/vespa/vespalib/hwaccelrated/private_helpers.hpp
baldersheim: Done

baldersheim:
@havardpe PTAL
Consider if the test is necessary mentioned is necessary.

searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
havardpe: By switching from the generic euclidean distance function to the HW version (for int8), I think you have forced the query tensors to be converted to int8 rather than float, making values like '200' become undefined (for HNSW).

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
```diff
@@ -43,7 +43,7 @@ struct SetupFixture
 
 TEST_F("require that blueprint can be created from factory", SetupFixture)
 {
-    EXPECT_TRUE(FTA::assertCreateInstance(f.blueprint, "euclideanDistance"));
+    EXPECT_TRUE(FTA::assertCreateInstance(f.blueprint, "squaredEuclideanDistance"));
```
havardpe, line 46: this seems unintended

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
```diff
@@ -94,20 +94,20 @@ struct ExecFixture
         }
     }
     void setupQueryEnvironment() {
-        test.getQueryEnv().getProperties().add("euclideanDistance.intquery", "[4 5 -6]");
-        test.getQueryEnv().getProperties().add("euclideanDistance.floatquery", "[4.1 15 0.001]");
+        test.getQueryEnv().getProperties().add("squaredEuclideanDistance.intquery", "[4 5 -6]");
```
havardpe, line 97: this seems unintended

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
```diff
@@ -94,20 +94,20 @@ struct ExecFixture
...
+        test.getQueryEnv().getProperties().add("squaredEuclideanDistance.floatquery", "[4.1 15 0.001]");
```
havardpe, line 98: this seems unintended

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
```diff
@@ -94,20 +94,20 @@ struct ExecFixture
...
     }
 
 };
 
 TEST_F("require that distance is calculated for integer vectors",
-        ExecFixture("euclideanDistance(aint,intquery)"))
+        ExecFixture("squaredEuclideanDistance(aint,intquery)"))
```
havardpe, line 104: this seems unintended

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
```diff
@@ -94,20 +94,20 @@ struct ExecFixture
...
 {
     EXPECT_TRUE(f.test.execute(11.789826, 0.000001));
 }
 
 TEST_F("require that distance is calculated for floating point vectors",
-        ExecFixture("euclideanDistance(afloat,floatquery)"))
+        ExecFixture("squaredEuclideanDistance(afloat,floatquery)"))
```
havardpe, line 110: this seems unintended

searchlib/src/vespa/searchlib/features/euclidean_distance_feature.cpp
```diff
@@ -52,7 +52,7 @@ EuclideanDistanceExecutor<DataType>::execute(uint32_t docId)
 
 
 EuclideanDistanceBlueprint::EuclideanDistanceBlueprint() :
-    Blueprint("euclideanDistance"),
+    Blueprint("squaredEuclideanDistance"),
```
havardpe, line 55: this seems unintended

havardpe (commented):
The fallback check is probably not needed right now, but might be useful if/when changing the code to not break float input vectors.

searchlib/src/vespa/searchlib/tensor/euclidean_distance.h
baldersheim: Now we are enforcing input both lhs and rhs to have same and expected type.
Test for mixed double and int8_t has been removed.

searchlib/src/tests/features/euclidean_distance/euclidean_distance_test.cpp
baldersheim: Yes, clion was not smart enough, or too smart for me.

PR #20228, opened 2021-11-25
Generate lb services config with systemname and weights

tokle:
- Add system name
- Support container endpoint weights

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ContainerEndpoint.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ContainerEndpoint.java
@@ -3,6 +3,9 @@ package com.yahoo.config.model.api;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
 
 /**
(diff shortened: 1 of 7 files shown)


config-model-api/src/main/java/com/yahoo/config/model/api/ContainerEndpoint.java
```diff
@@ -16,11 +18,17 @@
     private final String clusterId;
     private final ApplicationClusterEndpoint.Scope scope;
     private final List<String> names;
+    private final OptionalLong weight;
```
mpolden, line 21: Why long?

mpolden (approved):

mpolden (approved):

hmusum:
This broke config model api (but cannot be reverted since this PR also adds new methods):

 Caused by: java.lang.NoSuchMethodError: 'com.yahoo.config.model.api.ApplicationClusterEndpoint$DnsName com.yahoo.config.model.api.ApplicationClusterEndpoint$DnsName.sharedL4NameFrom(com.yahoo.config.provision.ClusterSpec$Id, com.yahoo.config.provision.ApplicationId, java.lang.String)'

PR #20241, opened 2021-11-26
Add query item limit

olaaun:

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/query/QueryCanonicalizer.java
+++ b/container-search/src/main/java/com/yahoo/prelude/query/QueryCanonicalizer.java
@@ -51,7 +51,7 @@ public class QueryCanonicalizer {
         ListIterator<Item> rootItemIterator = query.getItemIterator();
         CanonicalizationResult result = recursivelyCanonicalize(rootItemIterator.next(), rootItemIterator);
         if (query.isEmpty() && ! result.isError()) result = CanonicalizationResult.error("No query");
-        if (query.getTreeSize() > maxQueryItems) result = CanonicalizationResult.error("Query tree exceeds allowed item count");
+        if (query.getTreeSize() > maxQueryItems) result = CanonicalizationResult.error("Query tree exceeds allowed item count. Configured item limit: " + maxQueryItems);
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/prelude/query/QueryCanonicalizer.java
```diff
@@ -19,25 +20,38 @@
...
      * Canonicalizes this query
      * 
      * @return null if the query is valid, an error message if it is invalid
      */
-    public static String canonicalize(QueryTree query) {
+    private static String canonicalize(QueryTree query, Integer maxQueryItems) {
         ListIterator<Item> rootItemIterator = query.getItemIterator();
         CanonicalizationResult result = recursivelyCanonicalize(rootItemIterator.next(), rootItemIterator);
         if (query.isEmpty() && ! result.isError()) result = CanonicalizationResult.error("No query");
+        if (query.getTreeSize() > maxQueryItems) result = CanonicalizationResult.error("Query tree exceeds allowed item count");
```
geirst, line 54: Please add configured item count to the error message.

baldersheim:
@geirst PTAL

container-search/src/main/java/com/yahoo/search/query/QueryTree.java
```diff
@@ -179,4 +180,23 @@ private static void getPositiveTerms(Item item, List<IndexedItem> terms) {
         }
     }
 
+    /**
+     * @return The total number of items in this query tree
```
bratseth, line 184:
```suggestion
     * Returns the total number of items in this query tree.
```

container-search/src/main/java/com/yahoo/search/query/QueryTree.java
```diff
@@ -179,4 +180,23 @@ private static void getPositiveTerms(Item item, List<IndexedItem> terms) {
...
+     */
+    public int getTreeSize() {
```
bratseth, line 186: This isn't a getter so I think you should use treeSize(), or maybe countItems() to communicate that this isn't cheap.

bratseth (approved):
👍

PR #20247, opened 2021-11-26
internal api annotation

tokle:
To better be able to catch incompatible changes in config-model-api.

Suggestion: 
* Add new `@InternalApi` annotation
* Run abi-check-plugin twice in modules that contain both internal and public api (i.e. config-model-api)


hmusum (approved):

bjorncs (commented):
Consider checking for `@ExportPackage` instead of introducing a new annotation.

tokle:
After offline discussion with @bjorncs:
* Let abi-check-plugin create spec-file for multiple annotations
* Re-use _@ExportPackage_ annotation instead of new annotation.

PR #20253, opened 2021-11-27
Revert "read routingmethod"

hmusum:
Reverts vespa-engine/vespa#20250

App in cd fails to deploy with:

Failed to prepare application: No enum constant com.yahoo.config.model.api.ApplicationClusterEndpoint.RoutingMethod.exclusive


arnej27959 (approved):
seems legit 👍

PR #20255, opened 2021-11-27
Convert from double to signed data type for reference hamming distance.

toregge:
@aressem : please review
@arnej27959 , @havardpe : FYI


arnej27959 (approved):
looks OK, but could 'x' and 'y' just be int8_t, to make the code simpler?

toregge:
`__builtin_popcount` expects an unsigned int argument, thus we would likely need to pass `(uint8_t) (x ^ y)` to `__builtin_popcount` to avoid an off by 24 error if `x` and `y` were of type `int8_t`.

aressem (approved):

PR #20282, opened 2021-11-30
Recompute change to avoid setting a change which triggers no jobs

jonmv:


jonmv:
This avoid the scenario where someone forces a no-op change; this never triggers any jobs, and thus never gets cleared.

mpolden:
LGTM, but there's a failing test.

mpolden (approved):

PR #20283, opened 2021-11-30
Override GOPROXY if using default direct proxy

hakonhall:


toregge (approved):

hakonhall:
This made make -C client/go complete in 9 seconds instead of the typical >3 minutes.

mpolden:
Is this because SD uses an older Go compiler which doesn't use proxy.golang.org by default?

hakonhall:
yes

PR #20296, opened 2021-11-30
Reduce default bucket delete priority and fix priority inheritance for merge-triggered deletion [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Since deletes are now async in the backend, make them FIFO-order with
client feed by default to avoid swamping the queues with deletes.
Also, explicitly inherit priority for bucket deletion triggered by
bucket merging. This was actually missing previously and meant such
deletes got the default, very low priority.


geirst (approved):
👍

PR #20297, opened 2021-11-30
Create FileDownloader once [run-systemtest]

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
```diff
@@ -64,49 +64,55 @@ protected double maintain() {
...
+        else
+            connectionPool = useFileDistributionConnectionPool
+                    ? new FileDistributionConnectionPool(configSourceSet, supervisor)
+                    : new JRTConnectionPool(configSourceSet, supervisor);
+        return new FileDownloader(connectionPool, supervisor, downloadDirectory, Duration.ofSeconds(30));
     }
 
     @Override
     public void awaitShutdown() {
-        supervisor.transport().shutdown().join();
```
gjoranv, line 109: Don't we have to shutdown the transport thread now?

gjoranv (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
hmusum: It's done by `FileDownloader.close()`, but I agree this does not look good. Putting it back again, I'll look into other users of FileDownloader and see if this can be done better in another PR

gjoranv: Thanks, it looks much better now.

gjoranv (approved):
👍

PR #20298, opened 2021-11-30
Account for capacity policies when autoscaling

bratseth:


bratseth:
The changes that matters here are the ones in AllocatableClusterResources.from(...)

And the purpose is to get the new behavior demonstrated in the two new tests in AutoscalingTest.

freva (approved):

hmusum:
Looks like this broke code in internal repo

PR #20303, opened 2021-11-30
Don't let ignored bucket info reply be propagated out of (top-level) distributor

vekterli:
@geirst please review

If a reply arrives for a preempted cluster state it will be ignored.
To avoid it being automatically sent further down the storage chain
we still have to treat it as handled. Otherwise a scary looking but
otherwise benign "unhandled message" warning will be emitted in the
Vespa log.

Also move an existing test to the correct test fixture.


geirst (approved):
👍

baldersheim (approved):

PR #20310, opened 2021-12-01
Exit vespa cli with non-zero exit code on error

hakonhall:


client/go/cmd/vespa/main.go
```diff
@@ -6,8 +6,11 @@ package main
 
 import (
 	"github.com/vespa-engine/vespa/client/go/cmd"
+	"os"
 )
 
 func main() {
-	cmd.Execute()
+	if cmd.Execute() != nil {
```
mpolden, line 13: Nit: `if err := ...; err != nil {` is a common idiom to indicate that the returned value is in fact an error.

mpolden (approved):

client/go/cmd/vespa/main.go
mpolden: Too bad `cobra.Command{Run: ...}` does not return an error, because then we could pass all errors up to here.

hakonhall: Which kind of errors are masked by cobra.Command?

hakonhall: Fixed

PR #20316, opened 2021-12-01
Put node ID in the ID field

mpolden:
I did not find any important usages of `id`, so let's use it for the node ID,
and eventually remove `openStackId`.

@freva


mpolden:
Hold merge. Noticed that we should read `id` when adding nodes as well.

freva (approved):
:1st_place_medal:

freva:
It would be nice if nodes had unique IDs across allocations, then we could get rid of this hack in node-admin: https://github.com/vespa-engine/vespa/blob/master/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminImpl.java#L227

mpolden:
Agree, but that sounds independent of this change.

freva (approved):

PR #20323, opened 2021-12-01
Deprecate DocumentTypeManager and DocumentTypeManagerConfigurer. They…

baldersheim:
… should not be used directly by applications.

We should make a new one with an interface that preserves the type hierachy.

@arnej27959 and @bratseth @jonmv PR

### DIFF
--- a/application/src/main/java/com/yahoo/application/container/DocumentProcessing.java
+++ b/application/src/main/java/com/yahoo/application/container/DocumentProcessing.java
@@ -23,6 +23,7 @@ import java.util.Map;
  * @author Einar M R Rosenvinge
  */
 @Beta
+@SuppressWarnings("deprecation")
 public final class DocumentProcessing {
 
     private final DocumentProcessingHandler handler;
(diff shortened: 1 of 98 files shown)


arnej27959:
I agree that this shouldn't be used from users' code; but I think the Jon's will have to agree also.

jonmv:
Shouldn't the replacement be available at the time off deprecation, though?

baldersheim:
There will be no replacement that is public.
User are supposed to have the thing you made some sprints ago injected.

bratseth:
Also agree, but what should they use? The @deprecation Javadoc comment should say.

arnej27959:
I don't know what "thing you made some sprints ago" is, but in system tests I see that document processors
typically use `processing.getService().getDocumentTypeManager().getDocumentType(...)`
We should have something, possibly an interface with `getDocumentType(String name)` as the only method,
which can be used instead of DocumentTypeManager here.

jonmv:
If they use the document manager to access document meta data, the injectable document access won't help with that. It gives them a fully configure DocumentManager, but if they're not allowed to use it, that doesn't help much. Perhaps the `getDocumentType(String name)` should be added to this, then?

PR #20329, opened 2021-12-02
Arnej/config class should not be public

arnej27959:

@bratseth please review
@baldersheim @hmusum @toregge FYI


bratseth (approved):
👍

document/src/vespa/document/config/doctypecfg.h
```diff
@@ -0,0 +1,16 @@
...
+ * DocumenttypesConfig and DocumenttypesConfigBuilder
+ * (globally visible).
+ **/
+
+namespace document::config::internal {
+class InternalDocumenttypesType;
+}
+
+using DocumenttypesConfigBuilder = document::config::internal::InternalDocumenttypesType;
+using DocumenttypesConfig = const document::config::internal::InternalDocumenttypesType;
```
toregge, line 16: Consider moving the using statements to the `document::config` namespace, i.e. the same namespace that is used when the generated `config-documenttypes.h` is included.

Consider renaming the header file to something containing fwd, e.g. as `iosfwd` contains forward declarations for `iostream`.

toregge (approved):

arnej27959:
> Consider moving the using statements to the document::config namespace

I tried that first, but it felt like more work for less consistent and less readable code.

toregge:
> I tried that first, but it felt like more work for less consistent and less readable code.

But that means that `config-documenttypes.h` won't work as a drop-in replacement for `documenttypes_config_fwd.h`, you might need both if you depend on the using being in the global namespace and you need the actual config definition.

PR #20340, opened 2021-12-02
Cap merge-induced DeleteBucket priority to that of default feed priority

vekterli:
@geirst please review
@baldersheim FYI

This lets `DeleteBucket` operations FIFO with the client operations using the default feed priority (120).

Not doing this risks preempting feed ops with deletes, elevating latencies.


geirst (approved):
👍

PR #20351, opened 2021-12-03
Auditref is not always present

bjorncs:


andreer:
We should not allow that. But we should at least not throw a NPE so this is still an improvement

andreer (approved):

PR #20358, opened 2021-12-03
Ensure that we do not get more resizing than necessary.

baldersheim:
@geirst PR


geirst (approved):
👍

searchlib/src/vespa/searchlib/docstore/writeablefilechunk.cpp
```diff
@@ -484,11 +484,19 @@ void
 WriteableFileChunk::updateChunkInfo(const ProcessedChunkQ & chunks, const ChunkMetaV & cmetaV, size_t sz)
 {
     std::lock_guard guard(_lock);
+    uint32_t maxChunkId(0);
+    for (const auto & chunk : chunks) {
+        maxChunkId = std::max(chunk->getChunkId(), maxChunkId);
+    }
```
vekterli, line 490: Could this be done prior to the lock being taken?

baldersheim: Yes, #20360

PR #20375, opened 2021-12-06
Bag of mixed candy

jonmv:


client/go/cmd/testdata/tests/expected-suite.out
```diff
@@ -1,8 +1,11 @@
-Running my test: .... OK
-Running wrong-bool-value.json: Failed step 1:
-Unexpected value at /root/coverage/full: true
+My test: .... OK
+wrong-bool-value.json: failed
+Step 1:
```
hakonhall, line 3: `In step 1`  may be better?

client/go/cmd/testdata/tests/expected-suite.out
```diff
@@ -1,8 +1,11 @@
...
+Unexpected value at /root/coverage/full
```
hakonhall, line 4: Unexpected value **in response** at ...etc ?

hakonhall (approved):

client/go/cmd/testdata/tests/expected-suite.out
jonmv: Hmm, yes, but it doesn't really match `<step name>"` ...

client/go/cmd/testdata/tests/expected-suite.out
jonmv: Reponse has unexpected value at ... ?

jonmv: Otherwise the "unexpeccted value" becomes too detached from the relevant path.

jonmv: But that's just wordy, again ...

hakonhall: LGTM

hakonhall: Or perhaps `In step 1` without name, and `In step 1: <name>` with name?  That way people also learn what the step number means.

jonmv: That's a great idea!

jonmv: Does it make sense to do this with files, too?

hakonhall: How would that look?

jonmv: ```feed-and-search-test.json: Test feeding and searching basics: ... OK```

jonmv: Perhaps too much.

hakonhall: I suggest the stem of the filename **is** the test name, which avoids the duality and possibly duplicated prefix (<filename>:<testname>).  But we can revisit this later.

PR #20381, opened 2021-12-06
Remove PublicApi annotation for jdisc.service and jdisc.test packages.

gjoranv:
They do not contain any classes that are appropriate for external usage. Not currently used by any hosted Vespa apps.


gjoranv:
We can't make jdisc.service non-public yet. Must at least remove references from doc.

PR #20384, opened 2021-12-06
proposal for new documentmanager config

arnej27959:

@baldersheim for discussion

I have written some code to read this in DocumentTypeManagerConfigurer but that code needs some testing first, so I need to write the corresponding code in config-model to generate config to test it with also...

But it seems so far that this can work as the minimal amount of config we need.


arnej27959:
@baldersheim this is now ready for review

baldersheim (approved):

PR #20405, opened 2021-12-07
Prevent orphaned bucket replicas caused by indeterminate CreateBucket replies [run-systemtest]

vekterli:

Previously we'd implicitly assume a failed `CreateBucket` reply meant the
bucket replica was not created, but this does not hold in the general case.
A failure may just as well be due to connection failures etc between the
distributor and content node. To tell for sure, we now send an explicit
`RequestBucketInfo` to the node in the case of `CreateBucket` failures. If
it _was_ created, the replica will be reintroduced into the bucket DB.

We still implicitly delete the bucket replica from the DB to avoid
transiently routing client write load to a bucket replica that may likely not
exist. Any write load will implicitly re-create the replica if needed.

### DIFF
--- a/storage/src/tests/distributor/putoperationtest.cpp
+++ b/storage/src/tests/distributor/putoperationtest.cpp
@@ -51,9 +51,8 @@ public:
     document::BucketId createAndSendSampleDocument(vespalib::duration timeout);
 
     void sendReply(int idx = -1,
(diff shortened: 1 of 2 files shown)


toregge (approved):

baldersheim (approved):

geirst (commented):
👍

PR #20417, opened 2021-12-08
Deprecate methods using Guava ListenableFuture [run-systemtest]

bjorncs:
- com.yahoo.processing.Response.recursiveComplete()
- com.yahoo.processing.response.DataList.complete()
- com.yahoo.processing.response.IncomingData.completed()

Also fixes bug in FutureResponse where `get()` does not return value produced by async task.

### DIFF
--- a/container-core/src/main/java/com/yahoo/processing/response/DataList.java
+++ b/container-core/src/main/java/com/yahoo/processing/response/DataList.java
@@ -72,9 +72,9 @@ public interface DataList<DATATYPE extends Data> extends Data {
      * Making this call on a list which does not support future data always returns immediately and
      * causes no memory synchronization cost.
      */
-    CompletableFuture<DataList<DATATYPE>> future();
+    CompletableFuture<DataList<DATATYPE>> completeFuture();
 
-    /** @deprecated Use {@link #future()} instead */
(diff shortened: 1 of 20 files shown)


container-core/src/main/java/com/yahoo/processing/response/DataList.java
```diff
@@ -73,6 +72,10 @@
      * Making this call on a list which does not support future data always returns immediately and
      * causes no memory synchronization cost.
      */
+    CompletableFuture<DataList<DATATYPE>> future();
```
bratseth, line 75: Isn't "future" here (and the other places) a bit meaningless? It's a future where ...?

bratseth: How about completed()? (Or completeFuture())

bratseth (commented):
Great!

container-core/src/main/java/com/yahoo/processing/response/DataList.java
bjorncs: Good point - I agree. Your second proposal matches well with the existing method. I'll rename.

bjorncs:
@bratseth PTAL

bratseth (approved):
👍

PR #20420, opened 2021-12-08
Add CompactionSpec.

toregge:
Pass CompactionStrategy to DataStoreBase::startCompactWorstBuffers in
preparation for compacting more buffers in the same compaction round.

@geirst : please review
@baldersheim : FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/datastore/compaction_spec.h
+++ b/vespalib/src/vespa/vespalib/datastore/compaction_spec.h
@@ -16,13 +16,13 @@ class CompactionSpec
     bool _compact_memory;
     bool _compact_address_space;
 public:
-    CompactionSpec(bool compact_memory, bool compact_address_space)
-        : _compact_memory(compact_memory),
-          _compact_address_space(compact_address_space)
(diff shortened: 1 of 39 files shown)


vespalib/src/vespa/vespalib/datastore/compaction_spec.h
```diff
@@ -0,0 +1,28 @@
...
+{
+    bool _compact_memory;
+    bool _compact_address_space;
+public:
+    CompactionSpec(bool compact_memory, bool compact_address_space)
+        : _compact_memory(compact_memory),
+          _compact_address_space(compact_address_space)
+    {
+    }
+    bool get_compact_memory() const noexcept { return _compact_memory; }
```
geirst, line 24: As this function returns a bool, I think it reads better when removing the _get__ prefix. Applies to the function below as well.

geirst (approved):
👍

vespalib/src/vespa/vespalib/datastore/compaction_spec.h
toregge: Fixed.

geirst (approved):
👍

PR #20427, opened 2021-12-09
Add validation override for 'paged' removal

bratseth:


config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/search/AttributeChangeValidatorTest.java
```diff
@@ -202,4 +209,20 @@ public void changing_hnsw_index_property_neighbors_to_explore_at_insert_requires
...
                                                   "'neighbors-to-explore-at-insert' from '200' to '100'"));
     }
+
+    @Test
+    public void removing_paged_requires_override() throws Exception {
+        try {
+            new Fixture("field f1 type tensor(x[10]) { indexing: attribute \n attribute: paged }",
+                        "field f1 type tensor(x[10]) { indexing: attribute  }").
+                    assertValidation(newRestartAction(ClusterSpec.Id.from("test"),
+                                                      "Field 'f1' changed: add attribute 'huge'"));
```
geirst, line 219: Consider changing this string to something indicating that is doesn't matter. Empty string would be just fine.

geirst (approved):
👍

config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/search/AttributeChangeValidatorTest.java
bratseth: Fixed - thanks!

PR #20431, opened 2021-12-09
Arnej/add builder lambda convenience

arnej27959:

@gjoranv please review
@hmusum @baldersheim FYI


gjoranv (approved):
It would be nice if a few unit tests were updated to use the new method. Perhaps it's even scriptable?

arnej27959:
> It would be nice if a few unit tests were updated to use the new method.

Agreed, which unit tests did you have in mind?
Looking at .def files, all of these seem like candidates:

config-lib/src/test/resources/configdefinitions/test.function-test.def
config/src/test/resources/configdefinitions/function-test.def
config/src/test/resources/configs/def-files/function-test.def

gjoranv:
How about

- ConfigInstanceBuilderTest.createVariableAccessBuilder
- ConfigInstanceUtilTest

gjoranv:
Great!

PR #20436, opened 2021-12-09
Support glob-style credential matching of SAN_URI certificate fields [run-systemtest]

vekterli:
@bjorncs please review

This is much like the `DNS_SAN` matching, but with two major differences:
 * Implicit delimiting around `/` characters instead of `.` characters.
 * Only wildcard `*` globbing is supported. `?` may be present in a valid
   URI (even though it's silly for this purpose) and is matched as a literal
   character instead of _any_ single char.


bjorncs (approved):

bjorncs:
FYI @tokle

PR #20439, opened 2021-12-09
Add zone info to query profile context

bratseth:


bratseth:
Ping @arnej27959

arnej27959 (approved):
looks good to me

PR #20440, opened 2021-12-09
Remove use of Guava ListenableFuture from com.yahoo.jdisc.handler [run-systemtest]

bjorncs:
This change is not 100% API compatible. Many classes from this package inherited types from Guava.
The classes will have the same methods as before, but their type has obviously changed.
Two options; merge now with the small risk of breakage or wait for Vespa 8 branch.


bratseth (approved):
I think it's close enough :shipit:

gjoranv (approved):
Low risk

PR #20449, opened 2021-12-10
deprecate parts of DocumentTypeManager

arnej27959:
* we want to stop using IDs as unique key for DataTypes;
  these methods are only used in our own config setup.
* nothing should setup a self-subscribing manager; add a
  convenience to create a manager from a config file
  to use from unit tests.

@bratseth please review

@gjoranv @bjorncs @jonmv @baldersheim FYI

### DIFF
--- a/document/src/main/java/com/yahoo/document/DocumentTypeManagerConfigurer.java
+++ b/document/src/main/java/com/yahoo/document/DocumentTypeManagerConfigurer.java
@@ -159,7 +159,7 @@ public class DocumentTypeManagerConfigurer implements ConfigSubscriber.SingleSub
             return type;
         }
 
-
+        @SuppressWarnings("deprecation")
         private DataType getOrCreateType(int id) {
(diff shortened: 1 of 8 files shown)


jonmv (approved):

document/src/main/java/com/yahoo/document/DocumentTypeManagerConfigurer.java
```diff
@@ -159,7 +159,7 @@ private DataType createReferenceType(int id, DocumentmanagerConfig.Datatype.Refe
             return type;
         }
 
-
+        @SuppressWarnings("deprecation")
```
baldersheim, line 162: If you deprecate the DocumentManagerConfigurer you do not need supressions.
It should not be public either.
But that can also be done separately.

baldersheim (approved):

bratseth (approved):
Great!

PR #20452, opened 2021-12-10
Let accounting handle multiple commits in flight.

baldersheim:
@toregge and @geirst PR


toregge (approved):

geirst (approved):
👍  The complexity of tracking num operations and commits has increased, and I think it is harder to read the FeedHandler functions where updating of stats happens. I suggest putting the following variables into a separate class with a suitable API and let FeedHandler use an instance of this class:
```
_numOperationsStarted
_numOperationsCompleted
_numOperationsStartedAtLastCommitStart
_numCommitsStarted
_numCommitsCompleted
```

baldersheim:
Agree, will do in a separate PR

PR #20463, opened 2021-12-11
Include signal.h to get declarations of kill() and killpg() on Darwin.

toregge:
@baldersheim : please review

### DIFF
--- a/vespalog/src/logger/runserver.cpp
+++ b/vespalog/src/logger/runserver.cpp
@@ -5,7 +5,7 @@
 #include <fcntl.h>
 #include <cerrno>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


vespalog/src/logger/runserver.cpp
```diff
@@ -5,6 +5,7 @@
 #include <fcntl.h>
 #include <cerrno>
 #include <unistd.h>
+#include <signal.h>
```
baldersheim, line 8: csignal ?

toregge: Fixed.

baldersheim (approved):

PR #20470, opened 2021-12-11
gcc 9 does not support std::ranges::all_of.

toregge:
@baldersheim : please review


baldersheim (approved):
Time to require gcc 10 ?

toregge:
Perhaps. Ubuntu 20.04 uses gcc 9 and Ubuntu 22.04 has not yet been released.

PR #20476, opened 2021-12-12
Add exponential backoff if starting download fails [run-systemtest]

hmusum:


filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReferenceDownloader.java
```diff
@@ -75,6 +75,16 @@ private void waitUntilDownloadStarted(FileReferenceDownload fileReferenceDownloa
         downloads.remove(fileReference);
     }
 
+    private void backoff(int retryCount) {
+        if (retryCount > 0) {
+            try {
+                Thread.sleep((long) (Math.pow(2, retryCount)) * sleepBetweenRetries.toMillis());
```
baldersheim, line 81: I think you need a a maximum backoff here.

hmusum: `retryCount` is max. 4, but maybe it's best to have a max if that changes. Will fix.

baldersheim (approved):

PR #20488, opened 2021-12-13
Arnej/reapply produce new config

arnej27959:

@baldersheim please review


config-model/src/main/java/com/yahoo/vespa/configmodel/producers/DocumentManager.java
```diff
@@ -176,4 +196,271 @@ private void buildImportedFieldsConfig(Collection<String> fieldNames, Datatype.D
...
+                docTypeInheritOrder(inherited, builder, produced, indexMap);
+            }
+            docTypeBuild(documentType, builder, indexMap);
+            produced.put(documentType.getFullName(), documentType);
+        }
+    }
+
+    static private class IdxMap {
+        private Map<Integer, Boolean> doneMap = new HashMap<>();
+        private Map<Object, Integer> map = new IdentityHashMap<>();
```
baldersheim, line 229: If you are using IdentityHashMap I assume that it is not used for iteration where order matters.

baldersheim (approved):

PR #20491, opened 2021-12-13
Revert "Treat empty replica subset as inconsistent for GetOperation [run-systemtest]"

vekterli:
@geirst please review.

The original change introduced a subtle edge case in the 3-phase update metadata handling code (**NOT** enabled by default) where we'd trigger an assertion if the following held:

* The bucket has >= 2 replicas
* At least one replica is empty, but not all are (this is rare)
* The document cannot be found in any of the non-empty replicas

The 3-phase update code would assume that if the `GetOperation` indicated replica inconsistency, this would necessarily mean that at least 1 timestamp had been observed and assert on this being the case. But the change introduced would initially tag a `GetOperation` as inconsistent if at least one replica was empty, regardless of any timestamps being observed.

Will create an alternative fix, likely situated on the _outside_ of `GetOperation` instead.

Reverts vespa-engine/vespa#20382

### DIFF
--- a/storage/src/tests/distributor/getoperationtest.cpp
+++ b/storage/src/tests/distributor/getoperationtest.cpp
@@ -267,26 +267,6 @@ TEST_F(GetOperationTest, send_to_all_invalid_nodes_when_inconsistent) {
     EXPECT_EQ("newauthor", getLastReplyAuthor());
 }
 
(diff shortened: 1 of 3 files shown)


geirst (approved):
👍

PR #20495, opened 2021-12-13
Use FleetControllerContext in ZooKeeperDatabase

hakonhall:


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetControllerContext.java
```diff
@@ -16,4 +16,13 @@
     default void log(Logger logger, Level level, String message) { log(logger, level, () -> message); }
     void log(Logger logger, Level level, String message, Throwable t);
     void log(Logger logger, Level level, Supplier<String> message);
+
+    default void log(Logger logger, Level level, String format, Object first, Object... rest) {
```
vekterli, line 20: Can this cause any overloading ambiguities with the `Throwable` overload above?

vekterli (approved):
:+1:

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetControllerContext.java
hakonhall: Yes it can, good point.  There's no compilation error.  So the only issue would be if someone had a subclass of Throwable, but stored it in an Object reference (or cast it to Object), and wanted to invoke the method with Throwable parameter.

I think the only way out would be to name the new method something else.  Could name it logf(), but :shrug:

PR #20501, opened 2021-12-14
Log JVM options [run-systemtest]

hmusum:
Prepare for validating JVM options. This PR starts logging them at info level to deploy log. This way we can gather information about which options are in use and how/when we can validate and eventually fail deployment for some of these options


mpolden (approved):
👍

PR #20504, opened 2021-12-14
Arnej/struct inherit did not work

arnej27959:

@baldersheim please review
@bratseth FYI


config-model/src/main/java/com/yahoo/searchdefinition/TemporarySDTypeResolver.java
```diff
@@ -0,0 +1,77 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.searchdefinition;
+
+import com.yahoo.config.application.api.DeployLogger;
+import com.yahoo.searchdefinition.document.SDDocumentType;
+import com.yahoo.searchdefinition.document.TemporarySDDocumentType;
+
+import java.util.*;
```
baldersheim, line 8: Avoid * imports

baldersheim (approved):

bratseth:
Great work in this PR, but something has gone very wrong in this area when all this complexity is needed ...

baldersheim:
Definitely, around 8-10 years ago..., and from then it only got worse :)

PR #20507, opened 2021-12-14
Dump pending GC state for nodes when update inconsistency is detected

vekterli:
@geirst please review

This is to help debug a very rare edge case where the theory is that
an update operation may race with the implicit removal of said document
by asynchronous GC. By dumping the pending GC state for the bucket+node
we can get some good indications on whether this theory holds.


geirst (approved):
👍

PR #20508, opened 2021-12-14
Failing to find zk system state aborts tick [run-systemtest]

hakonhall:


vekterli (approved):
:+1:  I presume aborting the tick is due to the added re-throw of `KeeperException` -> `RuntimeException`?

hakonhall:
Yes

PR #20512, opened 2021-12-14
minimal implementation of "ignore some user agents"

arnej27959:

with some guessing I think I have identified where to plug in some sort of "ignore specific user agents" feature, but need some feedback:

1) is this the metric point we are looking for?
2) is it OK to just ignore these requests WRT metrics?
3) does the use of jetty "Request" object seem legit?
4) do we need some sort of pattern-matching or is string "equals" OK?

@frodelu @andreer @oyving @bjorncs @gjoranv please comment if you have any input


gjoranv (commented):
Please try to make all new methods package-private, to avoid expanding the public API. Otherwise, LGTM.

andreer:
I crunched some logs. Excluding two fixed user agent strings should remove ~60% of non-200 responses so it's probably a good start.

andreer (approved):
Makes sense to me. Let's configure it and see!

PR #20513, opened 2021-12-14
Simplify by allowing only constants in maps and arrays at grammar level

jonmv:
@bratseth please review.


jonmv:
Constants means parameters, literals, arrays and maps.

bratseth (approved):

PR #20517, opened 2021-12-14
1 - Make the first sample period at least 5s long.

baldersheim:
2 - Ensure that total cpu usage is sampled first on startup, and last on all remaining samples.
    This is to avoid the large skew you get when sample interval is shorter than the time used to collect the samples.
This should eliminate the sampling noise when metricsproxy is started.

@bjorncs or @gjoranv


bjorncs (approved):

gjoranv (commented):
👍

PR #20519, opened 2021-12-14
Bratseth/match exactly

bratseth:
### Document selection type matching and document type inheritance

When evaluating the document selection expression "myType" on a document, we have the choice of returning _true_ only if the document is of type myType (_exact matching_), or if it's of any subtype of "myType" (_isa matching_).

The current state is this:
- In Java, where selection expressions are used for routing: Exact matching is used for _Get_ and _Remove_ document operations, and isa matching for _Put_ and _Update_ operations.
- In C++, where selection expressions are used for garbage collection and "where" expressions selecting or updating documents: Isa matching is always used.
- The documentation does not specify one way or the other.
- Internal inconsistency: Content clusters require all exact types to have a document db to be configured in the cluster, which means that adding a subtype to a document type configured in a content cluster will lead document writes of the subtype to fail since the writes are routed to that cluster but there is no document db for it.

We need to choose either exact or isa matching and implement it consistently, and make any changes needed either before or at Vespa 8 depending on compatibility requirements.

Some users will want exact matching, while others will appreciate the convenience of sust adding subtypes and having them automatically work like the parent type. If we choose exact matching and the user want the same behavior for each subtype it is achieved by adding each type to any content cluster and selection expression. If we choose isa matching and user does not want the same behavior for each subtype though, it is harder to achieve - we would need additional syntax.

Therefore we should choose exact matching.

### This PR 

This PR prepares changing to always use exact matching in Java (commit #1-3), but preserves the current behavior with Vespa 8 TODO's (commit #4). In C++ the change seems straightforward so this only adds the TODO.

@jonmv AND @vekterli AND @baldersheim please review the above description.
@jonmv please review the code changes.

### DIFF
--- a/document/src/main/java/com/yahoo/document/select/rule/AttributeNode.java
+++ b/document/src/main/java/com/yahoo/document/select/rule/AttributeNode.java
@@ -27,18 +27,11 @@ import java.util.List;
 public class AttributeNode implements ExpressionNode {
 
     private ExpressionNode value;
-    private final List<Item> items = new ArrayList<>();
+    private final List<Item> items;
 
-    public AttributeNode(ExpressionNode value, List items) {
+    public AttributeNode(ExpressionNode value, List<Item> items) {
         this.value = value;
-        for (Object obj : items) {
-            if (obj instanceof Item) {
-                this.items.add((Item)obj);
-            } else {
-                throw new IllegalStateException("Can not add an instance of " + obj.getClass().getName() +
-                                                " as a function item.");
-            }
-        }
+        this.items = new ArrayList<>(items);
     }
 
     public ExpressionNode getValue() {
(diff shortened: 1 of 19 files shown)


bratseth:
@gjoranv FYI

baldersheim (approved):
Fully agree.

vekterli:
Agreed. There is already way too much implicit behavior in the document selection language, which causes surprises for users and backend developers alike... Could add `isa`/`instanceof` as an explicit document selection function or a separate binary operator if we want the legacy functionality.

document/src/main/java/com/yahoo/document/select/rule/AttributeNode.java
```diff
@@ -27,18 +27,11 @@
...
+    private final List<Item> items;
 
-    public AttributeNode(ExpressionNode value, List items) {
+    public AttributeNode(ExpressionNode value, List<Item> items) {
         this.value = value;
-        for (Object obj : items) {
-            if (obj instanceof Item) {
-                this.items.add((Item)obj);
-            } else {
-                throw new IllegalStateException("Can not add an instance of " + obj.getClass().getName() +
```
jonmv, line 38: 😀

document/src/vespa/document/select/doctype.cpp
```diff
@@ -15,6 +15,7 @@ namespace {
                                 vespalib::stringref name)
     {
         if (type.getName() == name) return true;
+        // TODO Vespa 8: Remove this for look on Vespa 8
```
jonmv, line 18:
```suggestion
        // TODO Vespa 8: Remove this for loop on Vespa 8
```

jonmv (approved):
LGTM!

PR #20521, opened 2021-12-15
Add metric for overdue upgrades

mpolden:
@jonmv


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/MetricsReporterTest.java
```diff
@@ -494,6 +497,60 @@ public void tenant_counter() {
...
+               .runJob(stagingTest)
+               .triggerJobs()
+               .assertRunning(productionUsWest1);
+        assertEquals("Upgrade is not overdue yet", Duration.ZERO, metric.get());
+
+        // Upgrade continues into block window
+        tester.clock().advance(Duration.ofHours(3)); // Tuesday at 02:00 (03:00 CET)
+        assertEquals("Upgrade is overdue measured relative to window 2", Duration.ofHours(2), metric.get());
+
+        tester.clock().advance(Duration.ofHours(6)); // Tuesday at 08:00 (09:00 CET)
```
jonmv, line 541: I find it more natural to take the union of all block windows, rather than to split them. Is there a particular reason for this choice?

jonmv (approved):
Nice :)

PR #20522, opened 2021-12-15
Extend proton config for controlling how many buffers to compact at

toregge:
the same time.

@geirst : please review


geirst (approved):
👍

PR #20523, opened 2021-12-15
Deprecate vespa-http-client

bjorncs:


vespa-http-client/src/main/java/com/yahoo/vespa/http/client/SyncFeedClient.java
```diff
@@ -19,7 +19,9 @@
  * (but it can and should be reused for multiple subsequent synchronous calls).
  *
  * @author bratseth
+ * @deprecated Vespa-http-client will be removed in Vespa 8. It's replaced by <a href="https://docs.vespa.ai/en/vespa-feed-client.html">vespa-feed-client</a>
  */
+@Deprecated
 public class SyncFeedClient implements AutoCloseable {
```
bratseth, line 25: I think we need a replacement for this in vespa-feed-client.

bratseth (approved):

PR #20524, opened 2021-12-15
Avoid creating many sets and lists containing the same list of consum…

baldersheim:
…ers.

Use a dictionary to keep track of the unique ones.

@bjorncs or @gjoranv PR


bjorncs (approved):

gjoranv (commented):
👍

PR #20529, opened 2021-12-15
Let bucket maintenance priority queue be FIFO ordered within priority class [run-systemtest]

vekterli:
@geirst please review

This avoids a potential starvation issue caused by the existing
implementation, which is bucket ID ordered within a given priority
class. The latter has the unfortunate effect that frequently reinserting
buckets that sort before buckets that are already in the queue may
starve these from being popped from the queue.

Move to a composite key that first sorts on priority, then on a strictly
increasing sequence number. Add a secondary index into this structure
that allows for lookups on bucket IDs as before.


geirst (approved):
👍

PR #20537, opened 2021-12-16
Arnej/improve mirror responsiveness

arnej27959:

@havardpe please review
@jonmv FYI

### DIFF
--- a/jrt/src/com/yahoo/jrt/TransportThread.java
+++ b/jrt/src/com/yahoo/jrt/TransportThread.java
@@ -20,13 +20,6 @@ public class TransportThread {
     private static final int CLOSING = 2;
     private static final int CLOSED  = 3;
 
-    private static final int SELECT_TIMEOUT_MS = getSelectTimeoutMs();
-
-    private static int getSelectTimeoutMs() {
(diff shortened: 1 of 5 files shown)


jrt/src/com/yahoo/jrt/TransportThread.java
```diff
@@ -294,7 +301,7 @@ private void run() {
 
             // perform I/O selection
             try {
-                selector.select(100);
+                selector.select(SELECT_TIMEOUT_MS);
```
havardpe, line 304: I think this should only be changed in order to improve the accuracy of timeouts. Then we would also need to change the scheduler tick size. C++ uses 10ms epoll timeout and scheduler tick size. We could do that in Java as well, but in a separate pull request. Note that using ScheduleNow when we get a reply will result in the task being run right afterwards (independent of select timeout). As far as I can see, the changes to Mirror.java will make the select timeout only relevant for the back-off case where we use an actual timeout value.

havardpe (commented):
The changes to Mirror.java looks good, but the select timeout should not be relevant here.

havardpe (approved):

PR #20549, opened 2021-12-16
Cover additional update failure edge cases with metrics

vekterli:
@geirst please review.

This change adds previously missing metric increments for test-and-set
condition failures when this happens in the context of a write-repair,
as well as adding missing wiring for the "notfound" failure metric.
The latter is incremented when an update is returned from the content
nodes having found no existing document, or when the read-phase of
a write-repair finds no document to update (and neither `create: true`
nor a TaS condition is set on the update).

Also remove some internal reply type erasure to make it more obvious
what the reply type must be.


geirst (approved):
👍

PR #20551, opened 2021-12-16
Don't split interlinear annotations.

toregge:
@geirst : please review

### DIFF
--- a/juniper/src/vespa/juniper/sumdesc.cpp
+++ b/juniper/src/vespa/juniper/sumdesc.cpp
@@ -33,28 +33,26 @@ bool wordchar(const unsigned char* s)
     }
 }
 
-bool wordchar_or_il_ann_anchor(const unsigned char* s)
+bool wordchar_or_il_ann_char(const unsigned char* s, ucs4_t annotation_char)
 {
     unsigned char c = *s;
     if (c & 0x80) {
         ucs4_t u = Fast_UnicodeUtil::GetUTF8Char(s);
         return Fast_UnicodeUtil::IsWordChar(u) ||
-            u == il_ann_anchor;
+            u == annotation_char;
     } else {
         return isalnum(c);
     }
 }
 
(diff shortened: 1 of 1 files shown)


juniper/src/vespa/juniper/sumdesc.cpp
```diff
@@ -29,6 +33,30 @@ bool wordchar(const unsigned char* s)
     }
 }
 
+bool wordchar_or_il_ann_anchor(const unsigned char* s)
+{
+    unsigned char c = *s;
```
geirst, line 38: Please move the content of this function into a helper function that both wordchar_or_il_ann_anchor() and wordchar_or_il_ann_terminator() can call.

juniper/src/vespa/juniper/sumdesc.cpp
```diff
@@ -40,6 +68,41 @@ bool nonwordchar(const unsigned char* s)
     }
 }
 
+bool
+il_ann_anchor_char(const unsigned char* s)
+{
+    unsigned char c = *s;
```
geirst, line 74: Please move the content of this function into a helper function that both il_ann_anchor_char(), il_ann_separator_char() and il_ann_terminator_char() can call.

juniper/src/vespa/juniper/sumdesc.cpp
```diff
@@ -66,10 +129,20 @@ int complete_word(unsigned char* start, ssize_t length,
     // Figure out if a word needs completion or if we are just going
     // to eliminate whitespace
     if (!wordchar(ptr)) {
-        whitespace_elim = true;
-        // Change direction of scan
-        increment = -increment;
-        chartest = wordchar;
+        if (increment > 0 && il_ann_anchor_char(ptr)) {
```
geirst, line 132: It would be nice with an extension of the description at L129-130 to match the extra checks added.

geirst (approved):
👍

geirst (approved):
👍

PR #20554, opened 2021-12-16
Add a BERT embedder

bratseth:


jobergum (commented):
I think that we should use WordPieceEmbedder, we have SentencePieceEmbedder which also can be used for BERT models (e.g mBERT).

bratseth:
Or BertWordPieceEmbedder since I think the "##" prefix is a BERT thing, long name though.
I guess I can just make that configurable and then it will be general enough to take the WordPiece name.

jobergum:
I'm fine with BertWordPieceEmbedder

bratseth:
PTAL @jobergum

jobergum (approved):

jobergum:
Once this is in we need to have a sprint on fixing sample apps to use this instead :)

bratseth:
Yes.

PR #20555, opened 2021-12-16
Faster reindexing when triggered

jonmv:
@baldersheim please review.


jonmv:
Oops, IntelliJ committed sosme other deletions here ...

PR #20563, opened 2021-12-17
numbers outside of the range of a long are forced to have a decimal p…

baldersheim:
…unctuation to enforce parsing as non integer number.

@gjoranv or @bratseth PR

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/json/JacksonUtil.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/json/JacksonUtil.java
@@ -11,6 +11,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
 import java.io.IOException;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
+import java.text.NumberFormat;
 import java.util.Locale;
 
 /**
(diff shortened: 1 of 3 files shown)


bratseth (approved):
We could just do it always ...

baldersheim:
I think it has been done like this to kind of preserve integer/floating point from reading of json to writing of json.
Main reason to use slime instead....

metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/json/JacksonUtil.java
```diff
@@ -19,12 +20,18 @@
  */
 public class JacksonUtil {
 
-    private static final DecimalFormat decimalFormat = createDecimalFormat();
-    private static DecimalFormat createDecimalFormat() {
+    private static final ThreadLocal<DecimalFormat> withinLongRangeFormat = ThreadLocal.withInitial(JacksonUtil::createWithinLongRangeFormat);
+    private static final ThreadLocal<DecimalFormat> outsideLongRangeFormat = ThreadLocal.withInitial(JacksonUtil::createOutsideLongRangeFormat);
```
gjoranv, line 24: Just curious, why do these have to be ThreadLocal?

bratseth:
We should always write floating points as floating point numbers and integers as integers? We always know the source type ...

metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/json/JacksonUtil.java
baldersheim: It was me wanted to be on the safe side due to the below comment in DecimalFormat.java.
From DecimalFormat.java
Synchronization
Decimal formats are generally not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

baldersheim:
In theory we could, but only if all json sources adhere to the same pattern.
1 - Always print doubles with decimal punctuation.
2 - Always print integer numbers within the range of long without decimals.
3 - Always print integer numbers outside of long range with decimal punctuation.

4 - And then never drop this information......

And we do not consistently comply with either of these.

baldersheim:
#20573 brings us one step closer on point 4

PR #20567, opened 2021-12-17
Arnej/speedup routing test case

arnej27959:

@jonmv please review
@gjoranv FYI


jonmv (approved):
Nice!

gjoranv:
Great. In IDEA, the test now runs in less than 1 s. However, when I use maven from the shell, the test still takes ~30 s. Could you please check if you see the same behaviour, @arnej27959 and @jonmv?

PR #20570, opened 2021-12-17
Complete wiring of OK/failure metric reporting during update write-repair [run-systemtest]

vekterli:
@geirst please review

This also changes the aggregation semantics of "not found" results.
An update reply indicating "not found" is protocol-wise reported back to
the client as successful but with a not-found field set that has to be
explicitly checked. But on the distributor the `notfound` metric is
under the `updates.failures` metric set, i.e. not counted as a success.

To avoid double bookkeeping caused by treating it as both OK and not,
we choose to not count "not found" updates under the `ok` metric. But to
avoid artificially inflating the `failures.sum` metric, we remove the
`notfound` metric from the implicit sum aggregation.

This means that visibility into "not found" updates requires explicitly
tracking the `notfound` metric as opposed to just the aggregate, but
this should be an acceptable tradeoff to avoid false failure positives.


geirst (approved):
👍

PR #20571, opened 2021-12-17
Even if we do not filter by consumer on the services, we can do it ea…

baldersheim:
…rly in the metrics-proxy.

This is especially benefiscal for the most frequent requests like autoscaling which only looks at a small set (1%) of the metrics.
This both reduces memory footprint and saves cpu.

@gjoranv or @bjorncs PR

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/core/MetricsConsumers.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/core/MetricsConsumers.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collector;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


bjorncs (approved):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/core/MetricsConsumers.java
```diff
@@ -31,6 +32,7 @@
 
     // All metrics for each consumer.
     private final Map<ConsumerId, List<ConfiguredMetric>> consumerMetrics;
+    private final Map<ConsumerId, Map<MetricId, ConfiguredMetric>> configuredMetricByMetricByConsumer;
```
gjoranv, line 35: Perhaps it's easier to just name it 'configuredMetricsByConsumer'.

gjoranv (commented):
👍

PR #20577, opened 2021-12-18
Fix calculation of initial value of last logged timestamp

hmusum:


config/src/main/java/com/yahoo/config/subscription/impl/JRTConfigRequester.java
```diff
@@ -67,8 +67,8 @@
         this.scheduler = scheduler;
         this.connectionPool = connectionPool;
         this.timingValues = timingValues;
-        // Adjust so that we wait 1 second with logging warning in case there are some errors just when starting up
-        timeForLastLogWarning = Instant.now().minus(delayBetweenWarnings.plus(Duration.ofSeconds(5)));
+        // Adjust so that we wait 5 seconds with logging warning in case there are some errors just when starting up
+        timeForLastLogWarning = Instant.now().minus(delayBetweenWarnings).plus(Duration.ofSeconds(5));
```
baldersheim, line 71: The devil is in the details :)

baldersheim (approved):

PR #20579, opened 2021-12-18
Balder/expose memory percentage on the container cluster

baldersheim:
@geirst or @bratseth PR


bratseth (approved):
👍

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
```diff
@@ -241,11 +241,14 @@ private void setupTuning(ContentCluster c, ModelElement tuning) {
         }
 
         /** Returns whether this hosts one of the given container clusters */
-        private boolean isCombined(String clusterId, Collection<ContainerModel> containers) {
-            return containers.stream()
-                             .map(model -> model.getCluster().getHostClusterId())
-                             .filter(Optional::isPresent)
-                             .anyMatch(id -> id.get().equals(clusterId));
+        private double fractionOfMemoryReserved(String clusterId, Collection<ContainerModel> containers) {
```
geirst, line 244: Please update function comment.

PR #20587, opened 2021-12-20
Log or fail deployment with invalid jvm gc options

hmusum:


config-model/src/test/java/com/yahoo/vespa/model/container/xml/JvmOptionsTest.java
```diff
@@ -166,31 +171,28 @@ public void requireThatInvalidJvmGcOptionsAreLogged() throws IOException, SAXExc
...
-        verifyLoggingOfJvmGcOptions(true,
-                                    "-XX:+ParallelGCThreads=8");
+        verifyLoggingOfJvmGcOptions(true, "-XX:+ParallelGCThreads=8");
+        verifyLoggingOfJvmGcOptions(true, "-XX:MaxTenuringThreshold"); // No + or - after colon
         verifyLoggingOfJvmGcOptions(false, "-XX:+UseConcMarkSweepGC");
     }
 
+    @Test
+    public void requireThatInvalidJvmGcOptionsFailDeployment() throws IOException, SAXException {
+        expectedException.expect(IllegalArgumentException.class);
```
baldersheim, line 181: I think using explicit try/catch is a lot easier to read.
try {
   some_code()
   fail();
   assertXX
}

baldersheim (approved):

config-model/src/test/java/com/yahoo/vespa/model/container/xml/JvmOptionsTest.java
hmusum: Hm, yeah, thanks for the input. I'll think about it the next time I'll need to test that an exception is thrown

PR #20589, opened 2021-12-20
Jonmv/reapply replace optimaize with open nlp

jonmv:
@bratseth please review. This runs ~3x faster than the previous attempt, "on my machine". I can't really say how accurate the results are, but I don't think they're worse than what we currently have. Is it a bit sketchy to change this on a minor, though?


linguistics/src/main/java/com/yahoo/language/opennlp/DefaultLanguageDetectorContextGenerator.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public String[] getContext(CharSequence document) {
+        int[] normalized = normalizer.normalize(document).codePoints().map(Character::toLowerCase).toArray();
+        Set<String> grams = new HashSet<>();
+        for (int i = 0; i < normalized.length; i++)
+            for (int j = minLength; j <= maxLength && i + j < normalized.length; j++)
+                grams.add(new String(normalized, i, j));
+
+        return grams.toArray(new String[0]);
```
jonmv, line 29:
```suggestion
        return grams.toArray(new String[grams.size()]);
```
Revert IntelliJ's silly suggestion of using empty array for `.toArray(...)`.

bratseth (approved):
👍 👍 

I don't see a problem with changing this on a minor.

PR #20590, opened 2021-12-20
Optimize handling of cleardoc updates to attributes with enum store

geirst:
@baldersheim please review

When retiring a content node a huge amount of cleardoc updates towards attributes are issued as part of deleteBucket operations. For some applications this has been observed as a bottleneck leading to increased latencies on client operations on the retired node. This optimization should fix that problem.

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/changevector.h
+++ b/searchlib/src/vespa/searchlib/attribute/changevector.h
@@ -25,32 +25,32 @@ struct ChangeBase {
         DIV,
         CLEARDOC
     };
-    enum {UNSET_ENUM = 0xffffffffu};
+    enum {UNSET_ENTRY_REF = 0xffffffffu};
 
     ChangeBase() :
         _type(NOOP),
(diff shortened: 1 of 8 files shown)


baldersheim (approved):
👍

searchlib/src/vespa/searchlib/attribute/changevector.h
```diff
@@ -25,32 +25,32 @@ struct ChangeBase {
         DIV,
         CLEARDOC
     };
-    enum {UNSET_ENUM = 0xffffffffu};
+    enum {UNSET_ENTRY_REF = 0xffffffffu};
```
toregge, line 28: Note that this value is within the range of valid entry refs thus we might (very rarely) consider a set entry ref to not being set. Consider using 0 instead.

geirst: Fixed in https://github.com/vespa-engine/vespa/pull/20633.

PR #20594, opened 2021-12-20
Use slf4j SimpleLogger instead of log4j

gjoranv:
This introduces a new dependency on slf4j-simple, but the zk-command-line-client module has no dependents, so it does not spread into other parts of Vespa. Besides, the slf4j-simple artifact seems innocent with just a handful of classes and a total of 15 KB.

I could not find a way to filter out the excessive INFO logging from zk with slf4j-jdk14 (its java.util.logging bridge)

FYI: @bratseth


gjoranv:
FYI: @aressem This will remove the last critical alert from sonatype.

hmusum (approved):

PR #20596, opened 2021-12-20
Don't trigger implicit ContentPolicy random send fallback on expected transient errors [run-systemtest]

vekterli:
@jonmv please review. I started looking at the existing "simulator" tests, then wisely retreated and left them alone in favor of some refactoring to be able to do explicit testing 😬
@yngveaasheim FYI

The `ContentPolicy` has a failure handling policy where more than _n_ error
replies (a small number in practice) will trigger an implicit random send instead
of using the cached cluster state. This is to force rediscovery of the actual
cluster state, and is useful if a node is bad but we're not sending feed to enough
other nodes to figure it out from them.

However, certain error codes may be used frequently by the content layer for
purposes that do _not_ indicate that a change in cluster state may have happened,
and should therefore not be counted as errors that may indicate a bad node:
 * `ERROR_TEST_AND_SET_CONDITION_FAILED`: may happen for any mutating operation
   that has an associated TaS condition. Technically an `APP_FATAL_ERROR` since
   resending doesn't make sense.
 * `ERROR_BUSY`: may happen for concurrent mutations and if distributors are
   in the process of changing bucket ownership and the grace period hasn't
   passed yet. Also sent if queues are full and client policy should back off
   a bit. None of these are errors as per se.

### DIFF
--- a/documentapi/abi-spec.json
+++ b/documentapi/abi-spec.json
@@ -1565,6 +1565,20 @@
       "protected final java.util.Random randomizer"
     ]
   },
+  "com.yahoo.documentapi.messagebus.protocol.ContentPolicy$InstabilityChecker": {
+    "superClass": "java.lang.Object",
+    "interfaces": [],
+    "attributes": [
+      "public",
+      "interface",
+      "abstract"
+    ],
+    "methods": [
+      "public abstract boolean tooManyFailures(int)",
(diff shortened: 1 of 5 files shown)


jonmv (approved):
TaS == bad node?! (((＼（✘෴✘）／)))

This happening in the first place also smells like layer violations :) Good fix.

PR #20598, opened 2021-12-20
Allow unquoted annotations

bratseth:
Apparently this was always the plan, but since ID consumed ":" it wouldn't work 😭


jonmv (approved):
( ´༎ຶㅂ༎ຶ`)

PR #20604, opened 2021-12-21
Validate JVM options [run-systemtest]

hmusum:


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -1121,9 +1122,21 @@ String buildLegacyOptions() {
...
+            if (jvmOptions == null || jvmOptions.isEmpty()) return;
+
+            String[] optionList = jvmOptions.split(" ");
+            List<String> invalidOptions = Arrays.stream(optionList)
+                                                .filter(option -> !option.isEmpty())
+                                                .filter(option -> !Pattern.matches(validPattern.pattern(), option))
+                                                .sorted()
+                                                .collect(Collectors.toList());
+
+            String message = "Invalid JVM options in services.xml: " + String.join(",", invalidOptions);
```
baldersheim, line 1135: I guess you should return here if invalidOptions is empty. If not you will throw or log an empty message.

baldersheim (changes requested):
Consider my comment on early return.

config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
hmusum: Right, thanks, will fix

baldersheim (approved):

PR #20605, opened 2021-12-21
Output YQL annotations without quotes

bratseth:


container-search/src/main/antlr4/com/yahoo/search/yql/yqlplus.g4
```diff
@@ -459,10 +459,14 @@ propertyNameAndValue
 	;
 
 propertyName
-	: IDENTIFIER
```
jonmv, line 462: This no longer matches `propertyref`, but the similar names don't _necessarily_ mean they are similar. Is this OK?

jonmv (approved):

container-search/src/main/antlr4/com/yahoo/search/yql/yqlplus.g4
jonmv: Yes, forget I asked!

PR #20606, opened 2021-12-21
Bratseth/allow keywords as yql properties

bratseth:


bratseth:
ping @jonmv

jonmv (approved):
pong


I'll admit I don't understand why quoting was necessary before, as an `IDENTIFIER` should match `offset` ... ?

bratseth:
Offset is a reserved word so it is consumed by that.

PR #20608, opened 2021-12-21
Use mechanism to dyanmically disable javadoc for non public modules.

aressem:

### DIFF
--- a/screwdriver/release-java-artifacts.sh
+++ b/screwdriver/release-java-artifacts.sh
@@ -40,6 +40,14 @@ find . -name "pom.xml" -exec sed -i'' -e "s,<version>.*SNAPSHOT.*</version>,<ver
      -e "s,<test-framework.version>.*project.version.*</test-framework.version>,<test-framework.version>$VESPA_RELEASE</test-framework.version>," \
      {} \;
 
+# We disable javadoc for all modules not marked as public API
+for MODULE in $(comm -2 -3 \
+                <(find . -name "*.java" | awk -F/ '{print $2}' | sort -u)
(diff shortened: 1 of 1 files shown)


tokle (approved):

aressem:
Not tested in PR build

screwdriver/release-java-artifacts.sh
```diff
@@ -40,6 +40,14 @@ find . -name "pom.xml" -exec sed -i'' -e "s,<version>.*SNAPSHOT.*</version>,<ver
      -e "s,<test-framework.version>.*project.version.*</test-framework.version>,<test-framework.version>$VESPA_RELEASE</test-framework.version>," \
      {} \;
 
+# We disable javadoc for all modules not marked as public API
+for MODULE in $(comm -2 -3 \
+                <(find . -name "*.java" | awk -F/ '{print $2}' | sort -u)
+                <(find . -name "package-info.java" -exec grep -HnE "@(com.yahoo.api.annotations.)?PublicApi.*" {} \; | awk -F/ '{print $2}' | sort -u)); do
+    mkdir -p $MODULE/src/main/javadoc
+    echo "No javadoc available for module" > $MODULE/src/main/javadoc/README
```
bratseth, line 48: Perhaps this should say something like
"Internal module - do not use. No javadoc available."

aressem: This will not display on `javadoc.io`, it is there only to be able to create the javadoc jar file. If we would like to display something on `javadoc.io` we would have to create an index.html file instead that `javadoc.io` would pick up.

PR #20626, opened 2022-01-03
Deprecate getConfigMd5() and setConfigMd5() [run-systemtest]

hmusum:
We don't use md5 checksum for config anymore, methods will be removed
in Vespa 8.


baldersheim (approved):

hmusum:
Cannot get PR builds to work, merging.

gjoranv:
👍

PR #20629, opened 2022-01-03
Support aborting and rolling back incomplete upgrade

mpolden:
This adds limited support for aborting an ongoing upgrade and rolling back the
system.

For example:
1. System is on version v0
2. System has started rolling out v1, where some zones have completed upgrading
3. We discover that v1 is severely broken and we need to halt the upgrade, and
   downgrade any infrastructure that has completed upgrading
   
Merge with internal PR.

@jonmv


mpolden:
Can you review this, @freva? I think Jon M. is OOO.

freva (approved):

PR #20638, opened 2022-01-03
Invalidate bucket DB replica statistics upon recovery mode entry [run-systemtest]

vekterli:
@geirst please review

The replica stats track the minimum replication factor for any bucket
for a given content node the distributor maintains buckets for. These
statistics may be asynchronously queried by the cluster controller
through the host info reporting API.

If we do not invalidate the statistics upon a cluster state change, there
is a very small window of time where the distributor may potentially
report back _stale_ statistics that were valid for the _prior_ cluster
state version but not for the new one. This can happen if the cluster
controller fetches host info from the node in between start of the recovery
period and the completion of the recovery mode DB scan. Receiving stale
replication statistics may cause the cluster controller to erroneously
believe that replication due to node retirements etc has completed earlier
than it really has, possibly impacting orchestration decisions in a sub-
optimal manner.

### DIFF
--- a/storage/src/tests/distributor/top_level_distributor_test.cpp
+++ b/storage/src/tests/distributor/top_level_distributor_test.cpp
@@ -90,6 +90,10 @@ struct TopLevelDistributorTest : Test, TopLevelDistributorTestUtil {
         return _distributor->getBucketSpacesStats();
     }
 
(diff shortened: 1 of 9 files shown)


geirst (approved):
👍

PR #20641, opened 2022-01-03
Add InfrastructureApplication in application-model [run-systemtest]

hakonhall:


gjoranv (approved):
Great! Some of these constants are also used in the internal repo, so that should be updated as well.

hakonhall:
Yes, I have some changes there too.

PR #20653, opened 2022-01-05
Let max summary filesize be 2% of available memory instead of an anci…

baldersheim:
…ent staircase, that only fits the perfect the ones with a perfect BMI anyway.

@geirst PR


geirst (approved):
👍

PR #20645, opened 2022-01-04
Use InfrastructureApplication

hakonhall:

PR #20656, opened 2022-01-05
Use InfrastructureApplication, try 2

hakonhall:
The original PR #20645 caused node-admin to fail to start because the application-model bundle was not found.  This PR extends the original by adding application-model to components.

### DIFF
--- a/node-admin/CMakeLists.txt
+++ b/node-admin/CMakeLists.txt
@@ -1,7 +1,7 @@
 # Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 install(DIRECTORY DESTINATION logs/vespa/node-admin)
 install(FILES target/node-admin-jar-with-dependencies.jar DESTINATION conf/node-admin-app/components)
-install_symlink(lib/jars/application-model-jar-with-dependencies.jar conf/node-admin-app/components/application-model.jar)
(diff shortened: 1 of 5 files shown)


node-admin/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 # Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 install(DIRECTORY DESTINATION logs/vespa/node-admin)
 install(FILES target/node-admin-jar-with-dependencies.jar DESTINATION conf/node-admin-app/components)
+install_symlink(lib/jars/application-model-jar-with-dependencies.jar conf/node-admin-app/components/application-model.jar)
```
gjoranv, line 4: Why not preserve the '-jar-with-dependencies' suffix?

hakonhall: For readability.  The "-jar-with-dependencies" part is ugly and long.  They are typically removed for config server, but preserved here for node-admin. I could just as easily have included the part for consistency - let me know if you prefer that.

gjoranv: Agree that it's ugly, we should have used just "-fat' or "-deploy" instead. But for now, let's be consistent. It's also convenient to see from the file names which jars are fat and not.

hakonhall: Done

gjoranv:
👍

gjoranv (approved):

PR #20663, opened 2022-01-05
Don't delete new sessions in file system that does not exist in zookeeper [run-systemtest]

hmusum:
Issue seen occasionally in system tests

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -572,8 +572,16 @@ public class SessionRepository {
     public void deleteExpiredSessions(Map<ApplicationId, Long> activeSessions) {
         log.log(Level.FINE, () -> "Purging old sessions for tenant '" + tenantName + "'");
         Set<LocalSession> toDelete = new HashSet<>();
+        Set<Long> newSessions = findNewSessionsInFileSystem();
         try {
             for (LocalSession candidate : getLocalSessionsFromFileSystem()) {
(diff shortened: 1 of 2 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -572,8 +572,16 @@ private void childEvent(CuratorFramework ignored, PathChildrenCacheEvent event)
     public void deleteExpiredSessions(Map<ApplicationId, Long> activeSessions) {
         log.log(Level.FINE, () -> "Purging old sessions for tenant '" + tenantName + "'");
         Set<LocalSession> toDelete = new HashSet<>();
+        Set<Long> newSessions = findNewSessionsInFileSystem();
         try {
             for (LocalSession candidate : getLocalSessionsFromFileSystem()) {
+                // Skip sessions newly added (we might have a session in the file system, but not in ZooKeeper,
+                // we don't want to touch any of them)
+                if (newSessions.contains(candidate.getSessionId())) {
+                    log.log(Level.INFO, () -> "Skipping session " + candidate.getSessionId() + ", newly created: ");
```
hakonhall, line 581: Nit: Since INFO is always logged, there is no need to pass a String supplier.

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Originally had it at level FINE, so forgot when changing. I'll change it to FINE in a few days if this change works as expected

PR #20665, opened 2022-01-05
Support an external list of terms in term list operators

bratseth:


container-search/src/main/java/com/yahoo/search/yql/ParameterListParser.java
```diff
@@ -0,0 +1,204 @@
...
+        while (s.peek() != '}') {
+            String key;
+            if (s.passOptional('\'')) {
+                key = s.stringTo(s.position('\''));
+                s.pass('\'');
+            }
+            else if (s.passOptional('"')) {
+                key = s.stringTo(s.position('"'));
+                s.pass('"');
+            }
```
jonmv, line 51: Keys like `foo'bar"baz:bax` are not possible then, but that's OK.

jonmv (approved):
Great. Would we want to allow this for other similar term lists as well?

bratseth:
Like what? This covers WeakAnd, WAND and DotProduct

jonmv:
Right, they all use the weighted set, of course.

PR #20669, opened 2022-01-05
Remove one of the tests for generic services.

gjoranv:
Keeping GenericServicesTest for now.

FYI: @bjorncs


hmusum (changes requested):
Test sounds useful and seem to have been added recently, I suggest to just remove `service` elements from services.xml

gjoranv:
Thanks, I didn't notice that the point here was the deployment spec. The services.xml was probably just copied from the generic services test.

PR #20673, opened 2022-01-06
Never modify cached runs

mpolden:
The cached runs can be read by multiple threads, which will cause a
ConcurrentModificationException if the same runs are being updated by another
thread.

@freva or @jonmv


freva (approved):

jonmv:
Thanks!

PR #20684, opened 2022-01-06
Use UUID for Linux container nodes

freva:
Motivation: We want to reset certain resources/caches in host-admin when a node is reallocated. Sometimes a node can be deallocated and allocated to a new application within a single host-admin tick so that it never sees the node unallocated. By using a unique ID, host-admin can simply compare against the node ID to find out when a node has been reallocated.


mpolden (approved):
Did you check if any internal tests fail on this?

freva:
Yes, no failures.

PR #20691, opened 2022-01-07
Use newest test framework version

jonmv:
@tokle or @bjorncs please review. We should revert this when our hosted apps have upgraded to the next release.


bjorncs (approved):

bjorncs:
Is this temporary or permanent?

jonmv:
Just temporary.

jonmv:
Also not sure we really need it.

jonmv:
Let's just work around this instead, _if_ it turns out to be a problem.

PR #20696, opened 2022-01-07
Allow comma in JVM options

hmusum:


config-model/src/test/java/com/yahoo/vespa/model/container/xml/JvmOptionsTest.java
```diff
@@ -246,7 +246,7 @@ public void requireThatJvmOptionsAreLogged()  throws IOException, SAXException {
         // Valid options, should not log anything
         verifyLoggingOfJvmOptions(true, "options", "-Xms2G");
         verifyLoggingOfJvmOptions(true, "options", "-verbose:gc");
-        verifyLoggingOfJvmOptions(true, "options", "-Djava.library.path=/opt/vespa/lib64:/home/y/lib64");
+        verifyLoggingOfJvmOptions(true, "options", "-Djava.library.path=/opt/vespa/lib64:/home/y/lib64 -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005");
```
mpolden, line 249: Do we need to allow this? The debug port won't be reachable in any case.

mpolden (approved):

config-model/src/test/java/com/yahoo/vespa/model/container/xml/JvmOptionsTest.java
mpolden: > Maybe we should disallow -Xrunjdwp:transport completely in hosted?

Yes, it will prevent confusion about why debugging doesn't work at least.

> Right now we warn or fail deployment for hosted (depending on feature flag value) and warn for non-hosted for JVM options like these. 

:+1:

mpolden (approved):

PR #20699, opened 2022-01-07
Support date range in block window

mpolden:
@bratseth

### DIFF
--- a/config-model-api/src/test/java/com/yahoo/config/application/api/TimeWindowTest.java
+++ b/config-model-api/src/test/java/com/yahoo/config/application/api/TimeWindowTest.java
@@ -3,7 +3,11 @@ package com.yahoo.config.application.api;
 
 import org.junit.Test;
 
+import java.time.DayOfWeek;
 import java.time.Instant;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import static java.time.DayOfWeek.FRIDAY;
 import static java.time.DayOfWeek.MONDAY;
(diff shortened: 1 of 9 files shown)


config-model-api/src/test/java/com/yahoo/config/application/api/TimeWindowTest.java
```diff
@@ -25,71 +28,106 @@
...
+
+            TimeWindow tw2 = TimeWindow.from("sun", "1", "CET", "2022-01-01", "2022-01-02");
+            Instant i6 = Instant.parse("2022-01-01T00:00:00.00Z"); // Wrong day
+            Instant i7 = Instant.parse("2022-01-02T01:00:00.00Z"); // Wrong hour because of time zone offset
+            Instant i8 = Instant.parse("2022-01-02T00:00:00.00Z");
+            assertOutside(tw2, i6);
+            assertOutside(tw2, i7);
+            assertInside(tw2, i8);
+
+            TimeWindow tw3 = TimeWindow.from("", "", "CET", "2022-01-02", "");
```
freva, line 83: Its not possible to define any range shorter than 7 days without also specifying `days`

bratseth (approved):
Great!

freva (commented):
We expose the hours and days block window in application API which is used in console, f.ex.:
image

https://github.com/vespa-engine/vespa/blob/a3bc6f9b6af95b97a2b55147a586ca19d1d98938/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java#L1162
I think we should change this to be based on _this_ week (as opposed to what it is now, which is the days/hours blocked without date range).

config-model-api/src/test/java/com/yahoo/config/application/api/TimeWindowTest.java
mpolden: Thanks! Fixed in https://github.com/vespa-engine/vespa/pull/20699

PR #20702, opened 2022-01-07
low-level code to sample CPU usage per thread

havardpe:

I think this is the low-level wrapper we want, to be able to build something that tracks CPU usage separated into different categories. The main point being able to sample the CPU usage of one thread from another thread.

Currently this code will only run on Linux. Is that OK? Do we need a dummy implementation that always reports 0 CPU usage? Do we need alternative implementations for other OSes?

@toregge please review
@baldersheim @geirst FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/util/cpu_usage.cpp
+++ b/vespalib/src/vespa/vespalib/util/cpu_usage.cpp
@@ -1,0 +1,26 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "cpu_usage.h"
+#include "require.h"
+
+namespace vespalib {
+
+namespace cpu_usage {
+
+ThreadSampler::ThreadSampler()
+  : _my_clock()
+{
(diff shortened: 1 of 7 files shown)


toregge (approved):

vespalib/src/vespa/vespalib/util/cpu_usage.cpp
```diff
@@ -0,0 +1,26 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "cpu_usage.h"
+#include "require.h"
+
+namespace vespalib {
+
+namespace cpu_usage {
```
baldersheim, line 8: I prefer vespalib::cpu_usage { }. Less level to track

baldersheim (approved):

vespalib/src/vespa/vespalib/util/cpu_usage.cpp
havardpe: When I add code to the vespalib namespace there will be more close/re-open if we combine them, not sure if that is better...

PR #20704, opened 2022-01-07
unify java warnings

arnej27959:
* these were stricter than in parent, but to simplify
  we can just use compiler args from parent

@baldersheim please review


baldersheim (approved):
Wrong direction, but I guess you intend make this default in parent soon.

PR #20709, opened 2022-01-07
After we close a MBusServer we are not busy, retrying will not make l…

baldersheim:
…ife better.

Shutdown is a better error code.

@bjorncs or @jonmv PR


jonmv (approved):
As long as clients retry this, this is perfectly fine. I know at least the new vespa-feed-client retires 500, which this ends up being. In any case, 500 is better than hanging for ever.

PR #20710, opened 2022-01-07
Enforce no log4j

gjoranv:
FYI: @bjorncs, @oyving, @tokle, @bratseth
The second commit is the important one.

This will fail any application that  depends on `log4j:log4j` or `org.apache.logging.log4j:log4j-core` in any version and any scope.

Things to consider:
1. Do we want to fail for cloud apps? If not, the enforcer config in `hosted-tenant-base` should be moved to `hosted-vespa-tenant-base`
2. Do we want to fail for the fixed versions of log4j2? If not, I will add a version range.

I honestly don't know if it makes sense for applications to pull in log4j, as we want to route all logging to vespa.log via slf4j bridges anyway.

### DIFF
--- a/container-dependencies-enforcer/pom.xml
+++ b/container-dependencies-enforcer/pom.xml
@@ -35,111 +35,99 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    <profiles>
-        <profile>
-            <id>enforce-container-deps</id>
-            <activation>
-                <activeByDefault>false</activeByDefault>
-                 <property>
-                     <!-- Dependency resolution is broken for old maven used in our CentOS docker containers -->
-                    <name>maven.version</name>
-                    <value>!3.0.5</value>
-                </property>
-            </activation>
-            <build>
-                <plugins>
-                    <plugin>
-                        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


aressem (approved):
We could potentially allow versions of log4j 2.x that are not vulnerable, but I think that we could just disallow the use of it now as we do the logging as you wrote.

oyving:
Do we know of any customers that currently include `log4j` in their apps?

aressem:
@gjoranv is scanning the application packages for `log4j` dependencies.

gjoranv:
All "homerun" apps have been updated, but there are still a few positives:

**_blink-dpa.sol_**
```
Found Banned Dependency: org.apache.logging.log4j:log4j-core:jar:2.5
Found Banned Dependency: log4j:log4j:jar:1.2.15
```

_**abu-p13n--krystal [-staging]**_
```
Found Banned Dependency: log4j:log4j:jar:1.2.17
```

**_gemini-native--csp_** (not in prod, and my version was not necessarily the latest)
```
Found Banned Dependency: log4j:log4j:jar:1.2.17
Found Banned Dependency: org.apache.logging.log4j:log4j-core:jar:2.5
```

PR #20715, opened 2022-01-09
Use latest version protonbuf-java on major version 3. 3.16.1 -> 3.19.2

baldersheim:
@bratseth or @bjorncs or @gjoranv  PR
@vekterli @toregge PR

### DIFF
--- a/vespa-hadoop/pom.xml
+++ b/vespa-hadoop/pom.xml
@@ -73,7 +73,7 @@
         <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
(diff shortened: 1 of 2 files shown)


bratseth (approved):

vespa-hadoop/pom.xml
```diff
@@ -73,7 +73,7 @@
         <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
-            <version>3.16.1</version>
+            <version>3.19.2</version>
```
gjoranv, line 76: Explicit version should just be removed.

PR #20727, opened 2022-01-10
dummy fall-back for non-linux platforms

havardpe:

@toregge please review

### DIFF
--- a/vespalib/src/vespa/vespalib/util/cpu_usage.h
+++ b/vespalib/src/vespa/vespalib/util/cpu_usage.h
@@ -7,6 +7,19 @@ namespace vespalib {
 
 namespace cpu_usage {
 
+// do not use this directly (use ThreadSampler)
+class DummyThreadSampler {
+private:
+    steady_time _start;
+    double _load;
+public:
+    DummyThreadSampler();
+    void expected_load(double load);
+    duration sample() const;
+};
+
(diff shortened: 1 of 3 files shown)


baldersheim:
Test failure is unrelated

vespalib/src/vespa/vespalib/util/cpu_usage.h
```diff
@@ -7,6 +7,19 @@ namespace vespalib {
...
+private:
+    steady_time _start;
+    double _load;
+public:
+    DummyThreadSampler();
+    void expected_load(double load);
+    duration sample() const;
+};
+
+#ifdef __linux__
```
baldersheim, line 21: Is it possible to do this in the cpp file only to avoid ifdefs in the header files.

havardpe: yes, but it will have a cost

havardpe: what about an is_linux.h that defines a constexpr bool that we can use?

toregge (approved):

havardpe:
@baldersheim @toregge PTAL

toregge (approved):

baldersheim (approved):
Perfect

PR #20734, opened 2022-01-10
onnxModel to onnx in summary/matchfeatures

lesters:
Fixes field name of `onnx` rank feature in summary- or matchfeatures to actually be `onnx` rather than `onnxModel` as it is today.


gjoranv:
@lesters What's the status on this?

PR #20738, opened 2022-01-10
Use an intermediate representation for service convergence [run-systemtest]

hmusum:
WAnt an intermediate representation that can be used e.g. for
checking convergence before restarting.  This is the first step,
should have no changes for clients checking service/config convergence


bjorncs (approved):
Nice 👍

PR #20741, opened 2022-01-10
Stem by linguistics in rule bases

bratseth:
Also add a @language directive to stem in other languages than english.


arnej27959 (approved):
ok

PR #20743, opened 2022-01-11
Support configuring ZooKeeper snapshot method [run-systemtest]

hmusum:


standalone-container/src/main/java/com/yahoo/container/standalone/CloudConfigInstallVariables.java
```diff
@@ -121,6 +121,12 @@
         return getInstallVariable("zts_url");
     }
 
+    @Override
+    public String zooKeeperSnapshotMethod() {
+        String vespa_zookeeper_snapshot_method = System.getenv("VESPA_ZOOKEEPER_SNAPSHOT_METHOD");
```
mpolden, line 126: Nit: Variable should be camel-case.

mpolden (approved):

standalone-container/src/main/java/com/yahoo/container/standalone/CloudConfigInstallVariables.java
hmusum: Yup, thanks

PR #20749, opened 2022-01-11
GC unused large-rank-expression-limit

baldersheim:
@hmusum PR


hmusum (approved):
Consider adding a `removeAfter` for `largeRankExpressionLimit()` in `ModelContext` and removing method in `TestProperties`

baldersheim:
Unfortunately there are some tests that rely on this interface for controlling the size.
I will need to figure out what to do there before it can be removed. Added a comment.

hmusum:
👍

PR #20750, opened 2022-01-11
Jonmv/more ignore apps

jonmv:
@mpolden please review.


jonmv:
Uh, no ... this is not right >_< Please ignore.

PR #20753, opened 2022-01-11
Upgrade minor version of icu4j.

gjoranv:
- Upgrading major beyond 57 breaks container-search:SortingTestCase

57.2 is from Apr-2019, while latest is 70.1 (Oct-2021)

FYI: @arnej27959, @bratseth


baldersheim (approved):

baldersheim:
I guess vespa 8 is a good time to upgrade to latest and update tests.

PR #20756, opened 2022-01-11
Bjorncs/bundle import validation [run-systemtest]

bjorncs:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
@@ -1,6 +1,9 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.application.validation;
 
+import aQute.bnd.header.Parameters;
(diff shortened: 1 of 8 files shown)


config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
```diff
@@ -66,13 +74,71 @@ public void validateOSGIHeaders(DeployLogger deployLogger, JarFile jarFile) thro
...
+            deployLogger.logApplicationPackage(Level.WARNING,
+                    String.format("For JAR file '%s': \n" +
+                            "Manifest imports the following Java packages from '%s': %s. \n" +
+                            "%s",
+                            jarPath, artifact.name, packagesInUse, artifact.description));
+        });
+    }
+
+    private enum DeprecatedArtifact {
+        ORG_JSON("org.json:json",
```
gjoranv, line 120: I suggest to add a dummy entry with impossible package/artifact name, and use that in the test, to avoid semicolon hack (for empty enum) and a test that expires as soon as we are on the 8 branch.

bjorncs: I'll do that in upcoming PR

PR #20760, opened 2022-01-11
Add metrics that offer insight into how throttling affects scheduling

vekterli:
@baldersheim and @geirst please review

Adds the following stripe-level metrics:
 * `throttled_rpc_direct_dispatches` - number of times an RPC thread
   could not directly dispatch an async operation directly to Proton
   because it was disallowed by the throttle policy.
 * `throttled_persistence_thread_polls` - number of times a persistence
   thread could not immediately dispatch a queued async operation because
   it was disallowed by the throttle policy.
 * `timeouts_waiting_for_throttle_token` - number of times a persistence
   thread timed out waiting for an available throttle policy token.


baldersheim (approved):

geirst (approved):
👍

PR #20767, opened 2022-01-12
- Synchronize when removing nodes not present anymore. This simplifie…

baldersheim:
…s code at neglible cost that removes the race on size().

- Also use explicit AtomicReference to get an Immutable snapshot when doing a random sample.
- Done after observing IllegalArgumentException due to negative argument to Random.nextInt(bound).

@vekterli or @hmusum or @jonmv PR


hmusum (approved):

vekterli (commented):
:+1:

PR #20768, opened 2022-01-12
Jonmv/reapply zk changes

jonmv:
@hakonhall please review. @hmusum FYI.

### DIFF
--- a/zookeeper-server/zookeeper-server-3.6.3/src/test/java/com/yahoo/vespa/zookeper/VespaZooKeeperTest.java
+++ b/zookeeper-server/zookeeper-server-3.6.3/src/test/java/com/yahoo/vespa/zookeper/VespaZooKeeperTest.java
@@ -1,0 +1,224 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.zookeper;
+
+import com.yahoo.cloud.config.ZookeeperServerConfig;
+import com.yahoo.net.HostName;
+import com.yahoo.vespa.zookeeper.ReconfigurableVespaZooKeeperServer;
+import com.yahoo.vespa.zookeeper.Reconfigurer;
+import com.yahoo.vespa.zookeeper.VespaZooKeeperAdminImpl;
+import com.yahoo.vespa.zookeeper.client.ZkClientConfigBuilder;
+import org.apache.zookeeper.CreateMode;
+import org.apache.zookeeper.KeeperException;
+import org.apache.zookeeper.ZooDefs;
+import org.apache.zookeeper.admin.ZooKeeperAdmin;
+import org.apache.zookeeper.data.ACL;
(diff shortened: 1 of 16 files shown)


hmusum (approved):

hakonhall (approved):
LGTM but Screwdriver failed

jonmv:
Probably because we try 400 fixed ports, and other unit tests in other modules could grab any of these ...

hmusum:
Ports in unit tests are reserved per module, defined in internal repo

zookeeper-server/zookeeper-server-3.6.3/src/test/java/com/yahoo/vespa/zookeper/VespaZooKeeperTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    static List<Integer> getPorts(int id) {
+        if (ports.size() < id * 3) {
+            int previousPort;
+            if (ports.isEmpty()) {
+                String[] version = System.getProperty("zk-version").split("\\.");
+                int versionPortOffset = 0;
+                for (String part : version)
+                    versionPortOffset = 32 * (versionPortOffset + Integer.parseInt(part));
+                previousPort = 20000 + versionPortOffset % 30000;
```
hakonhall, line 218: This 20000 and 30000 has  something to do with the module port range Harald mentioned?  Extract to constants?

hakonhall (approved):
This doesn't remove the port conflict issue, but it does reduce the chance of it happening right?

zookeeper-server/zookeeper-server-3.6.3/src/test/java/com/yahoo/vespa/zookeper/VespaZooKeeperTest.java
jonmv: Hmm, they're just picked from thin air, really. Well, those other modules prefer things below 20k, but staying below 50k is just by chance.

jonmv:
Yes, the port conflict may still happen, but there's not really any way to avoid that, when you have to know the port number up front, while not having it bound, i.e., in the interval between choosing it and ZK (re-)binding it. 
The ports are lazily populated to minimise the chances of conflict.

jonmv:
With a Java API for creating ZK servers, we could perhaps pass the sockets? That would eliminate the problem.

hakonhall:
what if we keep an int counter for each zk version, incremented for each new zk version.  This counter could serve as a bit to chose between 2 ranges.  As long as the 2 ranges are within this module's range port allocation should always succeed?

jonmv:
Modules don't really have a static port range. That is outdated. Many tests get random ports from `new ServerSocket(0)`, I think?

jonmv:
And, the ZK-version range stuff is still kept, to also minimise conflict chance.

hakonhall:
> Many tests get random ports from new ServerSocket(0), I think?

I find 4 test files that use it.

PR #20790, opened 2022-01-13
use less CPU in retention-enforcer

arnej27959:
* start it with "nice"
* put in "sleep 1" between each processing step, to make it
  less CPU intensive
* avoid doing work which can never be useful, by not looking
  at databases where the earliest possible timestamp is
  not older than the retention limit

@bratseth please review


bratseth (approved):
Great!

PR #20800, opened 2022-01-13
Add an interface that can post a list of task instead of only one at …

baldersheim:
…a time.

Intention is to make it cheaper to post many small tasks.
It requires that the implementation adds support if it find it worthwhile.

@geirst or @havardpe PR


geirst (approved):
👍

PR #20806, opened 2022-01-13
TemplateFileSync: keep file in sync with template

hakonhall:


baldersheim:
@hakonhall  Is this still ongoing, or should it be closed ?

hakonhall:
Should be kept for a while longer

PR #20814, opened 2022-01-14
Ensure new components config is always acted upon [run-systemtest]

jonmv:
@bratseth please review.


jonmv:
The change is just to inline the `getOptionalConfigs` and keep the knowledge that there was a new components config that we should possibly return, unless bootstrap configs have changed, in which case we'll be called again with a new components subscriber, with generation -1.

jonmv:
@gjoranv FYI, or review if you wish.

jonmv:
Superseded by #20821

PR #20815, opened 2022-01-14
track cpu usage across threads and usage categories

havardpe:

@baldersheim @arnej27959 please review

This is a first version. We should consider adding more elaborate testing to verify conflict resolution.


arnej27959 (approved):
👍

havardpe:
@arnej27959 @baldersheim PTAL

Removing the exponential decay on cpu load values make them less stable, but avoids hiding spikes. When correlating with rusage we see that we spend some cpu in threads before/after we are able to run code. CPU usage reported from getrusage and from thread-based pthread clocks seem to have good correlation.

baldersheim (approved):

PR #20817, opened 2022-01-14
Readuce minimum direct memory size from 75m to 16m

baldersheim:


baldersheim:
@jonmv @bjorncs @hmusum PR
Netty is very hungry for buffers and is not shy.
So for small nodes it is hard to constrain memory usage when most of it is due to the fixed component.
The default stems from 2004/2005 when heap size default was 1536m....

jonmv (approved):

PR #20822, opened 2022-01-14
Remove Templar

hakonhall:


freva (approved):
Needs to be removed from host-admin's `ApplicationPackage`

hakonhall:
I had made the changes, but didn't realize it was in another repo

PR #20826, opened 2022-01-14
Deprecate yolean chain

gjoranv:
I would have liked to use `forRemoval=true` but then we would have to switch off `-Werror` for yolean.


yolean/src/main/java/com/yahoo/yolean/chain/Chain.java
```diff
@@ -16,7 +16,7 @@
  * @author Tony Vaagenes
  * @deprecated Will be removed in Vespa 8 with no replacement.
  */
-@Deprecated(forRemoval = true, since = "7")
+@Deprecated
```
bjorncs, line 19: Use `@SuppressWarnings("removal")` to suppress when `forRemoval=true`

bjorncs (approved):

PR #20827, opened 2022-01-15
Stop extending logging request handler [run-systemtest]

gjoranv:
Base class is unchanged for SearchHandler and AbstractProcessingHandler. There should be no changes to public APIs here.

FYI: @bratseth


gjoranv:
Replaced by #20846

PR #20832, opened 2022-01-17
Hmusum/wait for config generation before restarting [run-systemtest]

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -218,6 +218,7 @@ private ApplicationRepository(TenantRepository tenantRepository,
         private Metric metric = new NullMetric();
         private SecretStoreValidator secretStoreValidator = new SecretStoreValidator(new SecretStoreProvider().get());
         private FlagSource flagSource = new InMemoryFlagSource();
+        private ConfigConvergenceChecker configConvergenceChecker = new ConfigConvergenceChecker();
```
jonmv, line 221: This component is injectable, probably because it holds an executor. Perhaps use this?

On a side note, that executor isn't closed (no `deconstruct()` method), which should also be fixed.

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
```diff
@@ -149,6 +156,9 @@ private void restartServices(ApplicationId applicationId) {
         RestartActions restartActions = configChangeActions.getRestartActions().useForInternalRestart(internalRedeploy);
 
         if ( ! restartActions.isEmpty()) {
+
+            waitForConfigToConverge(applicationId);
```
jonmv, line 160: Who does the waiting here? I guess it could take a long time.

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
```diff
@@ -165,6 +175,31 @@ private void restartServices(ApplicationId applicationId) {
...
+                log.info("services converged on new generation " + response.currentGeneration);
+                return;
+            } else {
+                log.info("services not converged on new generation, wanted generation: " + response.wantedGeneration +
+                                 ", current generation: " + response.currentGeneration + ", will retry");
+                try { Thread.sleep(10_000); } catch (InterruptedException e) { /* ignore */ }
+            }
+        } while (clock.instant().isBefore(end));
+
+        throw new RuntimeException("Config has not converged");
```
jonmv, line 200: I suppose this'll ultimately lead to a failed deployment, not retried by the controller during deployment jobs. Should we retry? If so, we need some way to communicate this.

jonmv (commented):
Some comments, otherwise LGTM!

configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
hmusum: Will look into it

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
hmusum: Yes, will do the same as in `waitForResources()`

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
hmusum: Yes, will look into it

hmusum:

jonmv (approved):
LGTM. I believe a mapping from this error to something that is retried is needed in the controller code if this should allow deployments to be retried there, but that can be added later.

PR #20840, opened 2022-01-17
Invalidate node object cache based on znode version

mpolden:
This is the same cache mechanism as the controller uses for applications.

Merge with internal PR.

@jonmv


jonmv (approved):

jonmv:
Nice :)

PR #20846, opened 2022-01-18
Deprecate logging request handler [run-systemtest]

gjoranv:
Base class is unchanged for SearchHandler and AbstractProcessingHandler. There should be no changes to public APIs here.

FYI: @bratseth


bjorncs (approved):

bjorncs (approved):

gjoranv:
Last force push was just to use the '$' for sublass in a string.

PR #20849, opened 2022-01-18
Use gz as zookeeper snapshot method by default

hmusum:


mpolden (approved):
Should we change the default value of `snapshotMethod` in `zookeeper-server.def` as well?

hmusum:
> Should we change the default value of `snapshotMethod` in `zookeeper-server.def` as well?

Yes, good point, will add that

mpolden (approved):

PR #20851, opened 2022-01-18
Revert "generate "jsonstring" as config value"

arnej27959:
Reverts vespa-engine/vespa#20834

While backend treats this the same, container-search has a special-case for "xmlstring" when rendering XML.


arnej27959:
@havardpe @hmusum @jonmv FYI

PR #20852, opened 2022-01-18
Tune persistence dynamic throttling policy

vekterli:
@geirst please review
@baldersheim FYI

Currently hardcoded to use values that have been empirically
observed to give good results. Can change this to be configurable
later if needed.


geirst (approved):
👍

PR #20854, opened 2022-01-18
Revert "Invalidate node object cache based on znode version"

jonmv:
Reverts vespa-engine/vespa#20840

I remember now why this doesn't work, and why I didn't to this for the node repo: state for one node is kept across multiple paths.


jonmv:
@mpolden

jonmv:
Or perhaps that's not exactly it ... but it's more complicated here, anyway.

jonmv:
I believe the problem is the interaction between reading live stats, and cached objects? 

Say a snapshot is taken, where node `a` is `reserved`. Then `a` is moved to `active`. Any users of the old snapshot may then be unable to find `a`, as failure to find it could be cached in `active`, and when they try to read it from `reserved`, the stat is gone?

freva (approved):

bratseth:
Another good reason to stop storing nodes in different paths depending on their state.

PR #20858, opened 2022-01-18
Update throttle metric name to use snake_case

vekterli:
@geirst please review. Not yet exposed to aggregators, so can be changed.


geirst (approved):
👍

PR #20860, opened 2022-01-18
Bjorncs/validate pom repositories

bjorncs:
I'm looking into a mechanism to inject internal links to the deployment warnings. That will be part of a next PR.

FYI @bjormel

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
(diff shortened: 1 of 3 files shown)


config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
```diff
@@ -152,16 +152,18 @@ private static void validateImportedPackages(DeployLogger deployLogger, String f
...
-        } catch (ParserConfigurationException e) {
-            throw new RuntimeException(e);
-        } catch (Exception e) {
-            deployLogger.log(Level.INFO, String.format("Unable to parse pom.xml from %s", jarFilename));
+    private void validatePomXml(DeployLogger deployLogger, boolean isHosted, String jarFilename, String pomXmlContent) {
+        if (isHosted) {
+            try {
+                Document pom = DocumentBuilderFactory.newDefaultInstance().newDocumentBuilder()
+                        .parse(new InputSource(new StringReader(pomXmlContent)));
+                validateDependencies(deployLogger, jarFilename, pom);
```
gjoranv, line 160: Validating dependencies can make sense for non-hosted as well, so perhaps we should do that regardless?

gjoranv (approved):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/BundleValidator.java
bjorncs: I'm looking into whether we can split the bundle validator into a public and internal part.

PR #20862, opened 2022-01-18
Support additional model validators

bjorncs:
Inject additional model validators through VespaModelFactory.
Useful for adding validators that are only relevant for hosted and cannot be open-sourced.


hmusum (approved):

gjoranv (commented):
👍

PR #20866, opened 2022-01-19
Support producing all clauses in a referenced condition

bratseth:


container-search/src/main/java/com/yahoo/prelude/semantics/engine/Evaluation.java
```diff
@@ -260,34 +272,49 @@ public void insertItem(Item item, CompositeItem parent, int index, TermType desi
...
     private boolean isEmpty(Item item) {
         if (item == null) return true;
         if (item instanceof QueryTree && ((QueryTree) item).isEmpty()) return true;
         return false;
     }
 
     /** Returns true if the desired type cannot have childCandidate as a child */
-    private boolean incompatible(TermType desiredParentType, Item childCandidate) {
-        return desiredParentType == TermType.EQUIV && childCandidate.getItemType() != Item.ItemType.EQUIV;
+    private boolean incompatible(TermType desiredParentType, CompositeItem parent) {
```
arnej27959, line 307: changed parameter name makes the comment (even more) incomprehensible

arnej27959 (approved):
This is probably OK, but I couldn't quite figure out what the result will be if you have EQUIV and PHRASE items, what will be allowed after this change?

container-search/src/main/java/com/yahoo/prelude/semantics/engine/Evaluation.java
bratseth: You are right; I did a round of cleanup now. Should hopefully address the phrase/equiv clarity issue also.

PR #20867, opened 2022-01-19
Let the Adaptive Executor have both soft and hard limit.

baldersheim:
@geirst or @havardpe PR


havardpe (commented):
This is more like a hard limit and a flag to ignore it. We could consider using a large value instead, like with the basic unbound executors.

baldersheim:
Ideally I agree, however the SingleExecutor requires a size for its lock free queue.
The intention is to remove the blocking part when throttling is rolled out.
Then we should be left with a 'reserved_queue_size'.
I chose to use the same terminology here as used for the single executor.
I suggest we defer making a better api/terminology until remove support for blocking behaviour.

havardpe (approved):

geirst (approved):
👍

PR #20868, opened 2022-01-19
Handle services using restartOnDeploy when checking config convergence [run-systemtest]

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
```diff
@@ -116,6 +140,26 @@ public ServiceResponse getServiceConfigGeneration(Application application, Strin
...
+    private boolean isNotContainer(ServiceInfo serviceInfo) {
+        return ! List.of(CONTAINER.serviceName, QRSERVER.serviceName, METRICS_PROXY_CONTAINER).contains(serviceInfo.getServiceType());
+    }
+
+    // Don't check service in a cluster which uses restartOnDeploy (new config will not be used until service is restarted)
+    private boolean serviceIsInClusterWhichShouldBeChecked(Application application, ServiceInfo serviceInfo) {
+        Set<ApplicationClusterInfo> excludeFromChecking = application.getModel().applicationClusterInfo()
+                                                                     .stream()
+                                                                     .filter(ApplicationClusterInfo::getDeferChangesUntilRestart)
+                                                                     .collect(Collectors.toSet());
```
jonmv, line 158: A bit simpler to map to `info.name()` here, and just use `Set.contains` below.

jonmv (approved):
Just a nit.

configserver/src/main/java/com/yahoo/vespa/config/server/application/ConfigConvergenceChecker.java
hmusum: Yup, will do that in next PR

PR #20869, opened 2022-01-19
Add comments to the tenant info classes

oyving:
fyi @ldalves

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
@@ -13,13 +13,35 @@ import java.util.Objects;
  * @author smorgrav
  */
 public class TenantInfo {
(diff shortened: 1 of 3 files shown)


kkraune (commented):
we need to discuss the email for notifications, otherwise good

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/tenant/TenantInfo.java
```diff
@@ -13,13 +13,35 @@
  * @author smorgrav
  */
 public class TenantInfo {
+    // Editable as 'Tenant Information - Company Name'
+    // Viewable in the 'Account - Profile' section as 'Company Name'
     private final String name;
+
+    // Editable as 'Tenant Information - Email'
+    // Not displayed outside of 'Edit profile'
```
kkraune, line 21: This is to be removed, and can be removed from console view

kkraune (approved):

PR #20875, opened 2022-01-19
Expicitly reject "none"-fieldset for external visitors

vekterli:
@geirst please review
@arnej27959 FYI

It does not make sense for a client to run a visitor that inherently
cannot return any data at all. Explicitly check for, and reject, `[none]`
fieldsets and emit a hopefully helpful error message to use `[id]`
instead.


geirst (approved):
👍

PR #20880, opened 2022-01-20
Use Object format args

hakonhall:


freva (commented):
Probably want to use `Locale.ENGLISH` inside the `String.format()` to get the decimal point right

freva (approved):

PR #20883, opened 2022-01-20
Arnej/deprecate vespalog internals

arnej27959:

@gjoranv please review
note: my plan is to move all the classes that are deprecated here to com.yahoo.log.vespa.internal (or something like that) after we branch for Vespa 8.


gjoranv (approved):
This is great! No hosted apps should be affected by this. The only usage of `com.yahoo.log` I've found is `LogSetup.initVespaLogging()`

We can consider sending a mail to vespa-architects about these changes.

PR #20884, opened 2022-01-20
benchmark std::atomic wait/notify wakeup

havardpe:
also measure cpu usage for waiting threads

@baldersheim please review

Does not seem like the miracle we were hoping for. Seems to behave like a spin-lock with some kind of sleep back-off.


baldersheim (approved):

baldersheim:
There are not that many silver bullets left. One was spent on Karl XII.

PR #20885, opened 2022-01-20
Use default value for ZooKeeper snapshot method for config servers [run-systemtest]

hmusum:
Use default value from config definition.


hmusum:
ping @mpolden

mpolden (approved):

PR #20895, opened 2022-01-21
Jonmv/deployment orchestration for long pipelines

jonmv:
@mpolden please review. We'll hold off documenting the new settings for now, and remove `latest` unless anyone requests it within a reasonable amount of time has passed with `sesparate` as the new default.


mpolden (approved):
Great!

ABI spec in config-model-api needs an update.

PR #20900, opened 2022-01-21
Replace storage operation throttler with vespalib implementation

vekterli:
@geirst please review.

Also move the remaining throttler unit tests to vespalib.


storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
```diff
@@ -143,16 +143,19 @@ selectSequencer(StorFilestorConfig::ResponseSequencerType sequencerType) {
...
 make_operation_throttler_from_config(const StorFilestorConfig& config, size_t num_threads)
 {
     const bool use_dynamic_throttling = (config.asyncOperationThrottlerType == StorFilestorConfig::AsyncOperationThrottlerType::DYNAMIC);
     if (use_dynamic_throttling) {
         auto config_win_size_incr = std::max(config.asyncOperationDynamicThrottlingWindowIncrement, 1);
         auto win_size_increment = std::max(static_cast<size_t>(config_win_size_incr), num_threads);
-        return SharedOperationThrottler::make_dynamic_throttler(win_size_increment);
+        vespalib::SharedOperationThrottler::DynamicThrottleParams params;
+        params.window_size_increment = win_size_increment;
+        params.min_window_size = win_size_increment;
```
geirst, line 155: Was this intentional, using the _win_size_increment_ as _min_window_size_?

geirst (approved):
👍

storage/src/vespa/storage/persistence/filestorage/filestormanager.cpp
vekterli: This is intentional. In the previous implementation the window size increment factor was implicitly used as the minimum window size as well (an artifact of how the `DynamicThrottlePolicy` was constructed). Now the two are separated out into their own parameters.

geirst: 👍

PR #20901, opened 2022-01-21
weighted set and map rendering

arnej27959:

@bratseth for discussion

note: this changes behavior for maps with non-string keys; also I haven't written unit tests yet, so we don't want to merge it.

But we need to decide if we're happy with something like this to paper over the inconsistencies in map and weightedset representations.


container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
```diff
@@ -160,9 +167,13 @@ protected static ObjectMapper createJsonCodec() {
     @Override
     public void init() {
         super.init();
-        debugRendering = false;
-        jsonMaps = false;
-        setGenerator(null, debugRendering);
+        fieldConsumerSettings.debugRendering = false;
```
bratseth, line 170: Since you already created an object for this you could delegate this to it.

container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
```diff
@@ -204,18 +213,25 @@ private void renderTiming() throws IOException {
...
-    private boolean getDebugRendering(Query q) {
-        return q != null && q.properties().getBoolean(DEBUG_RENDERING_KEY, false);
-    }
-
-    private boolean getTensorShortFormRendering(Query q) {
-        if (q == null || q.properties().get(TENSOR_FORMAT) == null)
-            return false;
-        return q.properties().getString(TENSOR_FORMAT).equalsIgnoreCase("short");
+    private void getSettings(Query q) {
+        if (q == null) {
```
bratseth, line 217: Since you already created an object for this you could delegate this to it.

bratseth (commented):
Yes, this looks good to me, and thanks for reminding me about the tensor short forms!
Once we have unit tests it'll be easier to see the consequence of each setting and then I think we should discuss it an at arch meeting.

arnej27959:
@bratseth please take another look now (unit test should be somewhat readable)

bratseth (approved):
LGTM!

arnej27959:
after merging with master branch I am confused by some of the deprecations / comments. The normal code path (after my changes) goes via
1 beginResponse
2 setGenerator
3 calls `createFieldConsumer(JsonGenerator generator, FieldConsumerSettings settings)`
4 calls constructor `FieldConsumer(JsonGenerator generator, FieldConsumerSettings settings)`

so if somebody wants to make their own FieldConsumer, they should overload 3 and from their own constructor call 4.
The comments says to
Use {@link #FieldConsumer(boolean, boolean, boolean)} instead

but if you do that you'll get generator=null and there's no way you can render anything?  It needs to be connected to the output somehow, right?

PR #20903, opened 2022-01-21
improve cpu usage tracking

havardpe:
- add cpu_usage::RUsage primitive used for getrusage
- reduce the number of categories (small initial set)
- use RUsage to fill OTHER category (more complete samples)
- threads no longer track OTHER category explicitly (less overhead)
- changing cpu category to itself is now nop (less overhead)
- make ThreadTracker and SampleConflict private (less API clutter)
- improved testing (sample conflict resolution)
- improves testing (thread tracker implementation)

@arnej27959 please review


havardpe:
The sample conflict resolution test fails. Probably due to a race (have been able to reproduce locally)

havardpe:
should be fixed now, just needed to make sure (multi-)conflict was unavoidable

arnej27959 (approved):
looks good to me

PR #20914, opened 2022-01-24
Support use of dynamic throttling in MergeThrottler component [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Now always uses `DynamicThrottlePolicy` instead of `StaticThrottlePolicy`,
but by default hardwires the min/max limits to _functionally_ act
as if a static policy were in place.

If dynamic throttling is enabled, forwarded (unordered) merges are
now busy-bounced instead of immediately accepted into the pending window
if the window size is exhausted. This so that dynamic throttling
windows on nodes earlier in the forwarding chain can be sized down
implicitly (thus the _dynamic_ aspect of it all). This is done to help
control resource usage and content node queue sizes.

Dynamic throttling may be enabled/disabled live via config. If enabled,
the following (live) config is supported:

 * Minimum window size
 * Maximum window size
 * Window size increment

More may be added (and defaults changed) as appropriate.

Also does a bit of tidying of the `MergeThrottler` code and its immediately
surrounding landscape.


geirst (approved):
👍

PR #20923, opened 2022-01-25
Reduce default timeout for metrics requests from 300s to 25s to ensur…

baldersheim:
…e it is shorter than shutdown timeout targeted to be 30s

@bjorncs or @gjoranv PR


bjorncs (approved):

gjoranv:
👍

PR #20929, opened 2022-01-25
Set a default query timeout

mpolden:
@bratseth

FYI @kkraune


bratseth (approved):
👍

kkraune:
Great! thanks for quick fix :-)

PR #20935, opened 2022-01-26
Support live reconfig of dynamic persistence throttling [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Start moving internal `stor-server` throttling config to use struct
config instead of separate fields, as this is more flexible and better
matches how we configure throttling elsewhere. For now, let dynamic
throttling be enabled via both the new and the old config enum.
Config model will be updated to use the new config struct shortly.


geirst (approved):
👍

PR #20927, opened 2022-01-25
Add feature flag for merge throttling policy

vekterli:
@geirst please review
@baldersheim FYI

Two options; `STATIC` (default) and `DYNAMIC`. Does not currently set
any window size configs based on the existing feature flag for max
pending, so the config defaults will be used for these when dynamic
throttling is enabled.

PR #20947, opened 2022-01-27
Vekterli/merge throttling policy feature flag 2 [run-systemtest]

vekterli:
@geirst please review. Same as #20927 but with one extra stylish override that brings the whole experience together.


geirst (approved):
👍

PR #20948, opened 2022-01-27
Stores all deployed app versions. Stores app/bundle hashes

olaaun:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
```diff
@@ -209,6 +219,13 @@ private static ValidationOverrides allValidationOverrides() {
         return ValidationOverrides.fromXml(validationOverridesContents.toString());
     }
 
+    // Hashes all files that require a deployment to be forwarded to configservers
+    private String calculateBundleHash() {
+        Predicate<String> entryMatcher = name -> !name.contains(deploymentFile) && !name.contains(buildMetaFile);
```
jonmv, line 224: Should at least be `endsWith`, also below.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
```diff
@@ -209,6 +219,13 @@ private static ValidationOverrides allValidationOverrides() {
...
+        SortedMap<String, Long> entryCRCs = ZipStreamReader.getEntryCRCs(new ByteArrayInputStream(zippedContent), entryMatcher);
+        return Hashing.sha1().hashInt(entryCRCs.hashCode()).toString();
```
jonmv, line 226: Hmm, why SHA1 a single number?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ZipStreamReader.java
```diff
@@ -59,6 +63,24 @@ public static void transferAndWrite(OutputStream out, InputStream in, String nam
...
 
+    public static SortedMap<String, Long> getEntryCRCs(InputStream in, Predicate<String> entryNameMatcher) {
+        SortedMap<String, Long> entryCRCs = new TreeMap<>();
+        byte[] buffer = new byte[2048];
+        try (ZipInputStream zipIn = new ZipInputStream(in);
+             OutputStream os = new ByteArrayOutputStream()) {
+            for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
+                if (!entryNameMatcher.test(entry.getName()))
+                    continue;
+                // CRC is not set until entry is read
```
jonmv, line 74: Hmm ... in that case, perhaps we should consider a different hash? I'm not sure which is the most appropriate, but we should try to avoid collisions. It may be that the CRC is OK for this purpose, of course.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ZipStreamReader.java
```diff
@@ -59,6 +63,24 @@ public static void transferAndWrite(OutputStream out, InputStream in, String nam
         }
     }
 
+    public static SortedMap<String, Long> getEntryCRCs(InputStream in, Predicate<String> entryNameMatcher) {
+        SortedMap<String, Long> entryCRCs = new TreeMap<>();
```
jonmv, line 67: Nice.

jonmv (commented):
What will we need the full hash (including deplohment.xml and build meta) for?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -453,9 +453,15 @@ public LockedApplication storeWithUpdatedConfig(LockedApplication application, A
...
-        for (ApplicationVersion version : application.get().versions()) {
-            if (!existingVersions.contains(version)) application = application.withoutVersion(version);
+        var oldVersions = application.get().versions()
+                .stream()
+                .filter(version -> !existingVersions.contains(version))
+                .sorted()
+                .collect(Collectors.toList());
+
+        // Remove any version not deployed anywhere - but keep one
+        for (int i = 0; i < oldVersions.size() - 1; i++) {
```
jonmv, line 463: Hmm. Still worried about getting holes. I was thinking of finding the oldest version deployed anywhere, and then remove everything older than this, but always keep one. That's what we do for packages we store.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
```diff
@@ -226,8 +228,16 @@ private void zoneIdToSlime(ZoneId zone, Cursor object) {
         object.setString(regionField, zone.region().value());
     }
 
-    private void versionsToSlime(List<ApplicationVersion> versions, Cursor object) {
-        versions.forEach(version -> toSlime(version, object.addObject()));
+    private void versionsToSlime(Application application, Cursor object) {
+        // Add all deployed versions as well
```
jonmv, line 232: And this'll be removed later, once we switch to using only the build number+direcly deployed bit as deployment keys :)

jonmv (approved):
Nice :)

jonmv:
Perhaps a unit test using a zip of a couple of files (deployment.xml, build-meta.json and a services.xml, or so)?

jonmv (approved):
Great!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
```diff
@@ -209,6 +219,13 @@ private static ValidationOverrides allValidationOverrides() {
         return ValidationOverrides.fromXml(validationOverridesContents.toString());
     }
 
+    // Hashes all files that require a deployment to be forwarded to configservers
+    private String calculateBundleHash() {
+        Predicate<String> entryMatcher = name -> !name.endsWith(deploymentFile) && !name.endsWith(buildMetaFile);
+        SortedMap<String, Long> entryCRCs = ZipStreamReader.getEntryCRCs(new ByteArrayInputStream(zippedContent), entryMatcher);
+        return String.valueOf(entryCRCs.hashCode());
```
freva, line 226: Shouldn't we calculate this ourselves? `hashCode()` provides no guarantees across execution and this also depends on the particular implementation of the `SortedMap` that we get.

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
jonmv: This is true, but as these are strings and numbers, the hash code won't change. They wouldn't dare XD Doesn't hurt to be more careful, of course.

PR #20960, opened 2022-01-27
Add issuer and fingerprint of peer certificate to connection log

bjorncs:

### DIFF
--- a/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
+++ b/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
@@ -30,6 +30,9 @@ import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
 import javax.net.ssl.StandardConstants;
 import java.net.InetSocketAddress;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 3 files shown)


container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
```diff
@@ -394,6 +402,14 @@ synchronized ConnectionLogEntry toLogEntry() {
             return builder.build();
         }
 
+        private static String certificateFingerprint(byte[] derEncoded) {
+            try {
+                return HexDump.toHexString(MessageDigest.getInstance("SHA-1").digest(derEncoded));
+            } catch (NoSuchAlgorithmException e) {
+                throw new RuntimeException(e);
```
oyving, line 409: It's probably not good to fail with an exception for an issue like this? Log error and return null instead? Or make the fingerprint be the error message? :)

oyving (approved):

oyving:
Can probably merge and then we can discuss the exception throwing separately.

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyConnectionLogger.java
bjorncs: The exception is thrown if the JVM implementation does not provide the SHA-1 digest algorithm. That should never happen.

PR #20969, opened 2022-01-28
Avoid clang warning when discarding values. DO NOT MERGE.

toregge:
@baldersheim : here are the locations triggering clang warnings due to return values being discarded.


baldersheim:
Superseded by #21083

PR #20976, opened 2022-01-28
Use half of available cores to compile query profiles.

baldersheim:
@bratseth PR

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -6244,7 +6244,7 @@
       "public"
     ],
     "methods": [
-      "public void <init>(com.yahoo.search.query.profile.config.QueryProfilesConfig)",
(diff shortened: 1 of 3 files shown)


bratseth (commented):
I don't think we can do this without knowing whether we're bootstrapping or reconfiguring?

baldersheim:
I thought that limiting to num_cores/2 concurrent tasks would be conservative enough.
Perhaps 1/3 or 1/4 are safer numbers.

bratseth:
We need more headroom then. Currently target load is 0.8 (when driven by queries not writes).
Either this cpu capacity will need to be left idle at all times except during reconfig, or reconfig coinciding with traffic load will always lead to overload+query failures.

I think we should leave it at 1 when reconfiguring and use all when bootstrapping.

bjorncs (approved):

PR #20983, opened 2022-01-30
- Storing references instead of just counting them has been default f…

baldersheim:
…or a while now, so making it default in code too.

- Remove option to do just do refcounting.

@bjorncs PR


bjorncs (approved):
Nice 👍

PR #20987, opened 2022-01-31
Set commitLogCount for cluster controller to 10

hmusum:
Other zookeeper server system properties are set in Configurator through config, so do that with `zookeeper.commitLogCount` as well. We could add a feature flag for this if needed.


hmusum:
Based on comments in https://jira.ouryahoo.com/browse/VESPA-22927 this will not work.

baldersheim:
Yes, unfortunately not.

PR #20993, opened 2022-01-31
Since we no longer have a minimum 1G summary file size, we should no …

baldersheim:
…longer have 1G initial buffer.

Reducing first allocation to 64M.

@geirst PR


geirst (approved):
👍

PR #21001, opened 2022-01-31
Deprecate legacy visitor functionality

vekterli:
@baldersheim please review. Ideally the implementation code using the 2nd pass statistics would be removed now as well, but due to the statistics wiring issues mentioned in #20997 that will have to wait until Vespa 8.


vekterli:
Missed a deprecation warning, will update

baldersheim (approved):

PR #21002, opened 2022-01-31
Jonmv/simultaneous rollouts

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
@@ -118,7 +118,7 @@ public class DeploymentStatus {
 
         return ! allJobs.matching(job -> criticalJobs.contains(job.id()))
                         .failing()
-                        .not().withStatus(RunStatus.outOfCapacity)
+                        .not().outOfTestCapacity()
                         .isEmpty();
     }
 
@@ -126,39 +126,29 @@ public class DeploymentStatus {
         if (visited.contains(current))
             return dependents.contains(current);
 
-        if (dependency == current) {
+        if (dependency == current)
             dependents.add(current);
-            visited.add(current);
-            return true;
-        }
+        else
+            for (StepStatus dep : current.dependencies)
+                if (fillDependents(dependency, visited, dependents, dep))
+                    dependents.add(current);
 
-        for (StepStatus dep : current.dependencies)
-            if (fillDependents(dependency, visited, dependents, dep))
-                dependents.add(current);
         visited.add(current);
-
         return dependents.contains(current);
     }
 
-    /** Whether any jobs of the given instance, or in an instance dependent on it, are failing with other errors than lack of capacity in a test zone. */
-    public boolean hasFailures(InstanceName instance) {
-        Set<StepStatus> dependents = new HashSet<>();
-        for (StepStatus instanceStep : instanceSteps().values())
-            fillDependents(instanceSteps().get(instance), new HashSet<>(), dependents, instanceStep);
-
-        Set<InstanceName> dependentInstances = dependents.stream()
-                                                         .map(StepStatus::instance)
-                                                         .collect(Collectors.toSet());
-        return ! allJobs.instance(dependentInstances)
-                        .failing()
-                        .not().withStatus(RunStatus.outOfCapacity)
+    /** Whether any job is failing on anything older than version, with errors other than lack of capacity in a test zone.. */
+    public boolean hasFailures(ApplicationVersion version) {
+        return ! allJobs.failing()
+                        .not().outOfTestCapacity()
+                        .matching(job -> job.lastTriggered().get().versions().targetApplication().compareTo(version) < 0)
                         .isEmpty();
     }
 
     /** Whether any jobs of this application are failing with other errors than lack of capacity in a test zone. */
     public boolean hasFailures() {
         return ! allJobs.failing()
-                        .not().withStatus(RunStatus.outOfCapacity)
+                        .not().outOfTestCapacity()
                         .isEmpty();
     }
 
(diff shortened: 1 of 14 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTrigger.java
```diff
@@ -321,19 +322,18 @@ private ApplicationController applications() {
...
-                      .ifPresent(readyAt -> {
-                          jobs.add(deploymentJob(status.application().require(job.application().instance()),
-                                                 versions,
-                                                 job.type(),
-                                                 status.instanceJobs(job.application().instance()).get(job.type()),
-                                                 readyAt));
-                      });
+        Map<JobId, List<DeploymentStatus.Job>> jobsToRun = status.jobsToRun();
+        jobsToRun.forEach((job, versionsList) -> {
+            versionsList.get(0).readyAt()
```
jonmv, line 327: Better to store the readyAt where we compute the job versions, and have all data available.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTrigger.java
```diff
@@ -348,27 +348,49 @@ private boolean isUnhealthyInAnotherZone(Application application, JobId job) {
...
+                  }
+              });
+    }
 
-        Run last = status.lastTriggered().get();
-        if (versionsList.stream().noneMatch(versions ->    versions.targetsMatch(last.versions())
-                                                        && versions.sourcesMatchIfPresent(last.versions())))
-            controller.jobController().abort(last.id());
+    /** Returns whether the job is free to start, and also aborts it if it's running with outdated versions. */
+    private boolean abortIfRunning(DeploymentStatus status, Map<JobId, List<DeploymentStatus.Job>> jobs, JobId job) {
```
jonmv, line 365: Should probably change the name of this ... :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTrigger.java
```diff
@@ -348,27 +348,49 @@ private boolean isUnhealthyInAnotherZone(Application application, JobId job) {
...
+                      controller.jobController().abort(last.id());
+                  }
+              });
+    }
 
-        Run last = status.lastTriggered().get();
-        if (versionsList.stream().noneMatch(versions ->    versions.targetsMatch(last.versions())
-                                                        && versions.sourcesMatchIfPresent(last.versions())))
-            controller.jobController().abort(last.id());
+    /** Returns whether the job is free to start, and also aborts it if it's running with outdated versions. */
```
jonmv, line 364: Hmm, also aborts an outdated, running, corresponding production test, if one exists ...

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTrigger.java
```diff
@@ -348,27 +348,49 @@ private boolean isUnhealthyInAnotherZone(Application application, JobId job) {
...
-        return false;
+        return ! blocked;
     }
 
     // ---------- Change management o_O ----------
 
     private boolean acceptNewApplicationVersion(DeploymentStatus status, InstanceName instance) {
         if (status.application().deploymentSpec().instance(instance).isEmpty()) return false; // Unknown instance.
-        if (status.hasFailures()) return true; // Allow changes to fix upgrade or previous revision problems.
+        if (status.hasFailures(instance)) return true; // Allow changes to fix upgrade or previous revision problems.
```
jonmv, line 391: This should be refined even further, to not allow a revision if it is the new revision that is failing eager system/staging tests ... :P

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTrigger.java
```diff
@@ -348,27 +348,49 @@ private boolean isUnhealthyInAnotherZone(Application application, JobId job) {
...
         DeploymentInstanceSpec spec = status.application().deploymentSpec().requireInstance(instance);
         Change change = status.application().require(instance).change();
-        if (change.platform().isPresent() && spec.upgradeRollout() == DeploymentSpec.UpgradeRollout.separate) return false;
```
jonmv, line 371: Woop.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
```diff
@@ -1168,48 +1171,65 @@ public void testDeployComplicatedDeploymentSpec() {
...
         app1.runJob(testUsEast3);
         app1.runJob(productionApSoutheast1);
+        tester.clock().advance(Duration.ofSeconds(1));
+        app1.runJob(productionUsEast3);      // R
+        tester.clock().advance(Duration.ofHours(2));
+        app1.runJob(productionEuWest1);      // R
+        tester.clock().advance(Duration.ofMinutes(330));
+        app1.runJob(testApNortheast1);       // R
+        app1.runJob(testApNortheast2);       // R
+        app1.runJob(testUsEast3);            // R
```
jonmv, line 1213: Should probably make a job in between production-us-east-3 and this fail, so we see that the upgrade-only run for test-us-east-3 is abandoned, and the revision is deployed earlier, to potentially fix that upgrade failure.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentTriggerTest.java
```diff
@@ -1224,41 +1244,48 @@ public void testDeployComplicatedDeploymentSpec() {
...
         assertEquals(Change.empty(), app3.instance().change());
     }
 
+    // TODO test new revision allowed exactly when dependent instance is failing
+    // TODO test multi-tier pipeline with various policies
+    // TODO test multi-tier pipeline with upgrade after new version is the candidate
+
     @Test
-    public void testChangeCompletion() {
+    public void testRevisionJoinsUpgradeWithSeparateRollout() {
```
jonmv, line 1271: And there should be a similar test for each of the other policies here ...

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
```diff
@@ -107,6 +110,51 @@ public JobList jobs() {
...
+                dependents.add(current);
+        visited.add(current);
+
+        return dependents.contains(current);
+    }
+
+    /** Whether any jobs of the given instance, or in an instance dependent on it, are failing with other errors than lack of capacity in a test zone. */
+    public boolean hasFailures(InstanceName instance) {
+        Set<StepStatus> dependents = new HashSet<>();
+        for (StepStatus instanceStep : instanceSteps().values())
```
jonmv, line 146: Hmm, this should use all steps as starting points, not just all instance steps.

jonmv: Urgh, this also doesn't catch implicit tests ...

jonmv: Perhaps the logic should instead be to check where there are any failures with an older revision than what we're considering deploying ...

jonmv: ... and just check all jobs. Tests are tricky, because they can be run under any instance—there could even be tests in an earlier instance that are failing because of a platform-revision combination in a later instance that just wasn't present in any of the earlier ones, so the "dependent instances" idea break down. :'( 

But considering revision ordering should always be correct for this use case ... I think!

mpolden (approved):
Great!

jonmv:
PTAL, will follow up with another PR.

mpolden (approved):

PR #21003, opened 2022-02-01
Balder/disable vespamalloc for small utilities

baldersheim:
@hmusum or @arnej27959 PR

### DIFF
--- a/vespabase/src/common-env.sh
+++ b/vespabase/src/common-env.sh
@@ -148,15 +148,15 @@ consider_fallback VESPA_VALGRIND_OPT "--num-callers=32 \
 --show-reachable=yes \
 ${VESPA_VALGRIND_SUPPREESSIONS_OPT}"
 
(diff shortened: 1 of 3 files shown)


hmusum (approved):

vespabase/src/common-env.sh
```diff
@@ -148,15 +148,15 @@ consider_fallback VESPA_VALGRIND_OPT "--num-callers=32 \
 --show-reachable=yes \
 ${VESPA_VALGRIND_SUPPREESSIONS_OPT}"
 
-consider_fallback VESPA_USE_HUGEPAGES_LIST  $(get_var "hugepages_list")
```
bratseth, line 151: Don't you have to escape the inner quotes?

baldersheim: I thought so too, but it worked like this.
All the quoting is shell scripts is a true nightmare.
I think we might need some test for this, as it did not work as intended.

PR #21006, opened 2022-02-01
Configure load balancer with nodes that will become active

mpolden:
Before this we only considered the current nodes (before activate-transaction is
committed), which worked in most cases because prepare does all the necessary
work. However, this is not the case when nodes are removed.

@bratseth


bratseth (approved):
👍

PR #21011, opened 2022-02-01
Bratseth/avoid schema dir

bratseth:


config-model/src/main/java/com/yahoo/searchdefinition/ApplicationBuilder.java
```diff
@@ -201,28 +195,28 @@ public Schema add(Schema schema) {
...
             throw new ParseException(stream.formatException(Exceptions.toMessageString(pe)));
         }
     }
 
-    private void addRankProfileFiles(Schema schema, String schemaPath) {
-        if (applicationPackage == null || schemaPath == null) return;
-        Path rankProfilePath = Path.fromString(schemaPath).append(schema.getName());
+    private void addRankProfileFiles(Schema schema) {
+        if (applicationPackage == null) return;
+        Path rankProfilePath = ApplicationPackage.SCHEMAS_DIR.append(schema.getName());
```
hmusum, line 219: Will not work with old schema dir (ApplicationPackage.SEARCH_DEFINITIONS_DIR), is that intended?

hmusum (approved):
One comment, LGTM

config-model/src/main/java/com/yahoo/searchdefinition/ApplicationBuilder.java
bratseth: Yes, but maybe it'll just cause confusion. I added it.

bratseth:
... adding that broke some tests.

PR #21017, opened 2022-02-02
Strip spaces from node count range

mpolden:
@jonmv


jonmv (approved):
This helps. I also got the same error trying to use the suggested default for node resources, even _after_ stripping the spaces, so perhaps there's more.

mpolden:
Will fix ranges inside resource spec in a separate PR.

jonmv:
Great!

PR #21022, opened 2022-02-02
Exit with non-zero if no arguments are given

mpolden:
@havardpe


havardpe (approved):

havardpe:
On the other hand, maybe Run for the root command should fail, since ending up calling it means you have some arguments that could not be matched to other (sub-)commands?

havardpe (approved):

PR #21025, opened 2022-02-02
Differentiate deployment method when clearing old app versions

olaaun:
Alternatively, we could delete directly deployed versions elsewhere, e.g. in `JobController.deploy`

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/ApplicationVersion.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/ApplicationVersion.java
@@ -185,7 +185,7 @@ public class ApplicationVersion implements Comparable<ApplicationVersion> {
         if (buildNumber().isEmpty() || o.buildNumber().isEmpty())
             return Boolean.compare(buildNumber().isPresent(), o.buildNumber.isPresent()); // Unknown version sorts first
 
-        if (deployedDirectly || o.deployedDirectly)
+        if (deployedDirectly != o.deployedDirectly)
             return Boolean.compare( ! deployedDirectly, ! o.deployedDirectly); // Directly deployed versions sort first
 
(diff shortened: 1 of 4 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/deployment/ApplicationVersion.java
```diff
@@ -185,7 +185,7 @@ public int compareTo(ApplicationVersion o) {
         if (buildNumber().isEmpty() || o.buildNumber().isEmpty())
             return Boolean.compare(buildNumber().isPresent(), o.buildNumber.isPresent()); // Unknown version sorts first
 
-        if (deployedDirectly || o.deployedDirectly)
+        if (deployedDirectly != o.deployedDirectly)
```
jonmv, line 188: Oops!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -445,19 +446,22 @@ public LockedApplication storeWithUpdatedConfig(LockedApplication application, A
...
                 .stream()
                 .flatMap(instance -> instance.deployments().values().stream())
                 .map(Deployment::applicationVersion)
-                .sorted()
-                .findFirst()
-                .orElse(ApplicationVersion.unknown);
+                .collect(toMap(
+                        ApplicationVersion::isDeployedDirectly,
+                        Function.identity(),
+                        (v1, v2) -> v1.compareTo(v2) < 0 ? v1 : v2
```
jonmv, line 458: Hmm, these have a separate counter per job ... Urgh :@

jonmv: The directly deployed ones, that is.

jonmv (approved):

PR #21035, opened 2022-02-02
Improve vespa curl documentation

mpolden:
@freva


freva (approved):
:+1: 
Consider also dropping the `;` at the end of YQLs, optional since #20595

PR #21041, opened 2022-02-02
Revert "Use LinkedHashMap to get deterministic config order"

bratseth:
Reverts vespa-engine/vespa#21031

well this made things less deterministic


baldersheim (commented):
👍

PR #21049, opened 2022-02-03
Handle node resource ranges

mpolden:
Deliberately un-fancy as I don't want to reimplement all the validation we
already have on deploy.

@jonmv


jonmv (approved):
Nice.

PR #21056, opened 2022-02-03
Use a conservative flush strategy before we reach the proton resource…

geirst:
… limits for memory and disk.

This should improve protons ability to have enough headroom to handle transient resource usage
(memory index and disk index fusion) before reaching feed blocked inside proton.

@baldersheim please review


baldersheim (approved):
Smooth 👍

PR #21063, opened 2022-02-03
Do not allocate nodes on suspended hosts

freva:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -728,7 +728,7 @@ public class Nodes {
         return canAllocateTenantNodeTo(host, zone.getCloud().dynamicProvisioning());
     }
 
-    public static boolean canAllocateTenantNodeTo(Node host, boolean dynamicProvisioning) {
+    public boolean canAllocateTenantNodeTo(Node host, boolean dynamicProvisioning) {
         if ( ! host.type().canRun(NodeType.tenant)) return false;
         if (host.status().wantToRetire()) return false;
         if (host.allocation().map(alloc -> alloc.membership().retired()).orElse(false)) return false;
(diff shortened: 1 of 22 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -728,17 +733,27 @@ public boolean canAllocateTenantNodeTo(Node host) {
         return canAllocateTenantNodeTo(host, zone.getCloud().dynamicProvisioning());
     }
 
-    public static boolean canAllocateTenantNodeTo(Node host, boolean dynamicProvisioning) {
+    public boolean canAllocateTenantNodeTo(Node host, boolean dynamicProvisioning) {
         if ( ! host.type().canRun(NodeType.tenant)) return false;
         if (host.status().wantToRetire()) return false;
         if (host.allocation().map(alloc -> alloc.membership().retired()).orElse(false)) return false;
+        if (suspended(host)) return false;
```
mpolden, line 740: Doesn't this mean that a given deployment will send one request per host in the node repository to the orchestrator? In larger zones that is ~400.

mpolden: Possibly double that as we do prepare, then prepareWithLocks.

freva: Yes, I not sure how expensive it is or if there is a more efficient way to do this, otherwise I guess we'll have to cache this?

mpolden: Maybe we can extend the orchestrator API to allow listing all suspended hosts, and use that? So we get away with one request instead of hundreds.

@hakonhall may have some thoughts on performance impact.

bratseth: It's cached in the orchestrator, which is a local component. Should be fine.

bratseth (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
hakonhall: The cheapest API would be to get all HostInfos for an application in the Orchestrator, which has the same performance per application as  getNodeStatus() has per node.  If there are any performance issues we could expose the former in the Orchestrator API.

PR #21067, opened 2022-02-04
Use installed config schema if none is given as option

hmusum:


baldersheim (approved):

gjoranv:
Is it possible that the schema does not exist in the configdefinitions folder? If so, what happens?

hmusum:
Yes, that might happen, no error checking in the original code, I'll try to add some

PR #21072, opened 2022-02-04
Shutdown reconfiguration thread in a more controlled way [run-systemtest]

bjorncs:
Use interrupts is a horrible mechanism as we cannot control which part of the code receives the signal.


container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
```diff
@@ -146,7 +148,7 @@ public void start() {
         ContainerBuilder builder = createBuilderWithGuiceBindings();
         configurer = createConfigurer(builder.guiceModules().activate());
         initializeAndActivateContainer(builder, () -> {});
-        startReconfigurerThread();
+        reconfigurerThread.start();
```
baldersheim, line 151: setDaemon() ?

baldersheim (approved):

PR #21073, opened 2022-02-04
Add hash function

bratseth:


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/HashExpression.java
```diff
@@ -0,0 +1,91 @@
...
+    @Override
+    protected void doVerify(VerificationContext context) {
+        String outputField = context.getOutputField();
+        if (outputField == null)
+            throw new VerificationException(this, "No output field in this statement: " +
+                                                  "Don't know what value to hash to.");
+        DataType outputFieldType = context.getInputType(this, outputField);
+        if (outputFieldType != DataType.INT && outputFieldType != DataType.LONG)
+            throw new VerificationException(this, "The type of the output field " + outputField +
+                                                  " is not an int or long but " + outputField);
```
geirst, line 72: I guess this should be outputFieldType instead of outputField?

indexinglanguage/src/test/java/com/yahoo/vespa/indexinglanguage/ScriptTestCase.java
```diff
@@ -99,6 +98,52 @@ public void testLiteralBoolean() throws ParseException {
...
+        adapter.setValue("myText", new StringFieldValue("input text"));
+        expression.setStatementOutput(new DocumentType("myDocument"), intField);
+
+        // Necessary to resolve output type
+        VerificationContext verificationContext = new VerificationContext(adapter);
+        assertEquals(DataType.LONG, expression.verify(verificationContext));
+
+        ExecutionContext context = new ExecutionContext(adapter);
+        context.setValue(new StringFieldValue("input text"));
+        expression.execute(context);
```
geirst, line 141: The setup and execution is identical to that used in testIntHash(), except the data type and field name. Consider refactor into a helper function. As it stands I think it is harder to read as I was looking for differences between the two tests.

geirst (commented):
👍

geirst (approved):

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/HashExpression.java
```diff
@@ -0,0 +1,91 @@
...
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Hashes a string value to a long or int (by type inference on the target value).
+ *
+ * @author bratseth
+ */
+public class HashExpression extends Expression  {
+
+    private final HashFunction hasher = Hashing.sipHash24();
```
baldersheim, line 22: Does the hash function matter ? If so I guess we should make it configurable.
We have made bad choices for hash function in the past.

bratseth: Yes it matters, but if we make it configurable we leave the choice to somebody who's less well able to make it. Here I think performance matters less than minimizing collisions given real text and then siphash seems like a good choice,. although I'm not sure it is the very best. 

We can also make it configurable later, with this as the default, if the need surfaces.

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/HashExpression.java
bratseth: Yes, thanks.

PR #21084, opened 2022-02-06
Deprecate binary encoding outside VDS

bratseth:


arnej27959:
Henning should look at this - can we switch to using protobuf serialization?

baldersheim (approved):

baldersheim:
Yes, we can add a new map entry and let the backend handle both.
If we are quick and enable protobuf now, we can drop the old one at 8.

bratseth:
I don't think we need a major version change to do that?

baldersheim:
Currently we require you have a recent MAJOR version in order to do rolling upgrade to next MAJOR.
We do not have any such requirement/procedure/enforcement for minor upgrades. We should have that as that would make it a lot easier to maintain our code base.

bratseth:
Good point, how about https://github.com/vespa-engine/frontpage/pull/232
Though we already say that if you don't want to maybe lose your data you need to upgrade content nodes through every minor so I'm not sure this makes any difference in practice.

baldersheim:
@geirst @vekterli FYI

bratseth:
How about changing the content node data compatibility to 2 months as well?
Requiring everyone to goo through every minor seems a bit unrealistic.

PR #21089, opened 2022-02-07
Revert "Shutdown reconfiguration thread in a more controlled way [run-systemtest]"

bjorncs:
Reverts vespa-engine/vespa#21072


bjorncs:
Looks like this broke the system tests on factory

hmusum (approved):

PR #21090, opened 2022-02-07
- Adde featureflag for controlling environment variables.

baldersheim:
- Use general environment mechanism for existing ad-hoc environment variables.

@hmusum or @bjorncs PR

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
@@ -198,6 +198,14 @@ public class PermanentFlags {
             APPLICATION_ID, HOSTNAME
     );
 
+    public static final UnboundListFlag<String> ENVIRONMENT_VARIABLES = defineListFlag(
+            "environment-variables", List.of(), String.class,
+            "A list of environment variables set for all services. " +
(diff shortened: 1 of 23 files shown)


hmusum (approved):

flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
```diff
@@ -198,6 +198,14 @@
             APPLICATION_ID, HOSTNAME
     );
 
+    public static final UnboundListFlag<String> ENVIRONMENT_VARIABLES = defineListFlag(
+            "environment-variables", List.of(), String.class,
+            "A list of environment variables set for all services. " +
+                    "Each item should be on the form <ENV_VAR>=<VALUE>",
+            "Takes effect on service restart",
+            ZONE_ID, APPLICATION_ID
```
bjorncs, line 206: Including cluster type/name as dimension would be useful, although not sure how we would wire it in through the `ModelContext`

bjorncs (approved):

flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
baldersheim: Future extension ...

PR #21092, opened 2022-02-07
Bratseth/presentation format tensors

bratseth:


container-search/src/main/java/com/yahoo/search/query/profile/types/FieldDescription.java
```diff
@@ -81,16 +81,14 @@ public FieldDescription(CompoundName name, FieldType type, String aliases, boole
     /**
      * Creates a field description
      *
-     * @param name the name of the field
+     * @param name the name of the field, emnpty means it describes the value held by  the query profile itself
```
lesters, line 84: emnpty -> empty

lesters (approved):
LGTM!

PR #21093, opened 2022-02-07
Bjorncs/dont use interrupts [run-systemtest]

bjorncs:
FYI @hmusum

### DIFF
--- a/config/abi-spec.json
+++ b/config/abi-spec.json
@@ -297,5 +297,16 @@
     "fields": [
       "public final java.lang.String payload"
     ]
+  },
+  "com.yahoo.config.subscription.SubscriberClosedException": {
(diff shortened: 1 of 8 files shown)


config/abi-spec.json
```diff
@@ -297,5 +297,16 @@
...
+    "interfaces": [],
+    "attributes": [
+      "public"
+    ],
+    "methods": [
+      "public void <init>()"
+    ],
+    "fields": []
   }
 }
```
baldersheim, line 312: Missing eol

baldersheim (approved):

config/abi-spec.json
bjorncs: The abispec file is automatically generated

PR #21096, opened 2022-02-07
Disable vespamalloc when address sanitizer is used.

toregge:
@baldersheim : please review


baldersheim (approved):

toregge:
vespamalloc is not built when the CMake variable VESPA_USE_SANITIZER is set. This breaks building the rpms.

baldersheim:
@toregge I this still relevant, or should it be closed.

toregge:
It can be closed. Vespamalloc is not built when address sanitizer is enabled and is thus inherently disabled.

PR #21097, opened 2022-02-07
Increase minimum number of Jetty threads [run-systemtest]

hmusum:
Ref. https://github.com/vespa-engine/vespa/pull/21095
(Error message when having too few threads is misleading, need 6 threads, not 5)


baldersheim (approved):

baldersheim:
What caused the requirement for number of threads to go up ?

hmusum:
That's unknown, no changes in the version (7.540.14) where this started failing that seem to be be related

PR #21099, opened 2022-02-07
Revert "Fail if schema file cannot be opened"

hmusum:
Reverts vespa-engine/vespa#21091

Breaks system tests


baldersheim (approved):

hmusum:
We cannot fail if schema is missing, as not all configs are Vespa configs and there will be no place to read schema file from in that case

PR #21106, opened 2022-02-08
- Generate mode modern c++ code.

baldersheim:
- Use existing predefined types.
- Allow for better management by allowing large string vectors to be mmapped.

@vekterli PR

### DIFF
--- a/config/src/tests/configgen/vector_inserter.cpp
+++ b/config/src/tests/configgen/vector_inserter.cpp
@@ -100,8 +100,9 @@ TEST("require that vector of bool can be inserted") {
     ASSERT_TRUE(vector[2]);
 }
 
-TEST("require that vector of string can be inserted") {
-    std::vector<vespalib::string> vector;
+template<typename V>
+void
+verify_vector_strings_can_be_inserted(V vector) {
     Slime slime;
     Cursor & root = slime.setArray();
     root.addString("foo");
@@ -115,19 +116,9 @@ TEST("require that vector of string can be inserted") {
     ASSERT_EQUAL("baz", vector[2]);
 }
 
-TEST("require that vector of string can be inserted") {
-    StringVector vector;
-    Slime slime;
-    Cursor & root = slime.setArray();
-    root.addString("foo");
-    root.addString("bar");
-    root.addString("baz");
-    VectorInserter inserter(vector);
-    root.traverse(inserter);
-    ASSERT_EQUAL(3u, vector.size());
-    ASSERT_EQUAL("foo", vector[0]);
-    ASSERT_EQUAL("bar", vector[1]);
-    ASSERT_EQUAL("baz", vector[2]);
+TEST("require that different vectors of strings can be inserted") {
+    verify_vector_strings_can_be_inserted(std::vector<vespalib::string>());
+    verify_vector_strings_can_be_inserted(StringVector());
 }
 
 TEST_MAIN() { TEST_RUN_ALL(); }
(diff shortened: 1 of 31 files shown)


config/src/tests/configgen/vector_inserter.cpp
```diff
@@ -106,7 +107,22 @@ TEST("require that vector of string can be inserted") {
...
-    VectorInserter<vespalib::string> inserter(vector);
+    VectorInserter inserter(vector);
+    root.traverse(inserter);
+    ASSERT_EQUAL(3u, vector.size());
+    ASSERT_EQUAL("foo", vector[0]);
+    ASSERT_EQUAL("bar", vector[1]);
+    ASSERT_EQUAL("baz", vector[2]);
+}
+
+TEST("require that vector of string can be inserted") {
```
vekterli, line 118: Consider renaming (and possibly factoring out) test since it's currently duplicated with the above test and only differs in vector type.

configgen/src/main/java/com/yahoo/config/codegen/CppClassBuilder.java
```diff
@@ -357,7 +330,7 @@ void writeConfigClassFunctionDeclarations(Writer w, String className, CNode node
     }
 
     void writeStructFunctionDeclarations(Writer w, String className, CNode node, String indent) throws IOException {
-        w.write(indent + className + "(const std::vector<vespalib::string> & __lines);\n");
+        w.write(indent + className + "(const " + vectorTypeDefs.get("vespalib::string") + " & __lines);\n");
```
vekterli, line 333: Tangentially related to this PR; we should stop using identifiers containing double underscores. These are reserved for internal compiler/libc usage and I can't see any point in actually having them here. Just makes the code less readable, and since we seem to be using config input names verbatim anyway this doesn't help if a name collides with a reserved C++ keyword...

config/src/vespa/config/configgen/value_converter.h
```diff
@@ -7,7 +7,7 @@
 
 namespace config::internal {
 
-void requireValid(const vespalib::string & __fieldName, const ::vespalib::slime::Inspector & __inspector);
+void requireValid(vespalib::stringref __fieldName, const ::vespalib::slime::Inspector & __inspector);
```
vekterli, line 10: Related to the other comment, we should stop using double underscores here as well

vekterli (approved):
:+1:

config/src/vespa/config/common/types.h
```diff
@@ -2,12 +2,14 @@
...
 
 #include <vespa/vespalib/stllike/string.h>
+#include <vespa/vespalib/stllike/allocator.h>
 #include <vector>
 #include <map>
 
 namespace config {
 
-using StringVector = std::vector<vespalib::string>;
+//using StringVector = std::vector<vespalib::string>;
```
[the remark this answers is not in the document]
baldersheim: Fixed

config/src/vespa/config/common/configparser.h
```diff
@@ -14,81 +14,95 @@ namespace config {
  */
 class ConfigParser {
 public:
-    using vsvector = StringVector;
+    class Cfg {
+    public:
+        Cfg(const std::vector<vespalib::string> & v)
```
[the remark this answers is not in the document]
baldersheim: Explicit implicit to get automatic conversion. Intention to make another pass at this to avoid the vector of strings completely in not too distant future.

configgen/src/main/java/com/yahoo/config/codegen/CppClassBuilder.java
baldersheim: Agree, but I think that belongs in a different PR

config/src/vespa/config/configgen/value_converter.h
baldersheim: Agree, but I think that belongs in a different PR

config/src/tests/configgen/vector_inserter.cpp
baldersheim: Fixed

PR #21118, opened 2022-02-09
Wire in mallopt(in param, int value) interface in vespamalloc and ver…

baldersheim:
…ify simple wiring.

@vekterli @havardpe @toregge  PR


toregge (approved):

vekterli (approved):
:+1: Nice, I didn't know `M_MMAP_THRESHOLD` was an existing thing

havardpe:
Are we planning to be able to revert back to a previously set value? How will this work? Will settings be inherited when starting new threads?

baldersheim:
It is thread local, so no effect on other threads for vespamalloc. Effect is implementation dependant for others.
Will also only be used if vespamalloc is used.

baldersheim:
The usage is scoped.  See #21120 This is just adapting to an existing incomplete API.

PR #21126, opened 2022-02-09
Request, list, and approve ssh access

olaaun:
As draft so we can use this as platform for discussion.

I assume `AccessControlService` can/should be reused for this, but for now it will communicate with different zms instances

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/athenz/AthenzAccessControlService.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/athenz/AthenzAccessControlService.java
@@ -5,6 +5,7 @@ package com.yahoo.vespa.hosted.controller.api.integration.athenz;
 import com.yahoo.config.provision.TenantName;
 import com.yahoo.vespa.athenz.api.AthenzDomain;
 import com.yahoo.vespa.athenz.api.AthenzGroup;
+import com.yahoo.vespa.athenz.api.AthenzIdentity;
 import com.yahoo.vespa.athenz.api.AthenzRole;
 import com.yahoo.vespa.athenz.api.AthenzUser;
 import com.yahoo.vespa.athenz.client.zms.ZmsClient;
@@ -39,7 +40,7 @@ public class AthenzAccessControlService implements AccessControlService {
         if(!isVespaTeamMember(user)) {
             throw new IllegalArgumentException(String.format("User %s requires manual approval, please contact Vespa team", user.getName()));
         }
-        Map<AthenzUser, String> users = zmsClient.listPendingRoleApprovals(dataPlaneAccessRole);
+        Map<AthenzIdentity, String> users = zmsClient.listPendingRoleApprovals(dataPlaneAccessRole);
         if (users.containsKey(user)) {
             zmsClient.approvePendingRoleMembership(dataPlaneAccessRole, user, expiry, Optional.empty());
             return true;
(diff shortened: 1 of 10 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/athenz/AthenzAccessControlService.java
```diff
@@ -53,6 +56,50 @@ public boolean approveDataPlaneAccess(AthenzUser user, Instant expiry) {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * @return Whether the ssh access role has any pending role membership requests
+     */
+    public boolean hasPendingAccessRequests(TenantName tenantName) {
+        var role = sshRole(tenantName);
+        var pendingApprovals = vespaZmsClient.listPendingRoleApprovals(role);
+        return !pendingApprovals.isEmpty();
```
freva, line 65: Check that `pendingApprovals` contains `vespaTeam`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -286,6 +287,8 @@ private HttpResponse handleGET(Path path, HttpRequest request) {
 
     private HttpResponse handlePUT(Path path, HttpRequest request) {
         if (path.matches("/application/v4/tenant/{tenant}")) return updateTenant(path.get("tenant"), request);
+        if (path.matches("/application/v4/tenant/{tenant}/access/ssh/request")) return requestSshAccess(path.get("tenant"), request);
+        if (path.matches("/application/v4/tenant/{tenant}/access/ssh/approve")) return approveAccessRequest(path.get("tenant"), request);
```
freva, line 291: Need to limit this API to tenant (administrators(?)) only in `controller-api` (PathGroup|Policy|RoleDefinition)

Maybe consider ordering path arguments so that approve is before ssh, then we can limit access to all ../access/* APIs

vespa-athenz/src/main/java/com/yahoo/vespa/athenz/client/zms/DefaultZmsClient.java
```diff
@@ -297,16 +297,15 @@ public boolean deletePolicyRule(AthenzDomain athenzDomain, String athenzPolicy,
 
         return roleEntity.roleMembers().stream()
                 .filter(RoleEntity.Member::pendingApproval)
-                .filter(re -> AthenzIdentities.USER_PRINCIPAL_DOMAIN.equals(AthenzIdentities.from(re.memberName()).getDomain()))
```
freva, line 300: Not sure why this was here to begin with, any possible side effects of removing this for the existing users of this method?

freva (approved):

vespa-athenz/src/main/java/com/yahoo/vespa/athenz/client/zms/DefaultZmsClient.java
olaaun: I think it should be fine - all users of the function are only interested in finding whether a specific identity is pending.

@tokle was there any specific reason for the filter?

tokle: Not sure, @andreer do you know?

PR #21142, opened 2022-02-11
Require shutdown to complete faster.

baldersheim:
@jonmv or @bjorncs PR


clustercontroller-reindexer/src/main/java/ai/vespa/reindexing/ReindexingMaintainer.java
```diff
@@ -100,17 +100,23 @@ public void deconstruct() {
                 reindexer.shutdown();
 
             executor.shutdown();
-            if ( ! executor.awaitTermination(45, TimeUnit.SECONDS))
+
+            executor.awaitTermination(5, TimeUnit.SECONDS); // Give it 5s to complete gracefully.
```
jonmv, line 104: I'm not sure 5s is enough in all cases, but reindexing a few documents an additional itme should be harmless anyway.

PR #21150, opened 2022-02-11
Only shutdown non-null configurer

mpolden:
Happens if `app.start()` throws before `configurer` is assigned in
https://github.com/vespa-engine/vespa/blob/master/jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java#L149-L162

@bjorncs


mpolden:
or @gjoranv

bjorncs (approved):

gjoranv:
👍

PR #21172, opened 2022-02-14
Deprecate service-alias and endpoint-alias

mpolden:
@tokle

FYI @gjoranv


mpolden:
Looks like the outer `<aliases>` element can be removed entirely. I'll deprecate that instead.

tokle (approved):

PR #21183, opened 2022-02-14
Let allocations larger than 1M be handled directly by mmap

baldersheim:
@havardpe or @vekterli PR


vekterli (approved):
:+1: might 1M be a bit low? Can envision lots of mmap activity for many allocs

PR #21188, opened 2022-02-14
Revert "Revert "Let allocations larger than 1M be handled directly by mmap""

baldersheim:
Reverts vespa-engine/vespa#21185
@vekterli PR


baldersheim:
After #21187

vekterli (commented):
:+1:

PR #21189, opened 2022-02-14
Revert "Revert "Revert "Let allocations larger than 1M be handled directly by mmap"""

baldersheim:
Reverts vespa-engine/vespa#21188
@vekterli PR


vekterli (commented):
:+1:

PR #21194, opened 2022-02-15
Don't verify zipped tests for now

jonmv:
@hakonhall or @mpolden please review and merge.

This code path currently fails when the tests are in a zip, because it assumes that isn't a zip ... 🤦 
We either need to unpack the zip, and verify any tests in it, or verify prior to zipping.


mpolden (approved):

mpolden:
Build failed:
` 00:01:59 one or more files need to be formatted: try make fmt to fix this automatically `

jonmv:
Sorry.

PR #21198, opened 2022-02-15
Support all services types in the curl subcommand

ean:
Required authentication is chosen based on which service is accessed, and if
you have Auth0 or api-key available.

### DIFF
--- a/client/go/cmd/curl.go
+++ b/client/go/cmd/curl.go
@@ -2,20 +2,25 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/vespa-engine/vespa/client/go/auth0"
 	"github.com/vespa-engine/vespa/client/go/curl"
+	"github.com/vespa-engine/vespa/client/go/vespa"
 )
 
 var curlDryRun bool
+var curlService string
 
 func init() {
 	rootCmd.AddCommand(curlCmd)
 	curlCmd.Flags().BoolVarP(&curlDryRun, "dry-run", "n", false, "Print the curl command that would be executed")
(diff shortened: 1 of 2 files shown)


ean:
@mpolden - Please take a look at the latest diff. Based on our discussion yesterday I think this follow the simple wrapper idea you had for the curl subcommand.

@oyving - FYI, this should could you the the functionality you asked for.

client/go/cmd/curl.go
```diff
@@ -60,8 +57,36 @@ $ vespa curl -- -v --data-urlencode "yql=select * from music where album contain
 		if err != nil {
 			return err
 		}
-		c.PrivateKey = privateKeyFile
-		c.Certificate = certificateFile
+		switch curlService {
+		case "deploy":
+			if !vespa.Auth0AccessTokenEnabled() {
```
mpolden, line 62: A test should be added for this branch.

ean: Added a test for the "target is local" case, would have been nice to have something on the Auth0-part as well, but I see no way to do that without mocking away the entire Auth0-flow, which will make the test useless.

client/go/cmd/curl_test.go
```diff
@@ -18,4 +18,8 @@ func TestCurl(t *testing.T) {
 		filepath.Join(homeDir, "t1.a1.i1", "data-plane-private-key.pem"),
 		filepath.Join(homeDir, "t1.a1.i1", "data-plane-public-cert.pem"))
 	assert.Equal(t, expected, out)
+
+	out, _ = execute(command{homeDir: homeDir, args: []string{"curl", "-s", "deploy", "-n", "/application/v4/tenant/foo"}}, t, httpClient)
+	expected = fmt.Sprintf("curl https://127.0.0.1:19071/application/v4/tenant/foo\n")
```
mpolden, line 23: `fmt.Sprintf` is not needed?

mpolden (approved):

PR #21199, opened 2022-02-15
Adjust attribute-address-space resource limits.

geirst:
Feed blocked is now always decided in the cluster controller,
and the new limit of 90% matches what was used in proton previously.

@baldersheim please review

### DIFF
--- a/searchcore/src/vespa/searchcore/config/proton.def
+++ b/searchcore/src/vespa/searchcore/config/proton.def
@@ -446,7 +446,7 @@ initialize.threads int default = 0
 
 ## Portion of max address space used in components in attribute vectors
 ## before put and update operations in feed is blocked.
-writefilter.attribute.address_space_limit double default = 0.9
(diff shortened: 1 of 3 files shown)


baldersheim (approved):

searchcore/src/vespa/searchcore/config/proton.def
```diff
@@ -446,7 +446,7 @@ initialize.threads int default = 0
 
 ## Portion of max address space used in components in attribute vectors
 ## before put and update operations in feed is blocked.
-writefilter.attribute.address_space_limit double default = 0.9
+writefilter.attribute.address_space_limit double default = 0.92
```
baldersheim, line 449: Intentional to change address_space_limit to 0.92 ?

geirst: Yes. The fail safe limit in proton should hit later than the limit used by the cluster controller.

PR #21202, opened 2022-02-15
Allow inhibiting all default bucket space merges if global merges pending [run-systemtest]

vekterli:
@geirst please review

If the cluster controller publishes a cluster state where at least
one node is marked as maintenance in the default space but is up
in the global space, global bucket merges are pending in the cluster.
We want these merges to complete ASAP, and to avoid starvation caused
by distributors that finish their global merges quicker than others
(or whose global bucket ops get preempted), allow for _completely_
inhibiting merges to buckets in the default space when global buckets
are pending.

Requires a live config flag to be set to actually be taken into account
in the bucket synchronization state checker.

Flag is tracked per distributor bucket space (per stripe) and is
automatically cleared once a cluster state is received without pending
global merges.


geirst (approved):
👍

PR #21210, opened 2022-02-15
Refactor Vespa CLI error handling

mpolden:
Before this we littered `fatalErr...` everywhere which meant that the control
flow when running the code directly differed from running in unit tests.

This change passes errors up the call stack by having all commands use `RunE`
instead of `Run`. Printing of errors and exiting with appropriate status can
then happen in a single place.

Errors are also wrapped with `%w` so that error printing can unwrap the most
interesting error if necessary.

@jonmv


jonmv (commented):
This is great, thank you!

PR #21217, opened 2022-02-16
Let number of warmup threads be a factor of cores, insteda of the fix…

baldersheim:
…ed number 4.

num_cores/8 but capped in the range [1, 4].

@geirst PR


geirst (approved):
👍

PR #21218, opened 2022-02-16
If threads are configured explicit to a positive number, obey it.

baldersheim:
If not scale it with num_cores/8 capped i the range [1, 4].
Let default configured value be 0 (unset) instead of 4.

@geirst PR


geirst (approved):
👍

PR #21219, opened 2022-02-16
Tag cpu category on service layer threads

geirst:
@vekterli please review
@baldersheim @havardpe FYI


havardpe (commented):
I think std::nullopt is better than {} for empty optionals

havardpe:
An alternative to optional cpu usage would be to use OTHER as the fallback. This will have a small additional overhead only in the cases where no cpu usage is ever tagged in the thread (~10ns per 5s, paid by the metrics hook thread).

geirst:
> I think std::nullopt is better than {} for empty optionals

Fixed.

vekterli (approved):
:+1: TIL `std::nullopt` is a thing

PR #21225, opened 2022-02-16
Add vespalib utility functions for atomic memory access

vekterli:
@havardpe and @toregge please review
@toregge feel free to add Clang on Darwin intrinsics if so desired 🦄 

Adds the following utilities:

* Atomic reference wrapper functions for accessing single memory
  locations as if they were `std::atomic` instances.
* Wrappers for less verbose `std::atomic` loads/stores that also
  sanity-check that accesses are always lock-free.

### DIFF
--- a/vespalib/src/vespa/vespalib/util/atomic.h
+++ b/vespalib/src/vespa/vespalib/util/atomic.h
@@ -100,7 +100,7 @@ template <typename T>
     static_assert(!detail::is_std_atomic_v<T>, "atomic ref function invoked with a std::atomic, probably not intended");
 #if __cpp_lib_atomic_ref
     static_assert(std::atomic_ref<const T>::is_always_lock_free);
-    return std::atomic_ref<const T>(a).load(std::memory_order_acquire);
+    return std::atomic_ref<const T>(a).load(std::memory_order_seq_cst);
 #else
     // TODO replace with compiler intrinsic
     std::atomic_thread_fence(std::memory_order_seq_cst);
--- a/vespalib/src/vespa/vespalib/util/atomic.h
+++ b/vespalib/src/vespa/vespalib/util/atomic.h
@@ -1,0 +1,155 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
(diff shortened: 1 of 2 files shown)


vespalib/src/vespa/vespalib/util/atomic.h
```diff
@@ -0,0 +1,155 @@
...
+
+namespace vespalib::atomic {
+
+//
+// std::atomic_ref<T> helpers
+//
+
+// No atomic_ref on clang on darwin (for now)
+#if defined(__clang__) && defined(__apple_build_version__)
+#  define VESPA_NO_ATOMIC_REF_SUPPORT
```
toregge, line 28: Consider `#include <version>` earlier in the file, and use `#if __cpp_lib_atomic_ref` instead of `#ifndef VESPA_NO_ATOMIC_REF_SUPPORT`

toregge (approved):

vespalib/src/vespa/vespalib/util/atomic.h
```diff
@@ -0,0 +1,155 @@
...
+    return a;
+#endif
+}
+
+template <typename T>
+[[nodiscard]] constexpr T load_ref_seq_cst(const T& a) noexcept {
+    static_assert(!detail::is_std_atomic_v<T>, "atomic ref function invoked with a std::atomic, probably not intended");
+#ifndef VESPA_NO_ATOMIC_REF_SUPPORT
+    static_assert(std::atomic_ref<const T>::is_always_lock_free);
+    return std::atomic_ref<const T>(a).load(std::memory_order_acquire);
```
havardpe, line 107: std::memory_order_seq_cst

vespalib/src/vespa/vespalib/util/atomic.h
vekterli: Neat! That's definitely better, will replace

vespalib/src/vespa/vespalib/util/atomic.h
vekterli: Good catch, fixed

toregge (approved):

havardpe (approved):

PR #21228, opened 2022-02-16
Resolve slice dimension

bratseth:
Proton requires slice on verbose form, e.g: myTensor{dimension:label} not myTensor{label}

To transform to that we need to resolve the tensor type from where this is serialized.
We're missing a bit of type resolving infrastructure to do that, so this implements it for the simple cases and outputs an easy-to-understand error message when not.


lesters (approved):
LGTM! 💯

PR #21232, opened 2022-02-16
Make CountDownLatch debug function thread safe

vekterli:
@baldersheim please review

### DIFF
--- a/vespalib/src/vespa/vespalib/util/count_down_latch.h
+++ b/vespalib/src/vespa/vespalib/util/count_down_latch.h
@@ -22,7 +22,7 @@ namespace vespalib {
 class CountDownLatch
 {
 private:
-    std::mutex              _lock;
+    mutable std::mutex      _lock;
     std::condition_variable _cond;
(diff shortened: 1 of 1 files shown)


vespalib/src/vespa/vespalib/util/count_down_latch.h
```diff
@@ -82,7 +82,10 @@ class CountDownLatch
      *
      * @return current count
      **/
-    uint32_t getCount() const { return _count; }
+    [[nodiscard]] uint32_t getCount() const noexcept {
+        std::unique_lock guard(_lock);
```
baldersheim, line 86: Enough to use a lock_guard ?

baldersheim (approved):

vespalib/src/vespa/vespalib/util/count_down_latch.h
vekterli: Yes, force of habit from spending too much time around condition variables 🙂  Will simplify

PR #21242, opened 2022-02-17
Remove unused CommonFields and QueryTreeUtil

hakonhall:


bratseth:
Technically we can remove these now, but we have waited 10-15 years, why not wait a few months longer?

bratseth (approved):

hakonhall:
These changes should be moved to the Vespa 8 branch once it has been created.

baldersheim:
These seems to have been applied already.

PR #21245, opened 2022-02-17
Handle different page size

hmusum:
Some ARM HW has 64K page size, try to update tests to handle this

### DIFF
--- a/vespalib/src/tests/array/array_test.cpp
+++ b/vespalib/src/tests/array/array_test.cpp
@@ -331,29 +331,33 @@ TEST("test move assignment")
 
 struct UnreserveFixture {
     Array<int> arr;
-    UnreserveFixture() : arr(1025, 7, alloc::Alloc::allocMMap(0))
+    UnreserveFixture() : arr(page_ints() + 1, 7, alloc::Alloc::allocMMap(0))
     {
-        EXPECT_EQUAL(1025u, arr.size());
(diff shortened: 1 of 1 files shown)


hmusum:
@toregge please take a look

vespalib/src/tests/array/array_test.cpp
```diff
@@ -333,26 +334,26 @@ struct UnreserveFixture {
...
+    EXPECT_FALSE(f.arr.try_unreserve(1_Ki));
 }
 
 TEST_F("require that try_unreserve() succeedes if mmap can be shrinked", UnreserveFixture)
 {
     int *oldPtr = &f.arr[0];
     f.arr.resize(512);
     EXPECT_TRUE(f.arr.try_unreserve(1023));
-    EXPECT_EQUAL(1_Ki, f.arr.capacity());
+    EXPECT_EQUAL(round_up_to_page_size(1) == 64_Ki ? 16_Ki : 1_Ki, f.arr.capacity());
```
toregge, line 356: Original test was close to the border, but is one lower to test that new capacity is larger than requested capacity. `page_ints() - 1` is probably a better argument to `try_unreserve()`, which results in the expected capacity being `page_ints()`

toregge (approved):

toregge (approved):

hmusum:
Thanks for the detailed feedback!

PR #21266, opened 2022-02-17
Add a base vespa timer frequency that is used as a base to control po…

baldersheim:
…ll frequency of various tasks.

@vekterli PR
Replaces #21260


vekterli (approved):
:+1:

PR #21272, opened 2022-02-18
Report current docker image & vespa version for coredump

freva:
Current version in node-repo may be wrong if core is dumped during upgrade. Only report the docker image of the currently running container and use the version from that docker image tag.

Fixes VESPA-23022.


hakonhall (approved):
We should consider storing to disk, what containers were started at what times and version.  It's typically there in the vespa.log, but it may be rotated and reading its own vespa log is not ideal.

freva:
Yes, that's probably the way to go if the container is already gone often enough when we report the cores.

PR #21277, opened 2022-02-18
Use EXPECT_NEAR when comparing

hmusum:
EXPECT_DOUBLE_EQ fails on arm64 for some reason

### DIFF
--- a/searchlib/src/tests/common/geogcd/geo_gcd_test.cpp
+++ b/searchlib/src/tests/common/geogcd/geo_gcd_test.cpp
@@ -49,7 +49,7 @@ TEST(GeoGcdTest, computed_distances_seem_legit) {
             double miles = km / 1.609344;
             EXPECT_GE(miles, 0);
             if (from.name == to.name) {
-                EXPECT_DOUBLE_EQ(miles, 0.0);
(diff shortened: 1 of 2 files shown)


hmusum:
or @toregge or @baldersheim

searchlib/src/tests/common/geogcd/geo_gcd_test.cpp
```diff
@@ -49,7 +49,7 @@ TEST(GeoGcdTest, computed_distances_seem_legit) {
             double miles = km / 1.609344;
             EXPECT_GE(miles, 0);
             if (from.name == to.name) {
-                EXPECT_DOUBLE_EQ(miles, 0.0);
+                EXPECT_NEAR(miles, 0.0, pow(10, -9)); // EXPECT_DOUBLE_EQ does not work on arm64 for some reason
```
toregge, line 52: Consider using just `1e-9` instead of `pow(10, -9)`.

toregge (approved):

searchlib/src/tests/common/geogcd/geo_gcd_test.cpp
hmusum: 👍

toregge (approved):

PR #21282, opened 2022-02-20
Make ConfigUri constructors explicit and use same context where possi…

baldersheim:
…ble in proton.

@vekterli PR

### DIFF
--- a/messagebus_test/src/tests/trace/cpp-server.cpp
+++ b/messagebus_test/src/tests/trace/cpp-server.cpp
@@ -41,7 +41,7 @@ Server::handleMessage(Message::UP msg) {
     msg->getTrace().trace(1, _name + " (message)", false);
     if (!msg->getRoute().hasHops()) {
         fprintf(stderr, "**** Server '%s' replying.\n", _name.c_str());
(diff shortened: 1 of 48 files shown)


messagebus_test/src/tests/trace/cpp-server.cpp
```diff
@@ -70,9 +70,9 @@ App::Main()
         return 1;
     }
     RPCMessageBus mb(ProtocolSet().add(std::make_shared<SimpleProtocol>()),
-                     RPCNetworkParams("file:slobrok.cfg")
+                     RPCNetworkParams(config::ConfigUri("file:slobrok.cfg"))
                      .setIdentity(Identity(_argv[1])),
-                     "file:routing.cfg");
+    config::ConfigUri("file:routing.cfg"));
```
vekterli, line 75: Consider reindenting

vekterli (approved):
:+1:

messagebus_test/src/tests/trace/cpp-server.cpp
baldersheim: Fixed

PR #21287, opened 2022-02-21
Retry deprovisioning if child is un-retired

mpolden:
@freva


freva (approved):
Host-admin tick is 30 sec.
`DynamicProvisioningMaintainer` interval is 5min, spread evenly across 3 configservers so that it runs every 5min/3 = 1m40sec.

Not the best odds, but I guess it'll converge eventually...

mpolden:
PTAL at last commit.

freva (approved):

PR #21291, opened 2022-02-21
Ean/make sample apps download atomic

ean:

### DIFF
--- a/client/go/cmd/clone.go
+++ b/client/go/cmd/clone.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net/http"
 	"os"
@@ -20,7 +21,10 @@ import (
 	"github.com/vespa-engine/vespa/client/go/util"
 )
 
-const sampleAppsCacheTTL = time.Hour * 168 // 1 week
+const (
+	sampleAppsCacheTTL = time.Hour * 168 // 1 week
+	sampleAppsFilename = "sample-apps-master.zip"
+)
 
 var listApps bool
 var forceClone bool
@@ -65,7 +69,7 @@ variable.`,
 }
 
 func cloneApplication(applicationName string, applicationDir string) error {
-	zipFile, err := getSampleAppsZip()
+	zipFile, err := openSampleAppsZip()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


client/go/cmd/clone.go
```diff
@@ -148,17 +135,50 @@ func getSampleAppsZip() (*os.File, error) {
...
+		if response.StatusCode != http.StatusOK {
 			return fmt.Errorf("could not download sample apps: github returned status %d", response.StatusCode)
 		}
-		if err := f.Truncate(0); err != nil {
-			return fmt.Errorf("could not truncate sample apps file: %s: %w", f.Name(), err)
-		}
 		if _, err := io.Copy(f, response.Body); err != nil {
 			return fmt.Errorf("could not write sample apps to file: %s: %w", f.Name(), err)
 		}
+		if err := os.Rename(f.Name(), destination); err != nil {
```
mpolden, line 144: `f` needs to be closed before moving (Linux/macOS is fine with this, but Windows isn't).

client/go/cmd/clone.go
```diff
@@ -148,17 +135,50 @@ func getSampleAppsZip() (*os.File, error) {
...
+		if err != nil {
+			return nil, errHint(fmt.Errorf("could not determine cache status: %w", err), "Try ignoring the cache with the -f flag")
+		}
+		if useCache {
+			log.Print(color.Yellow("Using cached sample apps ..."))
+			return f, nil
+		}
+		f.Close()
+	}
+	f, err = os.OpenFile(path, os.O_RDWR|os.O_CREATE, 0755)
```
mpolden, line 175: Instead of doing this, the method should simply return `os.Open(path)` at the end of this function (after the call to `fetchSampleAppsZip`)

client/go/cmd/clone.go
```diff
@@ -148,17 +135,50 @@ func getSampleAppsZip() (*os.File, error) {
...
 			return fmt.Errorf("could not write sample apps to file: %s: %w", f.Name(), err)
 		}
+		if err := os.Rename(f.Name(), destination); err != nil {
+			return fmt.Errorf("could not move sample apps to cache path")
+		}
 		return nil
 	})
+}
+
+func getSampleAppsZip() (*os.File, error) {
```
mpolden, line 151: Consider naming this `openSampleAppsZip` as it returns an open file handle.

client/go/cmd/clone.go
```diff
@@ -125,20 +120,12 @@ func useCache(cacheFile *os.File) (bool, error) {
 	return stat.Size() > 0 && time.Now().Before(expiry), nil
 }
 
-func getSampleAppsZip() (*os.File, error) {
-	f, err := openOutputFile()
+func fetchSampleAppsZip(cacheDir string, destination string) error {
+	f, err := ioutil.TempFile(cacheDir, "sample-apps")
```
mpolden, line 124: Instead of passing `cacheDir` as a separate argument this should just do `filepath.Dir(destination)` so that the temp file directory is always the parent of `destination`.

client/go/cmd/command_tester.go
```diff
@@ -76,7 +78,10 @@ func execute(cmd command, t *testing.T, client *mockHttpClient) (string, string)
 
 	// Execute command and return output
 	rootCmd.SetArgs(append(cmd.args, cmd.moreArgs...))
-	Execute()
+	err := Execute()
+	if cmd.failTestOnError {
+		require.Nil(t, err)
```
mpolden, line 83: Shouldn't err be non-nil? `assert.NotNil(t, err)`?

client/go/cmd/clone.go
ean: ... that was the clean solution I missed. Thanks.

client/go/cmd/command_tester.go
ean: The intent here is that you ca set failTestOnError to true if you want the test to fail if execute() returns an error.

The best solution here is probably to invert it and create a `failTestOnSuccess` so that tests that should fail on execution was explicitly marked as such, but there were too many cases of that to add that to this PR.

client/go/cmd/clone.go
ean: Simplified all of this, please take another look.

mpolden (approved):

PR #21299, opened 2022-02-21
Do not roll out new application package with pure deployment orchestr…

jonmv:
…ation changes

@mpolden please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/Application.java
```diff
@@ -118,6 +121,20 @@ public Application(TenantAndApplicationId id, Instant createdAt, DeploymentSpec
         return versions;
     }
 
+    /** Returns the currently deployed versions of the application */
```
jonmv, line 124:
```suggestion
    /** Returns all current application version which change deployments, relative to the previous one(s) */
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/Application.java
```diff
@@ -118,6 +121,20 @@ public Application(TenantAndApplicationId id, Instant createdAt, DeploymentSpec
...
+    public Collection<ApplicationVersion> deployableVersions(boolean ascending) {
+        Deque<ApplicationVersion> versions = new ArrayDeque<>();
+        String previousHash = "";
+        for (ApplicationVersion version : versions()) {
+            if (previousHash.equals(previousHash = version.bundleHash().orElse("")) && version.bundleHash().isPresent())
```
mpolden, line 129: Consider putting assignment on separate line. I don't think I've ever seen an equals call that also assigns a variable. 😆

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
```diff
@@ -254,14 +260,29 @@ public DeploymentContext submit(ApplicationPackage applicationPackage) {
...
+        return submit(applicationPackage, sourceRevision, salt.getAndIncrement());
+    }
+
+    /** Submit given application package for deployment */
+    public DeploymentContext submit(ApplicationPackage applicationPackage, Optional<SourceRevision> sourceRevision, long salt) {
+        var buffer = new ByteArrayOutputStream();
+        ZipStreamReader.transferAndWrite(buffer,
+                                         new ByteArrayInputStream(applicationPackage.zippedContent()),
+                                         "salt",
+                                         new byte[]{ (byte) (salt >> 56), (byte) (salt >> 48), (byte) (salt >> 40), (byte) (salt >> 32), (byte) (salt >> 24), (byte) (salt >> 16), (byte) (salt >> 8), (byte) salt });
```
mpolden, line 279: 🙂

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/Application.java
jonmv: I've seen it multiple places in the standard library 😬 But sure. It's just what felt natural writing. Maybe not the easiest thing to parse. though ... 😀

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentContext.java
jonmv: 🙈

mpolden (approved):

PR #21301, opened 2022-02-22
Move away from deprecated 'allocated-memory' attribute in nodes element

hmusum:
@gjoranv and/or @jonmv please review


jonmv (approved):

jonmv:
Thanks :)

PR #21304, opened 2022-02-22
Remove TODO on changing toString to YQL format

bratseth:
It would be nice but it breaks 842 tests for us,
and probably a lot for applications too, so it is just
not worth the effort.


arnej27959 (approved):
👍

PR #21309, opened 2022-02-22
Add more strict validation for node counts

ean:
Disallow count values that are zero, negative and ranges where max is less than
min.


mpolden (approved):
👍 
This will still allow 1 node, but I think it's better for the remote side to validate that.

PR #21311, opened 2022-02-22
Revert "Use common tranport for TlsClient"

hmusum:
Reverts vespa-engine/vespa#21293

Seems to have broken some unit tests


geirst (approved):
👍

PR #21312, opened 2022-02-22
Expose current active operation throttler token count

vekterli:
@geirst please review

Allows for inspecting _actual_ operation parallelism vs. the _maximum_
parallelism given by the internal policy window size. Remove the note
stating that waiting thread count is for testing only, as we want to be
able to expose that as well.

Available for both the dynamic and the unlimited implementations, as they
both explicitly track active token counts.


geirst (approved):
👍

PR #21316, opened 2022-02-22
Add API to download submitted tests

freva:


jonmv (approved):
You assume the two sets of versions are in sync, but we don't really enforce this. It's currently true, or at least almost always. I guess this is OK. I think perhaps we _should_ enforce it, too ...

PR #21318, opened 2022-02-22
Add support for creating ConfigContext with externally provided FNET_…

baldersheim:
…Transport object

@vekterli PR


vekterli (approved):
:+1:

PR #21323, opened 2022-02-22
Arnej/intermediate parser 2

arnej27959:

@bratseth and/or @baldersheim please review


bratseth (approved):
👍 

You should add @author tags and some class level javadoc which explains what this does.

I guess we should also have a big note in the current parser saying that new things must be added two places for now.

bratseth:
For things that are immutable we usually skip the "get" (as in "getName") as there will never be a "set".

baldersheim (approved):

baldersheim (approved):

PR #21331, opened 2022-02-23
Replace 'out of capacity' by 'node allocation failure'

bratseth:
This might lead to some temporary glitches during infrastructure upgrade, but I don't think it's worth doing something more complicated.


config-model/src/test/java/com/yahoo/vespa/model/test/VespaModelTester.java
```diff
@@ -109,7 +109,7 @@ public void useDedicatedNodeForLogserver(boolean useDedicatedNodeForLogserver) {
         this.useDedicatedNodeForLogserver = useDedicatedNodeForLogserver;
     }
 
-    /** Creates a model which uses 0 as start index and fails on out of capacity */
+    /** Creates a model which uses 0 as start index and fails on node allocation */
```
jonmv, line 112: I hope it doesn't _always_ fail on node allocation? 😜

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
```diff
@@ -377,7 +377,8 @@ static String valueOf(RunStatus status) {
     static RunStatus runStatusOf(String status) {
         switch (status) {
             case "running"                    : return running;
-            case "outOfCapacity"              : return outOfCapacity;
+            case "outOfCapacity"              : return nodeAllocationFailure;  // TODO: Remove after March 2022
```
jonmv, line 380: We don't re-write historic runs, so this needs to stay, unless we start re-writing all of those regularly as well.

bratseth: When do they expire?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/InternalStepRunner.java
```diff
@@ -812,7 +812,7 @@ private void updateConsoleNotification(Run run) {
             case success:
                 controller.notificationsDb().removeNotification(source, Notification.Type.deployment);
                 return;
-            case outOfCapacity:
+            case nodeAllocationFailure:
                 if ( ! run.id().type().environment().isTest()) updater.accept("lack of capacity. Please contact the Vespa team to request more!");
```
jonmv, line 816: This should dalso be updated, like the mail notification.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/RunSerializer.java
jonmv: When they're more than 65 runs old, or somehting, and they're also no longer the "first failing" or "lasts successful" run. Potentially never. I guess we should re-write these runs ...

jonmv: I can do that in a differnet PR.

jonmv: Never mind I'm just bad at reading code. We update these as well.

bratseth: Thanks. I'll remove the date for now.

jonmv (approved):

hmusum:
Users of `OutOfCapacityException` in internal repo also need to be updated (build is broken now)

PR #21335, opened 2022-02-23
Ean/device flow as default

ean:
@mpolden - please take a look at this

### DIFF
--- a/client/go/auth0/auth0.go
+++ b/client/go/auth0/auth0.go
@@ -234,6 +234,16 @@ func (a *Auth0) getSystem() (*System, error) {
 	return s, nil
 }
 
+// HasSystem checks if the system is configured
+// TODO: Used to print deprecation warning if we fall back to use tenant API key.
+//       Remove when this is not longer needed.
+func (a *Auth0) HasSystem() bool {
+	if _, err := a.getSystem(); err != nil {
+		return false
+	}
+	return true
+}
+
 // AddSystem assigns an existing, or new System. This is expected to be called
 // after a login has completed.
 func (a *Auth0) AddSystem(s *System) error {
--- a/client/go/auth0/auth0.go
+++ b/client/go/auth0/auth0.go
@@ -129,6 +129,7 @@ func (a *Auth0) IsLoggedIn() bool {
 	}
 
 	// Check if token is valid.
+	// TODO: Choose issuer based on system
 	if err = jwt.Validate(token, jwt.WithIssuer("https://vespa-cd.auth0.com/")); err != nil {
 		return false
 	}
--- a/client/go/cmd/config.go
+++ b/client/go/cmd/config.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"github.com/vespa-engine/vespa/client/go/auth0"
 	"github.com/vespa-engine/vespa/client/go/util"
 	"github.com/vespa-engine/vespa/client/go/vespa"
 )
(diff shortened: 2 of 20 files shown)


client/go/auth0/auth0.go
```diff
@@ -129,6 +129,7 @@ func (a *Auth0) IsLoggedIn() bool {
 	}
 
 	// Check if token is valid.
+	// TODO: Choose issuer based on system
```
bjorncs, line 132: This can be deduced Auth0 device flow config from api endpoint (https://api.vespa-external-cd.aws.oath.cloud:4443/auth0/v1/device-flow-config).

bjorncs (approved):

mpolden (approved):

client/go/cmd/config.go
```diff
@@ -179,19 +183,41 @@ func (c *Config) X509KeyPair(app vespa.ApplicationID) (KeyPair, error) {
...
+		return true
+	}
+
+	// If no Auth0 token is created, fall back to tenant api key, but warn that this functionality is deprecated
+	a, err := auth0.GetAuth0(c.AuthConfigPath(), getSystemName(), getApiURL())
+	if err != nil || !a.HasSystem() {
+		if _, err := os.Stat(c.APIKeyPath(tenantName)); err != nil {
+			return false
+		}
+		fmt.Fprintln(os.Stderr, "Defaulting to tenant API key is deprecated. Use Auth0 device flow: 'vespa auth login' instead")
```
mpolden, line 214: Should print to the variable `stderr`.

client/go/cmd/config.go
```diff
@@ -179,19 +183,41 @@ func (c *Config) X509KeyPair(app vespa.ApplicationID) (KeyPair, error) {
...
+	if _, err := c.Get(apiKeyFlag); err == nil {
+		return true
+	}
+	if _, err := c.Get(apiKeyFileFlag); err == nil {
+		return true
+	}
+
+	// If no Auth0 token is created, fall back to tenant api key, but warn that this functionality is deprecated
+	a, err := auth0.GetAuth0(c.AuthConfigPath(), getSystemName(), getApiURL())
+	if err != nil || !a.HasSystem() {
```
mpolden, line 210: Why do we need all these checks? Shouldn't we simply use the API key if it exists and print a warning that the user should setup Auth0 with `vespa auth login`?

client/go/cmd/config.go
```diff
@@ -179,19 +183,41 @@ func (c *Config) X509KeyPair(app vespa.ApplicationID) (KeyPair, error) {
...
+		if _, err := os.Stat(c.APIKeyPath(tenantName)); err != nil {
```
mpolden, line 211: Consider using `util.PathExists`.

client/go/cmd/helpers.go
```diff
@@ -270,19 +250,11 @@ func getDeploymentOpts(cfg *Config, pkg vespa.ApplicationPackage, target vespa.T
...
+			hint := "Try 'vespa auth cert'"
 			return vespa.DeploymentOpts{}, errHint(fmt.Errorf("missing certificate in application package"), "Applications in Vespa Cloud require a certificate", hint)
 		}
-		opts.APIKey, err = cfg.ReadAPIKey(deployment.Application.Tenant)
-		if !vespa.Auth0AccessTokenEnabled() {
-			if err != nil {
-				return vespa.DeploymentOpts{}, errHint(err, "Deployment to cloud requires an API key. Try 'vespa api-key'")
-			}
+		if cfg.UseAPIKey(deployment.Application.Tenant) {
+			opts.APIKey, err = cfg.ReadAPIKey(deployment.Application.Tenant)
```
mpolden, line 257: This error should be checked.

client/go/cmd/config.go
ean: This was never meant as a permanent solution, more of a way to support the old behaviour for some releases to nudge people to the device flow. It will give people some time to migrate over, but I am happy to rip it all out if we think this functionality is not part of any automated systems somewhere.

PR #21348, opened 2022-02-23
Application version JSON serialization

freva:
Wanted to expose `compileVersion` in `builds`, but then I saw that we have 3 ways of serializing `ApplicationVersion`. This PR changes it to 1 way. 

My understand is that we want to get rid of `SourceRevision` in favor of `sourceUrl`, with this PR the only users of `ApplicationVersion::source` are tests and ZK serialization code.

I've verified that `source` object, `hash` and `buildNumber` are not used anywhere except in 1 internal documentation page, which I'll update.


jonmv:
Are there conflicts on this branch now?

jonmv:
Otherwises LGTM!

freva:
All tests still pass on my machine rebased on master at least.

jonmv (approved):

PR #21349, opened 2022-02-23
Add more feature flags for tuning dynamic persistence throttling [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Adds the following feature flags:
 * `persistence-throttling-window-size` for setting min/max window size
   to the same value, effectively creating a static throttle policy.
 * `persistence-throttling-ws-resize-rate` for controlling the rate
   at which the dynamic throttle policy resizes its window.


geirst (approved):
👍

PR #21350, opened 2022-02-23
Ensure that internal state of ExternSlobrokPolicy and ExternPolicy is…

baldersheim:
… consistent with respect to transport, supervisor, and mirror.

@vekterli PR

PR #21351, opened 2022-02-23
In order to properly detach supervisor from the transport, do an asyn…

baldersheim:
…c close and proper sync of trasnport threads.

@havardpe PR
Must go in after #21350


fnet/src/vespa/fnet/frt/supervisor.cpp
```diff
@@ -27,8 +27,9 @@ FRT_Supervisor::FRT_Supervisor(FNET_Transport *transport)
 FRT_Supervisor::~FRT_Supervisor()
 {
     if (_connector != nullptr) {
-        _connector->SubRef();
+        _connector->Owner()->Close(_connector, /* needref */ false);
     }
+    _transport->sync();
```
havardpe, line 32: We could consider having an explicit shutdown sequence to avoid deadlocks for non-activated objects. Let us discuss today.

PR #21352, opened 2022-02-23
GC unused SessionHandler

baldersheim:
@havardpe PR


jrt/tests/com/yahoo/jrt/ConnectTest.java
```diff
@@ -11,33 +12,23 @@ public void testConnect() throws ListenFailedException {
...
         Acceptor acceptor = server.listen(new Spec(0));
 
-        assertTrue(server.checkLifeCounts(0, 0));
-        assertTrue(client.checkLifeCounts(0, 0));
-
         Target target = client.connect(new Spec("localhost", acceptor.port()));
 
         for (int i = 0; i < 100; i++) {
-            if (client.initCount == 1 && server.initCount == 1) {
+            if (target.isValid()) {
```
havardpe, line 18: A connection is valid until it fails (you can start using it while it connects). What we would like to test here is that the connection transitions into the CONNECTED state. We could consider adding an 'isConnected' function that checks for CONNECTED state while holding the lock (side note: isValid should also take the lock). Being able to poll and wait for a connections to become connected could be useful in other contexts (where batchInvoke was originally used, to separate tls handshake time from RPC timeout).

havardpe (commented):
we are no longer able to detect that we are able to connect... (see comment)

jrt/tests/com/yahoo/jrt/ConnectTest.java
baldersheim: I added a synchronized isConnected() method on the connection.
In addition I used the Connection directly to avoid adding the interface to the Target, and used isConnected/isClosed instead of isValid.

Both isValid and isClosed should have been synchronized, but they are used intensively in frequently called methods.
I guess the state should either have been volatile instead, or an Atomic. But that is a separate issue for later on.

PR #21353, opened 2022-02-23
Add grammar:none

bratseth:
Provide a way to handle text with various characters which was not written with any query syntax in mind.


arnej27959 (approved):
looks useful to me

PR #21363, opened 2022-02-24
Allow no tests in staging setup

tokle:


aressem (approved):

jonmv:
So long-term I think we should require staging-setup, but we can allow not having it for the oldest test framework. The basic HTTP tests already enforce it. (Well, _now_ they don't, but they always have until this.) 
Also, staging tests without the setup step make almost no sense.

PR #21365, opened 2022-02-24
Allow no sstaging setup tests

jonmv:
@aressem or @tokle


tokle:
We are fixing this in #21363.

PR #21369, opened 2022-02-24
- Create the common transport and threadpool in the main loop.

baldersheim:
- Also use the common transport for config subscriptions.
- Put The TransportConfig in the fnet namespace.

@havardpe PR
@geirst @vekterli FYI

### DIFF
--- a/searchcore/src/apps/proton/proton.cpp
+++ b/searchcore/src/apps/proton/proton.cpp
@@ -5,9 +5,13 @@
 #include <vespa/metrics/metricmanager.h>
 #include <vespa/vespalib/util/signalhandler.h>
 #include <vespa/vespalib/util/programoptions.h>
+#include <vespa/vespalib/util/size_literals.h>
 #include <vespa/vespalib/io/fileutil.h>
 #include <vespa/vespalib/util/exceptions.h>
 #include <vespa/config/common/exceptions.h>
+#include <vespa/config/common/configcontext.h>
+#include <vespa/fnet/transport.h>
+#include <vespa/fastos/thread.h>
 #include <vespa/fastos/app.h>
 #include <iostream>
 #include <thread>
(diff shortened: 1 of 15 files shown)


searchcore/src/apps/proton/proton.cpp
```diff
@@ -186,8 +190,14 @@ App::Main()
         LOG(debug, "serviceidentity: '%s'", params.serviceidentity.c_str());
         LOG(debug, "subscribeTimeout: '%" PRIu64 "'", params.subscribeTimeout);
         std::chrono::milliseconds subscribeTimeout(params.subscribeTimeout);
-        config::ConfigUri identityUri(params.identity);
-        protonUP = std::make_unique<proton::Proton>(identityUri, _argc > 0 ? _argv[0] : "proton", subscribeTimeout);
+        FastOS_ThreadPool threadPool(128_Ki);
+        uint32_t numProcs = std::thread::hardware_concurrency();
+        FNET_Transport transport(fnet::TransportConfig(std::max(1u, std::min(4u, numProcs/8))));
```
geirst, line 195: Consider adding a helper function for calculating TransportConfig.

searchcore/src/apps/proton/proton.cpp
```diff
@@ -228,6 +238,7 @@ App::Main()
                 EV_STOPPING("servicelayer", "clean shutdown");
             }
             protonUP.reset();
+            transport.ShutDown(true);
```
havardpe, line 241: consider shutting down transport before deleting proton, to reduce chance of core dumps

geirst (commented):
The changes in proton looks good. @havardpe should look at the other parts.

havardpe (commented):
deleting stuff using the shared transport before shutting down the transport may lead to unfortunate events.

searchcore/src/apps/proton/proton.cpp
baldersheim: Xixed

PR #21375, opened 2022-02-24
We do not need to maintain a separate executor just for handling rand…

baldersheim:
…om manual getProtonStatus rpc calls. Just use the one that proton owns.

@geirst PR


geirst (approved):
👍

PR #21376, opened 2022-02-24
Make it possible to configure what aspects of merges should be throttled [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Add live config for choosing whether merges should be throttled
on a per-feed operation (`MergeHandler`) level, or on an `ApplyBucketDiff`
persistence thread level.

This is intended to be a temporary feature while we do experiments, so
some liberties are taken with regards to how holes are punched in the
various abstraction layers.


geirst (approved):
👍

PR #21380, opened 2022-02-24
Allow multiple settings if they are equal

bratseth:


arnej27959 (approved):
This seems to be getting complicated quickly, and I'm afraid of feature creep.
As a user I would expect that if multiple inheritance works,  I can also inherit from one rank-profile setting rank-properties for fieldMatch(title) and another setting properties for nativeRank.fieldMatchWeight (just as an example).  Or inheriting multiple summary-features list to get the union of them.  Maybe we can do more resolving while setting up the RankProfile objects?
My related WIP for what is actually user-configurable is in

https://github.com/vespa-engine/vespa/blob/arnej/intermediate-parser-5/config-model/src/main/java/com/yahoo/searchdefinition/parser/ParsedRankProfile.java

bratseth:
Yes, there could be feature creep here, however the reason for this one is that some settings on rank profiles (at least filters) are propagated from the schema to all profiles, with the result that any such settings prevents multiple inheritance completely.

We'll have more practical options available when you are done with your work I think, but I'll merge this for now since it fixes a bug.

PR #21384, opened 2022-02-24
Revert "Revert "- Create the common transport and threadpool in the main loop.""

baldersheim:
Reverts vespa-engine/vespa#21383


baldersheim:
@havardpe @geirst PR
Only the last PR matters.

havardpe (approved):

PR #21387, opened 2022-02-25
Ean/cli test cleanup

ean:


mpolden (approved):
👍 

At some point I want to move each command to a separate package so this becomes less of a problem.

PR #21388, opened 2022-02-25
Allow `vespa auth cert` to work without application package

ean:
Let a users prepare certificates before creating their application package. If
an application package is not found, the user will be warned that the
certificate must be added to the application package at a later time using `vespa auth cert add`.

### DIFF
--- a/client/go/cmd/cert.go
+++ b/client/go/cmd/cert.go
@@ -4,7 +4,9 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


mpolden:
I think we should preserve `vespa cert` as it works today because that is the most common use case. To support the other use case we can simply add a flag, e.g. `--no-application-package` and skip copying if that flag is given.

ean:
I got a similar idea during lunch, but using a subcommand instead of argument.

`vespa auth cert --no-package` vs `vespa auth cert generate`

In conjunction with`vespa auth cert add` I think `generate` makes more sense. 
That also brings it together in a way that `cert` is `generate` + `add`.

What do you think?

mpolden:
Agree that it's slightly cleaner, but I don't think the overhead of the extra sub-commands and documentation of how they relate to each-other is worth it and will just cause confusion. I also think this is a rare use-case in practice.

How about keeping `vespa auth cert` as-is, and simply add `--no-package`?

client/go/cmd/cert.go
```diff
@@ -75,14 +110,35 @@ var deprecatedCertCmd = &cobra.Command{
...
+	Short: "Add certificate to application package",
+	Long: `Add an existing self-signed certificate for Vespa Cloud deployment to your application package.
+
+The certificate will be looked for in the Vespa CLI home directory (see 'vespa
+help config') by default.
+
+The location of the application package can be specified as an argument to this
+command (default '.').`,
+	Example: `$ vespa auth cert -a my-tenant.my-app.my-instance
+$ vespa auth cert -a my-tenant.my-app.my-instance path/to/application/package`,
```
mpolden, line 124: Examples should show how to use `add`.

client/go/cmd/cert.go
```diff
@@ -109,30 +167,91 @@ func doCert(_ *cobra.Command, args []string) error {
...
+	} else if err != nil {
+		return fmt.Errorf("could not open certificate file: %w", err)
+	}
+	defer src.Close()
+	flags := os.O_CREATE | os.O_RDWR
+	if overwriteCertificate {
+		flags |= os.O_TRUNC
+	} else {
+		flags |= os.O_EXCL
+	}
```
mpolden, line 244: Are all these flags portable?

client/go/cmd/cert.go
```diff
@@ -38,34 +62,45 @@ environment variables. This can be useful in continuous integration systems.
...
 	DisableAutoGenTag: true,
 	SilenceUsage:      true,
 	Args:              cobra.MaximumNArgs(1),
 	RunE:              doCert,
 }
 
 var deprecatedCertCmd = &cobra.Command{
-	Use:               "cert",
-	Short:             "Create a new private key and self-signed certificate for Vespa Cloud deployment",
-	Long:              longDoc,
```
mpolden, line 68: Why duplicate the documentation?

mpolden:
Also, maybe `--no-add` is a better name? That fits well with the new command:

```
vespa auth cert --no-add
vespa auth cert add
```

client/go/cmd/cert.go
ean: golang rewrite them into supported windows flags:
https://github.com/golang/go/blob/master/src/syscall/syscall_windows.go#L316
All the possible combinations are supported, if the flags are changed, they may not be.

client/go/cmd/cert.go
ean: The documentation for `auth cert` is updated with the `--no-add` argument, but I avoided adding the new functionality to the deprecated command.

ean:
Yes, `--no-add` fits well. Fixing that.

mpolden (approved):
💯

PR #21391, opened 2022-02-25
- Only wait with timeout if you have any recurring tasks to execute.

baldersheim:
- Increase timeout from 10 -> 100 milliseconds.
- Also consider the VESPA_TIMER_HZ control of timer frequency.

@vekterli PR


vekterli (approved):
:+1:

PR #21393, opened 2022-02-25
Take VESPA_TIMER_HZ into consideration.

baldersheim:
@vekterli PR


vekterli (approved):
:+1: maybe we should have a utility method in `SystemTimer` that adjusts a given input time and use that for both this case and that in #21391 ? Easier to understand a year from now when we look at the code again with our memories erased by the relentless passage of time.

baldersheim:
Good idea, I will make a separate one for that considering all usages.

PR #21395, opened 2022-02-25
Add a application/{app}/prod resource

oyving:
I sketched out this while thinking about issues with disappearing deployments and instances in our application view in the Console. I think the only way out of this is to always show a union of what is currently deployed/registered in ZooKeeper and what is mentioned in `deployment.xml`.

In this draft PR I created a new `/application/{app}/prod` resource that recursively shows all this information. The structure returned is

```yaml
tenant: name
application: application
instances:
  - name: instance-name
    deployments:
      - name: test
         declared: true
         active: null
     - name: prod.aws-us-east-1c
        declared: true
        active:
          <all deployment stuff here>
```

`declared` means the deployment is mentioned in the latest deployment spec. `active` that is non-null means there is a running deployment that we have information about.

@freva and @jonmv - is this a good approach to make sure we don't accidentally hide information about the system state to the user?

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
@@ -92,6 +92,7 @@ enum PathGroup {
     application(Matcher.tenant,
                 Matcher.application,
                 "/application/v4/tenant/{tenant}/application/{application}",
+                "/application/v4/tenant/{tenant}/application/{application}/prod",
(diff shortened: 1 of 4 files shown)


oyving:
Non-deployed instances already covered here: https://github.com/vespa-engine/vespa/blob/fc9db2992f799bffb80df2b1d57bd4f320da5e7e/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java#L1394

Only gap is test deployments that disappear.

PR #21396, opened 2022-02-25
- Use a utility method that recomputes a timeout if VESPA_TIMER_HZ de…

baldersheim:
…viates from the default of 1000hz.

@vekterli PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/concurrent/SystemTimer.java
+++ b/vespajlib/src/main/java/com/yahoo/concurrent/SystemTimer.java
@@ -1,6 +1,7 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.concurrent;
 
+import java.time.Duration;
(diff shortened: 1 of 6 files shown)


vespajlib/src/main/java/com/yahoo/concurrent/SystemTimer.java
```diff
@@ -32,8 +32,12 @@ public static int detectHz() {
         return hz;
     }
 
+    public static int adjustTimeoutByDetectedHz(int timeoutMS) {
```
vekterli, line 35: Should we explicitly mention the time unit in the name to make it more obvious what unit is used? E.g. `adjustMsTimeoutByDetectedHz` or `adjustTimeoutMsByDetectedHz`?

vekterli (approved):
:+1:

vespajlib/src/main/java/com/yahoo/concurrent/SystemTimer.java
baldersheim: I used a typesafe Duration instead.

PR #21399, opened 2022-02-25
Prepare for different semantics for getData and getChild member functions

toregge:
in internal btree node.

@geirst : please review


geirst (approved):
👍

PR #21402, opened 2022-02-25
Add member functions to get and set btree internal node child refs that

toregge:
will use relaxed memory order instead of acquire and release.

@geirst : please review


geirst (approved):
👍

PR #21407, opened 2022-02-25
Reuse node indexes

bratseth:


config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
```diff
@@ -68,6 +68,9 @@ public RegionName region() {
         return region;
     }
 
+    /** Returns the string "region.environment" */
```
mpolden, line 71: Should be "environment.region".

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
```diff
@@ -62,11 +65,13 @@
...
     public NodeRepositoryProvisioner(NodeRepository nodeRepository,
                                      Zone zone,
-                                     ProvisionServiceProvider provisionServiceProvider, FlagSource flagSource) {
+                                     ProvisionServiceProvider provisionServiceProvider,
+                                     FlagSource flagSource) {
         this.nodeRepository = nodeRepository;
         this.allocationOptimizer = new AllocationOptimizer(nodeRepository);
         this.capacityPolicies = new CapacityPolicies(nodeRepository);
         this.zone = zone;
+        this.flagSource = flagSource;
```
mpolden, line 74: This can already be accessed through `nodeRepository.flagSource()`.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeRepositoryProvisioner.java
```diff
@@ -113,7 +118,8 @@ public NodeRepositoryProvisioner(NodeRepository nodeRepository,
                                                                                  : requested.minResources().nodeResources();
             nodeSpec = NodeSpec.from(requested.type());
         }
-        return asSortedHosts(preparer.prepare(application, cluster, nodeSpec, groups), resources);
+        var reuseIndexes = Flags.REUSE_NODE_INDEXES.bindTo(flagSource).with(FetchVector.Dimension.ZONE_ID, zone.systemLocalValue()).value();
```
mpolden, line 121: Not sure if it's intended, but if you do it like this the flag value will only be read once, when this is constructed.

To read it every time you should pass the `BooleanFlag` to `prepare` and call `.value()` there.

bratseth: The line is in the prepare method.

mpolden: My mistake. 🙂

PR #21409, opened 2022-02-25
Readers of a btree use atomic read with acquire memory ordering when

toregge:
getting btree internal node child ref.

Writer of a btree uses atomic write with release memory ordering when
setting btree internal node child ref in node that is visible to readers
(e.g. when moving child node as part of compacting btree nodes).

@geirst : please review


geirst (approved):
👍

PR #21411, opened 2022-02-26
Add support for architecture in HostResources

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/custom/HostResources.java
```diff
@@ -51,6 +54,7 @@ public HostResources(@JsonProperty("vcpu") Double vcpu,
         this.storageType = validateEnum("storageType", validStorageTypes, storageType);
         this.clusterType = Optional.ofNullable(clusterType).map(cType -> validateEnum("clusterType", validClusterTypes, cType));
         this.containers = requirePositive("containers", containers);
+        this.architecture = validateEnum("architecture", validArchitectures, architecture);
```
freva, line 57: The intention is that this must be set (i.e. non-null)? In that case, `architecture()` can be simplified, but currently set `shared-host` flags must be updated.

hmusum: `shared-host` flag is updated in PR in internal repo, agree that this can be simplified when that PR is merged.

PR #21458, opened 2022-02-28
convert intermedate to DataType objects

arnej27959:
* simplify type variants; all simple builtins have the same handling
* fix wiring for nested type in array/wset
* all annotation types and struct types are owned by the
  document of their schema; move them there if they were
  declared outside.
* tag annotation types and struct types with the name of their
  owning document
* schema ordering must take document inheritance into account, so
  add extra "inheritByDocument" wiring.
* require up-front that a document block inside a schema must have
  the same name as the schema, to simplify the previous mechanism.

@bratseth please review
@baldersheim please review


baldersheim (approved):

bratseth (approved):
👍

PR #21460, opened 2022-02-28
Fix Slobrok race between completion callback and scheduled task [run-systemtest]

vekterli:
@arnej27959 please review

* 00af3469c33fc20fb3db166e2115a10a4e8226e3 adds locking inside `RegisterAPI::handleReqDone()` and `RegisterAPI::RequestDone()` since they may be invoked concurrently through a periodic task and a RPC completion callback. I can't see any ways this can deadlock, but please review to confirm.
* 9a304941c374f98ae93c3eda3884ba6c0f2a8373 only contains minor cleanups with no expected change in semantics.


vekterli:
@arnej27959 ping

arnej27959 (approved):
👍

PR #21466, opened 2022-02-28
Deprecate classes in com.yahoo.docproc

gjoranv:
As you can see in the last commit, we have some issues in the Processing class. We can discuss and resolve them now, or do it in a later PR.


baldersheim (approved):
Yes, we should figure out what to do for the 2 leaking issues.

gjoranv:
Can we merge this PR first? The DocumentTypeManager is currently not an injectable component. I'll try to make it injectable, but inconvenient to do it in this PR in case we need to revert.

PR #21467, opened 2022-02-28
Add a simple adjustTimeoutByDetectedHz to compute a suitable timeout/…

baldersheim:
…naptime based on base VESPA_TIMER_HZ to use for random waiting

when doing a poll based appraoch.

@havardpe and @vekterli PR

### DIFF
--- a/vespalib/src/vespa/vespalib/util/time.h
+++ b/vespalib/src/vespa/vespalib/util/time.h
@@ -92,5 +92,7 @@ public:
  * The default frequency (1000hz) for vespa timer, with environment override VESPA_TIMER_HZ capped to [1..1000]
  */
 uint32_t getVespaTimerHz();
+duration adjustTimeoutByDetectedHz(duration timeout);
+duration adjustTimeoutByHz(duration timeout, long hz);
 
 }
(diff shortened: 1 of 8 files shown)


havardpe (approved):

vespalib/src/vespa/vespalib/util/time.h
```diff
@@ -92,5 +92,7 @@ class Timer
  * The default frequency (1000hz) for vespa timer, with environment override VESPA_TIMER_HZ capped to [1..1000]
  */
 uint32_t getVespaTimerHz();
+duration adjustTimeoutByDetectedHz(duration timeout);
+duration adjustTimeoutByHz(duration timeout, long hz);
```
vekterli, line 96: Should these be in explicit milliseconds rather than `duration`?

vekterli (approved):
:+1:

vespalib/src/vespa/vespalib/util/time.h
baldersheim: Trying to avoid explicit time units and just stick to duration.

PR #21477, opened 2022-03-01
Use wait_until and a deadline instead of a timeout.

baldersheim:
This reduces the need to smaple the time significantly.

@vekterli PR


baldersheim:
Better name for timeout_end ?   deadline, doom ?

vekterli (approved):
:+1:

vekterli:
I personally prefer "deadline". It also seems to be the de facto term used in other software for similar purposes (gRPC et al).

PR #21483, opened 2022-03-01
Wait for deploy service to become ready

### LINKED ISSUE #20789: vespa-cli: Add a retry or two for http/network errors for vespa deploy
```
vespa deploy --wait 300
Uploading application package ... failed
Error: Post "http://127.0.0.1:19071/application/v2/tenant/default/prepareandactivate": EOF
```
Next immediate request succeeds - probably related to docker network

mpolden:
If `--wait` is given, we'll now wait for the deploy service to become ready. In
practice this rarely matters for cloud, but it's useful for local targets.

Fixes #20789

@ean


mpolden:
No functional changes in first commit. Just splitting the code into multiple files.

client/go/vespa/target.go
```diff
@@ -139,6 +134,17 @@ type responseFunc func(status int, response []byte) (bool, error)
...
 type requestFunc func() *http.Request
 
+// waitForOK queries url and returns its status code. If the url returns a non-200 status code, it is repeatedly queried
+// until timeout elapses.
+func waitForOK(url string, certificate *tls.Certificate, timeout time.Duration) (int, error) {
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return 0, err
+	}
+	okFunc := func(status int, response []byte) (bool, error) { return status/100 == 2, nil }
```
ean, line 144: Can we add a `Is2xxStatus()` function and use that for the cases where we do `status/100 ==/!= 2`?
I know you just moved code around now, but I had to think a bit too hard at this pattern to understand what it was trying to do.

ean (approved):

client/go/vespa/target.go
mpolden: It's only in one place now.

ean: I count it 5 places in this PR.

mpolden: I thought you meant the this particular function. PTAL last commit.

ean (approved):
Perfect! Then we just need to wait for factory to be operational I guess.

PR #21486, opened 2022-03-01
Make reconfiguring dynamic vs unlimited throttling not require restarting [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Instead of having one abstract throttler created from bootstrap config,
explicitly create one dynamic and one unlimited throttler and allow for
atomically switching between the two based on received config.

The `MergeHandler` component will now always fetch the current throttler
from the `FileStorHandler` instead of caching it at construction time.

This commit removes the `restart` annotation on the existing throttler
type config enums.


geirst (approved):
👍

PR #21487, opened 2022-03-01
Ean/application test zip issues

ean:
@mpolden

### DIFF
--- a/client/go/cmd/prod.go
+++ b/client/go/cmd/prod.go
@@ -157,8 +157,7 @@ $ vespa prod submit`,
 		}
 		isCI := os.Getenv("CI") != ""
 		if !isCI {
-			fmt.Fprintln(stderr, color.Yellow("Warning:"), "We recommend doing this only from a CD job")
-			fmt.Fprintln(stderr, color.Cyan("Hint:"), "See https://cloud.vespa.ai/en/getting-to-production")
+			printWarning("We recommend doing this only from a CD job", "See https://cloud.vespa.ai/en/getting-to-production")
 		}
 		opts, err := getDeploymentOptions(cfg, pkg, target)
 		if err != nil {
--- a/client/go/cmd/prod.go
+++ b/client/go/cmd/prod.go
@@ -152,11 +152,8 @@ $ vespa prod submit`,
 				"The application must be a Java maven project, or include basic HTTP tests under src/test/application/",
 				"See https://cloud.vespa.ai/en/getting-to-production")
 		}
-		// TODO: Always verify tests. Do it before packaging, when running Maven from this CLI.
-		if !pkg.IsZip() {
-			if err := verifyTests(pkg.TestPath, target); err != nil {
-				return err
-			}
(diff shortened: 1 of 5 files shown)


client/go/vespa/application.go
```diff
@@ -153,25 +154,23 @@ func FindApplicationPackage(zipOrDir string, requirePackaging bool) (Application
...
+			return ApplicationPackage{Path: path, TestPath: testPath}, nil
 		}
-		return ApplicationPackage{Path: filepath.Join(zipOrDir, "src", "main", "application")}, nil
+		return ApplicationPackage{Path: path}, nil
 	}
 	if util.PathExists(filepath.Join(zipOrDir, "services.xml")) {
 		return ApplicationPackage{Path: zipOrDir}, nil
 	}
-	return ApplicationPackage{}, errors.New("Could not find an application package source in '" + zipOrDir + "'")
+	return ApplicationPackage{}, fmt.Errorf("Could not find an application package source in '%s'", zipOrDir)
```
mpolden, line 224: Errors should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

client/go/cmd/prod.go
```diff
@@ -377,18 +374,29 @@ func prompt(r *bufio.Reader, question, defaultAnswer string, validator func(inpu
...
+	testPath := app.TestPath
+	if app.IsZip() {
+		path, err := app.Unzip(true)
+		if err != nil {
+			return err
+		}
+		defer os.RemoveAll(path)
+		testPath = path
 	}
-	for suite, required := range suites {
```
mpolden: I liked the map approach better as the destructuring here was more clear.

client/go/vespa/application.go
```diff
@@ -146,6 +146,55 @@ func (ap *ApplicationPackage) zipReader(test bool) (io.ReadCloser, error) {
...
+	}
+	tmp, err := os.MkdirTemp(os.TempDir(), "vespa-test-pkg")
+	if err != nil {
+		return "", err
+	}
+	path := ap.Path
+	if test {
+		path = ap.TestPath
+	}
+	f, err := zip.OpenReader(path)
```
mpolden, line 161: `f` should be closed in a `defer`.

mpolden, line 161: If this fails, `tmp` will still exist. How about this:
```
cleanTemp := true
tmp, err := os.MkdirTemp(os.TempDir(), "vespa-test-pkg")
if err != nil {
    return "", err
}
defer func() {
    if cleanTemp {
        os.RemoveAll(tmp)
    }
}()
```

then remove the explicit `os.RemoveAll`s below and simply do `cleanTemp = false` before returning.

client/go/cmd/prod.go
ean: agreed, and fixed.

client/go/vespa/application.go
ean: I like it.

mpolden (approved):
💯

PR #21491, opened 2022-03-01
Running the poll task in the singleCommitter thread removes the need …

baldersheim:
…to sync against the sma executor.

This prevents the potential for a deadlock between the tls_executor and the singleCommiter where the latter waits for a future produced the shared executor,
which in turn is blocked a triggerSyncNow doing a sync against the singleCommitter craeting a cycle.
Also remove waitPending -> triggerSyncNow -> waitPendingduring rotation of files as that should not be necessary
after the overhaul of the tls using a singlecommitter thread making all the progress.

@toregge and @geirst PR


toregge (approved):

geirst (commented):
👍

PR #21496, opened 2022-03-02
Ignore architecture when adding or subtracting NodeResources

hmusum:
Need to solve this long-term, but ignore for now as provisoning will
not work as this is now.


aressem (approved):

bratseth:
Code which does arithmetic on NodeResources where compatibility doesn't matter should call justNumbers() on the NodeResource instances first.

hmusum:
Retry in https://github.com/vespa-engine/vespa/pull/21736

PR #21498, opened 2022-03-02
Do an explicit sync of domain part when rotating tls file.

baldersheim:
@geirst or @toregge PR
Necessary to complete #21491


geirst (approved):
👍

PR #21500, opened 2022-03-02
Reduce warnings when using Vespa CLI in CI

mpolden:
@jonmv

### DIFF
--- a/client/go/cmd/config_test.go
+++ b/client/go/cmd/config_test.go
@@ -101,10 +101,17 @@ func TestUseAPIKey(t *testing.T) {
 	}
 }`
 	withEnv("VESPA_CLI_CLOUD_SYSTEM", "public", func() {
+		ci, ok := os.LookupEnv("CI")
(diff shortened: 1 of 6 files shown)


client/go/cmd/config_test.go
```diff
@@ -101,10 +101,17 @@ func TestUseAPIKey(t *testing.T) {
 	}
 }`
 	withEnv("VESPA_CLI_CLOUD_SYSTEM", "public", func() {
+		ci, ok := os.LookupEnv("CI")
```
jonmv, line 104: Hehe :)

jonmv (approved):
Nice!

client/go/cmd/config_test.go
mpolden: 🙈 
Commands need some refactoring so that the command itself and dependencies are contained in a struct. Tests can then pass mocks for things such as HTTP client and process environment into this struct.

PR #21505, opened 2022-03-02
Add replay feed token factory with optional tracking.

toregge:
Add feed operation type and serial number to replay feed tokens.

@geirst : please review
@baldersheim , @vekterli : FYI


geirst (approved):
👍

PR #21506, opened 2022-03-03
Print log lines sent as response to Application Prepare request

ean:
This will show required restart, re-feed and re-index action(s) to the caller.

### DIFF
--- a/client/go/cmd/deploy.go
+++ b/client/go/cmd/deploy.go
@@ -70,9 +70,9 @@ $ vespa deploy -t cloud -z perf.aws-us-east-1c`,
 			return err
 		}
 
-		var sessionOrRunID int64
+		var result vespa.PrepareResult
 		err = util.Spinner(stderr, "Uploading application package ...", func() error {
-			sessionOrRunID, err = vespa.Deploy(opts)
+			result, err = vespa.Deploy(opts)
 			return err
 		})
 		if err != nil {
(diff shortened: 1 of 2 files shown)


client/go/vespa/deploy.go
```diff
@@ -142,17 +154,22 @@ func Activate(sessionID int64, deployment DeploymentOptions) error {
 		return err
 	}
 	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("Got in activate:\n%s\n", string(body))
```
mpolden, line 161: Remove?

client/go/cmd/deploy.go
```diff
@@ -177,3 +180,17 @@ func waitForQueryService(sessionOrRunID int64) error {
...
+	for _, entry := range result.LogLines {
+		switch l := entry.Level; l {
+		case "ERROR":
+			fmt.Fprint(os.Stderr, color.Red(l))
+		case "WARNING":
+			fmt.Fprint(os.Stderr, color.Yellow(l))
+		default:
+			fmt.Fprint(os.Stderr, color.Reset(l))
+		}
+		fmt.Fprintf(os.Stderr, " %s", entry.Message)
```
mpolden, line 194: All of these should print to `stderr`. Also, shouldn't there be a newline printed here? Or does `entry.Message` always contain one?

client/go/vespa/deploy.go
ean: yes

client/go/cmd/deploy.go
ean: Fixed.
And it looks like entry.Message always contains a newline, so we don't need to append that ourselves.

mpolden (approved):
👍

PR #21514, opened 2022-03-03
Let the InvocationService drive the clock instead of having its own t…

baldersheim:
…icking loop.

Also use sleep_until to get intervals indendant of invoke cost as long as cost is within interval.
This also also saves a clock sample and simplifies implementation.

@havardpe PR
This replaces #21470


vespalib/src/tests/invokeservice/invokeservice_test.cpp
```diff
@@ -29,6 +29,25 @@ TEST("require that wakeup is called") {
     EXPECT_EQUAL(countAtStop, a._count);
 }
 
+TEST("require that now is moving forward") {
+    InvokeCounter a;
+    InvokeServiceImpl service(1ms);
+    EXPECT_EQUAL(0u, a._count);
+    steady_time prev = steady_clock::now();
```
havardpe, line 36: I would prefer moving this above the creation of the invoke service to ensure now > prev also the first time

havardpe (commented):
TestClock is missing...

staging_vespalib/src/vespa/vespalib/util/testclock.h
```diff
@@ -0,0 +1,32 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include "clock.h"
+
+namespace vespalib {
+
+class InvokeServiceImpl;
+
+/**
```
havardpe, line 10: this comment seems copied

baldersheim: Fixed

vespalib/src/tests/invokeservice/invokeservice_test.cpp
baldersheim: Fixed

PR #21519, opened 2022-03-03
relative std::this_thread::sleep_for is more efficient than absolute …

baldersheim:
…std::this_thread::sleep_until. This is the opposite of std::condition_variable::wait_until/wait_for.

@vekterli PR


vekterli (approved):
:+1:

PR #21524, opened 2022-03-03
Inline small getters

baldersheim:
@vekterli PR


vekterli (approved):
:+1: looks like these could also safely be tagged `noexcept`

baldersheim:
noexcept added.

PR #21527, opened 2022-03-03
Make global alloc hugepages failure flag atomic

vekterli:
@baldersheim please review

### DIFF
--- a/vespalib/src/vespa/vespalib/util/alloc.cpp
+++ b/vespalib/src/vespa/vespalib/util/alloc.cpp
@@ -1,5 +1,6 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #include "alloc.h"
+#include "atomic.h"
 #include "memory_allocator.h"
 #include "round_up_to_page_size.h"
 #include <sys/mman.h>
@@ -17,11 +18,13 @@
 #include <vespa/log/log.h>
 LOG_SETUP(".vespalib.alloc");
 
+using namespace vespalib::atomic;
+
 namespace vespalib {
 
 namespace {
 
-volatile bool _G_hasHugePageFailureJustHappened(false);
+std::atomic<bool> _G_hasHugePageFailureJustHappened(false);
 bool _G_SilenceCoreOnOOM(false);
 int  _G_HugeFlags = 0;
 size_t _G_MMapLogLimit = std::numeric_limits<size_t>::max();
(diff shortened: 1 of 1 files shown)


vespalib/src/vespa/vespalib/util/alloc.cpp
```diff
@@ -1,5 +1,6 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #include "alloc.h"
+#include "atomic.h"
```
baldersheim, line 3: <atomic> ?

vespalib/src/vespa/vespalib/util/alloc.cpp
```diff
@@ -328,8 +331,8 @@ MMapAllocator::salloc(size_t sz, void * wantedAddress)
         }
         buf = mmap(wantedAddress, sz, prot, flags | _G_HugeFlags, -1, 0);
         if (buf == MAP_FAILED) {
-            if ( ! _G_hasHugePageFailureJustHappened ) {
-                _G_hasHugePageFailureJustHappened = true;
+            if ( ! load_relaxed(_G_hasHugePageFailureJustHappened)) {
```
baldersheim, line 334: if ( ! _G_hasHugePageFailureJustHappened.exchange(true)) {  atomic load/and store

vespalib/src/vespa/vespalib/util/alloc.cpp
```diff
@@ -347,8 +350,8 @@ MMapAllocator::salloc(size_t sz, void * wantedAddress)
                 }
             }
         } else {
-            if (_G_hasHugePageFailureJustHappened) {
-                _G_hasHugePageFailureJustHappened = false;
+            if (load_relaxed(_G_hasHugePageFailureJustHappened)) {
```
baldersheim, line 353: if ( _G_hasHugePageFailureJustHappened.exchange(false)) { atomic load/and store

baldersheim (changes requested):

vespalib/src/vespa/vespalib/util/alloc.cpp
vekterli: Is it worth a `cmpxchg` to skip a debug log?

baldersheim: No, good point.

vespalib/src/vespa/vespalib/util/alloc.cpp
baldersheim: Ah, was not aware of those.

vespalib/src/vespa/vespalib/util/alloc.cpp
baldersheim: good point.

baldersheim (approved):

PR #21529, opened 2022-03-03
- Remove provider, Handler,Waitable, PollAble, Interruptable as separ…

baldersheim:
…ate interface.

- Hold the lock when closing the ConfigHolder so that noone risk getting stuck inside the wait_until.
- Use wait_until instead of poll.

@vekterli PR


vekterli (approved):
:+1:

PR #21546, opened 2022-03-04
process code

havardpe:

### DIFF
--- a/vespalib/src/tests/process/process_test.cpp
+++ b/vespalib/src/tests/process/process_test.cpp
@@ -27,13 +27,13 @@ TEST(ProcessTest, simple_run_ignore_output_failure) {
 
 TEST(ProcessTest, simple_run) {
     vespalib::string out;
-    EXPECT_TRUE(Process::run("echo -n foo", out));
+    EXPECT_TRUE(Process::run("/bin/echo -n foo", out));
     EXPECT_EQ(out, "foo");
 }
 
 TEST(ProcessTest, simple_run_failure) {
     vespalib::string out;
-    EXPECT_FALSE(Process::run("echo -n foo; false", out));
+    EXPECT_FALSE(Process::run("/bin/echo -n foo; false", out));
     EXPECT_EQ(out, "foo");
 }
 
--- a/vespalib/src/tests/process/process_test.cpp
+++ b/vespalib/src/tests/process/process_test.cpp
@@ -1,0 +1,112 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/vespalib/process/process.h>
+#include <vespa/vespalib/data/slime/slime.h>
+#include <vespa/vespalib/data/slime/json_format.h>
+#include <vespa/vespalib/data/simple_buffer.h>
+#include <vespa/vespalib/gtest/gtest.h>
+
+using vespalib::Input;
+using vespalib::Output;
+using vespalib::Process;
+using vespalib::SimpleBuffer;
+using vespalib::Slime;
+using vespalib::slime::JsonFormat;
+
+//-----------------------------------------------------------------------------
+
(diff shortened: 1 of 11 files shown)


vespalib/src/vespa/vespalib/process/process.cpp
```diff
@@ -0,0 +1,153 @@
...
+    REQUIRE(pipe_in.valid() && pipe_out.valid());
+    pid_t pid = fork();
+    REQUIRE(pid != -1);
+    if (pid == 0) {
+        dup2(pipe_in.read_end.fd(), STDIN_FILENO);
+        dup2(pipe_out.write_end.fd(), STDOUT_FILENO);
+        if (capture_stderr) {
+            dup2(pipe_out.write_end.fd(), STDERR_FILENO);
+        } else {
+            int dev_null = open("/dev/null", O_WRONLY);
```
toregge, line 37: Might need to check for failure. if dev_null < 0 then open failed, typical case when we have run out of available file descriptors.

baldersheim (approved):
Significantly better than the old code.

vespalib/src/vespa/vespalib/process/process.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+int
+Process::join()
+{
+    pid_t res;
+    int status;
+    do {
+        res = waitpid(_pid, &status, 0);
+    } while ((res == -1) && (errno == EINTR));
```
toregge, line 114: Note that waitpid() can return failure with errno set to ECHILD if child process has exited and SIGCHLD is ignored.

vespalib/src/tests/process/process_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+TEST(ProcessTest, simple_run_ignore_output_failure) {
+    EXPECT_FALSE(Process::run("false"));
+}
+
+//-----------------------------------------------------------------------------
+
+TEST(ProcessTest, simple_run) {
+    vespalib::string out;
+    EXPECT_TRUE(Process::run("echo -n foo", out));
```
toregge, line 30: Consider `/bin/echo` instead of `echo`. The `sh` builtin `echo` does not understand `-n` option on darwin.

vespalib/src/tests/process/process_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+    EXPECT_EQ(out, "foo");
+}
+
+TEST(ProcessTest, simple_run_failure) {
+    vespalib::string out;
+    EXPECT_FALSE(Process::run("echo -n foo; false", out));
```
toregge, line 36: Consider `/bin/echo` instead of `echo`. The `sh` builtin `echo` does not understand `-n` option on darwin.

toregge (approved):

vespalib/src/vespa/vespalib/process/process.cpp
havardpe: what should we do in that case? I figured dup2/close will be 'ignored' if open returns -1, which is 'best effort'

vespalib/src/vespa/vespalib/process/process.cpp
havardpe: how to handle that better than assert/REQUIRE? we probably have no way of finding out which child has exited?

vespalib/src/tests/process/process_test.cpp
havardpe: fixed

vespalib/src/tests/process/process_test.cpp
havardpe: fixed

vespalib/src/vespa/vespalib/process/process.cpp
toregge: I agree, both dup2() and close() will return EBADF.

PR #21551, opened 2022-03-04
Add note about verifying that the new vespa-cli version works

ean:
We had an instance where the brew bottles from the PR was not uploaded and made
available when bumping the version.  Their BrewTestBot should merge an approved
PR, when it does it will trigger a bottle upload to make the new version
available for install.

See PRs for details:
https://github.com/Homebrew/homebrew-core/pull/96173
https://github.com/Homebrew/homebrew-core/pull/96229

FYI @kkraune

### DIFF
--- a/client/go/Makefile
+++ b/client/go/Makefile
@@ -31,6 +31,9 @@ all: test checkfmt install
 
 # Bump the version of the vespa-cli formula and create a pull request to Homebrew repository.
 #
+# When the homebrew PR has been merged, verify that the prebuilt bottles are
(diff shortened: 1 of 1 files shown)


mpolden (commented):
> The problem was caused by a manual merge to the homebrew repository, instead of letting their automated BrewTestBot handle the merge, which also triggers a bottle upload to the correct place.

Can you mention that we should verify this?

ean:
Done

mpolden (approved):

PR #21555, opened 2022-03-04
Refactor cmd package to eliminate package-level state

mpolden:
Sorry for the large diff, but this was way overdue.

Everything is now encapsulated in the `CLI` struct which holds a reference to
the Cobra command tree and other dependencies.

@ean


mpolden:
This also ends the `util.ActiveHttpClient` fun-times.

client/go/util/http.go
```diff
@@ -38,20 +29,6 @@ func (c *defaultHttpClient) UseCertificate(certificates []tls.Certificate) {
...
-	return &defaultHttpClient{
-		client: &http.Client{Timeout: timeout},
-	}
-}
-
-func HttpDo(request *http.Request, timeout time.Duration, description string) (*http.Response, error) {
-	if request.Header == nil {
-		request.Header = make(http.Header)
-	}
-	request.Header.Set("User-Agent", fmt.Sprintf("Vespa CLI/%s", build.Version))
```
ean, line 51: Looks like this entirely removed the User-Agent override?
I think i would be best to keep that, then we can keep somewhat track of older versions in use.

ean, line 51: Looks like this entirely removed the User-Agent override?
I think i would be best to keep that, then we can keep somewhat track of older versions in use.

client/go/cmd/root.go
```diff
@@ -39,63 +102,77 @@ Vespa documentation: https://docs.vespa.ai`,
...
+	if f, ok := c.Stderr.(*os.File); ok {
+		c.Stderr = colorable.NewColorable(f)
+	}
+	if quiet, _ := c.config.get(quietFlag); quiet == "true" {
+		c.Stdout = ioutil.Discard
 	}
 	log.SetFlags(0) // No timestamps
-	log.SetOutput(stdout)
-	colorValue, _ := config.Get(colorFlag)
+	log.SetOutput(c.Stdout)
```
ean, line 169: There is a good mix of `log.Printf` and `fmt.Fprintf` usage, when do we use what?

ean, line 169: There is a good mix of `log.Printf` and `fmt.Fprintf` usage, when do we use what?

client/go/auth0/auth0.go
```diff
@@ -166,13 +166,7 @@ func (a *Auth0) PrepareSystem(ctx context.Context) (*System, error) {
 
 		res, err := tr.Refresh(ctx, a.system)
 		if err != nil {
-			// ask and guide the user through the login process:
```
ean, line 169: Not entirely sure about this one. Automatically renewing the token makes it all a bit easier for the user. But then again, if there is an intermittent network issue that casued the refresh to fail, starting the entire flow again is not really needed.

What is the lifetime of a refresh token?

ean, line 169: Not entirely sure about this one. Automatically renewing the token makes it all a bit easier for the user. But then again, if there is an intermittent network issue that casued the refresh to fail, starting the entire flow again is not really needed.

What is the lifetime of a refresh token?

ean (commented):
I am assuming no logical changes in this PR, hard to do a line by line review.

The overall structure looks good, and should make the entire CLI much easier to test moving forward.

client/go/cmd/root.go
mpolden: `log` should be preferred since that always point to `c.Stdout`. When using `fmt`, the caller needs to remember to pass the custom writer.

I didn't intend to update the use of `fmt` in this PR.

client/go/auth0/auth0.go
mpolden: Before my change this didn't work as the re-authentication output gets mangled with the spinner output. It's better to return the error and tell the user to re-authenticate in this case.

Note that this is different from refresh. Here you're doing a full re-authentication with the browser flow if refresh failed. Re-running `vespa auth login` is a very minor inconvenience in this case.

client/go/util/http.go
mpolden: Thanks, that was unintentional. Fixed.

client/go/cmd/root.go
ean: My intent was only to know what to prefer in the future. Not that all fmt callers should be changed.

Thanks.

client/go/auth0/auth0.go
ean: ok for me

ean (approved):

PR #21556, opened 2022-03-04
Revert "Fix Slobrok race between completion callback and scheduled task [run-systemtest]"

vekterli:
@geirst please review. TSan detected a possible mutex order inversion transitively caused by doing a `SubRef` when holding a mutex. Will need to choose another approach for fixing this.
@arnej27959 FYI

Reverts vespa-engine/vespa#21460


geirst (approved):
👍

PR #21558, opened 2022-03-04
populate extra fields

arnej27959:
* for consistency, ensure we call populateWithStructFields etc
  also for fields declared outside the document block.

@bratseth please review
NOTE: the main effect here seems to be extra index-info for position sub-fields; it's unclear to me that this is what we want, but it already happens for in-document fields, with this change it becomes consistent.


bratseth (approved):
👍

PR #21560, opened 2022-03-05
Doc type manager as injectable component [run-systemtest]

gjoranv:
I didn't expect it to be this simple, so let's see what factory says. There used to be self-subscribing of config that made this difficult, but that was probably removed some time back.


baldersheim (approved):
Complexity has been gradually reduced over time.

baldersheim:
@gjoranv I also suggest that you go over our code that has DocumentmanagerConfig injected, and let them take the DocumentTypeManager instead.
I see that atleast ReindexingV1ApiHandler and IndexingProcessor does that.

PR #21570, opened 2022-03-07
Use AtomicEntryRef in single value enumerated attribute for mapping

toregge:
from local doc id to enumerated value reference.

@geirst : please review


geirst (approved):
👍

PR #21571, opened 2022-03-07
allow function replacement

arnej27959:
* for backwards-compatibility, allow a rank-profile function
  to be replaced (with a warning)
* also, allow multiple attribute settings for same attribute

@bratseth or @baldersheim please review

### DIFF
--- a/config-model/src/main/javacc/IntermediateParser.jj
+++ b/config-model/src/main/javacc/IntermediateParser.jj
@@ -1909,8 +1909,9 @@ void function(ParsedRankProfile profile) :
         func.setInline(inline);
         var old = profile.addOrReplaceFunction(func);
         if (old != null) {
+            /* TODO disallow this for Vespa 8 */
             deployLogger.logApplicationPackage(Level.WARNING, "Function '" + func.name()
-                                               + "' replaces a previous function with the same name in rank profile '"
+                                               + "' is defined twice in rank profile '"
                                                + profile.name() + "'");
         }
     }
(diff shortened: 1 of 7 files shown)


baldersheim (approved):

config-model/src/main/javacc/IntermediateParser.jj
```diff
@@ -1910,7 +1907,12 @@ void function(ParsedRankProfile profile) :
     {
         func.setExpression(expression);
         func.setInline(inline);
-        profile.addFunction(func);
+        var old = profile.addOrReplaceFunction(func);
+        if (old != null) {
+            deployLogger.logApplicationPackage(Level.WARNING, "Function '" + func.name()
+                                               + "' replaces a previous function with the same name in rank profile '"
```
bratseth, line 1913: Shouldn't this just say "[function] is defined twice"?

config-model/src/main/javacc/IntermediateParser.jj
```diff
@@ -1910,7 +1907,12 @@ void function(ParsedRankProfile profile) :
     {
         func.setExpression(expression);
         func.setInline(inline);
-        profile.addFunction(func);
+        var old = profile.addOrReplaceFunction(func);
+        if (old != null) {
+            deployLogger.logApplicationPackage(Level.WARNING, "Function '" + func.name()
```
bratseth, line 1912: I suggest a TODO on changing this to an error on Vespa 8.

config-model/src/main/javacc/IntermediateParser.jj
arnej27959: > Shouldn't this just say "[function] is defined twice"?

I will change this (the unit test was very strict)

bratseth (approved):
👍

PR #21574, opened 2022-03-07
Ensure we match the service name exactly

ean:
Services 'container' and 'metrisproxy-container' would both match for
"container". Which one was chosen would depend on the order they were returned
from the sentinel.


bjorncs (approved):
👍

PR #21580, opened 2022-03-07
struct/document name collision fails

arnej27959:
* the unit tests claims that this works, and seems to verify that.
  But the generated config isn't valid.

this is basically the same problem as:
different structs (in unrelated .sd files) can't have the same name.


bratseth (approved):
👍

PR #21581, opened 2022-03-07
Make attribute Status fields relaxed atomic

vekterli:
@geirst please review. I only atomic-ified the fields written by `updateStatistics()` since that was what TSan complained about. Let me know if I should do the same for the remaining fields (`_lastSyncToken` was already atomic).

Fields could be written and read from separate threads.


geirst (approved):
👍 
@toregge FYI

toregge (approved):

baldersheim (approved):

PR #21590, opened 2022-03-08
Detect that we are running in docker-on-mac and automatically reduce …

baldersheim:
…timer frequency due to expensive

calls to sample time. This might reduce timeout accuracy from 1ms to 10ms, but that is fine for development.
docker-on-mac will never be supported as production environment.

@arnej27959 and @hmusum PR
@jobergum FYI

### DIFF
--- a/vespabase/src/common-env.sh
+++ b/vespabase/src/common-env.sh
@@ -118,10 +118,14 @@ add_valgrind_suppressions_file() {
 }
 
 optionally_reduce_base_frequency() {
-    os_release=`uname -r`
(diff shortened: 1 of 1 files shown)


arnej27959 (changes requested):
we can not echo to stdout here; it will appear as output from things like vespa-visit.  Use ":" not "echo".

baldersheim:
@arnej27959 echo removed. PTAL

arnej27959 (approved):
looks good to me

jobergum:
💯 🥇

PR #21595, opened 2022-03-08
Make MatchEngine node state polling atomic

vekterli:
State is polled across threads, so access must be well-defined.


vekterli:
@geirst please review. Hit submit a bit too soon 🦄

geirst (approved):
👍

PR #21602, opened 2022-03-08
Increase thread stack size for config proxy and logserver

hmusum:
448k or more is required on arm64

@baldersheim I think you added this setting originally


baldersheim (approved):
Any idea why this i sso different on arm64 vx x86_64 ?

hmusum:
Related to page size being larger on arm64 (64k) and min stack, thread stack guard zone and thread stack shadow zone needing to be a multiple of that, see https://bugs.openjdk.java.net/browse/JDK-8163363?focusedCommentId=14256491&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14256491 for a detailed explanation.

PR #21605, opened 2022-03-09
make some variables atomic to make tsan happy

havardpe:

@vekterli please review

### DIFF
--- a/fastos/src/tests/thread_test_base.hpp
+++ b/fastos/src/tests/thread_test_base.hpp
@@ -5,7 +5,7 @@
 #include <chrono>
 #include <thread>
 
-static volatile int64_t number;
+static std::atomic<int64_t> number;
 #define INCREASE_NUMBER_AMOUNT 10000
 
 using namespace std::chrono_literals;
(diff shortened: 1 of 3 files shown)


fastos/src/tests/thread_test_base.hpp
```diff
@@ -101,7 +101,7 @@ void ThreadTestBase::Run (FastOS_ThreadInterface *thread, void *arg)
 
          int sleepOn = (INCREASE_NUMBER_AMOUNT/2) * 321/10000;
          for (int i=0; i<(INCREASE_NUMBER_AMOUNT/2); i++) {
-            number = number + 2;
+            number += 2;
```
vekterli, line 104: Does this test try to check any thread safety issues by itself? If so, should this be a relaxed `fetch_add` to avoid introducing implicit barriers that may hide other things?

vekterli (approved):
:+1:

fastos/src/tests/thread_test_base.hpp
havardpe: might be, relaxing it

havardpe:
@vekterli PTAL

vekterli (approved):
:+1:

PR #21610, opened 2022-03-09
signal while holding lock

havardpe:

Not exactly sure why this becomes a problem with the adaptive version and not the normal version. However, not holding the lock while signalling will in general require more magic to be safe. Also, it is not even faster to do so, which means we should just stop doing it altogether.

@vekterli please review


vekterli (approved):
:+1:

baldersheim (approved):

PR #21612, opened 2022-03-09
Interrupt reconfig thread when shutting down

jonmv:
@bjorncs please reivew


jonmv:
It would appear we don't have sufficient control to do this.

PR #21615, opened 2022-03-09
Use app platform version during config model feature flags resolution

hakonhall:
Allows adding a feature flag condition like

```
{
    "type": "relational",
    "dimension": "vespa-version"
    "predicate": ">7.555.50"
}
```

to enable a config model property with the rollout of a Vespa release.  This ensures the property change is tested just as rigorously as any other change that rolls out with a Vespa release.


bjorncs (approved):
Nice!

PR #21618, opened 2022-03-09
fix tsan warnings

havardpe:
Portal code calls attach from the reactor thread, so most of these
warnings will not be relevant there. However, they indicate that we
need to be careful about creating the object that will get the
callback before creating the reactor token. Also, the token itself is
in need of synchronization (like any smart-ptr-like-thing) which makes
deleting it inside the call-back somewhat problematic unless there is
a lock or it was created in the reactor thread in the first place.

@vekterli please review


vekterli (approved):
:+1:

PR #21654, opened 2022-03-11
Ean/spinner issues in ci

ean:

### DIFF
--- a/client/go/cmd/login.go
+++ b/client/go/cmd/login.go
@@ -1,10 +1,21 @@
 package cmd
 
 import (
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
+	"github.com/vespa-engine/vespa/client/go/auth"
 	"github.com/vespa-engine/vespa/client/go/auth/auth0"
 )
 
+// newLoginCmd runs the login flow guiding the user through the process
+// by showing the login instructions, opening the browser.
+// Use `expired` to run the login from other commands setup:
+// this will only affect the messages.
 func newLoginCmd(cli *CLI) *cobra.Command {
 	return &cobra.Command{
 		Use:               "login",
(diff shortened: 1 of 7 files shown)


client/go/cmd/login.go
```diff
@@ -27,7 +37,57 @@ func newLoginCmd(cli *CLI) *cobra.Command {
...
+				return fmt.Errorf("could not start the authentication process: %w", err)
+			}
+
+			fmt.Printf("Your Device Confirmation code is: %s\n\n", state.UserCode)
+
+			fmt.Println("If you prefer, you can open the URL directly for verification")
+			fmt.Printf("Your Verification URL: %s\n\n", state.VerificationURI)
+
+			fmt.Println("Press Enter to open the browser to log in or ^C to quit...")
+			fmt.Scanln()
```
mpolden, line 51: These should print to `cli.Stdout` or just use `log.PrintXXX`. Same below.

client/go/cmd/root.go
```diff
@@ -203,6 +205,19 @@ func (c *CLI) configureFlags() {
 	c.flags = &flags
 }
 
+func (c *CLI) configureSpinner() {
+	// Explicitly disable spinner for Screwdriver. It emulates a tty but
+	// \r result in a newline, and output gets truncated.
+	_, screwdrvier := os.LookupEnv("SCREWDRIVER")
```
mpolden, line 211: This should access the environment through `c.Environment`.

mpolden: Typo: s/screwdrvier/screwdriver/

client/go/cmd/login.go
ean: Fixed, but we are left with the ugly fmt.Scanln()` duckling. But I leave that to be handled as part of the ongoing discussion we have about the UX of this command.

mpolden: `fmt.Fscanln(cli.Stdin)` should work.

mpolden (approved):

PR #21660, opened 2022-03-11
Add a restart action (with deferred activation) when switching major versions [run-systemtest]

jonmv:
@hmusum @mpolden please review. I see now that more or less the same problem was attacked in #21645.


jonmv:
Perhaps #21661 instead, then ...

PR #21662, opened 2022-03-11
Make B-tree bucket database values atomic to ensure well-defined access [run-systemtest]

vekterli:
@geirst and @toregge please review. This fixes a spurious TSan warning.
Have used the added test to run some 100M+ read/write interactions with this fix included and no further warnings have been observed.

Existing implementation already used explicit acquire/release fences to
ensure visibility from writer to concurrent readers, but the values written/read
were not of an atomic type and thus _technically_ considered a data race.

This commit adds an `AtomicValueWrapper` wrapper to vespalib which looks and acts
much like the existing `AtomicEntryRef`, but for primitive types that are not
related to `EntryRef`s. The bucket DB B-tree primitive u64 value type is replaced
with an atomic wrapper, and explicit memory fences are replaced with release stores
and acquire loads on the atomic values themselves to ensure they form correct
pairs between writer and readers.

### DIFF
--- a/storage/src/tests/distributor/btree_bucket_database_test.cpp
+++ b/storage/src/tests/distributor/btree_bucket_database_test.cpp
@@ -2,8 +2,11 @@
 
 #include "bucketdatabasetest.h"
 #include <vespa/storage/bucketdb/btree_bucket_database.h>
+#include <vespa/vespalib/util/count_down_latch.h>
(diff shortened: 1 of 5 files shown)


toregge (approved):

toregge (approved):

baldersheim (approved):

geirst (commented):
👍

PR #21666, opened 2022-03-13
Arnej/add alternative to global gettype

arnej27959:

@bratseth please review
@baldersheim @gjoranv FYI


document/src/main/java/com/yahoo/document/DocumentType.java
```diff
@@ -138,6 +139,39 @@ public StructDataType contentStruct() {
...
     }
 
+    /**
+     * Get a struct declared in this document (or any inherited
+     * document). Returns null if no such struct was found.
+     *
+     * @param name the name of the struct
+     * @return reference to a struct data type, or null
+     **/
+    public StructDataType getDeclaredStructType(String name) {
```
bratseth, line 149: Usually, we use "declared" to mean "in this only" (it was declared in this).

bratseth (approved):
Merging, but notice my comment.

PR #21667, opened 2022-03-14
Use entity tag to expire sample apps cache

mpolden:
@ean

### DIFF
--- a/client/go/cmd/clone.go
+++ b/client/go/cmd/clone.go
@@ -55,7 +55,8 @@ variable.`,
 			if len(args) != 2 {
 				return fmt.Errorf("expected exactly 2 arguments, got %d", len(args))
 			}
-			return cloneApplication(cli, args[0], args[1], forceClone)
+			cloner := &cloner{cli: cli, discardCache: forceClone}
+			return cloner.Clone(args[0], args[1])
 		},
 	}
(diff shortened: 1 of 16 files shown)


client/go/cmd/clone.go
```diff
@@ -108,71 +116,147 @@ func (c *cloner) Clone(applicationName, path string) error {
...
+			os.Remove(f.Name())
 		}
+	}()
+	if _, err := io.Copy(f, zipReader); err != nil {
+		return "", fmt.Errorf("could not write sample apps to file: %s: %w", f.Name(), err)
+	}
+	f.Close()
+	path := filepath.Join(c.cli.config.cacheDir, sampleAppsNamePrefix)
+	if etag != "" {
+		path += "_" + etag
```
ean, line 248: I am not entirely sold on this yet.
Won't be be less bookkeeping to keep one file for the content, and one for the etag. Then we don't need the directory traversal and garbage collection.

And if we default the etag to empty string for all etag file failure cases I think the rest should work itself out, even if we end up downloading the file a few more times than required.

mpolden: This way we guarantee that the data and its etag are in sync by not keeping state in a separate file. For example it's not possible to write the cached data, but not the etag, or vice-versa.

We also don't need special handling related to presence of an etag file, atomic writing of the etag file and parsing of its contents.

ean (approved):

PR #21672, opened 2022-03-14
Infer managed access through assertion existence

olaaun:


vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/AthenzAssertion.java
```diff
@@ -36,6 +36,12 @@ public static Builder newBuilder(AthenzRole role, AthenzResourceName resource, S
         return new Builder(role, resource, action);
     }
 
+    public boolean satisfies(AthenzAssertion other) {
```
tokle, line 39: also check `effect`

tokle (approved):

PR #21674, opened 2022-03-14
Convert rcu vector unit test to gtest.

toregge:
Factor out generation holder test to separate unit test.

@geirst : please review


baldersheim (approved):

geirst (commented):
👍

PR #21679, opened 2022-03-14
Make SourceSession pending count atomic to avoid data races

vekterli:
@havardpe please review

Could have used the existing mutex, but this value is polled frequently
by visitor threads so avoiding having to take a lock every time makes sense.

### DIFF
--- a/messagebus/src/vespa/messagebus/sourcesession.cpp
+++ b/messagebus/src/vespa/messagebus/sourcesession.cpp
@@ -9,7 +9,6 @@
 
 using vespalib::make_string;
 
-
 namespace mbus {
 
 SourceSession::SourceSession(MessageBus &mbus, const SourceSessionParams &params)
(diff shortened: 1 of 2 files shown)


messagebus/src/vespa/messagebus/sourcesession.cpp
```diff
@@ -81,16 +80,18 @@ SourceSession::send(Message::UP msg)
...
+                                make_string("Too much pending data (%d messages).", my_pending_count)),
                           std::move(msg));
         }
         msg->pushHandler(_replyHandler);
         if (_throttlePolicy) {
             _throttlePolicy->processMessage(*msg);
         }
-        my_pending_count = ++_pendingCount;
+        ++my_pending_count;
+        _pendingCount.store(my_pending_count, std::memory_order_relaxed);
```
baldersheim, line 94: Modification of _pendingCount is only done while lock is held I presume.

baldersheim (approved):

PR #21680, opened 2022-03-14
Jonmv/major upgrade orchestration

jonmv:
@mpolden please review.

### DIFF
--- a/component/src/main/java/com/yahoo/component/VersionCompatibility.java
+++ b/component/src/main/java/com/yahoo/component/VersionCompatibility.java
@@ -1,0 +1,109 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.component;
+
+import java.util.List;
+import java.util.NavigableMap;
+import java.util.TreeMap;
+
+/**
+ * Logic for what platform and compile versions are incompatible.
+ *
+ * This class is instantiated with a list of versions which are only compatible with other versions
+ * that are not older than themselves. Each entry in this list is represented by a string, and any
+ * major, minor or micro with the value @{code *} means all versions are such boundaries.
+ * Versions A, B are incompatible iff. there exists an incompatibility boundary X such that
+ * (A &ge; X) &ne; (B &ge; X).
+ *
+ * @author jonmv
+ */
+public class VersionCompatibility {
+
+    private final Node root;
+
+    private VersionCompatibility(Node root) {
+        this.root = root;
+    }
+
+    public static VersionCompatibility fromVersionList(List<String> versions) {
+        Node root = new Node();
+        for (String spec : versions) {
+            String[] parts = spec.split("\\.");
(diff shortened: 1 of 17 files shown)


jonmv:
PR build timed out, unrelated.

component/src/test/java/com/yahoo/component/VersionIncompatibilityTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertFalse(compatibility.accept(new Version(0, 0, 0), new Version(6, 0, 0)));
+    }
+
+    @Test
+    public void testIllegalIncompatibilities() {
+        try {
+            VersionCompatibility.fromVersionList(List.of("1", "*"));
+            fail();
+        }
+        catch (IllegalArgumentException expected) { }
```
mpolden, line 62: Consider extracting a method for this pattern.

component/src/main/java/com/yahoo/component/VersionCompatibility.java
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * This class is instantiated with a list of versions which are only compatible with other versions
+ * that are not older than themselves. Each entry in this list is represented by a string, and any
+ * major, minor or micro with the value @{code *} means all versions are such boundaries.
+ * Versions A, B are incompatible iff. there exists an incompatibility boundary X such that
+ * (A &ge; X) &ne; (B &ge; X).
+ *
+ * @author jonmv
+ */
+public class VersionCompatibility {
```
mpolden, line 19: Why do we need this flexibility in verifying incompatible versions? Seeing as we use semantic versioning I don't think we ever need to express incompatibility between 1.2.3 and 1.x.x? Major is sufficient.

component/src/test/java/com/yahoo/component/VersionIncompatibilityTest.java
jonmv: Hmm yes.

component/src/main/java/com/yahoo/component/VersionCompatibility.java
jonmv: The intention was that we could set a different range (`List.of("*.*.*")`) for a test app, in our CD tests.

jonmv: If we set that for a particular application, the system will treat all its packages as incompatible with other platform versions than the one used to compile the package, and we'll have a test for this that runs all the time. I don't think there's a cleaner way to do that.

mpolden: I see, good idea 🙂

mpolden (approved):

component/src/main/java/com/yahoo/component/VersionCompatibility.java
jonmv: Well, if there's only that special value we need, we could of course simplify this a lot ... >_<

Issue #21286, opened 2022-02-21
Support boolean fields in GC expressions


bratseth:

bratseth:
In addition, the Java parser does not support booleans.

vekterli:
> Well that is something we can change, to implement boolean support properly.

Do we want to do this at the same time as adding fundamental bool field support, or should we consider these separately? The former likely takes rather more effort than the latter.

vekterli:
It's not that I expect it to be "hard"™, but with how boolean constants are modelled and wired today it does require changing the C++ parser grammar and/or AST structure, so it's more work than just adding an extra type to the existing field value extraction logic. Hence the question on whether we should split this into two tasks (bool _field_ support vs. proper bool _expression_ support) 🎶🎷🦖

PR #21693, opened 2022-03-15
Basic support for bool fields in document selections [run-systemtest]

vekterli:
@bratseth / @baldersheim please review.

This adds basic support for bool field values in document selections in C++ and Java, re: #21286. No changes to parser grammar etc, so the caveats mentioned in the associated issue still apply.


bratseth (approved):
👍

PR #21702, opened 2022-03-16
Tune criterion for when jobs need to re-run, and some other fixes

jonmv:

This fixes an issue where a fully deployed application change would become "deploying" again, when an upgrade was triggered for the instance, because the production job "completedAt" was too strict.


mpolden (approved):
👍

PR #21710, opened 2022-03-16
Stop using deprecated ioutil functions

mpolden:
The oldest version we're building with is now Go 1.16.

@ean


ean (approved):
Some build failures on SD, but I think that was unrelated. Restarted the job, but please check that before merging.

PR #21711, opened 2022-03-16
Move ownership of target major to ApplicationController, and use for …

jonmv:
…compileVersion


mpolden (approved):

mpolden:
Great, but looks like there are test failures.

jonmv:
Heh, yes, IntelliJ was failing to run some of the webserver tests, so never got to the end ... where the `V` for `VersionStatusTest` is ... >_<

PR #21712, opened 2022-03-16
use atomics for disk footprint to avoid tsan warnings.

havardpe:

@vekterli please review
@baldersheim FYI

The disk footprint of a writeable chunk is written while holding locks A and B, and read while holding lock C. As long as only a single actor can write it this might be ok, @baldersheim please have a quick look.


vekterli (approved):
:+1:

baldersheim (approved):

PR #21714, opened 2022-03-16
Add NTokenGenerator

hakonhall:


vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/NTokenGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * @author hakonhall
+ */
+public class NTokenGenerator {
+    private final Signer signer;
+    private final Supplier<Instant> time;
```
freva, line 22: Use `java.time.Clock`?

freva (approved):

vespa-athenz/src/main/java/com/yahoo/vespa/athenz/api/NTokenGenerator.java
hakonhall: Done

freva (approved):

PR #21716, opened 2022-03-17
Use --release, not --source and --target, for tenant apps with old pa…

jonmv:
…rent

@gjoranv please review and merge.

### DIFF
--- a/tenant-base/pom.xml
+++ b/tenant-base/pom.xml
@@ -393,8 +393,10 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
-                    <source>${target_jdk_version}</source>
-                    <target>${target_jdk_version}</target>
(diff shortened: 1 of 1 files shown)


jonmv:
Or @hmusum :)

hmusum (approved):

tenant-base/pom.xml
```diff
@@ -393,8 +393,10 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>${maven-compiler-plugin.version}</version>
                 <configuration>
-                    <source>${target_jdk_version}</source>
-                    <target>${target_jdk_version}</target>
+                    <jdkToolchain>
```
gjoranv, line 396: jdkToolchain is only valid when using maven-toolchains-plugin with toolchains.xml. Please remove.

jonmv: Thanks.

PR #21722, opened 2022-03-17
Deprecate public org.json

bratseth:


arnej27959 (approved):
looks good to me
@gjoranv FYI

gjoranv:
The abi plugin does not consider protected methods in non-final classes.

Should `JSONString` have been declared final? Only gemini-native extend it, but it doesn't seem very useful:

```
public class JSONStringWrapper extends JSONString {


    public JSONStringWrapper(String content) {
        super(content);
    }

    public JSONStringWrapper() {
        super("");
    }
}
```

bratseth:
> The abi plugin does not consider protected methods in non-final classes.

That's a bug I think.

> Should JSONString have been declared final?

YYes, but that would be a breaking API change in itself so probably not worth it?

gjoranv:
I was fortunately wrong about the abi plugin. The protected method here was inside a private static class (FieldRenderer), so it's not accessible from the outside. I really don't understand why the method was declared public. It does not extend an interface, and is only used inside from the outer class.

As you say, making it final is not worth it. Plus, there are no real API leaks with org.json here.

PR #21741, opened 2022-03-18
fix tsan warnings for transaction log client test

havardpe:

@vekterli please review

### DIFF
--- a/searchlib/src/tests/transactionlog/translogclient_test.cpp
+++ b/searchlib/src/tests/transactionlog/translogclient_test.cpp
@@ -61,21 +61,42 @@ createDomainConfig(uint32_t partSizeLimit) {
                          .setEncoding(Encoding(Encoding::xxh64, Encoding::none_multi));
 }
 
+// Used to signal 'eof' when visiting in a way that makes sure the
+// stuff we have visited is visible to the test code.
+class Eof {
+private:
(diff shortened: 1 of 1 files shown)


searchlib/src/tests/transactionlog/translogclient_test.cpp
```diff
@@ -409,8 +433,7 @@ visitDomainTest(TransLogClient & tls, Session * s1, const vespalib::string & nam
     auto visitor = tls.createVisitor(name, ca);
     ASSERT_TRUE(visitor);
     EXPECT_TRUE( visitor->visit(0, 1) );
-    for (size_t i(0); ! ca._eof && (i < 60000); i++ ) { std::this_thread::sleep_for(10ms); }
-    EXPECT_TRUE( ca._eof );
+    EXPECT_TRUE( ca.wait_for_eof() );
```
vekterli, line 436: Should these eof-waits be `ASSERT_TRUE` instead if it does not make sense to continue the test execution if they do not hold true?

vekterli (approved):
:+1:

searchlib/src/tests/transactionlog/translogclient_test.cpp
havardpe: maybe, some were asserts, some were expects, I just left them as they were

PR #21744, opened 2022-03-18
Some minor improvements

jonmv:
@bratseth please review.


bratseth (approved):
Very nice

bratseth:
Needs ABI spec update

container-search/src/main/java/com/yahoo/prelude/query/CompositeItem.java
```diff
@@ -205,7 +201,7 @@ protected void appendBodyString(StringBuilder buffer) {
     }
 
     /** Composite items should be parenthized when not on the top level */
-    protected boolean shouldParenthize() {
```
jonmv, line 208: Ah, because of this XD

jonmv:
PTAL.

PR #21757, opened 2022-03-21
Fix visitor manager test TSan mutex inversion warning and refactor deadline handling [run-systemtest]

vekterli:
@geirst please review

Defer starting main message dispatch thread until test has enqueued a message
that will be immediately timed out. Avoids having to depend on taking an
internal mutex in the test to prevent racing with queue handoffs. Taking
said mutex triggered a mutex order inversion warning in TSan.

Also refactor visitor queue deadline handling by moving to a strongly typed
time point and monotonic clock. This removes some manual unit scaling arithmetic
that did not appear to be entirely correct in the first place.


geirst (approved):
👍

PR #21761, opened 2022-03-21
Lesters/multiple embedders

lesters:
@bratseth Please review.

### DIFF
--- a/linguistics/src/main/java/com/yahoo/language/process/Embedder.java
+++ b/linguistics/src/main/java/com/yahoo/language/process/Embedder.java
@@ -16,14 +16,14 @@ import java.util.Map;
 public interface Embedder {
 
     /** Name of embedder when none is explicity given */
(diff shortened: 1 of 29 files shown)


linguistics/src/main/java/com/yahoo/language/process/Embedder.java
```diff
@@ -15,9 +15,22 @@
  */
 public interface Embedder {
 
+    /** Name of embedder when none is explicity given */
+    String defaultEmbedderName = "defaultEmbedder";
```
bratseth, line 19: I think we should use "id" not "name" here, and just use the id "default".

bratseth (commented):
👍

linguistics/src/main/java/com/yahoo/language/process/Embedder.java
lesters: :+1:

bratseth (approved):

PR #21768, opened 2022-03-22
support project info

tokle:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/organization/IssueHandler.java
```diff
@@ -2,6 +2,8 @@
 package com.yahoo.vespa.hosted.controller.api.integration.organization;
 
 
+import com.yahoo.slime.Inspector;
+
```
jonmv, line 6:
```suggestion
```

Remove stray added ddependency

jonmv (approved):

PR #21784, opened 2022-03-23
Include architecture in metering

olaaun:
Remove unused metering API


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ResourceMeterMaintainer.java
```diff
@@ -150,14 +150,21 @@ private void reportResourceSnapshots(Collection<ResourceSnapshot> resourceSnapsh
...
-                                                zoneId))
-                                )).values();
+                                Collectors.collectingAndThen(Collectors.groupingBy(node -> node.resources().architecture()),
+                                        nodeList -> nodeList.entrySet()
+                                                .stream()
+                                                .collect(Collectors.toMap(
+                                                        Map.Entry::getKey,
+                                                        entry -> ResourceSnapshot.from(entry.getValue(), clock.instant(), zoneId))
+                                                ))
+                                ))
```
oyving, line 163: When combining collectors in this way it can help readability to assign them to a variable when you create them and use that later. It makes the code less noisy and gives the collector a semantic name.

oyving (changes requested):

oyving (approved):

PR #21792, opened 2022-03-23
Avoid signed integer overflow in hashCode() which could trigger undefined behavior.

toregge:
@geirst : please review

### DIFF
--- a/config/src/vespa/config/frt/frtconnectionpool.cpp
+++ b/config/src/vespa/config/frt/frtconnectionpool.cpp
@@ -97,7 +97,7 @@ namespace {
  * @return the hash value
  */
 int hashCode(const vespalib::string & s) {
-    int hashval = 0;
+    unsigned int hashval = 0;
 
     for (int i = 0; i < (int) s.length(); i++) {
         hashval = 31 * hashval + s[i];


config/src/vespa/config/frt/frtconnectionpool.cpp
```diff
@@ -97,7 +97,7 @@ namespace {
  * @return the hash value
  */
 int hashCode(const vespalib::string & s) {
```
baldersheim, line 99: I suggest changing signature to return uint32_t and remove the std::abs that are lying around.

config/src/vespa/config/frt/frtconnectionpool.cpp
```diff
@@ -97,7 +97,7 @@ namespace {
  * @return the hash value
  */
```
baldersheim, line 98: This is a very bad hash function. Since hostnames are often very similar I strongly suggest using xxhash.
It says here that the hash function should be the same in both java/cpp, but I do not see why.

geirst (approved):
👍

config/src/vespa/config/frt/frtconnectionpool.cpp
geirst: We will address this in a later PR.

PR #21800, opened 2022-03-24
Add comment to tsan-suppressions.txt describing why we suppress

toregge:
deadlock warning related to attribute context lock and enumerated
attribute enum lock.

@geirst : please review


geirst (approved):
👍

PR #21808, opened 2022-03-24
Revert "Use vespalib::string in TestAndSetCondition"

vekterli:
@baldersheim please review, I'll fix the tests and resubmit. I was fooled by the shiny green badge in the PR, I see now that CI for some reason wasn't actually run at all...!

Reverts vespa-engine/vespa#21805


geirst (approved):
👍

PR #21815, opened 2022-03-24
Use std::atomic_ref for updating and reading single value non-enumerated

toregge:
attribute vectors with primitive types.

@geirst : please review
@vekterli : FYI


geirst (approved):
👍

PR #21821, opened 2022-03-25
Always link vespa cli statically

mpolden:
@aressem


mpolden:
This will ensure that all binaries always end up statically linked. The homebrew/brew image has the dynamic loader in a custom path, so the binaries it builds may not work in other Linux environments.

aressem (approved):

PR #21823, opened 2022-03-25
Use std::atomic_ref for updating and reading single value non-enumerated

toregge:
attribute vectors with small numeric types.

@geirst : please review


geirst (approved):
👍

PR #21837, opened 2022-03-27
Avoid multiple inheritance.

baldersheim:
@vekterli PR

### DIFF
--- a/document/src/tests/documentselectparsertest.cpp
+++ b/document/src/tests/documentselectparsertest.cpp
@@ -139,9 +139,9 @@ DocumentUpdate::SP DocumentSelectParserTest::createUpdate(
     const DocumentType* type = _repo->getDocumentType(doctype);
     auto doc = std::make_shared<DocumentUpdate>(*_repo, *type, DocumentId(id));
     doc->addUpdate(std::move(FieldUpdate(doc->getType().getField("headerval"))
-                      .addUpdate(AssignValueUpdate(IntFieldValue(hint)))));
+                      .addUpdate(std::make_unique<AssignValueUpdate>(IntFieldValue(hint)))));
     doc->addUpdate(std::move(FieldUpdate(doc->getType().getField("hstringval"))
-                      .addUpdate(AssignValueUpdate(StringFieldValue(hstr)))));
+                      .addUpdate(std::make_unique<AssignValueUpdate>(StringFieldValue(hstr)))));
     return doc;
 }
 
--- a/document/src/tests/documentselectparsertest.cpp
+++ b/document/src/tests/documentselectparsertest.cpp
@@ -137,11 +137,11 @@ DocumentUpdate::SP DocumentSelectParserTest::createUpdate(
         const std::string& hstr)
 {
     const DocumentType* type = _repo->getDocumentType(doctype);
-    DocumentUpdate::SP doc(new DocumentUpdate(*_repo, *type, DocumentId(id)));
-    doc->addUpdate(FieldUpdate(doc->getType().getField("headerval"))
-                      .addUpdate(AssignValueUpdate(IntFieldValue(hint))));
-    doc->addUpdate(FieldUpdate(doc->getType().getField("hstringval"))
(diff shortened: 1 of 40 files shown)


document/src/vespa/document/update/fieldupdate.h
```diff
@@ -58,23 +53,19 @@ class FieldUpdate : public vespalib::Identifiable,
      * @param update A pointer to the value update to add to this.
      * @return A pointer to this.
      */
-    FieldUpdate& addUpdate(const ValueUpdate& update) {
-        update.checkCompatibility(_field); // May throw exception.
-        _updates.push_back(ValueUpdate::CP(update.clone()));
-        return *this;
-    }
+    FieldUpdate& addUpdate(const ValueUpdate& update);
```
vekterli, line 56: To avoid having to wrap `std::move` around `FieldUpdate(foo).addUpdate(bar)` etc it's possible to use reference qualifiers to add an overload here that can only be selected by the compiler when the function is called on an rvalue (such as a stack-constructed `FieldValue` temporary). Example adding an explicit lvalue and rvalue ref qualifier to the functions (note: have not compiled this locally):

```c++
FieldUpdate& addUpdate(const ValueUpdate& update) &;
FieldUpdate&& addUpdate(const ValueUpdate& update) &&
{
    addUpdate(update);
    return std::move(*this);
}
```
Since `this` implicitly becomes an lvalue inside an rvalue function, the `addUpdate` is a forwarding function and not an infinite recursion since the rvalue function won't resolve itself as a candidate. This is also why we still have to explicitly `std::move(*this)` for the return.

A function call such as `my_rvalue_accepting_func(FieldValue(foo).addUpdate(bar).addUpdate(baz))` should then work as expected.

document/src/tests/testxml.cpp
```diff
@@ -131,6 +127,7 @@ TEST(TestXml, testDocumentUpdate)
         "    <remove>789</remove>\n"
         "  </alter>\n"
         "</document>";
+    std::ostringstream os;
```
vekterli, line 130: Is this needed?

document/src/tests/documentselectparsertest.cpp
```diff
@@ -137,11 +137,11 @@ DocumentUpdate::SP DocumentSelectParserTest::createUpdate(
         const std::string& hstr)
 {
     const DocumentType* type = _repo->getDocumentType(doctype);
-    DocumentUpdate::SP doc(new DocumentUpdate(*_repo, *type, DocumentId(id)));
-    doc->addUpdate(FieldUpdate(doc->getType().getField("headerval"))
-                      .addUpdate(AssignValueUpdate(IntFieldValue(hint))));
-    doc->addUpdate(FieldUpdate(doc->getType().getField("hstringval"))
-                      .addUpdate(AssignValueUpdate(StringFieldValue(hstr))));
+    auto doc = std::make_shared<DocumentUpdate>(*_repo, *type, DocumentId(id));
+    doc->addUpdate(std::move(FieldUpdate(doc->getType().getField("headerval"))
```
vekterli, line 141: I think most (if not all?) of these explicit moves for chained calls can be removed; see comment for rvalue `addUpdate` overload

vekterli (approved):
:+1: see comment for possible simplification of chained calls

document/src/tests/testxml.cpp
baldersheim: No

document/src/vespa/document/update/fieldupdate.h
baldersheim: Thanks, that was better.
I already had a large followup touching same code, so in order to avoid massive merge conflict, I adde that too before reverting the std::move()

document/src/tests/documentselectparsertest.cpp
baldersheim: Indeed, gone

baldersheim:
@vekterli Now you can have a look again, note clone gone.

vekterli (approved):
:+1: one less `clone()` makes for a happier world

PR #21859, opened 2022-03-28
Remove jersey media bundles 3 [run-systemtest]

gjoranv:
Only the 4 last commits are new. Review only, please do not merge.

FYI: @hakonhall for orchestrator and jaxrs_client_utils


jonmv (approved):
This is too easy. We should add more stuff like this 😀

bjorncs (approved):

PR #21864, opened 2022-03-29
Ask if user wants login link in browser opened

oyving:

### DIFF
--- a/client/go/cmd/login.go
+++ b/client/go/cmd/login.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/pkg/browser"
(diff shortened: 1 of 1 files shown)


client/go/cmd/login.go
```diff
@@ -43,18 +44,18 @@ func newLoginCmd(cli *CLI) *cobra.Command {
 				return fmt.Errorf("could not start the authentication process: %w", err)
 			}
 
-			log.Printf("Your Device Confirmation code is: %s\n\n", state.UserCode)
+			log.Printf("Your Device Confirmation code is: %s\n", state.UserCode)
 
-			log.Println("If you prefer, you can open the URL directly for verification")
-			log.Printf("Your Verification URL: %s\n\n", state.VerificationURI)
+			auto_open := confirm("Allow Vespa CLI to open confirmation page in your default browser?")
```
ean, line 49: Remove the "Vespa CLI to" prefix, it reads a too permanent and scary.

client/go/cmd/login.go
```diff
@@ -93,3 +94,22 @@ func newLoginCmd(cli *CLI) *cobra.Command {
 		},
 	}
 }
+
+func confirm(question string) bool {
+	for {
+		var answer string
+
+		log.Printf("%s [Y/n] ", question)
```
ean, line 102: Use `fmt.Fprintf(cli.Stdout, ...)` instead of `log.Printf()`. That way we avoid a newline after the prompt.
At least for me that ensures that I catch the action that is required. When the prompt is on an empty line I just wait for more output.

client/go/cmd/login.go
```diff
@@ -93,3 +94,22 @@ func newLoginCmd(cli *CLI) *cobra.Command {
...
+		fmt.Scanln(&answer)
```
ean, line 103: `fmt.Fscanln(cli.Stdin, ...)`

ean (approved):

PR #21871, opened 2022-03-29
Remove const type qualifier from member functions that should only be

toregge:
called from writer in RcuVector, MultiValueMappingBase and
CondensedBitVector.

@geirst : please review


geirst (approved):
👍

PR #21880, opened 2022-03-29
Disallow relative paths and specs

mpolden:
@jonmv


container-core/src/main/java/com/yahoo/restapi/Path.java
```diff
@@ -26,29 +25,19 @@
...
     // Info about the last match
     private final Map<String, String> values = new HashMap<>();
     private String rest = "";
 
     public Path(URI uri) {
-        this(uri, "");
-    }
-
-    // TODO (freva): Remove, used by factory
-    public Path(URI uri, String optionalPrefix) {
```
mpolden, line 50: I removed this use from Factory.

jonmv (approved):
Nice!

PR #21882, opened 2022-03-29
Undeprecate string features and be explicit

bratseth:


container-search/src/main/java/com/yahoo/search/query/ranking/RankFeatures.java
```diff
@@ -41,11 +41,10 @@ public void put(String name, Tensor value) {
     }
 
     /**
-     * Sets a rank feature to a value represented as a string.
-     *
-     * @deprecated set either a double or a tensor
+     * Sets a rank feature to a string. This will be available as the hash value
+     * of the string in ranking, so it can be used in equality comparisons
+     * with other string, but not for any other purpose.
```
lesters, line 46: *strings

lesters (approved):
LGTM!

PR #21893, opened 2022-03-30
Jonmv/misc 3

jonmv:
@mpolden please review, and indicate whether you like the first or second variant better.


vespajlib/src/main/java/ai/vespa/validation/Hostname.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * A valid hostname.
+ *
+ * @author jonmv
+ */
+public class Hostname extends StringWrapper<Hostname> {
+
+    public static final Pattern hostnameLabel = Pattern.compile("([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]{0,61}[A-Za-z0-9])");
+    public static final Pattern hostnamePattern = Pattern.compile("(?=.{1,255})(" + hostnameLabel + "\\.)*" + hostnameLabel);
```
mpolden, line 16: In practice, hostnames are limited to 64 characters. At least on Linux.

jonmv: Hmm ... so there's a difference in "host part of a URI" and a "hostname", I guess. I used the former definition. Perhaps the latter is significantly different?

bratseth:
I don't think "validation" is a the best name for a package that contains value objects for common entities?
Put them in the most suitable existing package instead?

jonmv:
Yes, the package is odd now. It was OK when it was only ... validation. 
I think it should reside in vespajlib, as that's a very common dependency. Alternatively, it should be a new module without any other dependencies. Not that we have too few modules already ... 

I think the `Validation`, `StringWrapper` and `Name` could live in that package, though, and the `DomainName` and `Hostname` (and others we move/add later) should live elsewhere. How's that?

bratseth:
Sounds good to me.

vespajlib/src/main/java/ai/vespa/validation/Name.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.regex.Pattern;
+
+import static ai.vespa.validation.Validation.requireMatch;
+
+/**
+ * A name is a non-null, non-blank {@link String} which starts with a letter, and has up to
+ * 254 more characters which may be letters, numbers, dashes or underscores.
+ *
+ * Prefer domain-specific wrappers over this class, but prefer this over raw strings when possible. gT
```
mpolden, line 12: "gT"?

mpolden (commented):
vespajlib and preferring the most relevant existing package sounds good to me. There's already `com.yahoo.net.HostName`. `DomainName` fits well there too.

vespajlib/src/main/java/ai/vespa/validation/Name.java
jonmv: vim 4 life

jonmv:
Yes, that `HostName` is just utility methods ... but perhaps it can be both.

jonmv:
PTAL. I think we're ready to start trying to use this now.

mpolden (approved):
Good start!

PR #21899, opened 2022-03-30
Extend Mailer and Mail classes with support for templated messages

ean:

@mpolden

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/organization/TemplatedMail.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/organization/TemplatedMail.java
@@ -1,0 +1,46 @@
+package com.yahoo.vespa.hosted.controller.api.integration.organization;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * A message based on a configured template
+ *
+ * @author enygaard
+ */
+public class TemplatedMail implements Mail {
+
+    private final Collection<String> recipients;
+    private final String templateName;
+    private final Map<String, String> input;
+
+    public TemplatedMail(Collection<String> recipients, String templateName, Map<String, String> input) {
+        if (recipients.isEmpty())
+            throw new IllegalArgumentException("Empty recipient list is not allowed.");
+        recipients.forEach(Objects::requireNonNull);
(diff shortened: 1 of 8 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/organization/TemplatedMail.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * A message based on a configured template
+ *
+ * @author enygaard
+ */
+public class TemplatedMail implements Mail {
```
mpolden, line 15: Is there any particular reason we need to expose the templating through the `Mailer` / `Mail` types? It doesn't look like they can do any template substituion on their own so I think it would be better to keep `Mailer` / `Mail` as-is, and handle any templating on the calling side, i.e. in the code that builds a `Mail`.

ean: The current implementation uses the templating support that is part of AWS SES. That way we can fully decouple the mail content from the code, which allows us to easily update the emails independently of controller releases.

ean: Part of the discussion has been that we might want to do speedier changes of the content.

I am honestly not 100% sold on this setup, at least not the interface as it stands today. If we are going to go ahead supporting multiple mail providers I think we need to go with your approach, but I think that is not something that will give us any benefit long term. And that we should likely decide on one provider for all this moving forward. If that is AWS or something else is up to debate. But at the moment that is the quickest way to support it both in public and main.

If we decide to go with AWS using their templating system is for me the most obvious solution. It gives ut a well defined and documented template system which support both text and html.

As for testing and review, the content will live in git, and the automatically deployed. My main motivation here has been to reuse as much of the infrastructure as possible. And finding a flow that will work for people that don't live in Intellij all day.

mpolden: Ok. But if it we do it this way, it sounds like we can't guarantee that the template contents and variables made available by the code are in sync?

In any case I think I would simplify this to just adding `Map<String, String> templateValues` to `Mail` and leave it up to the implementation of the interface which fields it uses or not.

PR #21903, opened 2022-03-30
make tsan happy by using atomic operations on bits

havardpe:
... related to document existence

### DIFF
--- a/searchlib/src/vespa/searchlib/common/bitvector.h
+++ b/searchlib/src/vespa/searchlib/common/bitvector.h
@@ -5,6 +5,7 @@
 #include "bitword.h"
 #include <memory>
 #include <vespa/vespalib/util/alloc.h>
+#include <vespa/vespalib/util/atomic.h>
 #include <vespa/vespalib/util/generationholder.h>
 #include <vespa/fastos/types.h>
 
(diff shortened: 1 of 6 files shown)


baldersheim (approved):
Looks good, @toregge what say you.

toregge (approved):

searchlib/src/vespa/searchlib/common/bitvector.h
```diff
@@ -46,6 +47,11 @@ class BitVector : protected BitWord
     bool testBit(Index idx) const {
         return ((_words[wordNum(idx)] & mask(idx)) != 0);
     }
+    bool testBitSafe(Index idx) const {
+        auto my_words = vespalib::atomic::load_ref_acquire(_words);
+        auto my_word = vespalib::atomic::load_ref_acquire(my_words[wordNum(idx)]);
```
vekterli, line 52: Does this need to be an acquire? Since it's a primitive and we've already acquired the base pointer it should probably suffice with a relaxed load of the word itself

havardpe: It is to match the release fence used to ensure the meta-data for a document is visible when the bit indicating its existence is observed:

Fence-atomic synchronization
A release fence F in thread A synchronizes-with atomic [acquire operation](https://en.cppreference.com/w/cpp/atomic/memory_order) Y in thread B, if

there exists an atomic store X (with any memory order)
Y reads the value written by X (or the value would be written by [release sequence headed by X](https://en.cppreference.com/w/cpp/atomic/memory_order) if X were a release operation)
F is [sequenced-before](https://en.cppreference.com/w/cpp/atomic/memory_order) X in thread A
In this case, all non-atomic and relaxed atomic stores that are sequenced-before F in thread A will [happen-before](https://en.cppreference.com/w/cpp/atomic/memory_order) all non-atomic and relaxed atomic loads from the same locations made in thread B after Y.

vekterli: Ah of course, this bitvector can have transitive dependencies.

PR #21909, opened 2022-03-31
Handle previous app version being invalid

hmusum:
This should handle the case were we become stricter when validating application packages.


jonmv (approved):
Nice!

PR #21916, opened 2022-03-31
Skip test if onnx not available (as on arm64)

hmusum:
See https://github.com/vespa-engine/vespa/pull/21915


hmusum:
@aressem FYI

PR #21917, opened 2022-03-31
Improve error message for invalid config

jonmv:
@bratseth please review.

### DIFF
--- a/config-lib/src/main/java/com/yahoo/config/FileNode.java
+++ b/config-lib/src/main/java/com/yahoo/config/FileNode.java
@@ -1,6 +1,8 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.config;
 
+import java.nio.file.Path;
+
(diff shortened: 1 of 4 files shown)


config-lib/src/main/java/com/yahoo/config/FileNode.java
```diff
@@ -14,6 +16,8 @@ public FileNode() {
     public FileNode(String stringVal) {
         super(true);
         this.value = new FileReference(ReferenceNode.stripQuotes(stringVal));
+        if (Path.of(value.value()).normalize().startsWith(".."))
```
bratseth, line 19: Why not use vespajlib Path and you get this for free?

bratseth (approved):

config-lib/src/main/java/com/yahoo/config/FileNode.java
jonmv: This module didn't already depend on that. I guess I could have changed instead, but I opted for this.

PR #21918, opened 2022-03-31
Jonmv/misc 6

jonmv:
@mpolden please review.

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
@@ -12,22 +12,24 @@ import java.util.Objects;
  * @author vegard
  * @author bratseth
  */
-public final class ApplicationId implements Comparable<ApplicationId> {
+public class ApplicationId implements Comparable<ApplicationId> {
 
     private final TenantName tenant;
     private final ApplicationName application;
     private final InstanceName instance;
     private final String serializedForm;
 
-    public ApplicationId(ApplicationIdConfig config) {
-        this(TenantName.from(config.tenant()), ApplicationName.from(config.application()), InstanceName.from(config.instance()));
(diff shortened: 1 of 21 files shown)


jonmv:
I think we should cover existing application names with this, but I'm not entirely sure. 
If we think it's too sketchy, I can go through the code and fix the application-name nonsense first.

jonmv:
Is this too risky, @bratseth?

jonmv:
I can't fix the application name thingy. It's optionally used when preparing sessions. We use ruby tests name as application name in our system tests. I suppose others could use anything, but I think it's not very likely? We could just disallow the things that break serialisation, though, like `:` and `/` ...

config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
```diff
@@ -12,24 +12,26 @@
...
  */
-public final class ApplicationId implements Comparable<ApplicationId> {
+public class ApplicationId implements Comparable<ApplicationId> {
 
     private final TenantName tenant;
     private final ApplicationName application;
     private final InstanceName instance;
     private final String serializedForm;
 
-    public ApplicationId(ApplicationIdConfig config) {
```
bratseth, line 22: This API is used by old config models, so you need to be sure they are not using this. I think it is ok, but not sure it's worth the risk just to make final implicit?

bratseth (approved):
Great, I really think we should do this and that we don't have any actual applications breaking these rules.

config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
jonmv: I could only find two usages of this: a unit test, and the `NginxMetricsReporter` component.

PR #21933, opened 2022-04-01
Log stacktrace when aborting in vespamalloc.

baldersheim:
@vekterli PR

### DIFF
--- a/vespamalloc/src/vespamalloc/util/osmem.cpp
+++ b/vespamalloc/src/vespamalloc/util/osmem.cpp
@@ -18,7 +18,7 @@ Memory::Memory(size_t blockSize)
       _start(nullptr),
       _end(nullptr)
 { }
-Memory::~Memory() { }
+Memory::~Memory() = default;
 
 void *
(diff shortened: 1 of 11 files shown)


vespamalloc/src/vespamalloc/util/osmem.cpp
```diff
@@ -1,23 +1,33 @@
...
-#include <errno.h>
 
 namespace vespamalloc {
 
+Memory::Memory(size_t blockSize)
+    : _blockSize(std::max(blockSize, size_t(getpagesize()))),
+      _start(nullptr),
+      _end(nullptr)
+{ }
+Memory::~Memory() { }
```
vekterli, line 21: `= default` instead?

vekterli (approved):
:+1:

vespamalloc/src/vespamalloc/util/osmem.cpp
baldersheim: Fixed

PR #21952, opened 2022-04-02
Jonmv/http url

jonmv:
@mpolden or @bratseth please review.

### DIFF
--- a/container-core/src/main/java/com/yahoo/restapi/HttpURL.java
+++ b/container-core/src/main/java/com/yahoo/restapi/HttpURL.java
@@ -1,0 +1,352 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.restapi;
+
+import ai.vespa.validation.StringWrapper;
+import com.yahoo.net.DomainName;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.OptionalInt;
+import java.util.StringJoiner;
+import java.util.function.Function;
+
+import static ai.vespa.validation.Validation.require;
+import static ai.vespa.validation.Validation.requireInRange;
+import static java.net.URLDecoder.decode;
+import static java.net.URLEncoder.encode;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.unmodifiableMap;
+import static java.util.Objects.requireNonNull;
+import static java.util.function.Function.identity;
+
+/**
+ * This is the best class for creating, manipulating and inspecting HTTP URLs, because:
+ * <ul>
+ *     <li>It is more restrictive than {@link URI}, but with a richer construction API, reducing risk of blunder.
+ *         <ul>
+ *             <li>Scheme must be HTTP or HTTPS.</li>
+ *             <li>Authority must be a {@link DomainName}, with an optional port.</li>
+ *             <li>{@link Path} must be normalized at all times.</li>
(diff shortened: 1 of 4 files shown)


container-core/src/main/java/com/yahoo/restapi/HttpURL.java
```diff
@@ -0,0 +1,352 @@
...
+            return new Path<>(segments.subList(offset, segments.size()), trailingSlash, validator, inverse);
+        }
+
+        /** Returns a copy of this which eliminates the given number of segments, from the end up. */
+        public Path<T> headPath(int offset) {
+            return new Path<>(segments.subList(0, segments.size() - offset), trailingSlash, validator, inverse);
+        }
+
+        /** Returns a copy of this with the <em>decoded</em> segment appended at the end; it may not be either of {@code ""}, {@code "."} or {@code ".."}. */
+        public Path<T> with(T segment) {
```
bratseth, line 222: Missing validation of segment here?

bratseth (commented):
Great! I think we need lots more unit tests of this class to cover the API surface.

bratseth:
And do we really need the type argument for path segments?

container-core/src/main/java/com/yahoo/restapi/HttpURL.java
jonmv: Thanks! Yes, lost it when refactoring.

jonmv:
Isn't it nice that validating all segments is as easy as adding a constructor argument when creating the path? Also, all things you read out of it are already wrapped, if you want to send the wrapped strings further down. 
If you don't want to use it, you can simply ignore it too, as `<String>` is an option. 

Well, this was the main reason I wanted an early review! 

(And more unit tests are definitely needed.)

bratseth:
I guess some usage examples in tests would help.

jonmv:
The idea is that the `HttpURL` can be used both for parsing (and validating parsed content) of incoming `URI` objects, in handlers; and also be used to construct `URI`s for outgoing requests.

bratseth:
And why do segments have to be of multiple implementation types to do that?

jonmv:
Different APIs may have different levels of strictness? The /doc/v1 must accept path segments with `/` for example, but in general, we should not allow that.

jonmv:
We can't find a one-size-fits-all validation for these segments, so I thought the best thing was to make it as easy as possible to define your own validation.

bratseth:
I see. Wouldn't it be clearer to accept a validator then?

jonmv:
Isn't that exactly what's done here? `HttpURL.parse(uri, MyStrictSegment::from)` does that. But you don't want it to be part of the class signature? You would prefer its `url.path().segments()` to always return a `List<String>`, rather than `List<MyStrictSegment>`? But why? Wouldn't you want to pass on the wrapper that indicates what kind of string you're dealing with here? Also, when using it to construct a new URL, isn't it nice that you can see what kind of strings it will accept, if you're getting the `HttpURL` passed form somewhere else?

jonmv:
I think the best thing would be for API declarations to contain the validation for each segment, tbh., but that's more tricky. It's also not exclusive with this. This could be used to provide a baseline validation for "everything". We could also remove the `<String>` possibility, and have a more lax default, and then require an explicit `RawString` argument for allowing "any" string as segments/queries.

jonmv:
I can use this now in one of our handlers, as an example, perhaps?

bratseth:
> But you don't want it to be part of the class signature? You would prefer its url.path().segments() to always return a List<String>, rather than List<MyStrictSegment>?

Not prefer maybe, but it's what I would expect because I think of the path segments as short-lived and not worthy of having their own classes.

I you feel this is better then feel free to merge, but I think we should document it since I didn't get it.

jonmv:
Right, I understand that expectation. I'll try using it, and see if the type parameter is a blessing, or just a curse.

PR #21953, opened 2022-04-02
Use more Path

jonmv:
@bratseth or @baldersheim please review. This is #21924 without the `UnsupportedException`. The code path that triggers it was from a generated class :)

### DIFF
--- a/vespajlib/src/test/java/ai/vespa/validation/NameTest.java
+++ b/vespajlib/src/test/java/ai/vespa/validation/NameTest.java
@@ -4,6 +4,9 @@ package ai.vespa.validation;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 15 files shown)


vespajlib/src/test/java/ai/vespa/validation/NameTest.java
```diff
@@ -4,6 +4,9 @@
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.nio.file.Path;
```
baldersheim, line 7: These imports does not seem necessary.

baldersheim (approved):

vespajlib/src/test/java/ai/vespa/validation/NameTest.java
jonmv: Thankss.

PR #21963, opened 2022-04-04
Unify zip/tar archive reading [run-systemtest]

mpolden:


mpolden:
@jonmv or @bratseth 

Note that this limits the total size of extracted files in the application package to 2 GB (controller) and 8 GB (configserver), to protect against fun things like zip bombs. Are these OK defaults?

jonmv:
I don't think we need 2GB in the controller? We only extract meta data?

mpolden:
When calculating the bundle hash the controller reads all the files (to calculate checksum for every file).

mpolden:
I see I changed the bundle hash calculation to use `ArchiveStreamReader` directly, so yes, `ZipEntries` can probably have lower limit.

configserver/src/main/java/com/yahoo/vespa/config/server/application/CompressedApplicationInputStream.java
```diff
@@ -40,88 +43,71 @@
...
     /**
      * Close this stream.
      * @throws IOException if the stream could not be closed
      */
     public void close() throws IOException {
-        ais.close();
+        reader.close();
     }
 
     File decompress() throws IOException {
```
jonmv, line 68: Would be nice if this was cleaned up, rather than just added to .gitignore :)

mpolden: The gitignore change was unrelated to this. CompressedApplicationInputStreamTest doesn't always close though, is that what you mean?

jonmv (approved):
Nice!

configserver/src/main/java/com/yahoo/vespa/config/server/application/CompressedApplicationInputStream.java
jonmv: Ah, I noticed my /var/ under config-application-package, or something, was filled up after running these unit tests. I assumed that's why you added the ignore, and also because of this test.

jonmv:
Ah, no the bundle hash needs to iterate through all bytes, of course. Then shouldn't it be the same in config server and controller? And we just want a low per-file setting for meta data, in controller.

bratseth (approved):
Great! Should we limit the number of entries we read from the zip as well?

mpolden:
Yes, I can add that.

configserver/src/test/java/com/yahoo/vespa/config/server/application/CompressedApplicationInputStreamTest.java
```diff
@@ -63,9 +63,10 @@ private void assertTestApp(File outApp) {
     @Test
     public void require_that_valid_tar_application_can_be_unpacked() throws IOException {
         File outFile = createTarFile();
```
jonmv, line 65: Perhaps it's this file that should be deleted?

jonmv (approved):

PR #21969, opened 2022-04-04
Revert "Jonmv/reapply application name validation"

jonmv:
Reverts vespa-engine/vespa#21965

@hakonhall we need to allow longer names >_<


hakonhall (commented):
:+1:

PR #21973, opened 2022-04-04
Add dependencies used in this module explicitly.

gjoranv:
- Building for JDK 17, the Jackson deps were resolved to test or
  provided scope when not added explicitly.


bjorncs (approved):

gjoranv:
Looks like a bug. But this is correct anyway, as the deps were missing from the pom.

PR #21976, opened 2022-04-05
Use email as notification sink for newly created notifications

ean:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
@@ -49,7 +49,10 @@ public class MockMailer implements Mailer {
 
     /** Returns the list of mails sent to the given recipient. Modifications affect the set of mails stored in this. */
     public List<Mail> inbox(String recipient) {
-        return mails.get(recipient);
+        return mails.getOrDefault(recipient, List.of());
     }
 
+    public void reset() {
+        mails.clear();
+    }
 }
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
@@ -12,9 +12,25 @@ import java.util.Map;
 public class MockMailer implements Mailer {
 
     public final Map<String, List<Mail>> mails = new HashMap<>();
+    public final boolean blackhole;
+
+    public MockMailer() {
+        this(false);
+    }
+
+    MockMailer(boolean blackhole) {
+        this.blackhole = blackhole;
+    }
+
+    public static MockMailer blackhole() {
+        return new MockMailer(true);
+    }
 
     @Override
     public void send(Mail mail) {
+        if (blackhole) {
+            return;
+        }
         for (String recipient : mail.recipients()) {
             mails.putIfAbsent(recipient, new ArrayList<>());
             mails.get(recipient).add(mail);
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
@@ -1,0 +1,76 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.controller.notify;
+
+import com.yahoo.text.Text;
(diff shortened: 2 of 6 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
```diff
@@ -33,7 +33,13 @@ public String domain() {
 
     /** Returns the list of mails sent to the given recipient. Modifications affect the set of mails stored in this. */
     public List<Mail> inbox(String recipient) {
+        if (!mails.containsKey(recipient)) {
+            return List.of();
+        }
         return mails.get(recipient);
```
mpolden, line 39: Can be simplifed to `mails.getOrDefault(recipient, List.of())`.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/stubs/MockMailer.java
```diff
@@ -33,7 +33,13 @@ public String domain() {
...
     }
 
+    public void discard() {
+        mails.keySet().forEach(k -> mails.remove(k));
```
mpolden, line 43: Can be simplifed to `mails.clear()`. Consider naming the method `reset`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class Notifier {
+    private final CuratorDb curatorDb;
+    private final Mailer mailer;
+
+    private static final Logger log = Logger.getLogger(Notifier.class.getName());
+
+    public Notifier(CuratorDb curatorDb, Mailer mailer) {
+        this.curatorDb = curatorDb;
+        this.mailer = mailer;
```
mpolden, line 28: Nit: Assign to result of `Objects.requireNonNull`.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
```diff
@@ -0,0 +1,68 @@
...
+        });
+    }
+
+    private void dispatch(Notification notification, TenantContacts.Type type, Collection<? extends TenantContacts.Contact> contacts) {
+        switch (type) {
+            case EMAIL:
+                dispatch(notification, contacts.stream().map(c -> (TenantContacts.EmailContact) c).collect(Collectors.toList()));
+                break;
+            default:
+                log.fine("Unknown TenantContact Type: " + type);
```
mpolden, line 51: This should throw instead so that sending notifications for unknown contact types doesn't fail silently.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    private void dispatch(Notification notification, Collection<TenantContacts.EmailContact> contacts) {
+        mailer.send(mailOf(notification, contacts.stream().map(c -> c.email()).collect(Collectors.toList())));
+    }
+
+    private Mail mailOf(Notification n, Collection<String> recipients) {
+        var subject = new Formatter().format("[%s] Vespa Notification for %s", n.level().toString().toUpperCase(), n.type().name());
```
mpolden, line 60: Any reason to not use `Text.format`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
ean: I was a bit back and forth on that. Throwing here will also block a notification being persisted to zookeeper. Is that OK? We can alternatively move the dispatch to after the ZK transaction has been committed. But that again can lead to lost mails..

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
ean: Me not knowing about it. Changed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
mpolden: Sending can fail for other reasons too, i.e. the mailer service being unavailable. So it's better that sending the notification happens after the write and outside the write lock. Log an error if it fails.

If lost emails becomes a problem we should create a queue for them, and a separate maintainer that consumes it (and retries on failures).

mpolden (approved):

PR #21978, opened 2022-04-05
Deprecate load types from Java APIs [run-systemtest]

vekterli:
@jonmv please review. Load type tentacles reached further than I first thought 🦑
@gjoranv FYI

Load types have not been used in practice for years, and supporting
them in backend metrics etc. has long since been lacking. Prepare for
removing these on Vespa 8.

Most callsites are unchanged, aside from presumed safe changes such
as constructors used by dependency injection. Have added new overloads
without load types where these did not already exist to allow for
an orderly transition.


jonmv (approved):
Flames

PR #21981, opened 2022-04-05
Use a primitive to see if that makes the JIT compiler more predictable.

baldersheim:
@jonmv PR

### DIFF
--- a/document/src/main/java/com/yahoo/document/idstring/IdString.java
+++ b/document/src/main/java/com/yahoo/document/idstring/IdString.java
@@ -79,7 +79,7 @@ public abstract class IdString {
     }
 
     private static void validateTextString(String id) {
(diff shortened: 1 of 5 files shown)


jonmv (approved):

jonmv:
Need ABI spec update.

jonmv (approved):

document/src/main/java/com/yahoo/document/idstring/IdString.java
```diff
@@ -79,7 +79,7 @@ public static IdString createFromSerialized(String id) {
     }
 
     private static void validateTextString(String id) {
-        if (Text.isValidTextString(id)) {
+        if ( ! Text.isValidTextString(id)) {
```
jonmv, line 82: You got it right half the places!

PR #21982, opened 2022-04-05
Accept default values for URL types in config

lesters:
@hmusum For discussion.


hmusum (approved):
Looks good to me. The value is still a `File`, so I guess // will be transformed to /, but we can look into that later

hmusum (approved):
👍

hmusum:
Looks good

PR #21984, opened 2022-04-05
Address some more UBSan warnings

vekterli:
@baldersheim please review
@arnej27959 @havardpe FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/util/array.hpp
+++ b/vespalib/src/vespa/vespalib/util/array.hpp
@@ -60,7 +60,9 @@ Array<T>::Array(const Array & rhs)
     : _array(rhs._array.create(rhs.size() * sizeof(T))),
       _sz(rhs.size())
 {
-    construct(array(0), rhs.array(0), _sz, std::is_trivially_copyable<T>());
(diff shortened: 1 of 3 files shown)


vespalib/src/vespa/vespalib/util/array.hpp
```diff
@@ -60,7 +60,9 @@ Array<T>::Array(const Array & rhs)
     : _array(rhs._array.create(rhs.size() * sizeof(T))),
       _sz(rhs.size())
 {
-    construct(array(0), rhs.array(0), _sz, std::is_trivially_copyable<T>());
+    if (_sz > 0) [[likely]] {
```
baldersheim, line 63: Standard ? Replacement for __builtin_expect..... ?

baldersheim (approved):

vespalib/src/vespa/vespalib/util/array.hpp
vekterli: Yep, as of C++20 https://en.cppreference.com/w/cpp/language/attributes/likely

PR #21986, opened 2022-04-05
Upgrade dependencies and improve error message on unexpected artifact

bjorncs:
FYI @gjoranv

### DIFF
--- a/bundle-plugin/pom.xml
+++ b/bundle-plugin/pom.xml
@@ -21,12 +21,12 @@
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-plugin-api</artifactId>
-            <version>3.5.0</version>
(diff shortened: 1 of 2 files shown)


jonmv (approved):

bundle-plugin/pom.xml
```diff
@@ -21,12 +21,12 @@
         <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-plugin-api</artifactId>
-            <version>3.5.0</version>
+            <version>3.8.5</version>
```
gjoranv, line 24: Any specific reason why this was upgraded?

bjorncs: We were uncertain whether the observed behaviour was a bug in the Maven plugin api.

PR #21995, opened 2022-04-06
Start using new Curator locks in controller

hmusum:
One should not use locks with overlapping paths, e.g.

    /controller/v1/locks/<tenant>
    /controller/v1/locks/<tenant>/<application>
    /controller/v1/locks/<tenant>/<application>/<instance>

Change to use new paths, e.g.:

    /controller/v1/locks/<tenant>
    /controller/v1/locks/<tenant:application>
    /controller/v1/locks/<tenant:application:instance>

Need to take both locks in a transition period using
new class MultiplePathsLock, which can be removed when
all users use both new and old paths for locks (remove and just
use the new lock paths)

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
@@ -713,7 +713,7 @@ public class CuratorDb {
     }
 
     private Path lockPath(ApplicationId instance, ZoneId zone) {
-        return lockRoot.append(instance.serializedForm() + ":" + zone.region().value());
+        return lockRoot.append(instance.serializedForm() + ":" + zone.environment().value() + ":" + zone.region().value());
     }
 
     private Path lockPath(ApplicationId instance, JobType type) {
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -667,32 +682,48 @@ private Path lockPath(TenantName tenant) {
...
+        return legacyLockPath(instance, type)
                 .append(step.name());
     }
 
+    private Path lockPath(TenantAndApplicationId application) {
+        return lockRoot.append(application.tenant().value() + ":" + application.application().value());
+    }
+
+    private Path lockPath(ApplicationId instance, ZoneId zone) {
+        return lockRoot.append(instance.serializedForm() + ":" + zone.region().value());
```
jonmv, line 716: This should be both region and environment, I guess?

jonmv (commented):
LGTM, except that zone lock are missing environment!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
jonmv:
```suggestion
        return lockRoot.append(instance.serializedForm() + ":" + zone.environment().value() + ":" + zone.region().value());
```

PR #21998, opened 2022-04-06
Refactor to optimize for hot path.

baldersheim:
No reason to count characters when you already now there can only be one.
Nor is there any reason to check for surrogates when there can be none.

@jonmv and @bratseth PR
This gains 20% in benchmark, and we can hope that it will make JIT compilaton more predictable.

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/text/Text.java
+++ b/vespajlib/src/main/java/com/yahoo/text/Text.java
@@ -86,9 +86,7 @@ public final class Text {
         if (codepoint <  0xFFFFE)  return true;
         if (codepoint <= 0xFFFFF)  return false;
         if (codepoint <  0x10FFFE) return true;
-        if (codepoint <= 0x10FFFF) return false;
-
-        return true;
+        return false;
     }
 
     /**
@@ -128,7 +126,6 @@ public final class Text {
                 i++;
             } else {
                 if ( ! isTextCharAboveUsAscii(codePoint)) return false;
-                if ( ! Character.isValidCodePoint(codePoint)) return false;
                 if ( ! Character.isSupplementaryCodePoint(codePoint)) {
                     if (Character.isHighSurrogate((char)codePoint)) return false;
                     i++;
--- a/vespajlib/src/main/java/com/yahoo/text/Text.java
+++ b/vespajlib/src/main/java/com/yahoo/text/Text.java
@@ -118,18 +118,30 @@ public final class Text {
      * Validates that the given string value only contains text characters.
      */
     public static boolean isValidTextString(String string) {
-        for (int i = 0; i < string.length(); ) {
+        int length = string.length();
+        for (int i = 0; i < length; ) {
             int codePoint = string.codePointAt(i);
-            if ( ! Text.isTextCharacter(codePoint)) return false;
-
-            int charCount = Character.charCount(codePoint);
-            if (Character.isHighSurrogate(string.charAt(i))) {
-                if ( (charCount == 1) || !Character.isLowSurrogate(string.charAt(i+1))) return false;
+            if (codePoint < 0x80) {
+                if ( ! allowedAsciiChars[codePoint]) return false;
+                i++;
+            } else if (codePoint < Character.MIN_SURROGATE) {
+                i++;
+            } else {
+                if ( ! isTextCharAboveUsAscii(codePoint)) return false;
+                if ( ! Character.isValidCodePoint(codePoint)) return false;
+                if ( ! Character.isSupplementaryCodePoint(codePoint)) {
+                    if (Character.isHighSurrogate((char)codePoint)) return false;
+                    i++;
+                } else {
+                    if (Character.isHighSurrogate(Character.highSurrogate(codePoint))
+                        && ! Character.isLowSurrogate(Character.lowSurrogate(codePoint))) return false;
+                    i += 2;
+                }
             }
-            i += charCount;
         }
         return true;
     }
+
 
     /** Returns whether the given code point is displayable. */
     public static boolean isDisplayable(int codePoint) {
--- a/vespajlib/src/main/java/com/yahoo/text/Text.java
+++ b/vespajlib/src/main/java/com/yahoo/text/Text.java
@@ -50,9 +50,10 @@ public final class Text {
 
         return (codepoint < 0x80)
                 ? allowedAsciiChars[codepoint]
-                : isTextCharAboveUsAscii(codepoint);
+                : (codepoint <  Character.MIN_SURROGATE) || isTextCharAboveUsAscii(codepoint);
     }
     private static boolean isTextCharAboveUsAscii(int codepoint) {
+        if (codepoint <= Character.MAX_HIGH_SURROGATE) return false;
         if (codepoint <  0xFDD0)   return true;
         if (codepoint <= 0xFDDF)   return false;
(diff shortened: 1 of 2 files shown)


vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -118,19 +118,31 @@ public static OptionalInt validateTextString(String string) {
...
-            int charCount = Character.charCount(codePoint);
-            if (Character.isHighSurrogate(string.charAt(i))) {
-                if ( (charCount == 1) || !Character.isLowSurrogate(string.charAt(i+1))) return false;
+            if (codePoint < 0x80) {
+                if ( ! allowedAsciiChars[codePoint]) return false;
+                i++;
+            } else if (codePoint < Character.MIN_SURROGATE) {
+                i++;
+            } else {
+                if ( ! isTextCharAboveUsAscii(codePoint)) return false;
```
jonmv, line 130: Can't this return false instead of true, when the code point is out of range? 

That would save the check for isValidCodePoint.

And ... I'd assume this method is intended to do exactly what you do with the `isHighSurrogate(highSurrogate(codePoint))`?

jonmv: (That may of course not be the case ... but what is that method used for, if this isn't the case?)

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -118,19 +118,31 @@ public static OptionalInt validateTextString(String string) {
...
+                if ( ! Character.isValidCodePoint(codePoint)) return false;
+                if ( ! Character.isSupplementaryCodePoint(codePoint)) {
+                    if (Character.isHighSurrogate((char)codePoint)) return false;
+                    i++;
+                } else {
+                    if (Character.isHighSurrogate(Character.highSurrogate(codePoint))
+                        && ! Character.isLowSurrogate(Character.lowSurrogate(codePoint))) return false;
```
jonmv, line 137: From javadoc: ```If isSupplementaryCodePoint(x) is true, then isHighSurrogate(highSurrogate(x)) and toCodePoint(highSurrogate(x), lowSurrogate(x)) == x are also always true.```

jonmv:
Hmm, it looks to me like all the checking could just be replaced by using the `isTextCharAboveUsAscii`, if that returns `false` when none of the if clauses are true?

vespajlib/src/main/java/com/yahoo/text/Text.java
baldersheim: Agree that isValidCodePoint can be removed when inverting the last return value in isTextCharacter.

vespajlib/src/main/java/com/yahoo/text/Text.java
baldersheim: So that means that surrogate check in the else can be skipped ?

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -118,19 +118,31 @@ public static OptionalInt validateTextString(String string) {
      * Validates that the given string value only contains text characters.
      */
     public static boolean isValidTextString(String string) {
-        for (int i = 0; i < string.length(); ) {
+        int length = string.length();
```
bratseth, line 119: This doesn't gain anything?

bratseth (approved):
Nice

vespajlib/src/main/java/com/yahoo/text/Text.java
baldersheim: No, not in benchmark. Change is intentional to see if it effects inconsistent JIT compilation, which sometimes leaves
all the String.functions here non inlined. Here, even non-optimized we should see it being called fewer times than codePointAt().

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -50,9 +50,10 @@ public static boolean isTextCharacter(int codepoint) {
 
         return (codepoint < 0x80)
                 ? allowedAsciiChars[codepoint]
-                : isTextCharAboveUsAscii(codepoint);
+                : (codepoint <  Character.MIN_SURROGATE) || isTextCharAboveUsAscii(codepoint);
     }
     private static boolean isTextCharAboveUsAscii(int codepoint) {
+        if (codepoint <= Character.MAX_HIGH_SURROGATE) return false;
```
jonmv, line 56: That should be >=, but it's implicit, when the last if clause (below) has failed?

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -118,19 +117,24 @@ public static OptionalInt validateTextString(String string) {
...
+        for (int i = 0; i < length; ) {
             int codePoint = string.codePointAt(i);
-            if ( ! Text.isTextCharacter(codePoint)) return false;
-
-            int charCount = Character.charCount(codePoint);
-            if (Character.isHighSurrogate(string.charAt(i))) {
-                if ( (charCount == 1) || !Character.isLowSurrogate(string.charAt(i+1))) return false;
+            if (codePoint < 0x80) {
+                if ( ! allowedAsciiChars[codePoint]) return false;
+            } else if (codePoint < Character.MIN_SURROGATE) {
```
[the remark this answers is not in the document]
jonmv: Well, I suppose it's possible that branching to a different method on this if test could be beneficial for the JIT? You know this better than me :)

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -50,9 +50,10 @@ public static boolean isTextCharacter(int codepoint) {
 
         return (codepoint < 0x80)
                 ? allowedAsciiChars[codepoint]
-                : isTextCharAboveUsAscii(codepoint);
+                : (codepoint <  Character.MIN_SURROGATE) || isTextCharAboveUsAscii(codepoint);
```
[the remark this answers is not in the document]
baldersheim: Well it is now a sthe below check is MAX_HIGH_SURROGATE.

vespajlib/src/main/java/com/yahoo/text/Text.java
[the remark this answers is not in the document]
baldersheim: No currently.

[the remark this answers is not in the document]
baldersheim: The main reason for branching here is that you then do not need the isBmp check below to check if you need an extra i++ in the range [0x80, 0xD800>

vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -50,9 +50,10 @@ public static boolean isTextCharacter(int codepoint) {
 
         return (codepoint < 0x80)
                 ? allowedAsciiChars[codepoint]
-                : isTextCharAboveUsAscii(codepoint);
+                : (codepoint <  Character.MIN_SURROGATE) || isTextCharAboveMinSurrogate(codepoint);
```
jonmv, line 53: Doesn't this now fail all leading surrogates? They're `> MIN_SURROGATE` and `< MAX_HIGH_SURROGATE`, but what you _want_ to check is that it's either not a leading surrogate, or that it is a leading surrogate followed by a trailing one.

PR #22000, opened 2022-04-06
Use a Q with limited capacity for warmup executor to avoid unlimited …

baldersheim:
…buildup with swamped disk.

@geirst or @vekterli PR


geirst (approved):
👍

vekterli (approved):
:+1:

PR #22003, opened 2022-04-06
Arnej/rank filter in struct field

arnej27959:

@geirst please review
@jobergum FYI


config-model/src/main/java/com/yahoo/searchdefinition/derived/AttributeFields.java
```diff
@@ -134,6 +134,11 @@ private void deriveAttributeAsArrayType(ImmutableSDField field) {
         }
         Attribute attribute = field.getAttributes().get(field.getName());
         if (attribute != null) {
+            Ranking ranking = field.getRanking();
```
geirst, line 137: I suggest creating a helper function for this (e.g. applyRankingSettings() / considerApplyRankingSettings()) and use that in deriveAttribute() as well to ensure we handle this the same way in both code paths.

geirst (approved):
👍

arnej27959:
note: this only works with the new SD parser, we don't know for sure when that is enabled everywhere.

geirst (approved):
👍

PR #22007, opened 2022-04-06
Don't log that distribution key is redundantly present when it's actually not

vekterli:
@geirst please review.

Would log that `distribution-key` attribute was present even if just `name` attribute was present.


geirst (approved):
👍

PR #22010, opened 2022-04-06
Jonmv/more http url

jonmv:
@bratseth please review. First commits are from #21999


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -1472,7 +1472,7 @@ DocumentId id() {
             return new DocumentId("id:" + requireNonNull(path.get("namespace")) +
                                   ":" + requireNonNull(path.get("documentType")) +
                                   ":" + group.map(Group::docIdPart).orElse("") +
-                                  ":" + requireNonNull(path.getRest()));
+                                  ":" + String.join("/", requireNonNull(path.getRest()).segments())); // :'(
```
bratseth, line 1475: image

vespajlib/src/main/java/com/yahoo/path/Path.java
```diff
@@ -41,6 +41,11 @@ private Path(List<String> elements, String delimiter) {
         this.delimiter = delimiter;
     }
 
+    /** Creates a new path with the given segments. */
+    public static Path from(List<String> segments) {
```
bratseth, line 45: A ... signature would be nice I guess

bratseth (approved):
This is pretty great now

vespajlib/src/main/java/com/yahoo/path/Path.java
jonmv: 😄 😭

PR #22012, opened 2022-04-06
Extract some patterns from existing code

jonmv:
@baldersheim not sure this is any less enigmatic than the previous, but the `validateTextString` is twice as fast for me. The `isValidTextString` is the same, at ~3.6s.


baldersheim:
On openjdk 17.0.2 this lost the gain. I brought the & 0xffff further in #22013.
Suggest closing this one.

PR #22013, opened 2022-04-06
Add simplification of long if based on tip from @jonmv.

baldersheim:
@jonmv PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/text/Text.java
+++ b/vespajlib/src/main/java/com/yahoo/text/Text.java
@@ -56,6 +56,7 @@ public final class Text {
         if (codepoint <= Character.MAX_HIGH_SURROGATE) return false;
         if (codepoint <  0xFDD0)   return true;
         if (codepoint <= 0xFDDF)   return false;
+        if (codepoint <  0x10000)  return true;
         if (codepoint >= 0x10FFFE) return false;
         return (codepoint & 0xffff) < 0xFFFE;
     }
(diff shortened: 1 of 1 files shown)


vespajlib/src/main/java/com/yahoo/text/Text.java
```diff
@@ -56,38 +56,8 @@ private static boolean isTextCharAboveMinSurrogate(int codepoint) {
...
-        if (codepoint <= 0xCFFFF)  return false;
-        if (codepoint <  0xDFFFE)  return true;
-        if (codepoint <= 0xDFFFF)  return false;
-        if (codepoint <  0xEFFFE)  return true;
-        if (codepoint <= 0xEFFFF)  return false;
-        if (codepoint <  0xFFFFE)  return true;
-        if (codepoint <= 0xFFFFF)  return false;
-        if (codepoint <  0x10FFFE) return true;
-        return false;
+        if (codepoint >= 0x10FFFE) return false;
```
jonmv, line 59:
```suggestion
        if (codepoint < 0x10000) return false;
        if (codepoint >= 0x10FFFE) return false;
```

jonmv: Like so

baldersheim: Correct, but it should return true.... I guess

jonmv: Haha, yes.

jonmv (approved):

jonmv:
Do you have any idea why we're considering all these FFFE and FFFF values invalid, though?

baldersheim:
I have no idea. I assume someone has read some formal specs. There are some links in the code.

PR #22021, opened 2022-04-07
Jonmv/http url in configserver client

jonmv:
@bratseth please review and merge.


vespajlib/src/main/java/ai/vespa/http/DomainName.java
```diff
@@ -15,8 +15,8 @@
  */
 public class DomainName extends PatternedStringWrapper<DomainName> {
 
-    static final Pattern labelPattern = Pattern.compile("([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]{0,61}[A-Za-z0-9])");
-    static final Pattern domainNamePattern = Pattern.compile("(" + labelPattern + "\\.)*" + labelPattern);
+    public static final Pattern labelPattern = Pattern.compile("([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]{0,61}[A-Za-z0-9])");
```
bratseth, line 18: Doesn't look like these need to be public?

bratseth (approved):
poetic and/or correct

bratseth:
HttpConfigServerClientTest.testRetries:61 Unexpected exception type thrown ==> expected: <java.io.UncheckedIOException> but was: <java.lang.IllegalArgumentException>

vespajlib/src/main/java/ai/vespa/http/DomainName.java
jonmv: Not the label. The domain pattern is used by `HostName`. It's probably better to just replicate the pattern. Will do that.

PR #22022, opened 2022-04-07
Add flag for using new controller locks

hmusum:


jonmv (commented):
Good idea. We could also have a trinary value of this flag, to allow quick switching everywhere.

hmusum:
Good idea, PTAL

jonmv (approved):

PR #22027, opened 2022-04-07
Support new lock paths for application operations in controller

hmusum:


jonmv (approved):
Yeah, the inheritance probably caused some issues ... :)

PR #22031, opened 2022-04-07
Add more detailed visit failure metrics.

yngveaasheim:

@vekterli Please review.

Note: 
Some of the added metrics might not be relevant for visit operations, but are still output from the distributors. We should likely remove those from the PR before merging.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
@@ -855,6 +855,17 @@ public class VespaMetricSet {
         metrics.add(new Metric("vds.distributor.visitor.sum.latency.average")); // TODO: Remove in Vespa 8
         metrics.add(new Metric("vds.distributor.visitor.sum.ok.rate"));
         metrics.add(new Metric("vds.distributor.visitor.sum.failures.total.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.notready.rate"));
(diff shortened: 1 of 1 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
```diff
@@ -855,6 +855,17 @@ private static void addSearchNodeExecutorMetrics(Set<Metric> metrics, String pre
...
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.notconnected.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.wrongdistributor.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.safe_time_not_reached.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.storagefailure.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.timeout.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.busy.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.inconsistent_bucket.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.notfound.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.concurrent_mutations.rate"));
+        metrics.add(new Metric("vds.distributor.visitor.sum.failures.test_and_set_failed.rate"));
```
vekterli, line 868: `concurrent_mutations` and `test_and_set_failed` metrics do not need to be included, as they cannot happen for read-only operations

vekterli (approved):
:+1:

PR #22038, opened 2022-04-07
Arnej/allow struct types with same name

arnej27959:

@bratseth and @baldersheim please review


bratseth (commented):
Finally there! But are we changing ids of struct definitions that works today here?

arnej27959:
> But are we changing ids of struct definitions that works today here?

yes, this is alternative "A": change name (and id) of user-defined struct types so they can be registered in DocumentTypeManager simultaneously.

if feels like we can just go with alternative "B" which is to stop registering struct types in the DocumentTypeManager at all;  I'm going to test that tomorrow and see what the fallout is.

arnej27959:
now: this is alternative "B" and it passes some struct-related system tests; if there are no objections I'll merge this once factory is green.

baldersheim (approved):

bratseth:
But we're still changing ids of structs that can be deployed today - doesn't that mean that we'll be unable to read existing data stored in such structs once this version rolls out?

arnej27959:
Re: changing IDs of structs - these aren't serialized, instead the reader knows from config which struct type to use for deserializing.  Only field IDs are truly critical; these are calculated using a hash from field name and type ID, but can also be overridden when creating fields.

Anyway, now on alternative "B" we keep the name and ID of the struct types but stop registering them in (global) maps, so we can have different types with same name and ID.  This feels like a safer change, even if the user-visible changes are (in theory) bigger.

bratseth:
I don't know how structs are deserialized so if you're sure you know the consequences here feel free to merge, but I notice there's still a lot of changes to ids in documentmanager.cfg's in this PR which seems to contradict your second paragraph.

baldersheim:
Arne is correct struct ids has no effect on serialisation. Only name is used.

baldersheim (approved):

arnej27959:
Sorry about the confusion, the changes in documentmanager.cfg are only for "idx" numbers (those counting from 10000) which aren't type IDs at all, just indexes for referencing between types in that config.

arnej27959:
>  Arne is correct struct ids has no effect on serialisation. Only name is used.

actually name isn't used either - only the contents of the struct type (the fields) really matter.

PR #22044, opened 2022-04-07
Bratseth/flatten totalhits

bratseth:


searchlib/src/main/java/com/yahoo/searchlib/aggregation/Grouping.java
```diff
@@ -184,11 +182,10 @@ public List<GroupingLevel> getLevels() {
      *
      * @param level the level to add
      * @return this, to allow chaining
-     * @throws NullPointerException If <code>level</code> argument is null.
+     * @throws NullPointerException ff <code>level</code> argument is null
```
bjorncs, line 185:
```suggestion
     * @throws NullPointerException if <code>level</code> argument is null
```

bjorncs (approved):
The group count is approximate, but it should be exact for <256 groups

PR #22049, opened 2022-04-08
Jonmv/unify hostname classes

jonmv:
@freva please review.


freva (approved):

jonmv:
That looks likely, indeed.

PR #22073, opened 2022-04-09
Trim down search::multivalue::Value and search::multivalue::WeightedValue.

toregge:
Use traits to check for presence of a weight and to extract inner type.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #22074, opened 2022-04-09
Jonmv reapply hostname verification

jonmv:
@hmusum or @bratseth please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/Endpoint.java
```diff
@@ -591,8 +590,7 @@ public Endpoint in(SystemName system) {
                                 Objects.requireNonNull(scope, "scope must be non-null"),
                                 Objects.requireNonNull(system, "system must be non-null"),
                                 Objects.requireNonNull(port, "port must be non-null"),
-                                legacy,
-                                Objects.requireNonNull(routingMethod, "routingMethod must be non-null"));
```
jonmv, line 595: non-nullity asserted already, some lines up

toregge (approved):

PR #22076, opened 2022-04-09
Jonmv/deployment risk

jonmv:
@freva or @mpolden please review. 

This is in preparation for slimming down the ApplicationVersion as a key, to just the RevisionId. Then, we can put more stuff in ApplicationVersion, such as commit messages and other metadata, and whether the user has decided to cancel a revision.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
@@ -53,7 +53,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
 import java.util.logging.Level;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.yahoo.collections.Iterables.reversed;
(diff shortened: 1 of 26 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -422,19 +422,15 @@ public void finish(RunId id) throws TimeoutException {
                 logs.flush(id);
                 metric.jobFinished(run.id().job(), finishedRun.status());
 
-                DeploymentId deploymentId = new DeploymentId(unlockedRun.id().application(), unlockedRun.id().job().type().zone(controller.system()));
```
jonmv, line 421: Ah, never mind this commit—I read that boolean inverted. This was correct after all. This was fixed later, when I rewrite this to also update the revision history.

mpolden (approved):
Nice cleanup!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/RevisionHistory.java
```diff
@@ -0,0 +1,147 @@
...
+    public RevisionHistory with(ApplicationVersion revision) {
+        NavigableMap<RevisionId, ApplicationVersion> production = new TreeMap<>(this.production);
+        production.put(revision.id(), revision);
+        return new RevisionHistory(production, development);
+    }
+
+    /** Returns a copy of this with the new development revision added, and the previous version without a package. */
+    public RevisionHistory with(ApplicationVersion revision, JobId job) {
+        NavigableMap<JobId, NavigableMap<RevisionId, ApplicationVersion>> development = new TreeMap<>(this.development);
+        NavigableMap<RevisionId, ApplicationVersion> revisions = development.computeIfAbsent(job, __ -> new TreeMap<>());
```
freva, line 83: By returning existing here we will modify the original instance

jonmv: Yes, thanks, forgot a copy here!

PR #22078, opened 2022-04-11
Remove viper

mpolden:
Cobra no longer depends on Viper since 1.4.0. By implementing our own config
instead, we can get rid of Viper altogether and remove some hacks made necessary
by that library.

Bonus: Removes the large dependency tree pulled in by Viper, reducing the binary
size by ~1 MB.

@bratseth


bratseth (approved):
Great!

PR #22080, opened 2022-04-11
Use IMultiValueReadView in dot product feature instead of getRawValues().

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #22081, opened 2022-04-11
- Allow control of throttlepolicy per session.

baldersheim:
- Move tests to same namespace as code to avoid polluting the public interface.
- Expose the throttlepolicy from SourceSession to avoid double bookeeping.

@bratseth and @jonmv PR
@vekterli FYI


documentapi/src/main/java/com/yahoo/documentapi/Parameters.java
```diff
@@ -9,5 +13,12 @@
  * @author bratseth
  */
 public class Parameters {
-    // empty
+    ThrottlePolicy throttlePolicyOverride;
```
bratseth, line 16: I suggest removing "override" from the names. It doesn't add anything.

jonmv (approved):
Nice!

bratseth (approved):
👍

bratseth:
00:03:10 [INFO] --- abi-check-plugin:7-SNAPSHOT:abicheck (default) @ documentapi ---
00:03:10 [ERROR] Class com.yahoo.documentapi.Parameters: Extra method public void setThrottlePolicyOverride(com.yahoo.messagebus.ThrottlePolicy)
00:03:10 [ERROR] Class com.yahoo.documentapi.Parameters: Extra method public java.util.Optional getThrottlePolicyOverride()

documentapi/src/main/java/com/yahoo/documentapi/Parameters.java
baldersheim: Did so, and that revealed use in subclass VisitorParameters.
I am not sure about binary compatibility. Does moving an interface to parent brake binary compatibility ?

bratseth: No, that is a compatible change.

PR #22083, opened 2022-04-11
Use IWeightedSetEnumReadView in dot product feature instead of IWeightedIndexVector.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #22086, opened 2022-04-11
Add config unset command

mpolden:
This allows the user to unset/reset an option in their global or local config,
and use the default value instead. Wanted to add this for a while, but Viper
made it unnecessarily hard.

Example:

```
# user has a bunch of local options
$ vespa config get --local
application = vespa.canary-pipeline.mpolden
target = hosted
zone = perf.us-east-3

# reset zone and let the system choose it
$ vespa config unset --local zone
$ vespa config get zone
zone = <unset>
```

@bratseth

### DIFF
--- a/client/go/cmd/config.go
+++ b/client/go/cmd/config.go
@@ -153,6 +153,42 @@ $ vespa config set --local wait 600
 	return cmd
 }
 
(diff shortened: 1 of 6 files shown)


bratseth (approved):
👍 👍

PR #22095, opened 2022-04-11
Jonmv/application version to revision

jonmv:
@freva


jonmv:
Only the last commit is new

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/ApplicationSerializer.java
```diff
@@ -338,21 +347,20 @@ private Application fromSlime(Slime slime) {
...
-        List<ApplicationVersion> revisions = prodVersionsArray.valid() ? revisionsFromSlime(prodVersionsArray)
-                                                                       : revisionsFromSlime(versionsArray);
+    private RevisionHistory revisionsFromSlime(Inspector prodVersionsArray, Inspector devVersionsArray, TenantAndApplicationId id) {
+        List<ApplicationVersion> revisions = revisionsFromSlime(prodVersionsArray, null);
         Map<JobId, List<ApplicationVersion>> devRevisions = new HashMap<>();
-        devVersionsArray.traverse((ArrayTraverser) (__, devRevisionsObject) ->
-                devRevisions.put(jobIdFromSlime(id, devRevisionsObject), revisionsFromSlime(devRevisionsObject.field(versionsField))));
+        devVersionsArray.traverse((ArrayTraverser) (__, devRevisionsObject) -> {
+            JobId job = jobIdFromSlime(id, devRevisionsObject);
+            devRevisions.put(jobIdFromSlime(id, devRevisionsObject), revisionsFromSlime(devRevisionsObject.field(versionsField), job));
```
freva, line 362:
```suggestion
            devRevisions.put(job, revisionsFromSlime(devRevisionsObject.field(versionsField), job));
```

jonmv:
I can try to summarise its changes, though, as I was unable to split it:

* All use of `ApplicationVersion` as keys was replaced with `RevisionId`. 
* Each `ApplicationVersion` holds a `RevisionId`, which is either a production-id, or holds an additional `JobId` to point to the job where it was deployed. 
* The `JobId` in `RevisionId` is inferred from serialisation context. We _can_ change to an explicit format later, but for now it must be inferred. There are some assumptions here, such as no dev builds ever being in a `Change`, or in the production part of `RevisionHistory`, but I think both of those were guarded by assertions already. 
* For now, all serialisation of `RevisionId` is augmented to full `ApplicationVersion` data, to avoid data loss during upgrade. This will be removed on the next release.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/ContainerTester.java
```diff
@@ -38,7 +38,7 @@
  */
 public class ContainerTester {
 
-    private static final boolean writeResponses = false;
+    private static final boolean writeResponses = true;
```
freva, line 41:
```suggestion
    private static final boolean writeResponses = false;
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2546,14 +2543,16 @@ private long asLong(String valueOrNull, long defaultWhenNull) {
         }
     }
 
+    /*
     private void toSlime(Run run, Cursor object) {
```
freva, line 2547: Remove?

jonmv: Sure, will do later. Wasn't sure it was unused, because things didn't compile at the time XD

jonmv: Done

freva (approved):

PR #22096, opened 2022-04-11
Change search::multivalue::Value from class to using statement.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #22098, opened 2022-04-12
Optimize for the leaf groups that has no children since there are mos…

baldersheim:
…t of them.

@jonmv or @bratseth or @bjorncs PR

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/grouping/vespa/ResultBuilder.java
+++ b/container-search/src/main/java/com/yahoo/search/grouping/vespa/ResultBuilder.java
@@ -143,9 +143,9 @@ class ResultBuilder {
     }
 
     private class GroupBuilder {
-
+        private static final int CHILDLIST_SIZE_INCREMENTS = 4;
         boolean [] results = new boolean[8];
-        GroupListBuilder [] childLists = new GroupListBuilder[8];
+        GroupListBuilder [] childLists;
         int childCount = 0;
         final ResultId resultId;
         final com.yahoo.searchlib.aggregation.Group group;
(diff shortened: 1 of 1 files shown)


container-search/src/main/java/com/yahoo/search/grouping/vespa/ResultBuilder.java
```diff
@@ -173,18 +173,23 @@ Group fill(Group group) {
...
+                    if (child != null) {
+                        group.add(child.build());
+                    }
                 }
             }
             return group;
         }
 
         GroupListBuilder getOrCreateChildList(int tag, boolean ranked) {
             int index = tag + 1; // Add 1 to avoid the dreaded -1 default value.
```
jonmv, line 187: (((＼（✘෴✘）／)))

container-search/src/main/java/com/yahoo/search/grouping/vespa/ResultBuilder.java
```diff
@@ -173,18 +173,23 @@ Group fill(Group group) {
...
-            if (index >= childLists.length) {
-                childLists = Arrays.copyOf(childLists, tag + 8);
+            if (childLists == null || index >= childLists.length) {
+                int reservedSize = (((index + 1) + (CHILDLIST_SIZE_INCREMENTS -1))/CHILDLIST_SIZE_INCREMENTS) * CHILDLIST_SIZE_INCREMENTS;
```
jonmv, line 189:
```suggestion
                int reservedSize = ((index + CHILDLIST_SIZE_INCREMENTS) / CHILDLIST_SIZE_INCREMENTS) * CHILDLIST_SIZE_INCREMENTS;
```

jonmv (approved):
Consider the simplification of that one expression, otherwise 👍

bratseth (approved):

container-search/src/main/java/com/yahoo/search/grouping/vespa/ResultBuilder.java
baldersheim: I added minSize = index + 1 to make it clearer. I wanted to keep the (CHILDLIST_SIZE_INCREMENTS - 1) as together with the division and the multiplication is an often used pattern for rounding of integers.

PR #22101, opened 2022-04-12
Ensure that config get includes local-only options

mpolden:
`config get` shows the effective configuration (local + global), but didn't
account for options that were only set in the local config.

@bratseth


bratseth (approved):
👍

Issue #22075, opened 2022-04-09
Allow pre-filtering when combining nearestNeighbor with geoLocation


jobergum:
Given the following query 

```
where  {targetNumHits: 10000}nearestNeighbor(embedding, q) and geoLocation(locations, 11, -110, "100 km") 
```
The global filtering handling estimates that the geoLocation will match 100% of documents and it ends up as a post-filter after first retrieving 10K from the NN search, and then applying the geoLocation as a post filter. 

This can be seen from the following trace. I think this happens because geoLocation has  "allow_termwise_eval": 0 ?

```
 "global_filter": {
  "[type]": "GlobalFilter",
  "brute_force_limit": 0.05,
  "exists": true,
  "hit_ratio": 0.9999997487718713,
  "hits": 3980445
},
```
Blueprint:

```
{
    "optimized": {
        "[type]": "search::queryeval::AndBlueprint",
        "children": {
            "[0]": {
                "[type]": "search::queryeval::NearestNeighborBlueprint",
                "approximate": true,
                "attribute_tensor": "tensor<float>(x[128])",
                "docid_limit": 3980446,
                "estimate": {
                    "[type]": "HitEstimate",
                    "allow_termwise_eval": 0,
                    "cost_tier": 1,
                    "empty": false,
                    "estHits": 10000,
                    "tree_size": 1
                },
                "explore_additional_hits": 0,
                "fields": {
                    "[0]": {
                        "[type]": "Field",
                        "fieldId": 12,
                        "handle": 0,
                        "isFilter": false
                    },
                    "[type]": "FieldList"
                },
                "global_filter": {
                    "[type]": "GlobalFilter",
                    "brute_force_limit": 0.05,
                    "exists": true,
                    "hit_ratio": 0.9999997487718713,
                    "hits": 3980445
                },
                "has_index": true,
                "isTermLike": true,
                "query_tensor": "tensor<float>(x[128])",
                "sourceId": 4294967295,
                "target_num_hits": 10000,
                "top_k_found_hits": 10000
            },
            "[1]": {
                "[type]": "proton::documentmetastore::(anonymous namespace)::WhiteListBlueprint",
                "docid_limit": 3980446,
                "estimate": {
                    "[type]": "HitEstimate",
                    "allow_termwise_eval": 1,
                    "cost_tier": 1,
                    "empty": false,
                    "estHits": 3980446,
                    "tree_size": 1
                },
                "isTermLike": false,
                "sourceId": 4294967295
            },
[...]

PR #22104, opened 2022-04-12
let LocationPreFilter contribute to global filter

arnej27959:

@baldersheim please review
@jobergum @havardpe FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
@@ -253,6 +253,19 @@ public:
         }
     }
 
+    SearchIterator::UP
+    createFilterSearch(bool strict, FilterConstraint constraint) const override
+    {
+        if (constraint == FilterConstraint::UPPER_BOUND) {
+            auto wrapper = std::make_unique<FilterWrapper>(getState().numFields());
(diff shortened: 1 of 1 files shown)


arnej27959:
note: this needs testing to see if it helps or hurts performance.

baldersheim (approved):
This might be required too, but in #22075 there is only a LocationPostFilterBlueprint which will never allow term wise eval.
I assume https://github.com/vespa-engine/vespa/blob/master/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp#L333 is were it fails out. And that will be the case if hit estimates are larger than 10%

baldersheim:
@arnej27959 Should we proceed with this one anyway, or should we just close it ?

baldersheim:
@arnej27959 Ping ref comment above, still valid or close ?

baldersheim (approved):

geirst (changes requested):
Please add unit test as well: searchlib/src/tests/attribute/searchable/attributeblueprint_test.cpp.

aressem:
@arnej27959 @geirst Integrate or close ?

geirst:
The current behaviour is documented here: https://docs.vespa.ai/en/approximate-nn-hnsw.html#combining-approximate-nearest-neighbor-search-with-filters. We can just close this for now.

PR #22105, opened 2022-04-12
switch to new-style config

arnej27959:

@baldersheim or @bratseth please review


bratseth (approved):
I may be missing something in this diff but don't you need some non-cfg change to cause the tests to produce new-style config too?

baldersheim:
@bratseth I think these files are not produced at all. They are checked in and are only consumed.
The produced ones are in config-model module.

baldersheim (approved):

PR #22112, opened 2022-04-13
Allow -1 as project id to indicate NO project id, which disabled dep …

jonmv:
…orch


mpolden (approved):
👍

hmusum (approved):

PR #22113, opened 2022-04-13
Warn if config changes without generation changing

hmusum:


hakonhall:
AFAIK we don't know how the config can change (.equals()) if the hashes are equal?

hakonhall (approved):

PR #22119, opened 2022-04-13
Jonmv/no more serviceview/v1

jonmv:
@freva don' merge yet.


jonmv:
@gjoranv perhaps the second commit enables some more cleanup as well?

jonmv:
We must also expose `/orchestrator/v1/` through `/application/v4/` before we can do this, so we have a starting point for getting the services and hosts.

freva (approved):
:trophy:

gjoranv:
Please revert the changes related to jersey bundles. configserver provides jersey to the controller components.

jonmv:
Already done in #22154

gjoranv:
Thanks, and great job removing serviceview. Only orchestrator uses Jersey client in vespa-engine now.

PR #22122, opened 2022-04-14
Wait in mock destination session for RPC timeout

jonmv:

### DIFF
--- a/documentapi/src/test/java/com/yahoo/documentapi/messagebus/MessageBusDocumentApiTestCase.java
+++ b/documentapi/src/test/java/com/yahoo/documentapi/messagebus/MessageBusDocumentApiTestCase.java
@@ -12,6 +12,7 @@ import com.yahoo.documentapi.DocumentAccess;
 import com.yahoo.documentapi.DocumentOperationParameters;
 import com.yahoo.documentapi.ProgressToken;
 import com.yahoo.documentapi.Response;
+import com.yahoo.documentapi.Response.Outcome;
 import com.yahoo.documentapi.VisitorParameters;
 import com.yahoo.documentapi.VisitorSession;
 import com.yahoo.documentapi.messagebus.protocol.CreateVisitorReply;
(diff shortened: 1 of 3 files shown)


documentapi/src/test/java/com/yahoo/documentapi/messagebus/MessageBusDocumentApiTestCase.java
```diff
@@ -119,17 +120,24 @@ public void requireThatTimeoutWorks() throws InterruptedException {
...
-                                                          })
-                                                          .withDeadline(Instant.now().minusSeconds(1)))
-                          .isSuccess());
-        assertTrue(latch.await(60, TimeUnit.SECONDS));
-        assertNotNull(response.get());
-        assertEquals(Response.Outcome.TIMEOUT, response.get().outcome());
+
+        // The setup is broken. Zero or negative timeout has semantics in messagebus, so ensuring a timeout is impossible,as negative
+        // remaining time is converted to 1 ms further down. We therefore try tenfold, and accept > 0 timeouts as success ...  (；☉_☉)
+        // We could return a timeout from the destination, but then we're not testing what we want, namely that the sender times out.
```
vekterli, line 126: Are there any guarantees that this will trigger any timeouts at all? Looks like the PR build failed with all responses completing successfully. It looks like all messages are routed to an instance of the `Destination` test mock class, so could instead add functionality to it for deferring response sending. The test could then wait for a client-generated timeout response before allowing the server response to be sent. No need for a retry loop either, since it should be deterministic.

jonmv:
This didn't help, though >_<

jonmv:
We could solve this by using a deadline on the client side, and just pass the truncated time remaining across the wire, but then the client code diverges from the rest. Deadlines are probably the right thing, though

vekterli:
Does the destination automatically generate a timeout response if it has not itself sent a proper response before the deadline is reached? If so, that sounds a bit broken. I would have expected it to keep its mittens to itself and leave the auto-generation to the client.

jonmv:
I think once the message is sent, the client won't sent a timeout. I may be wrong. Let me check ...

vekterli (approved):
:+1:

PR #22123, opened 2022-04-15
Jonmv/disallow newer compile than parent version

jonmv:
@gjoranv please review.

I'm not sure this is 100% the best thing, because it assumes the parent is precisely one of our tenant parent poms, or at the very least, that it specifies no version of its own. We could traverse to find the uppermost parent, perhaps ... `vespaversion` shouldn't be set, and the test not fail, when this is not a hosted app.


jonmv:
Urgh, those two first commits were not supposed to be here >_<

PR #22124, opened 2022-04-15
Jonmv/require compile version leq parent

jonmv:
@gjoranv please review.


gjoranv (approved):
Duplicate code is never nice, but I can't think of anything better.

PR #22100, opened 2022-04-12
Remove compatibility serialisation for revisions, and add (de-) for job types

jonmv:
@freva please review.

PR #22125, opened 2022-04-16
Jonmv/serialize job type with zone

jonmv:
@freva please review. Must be merged _after_ https://github.com/vespa-engine/vespa/pull/22100 has rolled out.


hosted-tenant-base/pom.xml
```diff
@@ -32,7 +32,7 @@
     </scm>
 
     <properties>
-        <vespaversion>${project.version}</vespaversion>
+        <vespaversion>7-SNAPSHOT</vespaversion>
```
freva, line 35: Intentional?

freva (approved):

hosted-tenant-base/pom.xml
jonmv: Oops, no.

jonmv:
Ah, there were more places jobs were serialised, which I need to fix in both the previous PR and this ...

jonmv:
This PR now switches to serialisation with zone everywhere, and `JobType.fromJobName` is invoked with zones elsewhere. The last PR will remove the compatibility deserialisation code, and let all `JobType` objects hold a specific zone. `systemTest` and `stagingTest` will depend on the zone registry. Most uses are in unit tests, where the `ZoneRegistryMock` can now have a static set of zones, and then also a static member for those two job types; elsewhere, the controller or job controller can be used to instantiate them, or just check the environment of the job zones, if the current usage is `xxx.equals(systemTest)` .

freva (approved):

jonmv:
What, more conflicts?

jonmv:
:D

freva (approved):

PR #22139, opened 2022-04-19
JobType now always has a zone, which simplifies things

jonmv:
@freva please review, but do not merge :)


jonmv:
This commit removes the compatibility deserialisation of `JobType` based on job name, and replaces the contained map from system to zone, with just a specific zone. 
This is possible now since all serialised data contains this specific zone, and all other JobType creators provide the necessary info to find the right zone (`fromJobName(name, zones)`). 
The bijection between (zone, isProductionTest) and (job) simplifies things, and makes it possible to have more uniform test setup in unit tests.
The static job declarations were moved to test code, as they're still useful for making test code less error prone.

jonmv:
@mpolden FYI.

jonmv:
This must roll out strictly after https://github.com/vespa-engine/vespa/pull/22125

freva (approved):
:1st_place_medal:

mpolden:
Great!

PR #22141, opened 2022-04-19
Ean/make notifier hosted aware

ean:

@mpolden


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/notify/NotifyDispatcher.java
```diff
@@ -0,0 +1,15 @@
...
+import com.yahoo.vespa.hosted.controller.tenant.TenantContacts;
+
+import java.util.Collection;
+
+/**
+ * NotifyDispatcher defines the notify API
+ *
+ * @author enygaard
+ */
+public interface NotifyDispatcher {
```
mpolden, line 13: Why is this needed in addition to the existing `Mailer` interface?

ean:
We need the extra indirection to separate fetching of contact data from creating the mail content.
CuratorDB lives in vespa-engine/vespa, while hosted information like the dashboard url lives in the close repository. 

And to avoid exposing all of the Hosted System information in `vespa` the indirection here is needed.

It will also give us an extendable way to add other destinations in the future (Slack, IRC, ...) now that Notifier is no longer responsible for dispatching a notification to the destination.

PR #22146, opened 2022-04-19
Jonmv/more validation

jonmv:
@hmusum please review.


hmusum (approved):
LGTM. We are sure there are no applications or other data that uses now invalid names?

hmusum (changes requested):
Unit test failed:

eTest
15:05:05 [ERROR] test(com.yahoo.vespa.hosted.controller.api.integration.deployment.JobTypeTest)  Time elapsed: 0.064 s  <<< ERROR!
15:05:05 java.lang.IllegalArgumentException: region name must match '[a-z]([a-z0-9-]*[a-z0-9])*', but got: 'snøhetta'

jonmv:
Can't be absolutely sure. 

Haha, nice, I added those "regions" last week :)

jonmv:
I think I found that it's highly unlikely the application names will no longer match. The `.` is the only real issue here, and we shouldn't be using that ourselves, at least (in system tests).

PR #22147, opened 2022-04-19
Set permissions on all intermediate directories as well

freva:


vespabase/src/rhel-prestart.sh
```diff
@@ -89,29 +89,29 @@ fixdir () {
         echo "fixdir: Expected 4 params, got:" "$@"
         exit 1
     fi
-    mkdir -p "$4"
+
+    parent=`dirname "$4"`
+    if ! [ -d "$parent" ]; then
```
hakonhall, line 94:
```
```suggestion
    [ -d "$parent" ] || fixdir "$1" "$2" "$3" "$parent"
```

hakonhall (approved):

PR #22148, opened 2022-04-19
Only deregister once

jonmv:
@hakonhall or @freva please review and merge if this is an OK assumption, i.e., stop only needs to be called once per agent in this test.

### DIFF
--- a/node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integration/ContainerTester.java
+++ b/node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integration/ContainerTester.java
@@ -36,6 +36,7 @@ import java.util.Optional;
 import java.util.concurrent.Phaser;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Logger;
 
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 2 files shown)


hakonhall:
stopForHostSuspension is idempotent and may be called multiple times.  It will remove the container of the associated node, and is a no-op if it is already removed.

node-admin/src/test/java/com/yahoo/vespa/hosted/node/admin/integration/ContainerTester.java
```diff
@@ -92,13 +93,16 @@ public class ContainerTester implements AutoCloseable {
...
+                    AtomicBoolean stopped = new AtomicBoolean();
                     @Override public void converge(NodeAgentContext context) {
                         super.converge(context);
                         phaser.arriveAndAwaitAdvance();
                     }
                     @Override public void stopForHostSuspension(NodeAgentContext context) {
-                        super.stopForHostSuspension(context);
-                        phaser.arriveAndDeregister();
+                        if (stopped.compareAndSet(false, true)) {
+                            super.stopForHostSuspension(context);
```
hakonhall, line 103: So `super.stopForHostSuspension()` should be called for each invocation, but phaser.arriveAndDeregister() only on the first invocation per node?

jonmv: Ideally they would just call `arriveAndAwaitAdvance` on each tick, but I believe these agents may be removed, and no longer tick, is that right? In that case, we need to deregister ...

jonmv: Hmm, perhaps the problem is rather that they `tick` after being removed here ... Urgh.

freva: Not sure why we have deregister on `stopForHostSuspension()`. If you want it at as the last call, put it inside `stopForRemoval()`. After that `tick()` will never be called again.

jonmv: Yes, I must have thought suspension was removal when I added this synchronisation some time back. It looks right to just arrive as usual when stopping for suspension. Done now.

hakonhall: Node agent instances are normally preserved with stopForHostSuspension() and until jdisc deconstruction.  Only under very particular circumstances would a node agent may be removed after it has invoked stopForHostSuspension(), something like: the reason for the reboot no longer exists, and a has been removed from an application. E.g. RebootTest doesn't test for this.

hakonhall (approved):

PR #22152, opened 2022-04-19
Add multi value read views that handles mapping from enumerated value

toregge:
and copy to add/remove weight.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

geirst:
The screwdriver failure is unrelated: "16:21:50 [INFO] documentapi ........................................ FAILURE [01:07 min]"

PR #22156, opened 2022-04-19
Use com.yahoo.messagebus.Error instead of java.lang.Error

baldersheim:
@vekterli and @jonmv PR

### DIFF
--- a/documentapi/src/main/java/com/yahoo/documentapi/Result.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/Result.java
@@ -116,7 +116,7 @@ public class Result {
             case TRANSIENT_ERROR:
                 return new Error(ErrorCode.TRANSIENT_ERROR, ResultType.TRANSIENT_ERROR.name());
             case CONDITION_NOT_MET_ERROR:
-                return new Error(ErrorCode.TRANSIENT_ERROR, ResultType.CONDITION_NOT_MET_ERROR.name());
+                return new Error(ErrorCode.FATAL_ERROR, ResultType.CONDITION_NOT_MET_ERROR.name());
             case FATAL_ERROR:
                 return new Error(ErrorCode.FATAL_ERROR, ResultType.FATAL_ERROR.name());
         }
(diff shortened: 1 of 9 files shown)


documentapi/src/main/java/com/yahoo/documentapi/Result.java
```diff
@@ -85,5 +111,16 @@ public enum ResultType {
         @Deprecated(since = "7", forRemoval = true) // TODO: Remove on Vespa 8 — this is a Response outcome, not a Result outcome.
         CONDITION_NOT_MET_ERROR
     }
+    public static Error toError(ResultType result) {
+        switch (result) {
+            case TRANSIENT_ERROR:
+                return new Error(ErrorCode.TRANSIENT_ERROR, ResultType.TRANSIENT_ERROR.name());
+            case CONDITION_NOT_MET_ERROR:
```
bratseth, line 118: Condition not met isn't transient?

bratseth (approved):

documentapi/src/main/java/com/yahoo/documentapi/Result.java
baldersheim: No you will need to change your condition for it to have a bigger chance at success, and anyway it should not be here. See comment in Result.ResultType.

documentapi/src/main/java/com/yahoo/documentapi/Result.java
```diff
@@ -85,5 +111,16 @@ public enum ResultType {
...
+                return new Error(ErrorCode.TRANSIENT_ERROR, ResultType.CONDITION_NOT_MET_ERROR.name());
```
bratseth, line 119:
```suggestion
                return new Error(ErrorCode.FATAL_ERROR, ResultType.CONDITION_NOT_MET_ERROR.name());
```

baldersheim:
Failure is unrelated

jonmv (commented):
:

vekterli (commented):
:+1:

PR #22162, opened 2022-04-20
Allow reading API key from default path when Auth0 is not configured

mpolden:
Only last commit is new.

@ean

### DIFF
--- a/client/go/cmd/config_test.go
+++ b/client/go/cmd/config_test.go
@@ -157,6 +157,12 @@ func TestReadAPIKey(t *testing.T) {
 	require.Nil(t, err)
 	assert.Equal(t, []byte("foo"), key)
 
+	// Cloud CI does not read key from disk as it's not expected to have any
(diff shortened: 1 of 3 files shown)


client/go/cmd/config_test.go
```diff
@@ -157,6 +157,12 @@ func TestReadAPIKey(t *testing.T) {
 	require.Nil(t, err)
 	assert.Equal(t, []byte("foo"), key)
 
+	// Cloud CI does not read key from disk as it's not expected to have any
+	cli, _, _ = newTestCLI(t, "VESPA_CLI_CLOUD_CI=true")
+	key, err = cli.config.readAPIKey(cli, vespa.PublicSystem, "t1")
+	require.Nil(t, err)
+	assert.Nil(t, []byte(nil), key)
```
ean, line 164: `[]byte(nil)` should be removed I think.

ean (approved):
Fix test argument to verify that key is nil, then this is ok to merge.

PR #22164, opened 2022-04-20
Fix node is down status in /application/v4 nodes API

freva:


mpolden (approved):
👍 
Maybe nodes API should just return "down" as a separate field in its API? So we don't need to have this logic in multiple places.

freva:
I'll add it to nodes API. Merging this so it doesn't show nodes as down in the meantime.

PR #22167, opened 2022-04-20
Jonmv/require parent geq compile version

jonmv:
@gjoranv please review.

Only throw if the parent project is a hosted Vespa parent.
Use parentArtifact, which is defined even when parent project cannot be generated. 
Fallback to the version of the plugin itself for what the "vespaversion" is, for those without a vespa parent.

### DIFF
--- a/vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
+++ b/vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
@@ -70,7 +70,7 @@ public class ApplicationMojo extends AbstractMojo {
 
         boolean hasVespaParent = false;
         Artifact parentArtifact = current.getParentArtifact();
-        if (parentArtifact != null && parentArtifact.getGroupId().startsWith("com.yahoo.vespa")) {
+        if (parentArtifact != null && (parentArtifact.getGroupId().startsWith("com.yahoo.vespa.") || parentArtifact.getGroupId().startsWith("ai.vespa."))) {
             hasVespaParent = true;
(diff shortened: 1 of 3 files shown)


vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
```diff
@@ -56,15 +57,30 @@ public void execute() throws MojoExecutionException {
...
-                                        System.currentTimeMillis());
+        Version compileVersion = Version.from(vespaversion);
+
+        MavenProject current = project;
+        while (current.getParent() != null && current.getParent().getParentArtifact() != null)
+            current = current.getParent();
+
+        boolean hasVespaParent = false;
+        Artifact parentArtifact = current.getParentArtifact();
+        if (parentArtifact != null && parentArtifact.getGroupId().startsWith("com.yahoo.vespa")) {
```
gjoranv, line 73: Should we add a future proof `startsWith(ai.vespa.)` as well?

Suggest to add a trailing dot to com.yahoo.vespa anyway.

gjoranv (approved):

gjoranv:
The failing unit test is probably among the usual suspects 
```
MessageBusDocumentApiTestCase>AbstractDocumentApiTestCase.requireThatAsyncSessionWorks:123 » NullPointer
```

vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
jonmv: Good points, fixed.

jonmv:
Definitely is.

gjoranv:
👍

PR #22168, opened 2022-04-20
Do not push notifications for none-production environments

ean:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
@@ -1,6 +1,7 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.notify;
 
+import com.yahoo.config.provision.Environment;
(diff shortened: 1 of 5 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notify/Notifier.java
```diff
@@ -56,7 +57,12 @@ public void dispatch(List<Notification> notifications, NotificationSource source
 
     private boolean skipSource(NotificationSource source) {
         // Limit sources to production systems only. Dev and test systems cause too much noise at the moment.
-        return source.jobType().map(t -> !t.isProduction()).orElse(false);
+        if (source.zoneId().map(z -> z.environment() != Environment.prod).orElse(false)) {
+            return true;
+        } else if (source.jobType().map(t -> !t.isProduction()).orElse(false)) {
+            return true;
+        }
+        return false;
```
mpolden, line 65: Use `!source.isProduction()`?

ean: I would like to exclude staging and systemtest for now. Too many emails in the beginning will just be annoying.

mpolden (approved):

PR #22170, opened 2022-04-20
Report incorrect rank-profile.

baldersheim:
@geirst and @havardpe PR
I think this is better. If you agree I will remove current exception handling that will trigger in the case where you have no rank-profiles. Not even default.


havardpe (approved):

geirst (approved):
👍

PR #22174, opened 2022-04-20
Expose down in nodes API

freva:
After this has rolled out, I'll fix the TODO in this PR and create an internal PR to actually set the down field in `NodeRepositoryClient`


mpolden (approved):
👍

PR #22175, opened 2022-04-20
Disallow URI type resources for hosted Vespa

jonmv:


hmusum (approved):
LGTM, @lesters this is in accordance with the work you are doing?

lesters:
@hmusum No, I think this is for URI resources downloaded to content nodes (defined in document schema). Is that correct @jonmv ?

jonmv:
This is only used for ranking constants now.

jonmv:
URI types are also undocumented: https://docs.vespa.ai/en/reference/schema-reference.html#constant

PR #22181, opened 2022-04-20
Remove orphaned wanted state

hakonhall:
When
1. new config removes at least one node, or
2. if the loaded wanted state contains a wanted state for a non-configured node

then trigger the writing of the wanted state back to ZooKeeper.  This writing is based on the set of wanted states for the currently configured nodes, and will therefore shed any orphaned wanted state, which is what we want.


clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/DatabaseHandlerTest.java
```diff
@@ -104,4 +133,29 @@ public void empty_bundle_is_returned_if_no_db_connection() throws Exception {
...
+        // wanted state outside the existing nodes.
+        Node storageNode5 = Node.ofStorage(5);
+        NodeState maintenance = new NodeState(NodeType.STORAGE, State.MAINTENANCE);
+        databaseContext.getCluster().getNodeInfo(storageNode5).setWantedState(maintenance);
+        var expectedWantedStates = new TreeMap<>(Map.of(storageNode5, maintenance));
+
+        // Ensure database is connected to ZooKeeper
+        assertTrue(handler.doNextZooKeeperTask(databaseContext));
+
+        // Verify saveWantedStates() is invoked once
```
vekterli, line 153: Should this be "Verify storeWantedStates() is invoked once", not `saveWantedStates`?

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/DatabaseHandlerTest.java
```diff
@@ -104,4 +133,29 @@ public void empty_bundle_is_returned_if_no_db_connection() throws Exception {
         assertEquals(ClusterStateBundle.empty(), retrievedBundle);
     }
 
+    @Test
+    public void save_wanted_state_sheds_orphaned_states() throws Exception {
```
vekterli, line 137: Consider changing test name to something like `save_wanted_states_preserves_existing_wanted_state_mappings` to more accurately reflect its semantics

vekterli (approved):
:+1:

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/DatabaseHandlerTest.java
hakonhall: Done

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/DatabaseHandlerTest.java
hakonhall: Done

vekterli (approved):

PR #22187, opened 2022-04-20
Initialize logging earlier in container startup sequence [run-systemtest]

baldersheim:
@bjorncs @gjoranv @hmusum PR

### DIFF
--- a/jdisc_core/pom.xml
+++ b/jdisc_core/pom.xml
@@ -279,6 +279,19 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>3.0.0-M6</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>integration-test</goal>
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
                 <configuration>
                     <archive>
--- a/jdisc_core/pom.xml
+++ b/jdisc_core/pom.xml
@@ -66,7 +66,6 @@
             <groupId>com.yahoo.vespa</groupId>
             <artifactId>defaults</artifactId>
             <version>${project.version}</version>
-            <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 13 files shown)


hmusum (approved):

vespalog/src/main/java/com/yahoo/log/InitializeLog.java
```diff
@@ -0,0 +1,20 @@
+package com.yahoo.log;
+
+/**
+ * Sets up Vespa logging. Call a setup method to set up this.
+ *
+ * @author baldersheim
+ */
+public class InitializeLog {
```
gjoranv, line 8: Was it intentional to put this class in a PublicApi package?

vespalog/src/main/java/com/yahoo/log/InitializeLog.java
```diff
@@ -0,0 +1,20 @@
...
+    static {
+        LogSetup.initVespaLogging("Container");
```
gjoranv, line 10: The term "Container" seems a bit too specific for a generic module as vespalog.

jdisc_core/pom.xml
```diff
@@ -167,6 +167,11 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.vespa</groupId>
+            <artifactId>vespalog</artifactId>
```
gjoranv, line 172: I would like to verify the effects this dependency has on e.g. the package exports from jdisc_core (classpath) and the set of dependencies made visible via container-dev, before this PR is merged.

vespalog/src/main/java/com/yahoo/log/InitializeLog.java
baldersheim: Agree, changed to 'unused-default'

vespalog/src/main/java/com/yahoo/log/InitializeLog.java
baldersheim: No, that is not necessary. I moved it to the log.impl and annotated that with ExportPackage.
I assume that is necessary to access it from other modules.

jdisc_core/pom.xml
baldersheim: OK, I will hold until green light.

hmusum (approved):

baldersheim:
This seems to be a bit brittle, and there was already explicit support for this in jdisc.
So this is superceeded in #22456

PR #22189, opened 2022-04-20
Revert "Cheaper way to get number of children."

hmusum:
Reverts vespa-engine/vespa#22166

Some system tests started failing, could be that expected output in tests needs to be updated instead of this revert. @baldersheim please review


baldersheim:
Fixed in #22188 instead

PR #22197, opened 2022-04-21
Jonmv/warn on no tests

jonmv:
@freva please review.


jonmv:
I guess we also need a little something to show the new status in the console?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobStatus.java
```diff
@@ -90,18 +90,17 @@ static Optional<Run> lastCompleted(NavigableMap<RunId, Run> runs) {
 
     static Optional<Run> lastSuccess(NavigableMap<RunId, Run> runs) {
         return runs.descendingMap().values().stream()
-                   .filter(run -> run.status() == RunStatus.success)
+                   .filter(run -> run.hasEnded() && ! run.hasFailed())
```
freva, line 93:
```suggestion
                   .filter(Run::hasSucceeded)
```

freva (approved):

jonmv:
.

PR #22199, opened 2022-04-21
Always use MultiValueReadView for dot product feature on weighted set attributes.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #22226, opened 2022-04-22
Set default max query items

bratseth:
- Set default max query items to 10000
- Disallow setting maxQueryItems in requests
- Include maxQueryItems in the Query query profile type

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -6635,11 +6635,13 @@
     ],
     "methods": [
       "public void <init>()",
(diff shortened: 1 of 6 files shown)


olaaun (approved):

olaaun (changes requested):
ABI spec mismatch

```
[ERROR] Class com.yahoo.search.query.properties.DefaultProperties: Extra method public static void requireNotPresentIn(java.util.Map)
[ERROR] Class com.yahoo.search.query.properties.DefaultProperties: Extra field public static final com.yahoo.processing.request.CompoundName MAX_QUERY_ITEMS
```

olaaun (approved):

PR #22231, opened 2022-04-22
Build for 17

gjoranv:
NOTE: this is for 8 only

Build, unit test and javadoc should work for 17 now, verified in docker centos-stream8. However, a couple of unit tests fail consistently both on master+jdk11 and 8+17.

 Build command:
```
mvn clean install -T 1C -Dtest='!ConfigServerRestExecutorImplTest,!CertificateAuthorityApiTest' -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
```

### DIFF
--- a/abi-check-plugin/src/main/java/com/yahoo/abicheck/collector/AnnotationCollector.java
+++ b/abi-check-plugin/src/main/java/com/yahoo/abicheck/collector/AnnotationCollector.java
@@ -13,7 +13,7 @@ public class AnnotationCollector extends ClassVisitor {
   private final Set<String> annotations = new HashSet<>();
 
   public AnnotationCollector() {
-    super(Opcodes.ASM7);
+    super(Opcodes.ASM9);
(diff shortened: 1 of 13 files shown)


hmusum (approved):

gjoranv:
This had to be forcefully removed, as I forgot to "rebase and merge". It's been too long since 6->7.

hakonhall:
Out of curiosity - why do you specify `-DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false` ?

bjorncs (commented):
👍

gjoranv:
@hakonhall, because the build fails otherwise. When excluding test classes, I have previously only used `-DfailIfNoTests=false`, but using maven 3.8 and release 17, maven told me to add the latter option. Perhaps the first one isn't necessary anymore. I haven't tried without it.

PR #22259, opened 2022-04-25
Avoid logging error when client exits

hmusum:
Depend on spotbugs  to avoid issue with
edu.umd.cs.findbugs.annotations.SuppressFBWarnings that leads to compiler
warning

### DIFF
--- a/zookeeper-command-line-client/pom.xml
+++ b/zookeeper-command-line-client/pom.xml
@@ -66,19 +66,14 @@
       <artifactId>snappy-java</artifactId>
       <scope>compile</scope>
     </dependency>
-
+    <dependency>
+      <!-- Needed by zookeeper, which only has an optional dependency. -->
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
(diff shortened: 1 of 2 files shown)


zookeeper-command-line-client/pom.xml
```diff
@@ -70,6 +70,15 @@
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
+            <arg>-Xlint:all</arg>
```
gjoranv, line 78: Is this to switch off `-Werror`?

gjoranv: I suggest to add this dependency to the pom instead (see also comment in the parent pom for this dep).
```
    <dependency>
      <!-- Needed by zookeeper, which only has an optional dependency. -->
      <groupId>com.github.spotbugs</groupId>
      <artifactId>spotbugs-annotations</artifactId>
    </dependency>
```

hmusum: Thanks! Will use this

gjoranv (approved):

PR #22271, opened 2022-04-26
Revert "Pass parent version, rather than just whether there was a vespa parent"

hmusum:
Reverts vespa-engine/vespa#22242

Seems like version is unquoted, that will not work


jonmv:
https://github.com/vespa-engine/vespa/pull/22273

PR #22280, opened 2022-04-26
Use MultiValueReadView to read values in multivalue attribute dynamic field writer.

toregge:
@geirst : please review


geirst (approved):
👍

PR #22286, opened 2022-04-26
never use "qrserver" as service name

arnej27959:
@gjoranv please review and merge
note: cleanup to be done later, removing the flag entirely.


gjoranv (approved):
Awaiting merge until we have a green build

PR #22297, opened 2022-04-27
Validate file extension for files in application package [run-systemtest]

hmusum:
Start validating file extension for files application package subdirectories,
only a subset of subdirectories handled for now.


hmusum:
Build fails when creating temp directory, seems unrelated to changes (observed also in another PR).

bratseth (approved):

PR #22304, opened 2022-04-27
Create schema-info for global schema info

bratseth:
@arnej27959 Lester is away so perhaps you could do this one?

This is just creating a new config to hold schema info, the same as documentdb-info, but available independently of search clusters.

I plan to simplify by switching everything over to use it and get rid of documentdb-info since we don't actually need this to be different per cluster. But I don't want to do all in one PR.


container-search/src/main/java/com/yahoo/search/searchchain/Execution.java
```diff
@@ -140,7 +140,7 @@ public Context(SearchChainRegistry searchChainRegistry, IndexFacts indexFacts, S
         public Context(SearchChainRegistry searchChainRegistry, IndexFacts indexFacts,
                        SpecialTokenRegistry tokenRegistry, RendererRegistry rendererRegistry, Linguistics linguistics,
                        Executor executor) {
-            this(searchChainRegistry, indexFacts, SchemaInfo.empty(), tokenRegistry, rendererRegistry, linguistics, Runnable::run);
+            this(searchChainRegistry, indexFacts, SchemaInfo.empty(), tokenRegistry, rendererRegistry, linguistics, executor);
```
arnej27959, line 142: this change looks totally unrelated

container-search/src/main/resources/configdefinitions/container.search.schema-info.def
```diff
@@ -0,0 +1,25 @@
+# Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+namespace=container.search
+
+## The name of this schema
+schema[].name string
+
+## The id of the summary class
```
arnej27959, line 7: this id isn't used anymore, can be dropped

container-search/src/main/resources/configdefinitions/container.search.schema-info.def
```diff
@@ -0,0 +1,25 @@
...
+# schema[].summaryclass[].fields[].name string
+## The type of a field in the summary class
+# schema[].summaryclass[].fields[].type string
+## Whether this field is a dynamic snippet
+# schema[].summaryclass[].fields[].dynamic bool default=false
+
+## Information about rank profiles
+schema[].rankprofile[].name string
+schema[].rankprofile[].hasSummaryFeatures bool default=true
+schema[].rankprofile[].hasRankFeatures bool default=true
```
arnej27959, line 21: i can't find any place where hasRankFeatures is actually used.

arnej27959 (approved):
looks good to me, only minor comments

container-search/src/main/java/com/yahoo/search/searchchain/Execution.java
bratseth: It fixes a bug in an unused method preserved for compatibility, introduced in an earlier part of this work.

container-search/src/main/resources/configdefinitions/container.search.schema-info.def
bratseth: Yes, it is commented out. I'll be adding it back when I get rid of documentdb-config, but for that I need to rewrite the way the summary part of that config is generated as it currently is generated from another config, which is no-no.

container-search/src/main/resources/configdefinitions/container.search.schema-info.def
bratseth: This was easier to keep. It will be used when I move away from documentdb-info.

PR #22306, opened 2022-04-27
validate nested type for weighted sets

arnej27959:
* prepare to be more strict in Vespa 8

@bratseth please review
@gjoranv FYI
@baldersheim FYI


bratseth (approved):
👍

PR #22307, opened 2022-04-27
Jonmv/warn on old parent

jonmv:
@freva please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -744,6 +745,7 @@ private static void toSlime(Cursor cursor, Notification notification, boolean in
 
     private static String notificationTypeAsString(Notification.Type type) {
         switch (type) {
+            case submission:
```
freva, line 748: This will result in potentially 2 "lines" of application package warnings (In the UI): 1 with just 1 "message" and another with potentially several "messages".

I think it's fine, but we could consider combining this into a single notification by concatenating messages arrays.

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
freva: :disappointed:

PR #22319, opened 2022-04-27
Initialize transitive ctor dependencies before use

vekterli:
@arnej27959 please review. Maybe there are some sneaky hidden dependencies here that I'm not aware of.

Ensure that `SBEnv` fields that are accessed by `RPCHooks` ctor are
initialized before the `RPCHooks` field itself is created.

### DIFF
--- a/slobrok/src/vespa/slobrok/server/sbenv.cpp
+++ b/slobrok/src/vespa/slobrok/server/sbenv.cpp
@@ -105,15 +105,16 @@ SBEnv::SBEnv(const ConfigShim &shim)
       _shuttingDown(false),
       _partnerList(),
       _me(createSpec(_configShim.portNumber())),
-      _rpcHooks(*this),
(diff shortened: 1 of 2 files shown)


arnej27959:
while I can't see any sneaky dependencies, it looks like rpchooks ctor only takes some references to things, which was ok to do before those objects were fully initialized.

arnej27959 (approved):
👍

vekterli:
The problem was `RPCHooks::RPCHooks` calling `env.localHistory()` which in turn calls `_localRpcMonitorMap.history()` prior to its construction. As you mention, this is a non-virtual function which only returns a reference that will eventually be valid, but UBSan was not happy about the whole ordeal. Seems like UBSan always checks that _any_ method invoked on an object that has a vtable happens when the vptr is valid, even if the method itself is not virtual.

PR #22323, opened 2022-04-27
Validate file extension for app files, part 2

hmusum:


bratseth (approved):
Nice! 

I think models/ might contain lots of different endings now due to TF so we probably can't validate that dir until Vespa 8.

hmusum:
Right, I'll have a look at what shows up in logs after this has been released and do adjustments

PR #22325, opened 2022-04-27
Validate file extension for app files, part 3

hmusum:


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
```diff
@@ -751,35 +751,42 @@ static File applicationFile(File parent, Path path) {
...
+        }
     }
 
     static {
         validFileExtensions = Map.ofEntries(
                 Map.entry(Path.fromString(COMPONENT_DIR), Set.of(".jar")),
                 Map.entry(CONSTANTS_DIR, Set.of(".json", ".json.lz4")),
                 Map.entry(Path.fromString(DOCPROCCHAINS_DIR), Set.of(".xml")),
-                Map.entry(MODELS_DIR, Set.of(".model")),
+                // Map.entry(MODELS_DIR, Set.of(".model")),  TODO: Enable on Vespa 8
```
bratseth, line 779: Should add .onnx and .txt here at least. But since people can write custom models, like e.g embedders consuming files in any formats they like, and it could be logical to place these in models, perhaps we shouldn't validate this dir at all?

bratseth (approved):
👍

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
hmusum: Yes, I think we should skip validating files in models/ , there are too many (unknown) valid extensions.

PR #22329, opened 2022-04-28
Use parsedVersion. It was accidentally removed earlier on.

baldersheim:
@gjoranv or @hmusum or @bjorncs PR


hmusum (approved):

gjoranv:
@baldersheim If you haven't already, please deactivate the osgi plugin in IntelliJ. IIRC, that was the cause of this problem.

PR #22348, opened 2022-04-28
Jonmv/apache clustercontroller client [run-systemtest

jonmv:
@hakonhall please review.


orchestrator/src/main/java/com/yahoo/vespa/orchestrator/OrchestratorImpl.java
```diff
@@ -427,10 +425,7 @@ public boolean isQuiescent(ApplicationId id) {
                 ClusterControllerClient client = clusterControllerClientFactory.createClient(clusterControllers, cluster.clusterId().s());
                 for (ServiceInstance service : cluster.serviceInstances()) {
                     try {
-                        ClusterControllerStateResponse response = client.setNodeState(context,
-                                                                                      VespaModelUtil.getStorageNodeIndex(service.configId()),
-                                                                                      MAINTENANCE);
-                        if ( ! response.wasModified)
+                        if ( ! client.setNodeState(context, service.hostName(), VespaModelUtil.getStorageNodeIndex(service.configId()), MAINTENANCE))
```
hakonhall, line 428: isQuiescent() is not used except by tests.  Was this something we used for CC migration that we no longer need?  It is a nice method though.

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/controller/ClusterControllerClientTimeouts.java
```diff
@@ -36,40 +36,28 @@ public class ClusterControllerClientTimeouts implements JaxRsTimeouts {
...
-            throw new UncheckedTimeoutException("Timed out after " + timeBudget.originalTimeout().get());
-        }
+    public Duration connectTimeout() {
+        return CONNECT_TIMEOUT;
+    }
 
-        return serverTimeout;
+    public TimeBudget readBudget() {
+        // timeLeft = CONNECT_TIMEOUT + readTimeout
+        return timeBudget.withReserved(CONNECT_TIMEOUT);
```
hakonhall, line 60: Throw if non-positive?

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/model/StorageNodeImpl.java
```diff
@@ -69,41 +62,7 @@ public void setNodeState(OrchestratorContext context, ClusterControllerNodeState
...
-                    e);
-        }
-
-        if ( ! response.wasModified) {
-            throw new HostStateChangeDeniedException(
-                    hostName(),
-                    HostedVespaPolicy.SET_NODE_STATE_CONSTRAINT,
-                    "Failed to set state to " + wantedNodeState + " in cluster controller: " + response.reason);
-        }
+        client.setNodeState(context, storageService.hostName(), nodeIndex, wantedNodeState);
```
hakonhall, line 65: If a probe was denied, this method used to throw an exception which was important.

jonmv: Oh, looked to me like this was never a probe. Ok, fine.

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/controller/ClusterControllerClientTimeouts.java
jonmv: It does that when you read from the budget.

orchestrator/src/main/java/com/yahoo/vespa/orchestrator/OrchestratorImpl.java
jonmv: Hmm ... true, it was.

hakonhall (approved):

PR #22354, opened 2022-04-29
Jonmv/fix parent version

jonmv:
@freva

### DIFF
--- a/vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
+++ b/vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
@@ -70,7 +70,7 @@ public class ApplicationMojo extends AbstractMojo {
 
         Version parentVersion = null;
         Artifact parentArtifact = current.getParentArtifact();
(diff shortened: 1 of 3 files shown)


vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
```diff
@@ -93,6 +93,10 @@ private void addBuildMetaData(File applicationDestination) throws MojoExecutionE
         }
     }
 
+    static boolean isVespaParent(String groupId) {
+        return groupId.matches("(com\\.yahoo\\.vespa|ai\\.vespa)(\\..+)?");
```
ean, line 97: Add `^` to avoid matching with example.com.ai.vespa?

freva (approved):

vespa-application-maven-plugin/src/main/java/com/yahoo/container/plugin/mojo/ApplicationMojo.java
jonmv: Urgh, yes, it was "startsWith" earlier.

jonmv: It broken because the trailing `.` was added ...

PR #22363, opened 2022-04-29
Don't support user config under document

bratseth:
I want to remove support for user config under <document> to enable some other simplification.
I've never heard of anyone using that, it's not explicitly documented, and it doesn't make much sense to me.


baldersheim (approved):
pragmatic... I like it 🥇

gjoranv:
👍

PR #22374, opened 2022-04-29
Jonmv/remove last controller jersey client [run-systemtest]

jonmv:
@hakonhall please review.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/core/documentapi/VespaDocumentAccess.java
+++ b/container-core/src/main/java/com/yahoo/container/core/documentapi/VespaDocumentAccess.java
@@ -20,6 +20,7 @@ import com.yahoo.documentapi.messagebus.MessageBusParams;
 import com.yahoo.documentapi.messagebus.protocol.DocumentProtocolPoliciesConfig;
 import com.yahoo.messagebus.MessagebusConfig;
 import com.yahoo.vespa.config.content.DistributionConfig;
+import com.yahoo.yolean.concurrent.Memoized;
 
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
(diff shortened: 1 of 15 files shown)


container-core/src/main/java/com/yahoo/container/core/documentapi/VespaDocumentAccess.java
```diff
@@ -72,14 +64,7 @@ public void shutdown() {
...
     void protectedShutdown() {
-        delegate.updateAndGet(access -> {
-            super.shutdown();
-            shutDown = true;
-            if (access != null)
-                access.shutdown();
-
-            return null;
-        });
+        delegate.close();
```
jonmv, line 66: No longer calls `super.shutdown()`, but this does nothing for this setup anyway.

yolean/src/main/java/com/yahoo/yolean/concurrent/Memoized.java
```diff
@@ -0,0 +1,52 @@
...
+    public static <T extends AutoCloseable> Memoized<T, ?> of(Supplier<T> factory) {
+        return new Memoized<>(factory, AutoCloseable::close);
+    }
+
+    @Override
+    public T get() {
+        if (wrapped == null) synchronized (monitor) {
+            if (factory != null) wrapped = factory.get();
+            factory = null;
+            if (wrapped == null) throw new IllegalStateException("already closed");
```
hakonhall, line 37: Consider something like

```
synchronized (monitor) {
  if (wrapped == null) {
    if (factory == null) throw new IllegalStateException("already closed");
    wrapped = factory.get();
    factory = null;
  }
}
```
  
which I find easier to read.  Also, it prevents an exception with a wrong message if factory.get() returns null.

hakonhall (approved):

yolean/src/main/java/com/yahoo/yolean/concurrent/Memoized.java
jonmv: Yes, I want to handle exceptions from the factory, too.

hakonhall (approved):

hakonhall (approved):

PR #22381, opened 2022-05-02
Bratseth/schema info 1

bratseth:
This adds summary info to the new schema-info, not used for anything yet but I plan to replace documentdb-info.
Schema-info is now generated in the same way as other schema-derived configs.

### DIFF
--- a/config-model/src/main/java/com/yahoo/searchdefinition/derived/SchemaInfo.java
+++ b/config-model/src/main/java/com/yahoo/searchdefinition/derived/SchemaInfo.java
@@ -7,6 +7,8 @@ import com.yahoo.searchdefinition.RankProfileRegistry;
 import com.yahoo.searchdefinition.Schema;
 import com.yahoo.searchlib.rankingexpression.Reference;
 import com.yahoo.tensor.TensorType;
+import com.yahoo.vespa.config.search.SummarymapConfig;
+import com.yahoo.vespa.documentmodel.SummaryTransform;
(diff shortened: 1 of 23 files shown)


config-model/src/main/java/com/yahoo/searchdefinition/derived/SchemaInfo.java
```diff
@@ -0,0 +1,145 @@
...
+        var schemaBuilder = new SchemaInfoConfig.Schema.Builder();
+        schemaBuilder.name(schema.getName());
+        addSummaryConfig(schemaBuilder);
+        addRankProfilesConfig(schemaBuilder);
+        builder.schema(schemaBuilder);
+    }
+
+    private void addSummaryConfig(SchemaInfoConfig.Schema.Builder schemaBuilder) {
+        /*
+        for (SummaryConfig.Classes sclass : summaryConfig.classes()) {
```
arnej27959, line 66: this comment can probably be removed

container-search/src/main/resources/configdefinitions/container.search.schema-info.def
```diff
@@ -5,15 +5,15 @@ namespace=search.config
 schema[].name string
 
 ## The id of the summary class
-# schema[].summaryclass[].id int
+schema[].summaryclass[].id int
```
arnej27959, line 8: the corresponding documentdb[].summaryclass[].id is never used, so you can remove this id.

arnej27959 (approved):
minor nits only 👍

PR #22391, opened 2022-05-02
Use SchemaInfo

bratseth:


bratseth:
I'll merge this now to let the system tests have their say.

PR #22417, opened 2022-05-03
Revert "Set permissions on all intermediate directories as well"

aressem:
Reverts vespa-engine/vespa#22147


aressem:
This does not ensure that `/opt/vespa/tmp` is owned by $VESPA_USER if it exists and is owned by root when the script is executed. This broke the system test in PR builds where `/opt/vespa/tmp/systemtests` is created and owned by root before starting vespa. We could consider working around this for the PR case, but since the ownership of parents is not being handled correctly by this code we should probably fix it.

freva (approved):

PR #22419, opened 2022-05-03
Add feature flag for installing Java from latest-appstream repo

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -382,6 +382,13 @@ public class Flags {
             "Takes effect immediately",
             TENANT_ID);
 
+    public static final UnboundBooleanFlag INSTALL_JDK_FROM_LATEST_APPSTREAM_REPO = defineFeatureFlag(
+            "install-jdk-from-latest-appstream-repo", false,
+            List.of("hmusum"), "2022-05-03", "2022-06-03",
+            "Whether we should install JDK from latest-appstream repo on hosts (e.g. for security issues fixed in newer JDK versions)",
+            "Takes effect immediately",
+            TENANT_ID);
```
freva, line 390: `TENANT_ID` is a strange dimension to use for a host-level flag since all of them are `hosted-vespa`

freva (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hmusum: Indeed. Fixed

PR #22425, opened 2022-05-03
Revert "Add lock around visitor queue size read during metric snapshot callback"

toregge:
Reverts vespa-engine/vespa#22414

Unit test storageserver_gtest_runner_app started failing with thread sanitizer warnings about lock order reversal.


vekterli (approved):
:+1:

PR #22426, opened 2022-05-03
Use schema-info

bratseth:


bratseth:
This one also needs a round of testing.

PR #22430, opened 2022-05-03
GC unused metrics

baldersheim:
@vekterli PR


vekterli (approved):
:+1: ahh multioperations, a blast from the past

PR #22446, opened 2022-05-04
Consolidate allocated hosts from all model versions 

bratseth:
- Activate the superset of hosts allocated from all model versions
- Use the version of each host allocated by the newest model version

@hmusum or @freva please review.

### DIFF
--- a/configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployNodeAllocationTest.java
+++ b/configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployNodeAllocationTest.java
@@ -1,0 +1,119 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.config.server.deploy;
+
+import com.yahoo.component.Version;
+import com.yahoo.config.model.api.HostProvisioner;
+import com.yahoo.config.model.api.ModelFactory;
+import com.yahoo.config.provision.Capacity;
+import com.yahoo.config.provision.ClusterMembership;
+import com.yahoo.config.provision.ClusterSpec;
+import com.yahoo.config.provision.HostSpec;
+import com.yahoo.config.provision.NodeResources;
+import com.yahoo.config.provision.ProvisionLogger;
+import com.yahoo.config.provision.Zone;
+import com.yahoo.vespa.config.server.MockProvisioner;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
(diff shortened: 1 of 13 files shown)


configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployNodeAllocationTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+        @Override
+        public List<HostSpec> prepare(ClusterSpec cluster, Capacity capacity, ProvisionLogger logger) {
+            if (cluster.id().value().equals("container")) { // the container cluster from the app package: Use this to test
+                if (invocation == 0) { // Building the latest model version, 7.3: Always first
+                    invocation++;
+                    // Returns nodes which are on both versions to trigger building old versions
+                    return List.of(host("host0", resources(3), 0, "7.3", cluster),
+                                   host("host1", resources(3), 1, "7.3", cluster),
+                                   host("host2", resources(3), 2, "7.2", cluster));
```
freva, line 82:
```suggestion
                                   host("host2", resources(3), 2, "7.3", cluster));
```

freva (commented):
Code looks fine, but I'm not sure if it's any better to have nodes on different versions than to just have them expire?
* We may have made assumption(s) in code that all active nodes have the same wanted version
* Controller node convergence checker needs to support this, but how would it know that not it's OK that not all nodes are no version `x` even though it deployed with `x`?

configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployNodeAllocationTest.java
bratseth: It must be like this actually - this isn't the model version but the node's version, and these versions decide what model versions we'll build.

configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployNodeAllocationTest.java
```diff
@@ -0,0 +1,119 @@
...
+                } else if (invocation == 1) { // Building 7.2
+                    invocation++;
+                    return List.of(host("host3", resources(2), 2, "7.2", cluster),
+                                   host("host4", resources(2), 3, "7.2", cluster),
+                                   host("host4", resources(2), 4, "7.2", cluster));
```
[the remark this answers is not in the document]
bratseth: thanks!

bratseth:
(But note that due to the TODO I added in ModelsBuilder.versionsToBuild we will not actually discover model versions to build from nodes not requested by the latest model, so the only consequence of this for now is that we'll always use the node spec produced by the latest model version.)

freva (approved):

hmusum (approved):
👍

PR #22454, opened 2022-05-04
Bjorncs/annotations [run-systemtest]

bjorncs:

### DIFF
--- a/annotations/pom.xml
+++ b/annotations/pom.xml
@@ -9,7 +9,7 @@
     <relativePath>../parent/pom.xml</relativePath>
   </parent>
   <artifactId>annotations</artifactId>
-  <packaging>jar</packaging>
+  <packaging>bundle</packaging>
   <version>7-SNAPSHOT</version>
   <name>annotations</name>
   <description>Public API annotations</description>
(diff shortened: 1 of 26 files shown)


annotations/pom.xml
```diff
@@ -46,6 +46,31 @@
...
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>parse-version</id>
+            <goals>
+              <goal>parse-version</goal>
```
jonmv, line 56: There's a plugin for everything, huh?

jonmv (approved):
What provides the bundle, though?

jonmv:
There's no `annotations` bundle started in the container in that system test.

bjorncs:
@jonmv It's embedded and re-exported through container-disc bundle. It's a common pattern in the container.

bjorncs:
Fails on system test (which is still broken).

fileacquirer/pom.xml
```diff
@@ -22,6 +22,12 @@
       <artifactId>config</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.yahoo.vespa</groupId>
+      <artifactId>annotations</artifactId>
+      <version>${project.version}</version>
+      <scope>provided</scope>
```
gjoranv, line 29: fileacquirer is a regular jar, so provided does not make much sense.

jdisc_core/pom.xml
```diff
@@ -167,6 +167,12 @@
             <version>${project.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.vespa</groupId>
+            <artifactId>annotations</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
gjoranv, line 174: 'provided' deps do not make sense for jdisc_core, as it's on the classpath. For this to be usable, it must be compile, and the packages re-exported via jdisc_core.jar.

gjoranv:
Why is it necessary to export these annotations? Not necessarily wrong, but if we do it, they should be "exported" via jdisc_core, not container-disc.

bjorncs:
> Why is it necessary to export these annotations? Not necessarily wrong, but if we do it, they should be "exported" via jdisc_core, not container-disc.

It's exported for the `@Inject` annotation.

Why is it required exporting it from jdisc_core instead of container-disc? It's the DI framework of container-disc that understands the new annotation, not Guice/jdisc_core.

gjoranv:
It would be needed for user bundles that are not built with our bundle-plugin, which should be possible, of course. Apart from that, Inject does not really need to be exported.

These annotations were always embedded in jdisc_core and available via the classpath, so exporting from container-disc is a false flag.

gjoranv:
I can fix the latter when finished with #22483

PR #22455, opened 2022-05-04
Add 'input' as alias to 'ranking.feature'

bratseth:


container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
```diff
@@ -160,6 +155,7 @@ else if (key.first().equals("rankfeature") || key.first().equals("featureoverrid
 
     @Override
     public void set(CompoundName key, Object value, Map<String,String> context) {
+        System.out.println("Setting " + key);
```
jobergum, line 158: Not sure if you want this? :)

jobergum (changes requested):
Great, just a debug printline

container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
bratseth: Oops, tx

jobergum (approved):

PR #22456, opened 2022-05-04
Avoid the 'jdisc/' prefix used for logging during startup of configse…

baldersheim:
…rver.

Disable logging during startup of jdisc. No reason for doing this different than any other container.

@hakonhall or @hmusum or @bjorncs or @gjoranv


hmusum (approved):

hakonhall (approved):
Do you mind fixing 
https://github.com/vespa-engine/vespa/blob/master/standalone-container/src/main/sh/standalone-container.sh#L190-L192
too?

baldersheim:
Fixed in #22457

PR #22458, opened 2022-05-05
Bratseth/schema info 5 take 3

bratseth:
This one is ok with https://github.com/vespa-engine/vespa/pull/22440


arnej27959 (approved):
looks good to me

PR #22459, opened 2022-05-05
Revert "Simplify" MERGEOK

bratseth:
Reverts vespa-engine/vespa#22395

Reverting this just to see if it caused the breaking system tests.


jonmv:
👍

PR #22464, opened 2022-05-05
Name matching query properties using camelCase as done elsewhere.

geirst:
Also provide the lowercase version as alias for backwards compatibility.

@baldersheim please review
@jobergum FYI


baldersheim:
OK, but I do not like camelcase....

geirst:
The decision was made in the architect meeting last week (@bratseth FYI)

baldersheim (approved):

PR #22466, opened 2022-05-05
Don't create ConstArrayRef of empty node vector

vekterli:
@geirst please review

`ConstArrayRef` ctor takes ref of 0th vector element, which is undefined
if the vector is empty. As such, this is a general `ConstArrayRef` issue,
but I'm hesitant to add an extra branch to a very frequently called
function, so adding a workaround at this particular callsite instead.


geirst (approved):
👍

vekterli:
Will fix this with no additional cost in `ConstArrayRef` in a follow-up. Closing this since it will be redundant.

vekterli:
Superseded by #22493

PR #22467, opened 2022-05-05
Move setting of node resources to node repo

hmusum:
* Will use the same node resources for all model versions, so no
  need to worry about changing the resources and getting issues
  when changing topology due to retiring of nodes
* Will do resource settings in the same place as for other admin
  clusters

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
@@ -51,7 +51,6 @@ import com.yahoo.vespa.model.content.storagecluster.StorageCluster;
 import com.yahoo.vespa.model.search.IndexedSearchCluster;
 import com.yahoo.vespa.model.search.Tuning;
 import org.w3c.dom.Element;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


freva (approved):

bratseth:
This will also provision new nodes, but for all applications on infrastructure upgrade instead of when each application upgrade. It's a nicer location for this, but also more risky I think.

hmusum:
We build the newest model first on infrastructure upgrade as well, so any change to node resources in the old code would also happen then and not only on application upgrade (this is also the case for new code, as you noted).

Maybe I should have kept the note about only changing this with a feature flag as I originally intended.

bratseth:
Ah, yes that is true. That could also be solved by the TODO on having a better way to determine what versions we're currently running. We could also disregard any allocations made by a model version not in use on bootstrap if we wanted a quick fix for this, but I think this is more of a long term thing.

So, in summary, this PR won't cause any changes to the allocations being activated, since they will always be based on the latter model with my other recent PR and we always deploy the new model anyway. But _with_ this PR, we'll not be able to reduce risk from only letting such changes take effect on application upgrade.

_However_, without this the models may disagree on which set of nodes should be retired, and when that has consequences for the config emitted by the model it may lead to some inconsistent behavior across a cluster. So, on balance I think it's better to merge this.

jonmv:
Which Vespa version? That of the config server?

jonmv:
The doc says it's the version we want the nodes to run. Assuming that's correct, in the strictest way, it should work. If we rather mean a hypothetical "version we want the nodes to run _if using this version when we actually activate_" or something like that, then I guess it won't.

jonmv:
Idea is if we want to switch flavours, and the current version is 7.123, then we add an entry with the new flavour with key 7.124—the crux is that no applications must yet be on this version when the config servers upgrade. Since config servers upgrade before applications, it wouldn't be a problem if 7.124 was never released either; when upgrading to 7.125, the newest applications would be on 7.123, and the switch would happen when they upgrade to 7.125 as well.

PR #22475, opened 2022-05-05
Generate messages with same content as console when dispatching notif…

ean:
…ications

@mpolden I don't really like the name NotificationContent. It basically contains the context of a notification that we are going to dispatch somewhere, do you have some ideas for a better name? Maybe `DispatchContext`?

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/FormattedNotification.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/FormattedNotification.java
@@ -1,0 +1,40 @@
+package com.yahoo.vespa.hosted.controller.notification;
+
+import java.net.URI;
+import java.util.Objects;
+
+/**
+ * Contains formatted text that can be displayed to a user to give extra information and pointers for a given
+ * Notification.
+ *
+ * @author enygaard
+ */
+public class FormattedNotification {
+    private final String prettyType;
+    private final String messagePrefix;
+    private final URI uri;
+    private final Notification notification;
+
+    public FormattedNotification(Notification notification, String prettyType, String messagePrefix, URI uri) {
+        this.prettyType = Objects.requireNonNull(prettyType);
+        this.messagePrefix = Objects.requireNonNull(messagePrefix);
+        this.uri = Objects.requireNonNull(uri);
+        this.notification = Objects.requireNonNull(notification);
+    }
+
+    public String prettyType() {
+        return prettyType;
+    }
+
+    public String messagePrefix() {
+        return messagePrefix;
+    }
+
+    public URI uri() {
+        return uri;
+    }
+
+    public Notification notification() {
+        return notification;
+    }
+}
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/MissingOptionalException.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/MissingOptionalException.java
@@ -1,0 +1,18 @@
+package com.yahoo.vespa.hosted.controller.notification;
+
+/**
+ * Used to signal that an expected value was not present when creating NotificationContent
+ *
+ * @author enygaard
+ */
+class MissingOptionalException extends RuntimeException {
+    private final String field;
+    public MissingOptionalException(String field) {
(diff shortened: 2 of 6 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContent.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Objects;
+
+/**
+ * Contains formatted text that can be displayed to a user to give extra information and pointers for a given
+ * Notification.
+ *
+ * @author enygaard
+ */
+public class NotificationContent {
+    private final String prettyType, messagePrefix;
```
mpolden, line 13: Our convention is one field per line.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContent.java
```diff
@@ -0,0 +1,40 @@
...
+        return messagePrefix;
+    }
+
+    public URI getURI() {
+        return uri;
+    }
+
+    public Notification getNotification() {
+        return notification;
+    }
```
mpolden, line 39: For the most part we don't prefix getters with `get`, to avoid "get-get-get-get-get-get" when accessing nested members.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContentFormatter.java
```diff
@@ -0,0 +1,209 @@
...
+        var clusterId = expect(source.clusterId());
+        return Text.format("Cluster %s in %s.%s for %s.%s",
+                clusterId.value(),
+                zone.environment(), zone.region(),
+                application, instance);
+    }
+
+    private URI url(Notification n) {
+        var source = n.source();
+        if (source.application().isPresent()) {
```
mpolden, line 188: Nit: You can avoid the excessive nesting if you restructure a bit, i.e:

```
if (source.application().isEmpty()) { return ... }
if (source.instance().isEmpty()) { return ... }
if (source.jobType().isEmpty() || source.runNumber().isEmpy() { return ... }
return ...
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContentFormatter.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+                return zoneRegistry.dashboardUrl(ApplicationId.from(source.tenant(), source.application().get(), source.instance().get()));
+            }
+            return zoneRegistry.dashboardUrl(source.tenant(), source.application().get());
+        }
+        return zoneRegistry.dashboardUrl(source.tenant());
+    }
+
+
+    private static <T> T expect(Optional<T> optional) {
```
mpolden, line 206: Our convention is to name methods that verify some constraint `requireXXX`, e.g. `requirePresent`.

For this particular case I would suggest `requirePresent(Optional<T> optional, String fieldName)` and include `fieldName` in the exception message so that we can know which field is missing:

`requirePresent(source.application(), "application")`

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContent.java
ean: right-o

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationContentFormatter.java
ean: I removed the entire thing, it was only needed for unknown Notifications, and that is something that should not really happen.

mpolden:
It basically contains the context of a notification that we are going to dispatch somewhere, do you have some ideas for a better name?

ean:
> > It basically contains the context of a notification that we are going to dispatch somewhere, do you have some ideas for a better name?
> 
> How about simply `FormattedNotification` and `NotificationFormatter`?

That works. Thanks.

mpolden (approved):

PR #22478, opened 2022-05-05
Prepare for adding GCP ArtifactExpirer

freva:

* Rename `ContainerImageExpirer` to `ArtifactExpirer`
* Change service registry to return `ArtifactExpirer` as a function of `Cloud`
* Expire artifacts in all clouds where there's at least 1 zone in this system
* Will skip if no `ArtifactExpirer` is returned for the zone, but if an exception is thrown during maintenance, that will affect the `successFactor`
* Replace `architecture` field with just `tag` since that's easier to generalize to GCP


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/artifact/ArtifactRegistry.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A registry of artifacts (e.g. container image or RPM).
+ *
+ * @author mpolden
+ */
+public interface ArtifactRegistry {
+
+    /** Delete all given artifacts */
+    void deleteAll(List<Artifact> images);
```
mpolden, line 15: images -> artifacts

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
```diff
@@ -99,7 +99,7 @@ public interface ServiceRegistry {
 
     BillingDatabaseClient billingDatabase();
 
-    ContainerRegistry containerRegistry();
+    ArtifactRegistry containerRegistry();
```
mpolden: containerRegistry -> artifactRegistry

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/artifact/Artifact.java
```diff
@@ -0,0 +1,95 @@
...
+ * A registry artifact (e.g. container image or RPM)
+ *
+ * @author mpolden
+ */
+public class Artifact {
+
+    private final String id;
+    private final Optional<String> registry;
+    private final Optional<String> repository;
+    private final Optional<String> tag;
```
mpolden, line 20: Won't there always be some kind of tag/version? This should be non-optional.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/integration/ArtifactRegistryMock.java
```diff
@@ -13,26 +13,26 @@
...
  */
-public class ContainerRegistryMock implements ContainerRegistry {
+public class ArtifactRegistryMock implements ArtifactRegistry {
 
-    private static final Comparator<ContainerImage> comparator = Comparator.comparing(ContainerImage::registry)
-                                                                           .thenComparing(ContainerImage::repository)
-                                                                           .thenComparing(ContainerImage::version);
+    private static final Comparator<Artifact> comparator = Comparator.comparing((Artifact artifact) -> artifact.registry().orElse(""))
+                                                                           .thenComparing(artifact -> artifact.repository().orElse(""))
+                                                                           .thenComparing(Artifact::version);
```
mpolden, line 20: Nit: Bad alignment.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/ArtifactExpirerTest.java
```diff
@@ -2,35 +2,33 @@
...
+        ArtifactExpirer expirer = new ArtifactExpirer(tester.controller(), Duration.ofDays(1));
+        ArtifactRegistryMock registry = tester.controllerTester().serviceRegistry().artifactRegistry(CloudName.defaultName()).orElseThrow();
 
         Instant instant = tester.clock().instant();
-        ContainerImage image0 = new ContainerImage("image0", "registry.example.com", "vespa/vespa", instant, Version.fromString("7.1"), Optional.empty());
-        ContainerImage image1 = new ContainerImage("image1", "registry.example.com", "vespa/vespa", instant, Version.fromString("7.2"), Optional.of(Architecture.amd64));
-        ContainerImage image2 = new ContainerImage("image2", "registry.example.com", "vespa/vespa", instant, Version.fromString("7.4"), Optional.of(Architecture.amd64));
-
+        Artifact image0 = new Artifact("image0", "registry.example.com", "vespa/vespa", "7.4", instant, Version.fromString("7.1"));
+        Artifact image1 = new Artifact("image1", "registry.example.com", "vespa/vespa", "7.4-amd64", instant, Version.fromString("7.2"));
```
mpolden, line 30: Tags should match the version number.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/artifact/Artifact.java
mpolden: My point was that an artifact (be it a container image or RPM) will always have some sort of "tag", e.g. a version number, so it can be non-optional.

Also, maybe we should remove the separate `version` field  and parse it from `tag` in the constructor as those should never come from two separate sources.

mpolden (approved):
👍

PR #22482, opened 2022-05-05
Remove '.sum' form vds sum metrics.

baldersheim:
Remove '.sum' from metric names for storage node and also remove the average metrics for the same.
Remove '.sum' from distributor metrics set and remove distributor average metrics.
GC '.sum' from distributor metric names.
Remove '.alldisks' from metric names and update tests.
GC '.alldisks' from filestor metrics.

@vekterli PR
@yngveaasheim FYI


vekterli (approved):
:+1: nice!

We should include this change in the Vespa 8 change log if it's not already there.

PR #22497, opened 2022-05-06
Use valid mappings exception --> response

jonmv:
@bjorncs please review.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/billing/BillingApiHandlerV2.java
```diff
@@ -76,6 +76,7 @@ private static RestApi createRestApi(BillingApiHandlerV2 self) {
                 .addRoute(RestApi.route("/billing/v2/accountant/preview/tenant/{tenant}")
                         .get(self::previewBill)
                         .post(Slime.class, self::createBill))
+                .addExceptionMapper(RuntimeException.class, (__, e) -> ErrorResponse.internalServerError(e.getMessage()))
```
bjorncs, line 79: This is not really needed unless the client want to parse the error message. A 500 response is returned whenever a request handler throws an exception. It will result in a HTML response though.

bjorncs (approved):

PR #22506, opened 2022-05-07
Check that the value has dimensions before trying to decode.

baldersheim:
@havardpe PR As discussed over the phone today


havardpe (commented):
👍

PR #22519, opened 2022-05-09
Reapply setup vespalog in jdisc core

gjoranv:


container-disc/src/main/java/com/yahoo/container/jdisc/ConfiguredApplication.java
```diff
@@ -42,7 +42,7 @@
 import com.yahoo.jrt.Transport;
 import com.yahoo.jrt.slobrok.api.Register;
 import com.yahoo.jrt.slobrok.api.SlobrokList;
-import com.yahoo.log.LogSetup;
+import com.yahoo.log.impl.InitializeLog;
```
hakonhall, line 45: Stray import.

vespalog/src/main/java/com/yahoo/log/impl/InitializeLog.java
```diff
@@ -0,0 +1,21 @@
...
+ * Sets up Vespa logging. Call a setup method to set up this.
+ *
+ * @author baldersheim
+ */
+public class InitializeLog {
+    private static final Logger log = Logger.getLogger(InitializeLog.class.getName());
+
+    public static void init(String programName) {
+        LogSetup.initVespaLogging(programName);
+        log.log(Level.INFO, "Initialized logging for " + programName);
```
hakonhall, line 18: When sifting through large amounts of logs it is often useful to know when a process started.  That was the intention of the `Starting jdisc at version 7.581.34` message.  If the initialization of the logging is moved so should that message (and this message is redundant).

gjoranv: This is just a draft, not ready for review yet :)

gjoranv: The log message you mention could be moved to ApplicationLoader, but without outputting the version which is not available for jdisc_core. Making it available is outside the scope of this PR.

gjoranv:
This should be good now, @baldersheim

vespalog/src/main/java/com/yahoo/log/impl/InitializeLog.java
hakonhall: Not sure what's best.

baldersheim (approved):
Have you run any of the system tests that failed on last try ?
I think that is a good start to sanitise it prior to merge.

gjoranv:
Yes, I tried all tests that I could run locally, and I see no abnormal/unexpected logging now. I suspect the problem was the double call to initializeLog(). Must mention that I have a problem with the configserver shutdown in my container, so those tests never came to the log check, but the log looks fine. No curator/zk log errors/warnings now.

PR #22522, opened 2022-05-10
Configure zookeeper log file only for services that might use zookeeper

hmusum:


freva (commented):
Not sure if I understand this correctly, looks to me that with this PR we will avoid setting `ZOOKEEPER_LOG_FILE_PREFIX` environment variable for the other services, but this code runs unconditionally: https://github.com/vespa-engine/vespa/blob/6168edcdbf5e7abfcc2979845020efc8b66db8dd/container-disc/src/main/sh/vespa-start-container-daemon.sh#L303
And then https://github.com/vespa-engine/vespa/blob/6168edcdbf5e7abfcc2979845020efc8b66db8dd/vespalog/src/main/java/com/yahoo/log/LogSetup.java#L165:L169 will also run because `System.getProperty("zookeeper_log_file_prefix")` would return an empty string and not null?

Also could we add a temporary `else` that deletes the `ZOOKEEPER_LOG_FILE_PREFIX` file?

hmusum:
I tested this and it still worked, but I think it's better to not set the system property at all if we shouldn't use a zookeeper log file. Also fixed a bug and deleted `ZOOKEEPER_LOG_FILE_PREFIX` file

freva (approved):
:1st_place_medal:

PR #22524, opened 2022-05-10
Support constants on the same format as inputs

bratseth:
Now that we can parse tensors we can allow constants blocks to have the same syntax as inputs blocks.


lesters (approved):
👍

PR #22537, opened 2022-05-11
Add madvise(buf, len, MADV_HUGEPAGE) call to signal that you strongly…

baldersheim:
… prefer huge pages.

This will make a difference if /sys/kernel/mm/transparent_hugepage/enabled has been set to [madvise], and not [always].

@toregge or @vekterli or @havardpe or @geirst PR
@jobergum FYI


toregge (approved):

vekterli (commented):
:+1:

PR #22557, opened 2022-05-12
Synchronize access MERGEOK

bratseth:
I'm removing the need for this, but keep it thread safe for now.

@baldersheim or @lesters


lesters (approved):

baldersheim:
I think it would be sufficient to just just a ConcurrentHashHap similar to what is done in LargeRankExpressions.

baldersheim:
I needed it sorted for consistent config generation and added a method that returned a sorted list     public Collection<RankExpressionBody> expressions() {
        return expressions.values().stream().sorted().collect(Collectors.toUnmodifiableList());
    }

PR #22562, opened 2022-05-12
Ean/tenant

ean:

@bjorncs

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/TenantId.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/TenantId.java
@@ -1,0 +1,34 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.config.provision;
+
+import ai.vespa.validation.PatternedStringWrapper;
+import com.google.common.io.BaseEncoding;
+
+import java.security.SecureRandom;
+import java.util.regex.Pattern;
+
+/**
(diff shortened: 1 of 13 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/TenantId.java
```diff
@@ -0,0 +1,29 @@
...
+import ai.vespa.validation.PatternedStringWrapper;
+
+import java.util.regex.Pattern;
+
+/**
+ * Represents a tenant id in the provision API.
+ *
+ * @author enygaard
+ */
+public class TenantId extends PatternedStringWrapper<TenantId> {
```
bjorncs, line 13: Consider adding a `TenantId` unit test to verify validation logic

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -158,32 +161,35 @@ public Tenant tenantFrom(Slime slime) {
 
     private AthenzTenant athenzTenantFrom(Inspector tenantObject) {
         TenantName name = TenantName.from(tenantObject.field(nameField).asString());
+        TenantId id = TenantId.from(tenantObject.field(idField).asString());
```
bjorncs, line 164: How is old ZK data handled (where tenant id is not present)?

bjorncs (approved):
See comments

PR #22566, opened 2022-05-12
Wrap the FNET_Transport to make it exception safe.

baldersheim:
@havardpe or @toregge or @vekterli PR

### DIFF
--- a/searchcore/src/apps/proton/proton.cpp
+++ b/searchcore/src/apps/proton/proton.cpp
@@ -205,9 +205,11 @@ buildTransportConfig() {
 
 class Transport {
 public:
-    Transport(const fnet::TransportConfig & config)
+    Transport(const fnet::TransportConfig & config, FastOS_ThreadPool & threadPool)
         : _transport(config)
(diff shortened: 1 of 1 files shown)


searchcore/src/apps/proton/proton.cpp
```diff
@@ -268,10 +281,9 @@ App::main(int argc, char **argv)
         setupSignals();
         setup_fadvise();
         FastOS_ThreadPool threadPool(128_Ki);
-        FNET_Transport transport(buildTransportConfig());
-        transport.Start(&threadPool);
-        startAndRun(threadPool, transport, argc, argv);
-        transport.ShutDown(true);
+        Transport transport(buildTransportConfig());
+        transport.transport().Start(&threadPool);
```
havardpe, line 285: consider doing Start in the wrapper constructor (if Start is not called or it fails, WaitFinished will block forever)

havardpe (approved):

havardpe (approved):

PR #22576, opened 2022-05-12
Create distributable constants on deriving

bratseth:


bratseth:
Merging this to get an off hours test run.

lesters (commented):
:+1:

PR #22564, opened 2022-05-12
Vespa 8: Rename 'qrs' to 'access'

bjormel:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #22583, opened 2022-05-13
Revert "Rename 'qrs' to 'access'"

gjoranv:
Reverts #22564 (rename qrs -> access)


aressem:
I think @bjormel needs to comment on this. I was under the impression that this folder was to be named access for Vespa 8.

bjormel:
Reverting, if it broke something, I will assess next week, and redo

gjoranv:
Thanks, build is green again.

PR #22585, opened 2022-05-13
Split tenant info into separate rest resources

oyving:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -517,6 +545,141 @@ private SlimeJsonResponse tenantInfo(TenantInfo info, HttpRequest request) {
...
+            tenant.setString("website", info.website());
+
+            toSlime(info.address(), root); // will create "address" on the parent
+        }
+
+        return new SlimeJsonResponse(slime);
+    }
+
+    private SlimeJsonResponse putTenantInfo(String tenantName, HttpRequest request, BiFunction<CloudTenant, Inspector, SlimeJsonResponse> handler) {
+        return controller.tenants().get(tenantName)
```
freva, line 569: Use the same pattern for the GETs?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -517,6 +545,141 @@ private SlimeJsonResponse tenantInfo(TenantInfo info, HttpRequest request) {
...
+                .withAddress(mergedAddress);
+
+        validateMergedTenantInfo(mergedInfo);
+
+        controller.tenants().lockOrThrow(cloudTenant.name(), LockedTenant.Cloud.class, lockedTenant -> {
+            lockedTenant = lockedTenant.withInfo(mergedInfo);
+            controller.tenants().store(lockedTenant);
+        });
+
+        return new MessageResponse("Tenant info updated");
```
freva, line 596: Maybe this should be part of the common function? Why isn't `validateMergedTenantInfo()` called in the other `put*` methods?

freva (commented):
Test failure:
```
[ERROR]   PathGroupTest.uniqueMatches:49 Paths '/application/v4/tenant/{tenant}/info/' and '/application/v4/tenant/{tenant}/info/{*}' overlap.
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
oyving: Because it does not validate any of the information that is part of this update. We should probably split this into two different validation methods because part of the issue we are trying to solve here is that I can't update a tenant that has bad information in both console forms :)

freva (approved):

PR #22580, opened 2022-05-13
Align code on hugepage size

baldersheim:
@toregge PR

PR #22586, opened 2022-05-13
Revert "Align code on hugepage size"

baldersheim:
Reverts vespa-engine/vespa#22580
@toregge PR
Needs to get rid of all our small libs first


toregge (commented):
👍

PR #22587, opened 2022-05-13
Fix request body sanitizing

mpolden:
@freva


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/auditlog/AuditLog.java
```diff
@@ -112,16 +110,27 @@ public enum Method {
...
-                               return v.substring(0, maxDataLength);
-                           }
-                           return v;
-                       });
+        private static Optional<String> sanitize(byte[] data) {
+            StringBuilder sb = new StringBuilder();
+            for (byte b : data) {
+                char c = (char) b;
+                if (!printableAscii(c) && !tabOrLineBreak(c)) {
+                    break;
```
freva, line 118: Return `Optional.empty()` here? Don't want to return ASCII prefix of a binary file

mpolden: PTAL

freva (approved):
:100:

PR #22599, opened 2022-05-13
Use bitvector iterator if attribute term is not used by ranking

toregge:
(i.e. unpack not needed).

@geirst : please review
@baldersheim : FYI


baldersheim (approved):

geirst (commented):
@toregge Please consider adding unit test(s) to verify that bitvector is chosen in the different scenarios in _PostingListSearchContextT<DataT>::
createPostingIterator()_ (line 171).

PR #22600, opened 2022-05-14
Bratseth/constants cleanup 2

bratseth:


config-model/src/test/java/com/yahoo/vespa/model/container/ml/ModelsEvaluatorTest.java
```diff
@@ -18,8 +18,7 @@
 public class ModelsEvaluatorTest {
 
     @Test
-    public void testModelsEvaluatorTester() {
-        assumeTrue(OnnxEvaluator.isRuntimeAvailable());
```
hmusum, line 22: This is used so that test is not run on arm64, where we have no support for onnx at the moment

bratseth: Thanks. Locally, the assumption failed but the test worked ...

PR #22603, opened 2022-05-14
GC unused code and dependencies

baldersheim:
@toregge PR


vdslib/src/vespa/vdslib/distribution/distribution.cpp
```diff
@@ -10,7 +10,6 @@
 #include <vespa/config/print/asciiconfigreader.hpp>
 #include <vespa/vespalib/util/exceptions.h>
 #include <vespa/config-stor-distribution.h>
-#include <algorithm>
```
toregge, line 13: The include statement is needed to get definition of std::sort (cf. PR #21439)

toregge (changes requested):

vdslib/src/vespa/vdslib/distribution/distribution.cpp
baldersheim: K

PR #22613, opened 2022-05-16
Bratseth/constants cleanup 2 take 2

bratseth:


config-model/src/test/java/com/yahoo/vespa/model/container/ml/ModelsEvaluatorTest.java
```diff
@@ -18,8 +18,11 @@
 public class ModelsEvaluatorTest {
 
     @Test
-    public void testModelsEvaluatorTester() {
+    public void testModelsEvaluator() {
+        // Assumption fails but test passes on Intel macs
+        // Assumption fails and test fails on ARM64
```
lesters, line 23: Confusing comment, so this test always fails?

lesters (approved):
LGTM, but the comment in `testModelsEvaluator` is a bit confusing to me

config-model/src/test/java/com/yahoo/vespa/model/container/ml/ModelsEvaluatorTest.java
bratseth: It says "test passes" so I don't know how you summarize it as "always fails".
Anyway, it leaves out linux intel, where (I assume) both the assumption and test succeeds.

lesters: I understand "assumption fails" as that `assumeTrue` fails, and it was confusing how the test then can pass?

bratseth: Yes, that was surprising to me as well. I guess OnnxEvaluator.isRuntimeAvailable() is wrong.

PR #22618, opened 2022-05-16
prepare shared state

havardpe:
needed for non-numeric query features

@geirst please review
@baldersheim FYI (dotproductfeature system test passed locally with this)


geirst (approved):
👍

PR #22645, opened 2022-05-18
Bratseth/model syntax take 2

bratseth:
@lesters please review - same as last just disabling the test that is useful locally but don't work on ARM again


lesters:
👍

PR #22646, opened 2022-05-18
We only need to expose a single libstorage.so. No need to split up in…

baldersheim:
… storage storgae_storageserver and storage_distributor.

@vekterli PR


vekterli (approved):
:+1:

PR #22648, opened 2022-05-18
Add CMake function to install components symlink to fat jar

hakonhall:


functions.cmake
```diff
@@ -691,6 +691,21 @@ function(install_configserver_component NAME)
     install_symlink(lib/jars/${NAME}-${PARAM_CLASSIFIER}.jar conf/configserver-app/components/${NAME}.jar)
 endfunction()
 
+function(install_components_symlink APPNAME NAME)
```
aressem, line 694: Maybe have app in the function name like `install_app_components_symlink` ? This is also more similar to the above function `install_configserver_component`

aressem (approved):

functions.cmake
hakonhall: Alright, then I drop the plural in components too, as in "application component".

aressem (approved):

PR #22653, opened 2022-05-18
Avoid rank zero tensors

bratseth:


bratseth:
@havardpe & @jobergum fyi

lesters:
👍

PR #22655, opened 2022-05-18
Balder/collapse storageframework into storage again

baldersheim:
@vekterli PR


vekterli (approved):
:+1: ideally we'll shrink this sub-library down to zero over time...!

PR #22657, opened 2022-05-19
Rename the 'searchdefinition' package to 'schema'

bratseth:


baldersheim:
This is not reviewable...
It is green => go

PR #22658, opened 2022-05-19
RankProfiles are very expensive memory wise so we need to keep their …

baldersheim:
…lifetime very short.

Only remember exactly what you need.
In this case we only need the constants and the onnxModels.

@hmusum and @bratseth PR


bratseth (approved):

bratseth:
Maybe when there's a reminder in every class I won't do it again.

Please hold off merge until https://github.com/vespa-engine/vespa/pull/22657 it'll be easer to put this on top.

hmusum (approved):

baldersheim:
Will rebase and resolve

PR #22659, opened 2022-05-19
RankProfiles are very expensive memory wise so we need to keep their …

baldersheim:
…lifetime very short.

Only remember exactly what you need.
In this case we only need the constants and the onnxModels.

@hmusum @bratseth PR
This superceeds #22658


hmusum (approved):

bratseth:
👍  thanks

PR #22661, opened 2022-05-19
Support rank profile files in schemas/ and subdirs

hmusum:
Followup after https://github.com/vespa-engine/vespa/pull/22654


bratseth (approved):
LGTM! It works without this though ..,.

PR #22677, opened 2022-05-20
Use method static instead of global to enable use in other static/glo…

baldersheim:
…bal constructions.

@havardpe or @vekterli or @toregge PR


havardpe (approved):

vekterli (commented):
:+1:

PR #22681, opened 2022-05-20
read cluster state while holding lock

havardpe:
will also avoid error messages like 'message discarded due to cluster
being UP'

@vekterli please review


vekterli (approved):
:+1:

PR #22682, opened 2022-05-20
Correctly set scope override for some jupiter deps

jonmv:
@gjoranv or @hakonhall please review and merge. 
@bjorncs FYI (opentest4j was placed below junit-jupiter-api, which contained it, so that entry was just ignored.)

We generate import-package statements for packages exported from junit-jupiter-api, even when when these are used only from inside that bundle. 
Setting the scope of the re-exported dependencies (that's what they are) to test (instead of provided) gets rid of these false imports.


jonmv:
Hmm, think I have a better solution for this.

PR #22683, opened 2022-05-20
wait for locks to be released in the destructor

havardpe:
@vekterli please review

while this looks like exactly what we need to avoid the shutdown race and it passes all unit tests... I still get a faint feeling that it is too good to be true...

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -1174,7 +1174,6 @@ FileStorHandlerImpl::Stripe::release(const document::Bucket & bucket,
     if (!entry._exclusiveLock && entry._sharedLocks.empty()) {
         _lockedBuckets.erase(iter); // No more locks held
     }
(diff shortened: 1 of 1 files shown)


vekterli (approved):
:+1: should be safe to try

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -78,7 +78,10 @@ FileStorHandlerImpl::FileStorHandlerImpl(uint32_t numThreads, uint32_t numStripe
     _component.registerMetricUpdateHook(*this, framework::SecondTime(5));
 }
 
-FileStorHandlerImpl::~FileStorHandlerImpl() = default;
+FileStorHandlerImpl::~FileStorHandlerImpl()
+{
+    waitUntilNoLocks();
```
toregge, line 83: Consider not releasing lock before calling notify_all() at end of` FileStoreHandlerImpl::Stripe::release()`

vekterli (approved):

toregge (approved):

PR #22685, opened 2022-05-20
Avoid override, re-export packages instead

jonmv:
@gjoranv please review. @bjorncs FYI.

### DIFF
--- a/hosted-tenant-base/pom.xml
+++ b/hosted-tenant-base/pom.xml
@@ -235,12 +235,11 @@
                             <!-- note: ordering affects how overrides are evaluated! -->
                             <testBundleScopeOverrides>
                                 ${extraTestBundleScopeOverrides},
-                                com.yahoo.vespa:application:test,
-                                com.yahoo.vespa:container-test:runtime,
                                 org.junit.jupiter:junit-jupiter-api:provided,
-                                org.opentest4j:opentest4j:test,
                                 org.junit.jupiter:junit-jupiter-engine:test,
                                 org.junit.vintage:junit-vintage-engine:test,
+                                com.yahoo.vespa:application:test,
+                                com.yahoo.vespa:container-test:runtime,
                                 com.yahoo.vespa:vespa-feed-client:runtime, <!-- prevent effective compile scope of vespa-feed-client in test bundle -->
                                 com.yahoo.vespa:vespa-feed-client-api:provided,
                                 com.yahoo.vespa:tenant-cd-api:provided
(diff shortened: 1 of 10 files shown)


jonmv:
These packages are all compile-scoped into junit-jupiter-api, and have an import-package statement in that bundle. I suspect we could have this trouble with all of them, depending on what parts of the junit-jupiter-api user code ends up referring to.

hosted-tenant-base/pom.xml
```diff
@@ -235,12 +235,11 @@
...
                             <testBundleScopeOverrides>
                                 ${extraTestBundleScopeOverrides},
-                                com.yahoo.vespa:application:test,
-                                com.yahoo.vespa:container-test:runtime,
                                 org.junit.jupiter:junit-jupiter-api:provided,
-                                org.opentest4j:opentest4j:test,
                                 org.junit.jupiter:junit-jupiter-engine:test,
                                 org.junit.vintage:junit-vintage-engine:test,
+                                com.yahoo.vespa:application:test,
+                                com.yahoo.vespa:container-test:runtime,
```
jonmv, line 242: Moved these down, in case they ever end up exposing the above dependencies, as that would shadow those (silently).

hosted-tenant-base/pom.xml
```diff
@@ -235,12 +235,11 @@
                             <!-- note: ordering affects how overrides are evaluated! -->
                             <testBundleScopeOverrides>
                                 ${extraTestBundleScopeOverrides},
-                                com.yahoo.vespa:application:test,
-                                com.yahoo.vespa:container-test:runtime,
                                 org.junit.jupiter:junit-jupiter-api:provided,
-                                org.opentest4j:opentest4j:test,
```
jonmv, line 241: Re-exported instead.

tenant-cd-api/src/main/java/org/junit/jupiter/api/extension/support/package-info.java
```diff
@@ -0,0 +1,9 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+/**
+ * @author jonmv
+ */
+@ExportPackage(version = @Version(major = 5, minor = 8, micro = 1))
+package org.junit.jupiter.api.extension.support;
```
jonmv, line 6: This one seems to have just been forgotten when the original re-exports were added.

gjoranv (approved):

PR #22686, opened 2022-05-20
Avoid wraparound when calculating compacted size.

toregge:
Always flush active filechunk even when not compacting to it to ensure that
lastSyncToken() >= sync token for any remove that affected the compaction.

@baldersheim : please review


baldersheim (approved):
Good catch 🥇

PR #22721, opened 2022-05-23
Revert "Jonmv/get tester logs"

jonmv:
Reverts vespa-engine/vespa#22720

Deactivate tester now needs to depend on copy vespa logs


jonmv:
@freva please review.

freva (commented):
:+1:

PR #22722, opened 2022-05-23
Jonmv/deactivate tester depends on vespa log now

jonmv:

@freva please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
@@ -67,10 +67,12 @@ import static com.yahoo.vespa.hosted.controller.deployment.Step.Status.succeeded
 import static com.yahoo.vespa.hosted.controller.deployment.Step.Status.unfinished;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.copyVespaLogs;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.deactivateTester;
+import static com.yahoo.vespa.hosted.controller.deployment.Step.deployTester;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.endStagingSetup;
 import static com.yahoo.vespa.hosted.controller.deployment.Step.endTests;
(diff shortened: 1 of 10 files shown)


freva (approved):

jonmv:
I wonder though ... if the Vespa log gather from the tester needs to be more robust. We don't even wait for that service to be up during the install step, for some reason—probably because it's on a shared host, or something? 
Also, this may lead to log records being stored slightly out-of-order, I suppose ... but I'm not sure it's easy to improve significantly on this.

jonmv:
PTAL.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -179,11 +183,27 @@ public void updateVespaLog(RunId id) {
...
 
-            Instant from = run.lastVespaLogTimestamp().isAfter(deployment.get().at()) ? run.lastVespaLogTimestamp() : deployment.get().at();
+            Instant deployedAt = run.stepInfo(installInitialReal).or(() -> run.stepInfo(installReal)).flatMap(StepInfo::startTime).orElseThrow();
+            Instant from = run.lastVespaLogTimestamp().isAfter(run.start()) ? run.lastVespaLogTimestamp() : deployedAt.minusSeconds(10);
             List<LogEntry> log = LogEntry.parseVespaLog(controller.serviceRegistry().configServer()
                                                                   .getLogs(new DeploymentId(id.application(), zone),
                                                                            Map.of("from", Long.toString(from.toEpochMilli()))),
                                                         from);
+
+            if (run.hasStep(installTester) && run.versions().targetPlatform().isAfter(new Version("7.589.14"))) { // todo jonmv: remove
```
freva, line 193: Should the version be updated since there was a new release today?

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
jonmv: No, that's fine. It was the version which started setting up the tester logservers.

PR #22724, opened 2022-05-23
Revert "Count number of bits to create a better estimate of number of hits it…"

hmusum:
Reverts vespa-engine/vespa#22716

2 system tests started failing and some performance tests had worse performance after this change, this should be merged if that is not as expected.

@baldersheim  or @toregge please review.


baldersheim:
Not expected at all.... Why hard ....

PR #22737, opened 2022-05-24
try to avoid some int64 wrap-arounds (the simple ones)

havardpe:
also avoid nullptr deref

@vekterli please review


vekterli (approved):
:+1:

PR #22742, opened 2022-05-24
Add flags for file distribution compression features

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -402,6 +402,20 @@ public class Flags {
             "Takes effect on redeployment",
             APPLICATION_ID);
 
+    public static final UnboundStringFlag FILE_DISTRIBUTION_COMPRESSION_ALGORITHM = defineStringFlag(
+            "file-distribution-compression-algorithm", "gzip",
+            List.of("hmusum"), "2022-05-24", "2022-06-24",
+            "Which algorithm to use for compressing file references when distributing files. Valid values: none, gzip",
```
baldersheim, line 408: Only gzip ?

baldersheim (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hmusum: For now, yes. We want zstd, but needs some work to support it. lz4 is probably easier, we'll see.

PR #22744, opened 2022-05-25
Support compressing files (not just directories) in file distribution [run-systemtest]

hmusum:
Use file distribution compression feature flags and support compressing single
files


baldersheim (approved):
How do we detect files that are already compressed when compressing single files ?

hmusum:
We don't, but that's a good point and should be looked into

PR #22752, opened 2022-05-25
fix undefined behavior in messagebus unit tests

havardpe:
also avoids hanging/leaving processes after running all unit tests
with UBSAN without recovery.

@vekterli please review


vekterli (approved):
:+1:

PR #22763, opened 2022-05-25
Require jackson2-2.12.6.1

baldersheim:
@bjorncs or @gv or @hmusum PR

PR #22768, opened 2022-05-26
Revert "Require jackson2-2.12.6.1"

hmusum:
Reverts vespa-engine/vespa#22763

Met some issues when trying to do the needed changes in internal repo, revert until it can be looked into


jonmv (commented):
🎷

PR #22779, opened 2022-05-28
Revert "Revert "Require jackson2-2.12.6.1""

baldersheim:
@hmusum PR
Trying this one again

### DIFF
--- a/container-dependency-versions/pom.xml
+++ b/container-dependency-versions/pom.xml
@@ -71,16 +71,6 @@
                 <groupId>com.fasterxml.jackson.jaxrs</groupId>
                 <artifactId>jackson-jaxrs-json-provider</artifactId>
                 <version>${jackson2.version}</version>
-                <exclusions>
-                    <exclusion>
-                        <groupId>jakarta.activation</groupId>
-                        <artifactId>jakarta.activation-api</artifactId>
-                    </exclusion>
-                    <exclusion>
-                        <groupId>jakarta.xml.bind</groupId>
-                        <artifactId>jakarta.xml.bind-api</artifactId>
-                    </exclusion>
-                </exclusions>
             </dependency>
             <dependency>
                 <groupId>com.fasterxml.jackson.module</groupId>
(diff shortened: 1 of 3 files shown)


hmusum (approved):

baldersheim:
@bjorncs or @gjoranv Can you verify if this is the correct way of excluding jakarta stuff ?

bjorncs (approved):

gjoranv:
I'm looking into this now. Please don't merge.

container-dependency-versions/pom.xml
```diff
@@ -71,6 +71,16 @@
...
+                <exclusions>
+                    <exclusion>
+                        <groupId>jakarta.activation</groupId>
+                        <artifactId>jakarta.activation-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>jakarta.xml.bind</groupId>
+                        <artifactId>jakarta.xml.bind-api</artifactId>
+                    </exclusion>
+                </exclusions>
```
gjoranv, line 83: Jakarta should already be excluded in the necessary pom files, so this should be unnecessary. I have verified that neither `container-dev` nor `container` lead jakarta if these exclusions are removed. Also, if a user adds an explicit dependency on jackson-jaxrs-json-provider, these exclusions will take effect for them. We want to affect user deps as little as possible.

BTW, the jackson-jaxrs-json-provider will probably be completely removed from Vespa 8.

gjoranv (changes requested):

container-dependency-versions/pom.xml
baldersheim: But I would assume that if you added a dependency that is provided, you would not need to add any exclusions for it yourself. Then it should be done by the provider.
I have forund 3 places we do that ourselves.

gjoranv: No, we don't want to interfere with users' dependencies any more than we have to. We are actively trying to reduce any overlap between our maven deps and the users'. This will be gone on Vespa 8 anyway. I will add commits for what I consider the best workaround.

gjoranv: Users depend on `container`, where jakarta is excluded. If they for some reason want to add their own dependencies, we should not interfere.

PR #22786, opened 2022-05-30
Jonmv/dep cleanup

jonmv:
@bjorncs or @gjoranv please review.

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -114,6 +114,7 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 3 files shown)


bjorncs (approved):

gjoranv:
Note that order is significant in maven, so we should compare dependency:tree output before and after such changes for 'container' and 'container-dev'.

gjoranv (approved):
Great!

bjorncs (approved):
Fewer artifacts exposed is great 👍 🥇

jonmv:
I ran what we have of tests using `application`, in our two Vespa repos, without finding any container breakages. I suppose this _could_ break someone using using `httpclient` in their own unit tests, without depending on it, but that's an acceptable tradeoff, I guess :)

gjoranv:
httpclient is explicitly given in com.yahoo.vespa.tenant:base which is used by most hosted vespa apps.

PR #22787, opened 2022-05-30
backwards compatibility

arnej27959:
if pos.radius is negative, add the geoLocation item as rank-only

@baldersheim please review

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/QueryTree.java
+++ b/container-search/src/main/java/com/yahoo/search/query/QueryTree.java
@@ -110,6 +110,19 @@ public class QueryTree extends CompositeItem {
     // -------------- Facade
 
     /**
+     * Modifies this query to become the current query RANK with the given item.
+     *
+     * @return the resulting root item in this
+     */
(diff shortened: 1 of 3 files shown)


baldersheim (approved):
Looks good to me, but I @bratseth should have a look too.
There is no extisting interface for adding a rank term ?

container-search/src/main/java/com/yahoo/search/query/QueryTree.java
```diff
@@ -109,6 +109,19 @@ public boolean isEmpty() {
 
     // -------------- Facade
 
+    /**
+     * Modifies this query to become the current query RANK with the given item.
+     *
+     * @return the resulting root item in this
+     */
+    public Item rankWith(Item item) {
```
bratseth, line 117: Wouldn't it be more logical with "withRank(item)"?

But I think it would be best to just name it "rank" to match with the "and" methods which are also in this facade.

arnej27959:
and() works even with empty tree, so 'withRank' feels slightly better to me.

PR #22791, opened 2022-05-30
Remove support for 'aliases' element

mpolden:
@gjoranv


gjoranv (approved):
Have you checked for usage in system tests?

PR #22792, opened 2022-05-30
Exact doc type matching for document selections on Vespa 8

vekterli:
@bratseth please review


vekterli:
This is a follow-up of #20519

bratseth (approved):

PR #22793, opened 2022-05-30
Reject deployment spec using elements deprecated on previous major

mpolden:
@jonmv

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageValidator.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageValidator.java
@@ -59,6 +59,16 @@ public class ApplicationPackageValidator {
         validateEndpointChange(application, applicationPackage, instant);
         validateCompactedEndpoint(applicationPackage);
         validateSecurityClientsPem(applicationPackage);
+        validateDeprecatedElements(applicationPackage);
+    }
+
+    /** Verify that deployment spec does not use elements deprecated on a major version older than compile version */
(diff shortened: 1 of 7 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageValidator.java
```diff
@@ -59,6 +59,16 @@ public void validate(Application application, ApplicationPackage applicationPack
         validateEndpointChange(application, applicationPackage, instant);
         validateCompactedEndpoint(applicationPackage);
         validateSecurityClientsPem(applicationPackage);
+        validateDeprecatedElements(applicationPackage);
+    }
+
+    /** Verify that deployment spec does not use elements deprecated on a major version older than compile version */
+    private void validateDeprecatedElements(ApplicationPackage applicationPackage) {
+        for (var deprecatedElement : applicationPackage.deploymentSpec().deprecatedElements()) {
+            if (applicationPackage.compileVersion().isEmpty()) continue;
```
jonmv, line 68: Hmm ... so should we try to set a compile version for non-java projects as well, then? Assign the controller version at the time of submission, or use the one from deployment.xml if set?

jonmv (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageValidator.java
mpolden: Yes, that makes sense. I'll fix.

jonmv (approved):

jonmv:
Great!

PR #22800, opened 2022-05-31
Bratseth/schema info take 2

bratseth:


arnej27959:
looks good to me, @baldersheim should check the streaming cluster logic

baldersheim (approved):

PR #22804, opened 2022-05-31
Include vespa/searchcommon/attribute/config.h, needed for syntax chec…

toregge:
…king call to search::attribute::Config constructor.

@baldersheim : please review


baldersheim (changes requested):
Default values in parameter lists is not nice.
I suggest dropping it and add an extra constructor taking a single argument and implementation in hpp or cpp file adding the default config there.

PR #22836, opened 2022-06-01
Don't calculate score or weights when unpacking for a term in

toregge:
a filter field or for a term not used by ranking.

@geirst : please review


geirst (approved):
👍

PR #22846, opened 2022-06-01
test/provided apache client bundle for cloud users

jonmv:
@bjorncs


jonmv:
@gv FYI

jonmv:
This makes the local and tester-container "class paths" both contain the apache libraries, and they're needed by both.

bjorncs (approved):

PR #22850, opened 2022-06-01
Update flag lifetime, and GC some now defunct flags.

baldersheim:
@geirst or @vekterli or @hmusum PR


hmusum (approved):
LGTM, if some of these flags are permanent they could be moved to PermanentFlags

baldersheim:
Most of them will be removed on 8.

PR #22854, opened 2022-06-01
- Optimize get_tensor_ref optimizing for no branches on happy path.

baldersheim:
- Also drop check for reference as that is done in the next called method.
- Inline DirectTensorStore::get_tensor.

@toregge PR
@havardpe PR
Do we really need  acquire_entry_ref in get_tensor_ref.
In addition it does 2 acquire loads,
EntryRef acquire_entry_ref(DocId doc_id) const noexcept { return _refVector.acquire_elem_ref(doc_id).load_acquire(); }

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/direct_tensor_attribute.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/direct_tensor_attribute.cpp
@@ -122,10 +122,10 @@ DirectTensorAttribute::getTensor(DocId docId) const
 const vespalib::eval::Value &
 DirectTensorAttribute::get_tensor_ref(DocId docId) const
 {
-    if (docId >= getCommittedDocIdLimit()) return *_emptyTensor;
+    if (docId >= getCommittedDocIdLimit()) { return *_emptyTensor; }
 
     auto ptr = _direct_store.get_tensor(acquire_entry_ref(docId));
-    if ( ptr == nullptr) return *_emptyTensor;
+    if ( ptr == nullptr) { return *_emptyTensor; }
 
     return *ptr;
 }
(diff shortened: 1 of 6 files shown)


baldersheim:
@geirst and @vekterli FYI

searchlib/src/vespa/searchlib/tensor/direct_tensor_attribute.cpp
```diff
@@ -122,24 +122,18 @@ DirectTensorAttribute::getTensor(DocId docId) const
...
-        ref = acquire_entry_ref(docId);
-    }
-    if (ref.valid()) {
-        auto ptr = _direct_store.get_tensor(ref);
-        if (ptr) {
-            return *ptr;
-        }
-    }
-    return *_emptyTensor;
+    if (docId >= getCommittedDocIdLimit()) return *_emptyTensor;
```
toregge, line 125: Please keep braces in if statements, even though the body is a simple statement.

searchlib/src/vespa/searchlib/tensor/direct_tensor_attribute.cpp
```diff
@@ -122,24 +122,18 @@ DirectTensorAttribute::getTensor(DocId docId) const
...
+
+    auto ptr = _direct_store.get_tensor(acquire_entry_ref(docId));
+    if ( ptr == nullptr) return *_emptyTensor;
```
toregge, line 128: Please keep braces in if statements, even though the body is a simple statement.

toregge, line 128: Please keep braces in if statements, even though the body is a simple statement.

toregge (approved):

havardpe (approved):

searchlib/src/vespa/searchlib/tensor/direct_tensor_attribute.cpp
baldersheim: Fixed

PR #22856, opened 2022-06-02
make IndexModel predictable

arnej27959:
@bratseth please review
@gjoranv FYI

should fix unit test in other repo.


bratseth:
Thanks!

PR #22859, opened 2022-06-02
If all else is equal the compiler will assume the if branch is more l…

baldersheim:
…ikely than the else branch.

This is based on experience since gcc 2.8 and til today, and is not a definitive truth. Just more likely.
if (criteria) {
    most_likely_executed code
} else {
    less_likely_executed code
}

While a single branched if is assumed less likely.

@toregge or @havardpe PR


havardpe (approved):
Also consider using [[likely]] and [[unlikely]] directives. The documentation seems to indicate that they have more effect than branch ordering.

toregge (approved):

PR #22874, opened 2022-06-02
Let deadlock detector dump stack of potentially deadlocked threads

vekterli:
@geirst please review
@havardpe @baldersheim FYI

Enable cross-thread stack tracing as part of signal handler init
code in both storage and proton daemons.

Make deadlock detector parameters live configurable. Remove existing
`restart` config definition annotations to reflect this.

Remove dumping of bucket DB locks which hasn't really worked for a
long time now.


geirst (approved):
👍

PR #22896, opened 2022-06-03
Don't provide Jetty bundles from 'container' artifact

bjorncs:
Add dependency on 'jetty-http' with scope test instead of adding false dependencies with 'container-test'.

Next step is to embedded the core Jetty artifacts inside container-core + create wrapper for multipart-parsing.

### DIFF
--- a/container-dependency-versions/pom.xml
+++ b/container-dependency-versions/pom.xml
@@ -170,21 +170,6 @@
                 <version>${bouncycastle.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-http</artifactId>
-                <version>${jetty.version}</version>
-            </dependency>
-            <dependency>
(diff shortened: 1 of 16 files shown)


hmusum (approved):

container-dependency-versions/pom.xml
```diff
@@ -169,21 +169,6 @@
...
-            <dependency>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-io</artifactId>
-                <version>${jetty.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-util</artifactId>
-                <version>${jetty.version}</version>
-            </dependency>
```
gjoranv, line 186: I strongly suggest to move these to parent/pom.xml, and remove jetty.version in all the other poms.

PR #22904, opened 2022-06-03
Disallow impossible redundancy

hakonhall:
But allow it for hosted deployment in dev if the count attribute of the nodes element is absent, to support deployment of self-hosted apps to dev.


bratseth (approved):
👍

PR #22910, opened 2022-06-03
Let ImportedAttributesRepo return a reference to a shared_ptr instead…

baldersheim:
… of a shared_ptr since that is all you need.

@toregge or @havardpe PR


toregge (approved):

havardpe (commented):
Consider inheriting from shared_from_this and return a normal reference instead

PR #22917, opened 2022-06-05
Remove vespa-http-client on Vespa 8

bratseth:


gjoranv (approved):
I added a commit for vespa.spec.

Updates are also needed in vespa-yahoo.

bratseth:
Thanks, yes.

bratseth:
should be ok now

PR #22927, opened 2022-06-06
Remove vespa-http-client on Vespa 8

bratseth:


bratseth:
Simultaneous, to yours but merging this too because on the link updates.

gjoranv (commented):
👍

PR #22946, opened 2022-06-07
Remove application/ subdir support on Vespa 8

bratseth:


hmusum:
I think we can only do this if no application in hosted uses this subdir today, if we have verified that OK, otherwise I guess we need to wait until no app uses it?

bratseth:
I thought you knew the status on this one but maybe it's somebody else.
We need to do it now or wait until Vespa 9. Since they need to rebuild the package for Vespa 8 anyway, I don't think it's strictly necessary to check first.

bratseth:
Or wait, this will be disable it for Vespa 7 apps as well I guess. Yes, that won't necessarily work, but also we shouldn't allow it for Vespa 8 built apps ... Some more complicated work is needed here then.

hmusum:
I was thinking of those apps that are still on Vespa 7, yes. But the changes in maven plugin were done in 2019. Surely all apps must have been rebuilt since then? Or are there other places we do/did this?

bratseth:
They could zip themselves, but quite unlikely then. Maybe we just check if Vespa Cloud is ok and if so just merge this?

hmusum:
Sounds good to me

bratseth:
No application/ root folder in any Vespa Cloud prod apps.

mpolden:
https://github.com/vespa-engine/vespa/blob/8/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java#L813-L819 should be removed too, and the ability to create such packages in `ApplicationPackageBuilder`.

PR #22955, opened 2022-06-07
Stop suppressing warnings when using config library

hmusum:


hmusum:
Done in https://github.com/vespa-engine/vespa/pull/22956

PR #22957, opened 2022-06-07
Remove TODOs

hmusum:
These are for compatibility with Vespa 7 apps, will be removed when
there are no more Vespa 7 apps in hosted.


bratseth (commented):
I think we should have a TODO on it even though it doesn't say "Vespa 8"?

PR #22960, opened 2022-06-07
Remove application folder support

bratseth:


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
```diff
@@ -337,23 +337,15 @@ private static byte[] buildMeta(Version compileVersion) {
...
         ByteArrayOutputStream zip = new ByteArrayOutputStream();
         try (ZipOutputStream out = new ZipOutputStream(zip)) {
             out.setLevel(Deflater.NO_COMPRESSION); // This is for testing purposes so we skip compression for performance
-            writeZipEntry(out, dir + "deployment.xml", deploymentSpec());
-            writeZipEntry(out, dir + "validation-overrides.xml", validationOverrides());
-            writeZipEntry(out, dir + "search-definitions/test.sd", searchDefinition());
-            writeZipEntry(out, dir + "build-meta.json", buildMeta(compileVersion));
+            writeZipEntry(out, "deployment.xml", deploymentSpec());
+            writeZipEntry(out, "validation-overrides.xml", validationOverrides());
+            writeZipEntry(out, "search-definitions/test.sd", searchDefinition());
```
hmusum, line 345: `search-definitions`? Shouldn't it be either `schemas` or `searchdefinitions` (if there is some backwards compatibility issue stopping us from using `schemas`). Or maybe it's not used at all in controller, if so it can be removed.

hmusum (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/deployment/ApplicationPackageBuilder.java
bratseth: No issue changing this I guess, I didn't look for it.

PR #22967, opened 2022-06-07
Remove maxFirstPass support on Vespa 8

bratseth:


documentapi/src/main/java/com/yahoo/documentapi/messagebus/MessageBusVisitorSession.java
```diff
@@ -1107,11 +1102,8 @@ private void handleCreateVisitorReply(CreateVisitorReply reply) {
             trace.getRoot().addChild(reply.getTrace().getRoot());
         }
 
-        // TODO: Vespa 8 remove this unused functionality
-        if (params.getDynamicallyIncreaseMaxBucketsPerVisitor()
-                && (reply.getVisitorStatistics().getDocumentsReturned()
-                    < params.getMaxFirstPassHits() / 2.0))
-        {
+        // TODO: Vespa 8 remove this unused functionality (?)
+        if (params.getDynamicallyIncreaseMaxBucketsPerVisitor()) {
```
vekterli, line 1106: Since the other part of the condition was removed, I think this entire branch should go away as well. I don't know of anyone using this, and semantics-wise nothing changes for the client if we remove it entirely

vekterli (approved):
:+1: consider swinging the axe at the mentioned code block

PR #22997, opened 2022-06-08
Update default versions in VERSION and VersionTagger MERGEOK

gjoranv:
Don't know if this is ok, but bundle-plugin-test currently requires 8 major version.

(Merge the usual way from now on.)


hmusum (approved):

arnej27959 (commented):
👍

PR #23002, opened 2022-06-08
Clean up 'searchchain' and 'searchchains' syntax in services.xml

gjoranv:
This syntax has been disallowed in the xml schema for a long time, but never cleaned up.


hmusum (approved):
💯

bjorncs (approved):

PR #23006, opened 2022-06-08
Only enable -XX:+UseTransparentHugePages if heap is at least 2x hugep… [run-systemtest]

baldersheim:
…age size.

@hakonhall and @hmusum PR

### DIFF
--- a/config-proxy/src/main/sh/vespa-config-ctl.sh
+++ b/config-proxy/src/main/sh/vespa-config-ctl.sh
@@ -114,7 +114,7 @@ case $1 in
         # TODO Better way of extracting heap size arguments.
         heap_min=32
         heap_max=128
-        jvmopts="-Xms$(heap_min)M -Xmx$(heap_max)M -XX:+PreserveFramePointer $(get_jvm_hugepage_settings $heap_max) -XX:CompressedClassSpaceSize=32m -XX:MaxDirectMemorySize=32m -XX:ThreadStackSize=448 -XX:MaxJavaStackTraceDepth=1000 -XX:-OmitStackTraceInFastThrow"
+        jvmopts="-Xms${heap_min}M -Xmx${heap_max}M -XX:+PreserveFramePointer $(get_jvm_hugepage_settings $heap_max) -XX:CompressedClassSpaceSize=32m -XX:MaxDirectMemorySize=32m -XX:ThreadStackSize=448 -XX:MaxJavaStackTraceDepth=1000 -XX:-OmitStackTraceInFastThrow"
 
         VESPA_SERVICE_NAME=configproxy
         export VESPA_SERVICE_NAME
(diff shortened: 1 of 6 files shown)


hmusum (approved):
LGTM

config-proxy/src/main/sh/vespa-config-ctl.sh
```diff
@@ -109,7 +109,12 @@ case $1 in
         nohup nice sbin/vespa-retention-enforcer > ${LOGDIR}/vre-start.log 2>&1 </dev/null &
         configsources=`bin/vespa-print-default configservers_rpc`
         userargs=$VESPA_CONFIGPROXY_JVMARGS
-        jvmopts="-Xms32M -Xmx128M -XX:+PreserveFramePointer -XX:+UseTransparentHugePages -XX:CompressedClassSpaceSize=32m -XX:MaxDirectMemorySize=32m -XX:ThreadStackSize=448 -XX:MaxJavaStackTraceDepth=1000 -XX:-OmitStackTraceInFastThrow"
+        # Note that the hardcoded memory Xmx=128 migh be overriden in VESPA_CONFIGPROXY_JVMARGS,
+        # hence rendering get_jvm_hugepage_settings incorrect, but it is better than not.
+        # TODO Better way of extracting heap size arguments.
```
hakonhall, line 114: Something like:

```

heap_min=32
heap_max=128
for token in $VESPA_CONFIGPROXY_JVMARGS
do
    [[ "$token" =~ ^-X(mx|ms)([0-9]+)(.)$ ]] || continue
    type="${BASH_REMATCH[1]}"
    size="${BASH_REMATCH[2]}"
    unit="${BASH_REMATCH[3],,}" # lower-case
    case "$type" in
        ms)
            case "$unit" in
                k) heap_min=$(( $size / 1024 )) ;;
                m) heap_min="$size" ;;
                g) heap_min=$(( $size * 1024 )) ;;
                *) echo "Warning: Invalid unit in '$token'" >&2 ;;
            esac
            ;;
        mx)
            case "$unit" in
                k) heap_max=$(( $size / 1024 )) ;;
                m) heap_max="$size" ;;
                g) heap_max=$(( $size * 1024 )) ;;
                *) echo "Warning: Invalid unit in '$token'" >&2 ;;
            esac
            ;;
    esac
done
```

hakonhall (approved):

vespabase/src/common-env.sh
```diff
@@ -129,6 +129,35 @@ optionally_reduce_base_frequency() {
...
+get_hugepage_size_mb() {
+    while read -r name size rest
+    do
+        if [[ "$name" =~ ^Hugepagesize:$ ]]
+        then
+          hugepagesize="$size"
+          unit="${rest,,}"
+          break
+        fi
+    done < /proc/meminfo
```
hakonhall, line 141: I think we should only enable -XX:+UseTransparentHugePages if:

1. transparent huge pages is set to madvise (or defer+madvise?)

and then get the transparent huge page size from `/sys/kernel/mm/transparent_hugepage/hpage_pmd_size`.  `Hugepagesize:` from `/proc/meminfo` is actually a HugeTLB setting.

`hpage_pmd_size` depends on the CPU architecture.  On x64 it is hard coded to 2 MB.  On arm64/aarch64 it is a function of the kernel config CONFIG_ARM64_PAGE_SHIFT (subtracting 1 => 0.25x factor).

config-proxy/src/main/sh/vespa-config-ctl.sh
baldersheim: Trying to address that in #23122

PR #23015, opened 2022-06-09
Remove legacy distribution hash fallback

vekterli:
@geirst please review

Was used to handle rolling upgrades between versions with different
semantics a long time ago on the 7 branch.


geirst (approved):
👍

PR #23016, opened 2022-06-09
Update checked in source-of-truth message blob for CreateVisitorReply

vekterli:
@arnej27959 please review
@toregge FYI

Now contains all zero bytes for the ignored legacy fields for 2nd pass visitor document stats.


arnej27959 (approved):
👍

PR #23025, opened 2022-06-09
Use records in versions package

mpolden:
Trying out records. Looks like it will get rid of a lot of boilerplate. Only
downside I see is that there isn't a way to add Javadoc to a getter without
declaring the method explicitly.

@jonmv


jonmv (approved):
🎉

jonmv:
Just need to be careful when converting classes with custom hashcode/equals :)

PR #23032, opened 2022-06-09
Change user query default from ALL to WEAKAND on Vespa 8

bratseth:


arnej27959 (approved):
just one minor doubt 👍

PR #23033, opened 2022-06-09
Disallow model building failures on any version with a major MERGEOK

jonmv:
…y a deployed or target version

@bratseth and @hmusum please review and merge.


hmusum (approved):

bratseth (approved):

bratseth:
This test depends on Vespa being on the right major now, which it shouldn't. But fixed tests to let that pass for now.

bratseth:
... hm no. I guess head changed to major 8 by now. Ok, I'm out of time on this for right now.

jonmv:
No worries, I'll figure it out :)

jonmv:
Thanks!

jonmv:
I added a verification that the wanted version has the same major as what's specified in `deployment.xml`. This directly conflicts with the second of those failing tests, so I removed that test. I don't understand why we should try to deploy an application on a different major than what they have asked for. 

The first test was failing because it hadn't set the wanted version either, so it defaulted to an 8-version, which was different from the major 7 specified in the deployment.xml. Again, I don't think we ever do this in hosted Vespa, and if we do, it's wrong. For self-hosted, there isn't even a deployment.xml to specify a major version in, so this check should never fail.

bratseth (approved):

PR #23034, opened 2022-06-09
flip default

arnej27959:
@gjoranv review - very late addition to flipped defaults
should be in release notes, and also for log archive


gjoranv (approved):

gjoranv:
Update doc: https://github.com/vespa-engine/documentation/pull/2092

PR #23037, opened 2022-06-09
Use static create methods instead of having many public constructor o…

baldersheim:
…nly used for testing.

@hmusum or @bjorncs PR


hmusum (approved):
👍 , one suggestion

PR #23046, opened 2022-06-10
Bjorncs/vespa enforcer extensions

bjorncs:

### DIFF
--- a/vespa-enforcer-extensions/src/main/java/com/yahoo/vespa/maven/plugin/enforcer/DependencyEnforcer.java
+++ b/vespa-enforcer-extensions/src/main/java/com/yahoo/vespa/maven/plugin/enforcer/DependencyEnforcer.java
@@ -1,0 +1,150 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.maven.plugin.enforcer;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.versioning.ArtifactVersion;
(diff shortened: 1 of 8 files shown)


vespa-enforcer-extensions/src/main/java/com/yahoo/vespa/maven/plugin/enforcer/DependencyEnforcer.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * Enforces that all expected dependencies are present.
+ * Fails by default for rules that do not match any dependencies.
+ * Similar to the built-in banned-dependency rule in maven-enforcer-plugin.
+ *
+ * @author bjorncs
+ */
+public class DependencyEnforcer implements EnforcerRule {
```
gjoranv, line 37: Suggest to rename to _EnforceDependencies_ for consistency with most existing enforcer rules.. Rule name should also be _enforceDependencies_. Where is the actual rule name defined, is it just a non-capitalized version of the class name?

gjoranv (changes requested):

vespa-enforcer-extensions/src/main/java/com/yahoo/vespa/maven/plugin/enforcer/DependencyEnforcer.java
bjorncs: Yes, there is some reflection magic under the hood. That is how injection of the `allowed` list is done.

bjorncs:
@gjoranv PTAL

gjoranv (approved):
This is great!

PR #23047, opened 2022-06-10
Jonmv/more test runtime api

jonmv:
@freva please review.

### DIFF
--- a/configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployTest.java
+++ b/configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployTest.java
@@ -277,6 +277,32 @@ public class HostedDeployTest {
      * that are still using features that do not work on version 8.x)
      */
     @Test
+    public void testFailingOnOldMajorButNotNewInUpgradeScenario() {
+        int oldMajor = 7;
+        int newMajor = 8;
+        Version wantedVersion = new Version(newMajor, 1, 2);
+        Version oldVersion = new Version(oldMajor, 2, 3);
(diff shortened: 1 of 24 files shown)


tenant-cd-api/src/main/java/ai/vespa/hosted/cd/Deployment.java
```diff
@@ -11,4 +13,13 @@ public interface Deployment {
     /** Returns an Endpoint in the cluster with the given id. */
     Endpoint endpoint(String id);
 
+    /** The Vespa runtime version of the deployment, e.g., 8.16.32. */
+    String platformVersion();
+
+    /** The build number assigned to the application revision of the deployment, e.g., 496. */
+    long applicationVersion();
```
freva, line 20: Did we settle on a name for this? Build? Revision? Now version?

freva (approved):

tenant-cd-api/src/main/java/ai/vespa/hosted/cd/Deployment.java
jonmv: Hmm yes ... Not sure why I felt this name was more appropriate here. I found `platform` too terse, and `applicationVersion` fits better with `platformVersion`. Not all three terms are used >_<

configserver/src/test/java/com/yahoo/vespa/config/server/deploy/HostedDeployTest.java
```diff
@@ -271,6 +271,32 @@ public void testCreateModelVersionsForManuallyDeployedAppsWhenCreatingFailsForOn
         assertTrue("Newest model for latest major version is always included", factory720.creationCount() > 0);
     }
 
+    /**
```
jonmv, line 274: Urgh

PR #23049, opened 2022-06-10
Stabilize metrics-proxy tests

bjorncs:
Synchronize access to 'response' field. Handle null response.
Ensure output stream is closed. Specify timeout for request connection from pool.


bratseth (approved):
👍

PR #23051, opened 2022-06-10
Don't make changes to the hnsw index when the inserted tensor is unchanged

### LINKED ISSUE #22025: Option to skip hnsw indexing of a tensor which is unchanged from the current value
Some applications regularly reindex a corpus where much of the data is unchanged but there is not easy way to detect unchanged items from the outside.

It would be nice, and quite cheap, to do this detection inside Vespa by comparing the existing and to-be-written tensor values on the content node before reindexing.

geirst:
With this optimization we avoid doing unnecessary costly work, first removing the vector point, then inserting the same point.

This fixes https://github.com/vespa-engine/vespa/issues/22025.

@toregge please review


bratseth:
👍 👍 👍

toregge (approved):

PR #23052, opened 2022-06-10
common memory tools

havardpe:
@vekterli @arnej27959 please review

### DIFF
--- a/vespalib/src/vespa/vespalib/util/memory.h
+++ b/vespalib/src/vespa/vespalib/util/memory.h
@@ -12,25 +12,24 @@
 namespace vespalib {
 
 inline void *memcpy_safe(void *dest, const void *src, size_t n) noexcept {
-    if (n > 0) [[likely]] {
-        memcpy(dest, src, n);
+    if (n == 0) [[unlikely]] {
+        return dest;
     }
-    return dest;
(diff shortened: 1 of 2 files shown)


vespalib/src/vespa/vespalib/util/memory.h
```diff
@@ -11,6 +11,70 @@
 
 namespace vespalib {
 
+inline void *memcpy_safe(void *dest, const void *src, size_t n) noexcept {
+    if (n > 0) [[likely]] {
```
baldersheim, line 15: I prefer having the non likely if be guards instead.
In this case 
if (n==0) { return dest; }
 return memcpy(...);
The it is easier to follow the happy fast path which will be without indentation.

arnej27959 (approved):
looks good to me

vekterli (approved):
:+1: there are some philosophical questions I have not been able to reliably answer regarding how aliasing interacts with accessing memory via the `Unaligned` "proxy" pointer type -> `char[]` _member_ vs. a raw `char` pointer (which is well-defined). But trying to figure this out from the ISO text is a bit like trying to decipher an Elder Scroll.

baldersheim (approved):

PR #23059, opened 2022-06-12
More flexibly instance layout

jonmv:
@freva or @bratseth or @hmusum please review and merge :)


jonmv:
This is meant to allow arranging instances in a DAG as well. The implementation supports it, but it's stopped here.

hmusum (approved):

PR #23061, opened 2022-06-13
Add application id to IllegalArgumentException in AutoscalingMaintainer

hmusum:
No context as it is now, so impossible to find out which app is having issues


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
```diff
@@ -54,7 +55,12 @@ protected double maintain() {
     }
 
     private void autoscale(ApplicationId application, NodeList applicationNodes) {
-        nodesByCluster(applicationNodes).forEach((clusterId, clusterNodes) -> autoscale(application, clusterId, clusterNodes));
+        try {
+            nodesByCluster(applicationNodes).forEach((clusterId, clusterNodes) -> autoscale(application, clusterId, clusterNodes));
+        }
+        catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException("Illegal arguments for " + application + ": " + Exceptions.toMessageString(e));
```
baldersheim, line 62: Is Exceptions.toMessageString preferred over chaining the exceptions ?
Don't you loose initial exception callstack ?

hmusum: Probably better to chain exceptions, yes, fixed.

baldersheim (approved):

PR #23062, opened 2022-06-13
More info in autoscaling advice

hmusum:
Hard to debug why autoscaling does or does not happen, add some more info.


baldersheim (approved):

bratseth:
Where is it you need this? These messages are displayed in the console, and they are already in context of a cluster.

PR #23076, opened 2022-06-13
Revert "Test seems to fail occasionally with ConnectionClosedException"

hmusum:
Reverts vespa-engine/vespa#23074

Build failed several times, reverting


hmusum:
Ah, right. OK, someone needs to look more into this then.

PR #23078, opened 2022-06-13
Allow unnamed modules to access java.util non-opened packages

jonmv:
@bjorncs please review.


bjorncs:
@gjoranv Do we want to open more packages for reflection?

gjoranv:
Why are we doing this?

PR #23079, opened 2022-06-13
Set context class loader for JUnit

jonmv:
@bjorncs please review.


bjorncs (approved):

jonmv:
Broken something more critical in the JUnit reflection stack ...

PR #23090, opened 2022-06-14
GC unused and deprecated code

baldersheim:
@bratseth or @gjoranv or @bjorncs PR
forgotten ?


bratseth:
Not forgotten - this was deprecated very soon before 8.0. That said it's unlikely to be called by anyone so I may be too conservative here.

baldersheim:
April 21st are ages ago - 7+ weeks => agree, a bit on the conservative side.

gjoranv:
I couldn't find any occurrences of `new ExecutionFactory` in my (now rather outdated) snapshot of hosted apps. The problem now is that the first 8 has been release, as I see it.

baldersheim:
Some pragmatism, especially as no  opensource release has been made yet.

gjoranv:
👍

PR #23094, opened 2022-06-14
Arnej/fflag log compress

arnej27959:
@hmusum please review - should in theory be no visible changes
@bratseth @gjoranv @jonmv FYI


hmusum (approved):
👍

PR #23096, opened 2022-06-14
Refuse config only when there is a possibility of incompatible MERGEOK

jonmv:
... Java bundles

@hmusum please review.


hmusum (approved):

gjoranv:
LGTM!

PR #23098, opened 2022-06-14
Add control of wether shared string repo will reclaim unused enums, o…

baldersheim:
…r if they will just grow.

@havardpe PR


havardpe (commented):
I think the feature should be inverted to allow the flag to be false by default.

baldersheim:
@havardpe PTAL

havardpe (approved):

PR #23102, opened 2022-06-15
common default compression type

arnej27959:
@hmusum please review
should be no-op for HV


hmusum:
I think we need to handle the case where some files are already compressed when we switch compression type (we have observed files being double compressed (files with suffix like e.g. ".zst.gz", I guess the other way around is also possible).

hmusum (approved):

PR #23111, opened 2022-06-15
Kkraune/query builder misc

kkraune:
- update to new Vespa look
- update icon set
- remove broken feature in query api lookup


ldalves (approved):
not sure about using this font in the new app... we'll see how this will fit.

PR #23114, opened 2022-06-15
Use a compatible fallback version

jonmv:
@mpolden or @freva please review and merge.


jonmv:
Using `systemVersion` as a last resort fallback for platform doesn't work when the system has upgraded to a version incompatible with something older, and an app on this older version has no production deployments. So we must use the newest compatible version as a fallback instead.

mpolden (approved):
👍

PR #23116, opened 2022-06-15
Allow OS upgrades to start earlier in CD systems

mpolden:
To reduce chance of slow convergence due to upgrade during working hours.

@hmusum


hmusum (approved):
Nice, thanks!

PR #23137, opened 2022-06-16
enable making string handles directly from numbers

havardpe:
used in peek/slice to avoid having to convert small numbers to strings
before resolving the enum values.

@baldersheim please review
@arnej27959 FYI


baldersheim (approved):
Perfect 🥇

PR #23138, opened 2022-06-16
Bjorncs/metrics proxy

bjorncs:


baldersheim:
Hmm, I think we have to discuss this tomorrow.
How does this impact buffering ? I did some work last autumn to avoid having to read the whole payload before starting to consume it. Is that streaming behaviour preserved ?

bjorncs:
> Hmm, I think we have to discuss this tomorrow. How does this impact buffering ? I did some work last autumn to avoid having to read the whole payload before starting to consume it. Is that streaming behaviour preserved ?

Yes, no need for async client for that. The buffered input stream should buffer response content like previously.

baldersheim (approved):
K 👍

PR #23139, opened 2022-06-16
Java 17 examples

andreer:
There are more changes from 11 to 17 ... Would love to see more examples! Some I was not able to find good cases for are:
- The Vector API, see https://openjdk.org/jeps/414
- String::indent and String::transform
- File::mismatch
- Collectors.teeing

### DIFF
--- a/config/src/main/java/com/yahoo/vespa/config/protocol/SlimeTraceDeserializer.java
+++ b/config/src/main/java/com/yahoo/vespa/config/protocol/SlimeTraceDeserializer.java
@@ -39,19 +39,13 @@ public class SlimeTraceDeserializer {
     }
 
     private static Object decodePayload(Inspector entry) {
-        switch (entry.type()) {
-            case STRING:
-                return entry.asString();
-            case LONG:
-                return entry.asLong();
-            case BOOL:
-                return entry.asBool();
-            case DOUBLE:
-                return entry.asDouble();
-            case DATA:
-                return entry.asData();
-            default:
-                return null;
-        }
+        return switch (entry.type()) {
+            case STRING -> entry.asString();
+            case LONG -> entry.asLong();
+            case BOOL -> entry.asBool();
+            case DOUBLE -> entry.asDouble();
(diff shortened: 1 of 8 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/certificate/EndpointCertificatesTest.java
```diff
@@ -211,14 +211,17 @@ public void reprovisions_certificate_with_added_sans_when_deploying_to_new_zone(
...
+                        <deployment version="1.0">
+                          <instance id="default">
+                            <prod>
+                              <region active="true">aws-us-east-1a</region>
+                              <region active="true">ap-northeast-1</region>
+                            </prod>
+                          </instance>
+                        </deployment>
+                        """
+        );
```
jonmv, line 224: Nice! This went under my radar.

jonmv (approved):
Very nice!

bratseth:
👍

config/src/main/java/com/yahoo/vespa/config/protocol/SlimeTraceDeserializer.java
```diff
@@ -39,19 +39,13 @@ private static long decodeTimestamp(Inspector entry) {
...
-            default:
-                return null;
-        }
+        return switch (entry.type()) {
+            case STRING -> entry.asString();
+            case LONG -> entry.asLong();
+            case BOOL -> entry.asBool();
+            case DOUBLE -> entry.asDouble();
+            case DATA -> entry.asData();
+            default -> null;
```
mpolden, line 48: Consider matching every value as switch expression on enum is exhaustive, i.e. not matching a value is a compile error if there isn't a `default` clause.

mpolden (commented):
👍 There's probably a use-case for sealed classes somewhere too.

A minor downside of using records is that you can't add javadoc to the getter without declaring it.

bjorncs (approved):
🏅

PR #23141, opened 2022-06-16
Remove .sum from metric name

baldersheim:
@vekterli or @yngveaasheim PR


gjoranv (approved):

vekterli (commented):
:+1: so that's where it was hiding..!

PR #23146, opened 2022-06-17
remove TemporarySDDocumentType

arnej27959:
* track converted structs as we make them
* inherit directly instead of by name
* remove code to deal with the no-longer-needed temporaries

@bratseth please review


bratseth (approved):
Great!

PR #23147, opened 2022-06-17
Get the Vespa version for flags from the model version

bratseth:
A flag value which depends on Vespa version should not take
effect immediately when a deployment wants that version
but only for nodes which are actually upgraded to that version.

Take the flag version from the Vespa model version to achieve this.

@hakonhall and @hmusum and/or @arnej27959  please review.

(I'm going to clean this up a bit, but I wanted to keep the change small for now.)


jonmv (approved):
Good idea!

hakonhall (approved):
nice

arnej27959 (approved):
looks good to me

hmusum (approved):
👍

PR #23157, opened 2022-06-18
Return empty log if getting log fails and we just deployed

hmusum:
Return empty log if unable to get log up to 1 minute after deployment (this is what
happens in cd and test zones when new nodes need to be started before any logging is done). Now we get lots of warnings in config server vespa logs due to this

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/LogRetriever.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/LogRetriever.java
@@ -6,8 +6,12 @@ import com.yahoo.container.jdisc.HttpResponse;
 import com.yahoo.yolean.Exceptions;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
-
 import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/LogRetriever.java
```diff
@@ -16,11 +20,20 @@ public class LogRetriever {
...
 
-    public HttpResponse getLogs(String logServerUri) {
+    public HttpResponse getLogs(String logServerUri, Optional<Instant> deployTime) {
         HttpGet get = new HttpGet(logServerUri);
         try {
             return new ProxyResponse(httpClient.execute(get));
         } catch (IOException e) {
+            // It takes some time before nodes are up after first-time deployment, return empty log for up to 1 minute
+            // if getting logs fail
+            if (deployTime.isPresent() && Instant.now().isAfter(deployTime.get().minus(Duration.ofMinutes(1))))
```
olaaun, line 30: Won't this always be true if deployTime is present?
```suggestion
            if (deployTime.isPresent() && Instant.now().isBefore(deployTime.get().plus(Duration.ofMinutes(1))))
```

hmusum: You are right, thanks!

olaaun (approved):

PR #23160, opened 2022-06-20
Remove vespa-jvm-dumper and find-pid from vespa RPM

toregge:
(already present in vespa-base RPM).

@geirst : please review
@bjorncs : FYI


geirst (approved):
👍

PR #23162, opened 2022-06-20
Stop removing vespa user and vespa group when removing vespa-base RPM.

toregge:
@geirst : please review
@aressem  : FYI


geirst (approved):
👍

PR #23164, opened 2022-06-20
Use getActivatedTime() for last deployed time for an app [run-systemtest]

hmusum:
Use getActivatedTime() instead of getCreatedTime in lastDeployTime().
getCreatedTime() gives time a new session was created, not when it was
activated, which is what we usually want.


bratseth (approved):
👍

PR #23184, opened 2022-06-21
Inline hot path and keep error handling in .cpp file.

baldersheim:
@vekterli PR

### DIFF
--- a/vespalib/src/vespa/vespalib/util/compress.h
+++ b/vespalib/src/vespa/vespalib/util/compress.h
@@ -27,12 +27,35 @@ public:
      * @param unsigned number to compute compressed size of in bytes.
      * @return Will return the number of bytes this positive number will require
      **/
-    static size_t compressedPositiveLength(uint64_t n);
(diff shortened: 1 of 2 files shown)


vespalib/src/vespa/vespalib/util/compress.h
```diff
@@ -82,6 +105,9 @@ class Integer {
         }
         return numbytes;
     }
+private:
+    [[ noreturn ]] static void too_big(int64_t n);
+    [[ noreturn ]] static void too_big(uint64_t n);
```
vekterli, line 110: Consider renaming to `throw_too_big`, or similar. Just `too_big` could at a glance be easy to confuse with a function that just _checks_ if a number is too big.

vekterli (approved):
:+1:

vespalib/src/vespa/vespalib/util/compress.h
baldersheim: Good point.

PR #23185, opened 2022-06-21
Log when deconstruction of mbus server provider is complete

bjorncs:
Useful for debugging failed component graph deconstruction caused by
mbus server shutdown hanging.


baldersheim (approved):

container-messagebus/src/main/java/com/yahoo/container/jdisc/messagebus/MbusServerProvider.java
```diff
@@ -49,9 +47,12 @@ public MbusServer get() {
     @Override
     public void deconstruct() {
         log.log(Level.INFO, "Deconstructing mbus server: " + server);
+        long start = System.currentTimeMillis();
```
baldersheim, line 50: Perhaps use nanTime instead that will only move forward.

baldersheim: Yes, but when you do System.currentTime() - start, you might get surprised :)

PR #23190, opened 2022-06-21
Revert "Synchronize access to sessionStateWatchers map [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#23177

Seems like this introduced a possible deadlock, need to look more into this.


frodelu (approved):

hmusum:
Fixed in https://github.com/vespa-engine/vespa/pull/23446 by removing the need for synchronization

PR #23194, opened 2022-06-22
Jonmv/dependencies fixing

jonmv:
@gjoranv @bjorncs please review and merge.

### DIFF
--- a/container-test/pom.xml
+++ b/container-test/pom.xml
@@ -36,14 +36,16 @@
         </exclusion>
       </exclusions>
     </dependency>
-
-    <!-- All dependencies that should be visible in test classpath, but not compile classpath,
-         for user projects must be added in compile scope here. These dependencies are explicitly excluded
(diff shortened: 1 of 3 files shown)


container-test/pom.xml
```diff
@@ -30,12 +30,20 @@
...
     </dependency>
 
     <!-- All dependencies that should be visible in test classpath, but not compile classpath,
          for user projects must be added in compile scope here. These dependencies are explicitly excluded
          (or set to non-compile scope) in the container and/or container-dev modules. -->
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk15on</artifactId>
+    </dependency>
     <dependency>
```
gjoranv, line 47: Suggest to move this above the multi-line comment, and add a separate comment saying that this is added back for the exclusion above to work around a presumed bug in maven dep res.

gjoranv (approved):
Please consider updating container-test as commented. Otherwise, this is great!

PR #23195, opened 2022-06-22
Make sure to return response when request times out

hmusum:
Return with new error code 2. Clients only check for error code 0,
so a new error code is handled implicitly also by old clients


hmusum:
ping @baldersheim

baldersheim (approved):

baldersheim:
This one passed me by...

PR #23200, opened 2022-06-22
fix more undefined behavior

havardpe:
@vekterli please review
@baldersheim FYI


vespalib/src/vespa/vespalib/stllike/asciistream.cpp
```diff
@@ -429,22 +429,30 @@ printInt(unsigned long long r, char * tmp, uint8_t i)
     return i;
 }
 
+unsigned long long normalize(long long v, bool &negative) {
+    if (v < 0) {
+        negative = true;
+        if (v == std::numeric_limits<long long>::min()) {
+            return v;
```
vekterli, line 436: This is sufficiently magic that a comment would be good 🦄

vekterli (approved):
:+1:

PR #23209, opened 2022-06-23
rank and summary features are never moved as json anymore.

baldersheim:
@havardpe or @arnej27959 PR

### DIFF
--- a/searchsummary/src/vespa/searchsummary/docsummary/rankfeaturesdfw.cpp
+++ b/searchsummary/src/vespa/searchsummary/docsummary/rankfeaturesdfw.cpp
@@ -2,28 +2,21 @@
 
 #include "rankfeaturesdfw.h"
 #include "docsumstate.h"
-#include <vespa/searchlib/common/packets.h>
 #include <vespa/vespalib/data/slime/cursor.h>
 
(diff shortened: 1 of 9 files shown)


searchsummary/src/vespa/searchsummary/docsummary/rankfeaturesdfw.cpp
```diff
@@ -2,36 +2,30 @@
...
+    if ( !state->_rankFeatures ) {
         state->_callback.FillRankFeatures(state, _env);
         if (state->_rankFeatures.get() == nullptr) { // still no rank features to write
             return;
         }
     }
     const FeatureSet::StringVector & names = state->_rankFeatures->getNames();
     const FeatureSet::Value * values = state->_rankFeatures->getFeaturesByDocId(docid);
-    if (type == RES_FEATUREDATA && values != nullptr) {
+    assert(type == RES_FEATUREDATA);
```
arnej27959, line 27: i would not assert here as this is configured, not computed. same with other assert.

arnej27959:
consider just removing the asserts.

havardpe (approved):

PR #23218, opened 2022-06-23
Jonmv/multiple test and staging zones

jonmv:
@freva please review very carefully >_<


config-provisioning/src/main/java/com/yahoo/config/provision/zone/ZoneFilter.java
```diff
@@ -16,24 +15,13 @@ public interface ZoneFilter {
     /** Negates the next filter. */
     ZoneFilter not();
 
-    /** Zones which are upgraded by the controller. */
```
freva, line 19: Remember to update implementation in hosted

freva (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/zone/ZoneFilter.java
jonmv: Yes :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
```diff
@@ -538,11 +574,15 @@ public Map<JobId, List<Job>> testJobs(Map<JobId, List<Job>> jobs) {
...
                     }
-            });
-        }
+            }
+        });
         return Collections.unmodifiableMap(testJobs);
     }
 
+    private CloudName findCloud(JobType job) {
+        return zones.zones().all().among(job.zone()).zones().stream().findFirst().map(ZoneApi::getCloudName).orElse(null);
```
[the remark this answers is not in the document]
jonmv: Hmm, probably, yes ...

freva (approved):

jonmv:
What I really don't like here is that mutiple `JobType` with the same identity (job name) can have different zones. I think it would be appropriate to split the identity and the zone in two different classes. However, the `.zone()` is used everywhere, so this will be a bit of a change ... 
The alternative, still, is to use multiple job names, and keep the zone 1-1 with name. Urgh :)

PR #23226, opened 2022-06-24
client js base app/layout

ldalves:
@freva, PTAL.  There are a few TODOs, but let's keep moving...


freva (approved):

client/js/app/src/app/libs/router.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { Routes, Route, useParams, Navigate } from 'react-router-dom';
+import { Error } from 'app/components';
+
+const mainTitle = 'Vespa Console';
```
freva, line 5:
```suggestion
const mainTitle = 'Vespa App';
```

PR #23230, opened 2022-06-24
Use record for ComponentGraphResult [run-systemtest]

gjoranv:


bjorncs (approved):
FYI keep an eye on system tests. Spifly fails on weaving classes using records and other newer byte code features.

baldersheim (changes requested):
@bjorncs was correct, spifly is unhappy. Somethings need an upgrade ?

bjorncs:
@baldersheim I have filed a ticket to the Aries team (https://issues.apache.org/jira/browse/ARIES-2080).

PR #23232, opened 2022-06-24
Precompute a zone context to avoid always having to compute them on t…

baldersheim:
…he fly.

@bratseth PR


baldersheim:
Try to counter performance regression due introduced in #23080

bratseth (approved):

PR #23238, opened 2022-06-24
If the context map is empty just return the default

baldersheim:
@bratseth PR


baldersheim:
@bratseth I think we need to reconsider the changes in #23080. To me it seems it went from once per query to once per get that reaches a query profile. The effect on our simple performance tests were around 8% if you did around 8k qps.
That seems way too high. I wonder if we should change the context being passed around from a Map<String, String> to an object exposing only the methods needed. Then it is easy to add these kind of extra items without having to populate a map. Making a readonly Map<> which wraps the original context and the zoneContext, seems just complicated.

baldersheim:
@bratseth ping

bratseth:
Yes, agreed. I'm just a bit busy atm ...

PR #23240, opened 2022-06-26
Remove 'requireSequencing' as an option.

baldersheim:
As a consequence move protocol and params resolution to common code so that it is always handled in the decoding/encoding thread.

@vekterli PR

### DIFF
--- a/messagebus/src/vespa/messagebus/network/rpcsend.cpp
+++ b/messagebus/src/vespa/messagebus/network/rpcsend.cpp
@@ -55,10 +55,10 @@ private:
 
 }
 
-RPCSend::RPCSend() :
-    _net(nullptr),
-    _clientIdent("client"),
-    _serverIdent("server")
+RPCSend::RPCSend()
+    : _net(nullptr),
+      _clientIdent("client"),
+      _serverIdent("server")
 { }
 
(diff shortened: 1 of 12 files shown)


messagebus/src/vespa/messagebus/network/rpcsend.cpp
```diff
@@ -220,19 +220,19 @@ RPCSend::decode(vespalib::stringref protocolName, const vespalib::Version & vers
...
+            doHandleReply(std::move(reply));
         }));
         assert (!rejected);
     }
 }
 
 void
-RPCSend::doHandleReply(const IProtocol * protocol, Reply::UP reply) {
+RPCSend::doHandleReply(Reply::UP reply) {
+    const IProtocol * protocol = _net->getOwner().getProtocol(reply->getProtocol());
```
vekterli, line 235: Can this be `nullptr` or do we assume the protocol is always valid since we've already checked it as part of the initial request handling? Just wondering since the old code did a `if (!protocol` check (even though the check itself doesn't preclude the protocol pointer from potentially being dereferenced later in `doHandleReply`).

vekterli (approved):
:+1:

PR #23241, opened 2022-06-26
Upgrades from pre vespa-6.149 directly to 8.8 is not supported. GC co…

baldersheim:
…de from frontend.

@vekterli PR


vekterli (approved):
:+1:

PR #23248, opened 2022-06-27
Avoid building full Slime AST when decoding json just to throw it all…

baldersheim:
… away when making it into a string.

Instead try cheap stream parsing with jackson and keep strings for as long as possible.

@bratseth and @bjorncs PR


bratseth (approved):
👍

PR #23259, opened 2022-06-28
Remove uneeded version check

jonmv:
@bratseth please review. I think I added this condition just because I thought the controller would always ensure this, and it eased my mind, not because it was crucial. It's blocking upgrades from 7 to 8 for non-Java apps in the cloud.


bratseth (commented):
👍

PR #23268, opened 2022-06-29
Add cloud to SystemInfo

mpolden:
@bratseth

### DIFF
--- a/hosted-zone-api/src/main/java/ai/vespa/cloud/Cloud.java
+++ b/hosted-zone-api/src/main/java/ai/vespa/cloud/Cloud.java
@@ -1,7 +1,9 @@
 package ai.vespa.cloud;
 
 /**
- * The cloud provider of a Vespa Cloud instance.
+ * The cloud provider in which a cloud deployment may be running.
+ *
+ * This is "aws" when this runs in Amazon Web Services, and "gcp" when this runs in Google Cloud Platform.
  *
  * @author mpolden
  */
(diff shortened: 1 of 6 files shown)


hosted-zone-api/src/main/java/ai/vespa/cloud/Cloud.java
```diff
@@ -0,0 +1,9 @@
+package ai.vespa.cloud;
+
+/**
+ * The cloud provider of a Vespa Cloud instance.
```
bratseth, line 4:
```suggestion
 * The cloud provider in which a cloud deployment may be running.
 This is "aws" when this runs in Amazon Web Services, and "gcp" when this runs in Google Cloud Platform.
```

hosted-zone-api/src/main/java/ai/vespa/cloud/SystemInfo.java
```diff
@@ -13,12 +13,14 @@ public class SystemInfo {
 
     private final ApplicationId application;
     private final Zone zone;
+    private final Cloud cloud;
     private final Cluster cluster;
     private final Node node;
 
-    public SystemInfo(ApplicationId application, Zone zone, Cluster cluster, Node node) {
```
bratseth, line 19: Public API so we must keep the old constructor.

bratseth (changes requested):
Otherwise LGTM

mpolden:
PTAL

bratseth (approved):
👍

PR #23270, opened 2022-06-29
Detect when using .zone() on incomplete job type

jonmv:
@freva please review.


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/deployment-overview.json
```diff
@@ -191,7 +190,6 @@
       "jobName": "staging-test",
       "url": "http://localhost:8080/application/v4/tenant/tenant1/application/application1/instance/instance1/job/staging-test",
       "environment": "staging",
-      "region": "staging.us-east-3",
```
jonmv, line 194: Are these used?

freva: No

freva (approved):

PR #23274, opened 2022-06-29
Support mTLS connection-level capabilities and RPC access filtering in C++ [run-systemtest]

vekterli:
@havardpe @bjorncs please review

Adds the following:
 * Named capabilities and capability sets that represent (respectively)
   a single Vespa access API (such as Document API, search API etc)
   or a concrete subset of individual capabilities that make up a
   particular Vespa service (such as a content node).
 * A new `capabilities` array field to the mTLS authorization policies
   that allows for constraining what requests sent over a particular
   connection are allowed to actually do. Capabilities are referenced
   by name and may include any combination of capability sets and
   individual capabilities. If multiple capabilities/sets are configured,
   the resulting set of capabilities is the union set of all of them.
 * An FRT RPC-level access filter that can be set up as part of RPC
   method definitions. If set, filters are invoked prior to RPC methods.
 * A new `PERMISSION_DENIED` error code to FRT RPC that is invoked if
   an access filter denies a request.

This also GCs the unused `AssumedRoles` concept which is now deprecated
in favor of capabilities.

This PR does not wire access filtering to any existing RPC method definitions;
this will be added when the base functionality is in.

Note: this is **not yet** a public or stable API, and capability
names/semantics may change at any time!


bjorncs (approved):

fnet/src/vespa/fnet/frt/require_capability.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "request_access_filter.h"
+#include <vespa/vespalib/net/tls/capability_set.h>
+
+/**
+ * An RPC access filter which verifies that a request is associated with an auth
+ * context that contains, at minimum, a given set of capabilities. If one or more
+ * required capabilities are missing, the request is denied.
+ */
+class FRT_RequireCapability final : public FRT_RequestAccessFilter {
```
havardpe, line 12: consider plural in class name

vespalib/src/tests/net/tls/capabilities/capabilities_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+    EXPECT_EQUAL(Capability::content_metrics_api().name(),  "vespa.content.metrics_api"sv);
+    EXPECT_EQUAL(Capability::content_cluster_controller_internal_state_api().name(),
+                 "vespa.content.cluster_controller.internal_state_api"sv);
+}
+
+TEST("Capability instances can be stringified") {
+    EXPECT_EQUAL(Capability::content_storage_api().to_string(), "Capability(vespa.content.storage_api)");
+}
+
+TEST("All known capabilities can be looked up by name") {
```
havardpe, line 40: consider also matching the values against the actual capabilities

havardpe (approved):
some thoughts:
the CapabilityId enum might not need to be part of the API
bit_id and as_bit_set could be named value/as_set to 'hide' the implementation

fnet/src/vespa/fnet/frt/require_capability.h
vekterli: Fixed

vespalib/src/tests/net/tls/capabilities/capabilities_test.cpp
vekterli: Added checking of all returned values

vekterli:
@havardpe PTAL. Have refactored and moved `Capability` ID functionality around so that it is no longer exposed as part of the API. `bit_id`/`as_bit_set` is no longer available.

havardpe (approved):

PR #23283, opened 2022-06-30
Set provided scope for jupiter-api in dep-man

jonmv:
@bjorncs what do you think? Several users have had to exclude jupiter-api from, e.g., jupiter-params, to build their app ...

### DIFF
--- a/hosted-tenant-base/pom.xml
+++ b/hosted-tenant-base/pom.xml
@@ -94,6 +94,13 @@
 
             <dependency>
                 <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
+                <version>${vespa.junit.version}</version>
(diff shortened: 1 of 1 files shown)


hosted-tenant-base/pom.xml
```diff
@@ -92,6 +92,13 @@
                 <version>${vespaversion}</version>
             </dependency>
 
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
+                <version>${vespa.junit.version}</version>
+                <scope>provided</scope>
```
bjorncs, line 99: This should be test scope right?

jonmv: Oops, yes. But then I wonder why those applications were failing ...

PR #23289, opened 2022-06-30
Convert attribute and attribute combiner overrides using different fi…

toregge:
…eld as source to copy overrides for streaming search.

@geirst : please review


geirst (approved):
👍

baldersheim (approved):

PR #23292, opened 2022-06-30
Deprecate combined cluster

mpolden:
@jonmv

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
@@ -732,9 +732,14 @@ public class ContainerModelBuilder extends ConfigModelBuilder<ContainerModel> {
     private List<ApplicationContainer> createNodes(ApplicationContainerCluster cluster, Element containerElement, Element nodesElement, ConfigModelContext context) {
         if (nodesElement.hasAttribute("type")) // internal use for hosted system infrastructure nodes
             return createNodesFromNodeType(cluster, nodesElement, context);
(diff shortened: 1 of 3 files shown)


jonmv (approved):

config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -732,9 +732,14 @@ private static String getEnvironmentVariables(Element environmentVariables) {
     private List<ApplicationContainer> createNodes(ApplicationContainerCluster cluster, Element containerElement, Element nodesElement, ConfigModelContext context) {
         if (nodesElement.hasAttribute("type")) // internal use for hosted system infrastructure nodes
             return createNodesFromNodeType(cluster, nodesElement, context);
-        else if (nodesElement.hasAttribute("of")) // hosted node spec referencing a content cluster
-            return createNodesFromContentServiceReference(cluster, nodesElement, context);
-        else if (nodesElement.hasAttribute("count")) // regular, hosted node spec
+        else if (nodesElement.hasAttribute("of")) {// hosted node spec referencing a content cluster
+            // TODO: Remove support for combined clusters in Vespa 9
+            List<ApplicationContainer> containers = createNodesFromContentServiceReference(cluster, nodesElement, context);
+            log.log(WARNING, "Declaring combined cluster with <nodes of=\"...\"> is deprecated without " +
```
freva, line 738: I assume you want this displayed in console? In that case, this must be logged with `log.logApplicationPackage()`

mpolden: Yes, thanks!

PR #23302, opened 2022-06-30
Remove legacy MessageBus StorageAPI transport fallback [run-systemtest]

vekterli:
@geirst please review
@baldersheim FYI

Direct P2P RPC has been the preferred way for years, and we don't
need the fallback now that we're on Vespa 8.


geirst (approved):
👍

baldersheim (approved):
Red is a very nice colour in git.

PR #23310, opened 2022-07-01
QueryBuilder recreated in React

Ethnas:


ldalves:
It's a little harder to comment and requires a little more work when you have an entire app at once. Better if it's more interactions and small changes. :)

I see that some things here are not necessary as this is already provided by the base application.

Since this is not yet in use by anyone and is basically a copy of the current query builder app to learn a little React, let's move ahead and fix everything when merging with the base app.

ldalves (approved):

PR #23313, opened 2022-07-01
Bjorncs/bundles

bjorncs:
First commit contains changes previously reverted. Last commits fixes failing system tests.


baldersheim (approved):

config-model/src/main/java/com/yahoo/vespa/model/admin/clustercontroller/ClusterControllerContainer.java
```diff
@@ -49,7 +49,7 @@ public class ClusterControllerContainer extends Container implements
     private static final int DEFAULT_NETTY_NUM_DIRECT_ARENAS = 1; // Reduced from nettys default of 2*cores
     private static final int DEFAULT_NETTY_NUM_HEAP_ARENAS = 1; // Reduced from nettys default of 2*cores
 
-    private final Set<String> bundles = new TreeSet<>();
+    private final Set<String> bundles = new TreeSet<>(); // Ensure stable ordering
```
gjoranv, line 52: In which cases did LinkedHashSet not give stable ordering?

bjorncs: The comment is a bit imprecise. We want to ensure lexical ordering.

PR #23315, opened 2022-07-01
Empty docsummary output for undefined value.

toregge:
@geirst : please review
@baldersheim : FYI

### DIFF
--- a/searchsummary/src/vespa/searchsummary/docsummary/check_undefined_value_visitor.cpp
+++ b/searchsummary/src/vespa/searchsummary/docsummary/check_undefined_value_visitor.cpp
@@ -1,0 +1,140 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "check_undefined_value_visitor.h"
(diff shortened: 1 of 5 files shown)


geirst (approved):
👍

searchsummary/src/vespa/searchsummary/docsummary/check_undefined_value_visitor.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+void
+CheckUndefinedValueVisitor::visit(const document::Document&)
+{
+}
+
+void
+CheckUndefinedValueVisitor::visit(const document::DoubleFieldValue& value)
+{
+    if (isUndefined(value.getValue())) {
```
baldersheim, line 46: Why is this necessary ?
Single value fields are only undefined if they do not exist.
The special undefined value is only used in attributes, and should hence never make it into any document.

geirst: This has been resolved in https://github.com/vespa-engine/vespa/pull/23317.

PR #23317, opened 2022-07-01
Don't drop single value non-attribute byte/short/int/long field due to

toregge:
undefined value.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

baldersheim (approved):

PR #23323, opened 2022-07-02
Avoid map creation

bratseth:


container-search/src/main/java/com/yahoo/search/query/profile/QueryProfileProperties.java
```diff
@@ -300,13 +300,7 @@ private Map<String, String> contextWithZoneInfo(Map<String, String> context) {
...
         if (context == null || context.isEmpty()) return zoneContext;
         if (context == zoneContext) return context;
-        if (context.containsKey(ENVIRONMENT) && context.containsKey(REGION) && context.containsKey(INSTANCE)) return context;
-
-        Map<String, String> contextWithZoneInfo = new HashMap<>(context);
-        contextWithZoneInfo.putIfAbsent(ENVIRONMENT, zoneInfo.zone().environment().name());
-        contextWithZoneInfo.putIfAbsent(REGION, zoneInfo.zone().region());
-        contextWithZoneInfo.putIfAbsent(INSTANCE, zoneInfo.application().instance());
-        return Collections.unmodifiableMap(contextWithZoneInfo);
+        return new ChainedMap(zoneContext, context);
```
baldersheim, line 303: Since you were using putIfAbsent, I would expect context to be primary.

container-search/src/main/java/com/yahoo/search/query/profile/ChainedMap.java
```diff
@@ -0,0 +1,97 @@
...
+    private final Map<K, V> primary, secondary;
+
+    ChainedMap(Map<K, V> primary, Map<K, V> secondary) {
+        this.primary = primary;
+        this.secondary = secondary;
+    }
+
+    @Override
+    public int size() {
+        return keySet().size();
```
baldersheim, line 26: Let us hope this is never called....

baldersheim (changes requested):
Verify order of primary / secondary.

container-search/src/main/java/com/yahoo/search/query/profile/QueryProfileProperties.java
baldersheim: I reorder arguments and added some tests

container-search/src/main/java/com/yahoo/search/query/profile/ChainedMap.java
baldersheim: A bit faster

container-search/src/main/java/com/yahoo/search/query/profile/ChainedMap.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public Set<K> keySet() {
+        var keys = new HashSet<>(secondary.keySet());
+        keys.addAll(primary.keySet());
+        return keys;
+    }
+
+    @Override
+    public Collection<V> values() {
+        throw new UnsupportedOperationException();
```
baldersheim, line 89: I wonder if you should also implement the other readonly operations, like values, entrySet, equals, hashCode too ?

container-search/src/main/java/com/yahoo/search/query/profile/ChainedMap.java
bratseth: Yes, it's not.

container-search/src/main/java/com/yahoo/search/query/profile/QueryProfileProperties.java
bratseth: Yes, on reflection, overriding this in the query would be "not good".

container-search/src/main/java/com/yahoo/search/query/profile/ChainedMap.java
bratseth: They aren't used, so I don't think so.

PR #23326, opened 2022-07-03
Add noexcept specifier.

toregge:
@baldersheim : please review

### DIFF
--- a/metrics/src/vespa/metrics/metric.h
+++ b/metrics/src/vespa/metrics/metric.h
@@ -87,7 +87,7 @@ struct Tag
 
     Tag(vespalib::stringref k);
     Tag(vespalib::stringref k, vespalib::stringref v);
-    Tag(const Tag &);
+    Tag(const Tag &) noexcept;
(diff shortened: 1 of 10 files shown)


metrics/src/vespa/metrics/metric.h
```diff
@@ -87,7 +87,7 @@ struct Tag
 
     Tag(vespalib::stringref k);
     Tag(vespalib::stringref k, vespalib::stringref v);
-    Tag(const Tag &);
+    Tag(const Tag &) noexcept;
     Tag & operator = (const Tag &);
     Tag(Tag &&) = default;
```
baldersheim, line 92: noexcept on the move constructors too ?

baldersheim (approved):

metrics/src/vespa/metrics/metric.h
toregge: PR #23328 adds noexcept specifier for the move constructor.

PR #23351, opened 2022-07-04
Add option to ignore undefined fields

bratseth:


document/src/main/java/com/yahoo/document/json/readers/AddRemoveCreator.java
```diff
@@ -21,22 +21,22 @@ public class AddRemoveCreator {
 
     // yes, this suppresswarnings ugliness is by intention, the code relies on
     // the contracts in the builders
-    @SuppressWarnings({ "cast", "rawtypes", "unchecked" })
-    public static void createAdds(TokenBuffer buffer, Field field, FieldUpdate update) {
-        createAddsOrRemoves(buffer, field, update, false);
+    @SuppressWarnings("cast")
+    public static void createAdds(TokenBuffer buffer, Field field, FieldUpdate update, boolean ignoreUndefinedFields) {
+        createAddsOrRemoves(buffer, field, update, false, ignoreUndefinedFields);
     }
```
baldersheim, line 28: I suggest making the methods non-static and make the ignoreUndefinedFields a member.
Then you do not have to pollute all methods with ignoreUndefinedFields.
That goes for AddRemoveCreator, and all the Readers

baldersheim (approved):

document/src/main/java/com/yahoo/document/json/readers/AddRemoveCreator.java
bratseth: Yes, I agree that is better. However, there's an infinite amount of things to improve ... I'll put it on my list.

PR #23354, opened 2022-07-05
Support specifying accepted compression types in file distribution RPC requests [run-systemtest]

hmusum:
Add accepted compression types in file distribution requests.
Older clients (e.g. Vespa 7.x) will not have that in requests,
so fallback to gzip in that case.

### DIFF
--- a/filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReferenceCompressor.java
+++ b/filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReferenceCompressor.java
@@ -2,6 +2,8 @@
 package com.yahoo.vespa.filedistribution;
 
 import com.google.common.io.ByteStreams;
+import net.jpountz.lz4.LZ4BlockInputStream;
+import net.jpountz.lz4.LZ4BlockOutputStream;
 import org.apache.commons.compress.archivers.ArchiveEntry;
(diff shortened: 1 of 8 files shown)


filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReferenceCompressor.java
```diff
@@ -114,9 +118,17 @@ private static void writeFileToTar(ArchiveOutputStream taos, File baseDir, File
...
+        log.log(Level.FINE, () -> "Compressing with type " + type + " and compression type " + compressionType);
         switch (type) {
             case compressed:
-                return new GZIPOutputStream(new FileOutputStream(outputFile));
+                switch (compressionType) {
+                    case gzip:
+                        return new GZIPOutputStream(new FileOutputStream(outputFile));
+                    case lz4:
+                        return new LZ4BlockOutputStream(new FileOutputStream(outputFile));
+                    default:
```
baldersheim, line 129: What about zstandard ?

baldersheim (approved):

filedistribution/src/main/java/com/yahoo/vespa/filedistribution/FileReferenceCompressor.java
hmusum: output and input streams for zstandard not available yet, work in progress

PR #23360, opened 2022-07-05
Add a grace period before upgrading new nodes

bratseth:
A suggestion to wait a month after provisioning before starting to OS upgrade nodes.

I think this would be helpful for dev/perf as most apps are only worked on for less than this [citation needed], so they would avoid running into (and paying for) OS upgrades. 

Arguably also for dev, if new nodes are added to solve some problem you don't want to immediately introduce new changes, and if the capacity is temporary due to e.g autoscaling why do it at all.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/configserver/Node.java
@@ -41,6 +41,7 @@ public class Node {
     private final Version wantedVersion;
     private final Version currentOsVersion;
     private final Version wantedOsVersion;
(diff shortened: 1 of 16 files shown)


bratseth:
Spread budget + staggered start would be best?

mpolden:
Figured it was easier to just exclude nodes that are deferring their upgrade. PTAL

mpolden:
And upgrades by DelegatingOsUpgrader/RebuildingOsUpgrader are less impactful, so I removed the grace period for those.

bratseth (commented):
Fine ... making the grace depend on the type has higher entropy though.

mpolden:
Yes, but the strategies of each implementation are already different so this is just another unique property of RetiringOsUpgrader.

PR #23361, opened 2022-07-05
GC long gone config

baldersheim:
@bjorncs or @hmusum  PR

### DIFF
--- a/document/src/vespa/document/config/documentmanager.def
+++ b/document/src/vespa/document/config/documentmanager.def
@@ -5,9 +5,6 @@ namespace=document.config
 ## Whether attempts to set an undefined field should be ignored rather than causing an error
 ignoreundefinedfields bool default=false
 
-## Whether to enable compression in this process.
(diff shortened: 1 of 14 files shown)


document/src/vespa/document/config/documentmanager.def
```diff
@@ -5,9 +5,6 @@ namespace=document.config
 ## Whether attempts to set an undefined field should be ignored rather than causing an error
 ignoreundefinedfields bool default=false
 
-## Whether to enable compression in this process.
-enablecompression bool default=false
```
bjorncs, line 9: It is safe to remove a parameter used by older model currently in use (hosted/cloud)?

bjorncs (commented):
See comment

document/src/vespa/document/config/documentmanager.def
baldersheim: Yes, for our own config that shoul be fine. They belong to the model.

PR #23365, opened 2022-07-05
Avoid the host lock while failing the children

hakonhall:
If the host should be failed out,
1. any active children will be failed first (as before) and without holding the host lock.
2. The host will be failed in the next maintain().

This avoids a potential deadlock in (i) a dynamically provisioned zone, and (ii) when the replacement node requires the provisioning of a new host, i.e. there's not enough room on a shared host.  Because provisioning of a new host requires activating the host to include the new host in the host application.

NodeFailer throttling is changed to include wantToFail nodes when counting the number of recently failed nodes, and comparing those against (a) the percentage of of failed nodes threshold and (b) the number of failed hosts threshold.  This was necessary due to (2) above.

Fixes the following bug:  Active hosts with wantToFail, typical when an operator fails out a host (before this PR), was moved to fail without a redeployment.  This meant the Orchestrator and Service Monitor were out of sync with node repo until the next infra redeployment.


mpolden (approved):
Great!

PR #23367, opened 2022-07-05
Stop storing integer values in docsum blobs.

toregge:
@geirst : please review


geirst (approved):
👍

baldersheim:
@geirst or @toregge I suspect this one caused performance degradation between .29 and .33.
It seems like it is fetching summaries from disk ...

PR #23382, opened 2022-07-06
Restore original semantics of ResultClass::GetIndexFromName() and

toregge:
ResultClass::GetIndexFromEnumValue. Caller must check for entry being present.

Rename GeneralResult member functions that checks for entry being present.

@geirst : please review


geirst (approved):
👍

PR #23384, opened 2022-07-06
- Control number of rpc targets.

baldersheim:
- Control number of write events before waking up thread.
- For mbus in java, mbus in c++ and the rpc connection towards content nodes.

@vekterli PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/content/ContentNode.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/content/ContentNode.java
@@ -25,7 +25,7 @@ public abstract class ContentNode extends AbstractService
     private final boolean dispatch_on_decode;
     private final int mbus_threads;
     private final int mbus_network_threads;
-    private final int mbus_cpp_rpc_targets;
+    private final int mbus_rpc_targets;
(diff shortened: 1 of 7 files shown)


config-model/src/main/java/com/yahoo/vespa/model/content/ContentNode.java
```diff
@@ -25,6 +25,10 @@ public abstract class ContentNode extends AbstractService
     private final boolean dispatch_on_decode;
     private final int mbus_threads;
     private final int mbus_network_threads;
+    private final int mbus_cpp_rpc_targets;
+    private final int mbus_events_before_wakeup;
```
vekterli, line 29: I think it would make sense to rename `mbus_cpp_rpc_targets` to just `mbus_rpc_targets` to be in line with `mbus_events_before_wakeup`. C++ is already implied since it's the content node service

vekterli (approved):
:+1:

config-model/src/main/java/com/yahoo/vespa/model/content/ContentNode.java
baldersheim: Fixed

PR #23388, opened 2022-07-06
Render test output with less overhead, and limit pending network data

jonmv:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
@@ -40,7 +40,7 @@ class ProxyResponse extends HttpResponse {
 
     @Override
     public long maxPendingBytes() {
-        return 1 << 25; // 15MB
+        return 1 << 25; // 32MB
     }
 
 }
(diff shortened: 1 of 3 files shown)


jonmv:
ᕙ༼◕_◕༽ᕤ

configserver/src/main/java/com/yahoo/vespa/config/server/http/ProxyResponse.java
```diff
@@ -38,4 +38,9 @@ public void render(OutputStream outputStream) throws IOException {
         }
     }
 
+    @Override
+    public long maxPendingBytes() {
+        return 1 << 25; // 15MB
```
jonmv, line 43:
```suggestion
        return 1 << 25; // 32MB
```

vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/TestRunnerHandler.java
```diff
@@ -220,4 +258,36 @@ private static String traceToString(Throwable thrown) {
...
+        }
+
+        @Override
+        public String getContentType() {
+            return "application/json";
+        }
+
+        @Override
+        public long maxPendingBytes() {
+            return 1 << 25; // 15MB
```
jonmv, line 288:
```suggestion
            return 1 << 25; // 32MB
```

bjorncs (approved):

PR #23396, opened 2022-07-06
Add CMR tracking metric

olaaun:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VcmrMaintainer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VcmrMaintainer.java
@@ -362,12 +362,14 @@ public class VcmrMaintainer extends ControllerMaintainer {
     }
 
     private void updateMetrics() {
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VcmrMaintainer.java
```diff
@@ -357,4 +361,13 @@ protected ZonedDateTime getRetirementStartTime(ZonedDateTime plannedStartTime) {
         return time;
     }
 
+    private void updateMetrics() {
+        curator.readChangeRequests()
+                .stream()
+                .collect(Collectors.groupingBy(VespaChangeRequest::getStatus))
+                .forEach((status, cmrs) ->
+                        metric.set(TRACKED_CMRS_METRIC, cmrs.size(), metric.createContext(Map.of("status", status.name())))
```
andreer, line 369: Might be better to explicitly iterate over statuses, so we properly report 0 instead of no data

andreer (approved):

andreer (approved):

PR #23420, opened 2022-07-08
Bratseth/quota

bratseth:
Max resources are always set. If autoscaling is off, max==min.


bratseth:
Ping @oyving

bratseth:
Ping @oyving

bratseth:
Ping @oyving

bratseth:
Ping @oyving

PR #23427, opened 2022-07-08
Fail deployment of standalone services.xml in hosted if cluster controllers are specified [run-systemtest]

hmusum:
@oyving please review


hmusum:
ping @oyving

hakonhall:
It's a feature that developers can take a self-hosted application and deploy it to the dev environment.  @bratseth made this argument when it came to the node count and redundancy.  Perhaps we should allow cluster-controller element in dev too?

bratseth:
Yes, I think we should allow it in dev.

hmusum:
Right, I'll make a new PR where we allow this and make sure it works.

PR #23428, opened 2022-07-08
Stop storing integers of any size, floating point numbers and bool in docsum blobs.

toregge:
@geirst : please review


geirst (approved):
👍

PR #23432, opened 2022-07-08
Keep storing bool in docsum blob for now for streaming search.

toregge:
@geirst : please review

GroupingStreaming::global_max__STREAMING started failing with PR #23428 applied.


toregge:
Looks like the reason for new failures is that when bool is not stored in docsum blob then the field is not present in summary if not present in document. Adjusting the document feed in the test to always set the field would remove the need for this PR.

geirst (approved):
👍

toregge:
@baldersheim : FYI

PR #23434, opened 2022-07-08
Do not claim that you have asked nodes that were in a known failed st…

baldersheim:
…ate.

This makes coverage computation incorrect.

@bratseth


geirst:
Do we have some unit test we can update to test this?

bratseth:
Yes, taking a more optimistic view will mostly be more correct, especially since most have 2 searchable copies.
Although, with this why include nodes in coverage at all.

I agree we should have tests demonstrating these scenarios, but we can do it separately.

PR #23439, opened 2022-07-08
Keep storage node up while permanently down [run-systemtest]

hakonhall:
When the Orchestrator is asked about getting permission to remove a host (node), the Orchestrator asks the Cluster Controller to safely set the wanted state of the storage node (and distributor) to DOWN.

This PR will instead:
* Ask the CC of the same, but as a probe.  Which means it may fail for all the reasons a non-probe may fail, but if successful no wanted state will actually be set in the CC.
* Force-set the distributor to DOWN.

Ideally, both of these should be done atomically, but that will have to be done in a follow-up PR and wouldn't work for those on Vespa 7 anyways.

The net change of this PR affects hosted only: the storage node state is up (retired) in the period from when the Orchestrator has allowed the host to be removed from the application, to when it is actually removed through redeployment in RetiredExpirer.

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -58,6 +58,12 @@ public class Flags {
             "Takes effect at redeployment (requires restart)",
             ZONE_ID, APPLICATION_ID);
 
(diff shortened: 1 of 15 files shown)


vekterli (approved):
:+1:

PR #23447, opened 2022-07-11
Revert "Keep storing bool in docsum blob for now for streaming search."

toregge:
Reverts vespa-engine/vespa#23432

Depends on PR vespa-engine/system-test#2483

@geirst : please review


geirst (approved):
👍

PR #23450, opened 2022-07-11
Avoid the host lock while failing the children, 2. try

hakonhall:
Reverts vespa-engine/vespa#23449

Same as original vespa-engine/vespa#23365, except that
* wantToFail is only set on the parent host if it has no active children, and just before activate().
* The throttle on failing a minimum number of parent hosts is changed to count the number of parent hosts that are failed or have failed children.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailer.java
@@ -251,10 +251,8 @@ public class NodeFailer extends NodeRepositoryMaintainer {
                 }
             }
 
-            // A parent with children gets wantToFail to avoid getting more nodes allocated to it.
-            wantToFail(failing.node(), true, lock);
-
             if (activeChildrenToFail.isEmpty()) {
+                wantToFail(failing.node(), true, lock);
                 try {
                     deployment.get().activate();
                     return true;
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailer.java
```diff
@@ -251,10 +251,8 @@ private boolean failActive(FailingNode failing) {
                 }
             }
 
-            // A parent with children gets wantToFail to avoid getting more nodes allocated to it.
-            wantToFail(failing.node(), true, lock);
-
             if (activeChildrenToFail.isEmpty()) {
+                wantToFail(failing.node(), true, lock);
```
mpolden, line 255: Won't this allow allocation of new nodes on a host we're trying to fail? Maybe we'll just have to live with that.

mpolden (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeFailer.java
hakonhall: Yes, I think we will have to live with that.  We could invent some other marker to prevent allocation and ensure that stays in sync with the intention of the NodeFailer, but that's a bigger fix.

PR #23453, opened 2022-07-11
Add request access filter to internal StorageAPI RPC [run-systemtest]

vekterli:
@bjorncs @geirst please review
@havardpe FYI

This should always succeed today, as authz rules by default grant
all capabilities. But since this is a very hot call path, we'll
learn very quickly if the capability check incurs a measurable
overhead; it is not expected to do so in practice (really just a
virtual function call and a few bitwise ops).

Also added (buffered) logging to capability-checking request access filter.
In case of authz failure, log message will be in the following format:
```
Permission denied for RPC method 'my.cool.method'. Peer at tcp/127.0.0.1:12345 with
PeerCredentials(CN 'sneaky.example', DNS SANs ['localhost', 'sneaky.example']).
Call requires CapabilitySet({foo}), but peer has CapabilitySet({bar, baz})
```

### DIFF
--- a/fnet/src/vespa/fnet/frt/require_capabilities.cpp
+++ b/fnet/src/vespa/fnet/frt/require_capabilities.cpp
@@ -5,9 +5,25 @@
 #include <vespa/fnet/connection.h>
 #include <vespa/vespalib/net/connection_auth_context.h>
 
+#include <vespa/log/bufferedlogger.h>
+LOG_SETUP(".fnet.frt.require_capabilities");
+
+using namespace vespalib::net::tls;
+
 bool
 FRT_RequireCapabilities::allow(FRT_RPCRequest& req) const noexcept
 {
     const auto& auth_ctx = req.GetConnection()->auth_context();
-    return auth_ctx.capabilities().contains_all(_required_capabilities);
+    const bool is_authorized = auth_ctx.capabilities().contains_all(_required_capabilities);
+    if (!is_authorized) {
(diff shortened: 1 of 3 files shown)


fnet/src/vespa/fnet/frt/require_capabilities.cpp
```diff
@@ -5,9 +5,25 @@
...
 bool
 FRT_RequireCapabilities::allow(FRT_RPCRequest& req) const noexcept
 {
     const auto& auth_ctx = req.GetConnection()->auth_context();
-    return auth_ctx.capabilities().contains_all(_required_capabilities);
+    const bool is_authorized = auth_ctx.capabilities().contains_all(_required_capabilities);
+    if (!is_authorized) {
+        auto peer_spec = req.GetConnection()->GetPeerSpec();
+        std::string method_name(req.GetMethodName(), req.GetMethodNameLen());
+        LOGBT(warning, peer_spec, "Permission denied for RPC method '%s'. "
```
bjorncs, line 21: This may result in severe logging if a client uses non-persistent connections

bjorncs (approved):

fnet/src/vespa/fnet/frt/require_capabilities.cpp
vekterli: Good point. Will look at adding an API for getting a "reduced" spec once @havardpe is back. This will suffice for now.

geirst (approved):
👍

Issue #23223, opened 2022-06-23
Cannot start Vespa after Upgrade


sirganya:
After upgrading to the version 7.594.36 Vespa does not start

**To Reproduce**
yum update vespa

**Screenshots**

```
[2022-06-23 16:36:50] container-clustercontroller Cluster 'stack': Storing new cluster state version in ZooKeeper: 12
[2022-06-23 16:36:50] container-clustercontroller Cluster 'stack': Master moratorium complete: timed out waiting for all nodes to report in
[2022-06-23 16:36:50] container-clustercontroller Cluster 'stack': Publishing cluster state version 12
[2022-06-23 16:36:50] distributor      Distributor node ready. Done initializing. Giving out of sequence metric event. Config id is stack/distributor/0
[2022-06-23 16:36:50] distributor      Transitioning from baseline state 'Down' to 'Up' (cluster state version 12)
[2022-06-23 16:36:50] container-clustercontroller Cluster event type SYSTEMSTATE @1656002210111: Reset 1 start timestamps as all available distributors have seen newest cluster state.
[2022-06-23 16:38:36] configserver     Session 58 prepared successfully.
[2022-06-23 16:38:37] configserver     Session activated: 58
[2022-06-23 16:38:37] configserver     Session 58 activated successfully using no host provisioner. Config generation 58. Based on session 57. File references: [file '7578094d437997ed']
[2022-06-23 16:38:37] searchnode       Filedistributorrpc config is changed
[2022-06-23 16:38:37] container-clustercontroller Switching to the latest deployed set of configurations and components. Application config generation: 58
[2022-06-23 16:38:37] container-clustercontroller Starting deconstruction of 2 components and 0 bundles from generation 57
[2022-06-23 16:38:38] container-clustercontroller Completed deconstruction in 0.030 seconds
[2022-06-23 16:38:38] searchnode       Add document database: doctypename(file), configid(stack/search/cluster.stack/file)
[2022-06-23 16:38:38] container        Switching to the latest deployed set of configurations and components. Application config generation: 58
[2022-06-23 16:38:38] container        Starting deconstruction of 2 components and 0 bundles from generation 57
[2022-06-23 16:38:38] container        Completed deconstruction in 0.013 seconds
[2022-06-23 16:38:38] searchnode       Add document database: doctypename(filter), configid(stack/search/cluster.stack/filter)
[...]

PR #23467, opened 2022-07-12
Always use Vespa version in config request when serving config

hmusum:
This change means that nodes on a version that there exists no config
model for will not get new config, they will get an error instead.
This means that there will be no longer be possible to get incompatible
config for nodes on another version than one known by the config server.
https://github.com/vespa-engine/vespa/issues/23223

(Note: In hosted Vespa this has been set to true for a long time).

### DIFF
--- a/configdefinitions/src/vespa/configserver.def
+++ b/configdefinitions/src/vespa/configserver.def
@@ -47,7 +47,7 @@ zoneDnsSuffixes[] string
 # RPC protocol
 maxgetconfigclients int default=1000000
 maxoutputbuffersize int default=65536
-useVespaVersionInRequest bool default=false
+useVespaVersionInRequest bool default=true
 payloadCompressionType enum { UNCOMPRESSED, LZ4 } default=LZ4
 
 # Athenz config


bratseth (approved):

hmusum:
Yes, it does, good point!

geirst:
@bratseth @hmusum I guess we need to keep the detachment procedure for people still on Vespa 7. In order to upgrade to Vespa 8 they will upgrade via 7.594.36, as specified here: https://docs.vespa.ai/en/operations/live-upgrade.html

hmusum:
@geirst yes, doc change https://github.com/vespa-engine/documentation/pull/2151 has not been merged yet. There is no per major version doc, so we need to wait some time before merging that or specify how to do this both for vespa 7 and vespa 8

geirst:
@hmusum Perhaps we can add the detachment procedure to the Vespa 8 release notes?

hmusum:
Yes, done in https://github.com/vespa-engine/documentation/pull/2151

PR #23473, opened 2022-07-12
Bratseth/total cost

bratseth:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepoStatsData.java
```diff
@@ -16,6 +16,12 @@
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class NodeRepoStatsData {
 
+    @JsonProperty("totalCost")
+    public Double totalCost;
+
+    @JsonProperty("totalAllocatedCost")
+    public Double totalAllocatedCost;
```
mpolden, line 23: I think these will be `null` until the zone includes the new fields, which can throw on the unboxing below. Consider initializing to 0.

bratseth: I don't think it's worth it since it just potentially breaks a low-value operator view during this one upgrade?

mpolden (approved):

PR #23475, opened 2022-07-12
Disallow absolute root in MockApplicationPackage

mpolden:
Before this change, an absolute path was treated as a relative one. So test data
ended up in this (surprising) directory:

```
~/y/vespa master$ ll ./config-model/Users/mpolden/y/vespa/config-model/target/
total 0
drwxr-xr-x 3 mpolden staff 96 Jul 12 15:24 junit4128625600353247086
drwxr-xr-x 3 mpolden staff 96 Jul 12 15:24 junit8496400184577411104
```

@hmusum


hmusum (approved):
👍

PR #23493, opened 2022-07-13
Reduce zk snapshot count for cluster controller to 1000 [run-systemtest]

hmusum:


hakonhall (approved):
What's the default?

hmusum:
The default is 50 000

vekterli (approved):
:+1:

PR #23505, opened 2022-07-14
Fail deployment if node count is not given for content cluster in hosted

hmusum:
Node count must be given for other zones than those that are manually deployed.


bratseth (approved):
Great!

PR #23514, opened 2022-07-14
Find current load more reliably

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterModel.java
```diff
@@ -110,8 +108,8 @@ public double queryFractionOfMax() {
         return queryFractionOfMax = clusterTimeseries().queryFractionOfMax(scalingDuration(), clock);
     }
 
-    /** Returns average load during the last {@link #currentLoadDuration} */
-    public Load currentLoad() { return nodeTimeseries().averageLoad(clock.instant().minus(currentLoadDuration)); }
```
mpolden, line 114: Is the `averageLoad` method still used or can it be removed?

mpolden (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterModel.java
bratseth: Still used.

PR #23517, opened 2022-07-15
Build vespa-feed-client as multi-release JAR and use reflection as workaround for JDK8

bjorncs:
FYI @gjoranv @jonmv


hakonhall (approved):

gjoranv:
👍

jonmv:
Nice :)

PR #23519, opened 2022-07-15
Make ExecutionFactory a child of SearchHandler for correct chains config [run-systemtest]

gjoranv:
No magic spell needed. You just have to add the component in the right place in the model topography.

The central parts are in ContainerModelBuilder and SearchBuilderTest. The rest is mostly unrelated unification of platform bundles setup.


bratseth (approved):
👍 👍

bjorncs (commented):
👍

PR #23528, opened 2022-07-20
Bjorncs/capabilities [run-systemtest]

bjorncs:


jrt/tests/com/yahoo/jrt/InvokeErrorTest.java
```diff
@@ -157,4 +160,36 @@ public void testConnectionError() {
...
+    public void testFilterFailsRequest() {
+        Request r = new Request("test");
+        r.parameters().add(new Int32Value(42));
+        r.parameters().add(new Int32Value(0));
+        r.parameters().add(new Int8Value((byte)0));
+        filter.allowed = false;
+        assertFalse(filter.invoked);
+        target.invokeSync(r, timeout);
+        assertTrue(r.isError());
+        assertTrue(filter.invoked);
```
vekterli, line 185: Should we add an explicit check that the `test` method itself is _not_ invoked when the filter fails? (and vice versa in the case where the RPC passes the authz filter)

security-utils/src/main/java/com/yahoo/security/tls/PeerAuthorizer.java
```diff
@@ -35,9 +35,7 @@ public PeerAuthorizer(AuthorizedPeers authorizedPeers) {
     public ConnectionAuthContext authorizePeer(X509Certificate cert) { return authorizePeer(List.of(cert)); }
 
     public ConnectionAuthContext authorizePeer(List<X509Certificate> certChain) {
-        if (authorizedPeers.isEmpty()) {
-            return new ConnectionAuthContext(certChain, CapabilitySet.all(), Set.of());
-        }
+        if (authorizedPeers.isEmpty()) return ConnectionAuthContext.defaultAllCapabilities();
```
vekterli, line 38: Consider if we also want to expose the peer certificate information in `ConnectionAuthContext` even if it has no authz rules and is implicitly granted all capabilities

vekterli (approved):
:+1:

vekterli (approved):
:+1:

PR #23530, opened 2022-07-21
Remove ancient vespamalloc output parsing script

vekterli:
@aressem or @bjorncs please review

Most of the symbols that the script knows about are long, long gone.


bjorncs (approved):
🎉

PR #23534, opened 2022-07-25
Fix cost field name

freva:
Controller node-repository client expects camel case https://github.com/vespa-engine/vespa/blob/58a7c193360e6aca3da295592f8b5ab5fcf6424e/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepoStatsData.java#L19
The other fields in this API also use camel case


bratseth (approved):
Thanks!

PR #23535, opened 2022-07-25
add "vespa logfmt" for pre-review

arnej27959:
please review and suggest improvements
note: this is my first attempt at 'go' programming.


bratseth (commented):
That was fast! 

LGTM but @mpolden is the Go authority :-)

PR #23545, opened 2022-07-27
go logfmt in vespalog

arnej27959:
this is supposed to be a drop-in replacement or the perl script vespa-logfmt with same options and basically same behavior.  Probably needs some unit tests of the underlying implementation, so not ready for merge yet, but some review can happen now (while I'm on vacation).

Some unresolved questions:
* I hooked building the go code into cmake, is there any problem with that?
* if not, should we do that for the client/go directory too?
* I'm considering moving the actual logfmt code from client to vespalog, but some parts could be useful
* my first try with "vespa logfmt" should be removed, it doesn't fit with the rest of the "vespa xxx" commands and isn't compatible with perl script.


mpolden:
> I hooked building the go code into cmake, is there any problem with that?
Should be fine.

> if not, should we do that for the client/go directory too?
Sure. I believe it's already included in the vespa-clients RPM though.

> I'm considering moving the actual logfmt code from client to vespalog, but some parts could be useful
> my first try with "vespa logfmt" should be removed, it doesn't fit with the rest of the "vespa xxx" commands and isn't compatible with perl script.

So it's possible to build multiple binaries within the same Go module, just create a separate package containing the function, e.g. ```$ $ cat cmd/vespa-logfmt/main.go
package main

import "fmt"

func main() {
	fmt.Println("hello from vespa-logfmt")
}```

client/go/cmd/logfmt/runlogfmt.go
```diff
@@ -0,0 +1,79 @@
...
+	if ev == "" {
+		return "/opt/vespa"
+	}
+	return ev
+}
+
+func RunLogfmt(opts *Options, args []string) {
+	if len(args) == 0 {
+		if inputIsTty() {
+			args = append(args, vespaHome()+"/logs/vespa/vespa.log")
```
mpolden, line 30: Consider using `filepath.Join` to build path.

client/go/cmd/logfmt/runlogfmt.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		tailFile(opts, args[0])
+		return
+	}
+	for _, arg := range args {
+		file, err := os.Open(arg)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Cannot open '%s': %v\n", arg, err)
+		} else {
+			formatFile(opts, file)
```
mpolden, line 48: `file` should be explicitly closed.

vespalog/go/vespa-logfmt/cmd.go
```diff
@@ -0,0 +1,47 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// vespa-logfmt command
+// Author: arnej
+
+package main
+
+// var VespaVersion string = "0.0.0-devel"
```
mpolden, line 7: Remove?

arnej27959:
will make new PRs with squished commits, thanks for many useful comments

PR #23547, opened 2022-07-28
Trace converter added

Ethnas:
Added ability to convert traces to the form expected by Jaeger. Added new page for pasting Vespa trace and downloading the converted trace.


kkraune:
@freva I assume we can just merge this for now for testing?

freva (approved):

PR #23550, opened 2022-07-28
Don't add Providers to injected ComponentRegistry if any concrete ins… [run-systemtest]

gjoranv:
…tances exist.

This will break for users that depend on Providers being added to the component registry, but sounds very unlikely to affect anyone in practice.


gjoranv:
My assessment was wrong. We are relying upon this functionality ourselves, where `Container.requestHandlerRegistry` contains both "concrete" handlers and  those created via a Provider. Definitely not unthinkable that a user app that takes all instances of a type assumes the same thing.

PR #23564, opened 2022-08-02
Make some ConvergenceException increment the unhandled_exceptions metric

hakonhall:


hakonhall:
An alternative is to leave the unhandled_exceptions metric as it is now, and create a new metric for the sum.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/orchestrator/OrchestratorImpl.java
```diff
@@ -59,7 +59,7 @@ public void suspend(final String hostName) {
         } catch (HttpException e) {
             throw new OrchestratorException("Failed to suspend " + hostName + ": " + e.toString());
         } catch (ConnectionException e) {
-            throw new ConvergenceException("Failed to suspend " + hostName + ": " + e.getMessage());
+            throw ConvergenceException.ofTransient("Failed to suspend " + hostName + ": " + e.getMessage());
```
freva, line 62: `ConnectionException` is defined as error, feels like they should be the same (whatever it is)?

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/orchestrator/OrchestratorImpl.java
hakonhall: Who are "they"?

Suspension failures is expected to happen and expected to be transient.  During upgrades there may well be connection failures etc without that necessarily indicating anything wrong.  The strategy is to avoid incrementing unhandled_exceptions in the common cases of Vespa upgrades etc.  So that's why a ConnectionException is converted to a transient convergence exception here.

But connection exceptions in other circumstances should generally increment unhandled_exceptions.

PR #23581, opened 2022-08-05
Added defaultIndex processing for userInput items parsing

akolhun:


bratseth (approved):
Great!

Since we already access parse-specific state in the superclass I think we should move this there instead of passing it as an argument, but I can do that in a followup PR.

baldersheim:
I suspect this one broke system test https://factory.vespa.oath.cloud/testrun/48700357/test/Alias::test_alias__ELASTIC
Reverting to see

PR #23590, opened 2022-08-08
Cleanup xml schema validation [run-systemtest]

hmusum:
Cleanup and throw an exception if no XML schema was found.


config-application-package/src/main/java/com/yahoo/config/model/application/provider/SchemaValidators.java
```diff
@@ -63,8 +63,6 @@ public SchemaValidators(Version vespaVersion) {
             routingStandaloneXmlValidator = createValidator(schemaDir, routingStandaloneXmlSchemaName);
         } catch (IOException ioe) {
             throw new RuntimeException(ioe);
-        } catch (Exception e) {
-            throw e;
```
jonmv, line 67: 😆

jonmv (approved):

PR #23592, opened 2022-08-08
Support partial infrastructure upgrades

mpolden:
This allows infrastructure upgrades (Vespa or OS) to proceed with the next
step/zone after some subset of hosts have completed upgrading. The minimum
number of hosts is set by the upgrade policy.

@jonmv

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/zone/NodeSlice.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/zone/NodeSlice.java
@@ -1,0 +1,46 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.config.provision.zone;
+
+import java.util.Objects;
+import java.util.OptionalDouble;
+import java.util.OptionalLong;
+
+/**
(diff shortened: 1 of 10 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/zone/NodeSlice.java
```diff
@@ -0,0 +1,46 @@
...
+        }
+        if (fraction.isPresent() && fraction.getAsDouble() > 1.0D) {
+            throw new IllegalArgumentException("Fraction must be <= 1.0, got " + fraction.getAsDouble());
+        }
+    }
+
+    /** Returns whether this slice is satisfied by given node count, out of totalCount */
+    public boolean satisfiedBy(long count, long totalCount) {
+        if (fraction.isPresent()) {
+            return count >= totalCount * fraction.getAsDouble();
```
jonmv, line 31: I think we should add a little epsilon here, on the left side.

jonmv:
```suggestion
            return count + 1e-9 >= totalCount * fraction.getAsDouble();
```

jonmv (approved):
Nice!

PR #23602, opened 2022-08-09
Arnej/logfmt go code

arnej27959:
just the go code, without hooking into cmake for now. functionality should be complete and backwards compatible.


mpolden (approved):
👍

PR #23610, opened 2022-08-09
qrserver -> container, part 2 [run-systemtest]

hmusum:


arnej27959 (approved):
Looks good, but «search metrics» could be backend, consider «search chain metrics» instead.

PR #23616, opened 2022-08-10
Expose details of next change in /os/v1

mpolden:
..which can then be shown in Console.

@jonmv


jonmv (approved):

jonmv:
Nice. Unrelated test failure.

Issue #19840, opened 2021-11-03
Use default content-type in vespa-fbench for post queries


frodelu:
When using `vespa-fbench`, you need to specifically add the content-type header (`-H "Content-type:application/json"`) when using HTTP POST queries. 

When you specify post queries on the command line with the `-P` option, we should add the relevant content-type header as default unless another one is explicitly set with the `-H` option.

frodelu:
Alternative to defaulting to json content-type is to at least require a content-type explicitly specified (using `-H` option) if you use post requests (with the `-P` option).

baldersheim:
fixed in vespa-8.34.21

PR #23634, opened 2022-08-11
Default content-type is json when using post.

baldersheim:
@frodelu or @havardpe  PR


frodelu (approved):

baldersheim:
Fixes #19840

jonmv:
This should be accompanied by an update to the documentation, which still says you need to specify `application/json`.

havardpe (commented):
consider using a single space in the default content type header:
"The field value MAY be preceded by any amount of LWS, though a single SP is preferred"

PR #23643, opened 2022-08-12
Store timestamp on tenant when to expire user session

freva:
Must me merged with internal PR.

The application roles are no longer used, so the first commit removes the add/remove APIs for those.
Only the console uses the update role API, and it has long ago migrated to send an array of `roles` the user should have (second commit).


tokle (approved):
💯

PR #23647, opened 2022-08-12
Add method to use Duration as timeout to make unit explicit, and avoi…

baldersheim:
…d many magic conversions.

@hmusum or @bjorncs PR


hmusum (approved):
Nice! 🏅

bjorncs (commented):
🏅

PR #23648, opened 2022-08-12
Revert "Revert "Use a hash_set<int32_t> to quickly check if a field i…

baldersheim:
…s an index field.""

Handle structs too.

@toregge or @geirst PR


toregge (approved):

baldersheim:
Reverting as there is an issue related to the URI type

PR #23671, opened 2022-08-15
Revert "Revert "Revert "Revert "Use a hash_set<int32_t> to quickly check if a field i…""

baldersheim:
Reverts vespa-engine/vespa#23670


baldersheim:
Superseded by #23679

PR #23690, opened 2022-08-17
Compare apples to apples MERGEOK

jonmv:
@tokle or @freva please review and merge.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/LockedTenant.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/LockedTenant.java
@@ -12,6 +12,7 @@ import com.yahoo.vespa.hosted.controller.api.identifiers.Property;
 import com.yahoo.vespa.hosted.controller.api.identifiers.PropertyId;
 import com.yahoo.vespa.hosted.controller.api.integration.organization.Contact;
 import com.yahoo.vespa.hosted.controller.api.integration.secrets.TenantSecretStore;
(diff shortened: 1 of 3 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/LockedTenant.java
```diff
@@ -154,8 +155,11 @@ public CloudTenant get() {
 
         public Cloud withDeveloperKey(PublicKey key, Principal principal) {
             BiMap<PublicKey, Principal> keys = HashBiMap.create(developerKeys);
+            principal = new SimplePrincipal(principal.getName());
```
freva, line 158: Ideally we should change the type of the `BiMap` to also be `SimplePrincipal`, if someone changes deserialization code, this could stop working again?

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/LockedTenant.java
```diff
@@ -154,8 +155,11 @@ public CloudTenant get() {
 
         public Cloud withDeveloperKey(PublicKey key, Principal principal) {
             BiMap<PublicKey, Principal> keys = HashBiMap.create(developerKeys);
```
mpolden, line 157: Consider using `SimplePrincipal` in the map as well so that the equality contract is obvious.

jonmv:
The updated unit test fails if we no longer reject these updates, but I can make a follow-up PR with the suggested change.

PR #23694, opened 2022-08-17
Use SimplePrincipal internally in CloudTenant

jonmv:
@freva or @mpolden


freva (approved):

mpolden (approved):
Slightly related: `SimplePrincipal` looks like a good use-case for StringWrapper/PatternedStringWrapper.

jonmv:
Yes, it does, but I'm not sure what pattern to expect. Emails can be so many different things. Any suggestions?

PR #23698, opened 2022-08-17
Discard corrupt log data MERGEOK

jonmv:
@freva please review and merege


freva (approved):

jonmv:
Was able to doctor the data.

PR #23701, opened 2022-08-17
Add support for two-phase document garbage collection [run-systemtest]

vekterli:
@geirst please review

If enabled, garbage collection is performed in two phases (metadata
gathering and deletion) instead of just a single phase. Two-phase GC
allows for ensuring the same set of documents is deleted across all
nodes and explicitly takes write locks on the distributor to prevent
concurrent feed ops to GC'd documents from potentially creating
inconsistencies.

Two-phase GC is only used _iff_ all replica content nodes support
the feature _and_ it's enabled in config. An additional field has
been added to the feature negotiation functionality to communicate
support from content nodes to distributors.


storage/src/tests/storageapi/mbusprot/mbusprot.4.2.serialization.V_4_2_STABLE
```diff
@@ -1,66 +0,0 @@
-
```
geirst, line 1: @vekterli Was removal of this file intentional?

vekterli: It's intentional, although it could've/should've been its own micro-PR; I just did it here since I stumbled across it as part of adding 2-phase RemoveLocation to the storage API protocol. It's an ancient leftover file from the days where we used mbus for the StorageAPI and had a daisy-chain of protocol versions supported. The "4.2" version part indicates it was practically museum-ready anyway 🦖

geirst (approved):
👍

PR #23714, opened 2022-08-19
Handle response from prepare properly

hmusum:
Works as expected when testing this manually against a config server, but unit test fails with:

    deploy_test.go:17:
        	Error Trace:	deploy_test.go:127
        	            				deploy_test.go:17
        	Error:      	Expected nil, but got: &errors.errorString{s:"EOF"}
        	Test:       	TestPrepareZip
    deploy_test.go:17:
        	Error Trace:	deploy_test.go:128
        	            				deploy_test.go:17
        	Error:      	Not equal:
        	            	expected: "Success: Prepared testdata/applications/withTarget/target/application.zip with session 42\n"
        	            	actual  : ""

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,2 +1 @@
        	            	-Success: Prepared testdata/applications/withTarget/target/application.zip with session 42

        	Test:       	TestPrepareZip

@mpolden or @ean any idea what happens here? Looks like it fails with EOF when getting response from mock client? Not too familiar with Go and how the mock is connected to the tests, any help appreciated

### DIFF
--- a/client/go/cmd/deploy.go
+++ b/client/go/cmd/deploy.go
@@ -132,6 +132,7 @@ func newPrepareCmd(cli *CLI) *cobra.Command {
 				return fmt.Errorf("could not write session id: %w", err)
 			}
 			cli.printSuccess("Prepared ", color.CyanString(pkg.Path), " with session ", result.ID)
+			printPrepareLog(cli.Stderr, result)
 			return nil
 		},
 	}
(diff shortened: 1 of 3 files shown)


ean:
@hmusum take a look at https://github.com/vespa-engine/vespa/pull/23714/commits/d71b338105692e7b0f089a04827a942fb5fe0ba9
it fixes the test, but I am not sure if there should be any more data in the response body.

(The http mock client will return OK with empty body if all the specified responses have been used. Which is why you are getting and EOF error when decoding the json data)

hmusum:
Thanks! This looks good, I plan to revisit these tests and add more to tests, including more data in response

ean (approved):

mpolden:
👍

PR #23715, opened 2022-08-19
Make vespa-logfmt build on non-unix

mpolden:
Vespa CLI releases are built for non-unix platforms, so we need to explicitly
preserve compatibility when using unix-specific APIs.

@arnej or @bratseth


mpolden:
@arnej27959 *

arnej27959 (approved):
Is this for windows or macOS?

mpolden:
The workaround is for Windows, yes. macOS counts as UNIX.

bratseth (approved):

PR #23724, opened 2022-08-19
Avoid copying a vector of shared_ptrs when lifetime is guaranteed by …

baldersheim:
…the lock held.

@toregge @geirst @vekterli PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/persistenceengine/persistenceengine.cpp
+++ b/searchcore/src/vespa/searchcore/proton/persistenceengine/persistenceengine.cpp
@@ -199,9 +199,14 @@ PersistenceEngine::getHandlerSnapshot(const WriteGuard &) const
 }
 
 PersistenceEngine::HandlerSnapshot
-PersistenceEngine::getHandlerSnapshot(const ReadGuard &, document::BucketSpace bucketSpace) const
+PersistenceEngine::getSafeHandlerSnapshot(const ReadGuard &, document::BucketSpace bucketSpace) const
 {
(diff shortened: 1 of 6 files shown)


searchcore/src/vespa/searchcore/proton/persistenceengine/persistenceengine.cpp
```diff
@@ -513,17 +518,17 @@ PersistenceEngine::createIterator(const Bucket &bucket, FieldSetSP fields, const
                                   IncludedVersions versions, Context &context)
 {
     ReadGuard rguard(_rwMutex);
-    HandlerSnapshot snapshot = getHandlerSnapshot(rguard, bucket.getBucketSpace());
+    auto snap = getSafeHandlerSnapshot(rguard, bucket.getBucketSpace());
```
geirst, line 521: @baldersheim What is the difference between _createIterator()_ and the other functions, requiring us to use _getSafeHandlerSnapshot()_ ?

baldersheim: The snapshot acquired in createIterator is stored in the iterator which is returned from the createIterator method.
It is used for all subsequent iterate() calls.

All other usages are within the method and lifetime is less than the Read/Write guard held when acquiring the snapshot.

Keep in mind that the handle snapshot mechanism was made before we introduced the RW locks to make configuration safe. Most of the frequently called methods like put/update etc. have been optimised earlier on.

This takes car eof the other frequently called ones where getBucketInfo stands out the most.

geirst (approved):
👍

PR #23729, opened 2022-08-22
allow lowercase 'n' for newline dequoting

arnej27959:
@hakonhall please review
@hmusum FYI


arnej27959:
the Getopt in perl is too clever, and so is the cobra framework in go, but they are clever in different incompatible ways...

hakonhall (approved):

PR #23730, opened 2022-08-22
use conditional on windows instead

arnej27959:
our "go" is probably too old, we always got "tail is not supported on this platform"
@mpolden please review
@aressem FYI


mpolden (approved):
👍 

My bad. Looks like "unix" was just added in the most recent release.

PR #23737, opened 2022-08-22
Support capability enforcement environment variable in C++ [run-systemtest]

vekterli:
@havardpe @bjorncs please review

Mirrors Java enforce/log-only/disable semantics, defaulting to enforce.

Also fixes an issue where connection auth context and capabilities
would not be set if a server socket was running in mixed-mode. This
is not a problem in practice since mixed-mode is inherently completely
insecure since it must accept plain-text clients, which implicitly
have all capabilities granted.

### DIFF
--- a/fnet/src/vespa/fnet/frt/require_capabilities.cpp
+++ b/fnet/src/vespa/fnet/frt/require_capabilities.cpp
@@ -4,6 +4,7 @@
 #include "rpcrequest.h"
 #include <vespa/fnet/connection.h>
 #include <vespa/vespalib/net/connection_auth_context.h>
+#include <vespa/vespalib/net/tls/capability_env_config.h>
 
 #include <vespa/log/bufferedlogger.h>
 LOG_SETUP(".fnet.frt.require_capabilities");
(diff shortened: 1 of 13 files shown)


bjorncs (approved):

fnet/src/vespa/fnet/frt/require_capabilities.cpp
```diff
@@ -15,15 +16,22 @@ FRT_RequireCapabilities::allow(FRT_RPCRequest& req) const noexcept
...
         std::string method_name(req.GetMethodName(), req.GetMethodNameLen());
-        LOGBT(warning, peer_spec, "Permission denied for RPC method '%s'. "
+        LOGBT(warning, peer_spec, "%sPermission denied for RPC method '%s'. "
                                   "Peer at %s with %s. Call requires %s, but peer has %s",
+              ((mode == CapabilityEnforcementMode::LogOnly) ? "(Dry-run only, not enforced): " : ""),
               method_name.c_str(), peer_spec.c_str(),
               to_string(auth_ctx.peer_credentials()).c_str(),
               _required_capabilities.to_string().c_str(),
               auth_ctx.capabilities().to_string().c_str());
+        return (mode == CapabilityEnforcementMode::Enforce) ? false : true;
```
havardpe, line 35: prefer logic expression over conditional mapping of bool to bool

vespalib/src/vespa/vespalib/net/tls/capability_env_config.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+}
+
+const char* to_string(CapabilityEnforcementMode mode) noexcept {
+    switch (mode) {
+    case CapabilityEnforcementMode::Enforce: return "Enforce";
+    case CapabilityEnforcementMode::LogOnly: return "LogOnly";
+    case CapabilityEnforcementMode::Disable: return "Disable";
+    default: abort();
```
havardpe, line 37: prefer abort outside switch to enable compile-time check for enum completeness

havardpe (approved):

fnet/src/vespa/fnet/frt/require_capabilities.cpp
vekterli: Fixed

vespalib/src/vespa/vespalib/net/tls/capability_env_config.cpp
vekterli: Fixed

PR #23738, opened 2022-08-22
- Refactor to allow for different decay method.

baldersheim:
- Implement decay over constant time in addition to over requests.
- Default is decay over 500 requests, and 5.0s if you choose to decay over time.

@bratseth and @havardpe PR

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
@@ -246,7 +246,7 @@ public class LoadBalancer {
             public void decay(RequestDuration duration) {
                 double searchTime = Math.max(duration.duration().toMillis()/1000.0, MIN_QUERY_TIME);
                 double decayRate = LATENCY_DECAY_TIME;
-                double sampleWeight = toDouble(duration.timeSince(prev));
+                double sampleWeight = Math.min(decayRate/2, toDouble(duration.timeSince(prev)));
                 averageSearchTime = (sampleWeight*searchTime + (decayRate - sampleWeight) * averageSearchTime) / decayRate;
                 prev = duration;
             }
--- a/container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
@@ -96,12 +96,12 @@ public class LoadBalancer {
 
         interface Decayer {
             void decay(RequestDuration duration);
-            double averageSearchTime();
+            double averageCost();
         }
 
         static class NoDecay implements Decayer {
             public void decay(RequestDuration duration) {}
-            public double averageSearchTime() { return MIN_QUERY_TIME; }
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
```diff
@@ -201,13 +207,58 @@ private int nextScoreboardIndex(int current) {
...
 
     static class AdaptiveScheduler implements GroupScheduler {
-
+        enum Type {TIME, REQUESTS}
         private final Random random;
         private final List<GroupStatus> scoreboard;
 
-        public AdaptiveScheduler(Random random, List<GroupStatus> scoreboard) {
+        private static double toDouble(Duration duration) {
+            return duration.toNanos()/1_000_000_000.0;
```
bratseth, line 215:
```suggestion
            return duration.toSeconds();
```

container-search/src/main/java/com/yahoo/search/dispatch/RequestDuration.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+    private RequestDuration complete(long duration) {
+        endTime = startTime + duration;
+        return this;
+    }
+    Duration duration() {
+        return Duration.ofNanos(endTime - startTime);
+    }
+    Duration timeSince(RequestDuration prev) {
+        return Duration.ofNanos(Math.abs(endTime - prev.endTime));
```
bratseth, line 33: I think you should move the abs to the caller, or else rename this to "difference".

container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
baldersheim: No can do, This only gives number of whole seconds. Needs exact time as a floating point number in seconds.

container-search/src/main/java/com/yahoo/search/dispatch/LoadBalancer.java
```diff
@@ -93,50 +94,55 @@ public void releaseGroup(Group group, boolean success, Duration searchTime) {
...
-                double decayRate = Math.min(queries + MIN_LATENCY_DECAY_RATE, DEFAULT_LATENCY_DECAY_RATE);
-                averageSearchTime = (searchSeconds + (decayRate - 1) * averageSearchTime) / decayRate;
-                queries++;
+                decayer.decay(searchTime);
             }
         }
 
         Duration averageSearchTime() {
-            return Duration.ofNanos((long)(averageSearchTime*1000000000));
+            return Duration.ofNanos((long)(decayer.averageSearchTime()*1_000_000_000));
```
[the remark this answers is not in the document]
baldersheim: Needs the full resolution.

container-search/src/main/java/com/yahoo/search/dispatch/RequestDuration.java
baldersheim: Agree

baldersheim:
I have refactored to make it more clear when we are dealing with time and when we are dealing with a decaying cost.

PR #23743, opened 2022-08-23
Fix release day and add system-dependent cooldown

mpolden:
This will ensure that we schedule a precise version instead of a lower bound,
and that upgrades to the next release always happens in CD systems earlier than
public/main.

@jonmv


mpolden:
or @andreer (in case @jonmv is OOO)

andreer (approved):

PR #23750, opened 2022-08-23
Add capability filter to cluster controller API RPCs on content nodes

vekterli:
@geirst please review
@bjorncs FYI

### DIFF
--- a/storage/src/vespa/storage/storageserver/rpc/cluster_controller_api_rpc_service.cpp
+++ b/storage/src/vespa/storage/storageserver/rpc/cluster_controller_api_rpc_service.cpp
@@ -7,6 +7,7 @@
 #include <vespa/storage/storageserver/rpcrequestwrapper.h>
 #include <vespa/vdslib/state/clusterstate.h>
 #include <vespa/fnet/frt/supervisor.h>
+#include <vespa/fnet/frt/require_capabilities.h>
 #include <vespa/fnet/frt/rpcrequest.h>
 #include <vespa/storageapi/message/state.h>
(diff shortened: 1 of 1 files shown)


storage/src/vespa/storage/storageserver/rpc/cluster_controller_api_rpc_service.cpp
```diff
@@ -32,10 +33,21 @@ void ClusterControllerApiRpcService::close() {
     _closed.store(true);
 }
 
+namespace {
+
+std::unique_ptr<FRT_RequireCapabilities> make_cc_api_capability_filter() {
+    return std::make_unique<FRT_RequireCapabilities>(vespalib::net::tls::CapabilitySet::of({
+        vespalib::net::tls::Capability::content_cluster_controller_internal_state_api()
```
bjorncs, line 40: Consider using a name convention with `__` as "namespace" delimiter (e.g `content__cluster_controller__internal_state_api`

vekterli: Identifiers containing double underscores are technically reserved, so I'm keeping them single (at least for now) to avoid angering the compiler gods. 👹

https://en.cppreference.com/w/cpp/language/identifiers

> An identifier can be used to name objects, references, functions, enumerators, types, class members, namespaces, templates, template specializations, parameter packs (since C++11) goto labels, and other entities, with the following exceptions:
> (...)
> the identifiers with a double underscore anywhere are reserved;

geirst (approved):
👍

PR #23763, opened 2022-08-24
Bump hadoop to version 3.3.4

baldersheim:
@lesters or @gjoranv or @bjorncs PR


baldersheim:
Due to #23635 does not work right away.

gjoranv (approved):
Did hadoop and/or pig stop using log4j?

bjorncs (approved):

baldersheim:
Seems so. It was listed as a dependency on 3.3.2, but not on 3.3.3 or 3.3.4.
https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-common/3.3.3

PR #23764, opened 2022-08-24
Purge old containers dimension value

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminImpl.java
```diff
@@ -139,11 +140,20 @@ public void updateMetrics(boolean isSuspended) {
...
+            updateMemoryOverheadMetric(numContainers, meminfo.memTotalBytes() - meminfo.memAvailableBytes() - totalContainerMemoryBytes);
         }
     }
 
+    private void updateMemoryOverheadMetric(int numContainers, double memoryOverhead) {
+        final String name = "mem.system.overhead";
+        Dimensions dimensions = new Dimensions(Map.of("containers", Integer.toString(numContainers)));
+        metrics.declareGauge(Metrics.APPLICATION_HOST, name, dimensions, Metrics.DimensionType.DEFAULT)
+               .sample(memoryOverhead);
+        if (previousMemoryOverheadDimensions != null && !previousMemoryOverheadDimensions.equals(dimensions))
```
freva, line 152:
```suggestion
        if (!dimensions.equals(previousMemoryOverheadDimensions))
```

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeadmin/NodeAdminImpl.java
hakonhall: This would be wrong if previousMemoryOverheadDimensions is null.

freva: Doh! :facepalm:

PR #23773, opened 2022-08-24
add go code for vespa-deploy

arnej27959:
* using same logic as perl script as far as possible
* uses vespa-curl-wrapper for HTTP with TLS

note: as this is a port from perl it needs more rewriting to idiomatic go, i guess.

### DIFF
--- a/client/go/cmd/deploy/activate.go
+++ b/client/go/cmd/deploy/activate.go
@@ -1,0 +1,44 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// vespa-deploy command
+// Author: arnej
+
+package deploy
+
+import (
+	"fmt"
+	"strconv"
+)
+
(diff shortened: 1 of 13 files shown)


mpolden (commented):
Ideally Vespa CLI, vespa-logfmt, vespa-deploy and friends would just be "frontends" for the library code in the `vespa` package. That way we can eventually drop dependencies on curl, tar etc.

Rewriting the code might be out of scope? In that case porting is a good first step.

hmusum (approved):
Looks good to me!

arnej27959:
> Ideally Vespa CLI, vespa-logfmt, vespa-deploy and friends would just be "frontends" for the library code in the vespa
> package. That way we can eventually drop dependencies on curl, tar etc.

I agree very much.  I started out looking at the code in vespa/deploy.go where there is an uploadApplication() already, for example, but I think we need to reconsider structure and layering first.

PR #23774, opened 2022-08-24
fix copy-paste bug

arnej27959:
vespa-logfmt should handle multiple files just fine...

@geirst please review
@kkraune FYI


kkraune (approved):

geirst (commented):
👍

PR #23776, opened 2022-08-25
If you get a ClassCastException when doing clone by reflection it mig…

baldersheim:
…ht be that you are using a

staale method reference that has been replaced by new code from a new/reloaded bundle.
If so we clear the method cache and gives it a new try.

@bratseth @jonmv @gjoranv @bjorncs  PR


bratseth (approved):
👍

bjorncs (commented):
👍

jonmv:
👍

PR #23780, opened 2022-08-25
switch to vespa-deploy in go

arnej27959:
@hmusum please review
@aressem FYI - we may need to revert this later today if it breaks something


hmusum (approved):
I don't understand why the name is `vespalog_deploy` in `add_custom_target`, otherwise LGTM

PR #23783, opened 2022-08-25
Wire in query parameter profileDepth to control how deep down in the …

baldersheim:
…depency tree

profiling shall go.

@havardpe @bratseth PR

### DIFF
--- a/searchlib/src/protobuf/search_protocol.proto
+++ b/searchlib/src/protobuf/search_protocol.proto
@@ -24,6 +24,7 @@ message SearchRequest {
     bytes grouping_blob = 15;   // serialized opaquely like now, to be changed later
     string geo_location = 16; // to be moved into query_tree
     bytes query_tree_blob = 17; // serialized opaquely like now, to be changed later
+    int32 profileDepth = 18;
 }
 
 message TensorProperty {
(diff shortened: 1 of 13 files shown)


bratseth (approved):
- Does this number have any semantics which makes sense for the user? Otherwise, isn't it just level? Or, if it should be controlled separately, shouldn't level, set defaults? Not that we have traces from two systems perhaps level should be coarse-grained knob impacting everything and overridable per "topic"?
- The above is probably best discussed in the arch meeting needed for this anyway ...
- Why u keep removing my paragraph spacing in javadoc comments?

baldersheim:
Regarding the paragraph spacing, I suspect that might be the fresh IntelliJ in my fresh machine. I will see if I find some settings there. Will undo.

The depth has a meaning to the user. If the calltree is A -> B -> C -> D. At level 1 you will only see A. Everything called from A will be part of A (inclusive time). At level 2 you will see A and B, and any others directly called from A.

searchlib/src/protobuf/search_protocol.proto
```diff
@@ -24,6 +24,7 @@ message SearchRequest {
     bytes grouping_blob = 15;   // serialized opaquely like now, to be changed later
     string geo_location = 16; // to be moved into query_tree
     bytes query_tree_blob = 17; // serialized opaquely like now, to be changed later
+    int32 profileDepth = 18;
```
havardpe, line 27: should use snake case as the other fields (profile_depth)

baldersheim: Fixed

PR #23788, opened 2022-08-25
go function loading default-env.txt

arnej27959:
@mpolden please review

this will eventually replace some perl code inside vespa/libexec/vespa/common-env.sh

### DIFF
--- a/client/go/vespa/load_env.go
+++ b/client/go/vespa/load_env.go
@@ -53,6 +53,9 @@ func LoadDefaultEnv() error {
 			err = fmt.Errorf("unknown action '%s'", action)
 		}
 	}
+	if err == nil {
+		err = scanner.Err()
+	}
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


client/go/vespa/load_env_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLoadEnv(t *testing.T) {
+	// setup
+	tmp := os.TempDir() + "/load_env_test.tmp"
```
mpolden, line 14: Nit: You can use `t.TempDir()` which is automatically created and removed after test ends.

client/go/vespa/load_env.go
```diff
@@ -0,0 +1,106 @@
...
+		case "fallback":
+			if os.Getenv(varName) == "" {
+				os.Setenv(varName, varVal)
+			}
+		case "unset":
+			os.Unsetenv(varName)
+		default:
+			err = fmt.Errorf("unknown action '%s'", action)
+		}
+	}
```
mpolden, line 55: Consider checking if the scanner had any errors (`scanner.Err()`).

mpolden (approved):

PR #23792, opened 2022-08-25
remove duplicate target with same command

arnej27959:
@aressem or @hmusum or @baldersheim please review and merge


arnej27959:
Note: this duplicate caused the "go" build to run twice concurrently, causing failures.

hmusum (approved):

PR #23795, opened 2022-08-26
Wire in profile_depth

baldersheim:
@havardpe PR


searchlib/src/vespa/searchlib/engine/proto_converter.cpp
```diff
@@ -58,6 +58,7 @@ ProtoConverter::search_request_from_proto(const ProtoSearchRequest &proto, Searc
     request.maxhits = proto.hits();
     request.setTimeout(1ms * proto.timeout());
     request.setTraceLevel(proto.trace_level());
+    request.trace().setProfileDepth(proto.profile_depth());
```
havardpe, line 61: test should be extended to verify conversion of this field

havardpe (approved):

searchlib/src/vespa/searchlib/engine/proto_converter.cpp
baldersheim: Done

PR #23798, opened 2022-08-26
Bratseth/fastpath hashmap and set

bratseth:


jonmv (approved):

container-core/src/main/java/com/yahoo/processing/request/CloneHelper.java
```diff
@@ -7,6 +7,8 @@
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
```
baldersheim, line 11: Not used, was it the intention to add fast path for LinkedHashMap too ?

baldersheim: Not necessary as it extends hashMap it seems.

bratseth: Yes, that's what I found out too, I'll remove the import later ...

PR #23801, opened 2022-08-26
Do not include exception message when an unexpected error happens

mpolden:
So we don't leak anything potentially sensitive.

@jonmv

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
@@ -101,6 +101,7 @@ import com.yahoo.vespa.hosted.controller.maintenance.ResourceMeterMaintainer;
 import com.yahoo.vespa.hosted.controller.notification.Notification;
 import com.yahoo.vespa.hosted.controller.notification.NotificationSource;
 import com.yahoo.vespa.hosted.controller.persistence.SupportAccessSerializer;
+import com.yahoo.vespa.hosted.controller.restapi.ErrorResponses;
 import com.yahoo.vespa.hosted.controller.routing.RoutingStatus;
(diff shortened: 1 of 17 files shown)


mpolden:
FYI @tokle

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -232,7 +232,7 @@ public HttpResponse auditAndHandle(HttpRequest request) {
         }
         catch (RuntimeException e) {
             log.log(Level.WARNING, "Unexpected error handling '" + request.getUri() + "'", e);
-            return ErrorResponse.internalServerError(Exceptions.toMessageString(e));
+            return ErrorResponse.internalServerError("Unexpected error occurred");
```
jonmv, line 235: Should we perhaps generate a UUID here, and place with both the response, and in the log, to make it easier to find relevant log entries when a user asks what happened?

jonmv: ... or is that unnecessary?

mpolden:
True, I can do the same other controller handlers too.

mpolden:
> and then there are other exceptions which may leak

True, but the catch-all is likely the worst offender.

> as well as all the other handlers

Updated all handlers in the controller now. PTAL.

jonmv:
Nice. I wonder if perhaps a server timestamp, with millis, is better than a UUID, though, since the log is ordered on time? What do you think?

jonmv (approved):

mpolden:
No hard preference, but I don't see how a timestamp as ID is better? The access log and Vespa will already contain a timestamp.

jonmv:
Yes, the point was that with a server timestamp, it'll be super-easy to find the log entry, since we'd otherwise need the approximate time of the error as auxiliary data, before searching the log for the UUID.

PR #23803, opened 2022-08-26
Arnej/detect hostname and more

arnej27959:
@mpolden please review
not tested on windows; also needs testing in IPv6-only environment.
ideally various broken network setups should be checked also.

@havardpe could maybe take a look and compare with C++ code for hostname detection and validation.

Adds:
```
func ExportDefaultEnvToSh() error
func FindOurHostname() (string, error)
func FindVespaUser() string
func HasOnlyIpV6() bool
```

### DIFF
--- a/client/go/vespa/detect_hostname_test.go
+++ b/client/go/vespa/detect_hostname_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func TestDetectHostname(t *testing.T) {
-	os.Setenv("VESPA_HOSTNAME", "foo.bar")
+	t.Setenv("VESPA_HOSTNAME", "foo.bar")
 	got, err := FindOurHostname()
 	assert.Equal(t, "foo.bar", got)
 	os.Unsetenv("VESPA_HOSTNAME")
--- a/client/go/vespa/detect_hostname_test.go
+++ b/client/go/vespa/detect_hostname_test.go
@@ -12,16 +12,19 @@ import (
 
 func TestDetectHostname(t *testing.T) {
 	os.Setenv("VESPA_HOSTNAME", "foo.bar")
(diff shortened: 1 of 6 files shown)


havardpe:
C++ does canonical hostname lookup based on hostname while the go code seems to do reverse lookup based on ip address instead. The C++ code requires that all ip addresses returned by the dns lookup of a potential hostname actually belongs to us; the go code seems a bit more relaxed on this one. There does not seem to be any special handling of loopback interfaces in C++

client/go/vespa/detect_hostname_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDetectHostname(t *testing.T) {
+	os.Setenv("VESPA_HOSTNAME", "foo.bar")
```
mpolden, line 14: Consider using `t.Setenv`.

client/go/vespa/load_env_test.go
```diff
@@ -98,3 +99,59 @@ override VESPA_V2 v2
...
+	setup(t, `
+override VESPA_USER unprivuser
+`)
+	LoadDefaultEnv()
+	u = FindVespaUser()
+	assert.Equal(t, "unprivuser", u)
+}
+
+func TestExportEnv(t *testing.T) {
+	os.Setenv("VESPA_FOO", "was foo")
```
mpolden, line 120: Consider using `t.Setenv` which automatically clears the variables after this test finishes.

mpolden (approved):

mpolden (approved):
👍

PR #23815, opened 2022-08-27
Clone map with sufficient initial size

bratseth:
@baldersheim or @jonmv please review


jonmv (commented):
The config def changes weren't supposed to be here?

bratseth:
No, too quick with the -am. I'll redo this.

bratseth:
https://github.com/vespa-engine/vespa/pull/23820/files

PR #23831, opened 2022-08-29
Support specifying Athenz key pair through environment

mpolden:
This should fix basic HTTP testing for main.

@jonmv


jonmv (approved):
Nice! We should enable these tests for some on-prem integration tests next.

PR #23847, opened 2022-08-30
Cleanup after failed component graph [run-systemtest]

gjoranv:
FYI: @bjorncs

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
+++ b/container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
@@ -6,13 +6,14 @@ import com.yahoo.osgi.Osgi;
 import org.osgi.framework.Bundle;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 
 /**
  * Manages the set of installed and active/inactive bundles.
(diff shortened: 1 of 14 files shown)


container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
```diff
@@ -48,54 +48,76 @@ public ApplicationBundleLoader(Osgi osgi, FileAcquirerBundleInstaller bundleInst
...
+     * exclusively belongs to the latest (failed) application generation. Uninstalling must
+     * be done by the Deconstructor as they may still be needed by components from the failed gen.
+     */
+    public synchronized Collection<Bundle> revertToPreviousGeneration() {
+        reference2Bundle.putAll(obsoleteBundles);
+        bundlesFromNewGeneration.forEach(reference2Bundle::remove);
+        Collection<Bundle> ret = bundlesFromNewGeneration.values();
+
+        // No duplicate bundles should be allowed until the next call to useBundles.
+        osgi.allowDuplicateBundles(Set.of());
```
jonmv, line 75: This forcibly disallows duplicates, even though there could be running code from these bundles, as the new, failed graph isn't yet deconstructed. The normal flow is that a bundle is removed from the set of allowed duplicates (`allowedDuplicates` in `CollisionHook`) when that bundle is uninstalled. So can the `filter(...)` method of that class get called after we disallow duplicates _here_, but before the bundles are all cleaned out, and what happens in that case?

jonmv: If, instead, the argument was the set of new bundles, from the failed graph, which should be deconstructed, then these should be purged "in the normal fashion" as the revert was carried out, and code from both bundles would continue working normally until then ... no? (Assuming we keep the `CollisionHook` code which ensures the new allowed duplicates aren't duplicates of something that is _also_ an allowed duplicate, i.e., my temporary fix from earlier.)

jonmv: It's more complicated, I'll give you that. If what you propose here is safe, then it's the preferred choice :)

jonmv:
Just that one question, otherwise nice!

container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
gjoranv: Great catch. But now I don't see the problem of clearing the set of allowed duplicates, making it a 'set' operation instead of 'add'.  If you disagree, you are probably correct though.

jonmv: As long as we're guaranteed no new graph setup attempt is made before we complete deconstructing the failed graph, I think you're right; `SET` will be just fine, and it's cleaner.

jonmv: Ok, so if we have an ongoing revert, and a new deployment comes in, we have triplicated some of the bundles, and this isn't handled in any case, so as long as we don't also handle that, using a `SET` operation is fine. 
Deconstruction _is_ async, but having a new deployment concurrently with it seems unlikely.

jonmv (approved):

jonmv:
If we were to support that, I think we'd need to change the collision logic to work within generations of the graph. Probably doable, but also probably not worth the effort?

gjoranv:
Redeployment before deconstruction is finished could happen, and sooner or later it probably will. I added a comment in a followup PR #23858.

PR #23849, opened 2022-08-30
Include deployment in endpoint discovery error

mpolden:
@arnej27959

FYI @kkraune

### DIFF
--- a/client/go/cmd/deploy.go
+++ b/client/go/cmd/deploy.go
@@ -78,13 +78,13 @@ $ vespa deploy -t cloud -z perf.aws-us-east-1c`,
 			}
 
 			log.Println()
-			if opts.IsCloud() {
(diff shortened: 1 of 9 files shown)


kkraune:
Better! I am still not sure if the end users would immediately understand that -z would help, we could hint in the error message, like:

`no endpoints discovered for %s, use -z for endpoints in other zones`

kkraune:
Much better!

kkraune (approved):

mpolden:
Example: 
```
$ vespa status -z prod.us-north-1
Error: service 'query' is unavailable: no endpoints discovered for deployment of vespa-team.music.mpolden in prod.us-north-1
Hint: Confirm that you're communicating with the correct zone and cluster
Hint: The -z option controls the zone
Hint: The -C option controls the cluster
```

PR #23851, opened 2022-08-30
Revert "replace two perl snippets with calls to script-utils"

hmusum:
Reverts vespa-engine/vespa#23836


mpolden (approved):

mpolden:
Host-admin has no `/opt/vespa/libexec/script-utils`. It probably needs to be added to its RPM.

PR #23869, opened 2022-08-31
allow simple hex format for dense tensors of known type

arnej27959:


bratseth (approved):
👍   needs doc

PR #23877, opened 2022-08-31
Add * to valid jvm pattern for debug address, self-hosted

kkraune:
- I am quite sure * inside [] is a literal match, not a quantifier, so no escaping
- did not yet run all unit tests, can hold merge for now

When deploying, I get
```
Success: Deployed target/application.zip
WARNING Invalid or misplaced JVM options in services.xml: -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005. See https://docs.vespa.ai/en/reference/services-container.html#jvm
```

but deployment succeeds, and I am able to debug, so it is OK self-hosted - but I think it is better without the warning unless this has side-effects


hmusum:
`*` should be allowed, agreed, moved it so that `-` is not used to create a range

kkraune:
right! running all unit tests with this now. This is from the Intellij defaults

kkraune:
all tests passed

hmusum (approved):

PR #23878, opened 2022-08-31
Bratseth/model config

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ModelIdResolver.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public static void resolveModelIds(Element component, boolean hosted) {
+        if ( ! hosted) return;
+        for (Element config : XML.getChildren(component, "config")) {
+            for (Element value : XML.getChildren(config))
+                transformModelValue(value);
+        }
+    }
+
+    /** Expans a model config value into regular config values. */
```
lesters, line 38: Expands

lesters (approved):
Brilliant!

PR #23879, opened 2022-08-31
Only reserve 4k upfront, instead of max permissible length.

baldersheim:
@bratseth @geirst @toregge PR
Counter the effects of #23168


toregge (approved):

bratseth:
👍

PR #23880, opened 2022-08-31
Revert "allow simple hex format for dense tensors of known type"

baldersheim:
Reverts vespa-engine/vespa#23869
I suspect this broke the lightgbm test


baldersheim:
@bratseth or @arnej27959 PR

bratseth:
👍

PR #23891, opened 2022-09-01
Jonmv/handle major upgrade in block window better

jonmv:
@mpolden please review.


jonmv:
Unrelated test failure

mpolden (approved):
👍

PR #23893, opened 2022-09-01
try harder to avoid disk access

arnej27959:
@bratseth please review
FieldFiller would never make things cheaper, only fill more if the requested summary class did not contain all the selected fields.


bratseth (approved):
👍  much clearer too

PR #23897, opened 2022-09-01
Minor fixes like adding 'explicit', 'override' and general unification.

baldersheim:

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/singlenumericattribute.h
+++ b/searchlib/src/vespa/searchlib/attribute/singlenumericattribute.h
@@ -44,7 +44,7 @@ public:
     explicit SingleValueNumericAttribute(const vespalib::string & baseFileName);  // Only for testing
     SingleValueNumericAttribute(const vespalib::string & baseFileName, const AttributeVector::Config & c);
 
-    ~SingleValueNumericAttribute() final;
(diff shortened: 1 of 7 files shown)


baldersheim:
@toregge PR

searchlib/src/vespa/searchlib/attribute/singlenumericattribute.h
```diff
@@ -41,10 +41,10 @@ class SingleValueNumericAttribute final : public B {
     }
 
 public:
-    SingleValueNumericAttribute(const vespalib::string & baseFileName);  // Only for testing
+    explicit SingleValueNumericAttribute(const vespalib::string & baseFileName);  // Only for testing
     SingleValueNumericAttribute(const vespalib::string & baseFileName, const AttributeVector::Config & c);
 
-    ~SingleValueNumericAttribute();
+    ~SingleValueNumericAttribute() final;
```
toregge, line 47: Consider `override` instead of `final` (which is already implicit due to class marked final).

toregge (approved):

searchlib/src/vespa/searchlib/attribute/singlenumericattribute.h
baldersheim: Fixed

PR #23908, opened 2022-09-02
Balder/wire in summary fields for streaming too

baldersheim:
@vekterli and @toregge PR


vekterli (approved):
:+1: I'll defer reviewing the summary semantics to the experts™

PR #23912, opened 2022-09-02
Rank-property 'vespa.mutate.allow_query_override: true' must be set i…

baldersheim:
…n rank-profile to allow for query side control.

@bratseth or @havardpe PR


bratseth (approved):
👍

PR #23918, opened 2022-09-04
Change matched elements filter docsum field writer to insert an empty

toregge:
array if input field is empty or not set. This was the old behavior
when jsonstring entries were stored in docsum blobs.

@geirst : please review
@baldersheim : FYI


toregge:
Inserting empty array doesn't match behavior for attribute combiner dynamic field writer, which handles the case when all involved fields are backed by attribute vectors.

geirst (approved):
👍

baldersheim:
We should discuss this one.

geirst (approved):
👍

PR #23921, opened 2022-09-05
Remove support for 'setsystemstate2' RPC method in cluster controller [run-systemtest]

hmusum:
Cleanup, only used by very old Vespa versions


vekterli (approved):
:+1:

PR #23930, opened 2022-09-05
Clear target major

jonmv:
@mpolden please review. Let's merge after next release, and then revert after this is also out.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
@@ -274,7 +274,7 @@ public class CuratorDb {
     }
 
     public void clearTargetMajor() {
-        curator.delete(root.append("ugprader").append("targetMajorVersion"));
+        curator.delete(root.append("upgrader").append("targetMajorVersion"));
     }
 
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -273,6 +273,10 @@ public void writeVersionStatus(VersionStatus status) {
         curator.set(versionStatusPath(), asJson(versionStatusSerializer.toSlime(status)));
     }
 
+    public void clearTargetMajor() {
+        curator.delete(root.append("ugprader").append("targetMajorVersion"));
```
mpolden, line 277: Typo

jonmv:
```suggestion
        curator.delete(root.append("upgrader").append("targetMajorVersion"));
```

(((＼（✘෴✘）／)))

jonmv: ┏( ﾟ益ﾟ)ゞ

jonmv:
Thanks!

mpolden (approved):

PR #23934, opened 2022-09-05
Factor out XML string escaping and use for internal legacy status pages [run-systemtest]

vekterli:
@geirst please review
@bjorncs FYI

The best approach would of course be to turn things on the head and always escape everything that _isn't_ explicitly trusted, but that would require significant rewrites of the legacy rendering code for these internal pages.


bjorncs (approved):

geirst (approved):
👍

PR #23939, opened 2022-09-05
Cleanup after failed component graph take2 [run-systemtest]

gjoranv:
Only the three last commits are new and relevant.
FYI: @bjorncs

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/di/Container.java
+++ b/container-core/src/main/java/com/yahoo/container/di/Container.java
@@ -72,6 +72,7 @@ public class Container {
         });
     }
 
+    // TODO: try to simplify by returning the result even when the graph failed, instead of throwing here.
     public ComponentGraphResult waitForNextGraphGeneration(ComponentGraph oldGraph, Injector fallbackInjector, boolean isInitializing) {
         try {
             Collection<Bundle> obsoleteBundles = new HashSet<>();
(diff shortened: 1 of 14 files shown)


gjoranv:
Force push was just to amend the commit message.

container-core/src/main/java/com/yahoo/container/di/Container.java
```diff
@@ -70,20 +70,26 @@ public Container(SubscriberFactory subscriberFactory, String configId, Component
...
             Collection<Bundle> obsoleteBundles = new HashSet<>();
-            ComponentGraph newGraph = waitForNewConfigGenAndCreateGraph(oldGraph, fallbackInjector, isInitializing, obsoleteBundles);
-            newGraph.reuseNodes(oldGraph);
+            try {
+                newGraph = waitForNewConfigGenAndCreateGraph(oldGraph, fallbackInjector, isInitializing, obsoleteBundles);
+                newGraph.reuseNodes(oldGraph);
+            } catch (Throwable t) {
+                log.warning("Failed to set up component graph - uninstalling latest bundles. Bootstrap generation: " + getBootstrapGeneration());
+                osgi.completeBundleGeneration(Osgi.GenerationStatus.FAILURE);
+                throw t;
```
jonmv, line 81: Any started bundles here are never uninstalled, as in `bundle.uninstall()`, as far as I can see.

jonmv: It looks like it would be OK to create a similar failed graph deconstruction task.

gjoranv: Correct. I'll see if I can fix it easily, or add a TODO. They are not installed today, and not in the previous attempted fix either.

jonmv:
LGTM :)

gjoranv:
I think this is near bulletproof now. Deconstruction after a failed generation is done without delay, so the risk of in-between reconfig should be very low.

jonmv (approved):
Great!

bjorncs:
👍

PR #23940, opened 2022-09-06
Cluster controller cleanup, part 11 [run-systemtest]

hmusum:
Remove fields fleetControllerConfigId and slobrokConfigId.
Never set, so no use in keeping them or showing what they are.


baldersheim (approved):

hmusum:
@vekterli FYI

vekterli (commented):
:+1:

PR #23941, opened 2022-09-06
Revert "Clear target major"

jonmv:
Reverts vespa-engine/vespa#23930

@mpolden


jonmv:
Let's merge when 8.49 is out

mpolden (approved):

PR #23942, opened 2022-09-06
Cluster controller cleanup, part 12 [run-systemtest]

hmusum:
Remove unused config field, has not been used for since version 4.1


vekterli (approved):
:+1: `VDS 4.1 node state polling frequency` sounds like this was _well_ overdue for retirement 🦖

PR #23951, opened 2022-09-06
Make two-phase GC work for parent-child with subset of replicas indexed [run-systemtest]

vekterli:
@geirst please review

The previous iteration of GC 1st phase candidate set computation required
_all_ replicas to agree that a particular document should be removed
for it to be passed on to the second phase. I.e. the intersection of all
nodes' document sets. This does not work as expected when the GC
expression references imported fields _and_ `searchable-copies` is
less than `redundancy`, as the required index structures are not present
across all replicas. The result was that eligible documents were never
removed.

This commit changes the candidate set semantics to instead use a
union of document IDs, using the maximum observed timestamp in the case
of conflicts for the same ID. This mirrors the end result of the legacy
behavior, but does not require merging in order to propagate tombstones
from the indexed replicas to those without. It also greatly simplifies
the candidate computation code.


geirst (approved):
👍

PR #23954, opened 2022-09-07
GC unused StringEnum, and avoid including config in header files.

baldersheim:
@toregge or @geirst PR

PR #23955, opened 2022-09-07
Remove unused search::util::StringEnum class.

toregge:
@geirst : please review


toregge:
Class is also removed in #23954

PR #23957, opened 2022-09-07
Ensure comp.platform for all changes of change, and do not override p…

jonmv:
…resent ones

Without this, a pin to downgrade majors would be cleared by the relatively new code which aimed to ensure a previous pin from a previous downgrade didn't stop a new, attempted upgrade, across incompatible majors ... Gosh :) 
Also, I think it's better to ensure the compatibility platform is always added (when manipulating change through other means than making a new submission).

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
@@ -178,18 +178,18 @@ public class DeploymentStatus {
                                                  .map(application.revisions()::get)
                                                  .flatMap(ApplicationVersion::compileVersion);
 
-        // If the revision requires a certain platform for compatibility, add that here.
+        // If the revision requires a certain platform for compatibility, add that here, unless we're already deploying a compatible platform.
         VersionCompatibility compatibility = versionCompatibility.apply(instance);
         Predicate<Version> compatibleWithCompileVersion = version -> compileVersion.map(compiled -> compatibility.accept(version, compiled)).orElse(true);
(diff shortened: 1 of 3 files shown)


jonmv:
@bjormel FYI :)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
```diff
@@ -188,7 +188,10 @@ public Change withCompatibilityPlatform(Change change, InstanceName instance) {
                     .stream() // Pick the latest platform with appropriate confidence, which is compatible with the compile version.
                     .filter(compatibleWithCompileVersion)
                     .findFirst()
-                    .map(platform -> change.withoutPin().with(platform))
+                    .map(platform -> change.platform()
+                                           .filter(platform::equals)
```
jonmv, line 192: Hmm, perhaps this should instead be a check on whether that platform is compatible with the compile version. That could also be done outside all of this ... Let me fix.

vespajlib/src/main/java/com/yahoo/vespa/VersionTagger.java
```diff
@@ -68,6 +68,7 @@ private Map<String, String> readVtagMap(String path) {
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
+        System.err.println(map);
```
mpolden, line 71: Probably doesn't belong here?

jonmv: Thanks.

mpolden:
👍

PR #23963, opened 2022-09-07
Bratseth/warn on old major

bratseth:


jonmv (approved):
Sweet.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -597,6 +579,50 @@ public ApplicationVersion submit(TenantAndApplicationId id, Submission submissio
...
+                                                             "lower major version than the current Vespa Cloud version");
+            else
+                controller.notificationsDb().removeNotification(NotificationSource.from(id), Type.submission);
+        });
+   }
+
+    private void validateMajorVersion(TenantAndApplicationId id, Submission submission) {
+        submission.applicationPackage().deploymentSpec().majorVersion().ifPresent(explicitMajor -> {
+            if (explicitMajor < 8)
+                controller.notificationsDb().setNotification(NotificationSource.from(id),
```
freva, line 616: Notifications are key'ed on (source, type) (see https://github.com/vespa-engine/vespa/blob/ceb337dea57c909ac83bc168e1b79872eae6dbfe/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/NotificationsDb.java#L62), so any application with explicit major will either override the `applicationPackage` warnings from configserver prepare or clear them all together. This should probably be together with `submission`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -597,6 +579,50 @@ public ApplicationVersion submit(TenantAndApplicationId id, Submission submissio
...
+                                                             Type.applicationPackage,
+                                                             Notification.Level.warning,
+                                                             "Vespa 7 will soon be end of life, upgrade to Vespa 8 now:" +
```
freva, line 619: Missing space between text end and URL

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
jonmv: Good point, I didn't notice this. I'll fix it in a PR soon.

bratseth: It's a property of the application package, which I think holds for all the submission warnings, so submission is just created as a workaround to this problem? It doesn't seem like we have this entirely right - why keep warnings associated with the application package around if they are not generated by the current version of it? And given that why should code emitting individual warnings clear a _category_ of them?

I think we should 
- Combine applicationPackage and submission to one category, unless we really have warnings associated with the parameters of the submission.
- Always clear this category when a new submission is received.
- Remove any code clearing warnings from individual validations.

freva: The main issue is that the different notifications are generated in different places at different times, and because we also want to automatically clear the notification once the issue is fixed, there must be a single source for each type. 

The `submission` type is just application package warnings generated in controller, while `applicationPackage` are generated in the configserver (and stored in the controller ZK after a deploy, which happens at a different time and by a different maintainer). But both types are emitted as `applicationPackage` to the console https://github.com/vespa-engine/vespa/blob/5f19d58210dee17bf2c8161b6fbea66163cd79d6/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java#L881

bratseth: Okay, so "applicationPackage" are warnings issued by the config model etc., which are managed centrally already, while "submission" are application package warnings issued by the controller, which are currently not managed centrally.

bratseth: Clearing those owned by the controller on submission: https://github.com/vespa-engine/vespa/pull/23975

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
bratseth: Thanks, fixed.

PR #23969, opened 2022-09-07
Remove unused and undocumented partition state request support

hmusum:


vekterli (approved):
:+1: "partition" was a very generic term that only ever mapped to "spinning disk", and explicitly having these as part of the cluster state has been deprecated for ages

PR #23971, opened 2022-09-07
Always return the set of bundles to uninstall upon completeGen [run-systemtest]

gjoranv:
Cleaner, and no more return value parameter.

FYI: @bjorncs


jonmv (approved):
Ncie!

PR #23980, opened 2022-09-08
Add frame-ancestors: 'none' to CSP headers returned by internal server

vekterli:
@bjorncs please review
@havardpe FYI

We already have (and will continue to have) `X-Frame-Options: DENY`, but this is the more modern version of it.


bjorncs (approved):

bjorncs:
FYI @tokle

PR #23983, opened 2022-09-08
Use attribute vector type instead of summary field type in SingleAttr…

toregge:
…DFW.

@geirst : please review
@baldersheim : FYI


baldersheim (approved):

geirst (approved):
👍

PR #23988, opened 2022-09-08
Use int64_t and double when printing values for single value numeric …

toregge:
…attributes.

@geirst : please review
@baldersheim : FYI


toregge:
This PR fixes a regression introduced by #23983

bratseth (approved):

baldersheim (commented):
👍

PR #23989, opened 2022-09-08
Don't render field in search result if it is empty after matched elem…

toregge:
…ents filtering.

@geirst : please review
@baldersheim : FYI


baldersheim (approved):

geirst (approved):
👍

PR #23990, opened 2022-09-09
Expose ?query parameter for find user. Output is now array

oyving:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserApiHandler.java
@@ -101,7 +101,7 @@ public class UserApiHandler extends ThreadedHttpRequestHandler {
 
     private HttpResponse handleGET(Path path, HttpRequest request) {
         if (path.matches("/user/v1/user")) return userMetadata(request);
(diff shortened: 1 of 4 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserApiHandler.java
```diff
@@ -134,18 +134,44 @@ private HttpResponse handleOPTIONS() {
...
             var roles = users.listRoles(new UserId(user.email()));
-            return renderUserMetaData(user, Set.copyOf(roles));
+            var slime = new Slime();
+            var root = slime.setObject();
+            var usersRoot = root.setArray("users");
+            renderUserMetaData(usersRoot.addObject(), user, Set.copyOf(roles));
+            return new SlimeJsonResponse(slime);
         }
 
         return ErrorResponse.notFoundError("Could not find user: " + email);
```
freva, line 158: Return empty list instead?

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/user/UserApiHandler.java
oyving: 👍 will update.

freva (approved):

PR #23991, opened 2022-09-09
Allow setting cloud account for non-production environments

mpolden:
@jonmv


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/JobController.java
```diff
@@ -702,7 +702,7 @@ public void deploy(ApplicationId id, JobType type, Optional<Version> platform, A
         controller.applications().lockApplicationOrThrow(TenantAndApplicationId.from(id), application -> {
             if ( ! application.get().instances().containsKey(id.instance()))
                 application = controller.applications().withNewInstance(application, id);
-
+            controller.applications().validatePackage(applicationPackage, application.get());
```
jonmv, line 705: 👍

jonmv (approved):
Nice!

mpolden:
`com.yahoo.vespa.clustercontroller.core.MasterElectionTest` strikes again 😞

mpolden:
Tried running a full build locally, but then `FleetControllerTest.waitForStableSystem` failed. 😆

PR #23999, opened 2022-09-09
Revert "Allow setting cloud account for non-production environments"

jonmv:
Reverts vespa-engine/vespa#23991

It's not this PR itself that is the problem, but it makes integration tests with invalid deployment-spec for their system fail (which they should, they just haven't been validated because it was direct deployments).


jonmv:
@hmusum or @mpolden or anyone else, please merge.

mpolden (approved):

PR #24001, opened 2022-09-09
Add a 'binding' element under 'search'.

gjoranv:


baldersheim:
@tokle ping

gjoranv:
This is just adding an element that is allowed in the schema, but was not in any test file. Anyone can approve and merge.

PR #24003, opened 2022-09-10
Avoid multiple definitions of juniper::QueryItem.

toregge:
Use single definition of ItemCreator enumeration in C++.

@geirst : please review
@baldersheim : FYI


baldersheim (approved):

geirst (commented):
👍

PR #24007, opened 2022-09-11
No functional changes

bratseth:
Minor cleanup

### DIFF
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
@@ -63,7 +63,7 @@ public class LinguisticsAnnotator {
      * Annotates the given string with the appropriate linguistics annotations.
      *
      * @param text the text to annotate
-     * @return whether or not anything was annotated
+     * @return whether anything was annotated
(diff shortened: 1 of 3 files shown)


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
```diff
@@ -116,42 +116,39 @@ private static void addAnnotationSpan(String input, SpanList parent, Token token
             }
             if ( ! token.isIndexable()) return;
         }
-        String orig = token.getOrig();
-        int pos = (int)token.getOffset();
-        if (pos >= input.length()) {
-            throw new IllegalArgumentException("Token '" + orig + "' has offset " + pos + ", which is outside the " +
-                                               "bounds of the input string; " + input);
+        if (token.getOffset() >= input.length()) {
```
baldersheim, line 119: This does not come without a performance cost.
At least one of the Token implementations uses lazy lookup of the token when calling getOrig(),
and getOffset will cause a non cached jni call that we want to avoid.

baldersheim (changes requested):
Keep the caching of orig and offset.

bratseth:
Fixed the token implementation.

baldersheim (commented):
👍

PR #24011, opened 2022-09-12
GlobalFilter is now an interface

havardpe:
instead of a shared optional BitVector

@arnej27959 please review
@geirst FYI


geirst (approved):
👍

baldersheim (approved):

PR #24012, opened 2022-09-12
Validate application package before direct deployment

mpolden:
@jonmv

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
@@ -696,7 +696,8 @@ public class ControllerTest {
 
     @Test
     void testDevDeployment() {
-        ApplicationPackage applicationPackage = new ApplicationPackageBuilder().build();
+        // A package without deployment.xml is considered valid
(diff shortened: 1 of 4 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
```diff
@@ -696,7 +696,7 @@ void testDnsUpdatesForApplicationEndpoint() {
 
     @Test
     void testDevDeployment() {
-        ApplicationPackage applicationPackage = new ApplicationPackageBuilder().build();
+        ApplicationPackage applicationPackage = ApplicationPackageBuilder.fromDeploymentXml(DeploymentSpec.empty.xmlForm());
```
jonmv, line 699: How about no `deployment.xml` at all?

jonmv (commented):
One question, otherwise good :)

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
mpolden: It's fine, but I updated the test to use a completely empty package.

jonmv:
Great!

PR #24015, opened 2022-09-12
add CLI for get-cluster-state and friends

arnej27959:
@mpolden please review
@vekterli please review


mpolden (approved):
LGTM, but I'm not familiar with the original code.

A few tests would be nice. For example asserting that the expected output is printed given some cluster state JSON blob.

vekterli (commented):
:+1: agreed that it would be good to have some testing of producing/consuming expected JSON

✨ _Achievement unlocked: "learn" a new programming language to review a PR_ 🦄✨

PR #24018, opened 2022-09-12
Run test jobs only in clouds for which there are dependent deployments

jonmv:
@freva please review ┌(˘⌣˘)ʃ


freva (approved):
:shrug: :exploding_head:

PR #24024, opened 2022-09-12
Jonmv/refuse new apps on old majors

jonmv:
@mpolden please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VersionStatusUpdater.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VersionStatusUpdater.java
@@ -55,7 +55,6 @@ public class VersionStatusUpdater extends ControllerMaintainer {
             case legacy -> legacy;
             case normal -> normal;
             case high -> high;
-            default -> throw new IllegalArgumentException("Unexpected confidence '" + confidence + "'");
         };
     }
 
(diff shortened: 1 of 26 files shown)


jonmv:
It struck me that, since major-version from deployment.xml is used by the config server, this is really a property of the application version, not of the global deployment spec. I didn't fully fix this, but it should likely not be an issue due to the way we orchestrate things. Something to be aware of, though ...

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VersionStatusUpdater.java
```diff
@@ -47,14 +48,15 @@ protected double maintain() {
...
-            default: throw new IllegalArgumentException("Unexpected confidence '" + confidence + "'");
-        }
+        return switch (confidence) {
+            case aborted -> aborted;
+            case broken -> broken;
+            case low -> low;
+            case legacy -> legacy;
+            case normal -> normal;
+            case high -> high;
+            default -> throw new IllegalArgumentException("Unexpected confidence '" + confidence + "'");
```
mpolden, line 58: I think `default` can be removed since the switch is exhaustive.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/versions/VersionStatus.java
```diff
@@ -262,4 +265,14 @@ private static VespaVersion createVersion(DeploymentStatistics statistics,
                                 confidence);
     }
 
+    /** Whether no version on a newer major, with normal or higher confidence, can be deployed. */
+    public boolean isOnCurrentMajor(Version version) {
+        for (VespaVersion available : deployableVersions())
+            if (   available.confidence().equalOrHigherThan(Confidence.high)
```
mpolden, line 271: Shouldn't this be `equalOrHigherThan(Confidence.normal)`, or is the javadoc wrong?

jonmv: Wrong javadoc, thanks.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/VersionStatusUpdater.java
jonmv: Fixed, thanks.

mpolden (approved):
Great!

PR #24025, opened 2022-09-13
Disable splitting of same-element until we can also handle that with …

baldersheim:
…matched-elements-only.

@havardpe @geirst PR

### DIFF
--- a/searchcore/src/tests/proton/matching/unpacking_iterators_optimizer/unpacking_iterators_optimizer_test.cpp
+++ b/searchcore/src/tests/proton/matching/unpacking_iterators_optimizer/unpacking_iterators_optimizer_test.cpp
@@ -291,7 +291,7 @@ TEST(UnpackingIteratorsOptimizerTest, require_that_root_same_element_node_can_be
 }
 
 #if ENABLE_SAME_ELEMENT_SPLIT
-//TODO Enable once matched-elements-only and artifiscial terms are handled
+//TODO Enable once matched-elements-only and artificial terms are handled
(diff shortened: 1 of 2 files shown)


searchcore/src/tests/proton/matching/unpacking_iterators_optimizer/unpacking_iterators_optimizer_test.cpp
```diff
@@ -124,8 +124,14 @@ Node::UP make_query_tree() {
     builder.addOr(3);
     builder.addStringTerm("t2", view, id, weight);
     add_phrase(builder);
+#if ENABLE_SAME_ELEMENT_SPLIT
+    //TODO Enable once matched-elements-only and artifiscial terms are handled
```
geirst, line 128:
```suggestion
    //TODO Enable once matched-elements-only and artificial terms are handled
```

searchcore/src/tests/proton/matching/unpacking_iterators_optimizer/unpacking_iterators_optimizer_test.cpp
```diff
@@ -301,11 +290,14 @@ TEST(UnpackingIteratorsOptimizerTest, require_that_root_same_element_node_can_be
     EXPECT_EQ(actual3, expect);
 }
 
+#if ENABLE_SAME_ELEMENT_SPLIT
+//TODO Enable once matched-elements-only and artifiscial terms are handled
```
geirst, line 294:
```suggestion
//TODO Enable once matched-elements-only and artificial terms are handled
```

geirst (approved):
👍

PR #24041, opened 2022-09-14
Use snippet modifier on each element in array of string when dynamic …

toregge:
…summary is requested and snippet modifier is available (streaming search).

@geirst : please review


geirst (approved):
👍

PR #24044, opened 2022-09-14
Validate file path exists

tokle:


config-model/src/main/java/com/yahoo/vespa/model/admin/LogForwarder.java
```diff
@@ -5,6 +5,10 @@
 import com.yahoo.config.model.producer.AbstractConfigProducer;
 import com.yahoo.vespa.model.AbstractService;
 import com.yahoo.vespa.model.PortAllocBridge;
+
+import javax.swing.text.html.Option;
```
arnej27959, line 9: this import does not belong here...

config-model/src/main/java/com/yahoo/vespa/model/admin/LogForwarder.java
```diff
@@ -86,7 +90,11 @@ public Optional<String> getPreShutdownCommand() {
         getConfig(builder);
         var cfg = new LogforwarderConfig(builder);
         var home = cfg.splunkHome();
-        return Optional.of("/usr/bin/env SPLUNK_HOME="+home+" "+home+"/bin/splunk stop");
+        if (Files.exists(Path.of(home))) {
```
arnej27959, line 93: it's very brittle if we're checking if a directory exists **on the configserver where the model runs** to decide whether to set up splunk forwarding.  Maybe we need some zone-wide config or a feature flag instead?  checking if the directory exists on the actual node would make more sense to me.

arnej27959 (changes requested):
we should discuss whether to use a feature flag for this instead.

PR #24058, opened 2022-09-14
Move the BERT config def to model-integration

bratseth:


lesters (approved):
Ok, but need to change CMakeLists in configdefinitions as well:

```
18:12:21 CMake Error at configdefinitions/src/vespa/cmake_install.cmake:205 (file):
18:12:21   file INSTALL cannot find
18:12:21   "/sd/workspace/src/github.com/vespa-engine/vespa/configdefinitions/src/vespa/embedding.bert-base-embedder.def":
18:12:21   No such file or directory.
```

PR #24083, opened 2022-09-15
Adjust api for DocsumWriter and DocsumFieldWriter:

toregge:
Pass reference to state instead of pointer.
Drop unused ResType argument.

@geirst : please review


geirst (approved):
👍

PR #24086, opened 2022-09-15
Make validation messages clearer given multiple instances

bratseth:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/Instance.java
```diff
@@ -220,6 +220,6 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "application '" + id + "'";
+        return "application instance '" + id + "'";
```
jonmv, line 223:
```suggestion
        return "application instance '" + id.ToFullString() + "'";
```

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
```diff
@@ -175,12 +175,13 @@ void testDeployment() {
             fail("Expected exception due to illegal production deployment removal");
         }
         catch (IllegalArgumentException e) {
-            assertEquals("deployment-removal: application 'tenant.application' is deployed in us-west-1, but does not include this zone in deployment.xml. " +
-                    ValidationOverrides.toAllowMessage(ValidationId.deploymentRemoval),
-                    e.getMessage());
+            assertEquals("deployment-removal: application instance 'tenant.application' is deployed in us-west-1, " +
```
jonmv, line 178:
```suggestion
            assertEquals("deployment-removal: application instance 'tenant.application.default' is deployed in us-west-1, " +
```

jonmv (approved):
Yes. Let's wait for unit tests before merge now though :)

PR #24057, opened 2022-09-14
Balder/only consider clusterstate when making bucket ready

baldersheim:
@vekterli @geirst @toregge PR

PR #24090, opened 2022-09-16
Do not inhibit bucket replica activation under maintenance mode [run-systemtest]

vekterli:
@geirst @baldersheim please review

To prevent spurious indexing during rolling upgrades, this also adds an internal content node feature support flag which communicates that the content node will _not_ implicitly index a non-ideal replica marked explicitly as active. Activation during maintenance is only performed iff candidate content nodes have this flag set.

This PR does not change the heuristics where the distributor prefers activating the replica with the most documents (vs. the node with the highest ideal state score) if no `ready` replicas are available. One thing at a time 🐢

### DIFF
--- a/storage/src/tests/distributor/top_level_bucket_db_updater_test.cpp
+++ b/storage/src/tests/distributor/top_level_bucket_db_updater_test.cpp
@@ -2512,6 +2512,7 @@ TEST_F(TopLevelBucketDBUpdaterTest, node_feature_sets_are_aggregated_from_nodes_
         for (uint16_t i : {0, 1, 2}) {
             EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(i).unordered_merge_chaining);
             EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(i).two_phase_remove_location);
+            EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(i).no_implicit_indexing_of_active_buckets);
         }
     }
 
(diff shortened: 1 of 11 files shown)


vekterli:
We'll hold off merging this until we've observed the effects of #24057 for a bit.

vekterli:
As a side note, if this PR is merged and #24057 is later reverted, this PR **must** be reverted alongside it as well. Otherwise the content node will be shamelessly lying about its supported feature set.

storage/src/tests/distributor/top_level_bucket_db_updater_test.cpp
```diff
@@ -2532,10 +2534,15 @@ TEST_F(TopLevelBucketDBUpdaterTest, node_feature_sets_are_aggregated_from_nodes_
     for (auto* s : stripes) {
         EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(0).unordered_merge_chaining);
         EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(0).two_phase_remove_location);
+        EXPECT_FALSE(s->node_supported_features_repo().node_supported_features(0).no_implicit_indexing_of_active_buckets);
```
baldersheim, line 2537: **node_supported_**features_repo().**node_supported_**features(0)
Butter on the fat...

baldersheim (approved):

geirst (approved):
👍

PR #24095, opened 2022-09-16
Allow path prefix alias for legacy content status pages [run-systemtest]

vekterli:
@geirst please review
@andreer @tokle FYI

This lets status pages that today are served at the root `/` path be aliased under `/contentnode-status/v1/`. Legacy paths continue working as before. Also changed existing status page absolute paths to relative to avoid having to care about this particular detail internally.

Note: both distributor and search/storage node process status pages use the `/contentnode-status/v1/` prefix, as they're both technically processes that are part of a _logical_ content node.


geirst (approved):
👍

PR #24101, opened 2022-09-16
Use bash instead of perl

baldersheim:
@toregge or @arnej27959 PR


arnej27959 (approved):
Yes please

baldersheim:
On a second note I think it is better that we just do this directly in system-test module with https://github.com/vespa-engine/system-test/pull/2538. There is no other use of this script.
I will just remove the script instead.

arnej27959:
you'll need to update dist/vespa.spec

PR #24114, opened 2022-09-19
the standard vespa packages should not require 'perf'

arnej27959:
@baldersheim please review
@aressem FYI


aressem (approved):

baldersheim (commented):
We do require perf inside both test and prod containers

arnej27959:
made a couple of PRs to add perf in base docker images

PR #24119, opened 2022-09-19
Try/catch triggering new revisions, and return success rate

jonmv:
@freva


freva (approved):

jonmv:
There'll be one more ...

jonmv:
https://github.com/vespa-engine/vespa/pull/24121

PR #24121, opened 2022-09-19
Jonmv/no compat platform when already on old major

jonmv:
@freva


jonmv:
Only the last commit is new ...

freva (approved):

PR #24127, opened 2022-09-19
Control whether to report coverage based on active or target active d…

baldersheim:
…ocuments.

@vekterli or @bratseth PR

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/logging/Coverage.java
+++ b/container-core/src/main/java/com/yahoo/container/logging/Coverage.java
@@ -9,14 +9,19 @@ public class Coverage {
     private final long active;
     private final long targetActive;
     private final int degradedReason;
+    private final boolean useTargetActiveForCoverageComputation;
     private final static int DEGRADED_BY_MATCH_PHASE = 1;
     private final static int DEGRADED_BY_TIMEOUT = 2;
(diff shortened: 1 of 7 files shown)


container-core/src/main/java/com/yahoo/container/logging/Coverage.java
```diff
@@ -9,14 +9,19 @@ public class Coverage {
...
-    public Coverage(long docs, long active, long targetActive, int degradedReason) {
+    public Coverage(long docs, long active, long targetActive, int degradedReason, boolean useTargetActiveForCoverageComputation) {
         this.docs = docs;
         this.active = active;
         this.targetActive = targetActive;
         this.degradedReason = degradedReason;
+        this.useTargetActiveForCoverageComputation = useTargetActiveForCoverageComputation;
+    }
+    Coverage(long docs, long active, long targetActive, int degradedReason) {
+        this(docs, active, targetActive, degradedReason, true);
```
vekterli, line 24: Do we want the new behavior to be the default for this constructor?

vekterli (approved):
:+1:

container-core/src/main/java/com/yahoo/container/logging/Coverage.java
baldersheim: Yes, this is a private constructor only used in tests. No reason to postpone the inevitable.

PR #24133, opened 2022-09-20
Chose correct "redundancy" for coverage estimation. Current default i…

baldersheim:
…s to use searchable-copies.

But when we flip to use target-active directly from the backend we need to use redundancy.

@vekterli PR


vekterli (approved):
:+1: should we add an explicit test with this dispatch config flag active?

baldersheim:
Test added.

vekterli (approved):
:+1:

PR #24135, opened 2022-09-20
Short circuit boolean expressions

bratseth:
Short circuit boolean expressions by converting them to (nested) if expressions. This also fixes a bug in Java expression evaluation where evaluation of arithmetic operations with the same precedence would be from right to left rather than left to right.

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformer.java
+++ b/config-model/src/main/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformer.java
@@ -1,0 +1,105 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.schema.expressiontransforms;
+
+import com.yahoo.searchlib.rankingexpression.evaluation.BooleanValue;
+import com.yahoo.searchlib.rankingexpression.rule.ArithmeticNode;
+import com.yahoo.searchlib.rankingexpression.rule.ArithmeticOperator;
+import com.yahoo.searchlib.rankingexpression.rule.CompositeNode;
+import com.yahoo.searchlib.rankingexpression.rule.ConstantNode;
+import com.yahoo.searchlib.rankingexpression.rule.ExpressionNode;
+import com.yahoo.searchlib.rankingexpression.rule.IfNode;
+import com.yahoo.searchlib.rankingexpression.transform.ExpressionTransformer;
+import com.yahoo.searchlib.rankingexpression.transform.TransformContext;
+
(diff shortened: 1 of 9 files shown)


config-model/src/main/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformer.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.List;
+
+/**
+ * Transforms
+ *        a &amp;&amp; b into if(a, b, false)
+ * and
+ *        a || b into if(a, true, b)
+ * to avoid computing b if a is false and true respectively.
+ *
+ * This may increase performance since boolean expressions are not short-circuited.
```
baldersheim, line 27: not -> now

config-model/src/main/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformer.java
```diff
@@ -24,7 +24,7 @@
  *        a || b into if(a, true, b)
  * to avoid computing b if a is false and true respectively.
  *
- * This may increase performance since boolean expressions are not short-circuited.
```
bratseth: I really meant "not", as in "not short-circuited in the backend", but ok :-)

config-model/src/main/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformer.java
```diff
@@ -0,0 +1,105 @@
...
+
+        return node;
+    }
+
+    private ExpressionNode transformBooleanArithmetics(ArithmeticNode node) {
+        Iterator<ExpressionNode> child = node.children().iterator();
+
+        // Transform in precedence order:
+        Deque<ChildNode> stack = new ArrayDeque<>();
+        stack.push(new ChildNode(ArithmeticOperator.OR, child.next()));
```
baldersheim, line 49: This causes a + b + c + d to become an unbalanced tree width depth == number of operators. Not very stack friendly.
(((a+b) + c) + d)

I suggest that you put all children with same precedence in a single arithmetic node. If we require that precedence should be simpler too.

bratseth: It'll make this code uglier, and nobody is going to add together fifty thousand things like that test, but I'll give it a shot and see if it can be done reasonably nice without changing everything.

baldersheim: I found a way that is just a minor change. PR is coming.

PR #24142, opened 2022-09-20
Optimize resolving and usage of summary result class.

geirst:
If an explicit set of fields is specified in the docsum request, we avoid reading the document instance from disk if all those fields are generated on the fly.

@toregge please review

### DIFF
--- a/searchsummary/src/vespa/searchsummary/docsummary/resultclass.cpp
+++ b/searchsummary/src/vespa/searchsummary/docsummary/resultclass.cpp
@@ -33,17 +33,16 @@ ResultClass::AddConfigEntry(const char *name, std::unique_ptr<DocsumFieldWriter>
     }
 
     _nameMap[name] = _entries.size();
-    ResConfigEntry e;
-    e._name = name;
+    ResConfigEntry e(name);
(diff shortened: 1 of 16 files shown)


searchsummary/src/vespa/searchsummary/docsummary/resultclass.cpp
```diff
@@ -54,4 +53,18 @@ ResultClass::AddConfigEntry(const char *name)
     return AddConfigEntry(name, {});
 }
 
+bool
+ResultClass::all_fields_generated(const vespalib::hash_set<vespalib::string>& fields) const
+{
+    if (fields.empty()) {
+        return _dynInfo._generateCnt == GetNumEntries();
```
toregge, line 60: Consider checking this first even when fields is not empty. If all fields in the result class are generated then any subset of those fields will also all be generated.

toregge (approved):

searchsummary/src/vespa/searchsummary/docsummary/resultclass.cpp
geirst: Fixed.

PR #24152, opened 2022-09-21
restrict vespamalloc attributes

arnej27959:
* only "name tokens' allowed

@tokle please review
@baldersheim FYI


arnej27959:
note: we probably want to restrict from "xsd:string" to NMTOKEN or similar other places in the rnc files, too.

tokle (approved):

PR #24154, opened 2022-09-21
Make it possible to limit number of log lines returned

freva:


container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
```diff
@@ -36,6 +36,8 @@ public AsyncHttpResponse handle(HttpRequest request) {
                                .map(Long::valueOf).map(Instant::ofEpochMilli).orElse(Instant.MIN);
         Instant to = Optional.ofNullable(request.getProperty("to"))
                              .map(Long::valueOf).map(Instant::ofEpochMilli).orElse(Instant.MAX);
+        long maxLines = Optional.ofNullable(request.getProperty("maxLines"))
```
jonmv, line 39: Not sure the default should be limited, though? Well, maybe, since we use this only in hosted ...

jonmv (approved):

jonmv:
We should also propagate this parameter through proxies.

container-core/src/main/java/com/yahoo/container/handler/LogHandler.java
freva: It looks like the logserver node was dying (out of memory?) even though we have streaming, so having a limit by default makes sense. But I guess the primary user of this API is console anyway, which will set the limit anyway.

freva:
Looks like configserver proxies all parameters. Created internal PR to proxy the parameter in the controller.

PR #24155, opened 2022-09-21
Reduce special handling of struct fields.

toregge:

PR #24160, opened 2022-09-21
Reduce special handling of struct fields.

toregge:
@geirst : please review

This is a rebased version of #24155 , with static member function added to SlimeFiller.


geirst (approved):
👍

PR #24161, opened 2022-09-21
Balder/short circuit boolean

baldersheim:
@bratseth PR
Only the last commit has any semantic.

### DIFF
--- a/config-model/src/test/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformerTestCase.java
+++ b/config-model/src/test/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformerTestCase.java
@@ -4,11 +4,15 @@ package com.yahoo.schema.expressiontransforms;
 import com.yahoo.searchlib.rankingexpression.RankingExpression;
 import com.yahoo.searchlib.rankingexpression.evaluation.MapContext;
 import com.yahoo.searchlib.rankingexpression.evaluation.MapTypeContext;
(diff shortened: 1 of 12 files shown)


config-model/src/test/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformerTestCase.java
```diff
@@ -33,25 +37,48 @@ public void testIt() throws Exception {
...
+    }
+
     private void assertTransformed(String expected, String input) throws Exception {
         var transformedExpression = new BooleanExpressionTransformer()
                                             .transform(new RankingExpression(input),
                                                        new TransformContext(Map.of(), new MapTypeContext()));
 
         assertEquals(new RankingExpression(expected), transformedExpression, "Transformed as expected");
 
-        MapContext context = contextWithSingleLetterVariables();
```
bratseth: Not really any point of evaluating at all without this since all expressions will just be a bunch of operations on zeros, which will all come out as zero.

bratseth (approved):
Great! But I think we should keep the evaluation testing.

config-model/src/test/java/com/yahoo/schema/expressiontransforms/BooleanExpressionTransformerTestCase.java
baldersheim: I suspected so :) I removed since it was not used.
Now it is actually used too.

PR #24164, opened 2022-09-21
on panic(err), just print the message and exit

arnej27959:
@mpolden please review

this just avoids printing a stack trace when errors are handled with a simple panic(err).
I considered wrapping it into some abstraction, what do you think?


mpolden (approved):
LGTM

PR #24182, opened 2022-09-22
Simplify state version checks by requiring exact version match [run-systemtest]

vekterli:
@geirst please review

The existing state unification logic was likely to help ensure that various distributor availability-states were treated as if they were simply Up, but the distributor has not been able to even _be_ in other available states than Up for many years. So it's effectively pointless.

Remove unification entirely and instead require both the distributor and content node to be mutually in sync with the exact cluster state version.


geirst (approved):
👍

PR #24202, opened 2022-09-24
Add patched ZK files

jonmv:
@hmusum is this right? :>

### DIFF
--- a/zookeeper-server/zookeeper-server/src/main/java/org/apache/zookeeper/server/ZooKeeperServer.java
+++ b/zookeeper-server/zookeeper-server/src/main/java/org/apache/zookeeper/server/ZooKeeperServer.java
@@ -1,0 +1,2329 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


hmusum (approved):
Looks good to me. Do you log anything that will show that this code is used and not the original one?

jonmv:
Yes, changed the startup logging a bit

hmusum:
👍

jonmv:
@hakonhall FYI

zookeeper-server/zookeeper-server/src/main/java/org/apache/zookeeper/server/ZooKeeperServer.java
```diff
@@ -0,0 +1,2329 @@
...
+        this.initialConfig = initialConfig;
+
+        this.requestPathMetricsCollector = new RequestPathMetricsCollector();
+
+        this.initLargeRequestThrottlingSettings();
+
+        this.authHelper = new AuthenticationHelper();
+
+        LOG.info(
+            "Created patched server with"
```
jonmv, line 363: Tjis is the log line to look for

PR #24210, opened 2022-09-26
Arnej/common backticks

arnej27959:
@mpolden please review


arnej27959:
note: this is mostly useful when rewriting shell scripts that use
```
`foo 2>/dev/null`
`bar 2>&1`
`baz`
```
and we want or need to preserve the exact behavior.

mpolden (approved):
👍

PR #24221, opened 2022-09-26
Add feature flag for enabling CSRF token in console

bjorncs:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -542,6 +542,13 @@ public class Flags {
             "Takes effect at next tick",
             APPLICATION_ID);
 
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -542,6 +542,13 @@ public class Flags {
             "Takes effect at next tick",
             APPLICATION_ID);
 
+    public static final UnboundBooleanFlag CONSOLE_CSRF = defineFeatureFlag(
+            "console-csrf", false,
+            List.of("bjorncs", "tokle"), "2022-09-26", "2023-06-01",
+            "Enable CSRF token in console",
+            "Takes effect immediately",
```
freva, line 549: The flags are fetched when the user first loads the console

freva (approved):

PR #24225, opened 2022-09-26
Remove StateManager handling of deprecated Initializing state [run-systemtest]

vekterli:
@geirst please review

Neither distributors nor content nodes ever report their state as Initializing as part of their startup sequence; they go straight
from Down to Up. Remove complicated init progress delta reporting that is no longer needed.


geirst (approved):
👍

baldersheim (approved):

PR #24241, opened 2022-09-28
Add feature flag for setting up wireguard on configservers.

gjoranv:


andreer:
After looking at this, I think it's actually a good idea to have two feature flags - one for config servers (like here) and one for tenant hosts (which may have different dimensions, i.e. tenant). Go ahead and just merge this and we can look at the config server case first.

andreer (approved):

PR #24261, opened 2022-09-29
No longer including jackson-dataformat-xml, woodstox-core, stax2-api

baldersheim:
@gjoranv or @bjorncs PR


baldersheim:
Required for #24259

PR #24279, opened 2022-10-02
Jonmv/revert gc heaven commits

jonmv:
@bratseth config servers in CD have been struggling with deployments, due to really heavy GC, after this was merged. I can't see any other possible explanations, so I think we should revert and investigate.


jonmv:
@baldersheim @aressem

baldersheim (approved):

PR #24308, opened 2022-10-05
Use latest version of jackson-databind.

gjoranv:
FYI: @baldersheim, @bjorncs


gjoranv:
Looking into the test failure.

PR #24325, opened 2022-10-06
Add support for different output formats in logfmt

ean:
Support vespa (default, json and raw lines

### DIFF
--- a/client/go/cmd/logfmt/handleline.go
+++ b/client/go/cmd/logfmt/handleline.go
@@ -5,58 +5,122 @@
 package logfmt
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
 	"time"
 )
 
-// handle a line in "vespa.log" format; do filtering and formatting as specified in opts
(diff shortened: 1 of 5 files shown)


client/go/cmd/logfmt/handleline.go
```diff
@@ -5,58 +5,120 @@
...
+	nsecs := int64(secs * 1e9)
+	return time.Unix(0, nsecs), nil
+}
+
+func handleLineJson(_ *Options, fields *logFields) (string, error) {
+	timestamp, err := parseTimestamp(fields.timestamp)
+	if err != nil {
+		return "", err
+	}
+	outputFields := struct {
```
arnej27959, line 90: I would prefer this struct to be defined (with a name) outside the function, but that's just a minor style thing.

arnej27959 (dismissed):
looks basically good to me, some minor things to fix or consider.
Maybe the JSON should include the timestamp in its original format also? (to allow easy arithmetic)

arnej27959 (approved):
looks good to me

PR #24327, opened 2022-10-06
GC forward-issues-as-errors flag

baldersheim:
@arnej27959 PR


arnej27959 (commented):
are we sure we want to do this? it's a permanent flag for a reason: when we start reporting a new issue from the backend, some tenants may require overriding the feature flag until they fix whatever triggers the issue.

PR #24330, opened 2022-10-06
Handle internal server error in response from config server as transi… [run-systemtest]

hmusum:
…ent error


freva (commented):
`HttpHandler`/`SessionHandler`/`ApplicationApiHandler` in `configserver` should handle the internal server errors with `com.yahoo.vespa.hosted.controller.restapi.ErrorResponses::logThrowing` or we should drop the `e.getMessage()` here since it will have the entire error from node-repository?

freva (approved):

PR #24335, opened 2022-10-06
Support getting SSLContext for a peer with a public CA cert

hakonhall:


hakonhall:
Superceeded by #24397 .

PR #24338, opened 2022-10-06
much simpler CharSequenceNormalizer

arnej27959:
do we really need to run all these special-purpose normalizers that all use expensive regexp match/replace?
my opinion is that if we really need some of these, we can reimplement them in a more efficient manner.
but most of these look very out of place, and at least the Emoji thingy looks just wrong.

@jonmv and @bratseth : please discuss...


bratseth (approved):
This looks perfectly fine to me, JonM can speak to why they were added in the first place.

bratseth:
Let's try it then!

PR #24344, opened 2022-10-06
Return X-Vespa-Ignored-Fields if fields were ignored

bratseth:
I wasn't able to separate out the functional changes, sorry.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
@@ -274,15 +274,11 @@ public class StateHandler extends AbstractRequestHandler {
     }
 
     private static List<Tuple> collapseMetrics(MetricSnapshot snapshot, String consumer) {
-        switch (consumer) {
-            case HEALTH_PATH:
-                return collapseHealthMetrics(snapshot);
-            case "all": // deprecated name
-            case METRICS_PATH:
-                return flattenAllMetrics(snapshot);
-            default:
-                throw new IllegalArgumentException("Unknown consumer '" + consumer + "'.");
-        }
+        return switch (consumer) {
+            case HEALTH_PATH -> collapseHealthMetrics(snapshot); // deprecated name
+            case "all", METRICS_PATH -> flattenAllMetrics(snapshot);
+            default -> throw new IllegalArgumentException("Unknown consumer '" + consumer + "'.");
+        };
     }
 
     private static List<Tuple> collapseHealthMetrics(MetricSnapshot snapshot) {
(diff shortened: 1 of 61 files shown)


bratseth:
The point of this is in DocumentV1ApiHandler.commit(..), 
and is mostly done in StructReader and VespaJsonDocumentReader.

document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
```diff
@@ -1006,17 +1009,65 @@ public void nonExistingFieldCausesException()  throws IOException{
...
+//                "    'nonexisting2': { " +
+//                "      'assign': {",
+//                "        'blocks': {",
+//                "          'a':[2.0,3.0],",
+//                "          'b':[4.0,5.0]",
+//                "        }",
+//                "      }",
+//                "    },",
+                "    'field2': { 'assign': 'value2' }",
+//                "    'nonexisting3': { 'assign': 'ignored value' }",
```
jonmv, line 1058: Intentionally commented out?

jonmv (approved):
I think maybe that one unit test wasn't done? Otherwise LGTM!

document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
bratseth: No, thanks!

document/src/main/java/com/yahoo/document/json/readers/VespaJsonDocumentReader.java
```diff
@@ -185,6 +191,7 @@ private void addFieldPathUpdates(DocumentUpdate update, TokenBuffer buffer, Stri
             update.addFieldPathUpdate(fieldPathUpdate);
             buffer.next();
         }
+        return true; // TODO: Track fullyApplied for fieldPath updates
```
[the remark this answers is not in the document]
bratseth: Yes.

container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
```diff
@@ -274,15 +274,11 @@ private ObjectNode buildJsonForSnapshot(String consumer, MetricSnapshot metricSn
...
-            case HEALTH_PATH:
-                return collapseHealthMetrics(snapshot);
-            case "all": // deprecated name
-            case METRICS_PATH:
-                return flattenAllMetrics(snapshot);
-            default:
-                throw new IllegalArgumentException("Unknown consumer '" + consumer + "'.");
-        }
+        return switch (consumer) {
+            case HEALTH_PATH -> collapseHealthMetrics(snapshot); // deprecated name
```
hmusum, line 278: I think this one got misplaces, should be on "all". And maybe deprecate and add note about removing in Vespa 9?

bratseth: Yes, added a better comment in https://github.com/vespa-engine/vespa/pull/24354

PR #24348, opened 2022-10-07
JSON output format is not stabilized yet, and will likely change in t…

ean:
…he future.

@arnej27959 - we will likely need to change the JSON object somewhat to ensure it maps well to both default managed kubernetes setups in both AWS and GCP.


arnej27959 (approved):
👍

PR #24357, opened 2022-10-07
Jonmv/revert ignored fields header

jonmv:
@bratseth


jonmv:
https://github.com/vespa-engine/vespa/pull/24358 instead

PR #24375, opened 2022-10-10
Add a test using implicit default zone

bratseth:
Looks to me like we didn't have a test for this any more.


jonmv (commented):
Do you mean "default instance"? See `DeploymentSpecWithoutInstanceTest`.

bratseth:
Fine, no need for this then!

PR #24387, opened 2022-10-11
Add utilities for secure one-way single-use key exchange tokens using ECIES

vekterli:
@bjorncs please review. Note: we should not actually use this until we're on a BouncyCastle version that can enforce good algorithms as part of the ECIES cipher name (SHA-256 vs. SHA-1 for digests etc). Also, ideally this should use X25519 curves, but have not gotten this to work with BC's ECIES yet.

Lets a sender generate a random, single-use symmetric key and securely share this with a receiver, with the sender only knowing the public key of the receiver. The shared key is exchanged via an opaque token that can only be decoded by having the private key corresponding to the public key used for encoding it.

This is implemented using ECIES, a hybrid encryption scheme using Elliptic Curve Diffie-Hellman (ECDH) for ephemeral key exchange combined with a symmetric cipher using the ephemeral key for actual plaintext encryption/decryption.

In addition to the key exchange itself, utilities for creating encryption and decryption ciphers for AES-GCM-256 from the shared keys are included.

**Security note**: since the key is intended to be used for producing a single piece of ciphertext, a fixed Initialization Vector (IV) is used. The key MUST NOT be used to produce more than one ciphertext, lest the entire security model breaks down entirely.

### DIFF
--- a/security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
+++ b/security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
@@ -1,0 +1,118 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security;
+
+import org.bouncycastle.jcajce.provider.util.BadBlockException;
+
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.KeyPair;
+import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
+import java.security.SecureRandom;
+
+/**
+ * Implements both the sender and receiver sides of a secure, anonymous one-way
+ * key generation and exchange protocol implemented using ECIES; a hybrid crypto
+ * scheme built around elliptic curves.
+ *
+ * A shared key, once generated, may have its sealed component sent over a public
+ * channel without revealing anything about the underlying secret key. Only a
+ * recipient holding the private key corresponding to the public used for shared
+ * key creation may derive the same secret key as the sender.
+ *
+ * Every generated key is globally unique (with extremely high probability).
+ *
+ * The secret key is intended to be used <em>only once</em>. It MUST NOT be used to
+ * produce more than a single ciphertext. Using the secret key to produce multiple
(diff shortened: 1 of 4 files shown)


security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+public class SharedKeyGenerator {
+
+    private static final int    AES_GCM_AUTH_TAG_BITS = 128;
+    private static final String AES_GCM_ALGO_SPEC     = "AES/GCM/NoPadding";
+    private static final String ECIES_CIPHER_NAME     = "ECIES"; // TODO ensure SHA-256+AES. Needs BC version bump
+
+    public static SecretSharedKey generateForReceiverPublicKey(PublicKey receiverPublicKey, int keyId) {
+        try {
+            var keyGen = KeyGenerator.getInstance("AES");
+            keyGen.init(256, new SecureRandom());
```
bjorncs, line 46: Consider having `SecureRandom` as a member (might be costly to instantiate per request). You may utilize a deterministic random generator in unit tests by providing `RandomGenerator` instance as constructor parameter and injecting a `Random` with fixed seed in test cases.

security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
```diff
@@ -0,0 +1,118 @@
...
+
+            var sealedSharedKey = new SealedSharedKey(keyId, eciesPayload);
+            return new SecretSharedKey(secretKey, sealedSharedKey);
+        } catch (NoSuchAlgorithmException | InvalidKeyException | NoSuchPaddingException
+                | IllegalBlockSizeException | BadPaddingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static SecretSharedKey fromSealedKey(SealedSharedKey sealedKey, KeyPair receiverKeyPair) {
```
bjorncs, line 61: Only private key required

bjorncs (approved):

security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
bjorncs: Alternatively as static field if deterministic unit tests are not required. `SecureRandom` is thread-safe.

security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
vekterli: Fixed

security-utils/src/main/java/com/yahoo/security/SharedKeyGenerator.java
vekterli: Fixed; moved to static field since we don't test the generated key material bytes explicitly

bjorncs (approved):

PR #24394, opened 2022-10-11
Add root level aliases to parameter options

xgboosting:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ Testing
 install_manifest.txt
 *.cbp
 !/.copr/Makefile
+node_modules/
--- a/client/js/app/src/app/pages/querybuilder/context/parameters.jsx
+++ b/client/js/app/src/app/pages/querybuilder/context/parameters.jsx
@@ -24,6 +24,21 @@ export default {
   traceLevel: { name: 'traceLevel', type: 'Integer' },
   explainLevel: { name: 'explainLevel', type: 'Integer' },
   explainlevel: { name: 'explainlevel', type: 'Integer' },
+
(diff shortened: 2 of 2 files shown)


kkraune:
Thanks for submitting! Adding proper support for aliases is a bit complicated. In _some_ cases, the alias name does not collide with the parameter name (e.g. tracelevel / traceLevel), in others it does, like ranking. the tracelevel is already confusing with two entries.

Supporting both the structured _ranking_ and _ranking_ as an alias for ranking.profile (string) complicates things

After discussing with @freva we think a better approach is _not_ supporting aliases. The QueryBuilder is after all a way to build a query using an interface with structure. the ranking parameter is a good example if this.

We can link to https://docs.vespa.ai/en/reference/query-api-reference.html where the aliases are documented. Other ideas for how to improve is very welcome, @xgboosting and we appreciate your feedback

xgboosting:
> Thanks for submitting! Adding proper support for aliases is a bit complicated. In _some_ cases, the alias name does not collide with the parameter name (e.g. tracelevel / traceLevel), in others it does, like ranking. the tracelevel is already confusing with two entries.
> 
> Supporting both the structured _ranking_ and _ranking_ as an alias for ranking.profile (string) complicates things
> 
> After discussing with @freva we think a better approach is _not_ supporting aliases. The QueryBuilder is after all a way to build a query using an interface with structure. the ranking parameter is a good example of this.
> 
> We can link to https://docs.vespa.ai/en/reference/query-api-reference.html where the aliases are documented. Other ideas for how to improve is very welcome, @xgboosting and we appreciate your feedback

Sounds good, I understand.

PR #24395, opened 2022-10-11
mapped lookup

havardpe:
- change how stride works with gen specs to allow better control over how mapped tensors overlap (avoid always overlapping via label "0")

- extend eval fixture verify functionality by adding a function that takes param specs directly to verify a single specific case.

@geirst please review
@arnej27959 FYI


geirst (approved):
👍

PR #24399, opened 2022-10-12
- Use a common scratchpad for serializing the different parts of the …

baldersheim:
…query.

- Use a threadlocal for the scratchpad. This avoids costly resizing, or initialiing too large buffer for every query. Using a thread local is fine now that we limit the number of search threads to a reasonable number = #cores * 2.

@bratseth PR


bratseth (approved):
2x cores is way too little in general, but even worst case I think a 4k buffer per thread is ok.

PR #24406, opened 2022-10-12
Add instance tags [run-systemtest]

bratseth:
@jonmv please review, but we can hold off merge until after tomorrows arch meeting

### DIFF
--- a/config-provisioning/src/test/java/com/yahoo/config/provision/TagsTest.java
+++ b/config-provisioning/src/test/java/com/yahoo/config/provision/TagsTest.java
@@ -20,6 +20,11 @@ public class TagsTest {
         assertEquals(Tags.empty(), Tags.fromString(""));
         assertEquals(Tags.empty(), Tags.fromString(" "));
     }
+    
+    @Test
+    public void testDeserialization() {
+        assertEquals(new Tags(Set.of("tag1", "tag2")), Tags.fromString("  tag1     tag2  "));
+    }
 
     @Test
     public void testSerialization() {
--- a/config-provisioning/src/test/java/com/yahoo/config/provision/TagsTest.java
+++ b/config-provisioning/src/test/java/com/yahoo/config/provision/TagsTest.java
@@ -1,0 +1,52 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.config.provision;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * @author bratseth
+ */
+public class TagsTest {
+
+    @Test
+    public void testEmpty() {
+        assertEquals(Tags.empty(), Tags.fromString(null));
+        assertEquals(Tags.empty(), Tags.fromString(""));
+        assertEquals(Tags.empty(), Tags.fromString(" "));
+    }
+
+    @Test
+    public void testSerialization() {
+        Tags tags = new Tags(Set.of("a", "tag2", "3"));
+        assertEquals(tags, Tags.fromString(tags.toString())); // Required by automatic serialization
+        assertEquals(tags, Tags.fromString(tags.asString())); // Required by automatic serialization
+    }
+
+    @Test
+    public void testContains() {
+        Tags tags = new Tags(Set.of("a", "tag2", "3"));
+        assertTrue(tags.contains("a"));
+        assertTrue(tags.contains("tag2"));
+        assertTrue(tags.contains("3"));
+        assertFalse(tags.contains("other"));
+
+        Tags subTags = new Tags(Set.of("a", "3"));
+        assertTrue(tags.containsAll(subTags));
+        assertFalse(subTags.containsAll(tags));
+    }
+
+    @Test
+    public void testIntersects() {
+        Tags tags1 = new Tags(Set.of("a", "tag2", "3"));
+        Tags tags2 = new Tags(Set.of("a", "tag3"));
+        assertTrue(tags1.intersects(tags2));
+        assertTrue(tags2.intersects(tags1));
+    }
+
+}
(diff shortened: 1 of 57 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
```diff
@@ -202,7 +203,8 @@ private void preProcessAndPopulateCache() {
                                     new InputStreamReader(new ByteArrayInputStream(servicesXml.content()), UTF_8),
                                     InstanceName.defaultName(),
                                     Environment.prod,
-                                    RegionName.defaultName())
+                                    RegionName.defaultName(),
+                                    Tags.empty())
```
jonmv, line 207: Urgh, I guess this should preprocess with all configurations that may eventually be deployed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2035,7 +2036,7 @@ private HttpResponse createInstance(String tenantName, String applicationName, S
         if (controller.applications().getApplication(applicationId).isEmpty())
             createApplication(tenantName, applicationName, request);
 
-        controller.applications().createInstance(applicationId.instance(instanceName));
+        controller.applications().createInstance(applicationId.instance(instanceName), Tags.empty());
```
jonmv, line 2039: Hmm ... supporting `dev/perf` here isn't trivial, and perhaps not that useful?

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
```diff
@@ -273,6 +306,7 @@ public static PrepareParams fromHttpRequest(HttpRequest request, TenantName tena
                             .verbose(request.getBooleanProperty(VERBOSE_PARAM_NAME))
                             .timeoutBudget(SessionHandler.getTimeoutBudget(request, barrierTimeout))
                             .applicationId(createApplicationId(request, tenant))
+                            .tags(Tags.fromString(request.getProperty(TAGS_PARAM_NAME)))
```
jonmv, line 309: This PR doesn't set this value.

jonmv (approved):
Great! So much wiring 😝

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
bratseth: I don't know what this is for, hopefully it is just wasted work if there are variants?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
bratseth: I assumed not.

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
bratseth: How do you figure that? Afaik this payload is autogenerated from com.yahoo.vespa.hosted.controller.api.application.v4.model.DeploymentData, which sets tags on line 655 in ApplicationController.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
jonmv: It is for storing all relevant deployment meta data to our audit store. If a certain override includes a file which isn't included in the default path taken here, that file will not be stored. It's not critical. Perhaps we could instead run only the include processor, and not the override one? Would that include all possible variants? (We don't care about the generated XML, but all files read by the inclusion process will be loaded to the file cache, which is then flushed to the audit storage.)

jonmv: You don't need to fix this in this PR.

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
jonmv: The `DeploymentData` has the tags, but they're not read anywhere. The implementation of the `ConfigServer` needs to do that.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackage.java
jonmv: Ah, never mind, the include processor is run first.

configserver/src/main/java/com/yahoo/vespa/config/server/session/PrepareParams.java
jonmv: Sorry, this wasn't the relevant one. The `fromJson` is it. I'll make a PR for you :)

bratseth: The fromJson does reads the tags already?

jonmv: Yes, but they're not written anywhere.

config-provisioning/src/test/java/com/yahoo/config/provision/TagsTest.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author bratseth
+ */
+public class TagsTest {
+
+    @Test
+    public void testEmpty() {
+        assertEquals(Tags.empty(), Tags.fromString(null));
+        assertEquals(Tags.empty(), Tags.fromString(""));
+        assertEquals(Tags.empty(), Tags.fromString(" "));
+    }
```
jonmv, line 22: How about this?

```java
        assertEquals(new Tags(Set.of("tag1", "tag2")), Tags.fromString("  tag1     tag2  "));
```

jonmv: This currently fails with duplicate element (`""`).

jonmv:
```suggestion

    @Test
    public void testDeserialization() {
        assertEquals(new Tags(Set.of("tag1", "tag2")), Tags.fromString("  tag1     tag2  "));
    }
```

bratseth:
Thanks!

jonmv:
Also need a stable ordering for tests ...

PR #24408, opened 2022-10-12
Do not retry invalid applications (for the next 100 years)

jonmv:
@freva

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/HttpHandler.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/HttpHandler.java
@@ -51,7 +51,8 @@ public class HttpHandler extends ThreadedHttpRequestHandler {
         } catch (IllegalArgumentException | UnsupportedOperationException e) {
             return HttpErrorResponse.badRequest(getMessage(e, request));
         } catch (NodeAllocationException e) {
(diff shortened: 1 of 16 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/HttpHandler.java
```diff
@@ -51,7 +51,8 @@ public HttpResponse handle(HttpRequest request) {
         } catch (IllegalArgumentException | UnsupportedOperationException e) {
             return HttpErrorResponse.badRequest(getMessage(e, request));
         } catch (NodeAllocationException e) {
-            return HttpErrorResponse.nodeAllocationFailure(getMessage(e, request));
+            return e.retryable() ? HttpErrorResponse.nodeAllocationFailure(getMessage(e, request))
```
freva, line 54: Why not throw `InvalidApplicationException` instead?

freva (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/http/HttpHandler.java
jonmv: The NodeAllicationException, that is.

PR #24418, opened 2022-10-13
Implement RFC-5869 HMAC-based Extract-and-Expand Key Derivation Function (HKDF)

vekterli:
@bjorncs please review.

The HKDF is initialized ("extracted") from a (non-secret) salt and a secret key. From this, any number of secret keys can be derived ("expanded") deterministically.

When multiple keys are to be derived from the same initial keying/salting material, each separate key should use a distinct "context". This ensures that there exists a domain separation between the keys. Using the same context as another key on a HKDF initialized with the same salt+key results in the exact same derived key material as that key.

This implementation only offers HMAC-SHA256-based key derivation.

Tested with all HMAC-SHA256 test vectors in RFC-5869, with added edge case tests.

Analogous to BouncyCastle's `HKDFBytesGenerator`, but with a simpler API that tries to be very explicit in its operation, as well as fully thread safe due to not storing intermediate calculations in member fields.

### DIFF
--- a/security-utils/src/main/java/com/yahoo/security/HKDF.java
+++ b/security-utils/src/main/java/com/yahoo/security/HKDF.java
@@ -1,0 +1,221 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security;
+
+import javax.crypto.Mac;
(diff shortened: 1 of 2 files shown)


bjorncs (approved):
Great work! 👍 🎖️

PR #24423, opened 2022-10-13
unify return value handling using universal references

havardpe:
the requires clause is there to make sure we are only allowed to return values that are implicitly convertible to T; this will avoid calling explicit constructors 'implicitly' (as seem from the co_return statement).

(the internal optional requires assignability to also cover the case where the value is not constructed, but overwritten, which probably imposes a similar restriction, but if we stop using optional, the require becomes more useful)

@vekterli please review


vekterli (approved):
:+1:

PR #24434, opened 2022-10-13
Remove HostLivenessTracker

freva:
Nodes are never in state `ready` since we migrated to dynamic allocation. Hosts are only in state `ready` for less than a minute and do not request config.


bratseth (approved):
Great!

PR #24440, opened 2022-10-13
Add fallback variant when std::jthread is not available.

toregge:
@havardpe : please review


havardpe (commented):
This removes the purpose of using jthread in the first place (less code). If there is no jthread, we are better off always using thread.

PR #24441, opened 2022-10-14
Keep last 1000 lines sorted in log reader

jonmv:
@freva please review and mereg.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/handler/LogReader.java
+++ b/container-core/src/main/java/com/yahoo/container/handler/LogReader.java
@@ -29,6 +29,7 @@ import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
+import java.util.PriorityQueue;
 import java.util.TreeMap;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


jonmv:
Or @hmusum :)

freva (approved):
:1st_place_medal:

hmusum (approved):

jonmv:
This'll unfortunately only solve minor out-of-order issues. 
The real fix would be a sorted write-cache in whatever code receives log entries and writes them to disk. 
When receiving a log entry which was older than the oldest in the cache, it would have to flush the cache, and start a new file, before adding that entry to the (now empty) cache. 
This would work fine with the merge sort we do over all log files per hour, in the log reader code. (The log writer already produces multiple such files, with overlapping time intervals, occasionally.)

So, the problem is I never found the code that writes these log entires. Is it C++? @arnej27959 you probably know :)

arnej27959:
Is this for reading from log archive? It’s written by the logserver.

jonmv:
Yes, I was pointed in the right direction by @hmusum :)

PR #24451, opened 2022-10-14
Jonmv/configurable zk session timeout

jonmv:
@hakonhall please review, but don't merge.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
@@ -89,7 +89,7 @@ public final class ApplicationContainerCluster extends ContainerCluster<Applicat
 
     private MbusParams mbusParams;
     private boolean messageBusEnabled = true;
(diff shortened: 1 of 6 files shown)


hakonhall (approved):

hakonhall (approved):

jonmv:
@bratseth perhaps something like this, to allow users to set that session timeout themselves?

config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
```diff
@@ -305,6 +307,15 @@ public void getConfig(ZookeeperServerConfig.Builder builder) {
         }
     }
 
+    @Override
+    public void getConfig(CuratorConfig.Builder builder) {
+        if (getParent() instanceof ConfigserverCluster) return; // Produces its own config
```
bratseth, line 312:
```suggestion
```

bratseth (changes requested):
LGTM (but don't make getConfig a noop for configservers).

hakonhall (approved):

PR #24452, opened 2022-10-14
Skip reclaim_labels on original buffer when it has been copied as par…

toregge:
…t of compaction or fallback copy (due to datastore buffer fallback resize).

@geirst : please review


geirst (approved):
👍

PR #24456, opened 2022-10-14
Stop bypassing hold list for btree nodes that were never reachable fo…

toregge:
…r readers.

@geirst : please review


geirst (approved):
👍

PR #24458, opened 2022-10-14
Support locks with TTL in ZK curator

bjorncs:

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
@@ -354,13 +354,16 @@ public class Curator extends AbstractComponent implements AutoCloseable {
         }
     }
 
-    /** Create and acquire a re-entrant lock in given path */
-    public Lock lock(Path path, Duration timeout) {
-        create(path);
+    /** Create and acquire a re-entrant lock in given path with a TTL */
+    public Lock lock(Path path, Duration timeout, Duration ttl) {
+        create(path, ttl);
         Lock lock = locks.computeIfAbsent(path, (pathArg) -> new Lock(pathArg.getAbsolute(), this));
         lock.acquire(timeout);
         return lock;
     }
+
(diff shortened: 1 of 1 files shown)


bjorncs:
FYI @tokle

hmusum (approved):

zkfacade/src/main/java/com/yahoo/vespa/curator/Curator.java
```diff
@@ -354,14 +354,17 @@ public Optional<Stat> getStat(Path path) {
         }
     }
 
-    /** Create and acquire a re-entrant lock in given path */
-    public Lock lock(Path path, Duration timeout) {
-        create(path);
+    /** Create and acquire a re-entrant lock in given path with a TTL */
```
jonmv, line 357: What happens when TTL passes?

bjorncs: The node is deleted

jonmv: Yes, but what happens with the code that still _thinks_ it's holding the lock? There's no callback to notify of expiry.

jonmv: Locks could also expiry without TTL, due to session death. We don't handle that case either, but it's uncommon. 
Ideally, all writes should contain the lock as an argument, and be a transaction where the lock is verified somehow, together with the write operation. I think this is possible to implement, but I'm not convinced it's worth it for the case we already have, without TTL. 
Depending on how TTL for locks is intended used, however, we may have to consider this ... no?

bjorncs: I assume the TTL applies from last time the locked was taken.

bjorncs: Correct

PR #24477, opened 2022-10-17
- Always check if enough space to write.

baldersheim:
- Split large initSortData method.
- make buffer resizing explicit and readable.
- Make a better initial estimate for buffer size.

@geirst and @toregge PR


toregge (approved):

geirst (approved):
👍

PR #24497, opened 2022-10-18
Balder/precompute hashcode as it will always be used atleast once

baldersheim:
@bjorncs PR

### DIFF
--- a/container-core/src/main/java/com/yahoo/metrics/simple/Point.java
+++ b/container-core/src/main/java/com/yahoo/metrics/simple/Point.java
@@ -6,7 +6,6 @@ import java.util.List;
 import java.util.Map;
 
 import com.yahoo.api.annotations.Beta;
-import com.google.common.collect.ImmutableList;
 import com.yahoo.collections.Tuple2;
 import com.yahoo.jdisc.Metric.Context;
 
@@ -88,14 +87,14 @@ public final class Point implements Context {
      * Get an immutable list view of the values for each dimension.
      */
     public List<Value> location() {
-        return ImmutableList.copyOf(location);
+        return List.of(location);
     }
 
     /**
      * Get an immutable list view of the names of each dimension.
      */
     public List<String> dimensions() {
-        return ImmutableList.copyOf(dimensions);
+        return List.of(dimensions);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


container-core/src/main/java/com/yahoo/metrics/simple/PointBuilder.java
```diff
@@ -27,18 +27,12 @@ public enum Discriminator {
     }
 
     PointBuilder(Point p) {
-        dimensions = new ArrayList<>();
-        location = new ArrayList<>();
         if (p != null) {
             int size = p.dimensionality();
             dimensions = new ArrayList<>(size+2);
```
bjorncs, line 32: Consider rewriting as `new ArrayList<>(p.getDimensions())` if `size+2` matches length of `getDimensions()` output.

container-core/src/main/java/com/yahoo/metrics/simple/Point.java
```diff
@@ -58,32 +60,16 @@ private static Tuple2<String[], Value[]> buildParameters(Map<String, ?> properti
...
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + Arrays.hashCode(dimensions);
-        result = prime * result + Arrays.hashCode(location);
-        return result;
+        return hashCode;
     }
```
bjorncs, line 74: Consider using `List.of()` for immutable view for return value of `location()` and `dimensions()`. This may lead to a zero-copy optimization downstream.

bjorncs (approved):

container-core/src/main/java/com/yahoo/metrics/simple/Point.java
baldersheim: Done

container-core/src/main/java/com/yahoo/metrics/simple/PointBuilder.java
baldersheim: That will never happen. p.getDimensions().size() == size == p.dimensionality().
I think intention is add room for 2 more than there were in the original Point.

bjorncs (approved):

PR #24499, opened 2022-10-18
- Unify on the faster JsonRender. The default Value.toString has a ve…

baldersheim:
…ry high initial cost as it is intended for large json structures.

@arnej27959 or @bratseth PR


bratseth (approved):
👍

PR #24501, opened 2022-10-19
Use java immutable collections instead of guava

baldersheim:
@bjorncs PR


bjorncs (approved):

bjorncs (approved):

baldersheim:
This must be done more carefully.
com.google.collections.ImmutableSet/Map preserves order, while Set.copyOf/Map.copyOf does not.

PR #24510, opened 2022-10-19
Remove use of document::UrlDataType from backend

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
Great 👍  This code has never been used in the production pipeline.

PR #24511, opened 2022-10-19
Balder/reuse already acquired context

baldersheim:
@bjorncs PR


bjorncs (approved):

bjorncs:
Another approach to re-using metric context per request is storing it as attribute/context on request object.

PR #24518, opened 2022-10-20
support models without tokenTypeIds (like DistilBERT)

cjolif:
This is similar to what is done in sbert allowing to use models that don't support tokenTypeIds. This is done by configuring the embedder with empty tokenTypeIdsName. In addition to the "unit" test in the code I tested it with https://huggingface.co/sentence-transformers/distiluse-base-multilingual-cased-v1.


bratseth (approved):
Great!

PR #24537, opened 2022-10-22
Replace the global variable vespalib::test::local_spec with the function

toregge:
vespalib::test::make_local_spec().

@vekterli : please review
@havardpe : FYI


vekterli (approved):
:+1:

PR #24545, opened 2022-10-24
Move related non-local variables to same translation unit to ensure

toregge:
ordered dynamic initialization.

@vekterli : please review


vekterli (approved):
:+1:

PR #24546, opened 2022-10-24
Write provisioned hosts to ZK early

freva:

For certain cloud providers we may get enough information to be able to create `Node`s representing the hosts and write them to ZK early and then wait for the rest of the provision request to finish. This PR ensures that we this to ZK as soon as possible, so that if the config server is suddenly stopped, we wont lose track of these hosts.


mpolden (approved):
👍

PR #24549, opened 2022-10-24
add go code wih API similar to defaults.cpp

arnej27959:
@mpolden or @baldersheim please review

### DIFF
--- a/client/go/defaults/defaults.go
+++ b/client/go/defaults/defaults.go
@@ -37,10 +37,8 @@ const (
 	ENV_WEB_SERVICE_PORT     = "VESPA_WEB_SERVICE_PORT"
 )
 
-/**
- * Compute the path prefix where Vespa files will live;
- * note: does not end with "/"
- **/
+// Compute the path prefix where Vespa files will live.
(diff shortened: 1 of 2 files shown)


client/go/defaults/defaults.go
```diff
@@ -0,0 +1,251 @@
...
+	ENV_CONFIGSERVER_RPC_PORT    = "port_configserver_rpc"
+
+	CONFIGPROXY_RPC_PORT_OFFSET = 90
+	ENV_CONFIGPROXY_RPC_PORT    = "port_configproxy_rpc"
+
+	DEFAULT_WEB_SERVICE_PORT = 8080
+	ENV_WEB_SERVICE_PORT     = "VESPA_WEB_SERVICE_PORT"
+)
+
+/**
```
mpolden, line 40: This looks like javadoc syntax. Should be simply 
```
/*
foo 
bar 
*/
```

https://go.dev/doc/comment

mpolden (approved):

baldersheim (approved):

PR #24555, opened 2022-10-24
Add email verification

olaaun:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
@@ -255,7 +255,10 @@ enum PathGroup {
 
     /** Paths used to approve requests to access tenant resources */
     accessRequestApproval(Matcher.tenant, "/application/v4/tenant/{tenant}/access/approve/operator",
-            "/application/v4/tenant/{tenant}/access/managed/operator");
+            "/application/v4/tenant/{tenant}/access/managed/operator"),
+
+    /** Path used for email verification */
+    emailVerification("/user/v1/email-verification");
 
     final List<String> pathSpecs;
     final List<Matcher> matchers;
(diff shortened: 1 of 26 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/role/PathGroup.java
```diff
@@ -255,7 +256,10 @@ enum PathGroup {
 
     /** Paths used to approve requests to access tenant resources */
     accessRequestApproval(Matcher.tenant, "/application/v4/tenant/{tenant}/access/approve/operator",
-            "/application/v4/tenant/{tenant}/access/managed/operator");
+            "/application/v4/tenant/{tenant}/access/managed/operator"),
+
+    /** Path used for email verification */
+    emailVerification("/user/v1/email-verification");
```
oyving, line 262: `/user/v1/email/verify` or maybe even just `/user/v1/verify`?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/MailVerifier.java
```diff
@@ -0,0 +1,114 @@
...
+        this.curatorDb = curatorDb;
+        this.clock = clock;
+    }
+
+    public PendingMailVerification sendMailVerification(TenantName tenantName, String email, PendingMailVerification.MailType mailType) {
+        var verificationCode = UUID.randomUUID().toString();
+        var verificationDeadline = clock.instant().plus(VERIFICATION_DEADLINE);
+        var pendingMailVerification = new PendingMailVerification(tenantName, email, verificationCode, verificationDeadline, mailType);
+        mailer.sendVerificationMail(pendingMailVerification);
+        writePendingVerification(pendingMailVerification);
```
oyving, line 42: Maybe it's better to write the pending verification before we send the email? That way we won't send something that we then fail to write.

oyving (changes requested):

oyving (approved):

PR #24564, opened 2022-10-25
QueryBuilder: support ranking.features.featurename

freva:
* Removes `traceLevel`, `explainLevel` and `explainlevel` aliases
* Adds 
  * `model.{filter,locale}`
  * `ranking.{rerankCount,matching}`
  * `grouping.{defaultMaxGroups,defaultMaxHits,globalMaxGroups,defaultPrecisionFactor}`
  * `presentation.summary`
  * `trace.{level,explainLevel,profileDepth,query}`
  * `dispatch.topKProbability`
  * `hitcountestimate`
  * `weakAnd.replace`
  * `wand.hits`
* Replaces `presentation.format` with `presentation.format.tensors`
* Query builder now has all parameters documented in https://docs.vespa.ai/en/reference/query-api-reference.html except:
  * `noCache` and `streaming.{userid,groupname,selection,priority,maxbucketspervisitor}`: supported by query builder but are not documented (on that page)
  * `select`: Seems like there are 2 types of select and at least 1 of them is covered by `yql`?
* Allows user defined key-values under `ranking.features` (Fixes #24475)

### DIFF
--- a/client/js/app/src/app/pages/querybuilder/context/__test__/query-builder-provider.test.jsx
+++ b/client/js/app/src/app/pages/querybuilder/context/__test__/query-builder-provider.test.jsx
@@ -52,11 +52,20 @@ test('manipulates inputs', () => {
     [ACTION.INPUT_ADD, { id: '2', type: 'location' }],
     [ACTION.INPUT_ADD, { id: '2', type: 'matchPhase' }],
     [ACTION.INPUT_UPDATE, { id: '2.0', value: 'us' }],
+    [ACTION.INPUT_ADD, { id: '2', type: 'features' }],
+    [ACTION.INPUT_ADD, { id: '2.2', type: '' }],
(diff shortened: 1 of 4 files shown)


kkraune (approved):
Great! will try it out

kkraune:
> select: Seems like there are 2 types of select and at least 1 of them is covered by yql?

yes. See https://github.com/vespa-engine/documentation/pull/2383 - the `select` parameter is there, but not something we want to promote, so good to not have it in the query builder

the `streaming.` parameters are documented in https://docs.vespa.ai/en/streaming-search.html for the same reason - we do not want these in the querybuilder either.

so everything looks right now, except the missing `noCache` - will look into

kkraune:
Added `noCache` to https://github.com/vespa-engine/documentation/pull/2383, too

PR #24570, opened 2022-10-25
re-throw runtime errors to get their stack trace

arnej27959:
@mpolden please review

note: long-term I think we may be better off with having something specific in our panic() calls, which would signal that we really *really* want to just exit with a simple message.

### DIFF
--- a/client/go/script-utils/main.go
+++ b/client/go/script-utils/main.go
@@ -6,6 +6,7 @@ package main
 import (
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


mpolden (approved):
> long-term I think we may be better off with having something specific in our panic() calls, which would signal that we really really want to just exit with a simple message.

Yes. We could use our own error type for those errors .

baldersheim (approved):

PR #24573, opened 2022-10-25
Remove existing load balancer if account changes

mpolden:
If the tenant changes the account for a deployment, e.g. from the default Vespa
account to the tenant's own account, we must ensure that the LB is provisioned
in the correct account.

@bratseth

FYI @ean


bratseth (approved):
Great! Should we have a validation override for changing account?

bratseth (approved):

mpolden:
Will look into the validation override together with how cloud account change affects node allocation.

bratseth:
Great, yes it's not directly related to this.

PR #24575, opened 2022-10-25
Arnej/add just exit error

arnej27959:
@mpolden please review
I'm not quite sure what the "go way" to do this is, suggestions welcome

### DIFF
--- a/client/go/util/just_exit.go
+++ b/client/go/util/just_exit.go
@@ -31,20 +31,20 @@ func (j *JustExitError) Error() string {
 	return j.String()
 }
 
-func JustExit(message string) *JustExitError {
+func JustExitMsg(message string) {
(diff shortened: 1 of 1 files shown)


mpolden (approved):
This works, but the "go way" leans more towards passing errors explicitly. 

In Vespa CLI all errors are passed up the stack and then we deal with it in `cli.Run()`:
https://github.com/vespa-engine/vespa/blob/2e05df2de19c2d5b87befa2ee6c4fd182dcb5630/client/go/cmd/root.go#L441-L454

`main` is only concerned with exiting with the wanted status:
https://github.com/vespa-engine/vespa/blob/2e05df2de19c2d5b87befa2ee6c4fd182dcb5630/client/go/cmd/vespa/main.go#L21-L33

PR #24576, opened 2022-10-25
Use a larger content node flavor by default

bratseth:
The previous flavor makes people run of out resources too often when trying vector search.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
@@ -90,23 +90,33 @@ public class CapacityPolicies {
             }
 
             return (requiresExclusiveHost(clusterSpec.type(), exclusive)
-                    ? versioned(clusterSpec, Map.of(new Version("0"), smallestExclusiveResources()))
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
```diff
@@ -90,23 +90,33 @@ public NodeResources defaultNodeResources(ClusterSpec clusterSpec, ApplicationId
...
+                    ? versioned(clusterSpec, Map.of(new Version(0), smallestExclusiveResources()))
+                    : versioned(clusterSpec, Map.of(new Version(0), smallestSharedResources())))
                     .with(architecture);
         }
 
-        return zone.getCloud().dynamicProvisioning()
-               ? versioned(clusterSpec, Map.of(new Version("0"), new NodeResources(2.0, 8, 50, 0.3)))
-               : versioned(clusterSpec, Map.of(new Version("0"), new NodeResources(1.5, 8, 50, 0.3)));
+        if (clusterSpec.type() == ClusterSpec.Type.content) {
+            // TODO: Simplify when no application is on an older version than 8.75
```
jonmv, line 99: Simplify the below, with cc flavours, as well then ;)

jonmv (approved):
So happy that someone got to use this version map again! :D

Issue #24476, opened 2022-10-17
Regex not working as expected


shubh9194:
**Describe the bug**
I am trying to use regex with attribute field in select query but not getting expected results.

Defined schema of fields used in query.
schema amDocument {

    document amDocument {
        field title type array<string> {
            indexing: summary | attribute | index
        }
       field collection type array<string> {
            indexing: summary | index
        }
      }
     field collection_attr type array<string> {
           indexing: input collection | attribute
           attribute: fast-search
      }
}

Document Ingested through postman
{
  "fields": {
 
   "table": "com.gs.document.store",
   "title": ["this is test document"],
   "collection":["public"],
    "sourcepath": [
      "2016-02.pdf"
    ],
    "filename": [
      "2016-02.pdf"
    ],
    "mimetype": [
      "application/pdf"
    ] 
  }
}

query used;
{
    "yql": "select * from sources * where collection_attr matches \"com\" ;",
    "timeout": "120s",
    "hits": 20
}
This should have returned no documents but it is returning the above ingested doc.

{
    "yql": "select * from sources * where title matches \"test document\" ;",
    "timeout": "120s",
    "hits": 20
}
this should have returned the document but it gives 0 documents
following query also not returning anything 
"yql": "select * from sources * where title matches \"test.*\" ;",

with title field, I am still able to get some results. like using title matches \"test\"
But with collection_attr, I get the document with almost any value in matches.

I am not able to understand why the above queries not working.
**Screenshots**

**Environment (please complete the following information):**
 - Infrastructure: Kubernetes
 - Versions [e.g. 22]

**Vespa version**
7.589

**Additional context**

kkraune:
@geirst , I looked into the `matches` in an `array<string>` field - this works well:

````
schema music {
    document music {
        field album type array<string> {
            indexing: summary | attribute
        }
    }
}

{
    "put": "id:mynamespace:music::a-head-full-of-dreams",
    "fields": {
        "album": ["A Head Full of Dreams"]
    }
}

vespa query 'select * from sources * where album matches "notindoc"'
````

This gives 0 results.

However, adding `fast-search` makes anything match:
````
schema music {
    document music {
        field album type array<string> {
            indexing: summary | attribute
            attribute: fast-search
        }
    }
}

{
    "put": "id:mynamespace:music::a-head-full-of-dreams",
    "fields": {
        "album": ["A Head Full of Dreams"]
    }
}


[...]

PR #24579, opened 2022-10-25
Verify that single dictionary entry from lookupRange() is a match.

toregge:
@geirst : please review

This is related to #24476


geirst (approved):
👍

baldersheim (approved):

PR #24580, opened 2022-10-25
use common algorithms for filter creation

havardpe:
add more testing
keep children blueprints in unique pointers

@geirst please review


havardpe:
@arnej27959 FYI

geirst (approved):
👍

PR #24581, opened 2022-10-25
Just check name of exception to avoid an issue if ClassCastException …

baldersheim:
…comes from different class loader

@bjorncs or @jonmv PR

### DIFF
--- a/container-core/src/main/java/com/yahoo/processing/request/CloneHelper.java
+++ b/container-core/src/main/java/com/yahoo/processing/request/CloneHelper.java
@@ -98,7 +98,7 @@ public class CloneHelper {
         try {
             return cloneByReflection(object);
         } catch (IllegalArgumentException e) {
-            if ( ! (e.getCause() instanceof ClassCastException))
+            if ( ! (e.getCause() != null && ClassCastException.class.getName().equals(e.getCause().getClass().getName())))
                 throw e;
 
             // When changing bundles you might end up having cached the old method pointing to old bundle,


jonmv (commented):
Are you sure this can happen?

jonmv:
I guess this doesn't hurt, but I didn't think this was possible.

bratseth (approved):
I think this needs a comment or it might quickly revert back

bjorncs (commented):
Sure about this?

baldersheim:
No, this is just be my guess to how you can end up with
java.lang.IllegalArgumentException: java.lang.ClassCastException@1b0496ad
	at jdk.internal.reflect.GeneratedMethodAccessor304.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at com.yahoo.processing.request.CloneHelper.cloneByReflection(CloneHelper.java:121)
	at com.yahoo.processing.request.CloneHelper.objectClone(CloneHelper.java:99)
	at com.yahoo.search.query.properties.CloneHelper.objectClone(CloneHelper.java:17)

bjorncs:
We need the full stack trace to be sure. My guess is that whatever code path logging the exception only includes the cause (not outer exception), or some code of the stack is wrapping or partially unwrapping the exception before rethrowing.

container-core/src/main/java/com/yahoo/processing/request/CloneHelper.java
```diff
@@ -98,7 +98,7 @@ else if (object instanceof HashSet)
         try {
             return cloneByReflection(object);
         } catch (IllegalArgumentException e) {
-            if ( ! (e.getCause() instanceof ClassCastException))
+            if ( ! (e.getCause() != null && ClassCastException.class.getName().equals(e.getCause().getClass().getName())))
```
bjorncs, line 101: Log exception before `if` with level `FINE` and enable verbose logging for this class through feature flag.

bjorncs (changes requested):

jonmv:
We have strong confirmation the cause is indeed `null`, and we're confused by a message string which is `ClassCastException@<some hash>` ... Ü

PR #24582, opened 2022-10-25
Fall back to comparing exception message

jonmv:
@baldersheim @bratseth @bjorncs please review and merge.


baldersheim (approved):

jonmv:
Reading the javadoc for `Method.invoke`, it looks to me like all possible causes of `IllegalArgumentException`, in _this_ case, are the class cast ones, so perhaps we should just do less checking instead ...

jonmv:
Or just use the actual class as key in the method cache, rather than the class name ... ?

baldersheim:
We should not catch exceptions unless we are sure. Full exposure, no hiding.
Nor should not put anything in the cache that is not strictly necessary.
Keep in mind that the cache should not really be there. I think I will add a warning statement when we put something in the cache, as it is better to fix the code.

PR #24584, opened 2022-10-25
Add a hook for tracking inserts into the method cache and log wheneve…

baldersheim:
…r clone methods are added.

@jonmv @bjorncs or @bratseth PR


bratseth (approved):
Good idea.

bjorncs (commented):
👍

PR #24586, opened 2022-10-25
Jonmv/use classes as keys

jonmv:
@baldersheim @bratseth please review.


baldersheim (changes requested):
This will leave code from old bundles in this cache forever. That is definitely not what you want.
You want to detect that there is a change and drop the cache. and the old code.

jonmv:
If only the map was a `WeakHashMap` ...

jonmv:
Ok, let's clear entries whenever a new class has the same name, instead.

jonmv:
Or would you rather we drop the entire cache whenever this happens?

jonmv:
That's probably the right thing to do.

vespajlib/src/main/java/com/yahoo/collections/MethodCache.java
```diff
@@ -35,7 +35,7 @@ public Method get(Object object) {
     }
     public Method get(Object object, Consumer<String> onPut) {
         Method m = cache.get(object.getClass().getName());
-        if ( ! m.getDeclaringClass().isAssignableFrom(object.getClass())) {
+        if ( m != null && ! m.getDeclaringClass().isAssignableFrom(object.getClass())) {
```
jonmv, line 38: If these calls are slow, we could also just use a composite value of (class, method), and compare class for identity.

jonmv:
Hang on, test on the way.

jonmv:
Ungh ... the weak-ref points to that pair, which isn't owned by anyone else. Need to use refs to the method and object class, of obviously ...

vespajlib/src/test/resources/dummy
```diff
@@ -0,0 +1 @@
+dummy
```
baldersheim, line 1: Missing EOL at EOF

vespajlib/src/main/java/com/yahoo/collections/MethodCache.java
```diff
@@ -33,18 +33,23 @@ public void clear() {
...
-        Method m = cache.get(object.getClass().getName());
-        if (m == null) {
-            m = lookupMethod(object);
-            if (m != null) {
-                if (onPut != null)
-                    onPut.accept(object.getClass().getName());
-                cache.put(object.getClass().getName(), m);
-            }
+        Pair<Class<?>, Method> pair = cache.get(object.getClass().getName());
+        if (pair != null && pair.getFirst() != object.getClass()) {
```
baldersheim, line 39: You need a comment similar to the one above to explain the dropping of the cache.

baldersheim (changes requested):
Functionality should be good now, but at least a comment when dropping the cache.
Perhaps also reason for having pair of class and method as value.

vespajlib/src/main/java/com/yahoo/collections/MethodCache.java
jonmv:
```suggestion
        // When changing bundles, you might end up having cached the old method pointing to the old bundle.
        // That will then lead to a class cast exception when invoking the wrong clone method.
        // Whenever we detect a new class with the same name, we therefore drop the entire cache. 
        // This is also the reason for caching the pair of method and original class—not just the method.
        if (pair != null && pair.getFirst() != object.getClass()) {
```

PR #24590, opened 2022-10-26
Revert mpolden/lb account change

bratseth:


jonmv (approved):
I guess this was meant to fix an issue with load balancers in the wrong account, and that issue remains after the revert?

bratseth:
Yes, we still need the PR, but not on the same Vespa release as that will stop config server upgrade. I think the issue it fixes is supporting _switching_ account.

PR #24594, opened 2022-10-26
Jonmv/restore json deps test scope

jonmv:
@gjoranv

### DIFF
--- a/container-test/pom.xml
+++ b/container-test/pom.xml
@@ -101,6 +101,12 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
+      <!-- required for container-search code using org.json -->
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
       <!-- TODO: this, and probably others, could be removed from here if we make the fat jar the default artifact for jdisc_core -->
       <groupId>org.lz4</groupId>
       <artifactId>lz4-java</artifactId>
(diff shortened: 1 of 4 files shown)


bratseth:
Could we put a `// TODO: Remove on Vespa 9 `on these?

jonmv:
That would be nice idneed.

container-test/pom.xml
```diff
@@ -100,6 +100,12 @@
       <artifactId>commons-compress</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <!-- not used by Vespa, but was historically on test classpath -->
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <scope>compile</scope>
```
gjoranv, line 107: Please don't use explicit 'compile' scope. It could imply that we manage dependency scopes, which we shouldn't.

gjoranv (approved):

container-test/pom.xml
jonmv:
```suggestion
```

container-test/pom.xml
```diff
@@ -100,6 +100,11 @@
       <artifactId>commons-compress</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <!-- not used by Vespa, but was historically on test classpath -->
```
gjoranv, line 104: Please add Vespa 9 TODO here as well.

parent/pom.xml
```diff
@@ -994,6 +994,11 @@
                 <artifactId>xercesImpl</artifactId>
                 <version>2.12.2</version>
             </dependency>
+            <dependency>
```
gjoranv, line 997: Please add Vespa 9 TODO here as well.

PR #24599, opened 2022-10-26
Fall back to default cloud account in node spec

mpolden:
This ensures that node spec always contains the cloud account being used (either
custom or default).


mpolden:
Merge with internal PR.

bratseth (commented):
I suggest making CloudAccount non-optional in Cloud and NodeSpec, and adding a CloudAccount.empty() instead. All these optionals are hard to deal with, and we will eventually have explicit accounts most places or maybe everywhere.

Otherwise LGTM!

mpolden:
PTAL

bratseth:
👍

PR #24603, opened 2022-10-26
ConfigurationRuntimeException when building model means invalid package

jonmv:
@bratseth please review and merge if you think it's right. Otherwise, errors in, e.g., services.xml with config for nonexistent keys results in a HTTP 500 and no usable information propagated back through the controller.


bratseth:
I think we should reclassify those as IllegalArgumentExceptionm instead. I see this is also thrown in all kinds of other situations which are not related to user input.

PR #24607, opened 2022-10-26
Use IllegalArgumentException when appropriate

bratseth:
Like this is better imho


jonmv (changes requested):
Please re-run `./bootstrap.sh` to get the newest abi-checker to avoid reformatting this abi spec again.

jonmv (approved):
Aggreed, this is better.

PR #24621, opened 2022-10-27
Allow turning off jobs in a zone, optionally per application

jonmv:
@freva please review and mereg.

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
@@ -296,6 +296,13 @@ public class PermanentFlags {
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundListFlag<String> DISABLED_DEPLOYMENT_ZONES = defineListFlag(
+            "disabled-deployment-zones", List.of(), String.class,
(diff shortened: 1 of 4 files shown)


jonmv:
Could consider wiring this through `DeploymentStatus` alongside `coolingDownUntil` et al to show, if we end up using it much ... but it's not straightforward to get it all right.

flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
```diff
@@ -296,6 +296,13 @@ public class PermanentFlags {
             "Takes effect at redeployment",
             ZONE_ID, APPLICATION_ID);
 
+    public static final UnboundListFlag<String> DISABLED_DEPLOYMENT_ZONES = defineListFlag(
```
freva, line 299: Why not boolean flag with `ZONE_ID` dimension?

jonmv: AFAIK that zone dimension is the zone of the requestor, but that's always the controller zone in this case?

jonmv: Ah, maybe that's not the case ... Then that's obviously the right thing.

freva (approved):

PR #24628, opened 2022-10-27
Node-Repo: Always lock by application ID

freva:
I've looked at the usage of all the locking methods that I could find in node-repo and created a graph of all paths where we take more than 1 lock concurrently:
image

Red boxes are the application locks. Green is the "unallocated" lock, i.e. the lock we take when we want to modify a node that does not have any allocation, when we are allocating or doing anything that could affect allocation (e.g. adding/removing nodes). The cyan boxes are the `lock(Node)` methods which will take either the application or the unallocated lock depending on if the given node has `Allocation` set or not. Finally, yellow is the session lock in `configserver` module, probably not relevant for this.

Gradient that goes top-down means we first take the top colored lock, then while still holding it, the bottom colored lock. Gradient that goes sideways means first take the left colored lock, then release it, then take the right colored lock.

We have already eliminated the obvious deadlocks, but they still occasionally happen. The reason for that are the cyan boxes which may take either lock depending on the node allocation state, before taking the other locks.

This PR turns all the cyan boxes red:
* All infrastructure nodes will lock by their application ID, which we know ahead of time even if the node is not yet allocated.
* Tenant nodes will continue locking by the application in their `Allocation`, but we will no longer ever remove `Allocation` from a node. Since nodes are created with an `Allocation`, that means that all tenant nodes should effectively always have it (though actually enforcing it will require rewriting many unit tests :cry: )

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -402,15 +402,6 @@ public final class Node implements Nodelike {
                         exclusiveToClusterType, switchHostname, trustStoreItems, cloudAccount);
     }
 
-    /** Returns a new Node without an allocation. */
-    public Node withoutAllocation() {
-        return new Node(id, ipConfig, hostname, parentHostname, flavor, status, state,
-                        Optional.empty(), history, type, reports, modelName, reservedTo,
-                        exclusiveToApplicationId, exclusiveToClusterType, switchHostname, trustStoreItems,
-                        cloudAccount);
-    }
-
-
     /** Returns a copy of this node with IP config set to the given value. */
     public Node with(IP.Config ipConfig) {
         return new Node(id, ipConfig, hostname, parentHostname, flavor, status, state,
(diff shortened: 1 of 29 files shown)


bratseth:
Great!

"Tenant nodes will continue locking by the application in their Allocation, but we will no longer ever remove Allocation from a node. Since nodes are created with an Allocation, that means that all tenant nodes should effectively always have it (though actually enforcing it will require rewriting many unit tests 😢 )"

Yes, we are testing a lot of stuff which we never do any more. Maybe it is time to fix this, I think it will unlock other simplifications too.

bratseth (approved):

freva:
Besides all the unit tests, there is also the allocation code that briefly creates `Node`s without allocation via `NodeCandidate::withNode`/`Node::reserve` (the last one is used in a few classes outside of node-repository). This one is harder to fix since it requires `Membership`, which requires index, which we cant assign before we actually decide to accept the node. So need to rewrite the allocation code to work on `NodeCandidate` or something all the way through.

freva:
Requiring allocation for non-tenant nodes this is much harder since these nodes can be added at any time via the REST API and the proxy app also depends on application package

mpolden (approved):
Great!

> Since nodes are created with an Allocation, that means that all tenant nodes should effectively always have it (though actually enforcing it will require rewriting many unit tests)

We did this multiple times in controller-server after the initial rewrite. Very boring, but worth it since we can simplify a lot of code.

> Yes, we are testing a lot of stuff which we never do any more. Maybe it is time to fix this, I think it will unlock other simplifications too.

Agree! One example is that we still kind of support allocating directly on hosts. We never do this in practice, but tests do.

PR #24638, opened 2022-10-28
Better error handling if both schemas/ and searchdefinitions/ exist [run-systemtest]

hmusum:
If something fails when having both directories, throw an InvalidApplicationException with an explanation about what to do

Since we just rethrow the exception this should not change behavior (e.g. when bootstrapping config servers), only what feedback the customer gets.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -841,15 +841,24 @@ public class SessionRepository {
         File schemasDir = applicationDir.resolve(ApplicationPackage.SCHEMAS_DIR.getRelative()).toFile();
         File sdDir = applicationDir.resolve(ApplicationPackage.SEARCH_DEFINITIONS_DIR.getRelative()).toFile();
         if (sdDir.exists() && sdDir.isDirectory()) {
-            File[] sdFiles = sdDir.listFiles();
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -841,15 +841,24 @@ private void moveSearchDefinitionsToSchemasDir(java.nio.file.Path applicationDir
...
+                if (sdFiles != null) {
+                    Files.createDirectories(schemasDir.toPath());
+                    Arrays.asList(sdFiles).forEach(file -> Exceptions.uncheck(
+                            () -> Files.move(file.toPath(),
+                                             schemasDir.toPath().resolve(file.toPath().getFileName()),
+                                             StandardCopyOption.REPLACE_EXISTING)));
+                }
+                Files.delete(sdDir.toPath());
+            } catch (IOException | UncheckedIOException e) {
+                if (schemasDir.exists() && schemasDir.isDirectory())
```
hakonhall, line 855: Failing if both directories exists would likely break too many applications?

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: Yes, that's the assumption

PR #24641, opened 2022-10-28
create optimized filter search

havardpe:
test strict/constraint propagation
test filters dropped due to short-circuit

@geirst please review


geirst (approved):
👍

PR #24647, opened 2022-10-28
QueryBuilder: Use different input types depending on parameter type

freva:
This will use a switch/toggle for boolean types, `<input` with `type="number"` for numerical inputs, additionally I've added min/max values where documented and it'll show an error if value is outside range.
image


kkraune (approved):
awesome!

PR #24653, opened 2022-10-31
Add encryption capabilities to core dump handler

vekterli:
@freva please review
@bjorncs FYI

Once wired in (not currently the case), a `Supplier` of non-null `SecretSharedKey` instances will trigger:

 1. Wrapping the output stream with an encrypting output stream using the secret component of the supplied key. Zstd compression is handled on the input stream, so this should transparently encrypt compressed data. To disambiguate, encrypted core dumps are suffixed with an additional `.enc` file extension.
 2. Emitting a public decryption token as part of the metadata using the shared component of the supplied key.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
@@ -1,7 +1,9 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.node.admin.maintenance.coredump;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.yahoo.security.SecretSharedKey;
 import com.yahoo.security.SharedKeyGenerator;
 import com.yahoo.vespa.hosted.node.admin.configserver.noderepository.NodeSpec;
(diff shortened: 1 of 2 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -175,6 +184,7 @@ String getMetadata(NodeAgentContext context, ContainerPath coredumpDirectory, Su
                     .resolve(context.containerName().asString())
                     .resolve(coredumpDirectory.getFileName().toString())
                     .resolve(coredumpFile.getFileName().toString()).toString());
+            decryptionToken.ifPresent(token -> metadata.put("decryption_token", token));
```
freva, line 187: This JSON is sent to a Vespa application, so need to add this field there (Not an issue for this PR since this is not enabled yet, just reminder :smile: )

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -151,10 +158,12 @@ Optional<ContainerPath> enqueueCoredump(ContainerPath containerCrashPath, Contai
 
     void processAndReportSingleCoredump(NodeAgentContext context, ContainerPath coredumpDirectory, Supplier<Map<String, Object>> nodeAttributesSupplier) {
         try {
-            String metadata = getMetadata(context, coredumpDirectory, nodeAttributesSupplier);
+            Optional<SecretSharedKey> sharedCoreKey = Optional.ofNullable(secretSharedKeySupplier.get());
+            Optional<String> decryptionToken = sharedCoreKey.map(k -> k.sealedSharedKey().toTokenString());
+            String metadata = getMetadata(context, coredumpDirectory, nodeAttributesSupplier, decryptionToken);
```
freva, line 163: `getMetadata()` writes the resulting metadata to a file, but if the file already exists, it'll return the data from the file ignoring the arguments. If we fail to report this coredump, we will try again later, but then we'll end up encrypting the core with the new key while the metadata will have the old `decryption_token`.

freva (commented):
... otherwise LGTM.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
freva: No problem overwriting it, this is done to avoid having to reprocess the coredump with `gdb` etc.

vekterli: Understood. Presumably this should only happen if either a) core dump reporting fails, or b) writing the compressed/encrypted core dump fails? I'd expect both of those to be fairly rare occurrences. For now I think the pragmatic route with overwriting should be fine; we can later consider other options such as patching in a new token into the existing metadata if we feel it makes sense (it's "just JSON"™ after all).

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
vekterli: Thanks, I'll make sure to update the schema prior to wiring this in anywhere.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
freva: Agree :+1:

vekterli: Patching in the new token value turned out to be not that much extra work after all, so I went with that. PTAL.

freva (approved):

PR #24665, opened 2022-10-31
Support standard IO streams for several encryption tool commands

vekterli:
@bjorncs please review

Useful for avoiding the need for intermediate files, such as when piping the output of decryption to a Zstd decompressor.

Adds stdio support to:
 * Encryption input (can't add to output since token is printed verbatim to stdout already)
 * Decryption input
 * Decryption output

Specified by substituting the file name with a single `-` character.

### DIFF
--- a/vespaclient-java/src/main/java/com/yahoo/vespa/security/tool/CliUtils.java
+++ b/vespaclient-java/src/main/java/com/yahoo/vespa/security/tool/CliUtils.java
@@ -27,7 +27,7 @@ public class CliUtils {
             return stdIn;
         } else {
             var inputPath = Paths.get(pathOrDash);
-            if (!inputPath.toFile().exists()) {
+            if (!Files.exists(inputPath)) {
                 throw new IllegalArgumentException("Input file '%s' does not exist".formatted(inputPath.toString()));
             }
             return Files.newInputStream(inputPath);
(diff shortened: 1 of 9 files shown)


vespaclient-java/src/main/java/com/yahoo/vespa/security/tool/CliUtils.java
```diff
@@ -16,4 +22,25 @@ public static String optionOrThrow(CommandLine arguments, String option) {
         return value;
     }
 
+    public static InputStream inputStreamFromFileOrStream(String pathOrDash, InputStream stdIn) throws IOException {
+        if ("-".equals(pathOrDash)) {
+            return stdIn;
+        } else {
+            var inputPath = Paths.get(pathOrDash);
+            if (!inputPath.toFile().exists()) {
```
bjorncs, line 30: Use `Files.exists(path)` instead of converting to `File`. The latter is basically the legacy API for file management.

bjorncs (approved):

vespaclient-java/src/main/java/com/yahoo/vespa/security/tool/CliUtils.java
vekterli: Fixed here and in some other places as well.

bjorncs (approved):

PR #24678, opened 2022-11-01
Go back to previous variant of default node resources for content clu…

hmusum:
…sters on-prem.

We don't have HW capacity on-prem to use 300 Gb disk (and probably not 16 Gb memory either), so go back to what we had before version 8.75


bratseth:
We did this for a good reason - people getting started with vector search quickly running out of resources and getting a bad experience -, so I think we should only change back on premise.

hmusum:
Agreed, isn't that what my change does?

bratseth (approved):
Um, yes.

PR #24681, opened 2022-11-01
Control transitive dependencies of pig for tomcat:jasper-runtime, tom…

baldersheim:
…cat:jasper-compiler and groovy-all

@bjorncs or @gjoranv PR


gjoranv (commented):
👍

PR #24687, opened 2022-11-01
Simplify

hmusum:


jonmv (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
```diff
@@ -115,9 +114,7 @@ private NodeResources clusterControllerResources(ClusterSpec clusterSpec, boolea
         if (requiresExclusiveHost(clusterSpec.type(), exclusive)) {
             return versioned(clusterSpec, Map.of(new Version(0), smallestExclusiveResources()));
         }
-        return versioned(clusterSpec, Map.of(new Version(0), new NodeResources(0.25, 1.14, 10, 0.3),
-                                             new Version(7, 586, 50), new NodeResources(0.25, 1.333, 10, 0.3),
-                                             new Version(7, 586, 54), new NodeResources(0.25, 1.14, 10, 0.3)));
```
jonmv, line 120: Hehe :) Can't remember if it was left as a sample for later, but now we have another, so ... :)

PR #24692, opened 2022-11-01
- Handle CancellationException.

baldersheim:
- Check stopped on every iteration.
- Resort to Thread.interrupt() if nothing else helps, and log a warning.

@bjorncs or @jonmv PR


bjorncs (approved):
This is could also be solved with`volatile`

jonmv:
👍

PR #24633, opened 2022-10-28
Avoid stalling akamai requests during routing table reload.

baldersheim:
@mpolden @bjorncs PR

PR #24695, opened 2022-11-01
Hold monitor on reload too.

baldersheim:
@hakonhall or @mpolden PR
This fixes a race condition introduced in #24633


mpolden (commented):
👍

PR #24697, opened 2022-11-01
Jonmv/more synch in config sub unit test

jonmv:
@bjorncs or @baldersheim or @hmusum  please review.

I'm not super-happy about this. I think the `ConfigSubscription` is weird, and way too hard to please. It seems some of its subclasses were wrong before, in certain scenarios, and they still are; e.g., I worked around that by adding some "atomicity guarantee" to the config effectively served by the `DirConfigSource` in container-core tests. I also imitated the `JRTConfigSubscription` in the `ConfigSetSubscription` for no good reason, other than to avoid warning bonanza—the state induced by running the different code paths is identical. 

At least the `reloadConfig` wiring is gone now, and replaced with a notion of a "generation" on those test config sources (file, set) that has to be manually incremented—better explicit and dumb, than magical and wrong.

### DIFF
--- a/config/src/main/java/com/yahoo/config/subscription/ConfigSet.java
+++ b/config/src/main/java/com/yahoo/config/subscription/ConfigSet.java
@@ -5,6 +5,7 @@ import com.yahoo.config.ConfigInstance;
 import com.yahoo.vespa.config.ConfigKey;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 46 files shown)


baldersheim (changes requested):
This is way too much to review in one go.
And config lib stuff is complicated enough and needs a careful review.
The PR must be split in smaller commits handling one thing at the time.

jonmv:
#24702 has the fix for the unit test, along with some cleanup.

PR #24698, opened 2022-11-01
There is no guarantee for when RoutingGenerator.routingTable() will r…

baldersheim:
…eflect last value.

No option but wait until it is so.

@jonmv or @hakonhall or @mpolden  PR


jonmv (changes requested):
https://github.com/vespa-engine/vespa/pull/24699 instead

jonmv (approved):

PR #24700, opened 2022-11-01
Move hadoop dependency management to parent pom.[run-systemtest]

baldersheim:
@bjorncs or @gjoranv  PR

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -442,6 +442,7 @@
             </build>
         </profile>
     </profiles>
+
     <dependencyManagement>
         <dependencies>
             <!-- Please keep this list lexically sorted by groupId, then artifactId.-->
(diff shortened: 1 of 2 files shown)


vespa-hadoop/pom.xml
```diff
@@ -17,109 +17,28 @@
...
-        <sun-jersey-json.version>1.19.4</sun-jersey-json.version>
-        <jettison.version>1.5.1</jettison.version>
-        <pig.version>0.16.0</pig.version>
-        <tomcat-jasper.version>5.5.23</tomcat-jasper.version>
     </properties>
 
-    <dependencyManagement>
-        <dependencies>
-            <!-- Force newer version of jetty-util (to match Jetty version of jdisc) -->
-            <dependency>
```
bjorncs, line 31: FYI the unit test framework for hadoop/pig requires Jetty 9.4. We'll have to reintroduce dependency management to force 9.4 once we migrate to Jetty 11.

bjorncs (approved):

parent/pom.xml
```diff
@@ -1084,6 +1085,89 @@
...
+                <!-- Transitive dependencies from pig-0.16 up-to-date -->
+                <groupId>tomcat</groupId>
+                <artifactId>jasper-compiler</artifactId>
+                <version>${tomcat-jasper.version}</version>
+            </dependency>
+            <!-- Hadoop dependencies -->
+            <dependency>
+                <groupId>org.apache.hadoop</groupId>
+                <artifactId>hadoop-common</artifactId>
+                <version>${hadoop.version}</version>
```
bjorncs, line 1122: @gjoranv Do we want to add exclusions to dependency management in parent pom? We'll have to reintroduce slf4j wherever we use these artifacts outside jdisc osgi context.

gjoranv: No, unless there are really special circumstances. Nasty surprices will be the result.

gjoranv (commented):
Artifacts only used in vespa-hadoop can have their dep management kept there. 

We should limit management of transitive deps to a minimum, due to pom maintenance load and risk of failure due to incompatible deps. IMO, we should only add it to fix serious security issues.

vespa-hadoop/pom.xml
baldersheim: I noticed, better to start off from scratch.

parent/pom.xml
baldersheim: The special circumstance is the log4j security issue that we **never** want anywhere.

PR #24702, opened 2022-11-01
Jonmv/config lib more sync 2

jonmv:
@baldersheim please review. A cleaner take on #24697, without any real changes to config lib code, just the test.

### DIFF
--- a/container-core/src/test/java/com/yahoo/container/di/ContainerTest.java
+++ b/container-core/src/test/java/com/yahoo/container/di/ContainerTest.java
@@ -282,8 +282,9 @@ public class ContainerTest extends ContainerTestBase {
                      () -> getNewComponentGraph(container, currentGraph));
 
         ExecutorService exec = Executors.newFixedThreadPool(1);
+        dirConfigSource.clearCheckedConfigs();
         Future<ComponentGraph> newGraph = exec.submit(() -> getNewComponentGraph(container, currentGraph));
-
+        dirConfigSource.awaitConfigChecked(10_000);
         try {
             newGraph.get(1, TimeUnit.SECONDS);
             fail("Expected waiting for new config.");
--- a/container-core/src/test/java/com/yahoo/container/di/ContainerTest.java
+++ b/container-core/src/test/java/com/yahoo/container/di/ContainerTest.java
@@ -112,10 +112,8 @@ public class ContainerTest extends ContainerTestBase {
         // bundle-2 is installed, bundle-1 has been uninstalled
         assertEquals(1, osgi.getBundles().length);
         assertEquals("bundle-2", osgi.getBundles()[0].getSymbolicName());
-    }
 
-    //@Test TODO
-    public void deconstructor_is_given_guice_components() {
+        container.shutdownConfigRetriever();
     }
 
     @Test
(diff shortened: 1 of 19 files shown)


jonmv:
It seems possible to replace the `reload(...)` code path with a reload mechanism on the test usages of file and set sources, since the sources are known to the caller in those cases. That requires the reload mechanism to _not be in real use_. I _think_ that is the case, but it _is_ `@PublicApi` in `messagebus` ...

container-core/src/test/java/com/yahoo/container/di/DirConfigSource.java
```diff
@@ -2,60 +2,72 @@
...
     }
 
-    public ConfigSource configSource() {
-        return configSource;
+    synchronized void clearCheckedConfigs() {
+        checked.clear();
+        doubleChecked = false;
     }
 
-    public void cleanup() {
```
baldersheim, line 49: No cleanup ?

container-core/src/test/java/com/yahoo/container/di/ContainerTestBase.java
```diff
@@ -31,20 +30,14 @@ public class ContainerTestBase {
...
     public void setup() {
-        dirConfigSource = new DirConfigSource(tmpDir, "ContainerTest-");
+        dirConfigSource = new DirConfigSource(tmpDir);
         componentGraph = new ComponentGraph(0);
         osgi = new TestOsgi(BundleTestUtil.testBundles());
     }
 
-    @AfterEach
-    public void cleanup() {
-        dirConfigSource.cleanup();
```
baldersheim, line 41: How is cleanup done ?

jonmv: The @TempDir takes care of it. The cleanup that was done here did nothing but return `false` (indicating it failed to delete the non-empty directory).

container-core/src/test/java/com/yahoo/container/di/DirConfigSource.java
jonmv: This was broken and did nothing.

container-core/src/test/java/com/yahoo/container/di/ContainerTest.java
```diff
@@ -282,8 +282,9 @@ void getNewComponentGraph_hangs_waiting_for_valid_config_after_invalid_config()
                      () -> getNewComponentGraph(container, currentGraph));
 
         ExecutorService exec = Executors.newFixedThreadPool(1);
+        dirConfigSource.clearCheckedConfigs();
         Future<ComponentGraph> newGraph = exec.submit(() -> getNewComponentGraph(container, currentGraph));
-
+        dirConfigSource.awaitConfigChecked(10_000);
```
jonmv, line 287: Without this, the reload below can be partially eaten by the other thread, resulting in a state where it fails the first run through all configs because of different generations, and never completes the second run because some configs don't have anything new. 
That this is possible doesn't seem right at all. I hope it's not the case for the actual JRT subscription, i.e., that that one ensures atomicity in some way, to avoid this state?

PR #24709, opened 2022-11-02
Bratseth/query builder

bratseth:


gjoranv (approved):

client/src/main/java/ai/vespa/client/dsl/EndQuery.java
```diff
@@ -76,15 +76,21 @@ public EndQuery limit(int limit) {
     }
 
     /**
-     * Semicolon.
-     * turn a query into fixed query.
+     * Calls fix()
      *
-     * @return the fixed query
+     * @deprecated use {link #fix}
```
gjoranv, line 81: Forgotten, or is the replacement TBD?

PR #24714, opened 2022-11-02
Enable the report-cores-via-cfg flag by default

hakonhall:
AND log progress of core dump processing.


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -154,29 +158,40 @@ Optional<ContainerPath> getCoredumpToProcess(ContainerPath containerCrashPath, C
      *
      * @return path to directory inside processing directory which contains the enqueued core dump file
      */
-    Optional<ContainerPath> enqueueCoredump(ContainerPath containerCrashPath, ContainerPath containerProcessingPath) {
+    Optional<ContainerPath> enqueueCoredump(NodeAgentContext context, ContainerPath containerCrashPath, ContainerPath containerProcessingPath) {
+        Predicate<String> isCoreDump = filename -> !HS_ERR_PATTERN.matcher(filename).matches();
```
freva, line 162: FYI there's `Pattern::asMatchPredicate` that does this

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
hakonhall: Nice to know, but it didn't improve the readability.

PR #24715, opened 2022-11-02
Encapsulate key identifier in own object

vekterli:
@bjorncs please review.

Enforces invariants and avoids having to pass raw byte arrays around.

### DIFF
--- a/security-utils/src/main/java/com/yahoo/security/KeyId.java
+++ b/security-utils/src/main/java/com/yahoo/security/KeyId.java
@@ -1,0 +1,89 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+import static com.yahoo.security.ArrayUtils.fromUtf8Bytes;
+import static com.yahoo.security.ArrayUtils.toUtf8Bytes;
+
+/**
+ * Represents a named key ID comprising an arbitrary (but length-limited)
+ * sequence of valid UTF-8 bytes.
+ *
+ * @author vekterli
+ */
+public class KeyId {
+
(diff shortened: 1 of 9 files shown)


security-utils/src/main/java/com/yahoo/security/KeyId.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Construct a KeyId containing the given sequence of bytes.
+     *
+     * @param keyIdBytes array of valid UTF-8 bytes. May be zero-length, but not null.
+     *                   Note: to avoid accidental mutations, the key bytes are deep-copied.
+     * @return a new KeyId instance
+     */
+    public static KeyId ofBytes(byte[] keyIdBytes) {
+        Objects.requireNonNull(keyIdBytes);
+        return new KeyId(keyIdBytes.clone());
```
bjorncs, line 41: Use `Arrays.copyOf()`. `Object.clone()` is a sad part of Java's history and should never be used...

security-utils/src/test/java/com/yahoo/security/KeyIdTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+        var id2 = KeyId.ofBytes(barBytes);
+        assertEquals("bar", id2.asString());
+        assertArrayEquals(barBytes, id2.asBytes());
+    }
+
+    @Test
+    void key_id_bytes_are_deep_copied_when_constructed_from_raw_byte_array() {
+        byte[] keyBytes = new byte[]{'f','o','o'};
+        byte[] expected = keyBytes.clone();
```
bjorncs, line 55: See previous comment

bjorncs (changes requested):

security-utils/src/main/java/com/yahoo/security/KeyId.java
vekterli: Array copying fixed to be more cheerful, less sad

bjorncs (approved):

PR #24716, opened 2022-11-02
Bjorncs/enforcer extension

bjorncs:


jonmv (approved):

gjoranv:
Are you sure that we now require exact version in the enforcers? I'm quite sure that I saw some explicitly given versions than what we actually use.

bjorncs:
It's gone now. I had to update several version number in enforcer poms after removing the range support.

PR #24718, opened 2022-11-02
open port for wireguard in acl

andreer:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #24723, opened 2022-11-02
Revert "open port for wireguard in acl"

andreer:
Reverts vespa-engine/vespa#24718

This doesn't do the trick, we need UDP not tcp ... 🤦‍♂️


baldersheim (approved):

gjoranv:
👍

PR #24724, opened 2022-11-02
Split up DynamicProvisioningMaintainer

freva:
Splits up the old `DynamicProvisioningMaintainer` into 4 maintainers:
* `HostCapacityMaintainer`: Is tasked with provisioning new hosts if required per `preprovsion-capacity` feature flag and mark excess hosts as `wantToDeprovsion`
* `HostResumeProvisioner`: Resumes provisioning of hosts in `provisioned`
* `HostDeprovisioner`: Deprovisions and removes from node-repo host's that have finished retiring. TODO: These should go via `dirty` so that host-admin gets a chance to sync logs to archive.
* `DiskReplacer`: Replaces root disk on hosts that `wantToRebuild`

By splitting it up into separate maintainers it should make things happen quicker, we no longer need to wait for provisioning of preprovision capacity to deprovision excess host. Additionally, the actual deprovisioning no longer holds any locks.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
@@ -104,9 +104,9 @@ public class HostCapacityMaintainer extends NodeRepositoryMaintainer {
 
                     // Retire the host to parked if possible, otherwise move it straight to parked
                     if (EnumSet.of(Node.State.reserved, Node.State.active, Node.State.inactive).contains(host.state())) {
-                        Node retiredHost = host.withWantToRetire(true, true, Agent.DynamicProvisioningMaintainer, nodeRepository().clock().instant());
+                        Node retiredHost = host.withWantToRetire(true, true, Agent.HostCapacityMaintainer, nodeRepository().clock().instant());
                         nodeRepository().nodes().write(retiredHost, mutex);
-                    } else nodeRepository().nodes().park(host.hostname(), true, Agent.DynamicProvisioningMaintainer, "Parked for removal");
+                    } else nodeRepository().nodes().park(host.hostname(), true, Agent.HostCapacityMaintainer, "Parked for removal");
                 } catch (UncheckedTimeoutException e) {
                     log.log(Level.WARNING, "Failed to mark " + host.hostname() +
                             " for deprovisioning: Failed to get lock on node, will retry later");
@@ -226,7 +226,7 @@ public class HostCapacityMaintainer extends NodeRepositoryMaintainer {
                                            osVersion, HostSharing.shared, Optional.empty(), nodeRepository().zone().cloud().account(),
                                            provisionedHosts -> {
                                                hosts.addAll(provisionedHosts.stream().map(ProvisionedHost::generateHost).toList());
-                                               nodeRepository().nodes().addNodes(hosts, Agent.DynamicProvisioningMaintainer);
+                                               nodeRepository().nodes().addNodes(hosts, Agent.HostCapacityMaintainer);
                                            });
             return hosts;
         } catch (NodeAllocationException | IllegalArgumentException | IllegalStateException e) {
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
@@ -37,6 +37,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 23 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostResumeProvisioner.java
```diff
@@ -65,7 +67,7 @@ protected double maintain() {
                                       " children, failing out the host recursively", e);
                 // Fail out as operator to force a quick redeployment
                 nodeRepository().nodes().failOrMarkRecursively(
-                        host.hostname(), Agent.DynamicProvisioningMaintainer, "Failed by HostProvisioner due to provisioning failure");
+                        host.hostname(), Agent.operator, "Failed by HostProvisioner due to provisioning failure");
```
freva, line 70: This was `operator` originally (see comment above), changed to DPM in #23345

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -77,99 +73,44 @@ public class DynamicProvisioningMaintainer extends NodeRepositoryMaintainer {
...
+        try (var lock = nodeRepository().nodes().lockUnallocated()) {
+            NodeList nodes = nodeRepository().nodes().list(); // Reread nodes under lock
+            for (Node host : excessHosts) {
+                Optional<NodeMutex> optionalMutex = nodeRepository().nodes().lockAndGet(host, Duration.ofSeconds(10));
+                if (optionalMutex.isEmpty()) continue;
+                try (NodeMutex mutex = optionalMutex.get()) {
+                    host = mutex.node();
+                    if (!canRemoveHost(host)) continue;
+                    if (!nodes.childrenOf(host).stream().allMatch(HostCapacityMaintainer::canDeprovision))
+                        continue;
```
hakonhall, line 103: These are linked to candidatesForRemoval --- perhaps there is a way for both to use the same function somehow?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -77,99 +73,44 @@ public class DynamicProvisioningMaintainer extends NodeRepositoryMaintainer {
...
+
+                    // Retire the host to parked if possible, otherwise move it straight to parked
+                    if (EnumSet.of(Node.State.reserved, Node.State.active, Node.State.inactive).contains(host.state())) {
+                        Node retiredHost = host.withWantToRetire(true, true, Agent.HostCapacityMaintainer, nodeRepository().clock().instant());
+                        nodeRepository().nodes().write(retiredHost, mutex);
+                    } else nodeRepository().nodes().park(host.hostname(), true, Agent.HostCapacityMaintainer, "Parked for removal");
+                } catch (UncheckedTimeoutException e) {
+                    log.log(Level.WARNING, "Failed to mark " + host.hostname() +
```
hakonhall, line 111: Consider adding reporting of success factor for this maintainer.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostResumeProvisioner.java
hakonhall: hm, could quick redeployment be triggered for a HostResumeProvisioner agent too?

freva: Done

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -77,99 +73,44 @@ public class DynamicProvisioningMaintainer extends NodeRepositoryMaintainer {
...
-                host = mutex.node();
-                if (!host.status().wantToRebuild()) {
-                    continue;
+    private void markForRemoval(List<Node> excessHosts) {
+        if (excessHosts.isEmpty()) return;
+
+        try (var lock = nodeRepository().nodes().lockUnallocated()) {
+            NodeList nodes = nodeRepository().nodes().list(); // Reread nodes under lock
+            for (Node host : excessHosts) {
+                Optional<NodeMutex> optionalMutex = nodeRepository().nodes().lockAndGet(host, Duration.ofSeconds(10));
```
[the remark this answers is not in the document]
freva: Right, changed to group nodes by node type and take the application lock + unallocated lock together. This does make assumptions on implementation of `lock(Node)`, but without that we'd have to retake unallocated lock for ever host...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
freva: Done

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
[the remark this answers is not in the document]
hakonhall: There could be a user-friendly one-liner to remind the viewer about this - say throwing if application on current node is different from the application of the first node.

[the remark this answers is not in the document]
freva: The problem is that the application on the node does not matter, the lock is taken on an application based on node-type: https://github.com/vespa-engine/vespa/blob/d76536ad6256ae1dfe780bb9be34e1f4574b13fb/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java#L853:L859
So from this method, we could see that the host went from unallocated to allocated, for example, but the lock it holds would still be correct.

hakonhall (approved):

PR #24725, opened 2022-11-03
Draft streamed application package

jonmv:
@mpolden for discussion. I don't think using this will be too hard. 

The idea is that we can use it only for deployments. There's one gotcha, and that is that we have to guess which files will be included by XML preprocessing, but that only allows valid XML in any case, so I think we may as well include all XML files we find? 

See the unit test for example usage. I think we'll need to convert the certificate metadata to be held in a lazy input stream, which generates its data after the package has been written as the first multi-part form data. 

Replacing zip entries (cert, trust store, tester files) should be straightforward.


jonmv:
Yes, forgot to mention, but we gobble up all the input from the HTTP layer, in our handlers, so to actually stream the package through on upload, we'd have to 

1. maybe throttle input read from jetty
2. ensure meta data is first in the multi-part form
3. change to a stream multi-part form parser which parses lazily

I think that's too much. I think we're far away from having problems because of this load; rather, it's the concurrent deployments which hold on to the application package for the entire duration of the deployment action.

jonmv:
@freva please review.

mpolden (approved):
👍

freva (approved):

PR #24726, opened 2022-11-03
Control the heas size percentage used as default for non-combined app…

baldersheim:
…lication clusters.

@hmusum or @bjorncs PR
@jobergum @yngveaasheim FYI


hmusum (approved):

bjorncs (commented):
👍

PR #24729, opened 2022-11-03
* add utility for directory/file fixups

arnej27959:
* use "sudo go test" to exercise chown/chmod in test
* also add test for IsExecutableFile

@mpolden or @baldersheim please review


baldersheim (approved):

baldersheim (changes requested):
Tests are unhappy

PR #24736, opened 2022-11-03
Since large constants are always tensors avoid creating a string just…

baldersheim:
… to convert it back to tensors.

That is a lot of work for nothing.

@lesters or @bratseth PR


baldersheim:
@hmusum FYI

bratseth (approved):
👍   this is for _small_ constants though ...

PR #24733, opened 2022-11-03
Balder/remove cluster from system info

baldersheim:
@bratseth and @gjoranv PR

PR #24738, opened 2022-11-03
Add a provider for cluster information.

baldersheim:
@gjoranv or @bratseth or @hakonhall PR
This is step 1 of #24733


jonmv (approved):

container-disc/src/main/java/com/yahoo/container/jdisc/ClusterInfoProvider.java
```diff
@@ -0,0 +1,29 @@
...
+import com.yahoo.cloud.config.ClusterInfoConfig;
+import com.yahoo.component.annotation.Inject;
+import com.yahoo.component.AbstractComponent;
+import com.yahoo.container.di.componentgraph.Provider;
+
+/**
+ * Provides information about the zone in which this container is running.
+ * This is available and can be injected when running in a cloud environment.
+ *
+ * @author bratseth
```
gjoranv, line 15: Suggest to add a comment why we need a Provider for this. Also, update author.

PR #24740, opened 2022-11-04
Wire core dump encryption public key ID to core dump handler logic

vekterli:
@hakonhall / @freva please review. The redundant constructor will be removed once the internal repo is updated. Adding it now so we don't have to do any kind of synchronized merging dance.

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
@@ -5,12 +5,14 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.yahoo.config.provision.DockerImage;
+import com.yahoo.security.KeyId;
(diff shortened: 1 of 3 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -196,9 +210,16 @@ Optional<ContainerPath> enqueueCoredump(NodeAgentContext context, ContainerPath
         return Optional.of(enqueuedDir);
     }
 
+    private String corePublicKeyFlagValue(NodeAgentContext context) {
+        return coreEncryptionPublicKeyIdFlag.with(FetchVector.Dimension.NODE_TYPE, context.nodeType().name()).value();
```
freva, line 214: Where will the key ID come from when this has rolled out?

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
vekterli: Discussed this offline; key id propagation etc. will change Soon™ after some longer term design decisions have been made.

hakonhall (approved):

PR #24745, opened 2022-11-04
Balder/model importing code in config model [run-systemtest]

baldersheim:
@bratseth @hmusum @gjoranv PR
Since we no longer rely on JNI for model importing we can keep the import code in the model too, as we do for the rest.
We nede it in both places until all models without it are gone.

### DIFF
--- a/config-model-fat/pom.xml
+++ b/config-model-fat/pom.xml
@@ -37,12 +37,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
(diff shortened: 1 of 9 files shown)


bratseth (commented):
Don't we need to move this to the model and then change the API once no models are using it towards the config server?

baldersheim:
Yes, you are right. Need to add alternative methods there first.

bratseth (approved):

PR #24747, opened 2022-11-04
Read the node after taking the lock when moving to ready

freva:
Also avoids taking unallocated lock

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -189,29 +189,15 @@ public class Nodes {
         }
     }
 
-    /** Sets a list of nodes ready and returns the nodes in the ready state */
-    public List<Node> setReady(List<Node> nodes, Agent agent, String reason) {
-        try (Mutex lock = lockUnallocated()) {
-            List<Node> nodesWithResetFields = nodes.stream()
-                                                   .map(node -> {
-                                                       if (node.state() != Node.State.provisioned && node.state() != Node.State.dirty)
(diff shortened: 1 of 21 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -189,29 +189,15 @@ public List<Node> addNodes(List<Node> nodes, Agent agent) {
...
-                                                   .collect(Collectors.toList());
-            return db.writeTo(Node.State.ready, nodesWithResetFields, agent, Optional.of(reason));
-        }
-    }
+    /** Sets a node to ready and returns the node in the ready state */
+    public Node setReady(NodeMutex nodeMutex, Agent agent, String reason) {
+        Node node = nodeMutex.node();
+        if (node.state() != Node.State.provisioned && node.state() != Node.State.dirty)
+            illegal("Can not set " + node + " ready. It is not provisioned or dirty.");
+        if (!node.status().wantToDeprovision()) // Do not reset status if wantToDeprovision
```
hakonhall, line 197: This suggests it is possible to end up with a ...host in ready but with wantToDeprovision -- why do we want to allow that? I thought the idea was that with wantToDeprovision, when host-admin reports the host is ready, the config server should instead move it to parked.

Also, the check in markNodeAvailableForNewAllocation() about reasonsToFailHost must be ignored with wantToDeprovision.

hakonhall: Oh OK, you're actually not changing anything other than fixing the locking here

hakonhall (approved):

PR #24755, opened 2022-11-04
Jonmv/avoid npe

jonmv:
@freva please review and mergen


jonmv:
First commit belongs to #24752

freva (approved):

PR #24756, opened 2022-11-04
- Remove cluster() from SystemInfo to only keep static information th…

baldersheim:
…ere.

- Instead provide only the cluster name.
- If you need cluster information have Cluster injected directly.

@bratseth @gjoranv @hakonhall PR
This is the second part of #24733
Do not merge until 8.79.16 or later has been released


hakonhall (approved):
This change can only reach main AFTER stream-ranking has upgraded to at least 8.79?

gjoranv (approved):

bratseth (approved):

PR #24757, opened 2022-11-04
Set cloud account on child nodes as well

freva:
To be able to show in the Console that a node is running on a host that is in an external account without having to look up the host.


gjoranv (approved):

bratseth (approved):
👍

PR #24767, opened 2022-11-05
Revert "Jonmv/close streams and add spammy log"

jonmv:
@freva @hmusum @hakonhall


jonmv:
Nevermind, streams should be closed in any case, and just the logging should be reverted.

PR #24772, opened 2022-11-07
Revert "Revert "Revert "Revert "Move $VESPA_HOME/tmp to $VESPA_HOME/var/tmp [run-systemtest]""""

ean:
Reverts vespa-engine/vespa#24771


ean:
@aressem

dist/vespa.spec
```diff
@@ -587,6 +586,10 @@ exit 0
 %systemd_postun_with_restart vespa-configserver.service
 %endif
 
+%post base
+
+ln -s %{_prefix}/var/tmp %{buildroot}/%{_prefix}/tmp
```
aressem, line 591: This will fail on upgrade, but `ln -sf ...` should work.

aressem (changes requested):

aressem (approved):

dist/vespa.spec
```diff
@@ -586,6 +586,10 @@ exit 0
 %systemd_postun_with_restart vespa-configserver.service
 %endif
 
+%post base
+
+ln -sf %{_prefix}/var/tmp %{_prefix}/tmp
```
toregge, line 591: This might work as expected if `%{_prefix}/tmp` is a directory, since it would make a symlink from `%{_prefix}/tmp/tmp` to `%{_prefix}/var/tmp`

PR #24785, opened 2022-11-07
Revert "Remove old and now unsed archive-access endpoint"

baldersheim:
Reverts vespa-engine/vespa#24779
@ean or @freva PR
I suspect it is not completely unused yet
https://factory.vespa.oath.cloud/testrun/52765414/test/XGBoostServing::test_xgboost__ELASTIC


freva:
This just changes a REST API in the controller, which AFAIK, we never even start in system tests. I'm guessing the failure is more likely caused by some delayed reaction to moving `$VESPA_HOME/tmp`?

baldersheim:
Hmm, you are probably right.
However it was the only when test started failing between 8.80.12 and 8.80.13 that I could see.

PR #24786, opened 2022-11-07
Don't wait for maintenance redeploy if autoscaling deploy fails once

bratseth:


bratseth:
... because AutoscalingMaintainer runs every 5 minutes and MaintenanceRedeploy only once per hour.

freva (approved):

PR #24792, opened 2022-11-08
Use DocVectorAccess API when building hnsw index during load of tensor

toregge:
attribute.

@geirst : please review


geirst (approved):
👍

PR #24798, opened 2022-11-08
GC metric that has not had any meaning since we removed fdispatch.

baldersheim:
@bratseth PR


container-search/src/test/java/com/yahoo/prelude/cluster/ClusterSearcherTestCase.java
```diff
@@ -28,15 +27,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
bratseth, line 30: New IntelliJ again? :stuck_out_tongue_closed_eyes:

bratseth (approved):

container-search/src/test/java/com/yahoo/prelude/cluster/ClusterSearcherTestCase.java
baldersheim: Sigh, why can the default just be correct. #24800

PR #24812, opened 2022-11-09
Jonmv/singleton fixes

jonmv:
@hakonhall

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
@@ -310,6 +310,7 @@ class SingletonManager {
         private void cleanOrphans() {
             List<String> orphans = null;
             try {
+                // Only the ephemerals owned by this client session are listed here, and this client should only ever attempt this lock from this thread, i.e., 0 or 1 nodes. 
                 for (String orphan : orphans = curator.framework().getZookeeperClient().getZooKeeper().getEphemerals(path.getAbsolute()))
                     curator.delete(path.append(orphan));
             }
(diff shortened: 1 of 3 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
```diff
@@ -302,9 +306,20 @@ private void renewLease() {
             }
         }
 
+        private void cleanOrphans() {
+            List<String> orphans = null;
+            try {
+                for (String orphan : orphans = curator.framework().getZookeeperClient().getZooKeeper().getEphemerals(path.getAbsolute()))
+                    curator.delete(path.append(orphan));
```
jonmv, line 313: Before you ask:

1. When a new singleton is registered, potentially as a result of a reload of user components (only), it will be registered with this same singleton manager, for which there should only ever be one lock.
2. When a new singleton manager is created in the same container, as a result of curator or zookeeper config changing, the new curator will use a different client, and see a different set of ephemeral nodes.

zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
```diff
@@ -302,9 +306,20 @@ private void renewLease() {
             }
         }
 
+        private void cleanOrphans() {
```
jonmv, line 309:
```suggestion
        /** This is a work-around for a bug where Curator leaks lock nodes, through "protection mode", probably because unexpected KeeperException children are thrown? */
        private void cleanOrphans() {
```

zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
hakonhall: Leave a comment on getEphemerals()?  It only fetches the ephemerals owned by this session, which is not obvious but important.

hakonhall (approved):

PR #24814, opened 2022-11-09
Revert "Revert "Revert "Balder/model importing code in config model [run-systemtest]"""

baldersheim:
Reverts vespa-engine/vespa#24813
@bratseth PR


jonmv (commented):
👍

PR #24821, opened 2022-11-10
fail on bad base64 POST data

havardpe:
@balder please review

using existing error reporting framework rather than calling std::_Exit(1) directly from http client thread

### DIFF
--- a/fbench/src/fbench/client.cpp
+++ b/fbench/src/fbench/client.cpp
@@ -227,7 +227,12 @@ Client::run()
             const char* content = urlSource.content();
             std::string base64_decoded;
             if (_args->_usePostMode && _args->_base64Decode) {
-                base64_decoded = Base64::decode(content, cLen);
+                try {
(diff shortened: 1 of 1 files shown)


fbench/src/fbench/client.cpp
```diff
@@ -227,7 +227,12 @@ Client::run()
             const char* content = urlSource.content();
             std::string base64_decoded;
             if (_args->_usePostMode && _args->_base64Decode) {
-                base64_decoded = Base64::decode(content, cLen);
+                try {
+                    base64_decoded = Base64::decode(content, cLen);
+                } catch (...) {
+                    _status->SetError("POST request contains invalid base64 encoded data");
```
baldersheim, line 233: I think you should add the message from the exception too.

havardpe:
@baldersheim PTAL

baldersheim (approved):

PR #24831, opened 2022-11-11
Use a thread executor to fire off subcribes to multiple keys concurre…

baldersheim:
…ntly.

@gjoranv or @bjorncs PR
This cuts 4-7s from startup of all containers, metricsproxy, clustercontroller, container.
This should also cut similar amount from all our system tests.

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
+++ b/container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
@@ -54,7 +54,7 @@ public class CloudSubscriber  implements Subscriber {
         return generation;
     }
 
-    //mapValues returns a view,, so we need to force evaluation of it here to prevent deferred evaluation.
+    //mapValues returns a view, so we need to force evaluation of it here to prevent deferred evaluation.
     @Override
     public Map<ConfigKey<ConfigInstance>, ConfigInstance> config() {
         Map<ConfigKey<ConfigInstance>, ConfigInstance> ret = new HashMap<>();
--- a/container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
+++ b/container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
@@ -10,6 +10,9 @@ import com.yahoo.vespa.config.ConfigKey;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.FINE;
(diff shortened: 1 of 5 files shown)


container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
```diff
@@ -29,10 +32,16 @@ public class CloudSubscriber  implements Subscriber {
...
+    CloudSubscriber(ExecutorService executor, String name, ConfigSource configSource, Set<ConfigKey<ConfigInstance>> keys) {
         this.name = name;
         this.subscriber = new ConfigSubscriber(configSource);
-        keys.forEach(k -> handles.put(k, subscriber.subscribe(k.getConfigClass(), k.getConfigId())));
+        Map<ConfigKey<ConfigInstance>, Future<ConfigHandle<ConfigInstance>>> futureHandles = new HashMap<>();
+        keys.forEach(k -> futureHandles.put(k, executor.submit(() -> subscriber.subscribe(k.getConfigClass(), k.getConfigId()))));
+        futureHandles.forEach((k, f) -> {
+            try {
+                handles.put(k, f.get());
+            } catch (InterruptedException | ExecutionException e) {}
```
bjorncs, line 43: Do we want to ignore any exceptions thrown by submitted task (`ExecutionException`)?

container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
```diff
@@ -29,10 +32,16 @@ public class CloudSubscriber  implements Subscriber {
     // if waitNextGeneration has not yet been called, -1 should be returned
     private long generation = -1L;
 
-    CloudSubscriber(String name, ConfigSource configSource, Set<ConfigKey<ConfigInstance>> keys) {
+    CloudSubscriber(ExecutorService executor, String name, ConfigSource configSource, Set<ConfigKey<ConfigInstance>> keys) {
         this.name = name;
         this.subscriber = new ConfigSubscriber(configSource);
-        keys.forEach(k -> handles.put(k, subscriber.subscribe(k.getConfigClass(), k.getConfigId())));
+        Map<ConfigKey<ConfigInstance>, Future<ConfigHandle<ConfigInstance>>> futureHandles = new HashMap<>();
```
bjorncs, line 38: Consider using vespajlib's `CompletableFutures.allOf()`

baldersheim: Requires a rewrite that does seem straight forward.

container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
baldersheim: Some unit test started failing if i tried catching and rethrowing.
Will investigate further to see what is the better solution.

container-core/src/main/java/com/yahoo/container/di/CloudSubscriberFactory.java
```diff
@@ -24,11 +29,25 @@ public class CloudSubscriberFactory implements SubscriberFactory {
...
+    }
+
+    @Override
+    public void close() {
+        executor.shutdown();
+        try {
+            if ( ! executor.awaitTermination(10, TimeUnit.SECONDS)) {
+                executor.shutdownNow();
+            }
+        } catch (InterruptedException e) { }
```
gjoranv, line 50: Would it be appropriate to log something here?

container-core/src/main/java/com/yahoo/container/di/CloudSubscriber.java
bjorncs: Even when unwrapping the cause from `ExecutionException`?

baldersheim: Yes, then I got a test that never finished.

container-core/src/main/java/com/yahoo/container/di/CloudSubscriberFactory.java
baldersheim: Yes, I will investigate a bit more if I can unwarp on ExecutionException or log.

PR #24837, opened 2022-11-11
Make is possible to say cloud-account="default"

oyving:

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
@@ -1335,7 +1335,7 @@ public class ControllerTest {
                 .region(prodZone2.region(), "default")
                 .build();
 
-        // I don't know why this makes the test pass :(
+        // Allow use of custom account (zone 1)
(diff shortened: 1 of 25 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/ControllerTest.java
```diff
@@ -1324,6 +1324,33 @@ void testCloudAccount() {
...
+        var prodZone1 = productionUsEast3.zone();
+        var prodZone2 = productionUsWest1.zone();
+        var cloudAccount = "012345678912";
+        var application = new ApplicationPackageBuilder()
+                .cloudAccount(cloudAccount)
+                .region(prodZone1.region())
+                .region(prodZone2.region(), "default")
+                .build();
+
+        // I don't know why this makes the test pass :(
```
oyving, line 1339: @mpolden Please explain and I'll remove the comment 😄

mpolden: The `012345678912` account is used for `prodZone1` and access to do this must be granted through the feature flag.

mpolden:
Added a few commits. PTAL and merge together with internal PR.

PR #24847, opened 2022-11-14
Improve exception message grammar, javadoc

jonmv:
@freva

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
@@ -98,6 +98,29 @@ public interface VespaCurator {
      * containers in the cluster, and across all component generations.
      * <p>
      * <br>
+     * Notes to implementors:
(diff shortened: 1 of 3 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/api/VespaCurator.java
```diff
@@ -129,15 +152,16 @@ record Meta(int version) { }
      *
      *     &#064;Override
      *     public void activate() {
-     *         resource.open(); // Verify resource works here, and propagate any errors out.
+     *         try { resource.open(5, TimeUnit.SECONDS); } // Verify resource works here, and propagate any errors out.
+     *         catch {Exception e) { resource.close(); throw new RuntimeException("failed opening " + resource, e); }
```
freva, line 156:
```suggestion
     *         catch (Exception e) { resource.close(); throw new RuntimeException("failed opening " + resource, e); }
```

freva (approved):

PR #24863, opened 2022-11-15
Add 'wireguardPubKey' field to Node

gjoranv:
For now, this will always be empty.

FYI: @andreer

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/WireguardKey.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/WireguardKey.java
@@ -13,7 +13,7 @@ import java.util.regex.Pattern;
 public class WireguardKey extends PatternedStringWrapper<WireguardKey> {
 
     // See https://lists.zx2c4.com/pipermail/wireguard/2020-December/006222.html
-    private static final Pattern pattern = Pattern.compile("^[A-Za-z0-9+/]{42}[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$");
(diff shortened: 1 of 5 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
```diff
@@ -57,6 +58,9 @@ public final class Node implements Nodelike {
     private final List<TrustStoreItem> trustStoreItems;
     private final CloudAccount cloudAccount;
 
+    // Only set for enclave nodes
```
hakonhall, line 61: Nit: Make this a documentation comment.

config-provisioning/src/main/java/com/yahoo/config/provision/WireguardKey.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Wraps a Wireguard key.
+ * For security reasons, this should only be used for public keys, although private keys use the same format.
+ *
+ * @author gjoranv
+ */
+public class WireguardKey extends PatternedStringWrapper<WireguardKey> {
+
+    // See https://lists.zx2c4.com/pipermail/wireguard/2020-December/006222.html
+    private static final Pattern pattern = Pattern.compile("^[A-Za-z0-9+/]{42}[A|E|I|M|Q|U|Y|c|g|k|o|s|w|4|8|0]=$");
```
freva, line 16: Square brackets define a list of characters of which 1 should be matched, so the `|` are redundant and as it is now, it'll also allow a `|` characters in second last spot
```suggestion
    private static final Pattern pattern = Pattern.compile("^[A-Za-z0-9+/]{42}[AEIMQUYcgkosw480]=$");
```

hakonhall (approved):

PR #24880, opened 2022-11-15
Separate dynamic provisioning and host sharing

bratseth:


node-repository/src/test/java/com/yahoo/vespa/hosted/provision/autoscale/Fixture.java
```diff
@@ -139,13 +139,23 @@ public Fixture.Builder zone(Zone zone) {
...
+                                      .dynamicProvisioning(dynamicProvisioning)
+                                      .allowHostSharing(zone.cloud().allowHostSharing())
+                                      .build(),
+                                 zone.system(),
+                                 zone.environment(),
+                                 zone.region());
+            return this;
+        }
+
+        /** Set to true to behave as if hosts are provisioned dynamically. */
```
freva, line 154:
```suggestion
        /** Set to true to allow multiple nodes be provisioned on the same host. */
```

freva (approved):

PR #24888, opened 2022-11-16
Enclave access service

jonmv:
@hakonhall or @freva  please review.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
@@ -87,7 +87,7 @@ public interface ServiceRegistry {
 
     ResourceTagger resourceTagger();
 
-    EnclaveAccessService amiService();
+    EnclaveAccessService enclaveAccessService();
 
     RoleService roleService();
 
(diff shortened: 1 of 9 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
```diff
@@ -86,6 +87,8 @@ public interface ServiceRegistry {
 
     ResourceTagger resourceTagger();
 
+    EnclaveAccessService amiService();
```
hakonhall, line 90: rename to fit type

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/aws/EnclaveAccessService.java
```diff
@@ -0,0 +1,15 @@
...
+import com.yahoo.config.provision.CloudAccount;
+
+import java.util.Set;
+
+/**
+ * @author jonmv
+ */
+public interface EnclaveAccessService {
+
+    /** Makes the current AMIs available to the given accounts. */
```
hakonhall, line 12: AMIs are currently 1/3 of what needs to be fixed.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/AMISharer.java
```diff
@@ -0,0 +1,43 @@
...
+import com.yahoo.vespa.hosted.controller.tenant.Tenant;
+
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.logging.Logger;
+
+import static java.util.logging.Level.WARNING;
+
+public class AMISharer extends ControllerMaintainer {
```
hakonhall, line 14: Rename class as it's not AMIs (only)

freva (approved):

PR #24905, opened 2022-11-17
Retire nodes on wrong parent host flavor

freva:
Nodes that run on host with the wrong flavor (due to feature flag change) should be retired and migrated instead of being immediately removed from the application.


mpolden (approved):
👍

PR #24902, opened 2022-11-16
Enclaves api

gjoranv:
As can be seen, the testing was the more difficult part. Note that I kept "yahoo.com" url for nodes in "tenant account", as changing it led to unpredictable changes in node order.

PR #24908, opened 2022-11-17
Revert "Enclaves api"

jonmv:
Reverts vespa-engine/vespa#24902

@gjoranv broke our build


freva:
Fixed in internal repo

PR #24909, opened 2022-11-17
Support GPU in node specification

mpolden:
@bratseth


config-model/src/test/java/com/yahoo/config/model/provision/ModelProvisioningTest.java
```diff
@@ -1381,6 +1381,13 @@ public void testRequestingSpecificNodeResources() {
                          <resources vcpu='12' memory='10Gb' disk='30Gb' architecture='arm64'/>
                       </nodes>
                    </container>
+                   <container version='1.0' id='container2'>
+                      <nodes count='2'>
+                         <resources vcpu='4' memory='16Gb' disk='125Gb'>
+                           <gpu count='1' memory='16Gb'/>
```
baldersheim, line 1387: Given the already long list of 13 types in AWS already with a range of core numbers, I think we need a 'type'.
If count is number of cores, I think that vcpu is a better term.

mpolden: > I think we need a 'type'

I'd rather not have the exact GPU type as it may change over time and vary between clouds. Count+memory should be sufficient as there aren't that many Nvidia memory combinations.

> If count is number of cores

It's a full GPU, not cores.

We can discuss the finer points of the syntax in the next architect review.

config-provisioning/src/main/java/com/yahoo/config/provision/serialization/AllocatedHostsSerializer.java
```diff
@@ -97,6 +97,10 @@ private static void toSlime(NodeResources resources, Cursor resourcesObject) {
         resourcesObject.setString(diskSpeedKey, diskSpeedToString(resources.diskSpeed()));
         resourcesObject.setString(storageTypeKey, storageTypeToString(resources.storageType()));
         resourcesObject.setString(architectureKey, architectureToString(resources.architecture()));
+        if (!resources.gpuResources().isDefault()) {
```
bratseth, line 100: I think it would be nicer to create a sub-object for this.

bratseth (approved):
👍

PR #24910, opened 2022-11-17
Mostly test with AWS setup

bratseth:
- AWS adds the complexity of advertised vs. real which we want to test as much as possible.
- Testing with host sharing as that provuces accurate resource numbers.

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
@@ -259,8 +259,7 @@ public class NodeResources {
     @Override
     public boolean equals(Object o) {
         if (o == this) return true;
-        if ( ! (o instanceof NodeResources)) return false;
-        NodeResources other = (NodeResources)o;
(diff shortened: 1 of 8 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
```diff
@@ -333,8 +332,34 @@ public boolean satisfies(NodeResources other) {
         return true;
     }
 
-    /** Returns true if all the resources of this are the same as or compatible with the given resources */
-    public boolean compatibleWith(NodeResources other) {
+    /**
+     * Returns true if all the resources of this are the same as or compatible with the requested resources:
+     * - Equal numbers only where request implies it (i.e not for disk if storage is any/remote, and not for bandwith
```
freva, line 337:
```suggestion
     * - Equal numbers only where request implies it (i.e not for disk if storage is any/remote, and not for bandwidth
```

freva (approved):

jonmv:
Broke the build, test code PR opened in internal repo.

PR #24926, opened 2022-11-18
Support interactive private key entry when not using stdio redirection

vekterli:
@bjorncs please review

Avoids having to use a file indirection for inputting a private key. Only available when the JVM is running under an interactive console and none of the input/output files use standard streams.


vespaclient-java/src/test/resources/expected-reseal-help-output.txt
```diff
@@ -13,6 +13,10 @@ Prints new token to STDOUT.
  -i,--key-id <arg>                 ID of recipient key
  -k,--private-key-file <arg>       Private key file in Base58 encoded
                                    format
+    --no-interactive               Never ask for private key interactively
```
bjorncs, line 16: No 'short' variant of parameter name?

bjorncs (approved):

PR #24935, opened 2022-11-20
Hash to 64 bits using xxhash

bratseth:


searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
```diff
@@ -31,10 +35,15 @@ public StringValue(String value) {
...
+    /**
+     * Returns the XXHash hashcode of this, to enable strings to be encoded (with reasonable safely)
+     * as doubles for optimization.
+     */
     @Override
     public double asDouble() {
-        return value.hashCode();
+        XXHash64 hasher = XXHashFactory.fastestInstance().hash64();
+        byte[] data = value.getBytes(StandardCharsets.UTF_8);
+        return Double.longBitsToDouble(hasher.hash(data, 0, data.length, 0));
```
baldersheim, line 46: I guess we want to avoid the patterns that gives NaN as it does not compare well.
There is a suggestion in #24938 after reading the documentation for Double.longBitsToDouble

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
```diff
@@ -31,10 +35,15 @@ public StringValue(String value) {
...
 
-    /** Returns the hashcode of this, to enable strings to be encoded (with reasonable safely) as doubles for optimization */
+    /**
+     * Returns the XXHash hashcode of this, to enable strings to be encoded (with reasonable safely)
+     * as doubles for optimization.
+     */
     @Override
     public double asDouble() {
-        return value.hashCode();
+        XXHash64 hasher = XXHashFactory.fastestInstance().hash64();
```
baldersheim, line 44: Does the hash function need to be the same in both java and c++. We are using the faster XXH3_64 as default for all hashing in the backend.

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
bratseth: Ugh, ok I'll do it

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
bratseth: I don't think we strictly depend on that any more, but it would feel safer to keep them the same ...

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
baldersheim: I applied the same trick here as in the backend.

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
baldersheim: We have no tests verifying equal hashing so I think we are fine there.
The backend have standardized on XXH3, and are using that for the string repo used for tensors and various other use cases. We probably want to use this for string attributes too in the not too distant future. So it is better that we just break it now. Once XXH3 is available for java too we can start using it there too.

baldersheim (approved):

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
bratseth: XXH3 is available not just i the jpountz implementation ... trying it now.

havardpe (approved):

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
baldersheim: I guess we should see where we are using jpountz library and unify usage - later.

bratseth:
I've already done that Henning ...

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
bratseth: Yes, file distribution I think.

searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/evaluation/StringValue.java
```diff
@@ -31,10 +33,20 @@ public StringValue(String value) {
...
+     * as doubles for optimization.
+     */
     @Override
     public double asDouble() {
-        return value.hashCode();
+        // Hash using the xxh3 algorithm which is also used on content nodes
+        byte[] data = value.getBytes(StandardCharsets.UTF_8);
+        long h = LongHashFunction.xx3().hashBytes(data);
+        if ((h & 0x7ff0000000000000L) == 0x7ff0000000000000L) {
+            // Avoid nan
```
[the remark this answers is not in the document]
baldersheim: Comment updated in #24951

baldersheim:
Did not notice comment .... Cleaned up duplicates in #24951

PR #24938, opened 2022-11-20
Balder/use hash2d

baldersheim:
@havardpe PR
@bratseth FYI
Similar to #24935

### DIFF
--- a/eval/src/vespa/eval/eval/basic_nodes.h
+++ b/eval/src/vespa/eval/eval/basic_nodes.h
@@ -6,6 +6,7 @@
 #include "string_stuff.h"
 #include <vespa/vespalib/util/hdr_abort.h>
 #include <vespa/vespalib/stllike/string.h>
+#include <vespa/vespalib/stllike/hash_fun.h>
 #include <vespa/vespalib/util/string_hash.h>
 #include <memory>
 #include <map>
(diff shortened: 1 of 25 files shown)


eval/src/vespa/eval/eval/basic_nodes.h
```diff
@@ -120,9 +121,9 @@ class String : public Leaf {
 public:
     String(const vespalib::string &value_in) : _value(value_in) {}
     bool is_const_double() const override { return true; }
-    double get_const_double_value() const override { return hash(); }
+    double get_const_double_value() const override { return hash2d(_value); }
     const vespalib::string &value() const { return _value; }
-    uint32_t hash() const { return hash_code(_value.data(), _value.size()); }
+    uint64_t hash() const { return hashValue(_value.data(), _value.size()); }
```
havardpe, line 126: this needs to agree with get_const_double_value or go away

vespalib/src/vespa/vespalib/util/string_hash.cpp
```diff
@@ -1,19 +1,23 @@
...
-        hash = (hash << 5) - hash + str[i];
+double hash2d(const char *str, size_t len) {
+    size_t h = hashValue(str, len);
+    if ((h & 0x7ff0000000000000ul) == 0x7ff0000000000000ul) {
+        // Avoid nan
+        h = h & 0xffeffffffffffffful;
     }
-    return hash;
+    double d = 0;
+    memcpy(&d, &h, sizeof(d));
```
havardpe, line 15: should have static_assert that size_t is the same size as a double

havardpe (commented):
I am not sure if changing the hash function will break stuff or not. Hopefully not.

vespalib/src/vespa/vespalib/util/string_hash.cpp
baldersheim: Fixed

eval/src/vespa/eval/eval/basic_nodes.h
baldersheim: Using get_const_double_value to generate the key and removed hash method.

PR #24957, opened 2022-11-22
Fix files and directories in new location for ZooKeeper config [run-systemtest]

hmusum:
@arnej27959 please review


arnej27959 (approved):
👍

PR #24958, opened 2022-11-22
hunting for fast (compiler inlined) generators

havardpe:
- drop support for recursive generators to allow better inlining
- a bit more noexcept where possible
- 'recursive' -> 'nested' when testing generators
- drop iterator '->' operator since it is optional

@geirst please review

clang is able to inline non-recursive generators (gcc not yet). Postponing recursive generators until std::generator


geirst (approved):
👍

PR #24963, opened 2022-11-23
Allow negative GPU resources

mpolden:
Node candidate sorting may result in negative resources (e.g. when comparing
skew).

@bratseth


mpolden:
or @freva

freva (approved):

PR #24979, opened 2022-11-24
Configure VPC endpoint services for container clusters

jonmv:
@freva and perhaps @mpolden please review.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
@@ -20,6 +20,7 @@ import com.yahoo.vespa.hosted.controller.api.integration.deployment.ApplicationS
 import com.yahoo.vespa.hosted.controller.api.integration.deployment.ArtifactRepository;
 import com.yahoo.vespa.hosted.controller.api.integration.deployment.TesterCloud;
 import com.yahoo.vespa.hosted.controller.api.integration.dns.NameService;
+import com.yahoo.vespa.hosted.controller.api.integration.dns.VpcEndpointService;
 import com.yahoo.vespa.hosted.controller.api.integration.entity.EntityService;
 import com.yahoo.vespa.hosted.controller.api.integration.horizon.HorizonClient;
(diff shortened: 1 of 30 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/SharedLoadBalancerService.java
```diff
@@ -28,12 +29,14 @@ public SharedLoadBalancerService(String vipHostname) {
 
     @Override
     public LoadBalancerInstance create(LoadBalancerSpec spec, boolean force) {
+        if (spec.settings() != LoadBalancerSettings.empty) throw new IllegalArgumentException("custom load balancer settings are not supported with " + getClass());
```
mpolden, line 32: Should be `equals`, so non-empty settings always throws.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
```diff
@@ -57,6 +58,8 @@ public interface ServiceRegistry {
 
     NameService nameService();
 
+    VpcEndpointService vpcEndpointService();
```
mpolden, line 61: Consider a more generic name. `PrivateEndpointService`?

mpolden (commented):
PR title looks wrong. 🙂

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/SharedLoadBalancerService.java
jonmv: Yes, replaced with `isEmpty()` method.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/ServiceRegistry.java
jonmv: Isn't VPC "virtual private cloud", and that concept is used in other clouds too? 
It should in any case be a `...EndpointServiceService` >_<

freva (approved):

PR #24980, opened 2022-11-24
Don't apply remove to attribute if last sync token on attribute is eq…

toregge:
…ual to serialNum.

@geirst : please review
@havardpe, @baldersheim : FYI


geirst (approved):
👍

PR #24996, opened 2022-11-25
Faster controller-server unit tests

jonmv:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -40,6 +40,7 @@
 import com.yahoo.vespa.hosted.controller.versions.OsVersionTarget;
 import com.yahoo.vespa.hosted.controller.versions.VersionStatus;
 import com.yahoo.vespa.hosted.controller.versions.VespaVersion;
+import org.apache.zookeeper.data.Stat;
```
jonmv, line 43:
```suggestion
```

mpolden (approved):
👍

PR #24999, opened 2022-11-25
Make internal cluster/node state propagation noexcept

vekterli:
@geirst please review

It is not well defined what to do if an implementation of the propagation callback throws, so make it `noexcept` to core the process if it does happen.


geirst (approved):
👍

PR #25010, opened 2022-11-26
Move ArchiveStreamReader and its large dependency from vespajlib to [run-systemtest]

baldersheim:
…pplication-model.

This avoids it being pulled in to the config-model-fat that make the install larger than necessary.

@bjorncs or @jonmv PR


jonmv (approved):

jonmv (approved):

bjorncs (commented):
👍

PR #25012, opened 2022-11-26
Split out opennlp-linguistics [run-systemtest]

baldersheim:
@bjorncs PR


container-core/pom.xml
```diff
@@ -82,6 +82,31 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
bjorncs, line 85: This should probably be moved to container-search or similar. Container-core/disc should ideally only contain core jdisc functionality.

bjorncs (approved):

container-core/pom.xml
baldersheim: I agree. It is due to com.yahoo.language.provider.DefaultLinguisticsProvider in container-core.

bratseth:
Shouldn't this go in linguistics-components?

PR #25013, opened 2022-11-26
Use a simple tokenizer and replace the larger than container-search [run-systemtest]

baldersheim:
…iz.aQute.bndlib

@bjorncs PR


bjorncs (approved):
It's less risky to keep using `bnd` for parsing OSGi attributes. There is a risk that our new parser is not fully compliant.

baldersheim:
As far as I can tell the Import-package spec is not that hard. And not enough to warrant the largest 3rd. part library to be brought in.

PR #25018, opened 2022-11-28
Upgrade Curator to 5.4.0 [run-systemtest]

hmusum:


hmusum:
I don't like the last commit here, but it's the only way I found to avoid compiler warnings (which will fail the build) in controller-server module

gjoranv (approved):

PR #25022, opened 2022-11-28
Allow GCP project IDs in CloudAccount

hakonhall:


config-provisioning/src/main/java/com/yahoo/config/provision/CloudAccount.java
```diff
@@ -25,8 +29,9 @@ public boolean isUnspecified() {
 
     public static CloudAccount from(String cloudAccount) {
         return switch (cloudAccount) {
+            // TODO: Remove "default" as e.g. it is a valid GCP project ID
             case "", "default" -> empty;
-            default -> new CloudAccount(cloudAccount);
+            default -> new CloudAccount(cloudAccount, EMPTY + "|" + AWS_ACCOUNT_ID + "|" + GCP_PROJECT_ID, "cloud account");
```
freva, line 34: Do we need to allow empty here? Would be nice if the empty was a singleton created above.

freva (approved):

PR #25024, opened 2022-11-28
Use correct encoding base in test

vekterli:
@jonmv please review.

Test was written prior to base change. Wrong base was "close enough" that test seemingly worked most of the time...! 👀


baldersheim (approved):

jonmv (approved):
(((＼（✘෴✘）／)))

PR #25026, opened 2022-11-28
Revert "Use a simple tokenizer and replace the larger than container-search [run-systemtest]"

baldersheim:
Reverts vespa-engine/vespa#25013
@bjorncs PR


bjorncs (commented):
👍

PR #25029, opened 2022-11-28
Avoid mass pending GC on config enable edge [run-systemtest]

vekterli:
@geirst please review

If a system is running in a stable state with no GC enabled, per-bucket last GC timestamps in the DB will end up further and further in the past. If GC is then enabled in config, we must ensure that GC timestamps are reset to the current time to avoid suddenly ending up with _every single_ bucket having exceeded its GC deadline, causing pending GC _en masse_.

Resetting is edge-triggered, so it should not happen if GC is enabled in both the old and new configs.


geirst (approved):
👍

PR #25033, opened 2022-11-28
No need to bring in container dependencies in the feeder client.

baldersheim:
@bjorncs or @gjoranv PR


gjoranv (approved):

vespaclient-core/src/main/java/com/yahoo/feedhandler/InputStreamRequest.java
```diff
@@ -16,10 +14,10 @@
  */
```
bjorncs, line 14: Consider updating the description...

bjorncs (commented):
🎖️

PR #25037, opened 2022-11-29
Revert "Split out filedistribution client into its own module [run-systemtest]"

hmusum:
Reverts vespa-engine/vespa#25035

Fails on vespa-dependencies-enforcer


baldersheim:
Fixed in https://github.com/vespa-engine/vespa/pull/25036
I have also run system tests, working

PR #25044, opened 2022-11-29
Defer MBus DestinationSession registration during content node init [run-systemtest]

vekterli:
@geirst please review
@toregge FYI

Creating a `DestinationSession` that is immediately registered as available for business means we may theoretically start receiving messages over the session even before the call returns to the caller. Either way there would be no memory barrier that ensures that `_messageBusSession` would be fully visible to the MessageBus threads (since it's written after return).

To avoid this sneaky scenario, defer registration (and thus introduce a barrier) until _after_ we've initialized our internal member variables.

This addresses a TSan warning.


geirst (approved):
👍

PR #25045, opened 2022-11-29
Balder/reduce jdisc http httprequest infestation

baldersheim:
@bjorncs or @bratseth PR


bratseth (approved):
Nice

container-search/src/main/java/com/yahoo/search/Query.java
```diff
@@ -912,7 +913,9 @@ private void copyPropertiesTo(Query clone) {
      * Return the HTTP request which caused this query. This will never be null
      * when running with queries from the network.
      */
-     public HttpRequest getHttpRequest() { return httpRequest; }
+    public HttpRequest getHttpRequest() { return httpRequest; }
+
+    public URI getUri() { return httpRequest != null ? httpRequest.getUri() : null; }
```
bjorncs, line 918: When will `httpRequest` be null? Should the return value be `Optional<URI>`?

bjorncs (commented):
👍

container-search/src/main/java/com/yahoo/search/Query.java
baldersheim: Currently it can not be null.

bjorncs: So that will be changed at a later time?

baldersheim: Hard to tell. If a query, fx created for testing, makes sense without URI we should make it Optional.

PR #25050, opened 2022-11-30
Revert "Re-create container if wanted create command changes" MERGEOK

mpolden:
Reverts vespa-engine/vespa#24998

Probably not feasible since there's multiple subtle scenarios where the create command and actual container state are out of sync.


mpolden:
Merge with internal PR.

@freva

andreer (approved):

PR #25053, opened 2022-11-30
Add range checks for BitVector setBits and clearBits member functions,

toregge:
enabled when VESPA_BITVECTOR_RANGE_CHECK environment variable is set to "yes".

@baldersheim : please review
@geirst : FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/common/bitvector.cpp
+++ b/searchlib/src/vespa/searchlib/common/bitvector.cpp
@@ -9,8 +9,8 @@
 #include <vespa/vespalib/util/size_literals.h>
 #include <vespa/vespalib/objects/nbostream.h>
 #include <vespa/fastos/file.h>
-#include <stdlib.h>
 #include <cassert>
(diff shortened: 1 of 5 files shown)


searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -9,6 +9,7 @@
 #include <vespa/vespalib/util/size_literals.h>
 #include <vespa/vespalib/objects/nbostream.h>
 #include <vespa/fastos/file.h>
+#include <stdlib.h>
```
baldersheim, line 12: cstdlib

baldersheim (approved):

searchlib/src/vespa/searchlib/common/bitvector.cpp
toregge: Done.

baldersheim (approved):

PR #25054, opened 2022-11-30
add /opt/vespa-deps/lib64 as a location for native libraries

arnej27959:
@baldersheim please review


client/go/jvm/jdk_properties.go
```diff
@@ -13,7 +13,7 @@ func (opts *Options) AddCommonJdkProperties() {
 	secOvr := defaults.UnderVespaHome("conf/vespa/java.security.override")
 	opts.fixSpec.FixDir(tmpDir)
 	opts.AddOption("-Djava.io.tmpdir=" + tmpDir)
-	opts.AddOption("-Djava.library.path=" + libDir)
+	opts.AddOption("-Djava.library.path=" + libDir + ":/opt/vespa-deps/lib64")
```
baldersheim, line 16: Is there a way similar to defaults.UnderVespaHome to avoid hardcoding ?

baldersheim (approved):

PR #25055, opened 2022-11-30
always use "go" version of vespa-start-container-daemon

arnej27959:
@baldersheim please review


baldersheim (approved):
I prefer changing the default value of the flag in one release before removing the code using the flag.

arnej27959:
We must never, ever change the default for a flag in the flags module. In this case, doing so would trigger immediate restart of all containers in all clusters for all versions.

baldersheim:
Yes, you are right about stuff that changes the sentinel config as that will trigger a restart.

PR #25056, opened 2022-11-30
avoid race caused by using 'accepted' after destruction

havardpe:
also tag await_suspend functions with noinline to make sure clang does not put local variables into the coroutine state (this was only an issue for try_schedule).

@geirst please review


geirst (approved):
👍

PR #25057, opened 2022-11-30
Move synchronous policy init inside lock [run-systemtest]

vekterli:
@geirst please review
@toregge FYI

Avoids a race when multiple threads try to synchronously initialize a MessageBus policy.

This resolves a TSan warning.


geirst (approved):
👍

PR #25064, opened 2022-12-01
Revert "- Reduce usage of guava."

baldersheim:
Reverts vespa-engine/vespa#25062
@hmusum PR
There were more non deterministic ordering


hmusum (commented):
👍

PR #25074, opened 2022-12-01
Write new settings only on activation

jonmv:


jonmv:
Wires LB settings through session ZK data, and writes it to an LB instance only during activation. Other LB instance users retain whatever settings are already present, if any.

freva (approved):

PR #25075, opened 2022-12-01
add container-onnxruntime-bundle

arnej27959:
@gjoranv and/or @bjorncs please review
@baldersheim @lesters FYI


container-dev/pom.xml
```diff
@@ -78,6 +78,17 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
bjorncs, line 81: `container-onnxruntime-bundle` should be added to `container-dev` if an application is allowed to use onnxruntime directly without adding container-onnxruntime-bundle as direct dependency in pom.xml.

container-onnxruntime-bundle/pom.xml
```diff
@@ -0,0 +1,86 @@
+<?xml version="1.0"?>
+<!-- Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                          http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>container-onnxruntime-bundle</artifactId>
```
bjorncs, line 8: Consider removing redundant `-bundle` suffix. Modules with prefix `container-` are all bundles.

config-model/pom.xml
```diff
@@ -163,6 +163,12 @@
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
```
bjorncs, line 166: I don't think this dependency is necessary.

bjorncs (commented):
See comments

container-onnxruntime-bundle/pom.xml
gjoranv: Agree, consistent naming is always good.

container-onnxruntime-bundle/src/main/java/ai/onnxruntime/vespa/OnnxBundleActivator.java
```diff
@@ -0,0 +1,51 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.                                                                                                                           
+package ai.onnxruntime.vespa;
```
gjoranv, line 2: Why not `vespa.ai.onnxruntime`?

container-onnxruntime-bundle/src/main/java/ai/onnxruntime/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+/**
+ * @author arnej
+ */
+@ExportPackage
+package ai.onnxruntime;
```
gjoranv, line 6: Why is this package exported? It does not seem to contain any code.

container-onnxruntime-bundle/src/main/java/ai/onnxruntime/providers/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+/**
+ * @author arnej
+ */
+@ExportPackage
+package ai.onnxruntime.providers;
```
gjoranv, line 6: Why is this package exported? It does not seem to contain any code.

container-dev/pom.xml
gjoranv: We probably want it to be available by default, because the bundle is always added to the container cluster. See also the comment in `ContainerModelBuilder.addModelEvaluationBundles` above.

gjoranv (changes requested):

container-onnxruntime-bundle/src/main/java/ai/onnxruntime/vespa/OnnxBundleActivator.java
arnej27959: will put the bundle activator in "our" package namespace (ai.vespa.*)

container-onnxruntime-bundle/src/main/java/ai/onnxruntime/package-info.java
arnej27959: that's just seemingly, it exports the actual onnruntime classes.

bjorncs (approved):

gjoranv (commented):
👍

PR #25098, opened 2022-12-03
Store and share private service names

jonmv:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -1960,6 +1962,22 @@ private HttpResponse setGlobalRotationOverride(String tenantName, String applica
                                                  instance.id().toShortString(), zone, inService ? "in" : "out of"));
     }
 
+    private HttpResponse getPrivateServiceInfo(String tenantName, String applicationName, String instanceName, String environment, String region) {
+        List<LoadBalancer> lbs = controller.serviceRegistry().configServer().getLoadBalancers(ApplicationId.from(tenantName, applicationName, instanceName),
+                                                                                              ZoneId.from(environment, region));
+        Slime slime = new Slime();
+        Cursor lbArray = slime.setObject().setArray("load-balancers");
```
freva, line 1969: Use camel case like we do for the other APIs and `serviceId` in this API? Also in /nodes/v2

freva (approved):

jonmv:
Thanks, this was an oversight.

freva (approved):

PR #25113, opened 2022-12-05
Add tensor short form option to vespa-visit and vespa-get

### LINKED ISSUE #23470: Add option for short tensor format for vespa-get and vespa-visit
Vespa 8 changes http document api to use short tensor format, 
but vespa-visit and vespa-get tooling still output in long format.

vekterli:
@bjorncs please review
@jobergum FYI

Specified with `--tensor-short-form`. No single-char option alias, as short form output will be the default on Vespa 9 and we're running out of usable option characters for these tools anyway.

This resolves #23470

### DIFF
--- a/vespaclient-java/src/main/java/com/yahoo/vespaget/CommandLineOptions.java
+++ b/vespaclient-java/src/main/java/com/yahoo/vespaget/CommandLineOptions.java
@@ -39,6 +39,7 @@ public class CommandLineOptions {
     public static final String PRIORITY_OPTION = "priority";
     public static final String JSONOUTPUT_OPTION = "jsonoutput";
     public static final String XMLOUTPUT_OPTION = "xmloutput";
+    public static final String TENSOR_SHORT_FORM_OPTION = "tensor-short-form";
 
     private final Options options = createOptions();
     private final InputStream stdIn;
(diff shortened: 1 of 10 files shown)


vespaclient-java/src/main/java/com/yahoo/vespavisit/VdsVisit.java
```diff
@@ -348,6 +348,13 @@ protected static Options createOptions() {
                         FixedBucketSpaces.defaultSpace(), FixedBucketSpaces.globalSpace(), FixedBucketSpaces.defaultSpace()))
                 .build());
 
+        // TODO Vespa 9 replace with --tensor-long-form ?
+        options.addOption(Option.builder()
+                .longOpt("tensor-short-form")
```
bjorncs, line 353: Consider not using dash in parameter name (the other parameters don't)

vespaclient-java/src/main/java/com/yahoo/vespaget/CommandLineOptions.java
```diff
@@ -131,6 +132,13 @@ private static Options createOptions() {
                 .desc("XML output")
                 .longOpt(XMLOUTPUT_OPTION).build());
 
+        // TODO Vespa 9: replace with --tensor-long-form ?
+        options.addOption(Option.builder()
+                .longOpt(TENSOR_SHORT_FORM_OPTION)
```
bjorncs, line 137: See comment for vespa-visit

bjorncs (approved):

vespaclient-java/src/main/java/com/yahoo/vespavisit/VdsVisit.java
vekterli: I considered it, but I think writingoptionslikethisispointlesslyhardtoread so I went for the "be the change you want to see" approach. Save for entirely replacing the tool it's not given how we can ever transition away from a (IMHO) very suboptimal naming strategy if it cannot be done gradually.

But it might be less horrible if I rename it to something like `--shorttensors` instead, so I think I'll go with that.

vekterli: Renamed

bjorncs (approved):

PR #25118, opened 2022-12-05
Decrease lock timeout

bratseth:
**10** min -> **1** min default node repo lock timeout.


mpolden (approved):

hmusum (approved):
I think this is the right way to go, but just FYI lock timeout was increased due to internal ticket 20741

bratseth:
Ah, right. But I think we have a considerably more tenacious retry mechanism now?

PR #25127, opened 2022-12-06
Turn off FullSearch optimization for active lids blueprint createLe…

toregge:
…afSearch().

@geirst : please review

This replaces #25119


geirst (approved):
👍

geirst:
The FullSearch optimization doesn't always work when the query doesn't have any constraints, e.g. "select * from mycluster where true" after removing all documents (as seen in https://github.com/vespa-engine/vespa/issues/25108). This is fixed by this PR. The optimization is now only used when evaluating a pre-filter used in combination with ANN search.

PR #25129, opened 2022-12-06
add and use util.Array

arnej27959:
@mpolden please review

### DIFF
--- a/client/go/util/array.go
+++ b/client/go/util/array.go
@@ -1,0 +1,89 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// Author: arnej
+
+// generic utilities
+package util
+
+type Array[E comparable] []E
+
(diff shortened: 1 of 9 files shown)


client/go/util/array.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (arrayP *Array[E]) Append(elem E) {
+	*arrayP = append(*arrayP, elem)
+}
+
+func (arrayP *Array[E]) AppendElements(elems ...E) {
+	arrayP.AppendAll(elems)
+}
+
+func (arrayP *Array[E]) AppendAll(elemsToAppend []E) {
```
mpolden, line 25: Do we need both? If you have a slice you can unpack it to args, e.g. `AppendElements(myslice...)`

arnej27959:
good points, please take another look now

mpolden (approved):
👍

PR #25135, opened 2022-12-06
test more widely

arnej27959:
@hakonhall please review


hakonhall (approved):
This will test in main us-west-1 too?

arnej27959:
> This will test in main us-west-1 too?
> 

Yes, that is the intent.

PR #25143, opened 2022-12-07
Support not specifying vcpu, memoryGb and diskGb for ClusterCapacity

hmusum:


flags/src/test/java/com/yahoo/vespa/flags/custom/ClusterCapacityTest.java
```diff
@@ -38,4 +39,29 @@ void serialization2() throws IOException {
...
+        assertEquals("{\"count\":7}", json);
+
+        ClusterCapacity deserialized = mapper.readValue(json, ClusterCapacity.class);
+        assertEquals(0.0, deserialized.vcpu(), 0.0001);
+        assertEquals(0.0, deserialized.memoryGb(), 0.0001);
+        assertEquals(0.0, deserialized.diskGb(), 0.0001);
+        assertEquals(1.0, deserialized.bandwidthGbps(), 0.0001);  // 1.0 is used as fallback
+        assertEquals(7, deserialized.count());
+
+        // Test that no values for node resources is allowed in json also
```
freva, line 57: The second below seems to be the same as the above, `json == jsonWithNoValuesForNodeResources == {"count":7}`, so both the `deserialized` are identical.

freva (approved):
Nice

flags/src/test/java/com/yahoo/vespa/flags/custom/ClusterCapacityTest.java
hmusum: Yes, you are right, went back and forth and forgot to remove this. Fixing

freva (approved):

PR #25147, opened 2022-12-07
fix unintended change

arnej27959:
@mpolden or @baldersheim please review and merge


baldersheim (changes requested):
It has merge conflicts

PR #25148, opened 2022-12-07
Have to switch to python39-pybind11-devel. pybind11-devel has been re…

aressem:
…moved from EPEL and not added to powertools for EL8.

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -404,7 +404,7 @@ Requires: %{name}-base-libs = %{version}-%{release}
 Requires: %{name}-libs = %{version}-%{release}
 %if 0%{?el8}
 %if 0%{?_centos_stream}
-BuildRequires: python39
+Requires: python39
 %else
 Requires: python36
 %endif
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -52,19 +52,21 @@ BuildRequires: gcc-toolset-11-gcc-c++
 BuildRequires: gcc-toolset-11-binutils
 BuildRequires: gcc-toolset-11-libasan-devel
 BuildRequires: gcc-toolset-11-libatomic-devel
+BuildRequires: python39-pybind11-devel
+BuildRequires: python39-devel
```
toregge, line 56: The requires for vespa-ann-benchmark subpackage needs adjustment too.

toregge (changes requested):

dist/vespa.spec
aressem: Adjusted

toregge (approved):

PR #25164, opened 2022-12-08
Add more xxe preventions

tokle:


hmusum (approved):

bratseth:
I think you should unify this with com.yahoo.text.XML.getDocumentBuilder
by splitting that method in two.

PR #25166, opened 2022-12-08
Support diskSpeed, storageType and architecture in ClusterCapacity

hmusum:


freva (approved):
Consider adding some validation of the string input like we have for `HostResources`: https://github.com/vespa-engine/vespa/blob/43680c094c227807695a67f8a2c5d7206dd1c98f/flags/src/main/java/com/yahoo/vespa/flags/custom/HostResources.java#L21

Then a typo on these fields would be caught in the feature-flag change PR.

hmusum:
Yup, good idea, PTAL

freva (approved):

PR #25173, opened 2022-12-08
Add metric tracking internal `StateManager` state propagation latency

vekterli:
@geirst please review

Adds a new top-level metric set with the single metric
```
vds.state_manager.invoke_state_listeners_latency
```
which tracks the latency for invoking the set of all state listeners.


geirst (approved):
👍

PR #25176, opened 2022-12-08
Improve error response when trying to use an imported field in a condition

vekterli:
@geirst please review
@jonmv FYI

We don't support using imported fields in conditional mutations, so catch attempts at doing this during the field enumeration that is done as part of the condition evaluation. Would previously get an internal error response with an ugly stack trace since the exception would propagate up to a generic exception-to-response handler.

Will now generate an `ILLEGAL_PARAMETERS` error response with a hopefully helpful error message.


geirst (approved):
👍

PR #25179, opened 2022-12-08
Move sessionmanager from documentdb to proton and use a SchedulerForw…

baldersheim:
…ardExecutor to drive pruning instead of a maintenance job.

@geirst PR


geirst (approved):
👍

PR #25190, opened 2022-12-09
Return a handle that will deregister the recurring task when it goes …

baldersheim:
…out of scope.

This requires the user to consider lifetime, and allows for finegrained control.

@geirst and @toregge PR


geirst (approved):
👍 
@toregge please review as well

toregge (approved):

PR #25195, opened 2022-12-09
Avoid refering an empty vector.

baldersheim:
@toregge PR


searchlib/src/vespa/searchlib/expression/resultnodes.cpp
```diff
@@ -213,7 +213,10 @@ RawResultNode::min(const ResultNode & b)
     char buf[32];
     ConstBufferRef s(b.getString(BufferRef(buf, sizeof(buf))));
 
-    if (memcmp(&_value[0], s.data(), std::min(s.size(), _value.size())) > 0)  {
+    size_t min_sz = std::min(s.size(), _value.size());
+    if (min_sz == 0) {
+        setBuffer("", 0);
```
toregge, line 218: If _value.size() is 0 then setBuffer() should not be needed.

toregge (approved):

PR #25200, opened 2022-12-09
Disallow exclusions in access-control

tokle:


config-model/src/test/java/com/yahoo/vespa/model/application/validation/AccessControlFilterExcludeValidatorTest.java
```diff
@@ -0,0 +1,108 @@
+package com.yahoo.vespa.model.application.validation;
```
bjorncs, line 1: Missing copyright header and author

bjorncs (commented):
Validator is not enabled - intentional?

bjorncs (approved):

PR #25209, opened 2022-12-12
extend lifetime to end of next year

arnej27959:
for new vespa versions we won't use this feature flag, but we need to get rid of Vespa 7 before final cleanup.


arnej27959:
@hmusum please review

hmusum (approved):

PR #25218, opened 2022-12-12
Keep the original task in a shared_ptr in the wrapping lambda and mak…

baldersheim:
…e a copy of it and pass to the actual task diaptch on another executor to ensure lifetime.

@geirst or @toregge PR


geirst (approved):
👍

PR #25229, opened 2022-12-13
Revert "Revert "Use the forward scheduler in proton also for disk-mem-util-sampler"" 

baldersheim:
Reverts vespa-engine/vespa#25228


baldersheim:
@geirst or @toregge PR

toregge (approved):

PR #25236, opened 2022-12-13
Arnej/no dependency reduced pom

arnej27959:
@gjoranv please review
@bjorncs @baldersheim FYI


arnej27959:
note: needs to be merged at the same time as PR in another repo.

gjoranv (approved):

PR #25247, opened 2022-12-13
support for disablerepo in yum command

bjormel:

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
@@ -35,6 +35,7 @@ public abstract class YumCommand<T extends YumCommand<T>> {
             builder::setName, builder::setEpoch, builder::setVersion, builder::setRelease, builder::setArchitecture);
 
     private List<String> enabledRepos = List.of();
+    private List<String> disabledRepos = List.of();
     private final Terminal terminal;
 
     protected YumCommand(Terminal terminal) {
@@ -47,11 +48,18 @@ public abstract class YumCommand<T extends YumCommand<T>> {
         return getThis();
     }
 
+    /** Enables the given repos for this command */
+    public T disableRepo(String... repo) {
+        disabledRepos = List.of(repo);
+        return getThis();
+    }
+
     protected abstract T getThis(); // Hack to get around unchecked cast warning
 
     protected void addParametersToCommandLine(CommandLine commandLine) {
         commandLine.add("--assumeyes");
         enabledRepos.forEach(repo -> commandLine.add("--enablerepo=" + repo));
+        disabledRepos.forEach(repo -> commandLine.add("--disablerepo=" + repo));
     }
 
     public abstract boolean converge(TaskContext context);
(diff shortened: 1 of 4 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
```diff
@@ -47,11 +48,18 @@ public T enableRepo(String... repo) {
         return getThis();
     }
 
+    /** Enables the given repos for this command */
+    public T disableRepo(String... repo) {
```
hakonhall, line 52: Consider special-casing an empty list, to disable all (add List.of("*")).

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
```diff
@@ -47,11 +48,18 @@ public T enableRepo(String... repo) {
...
+        disabledRepos = List.of(repo);
+        return getThis();
+    }
+
     protected abstract T getThis(); // Hack to get around unchecked cast warning
 
     protected void addParametersToCommandLine(CommandLine commandLine) {
         commandLine.add("--assumeyes");
         enabledRepos.forEach(repo -> commandLine.add("--enablerepo=" + repo));
+        disabledRepos.forEach(repo -> commandLine.add("--disablerepo=" + repo));
```
hakonhall, line 62: Move disableRepos before enableRepos.  I have only seen this order, and don't know if this order will work.  Unless you know it's OK to put disable last?

hakonhall (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
bjormel: nice!

bjormel: The `"*"` triggers escaping , which leads to `yum install --assumeyes "--disablerepo=*" --enablerepo...`, I guess that it will cause the command to fail?

bjormel: Tested with escaped, and it works
```
sudo yum install "--disablerepo=*" --enablerepo
...
Complete!
```

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/yum/YumCommand.java
```diff
@@ -34,13 +34,20 @@
...
     private List<String> enabledRepos = List.of();
     private final Terminal terminal;
 
     protected YumCommand(Terminal terminal) {
         this.terminal = terminal;
     }
 
+    /** Disables the given repos for this command */
+    public T disableRepo(String... repo) {
+        disabledRepos = List.of(repo);
```
hakonhall, line 47: I was thinking

```
disabledRepos = repo.length == 0 ? List.of("*") : List.of(repo)
```
and then only do the 
```
 disabledRepos.forEach(repo -> commandLine.add("--disablerepo=" + repo));
```
below.  But this works too.

hakonhall (approved):
The argument `--disablerepo=*` is escaped in the TestTerminal's command string because it contains `*`.  It it actually not escaped by TerminalImpl when the yum program is invoked.

PR #25260, opened 2022-12-14
Keep a static representation of strings for low indexes [0...999].

baldersheim:
This will avoid blowing up the heap with many duplicate strings.

@bratseth or @lesters or @bjorncs PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
@@ -13,6 +13,21 @@ import java.util.stream.Collectors;
  * @author bratseth
  */
 public abstract class TensorAddress implements Comparable<TensorAddress> {
+    private static final String [] SMALL_INDEXES = createSmallIndexesAsStrings(1000);
+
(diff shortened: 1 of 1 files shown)


vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
```diff
@@ -13,6 +13,21 @@
  * @author bratseth
  */
 public abstract class TensorAddress implements Comparable<TensorAddress> {
+    private static final String [] SMALL_INDEXES = createSmallIndexesAsStrings(1000);
```
bjorncs, line 16: Space between `String` and `[]` looks very much like a C programmer coding Java ;)

bjorncs (approved):

vespajlib/src/main/java/com/yahoo/tensor/TensorAddress.java
baldersheim: :)

kkraune: priceless

bratseth (commented):
👍

PR #25264, opened 2022-12-15
Fix check for file reference being in use

hmusum:
Do this in a way that actually works.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDirectory.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDirectory.java
@@ -9,6 +9,7 @@ import com.yahoo.concurrent.Locks;
 import com.yahoo.config.FileReference;
 import com.yahoo.io.IOUtils;
 import com.yahoo.text.Utf8;
+import com.yahoo.vespa.config.server.ApplicationRepository;
 import com.yahoo.vespa.defaults.Defaults;
 import com.yahoo.vespa.flags.BooleanFlag;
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDirectory.java
```diff
@@ -127,13 +127,13 @@ public FileReference addFile(File source) throws IOException {
...
     }
 
-    public void delete(FileReference fileReference, Function<FileReference, Boolean> canBeDeleted) {
+    public void delete(FileReference fileReference, ApplicationRepository applicationRepository) {
         if (useLock.value())
             try (Lock lock = locks.lock(fileReference)) {
-                if (canBeDeleted.apply(fileReference))
-                    deleteDirRecursively(destinationDir(fileReference));
-                else
+                if (applicationRepository.isFileReferenceInUse(fileReference))
```
baldersheim, line 133: It was reversed ?

baldersheim (approved):

baldersheim (changes requested):
Requires a fix in configserver too, build not happy.

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDirectory.java
hmusum: No, expression was evaluated immediately, not when calling `apply`.

hmusum:
A seconds commit was missing

hmusum:
@baldersheim please take a look again

PR #25280, opened 2022-12-16
Fix dimension hashing bug

gjoranv:
FYI: @hakonhall

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
@@ -38,6 +38,8 @@ public class MetricsParser {
     static void parse(InputStream data, Consumer consumer) throws IOException {
         parse(jsonMapper.createParser(data), consumer);
     }
+
+    // Top level 'metrics' object, with e.g. 'time', 'status' and 'metrics'.
     private static void parse(JsonParser parser, Consumer consumer) throws IOException {
         if (parser.nextToken() != JsonToken.START_OBJECT) {
             throw new IOException("Expected start of object, got " + parser.currentToken());
@@ -55,6 +57,7 @@ public class MetricsParser {
             }
         }
     }
+
     static private Instant parseSnapshot(JsonParser parser) throws IOException {
         if (parser.getCurrentToken() != JsonToken.START_OBJECT) {
             throw new IOException("Expected start of 'snapshot' object, got " + parser.currentToken());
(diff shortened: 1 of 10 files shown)


hakonhall (approved):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
```diff
@@ -142,25 +143,42 @@ static private void handleValue(JsonParser parser, Instant timestamp, Consumer c
...
+                                                        dim -> toDimensionId(dim.id), dim -> dim.value)));
+    }
+
+    static long dimensionsHashCode(List<Dimension> dimensions) {
+        int keyHash = 0;
+        int valueHash = 0;
+
+        for (Dimension dim : dimensions) {
+            keyHash += dim.id.hashCode();
+            valueHash += dim.value.hashCode();
```
baldersheim, line 172: Very good catch @gjoranv. This was way to naive.
We can not treat key and value separately.
I suggest using com.yahoo.compress.Hasher.xxh3 and just keep one 64bit.
Either that or follow @hakonhall's suggestion.

metrics-proxy/src/test/java/ai/vespa/metricsproxy/service/MetricsParserTest.java
```diff
@@ -24,6 +25,21 @@ public void consume(Metric metric) {
...
+        var dimensions1 = List.of(
+                new MetricsParser.Dimension("cluster", "CLUSTER-1"),
+                new MetricsParser.Dimension("clusterid", "CLUSTER-1"));
+
+        var dimensions2 = List.of(
+                new MetricsParser.Dimension("cluster", "CLUSTER-2"),
+                new MetricsParser.Dimension("clusterid", "CLUSTER-2"));
+
+        System.out.println(dimensionsHashCode(dimensions1));
+        System.out.println(dimensionsHashCode(dimensions2));
```
baldersheim: I added test and also removed the 2 System.out.println lines assuming they were debug.

baldersheim (changes requested):
Either a set as suggested by @hakonhall or 
hash += Hasher.xx3(dim.id.getBytes(UTF8)) + Hasher.xx3(dim.value.getBytes(UTF8))

baldersheim (commented):
Looking at the code and the history the key point was to detect unique Dimension sets.
And using a set seems a lot safer.

baldersheim (approved):

gjoranv:
👍

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
```diff
@@ -142,25 +143,42 @@ static private void handleValue(JsonParser parser, Instant timestamp, Consumer c
...
             }
         }
-        Long uniqueKey = (((long) keyHash) << 32) | (valueHash & 0xffffffffL);
-        return uniqueDimensions.computeIfAbsent(uniqueKey, key -> dims.stream().collect(Collectors.toUnmodifiableMap(e -> toDimensionId(e.getKey()), Map.Entry::getValue)));
+        return uniqueDimensions.computeIfAbsent(dimensionsHashCode(dimensions),
+                                                key -> dimensions.stream().collect(Collectors.toUnmodifiableMap(
+                                                        dim -> toDimensionId(dim.id), dim -> dim.value)));
+    }
+
+    static long dimensionsHashCode(List<Dimension> dimensions) {
```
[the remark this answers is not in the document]
baldersheim: Yes, it could

baldersheim:
A benchmark after that change show parse time increased around 15-20% when using set. No effect on the memory footprint which is the most important measure here.

PR #25281, opened 2022-12-16
Now all threads have a fixed stack and there is no use in creating an…

baldersheim:
…d extra thread with larger stack.

No longer possible, so just do the work yourself.

@toregge or @havardpe or @arnej27959 PR


havardpe (commented):
does this mean that Darwin compilation is no longer possible?

baldersheim:
@arnej27959 fixed the root cause some time ago.
And @toregge has not alerted about this not working after control of stack size was removed.
So my conclusion is that we are all clear here.

PR #25282, opened 2022-12-16
Avoid flushing components in proton more than necessary.

geirst:
Don't consider TLSSIZE ordering if there exists an active (ongoing) flush (for the same flush handler) that started before the last flush time of the flush target to evaluate. Instead we should wait for the active (ongoing) flush to be finished before doing another evaluation.

@baldersheim please review
@toregge FYI

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/flushengine/active_flush_stats.h
+++ b/searchcore/src/vespa/searchcore/proton/flushengine/active_flush_stats.h
@@ -1,0 +1,32 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include <vespa/vespalib/stllike/hash_map.h>
+#include <vespa/vespalib/util/time.h>
+#include <optional>
+
+namespace proton::flushengine {
+
+/**
+ * Tracks the oldest start time of active (ongoing) flushes in each flush handler.
+ */
+class ActiveFlushStats {
+public:
+    using OptionalTime = std::optional<vespalib::system_time>;
+
+private:
+    using StatsMap = vespalib::hash_map<vespalib::string, vespalib::system_time>;
+    StatsMap _stats;
+
+public:
+    ActiveFlushStats();
+    /**
+     * Set the start time for a flush in the given flush handler.
+     * A start time is only updated if it is older than the current oldest one.
+     */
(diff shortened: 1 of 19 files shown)


searchcore/src/vespa/searchcore/proton/flushengine/flushengine.h
```diff
@@ -23,31 +23,33 @@ class FlushEngine final : public FastOS_Runnable
 public:
     class FlushMeta {
     public:
-        FlushMeta(const vespalib::string & name, uint32_t id);
+        FlushMeta(const vespalib::string& handler_name, const vespalib::string& target_name, uint32_t id);
         ~FlushMeta();
         const vespalib::string & getName() const { return _name; }
+        const vespalib::string& handler_name() const { return _handler_name; }
         vespalib::system_time getStart() const { return vespalib::system_clock::now() - std::chrono::duration_cast<vespalib::system_time::duration>(elapsed()); }
```
baldersheim, line 30: rename getStart -> getStartInEpoch and make repurpose getStart to return steady_time.

searchcore/src/vespa/searchcore/proton/flushengine/active_flush_stats.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Tracks the oldest start time of active (ongoing) flushes in each flush handler.
+ */
+class ActiveFlushStats {
+public:
+    using OptionalTime = std::optional<vespalib::system_time>;
+
+private:
+    using StatsMap = vespalib::hash_map<vespalib::string, vespalib::system_time>;
```
baldersheim, line 18: use vespalib::steady_time instead. That is a lot safer than using system_time.

baldersheim (changes requested):
I would prefer using steady_time over system_time. We should try to avoid that.

geirst:
Discussed with @baldersheim offline. We decided to change to using vespalib::steady_time in a follow-up PR as it involves changing all implementations of IFlushTarget (getLastFlushTime).

baldersheim (approved):
Using steady_time should be done in separate PR as it requires some more changes.

geirst:
After further investigation we found that changing to steady_time is not feasible as the implementations of IFlushTarget::getLastFlushTime() use time stamps on disk files (system_time).

PR #25284, opened 2022-12-17
Always --disablerepo * when enablerepo

bjormel:


hakonhall:
Some/most repos are disabled by default, and needs to be enabled when installing packages from them.  But those packages may still depend on packages from default-enabled repos, which this PR will break.  You want to hunt down all those cases?

An implicit --disablerepo with --enablerepo is counter-intuitive IMO, because YumCommand is mostly 1:1 with the `yum` command.  So whatever options you pass to yum you should pass to YumCommand, and vice versa.

bjormel:
So much that I do not know :/

bjormel:
Back to ##25279 then

PR #25285, opened 2022-12-18
Clean up /opt/vespa/var/tmp

bjormel:


freva (approved):
Why is this limited to containers only?

bjormel:
> Why is this limited to containers only?
Just me not reading surounding code 🙄

freva (approved):

PR #25289, opened 2022-12-19
Avoid PTR verification when provisioning in Enclave

hakonhall:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/testutils/MockNameResolver.java
```diff
@@ -73,9 +76,23 @@ public Set<String> resolveAll(String name) {
 
     @Override
     public Set<String> resolve(String name, RecordType first, RecordType... rest) {
-        return resolveAll(name);
+        Set<RecordType> types = new HashSet<>(1 + rest.length);
+        types.add(first);
+        Collections.addAll(types, rest);
```
freva, line 81: Use `EnumSet.of(first, rest)` ?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/testutils/MockNameResolver.java
```diff
@@ -73,9 +76,23 @@ public Set<String> resolveAll(String name) {
...
+                .stream()
+                .filter(addressString -> {
+                    if (types.contains(RecordType.A) && IP.isV4(addressString)) return true;
+                    if (types.contains(RecordType.AAAA) && IP.isV6(addressString)) return true;
+                    return false;
+                })
+                .collect(Collectors.toSet());
     }
 
+    private static boolean isIpv6(String ipAddress) { return ipAddress.contains(":"); }
```
freva, line 93: Not used?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/IP.java
```diff
@@ -415,6 +426,27 @@ private static Allocation ofIpv4(String ipAddress, NameResolver resolver) {
             return new Allocation(hostname4, Optional.of(addresses.get(0)), Optional.empty());
         }
 
+        private static Allocation fromAddress(Address address, NameResolver resolver, IpAddresses.Protocol protocol) {
+            Optional<String> ipv4Address = resolveOptional(address.hostname(), resolver, RecordType.A);
+            if (protocol != IpAddresses.Protocol.ipv6 && ipv4Address.isEmpty())
```
freva, line 431: Avoid resolving `A` records in the first place if protocol is `IPv6`?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/testutils/MockNameResolver.java
hakonhall: Great, exactly what I was looking for.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/testutils/MockNameResolver.java
hakonhall: Removed

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/IP.java
hakonhall: :thinking: or, verify no A records are found for a supposedly IPv6-only host?  We expect later resolutions in host admin or various tools to blindly find all A and AAAA records, so let's ensure they are exactly as expected?  Implemented the latter.

freva (approved):

PR #25301, opened 2022-12-20
Balder/allow multiple concurrent requests [run-systemtest]

baldersheim:
@havardpe or @geirst PR
Only the second commit differs from #25292


config/src/vespa/config/frt/frtsource.cpp
```diff
@@ -76,7 +76,7 @@ FRTSource::getConfig()
 void
 FRTSource::erase(FRT_RPCRequest * request) {
     std::lock_guard guard(_lock);
-    _inflight.erase(request);
+    assert(1 == _inflight.erase(request));
```
havardpe, line 79: you should never put actual code inside an assert

baldersheim: Ouch..

geirst (commented):
👍

PR #25303, opened 2022-12-20
Balder/make frtconnection thread safe [run-systemtest]

baldersheim:
@havardpe or @geirst PR
@toregge FYI

### DIFF
--- a/config/src/vespa/config/frt/frtconnection.cpp
+++ b/config/src/vespa/config/frt/frtconnection.cpp
@@ -14,19 +14,21 @@ namespace config {
 
 FRTConnection::FRTConnection(const vespalib::string& address, FRT_Supervisor& supervisor, const TimingValues & timingValues)
     : _address(address),
+      _transientDelay(timingValues.transientDelay),
       _fatalDelay(timingValues.fatalDelay),
       _supervisor(supervisor),
+      _lock(),
       _target(0),
       _suspendedUntil(),
       _suspendWarned(),
       _transientFailures(0),
-      _fatalFailures(0),
-      _transientDelay(timingValues.transientDelay)
+      _fatalFailures(0)
 {
 }
 
 FRTConnection::~FRTConnection()
 {
+    std::lock_guard guard(_lock);
     if (_target != nullptr) {
         LOG(debug, "Shutting down %s", _address.c_str());
         _target->SubRef();
@@ -37,6 +39,7 @@ FRTConnection::~FRTConnection()
 FRT_Target *
 FRTConnection::getTarget()
 {
+    std::lock_guard guard(_lock);
     if (_target == nullptr) {
         _target = _supervisor.GetTarget(_address.c_str());
     } else if ( ! _target->IsValid()) {
@@ -78,6 +81,7 @@ FRTConnection::setError(int errorCode)
 
 void FRTConnection::setSuccess()
 {
+    std::lock_guard guard(_lock);
     _transientFailures = 0;
     _fatalFailures = 0;
     _suspendedUntil = steady_time();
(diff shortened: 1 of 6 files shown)


config/src/vespa/config/frt/frtconnection.cpp
```diff
@@ -78,41 +81,29 @@ FRTConnection::setError(int errorCode)
...
+        delay = std::min(6u, ++_transientFailures) * _transientDelay;
         LOG(warning, "Connection to %s failed or timed out", _address.c_str());
         break;
     case FATAL:
-        _fatalFailures.fetch_add(1);
-        delay = _fatalFailures.load(std::memory_order_relaxed) * getFatalDelay();
-        if (delay > getMaxFatalDelay()) {
-            delay = getMaxFatalDelay();
-        }
+        delay = std::min(6u, ++_fatalFailures) * _fatalDelay;
```
geirst, line 101: What is the magic constant 6? Consider giving it a name.

geirst (approved):
👍

havardpe (commented):
Consider giving out a target with an independent reference to avoid use after delete (return value from getTarget vs other call to getTarget or the destructor deleting the previously returned target).

config/src/vespa/config/frt/frtconnection.cpp
```diff
@@ -14,19 +14,21 @@ namespace config {
...
-      _fatalFailures(0),
-      _transientDelay(timingValues.transientDelay),
-      _fatalDelay(timingValues.fatalDelay)
+      _fatalFailures(0)
 {
 }
 
 FRTConnection::~FRTConnection()
 {
+    std::lock_guard guard(_lock);
```
havardpe, line 31: this does not seem like it helps...
(any conflicting thread that gets the lock after us will try to lock the mutex after it has been destructed)

config/src/vespa/config/frt/frtconnection.cpp
```diff
@@ -78,41 +81,29 @@ FRTConnection::setError(int errorCode)
...
+    steady_time now = steady_clock::now();
+    std::lock_guard guard(_lock);
     switch(type) {
     case TRANSIENT:
-        _transientFailures.fetch_add(1);
-        delay = _transientFailures.load(std::memory_order_relaxed) * getTransientDelay();
-        if (delay > getMaxTransientDelay()) {
-            delay = getMaxTransientDelay();
-        }
+        delay = std::min(6u, ++_transientFailures) * _transientDelay;
```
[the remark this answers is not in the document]
baldersheim: MAX_DELAY_MULTIPLIER

config/src/vespa/config/frt/frtconnection.cpp
baldersheim: MAX_DELAY_MULTIPLIER

config/src/vespa/config/frt/frtconnection.cpp
baldersheim: That was only intended to ensure thread visibility.

baldersheim:
The target never escapes the class. getTarget is private and is only called from invoke(). So I think that should be fine.
FRTConnection lifetime should be handled at a higher level.

config/src/vespa/config/frt/frtconnection.cpp
havardpe: If you need to ensure thread visibility here, it kind of implies that there is no synchronization between using the object and deleting it, which could lead to locking the deleted mutex. In other words; if the locking was added to make TSAN happy, this will probably not fix the race, just hide it.

havardpe:
if the target never escapes the class and there is no chance of 'someone else' calling getTarget (invoke) to delete our target under our feet; why do we need a lock there in the first place...?

PR #25306, opened 2022-12-20
Add an eof object that can be sent to the Q to wake up the ones waiti…

baldersheim:
…ng for config.

This enables faster close.

@hmusum PR
This makes container shutdown faster and more deterministic.


hmusum (approved):
Great!

PR #25308, opened 2022-12-20
Revert "Add an eof object that can be sent to the Q to wake up the ones waiti…"

baldersheim:
Reverts vespa-engine/vespa#25306
Revert and see if this is causing system test issues.


hmusum:
👍

PR #25316, opened 2022-12-21
Reduce NodeFailer activate timeout

hakonhall:


freva (commented):
Ideally there should be a way for `ExpeditedChangeApplicationMaintainer` or some other maintainer to quickly pick up this application and retry the deployment. With this PR, in dynamically provisioned zones, it'll almost always timeout due to the need to wait for the newly provisioned host to come up.
If the deployment is not retried quickly enough, the reserved node will be expired, making the newly provisioned host no longer needed, thus taking us back to square 1.

hakonhall:
The NodeFailer interval is 7m so it should be retried in minutes in normal circumstances.

freva (approved):

PR #25319, opened 2022-12-21
park expirer

bjormel:

Strongly inspired by the https://github.com/vespa-engine/vespa/blob/master/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/FailedExpirer.java

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
@@ -1,0 +1,90 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.provision.maintenance;
+
+import com.yahoo.config.provision.Environment;
+import com.yahoo.config.provision.NodeType;
+import com.yahoo.config.provision.Zone;
+import com.yahoo.jdisc.Metric;
+import com.yahoo.vespa.hosted.provision.Node;
+import com.yahoo.vespa.hosted.provision.NodeList;
+import com.yahoo.vespa.hosted.provision.NodeRepository;
+import com.yahoo.vespa.hosted.provision.node.Agent;
+import com.yahoo.vespa.hosted.provision.node.History;
+import com.yahoo.vespa.hosted.provision.node.Report;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+import java.util.logging.Logger;
+
+import static com.yahoo.vespa.hosted.provision.maintenance.FailedExpirer.maxAllowedFailures;
+
+/**
+ * This moves expired parked nodes:
+ *
+ * - To dirty: if it does not have a hardware failure.
+ * - Otherwise the node will remain in parked.
+ *
+ * Nodes with detected hardware issues will not be recycled.
+ *
+ * @author bjormel
+ */
+public class ParkedExpirer extends NodeRepositoryMaintainer {
+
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,90 @@
...
+        for (Node candidate : nodes) {
+            if (candidate.type().isHost()
+                && (candidate.status().failCount() >= maxAllowedFailures
+                   || candidate.reports().getReports().stream().anyMatch(report -> report.getType().equals(Report.Type.HARD_FAIL)))) {
+                log.info(String.format("Not recycling due to hardware failure or max allowed failures: %s", candidate.hostname()));
+            } else {
+                nodesToRecycle.add(candidate);
+            }
+        }
+        nodeRepository.nodes().deallocate(nodesToRecycle, Agent.ParkedExpirer, "Expired by ParkedExpirer");
```
freva, line 87: Note that `deallocate()` will `park` the node instead of `dirty` if `Nodes::parkOnDeallocationOf`, so this class should probably also check with that method to avoid infinite loops.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    protected double maintain() {
+        NodeList allNodes = nodeRepository.nodes().list();
+        List<Node> remainingNodes = new ArrayList<>(allNodes.state(Node.State.parked)
+                .nodeType(NodeType.tenant, NodeType.host)
+                .asList());
+
+        recycleIf(node -> node.allocation().isEmpty(), remainingNodes);
```
freva, line 61: Not sure if this is warranted, just expire after the duration always?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,90 @@
...
+        recycleIf(node -> node.allocation().isPresent() &&
```
freva, line 62: You also want to make sure that neither `wantToRebuild` nor `wantToDeprovision` is set, otherwise this host may already be in the process of being deprovisioned. (And `Nodes::deallocate` with throw an exception)

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/ParkedExpirer.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Logger log = Logger.getLogger(ParkedExpirer.class.getName());
+
+    private final NodeRepository nodeRepository;
+    private final Duration expiry; // Grace period to allow operators to inspect
+
+    ParkedExpirer(NodeRepository nodeRepository, Zone zone, Duration interval, Metric metric) {
+        super(nodeRepository, interval, metric);
+        this.nodeRepository = nodeRepository;
+        if (zone.system().isCd()) {
+            expiry = Duration.ofMinutes(30);
```
freva, line 44: This seems rather short. The reason we expire hosts from `provisioned` to `parked` is because, in general, that should not happen and if it does, we want someone to take a look and see if the fix can be automated. This is hard to do if the hosts are deprovisioned before anyone has a chance to look at the issue. 
Note that we do not sync system logs to archive storage, without those or the actual host it may be impossible to determine why a host failed to come up.

bjormel:
Thanks @freva , much appreciate, did not expect any feedback this year 😄

PR #25335, opened 2022-12-23
owner, shutdown and error propagation for async io

havardpe:
@geirst please review


geirst (approved):
👍  I assume the usage of _fprintf_ is temporary.

PR #25336, opened 2022-12-23
Revert "Revert "Add an eof object that can be sent to the Q to wake up the ones waiti…""

baldersheim:
Reverts vespa-engine/vespa#25308
@aressem or @hmusum PR
Retrying this one as there were other reasons for failing performance tests.


hmusum:
👍

PR #25337, opened 2022-12-27
Output more info if it's likely OOM killer has run

hmusum:
We get some questions/support about this, try to inform user if OOM killer has run


arnej27959 (approved):
good idea!

PR #25338, opened 2022-12-29
Don't warn when 1 node is down

bratseth:


container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -252,11 +256,15 @@ private void trackGroupCoverageChanges(Group group, boolean fullCoverage, long m
...
-                            "Active documents: " + group.activeDocuments() + "/" + medianDocuments + ", " +
-                            "Target active documents: " + group.targetActiveDocuments() + ", " +
-                            "working nodes: " + group.workingNodes() + "/" + group.nodes().size() +
-                            ", unresponsive nodes: " + (unresponsive.toString().isEmpty() ? " none" : unresponsive));
+                String message = "Cluster " + clusterId + ": " + group + " has reduced coverage: " +
+                                 "Active documents: " + group.activeDocuments() + "/" + medianDocuments + ", " +
+                                 "Target active documents: " + group.targetActiveDocuments() + ", " +
+                                 "working nodes: " + group.workingNodes() + "/" + group.nodes().size() +
+                                 ", unresponsive nodes: " + (unresponsive.toString().isEmpty() ? " none" : unresponsive);
+                if (nonWorkingNodeCount() == 1) // That is normal
```
baldersheim, line 264: This changes for both single node and multi node groups. Is that intentional ?
Is there another reason why this has only been reported as an issue for single node groups ?
You might be just masking a symptom now.

bratseth: Yes, intentional since whether we have one or many nodes in a group it is normal to take at least one node down.

baldersheim: OK, so we assume that we have the same logging for groups with multiple nodes too.
If so this is fine.

PR #25339, opened 2022-12-29
Compile sample apps in component build.

aressem:

### DIFF
--- a/screwdriver.yaml
+++ b/screwdriver.yaml
@@ -149,7 +149,7 @@ jobs:
           ccache --show-stats
           time make -j ${NUM_THREADS} install DESTDIR=$WORKDIR/vespa-install
       - build-sample-apps: |
-          (cd sample-apps && time mvn -T $NUM_THREADS $VESPA_MAVEN_EXTRA_OPTS package)
+          (cd $WORKDIR/sample-apps && time mvn -T $NUM_THREADS $VESPA_MAVEN_EXTRA_OPTS package)
       - verify-rpm-build: |
           cd $WORKDIR
           ulimit -c 0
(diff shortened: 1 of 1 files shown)


screwdriver.yaml
```diff
@@ -124,8 +124,10 @@ jobs:
           (cd vespa && git checkout $VESPA_REF && git tag v$VESPA_VERSION)
           git clone -q https://github.com/vespa-engine/system-test
           (cd system-test && git checkout $SYSTEM_TEST_REF)
+          git clone -q https://github.com/vespa-engine/sample-apps
           # Set correct version in pom.xml files
           (cd vespa && screwdriver/replace-vespa-version-in-poms.sh $VESPA_VERSION $(pwd) )
+          (cd sample-apps && find . -name "pom.xml" -exec sed -i -e "s,<vespa_version>.*</vespa_version>,<vespa_version>$VESPA_VERSION</vespa_version>," {} \;)
```
arnej27959, line 130: consider extending screwdriver/replace-vespa-version-in-poms.sh to handle vespa_version and using it here

arnej27959 (approved):
looks good to me, only cosmetic suggestion

bjormel (approved):

screwdriver.yaml
aressem: I considered doing this, but the `<vespa_version>` tag is very specific to the sample apps repo. Also it has a mix of fixed version using `<vespa_version>` and ranged `<version>[8,9)</version>` tags that does not match the pattern in `replace-vespa-version-in-poms.sh` (but works since the latest 8 is already in .m2). So I will leave this as is for now.

PR #25340, opened 2022-12-29
add fork of airlift zstd code

arnej27959:
@gjoranv please verify that this is OK with enforcer rules etc
@vekterli please take a look at the zstd code
@jonmv FYI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ vespa_install_data(tsan-suppressions.txt etc/vespa)
 # Include vespa config definitions in every target
 include_directories(BEFORE ${CMAKE_BINARY_DIR}/configdefinitions/src)
 
+add_subdirectory(airlift-zstd)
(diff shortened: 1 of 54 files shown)


vekterli (approved):
:+1: with this I can finally retire my hacky branch with manually patched files. As discussed offline, it would be good to get at least a subset of the original unit tests, but this can be a follow-up since we're not changing the semantics of the forked code (and the tests cannot be unreasonably expensive to run).

gjoranv (approved):
It seems like the original aircompress is now unused, and could be removed. Should we keep it to avoid extra work if/when they decide to merge the mentioned PR?

baldersheim:
I suggest removing it. It is not that hard to add it back. And just blurs how the dots are connected.

arnej27959:
Thanks @gjoranv - this looks good to me

baldersheim (approved):

PR #25347, opened 2023-01-02
Make zookeeper server non-reconfigurable in self-hosted [run-systemtest]

hmusum:


zookeeper-server/zookeeper-server/src/main/java/com/yahoo/vespa/zookeeper/ConfigServerZooKeeperServer.java
```diff
@@ -0,0 +1,43 @@
...
+package com.yahoo.vespa.zookeeper;
+
+import com.yahoo.cloud.config.ZookeeperServerConfig;
+import com.yahoo.component.AbstractComponent;
+import com.yahoo.component.annotation.Inject;
+import java.nio.file.Path;
+
+/**
+ *
+ * Server used for starting config server, needed to be able to be able to have different behavior for hosted and
```
hakonhall, line 11:
```suggestion
 * Server used for starting config server, needed to be able to have different behavior for hosted and
```

hakonhall (approved):

PR #25348, opened 2023-01-02
Balder/extend flag eol

baldersheim:
@aressem or @hmusum PR


hmusum (approved):

aressem (approved):

aressem:
Any of these that should be considered moved to permanent flags ?

baldersheim:
Will consider when we see what is left after vespa 7 is gone.

PR #25349, opened 2023-01-02
Test jobs have no permanent runtime versions

jonmv:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -324,15 +324,16 @@ private Optional<Version> oldestInstalledPlatform(JobId job) {
     /** Returns the oldest Vespa version installed on any active or reserved production node for the given application. */
     public Optional<Version> oldestInstalledPlatform(Application application) {
         return controller.jobController().deploymentStatus(application).jobs()
-                         .production().asList()
-                         .stream()
+                         .production()
```
mpolden, line 327: Doesn't `production()` imply `not().test()`?

jonmv: Production tests are production tests Ü

mpolden (approved):

PR #25353, opened 2023-01-02
Ensure lock is cleared when lease is invalidated

jonmv:


jonmv:
Without this, the following could happen:

take lock
lease invalidated from curator
update status, doing nothing, in particular not clearing the lock

renew lease again, but not requiring lock, since this wasn't cleared
ping ZK, verifying we are live (and our lock is valid)
proceed with activation

hakonhall (approved):

PR #25361, opened 2023-01-02
Use mutex/condition_variable instead of atomic/sleep for synchronizat…

baldersheim:
…ion.

@toregge PR


baldersheim:
Get a run with thread sanitizer

toregge (commented):
👍

PR #25365, opened 2023-01-03
Remove unsupported docproc xml elements

hmusum:
`<docproc>` hasn't been supported for a long time, not sure why this hasn't been removed before.


baldersheim (approved):
I think we can remove all the attributes in the docproc cluster.
I will go through the code and clean up some more.

gjoranv (approved):

PR #25374, opened 2023-01-03
Avoid uploading metrics-proxy access logs

jonmv:
@freva

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfo.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfo.java
@@ -82,7 +82,7 @@ public class SyncFileInfo {
             compression = filename.endsWith(".zst") ? Compression.NONE : Compression.ZSTD;
             if (rotatedOnly && compression != Compression.NONE)
                 dir = null;
-            else if (filename.matches(".*\\.metrics-proxy\\..*")) // See AccessLogComponent.java for filename.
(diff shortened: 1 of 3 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfo.java
```diff
@@ -82,12 +82,14 @@ public static Optional<SyncFileInfo> forLogFile(URI uri, Path logFile, boolean r
             compression = filename.endsWith(".zst") ? Compression.NONE : Compression.ZSTD;
             if (rotatedOnly && compression != Compression.NONE)
                 dir = null;
+            else if (filename.matches(".*\\.metrics-proxy\\..*")) // See AccessLogComponent.java for filename.
```
freva, line 85: Maybe `filename.contains(".metrics-proxy.")` to avoid regex?

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/sync/SyncFileInfo.java
jonmv: Of course ...

jonmv:
```suggestion
            else if (filename.contains(".metrics-proxy.")) // See AccessLogComponent.java for filename.
```

PR #25377, opened 2023-01-03
remove _shutdown flag

havardpe:
- remove debug printf
- move shutdown code into RunGuard class
- let readable/writable return bool (ok/cancel)

@geirst please review


geirst (approved):
👍

PR #25385, opened 2023-01-04
Enforce that we also get correct commons-codec in tentant-base and ot…

baldersheim:
…her related places.

@bjorncs @gjoranv PR
This is necessary when upgrading apache.httpcomponents.core5


bjorncs (changes requested):
`container-dependencies-versions` should only manage dependencies that are provided as bundles by container. Move to `parent/pom.xml`.

bjorncs (approved):

PR #25394, opened 2023-01-04
When resorting to paged structures you are will use all memory and more.

baldersheim:
That is too large to get in a coredump, so ignore it.

@vekterli @toregge @geirst PR


vekterli (approved):
:+1:

toregge (approved):
Note that Darwin does not define MADV_DONTDUMP.

PR #25401, opened 2023-01-04
Extract virtual fields in the index environment.

geirst:
Fields that are represented by a set of attributes in the backend are considered virtual fields. Currently, this is map or array of struct fields (from the SD file) with struct-field attributes.

@havardpe please review


havardpe (approved):
any reason not to use the already extracted fields as input when detecting virtual fields?
(consider also testing with non-attribute and shadowing fields)

PR #25404, opened 2023-01-04
if the Document to be processed is the wrong type, convert it

arnej27959:
* this happens when you have a concrete document
* the indexing processor works better with the "normal" format, especially if there are any complex structure in the schema such as array<string> that needs processing
* convert to normal format by serializing and deserializing

@baldersheim please review
@bratseth FYI

### DIFF
--- a/docprocs/src/main/java/com/yahoo/docprocs/indexing/IndexingProcessor.java
+++ b/docprocs/src/main/java/com/yahoo/docprocs/indexing/IndexingProcessor.java
@@ -17,6 +17,9 @@ import com.yahoo.document.DocumentRemove;
 import com.yahoo.document.DocumentType;
 import com.yahoo.document.DocumentTypeManager;
 import com.yahoo.document.DocumentUpdate;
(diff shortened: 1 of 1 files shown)


docprocs/src/main/java/com/yahoo/docprocs/indexing/IndexingProcessor.java
```diff
@@ -93,14 +96,24 @@ DocumentTypeManager getDocumentTypeManager() {
...
         if (script == null) {
             log.log(Level.FINE, "No indexing script for document '%s'.", prev.getId());
             out.add(prev);
             return;
         }
         log.log(Level.FINE, "Processing document '%s'.", prev.getId());
-        Document next = script.execute(adapterFactory, prev.getDocument());
+        DocumentType wantType = docTypeMgr.getDocumentType(hadType.getName());
+        Document prevDoc = prev.getDocument();
+        if (hadType != wantType) {
```
baldersheim, line 109: I think that a comment here would be good. Not obvious....

baldersheim (approved):

PR #25405, opened 2023-01-04
DJL-based HuggingFaceEmbedder prototype

dnmca:
---
This is a prototype of custom embedder that uses DJL's HuggingFaceTokenizer with ONNX-converted model.

This code was tested with single model type: XLMRoberta, but should work with larger range of models from HuggingFace.

I have tested the code with my model files, but unfortunately I can't share them.

Model's onnx file was obtained by using `transformers.onnx` utility:
```
python -m transformers.onnx --model=$HF_MODEL_PATH onnx/
```


bratseth (approved):
Great! Some work is needed to allow the huggingface dependency, but I can do that.

PR #25408, opened 2023-01-05
Avoid failing assert when racing with deploy thread, in unit test

jonmv:
@freva or @mpolden please review and merge.


jonmv:
Urgh ...

jonmv:
@vekterli FYI.

mpolden (approved):

PR #25411, opened 2023-01-05
reject new external messages while closing and flushing

havardpe:
@vekterli please review


vekterli:
After looking at the code in question more closely it seems it's a bit more Complicated™ to do this. Returning false from the message handler (indicating message was not handled) dispatches the message further down the component chain, but there is nothing there to accept the message, so we'd get an unhandled message warning. We might want to reconsider the design for component flushing/closing to ensure we have sufficient barriers in place further up the chain (closer to the RPC/mbus source) that no messages can arrive at the distributor during the flush/close step.

Closing this PR for further pondering... 🦄

PR #25412, opened 2023-01-05
Arnej/add unit tests for zstd

arnej27959:
@gjoranv please review changes to pom.xml and allowed-maven-dependencies


gjoranv (approved):
Sorry for late review.

PR #25421, opened 2023-01-05
Improve shared hosts resource tagging

freva:

Hosts that were exclusive to a cluster type will now we tagged with `hosted-vespa.shared-host.<cluster-type>` (previously `hosted-vespa.intrastructure.default`).


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/noderepository/NodeRepositoryNode.java
```diff
@@ -105,6 +105,8 @@ public class NodeRepositoryNode {
     private String reservedTo;
     @JsonProperty("exclusiveTo")
     private String exclusiveTo;
+    @JsonProperty("exclusiveToClusterType")
+    private String exclusiveToClusterType;
```
hakonhall, line 109: add to toString()

hakonhall (approved):

hakonhall (approved):

PR #25423, opened 2023-01-06
Improve user friendliness when class resolution fails.

gjoranv:


baldersheim (approved):

container-core/src/main/java/com/yahoo/osgi/OsgiImpl.java
```diff
@@ -62,6 +62,11 @@ public Class<?> resolveClass(BundleInstantiationSpecification spec) {
         if (bundle != null) {
             return resolveFromBundle(spec, bundle);
         } else {
+            if (spec.bundle != spec.classId) {
```
bjorncs, line 65: Should be `.equals()`?

bjorncs (approved):

container-core/src/main/java/com/yahoo/osgi/OsgiImpl.java
gjoranv: Thanks, fixed.

bjorncs (approved):

PR #25430, opened 2023-01-06
Arnej/move consensus time metric

arnej27959:
@baldersheim please review


arnej27959:
with these changes I got the new metric through metrics-proxy at `/metrics/v1/values?consumer=vespa`

baldersheim (approved):

PR #25431, opened 2023-01-06
Exclusive containers in public

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/NodeRepository.java
```diff
@@ -205,6 +205,7 @@ public NodeRepository(NodeFlavors flavors,
      */
     public boolean exclusiveAllocation(ClusterSpec clusterSpec) {
         return clusterSpec.isExclusive() ||
+               ( clusterSpec.type().isContainer() && zone.system().isPublic() &&  !zone.environment().isTest() ) ||
```
mpolden, line 208: Are we sure we want to limit this to containers only? History shows that there are ways to run custom code on other cluster types.

bratseth: Debatable, but we should definitely not allow that, so disallowing it would be from a defence in depth argument, where it must be weighted against the potential cost of this, which is much higher on content nodes, as that's where most of the resource cost and therefore cost optimization is spent, and I just don't think it outweights that. For example, it is much harder to defend against lateral movement from the host than disallowing content node code execution, and we still trust ourselves to do it.

mpolden (approved):

PR #25432, opened 2023-01-06
Use a thread pool for HTTP dispatch, especially faster with compression

jonmv:
@bjorncs please review. 

I don't think this is ideal. I wonder if the client has changed enough since its simple conception, that we should revisit some of the code. For now, this is a simple fix, and increases through both with and without compression significantly, on a single-host test. We should also test further with a multi-host setup.


bjorncs (approved):
👍 
Agree, we should look into using the http client's async request producer interface

jonmv:
Implementing the async entity producer, with zipping in client threads, is _slower_, though ... "on my machine".

PR #25439, opened 2023-01-07
Revert "add some logctl tuning"

hmusum:
Reverts vespa-engine/vespa#25435

System tests failing when run as part of build, can't find any other commits than this one causing it


arnej27959:
you're correct, this introduced a problem, but should be fixed now.

PR #25443, opened 2023-01-08
More user friendly class resolution

gjoranv:
Only the last commit is new since #25423

com.yahoo.jdisc.application is not PublicApi


baldersheim (approved):

bjorncs (commented):
👍

PR #25445, opened 2023-01-09
Reapply "Let vespa own the /conf/vespa dir."

aressem:
Reverts vespa-engine/vespa#25437


toregge (approved):

arnej27959:
I think we want to leave the conf/vespa directory as "configuration from outside", so no vespa processes should need to write to it, and it should not be owned by the vespa user.

aressem:
Sounds good. I seem to have forgotten the reasoning behind this change.

PR #25449, opened 2023-01-09
Use gzip by default for payloads > 512b

jonmv:
@bjorncs


bjorncs (approved):
Consider making compression a tri-state: default,off,gzip

bjorncs:
FYI @baldersheim

baldersheim:
Agree with default,none,gzip enum. I assume there might soon be zstd too.

jonmv:
Do you want an explicit `default`, which is what the current <no setting> is? Is that needed?

jonmv:
Updating --help in https://github.com/vespa-engine/vespa/pull/25451

jonmv:
`auto`, perhaps?

PR #25458, opened 2023-01-09
Reduce vespa-zkctl log spam

mpolden:
In #20594 we switched to log4j-over-slf4j to get rid of log4j. This changed the
underlying logging implementation to Logback, which must be configured in a
different way.

@hmusum


mpolden:
I see some commits do not belong here, removing.

hmusum (approved):
Great!

PR #25464, opened 2023-01-09
Patch wantToDeprovision recursively

hakonhall:
A host can be deprovisioned if its children are (a) failed or parked, and (b) has wantToDeprovision.  This is utilized e.g. in ProvisionedExpirer to deprovision hosts when there are too many parked nodes already.  This cannot be replicated by an operator (trivially), as setting wantToRetire+wantToDeprovision on a host is not recursive in wTD.  This PR makes wTD recursive.

This is aligned with https://github.com/vespa-engine/vespa/pull/16963.

It also reduces the max number of parked nodes before automatically deprovisioning hosts from 20 to 5.


freva (approved):

bjormel:
🏅

PR #25465, opened 2023-01-09
Use 'null' bundle instead of the invalid 'container-core'

gjoranv:
- Worked by accident because of the class lookup in the container falls back to container-disc.


baldersheim (approved):

bjorncs (commented):
👍

PR #25476, opened 2023-01-10
Revert "Require non-null compression, and update doc"

aressem:
Reverts vespa-engine/vespa#25471

Broke some system tests:
https://factory.vespa.oath.cloud/testrun/55612941


freva:
@jonmv fixed in #25477

aressem:
👍

PR #25478, opened 2023-01-10
Read node objects from unified path

mpolden:
We still write to the old paths. I'll remove those after this change is released
everywhere.

@bratseth


bratseth (approved):
Great!

PR #25480, opened 2023-01-10
Add units to the global container metrics enum.

gjoranv:
The units defined so far are base on https://github.com/DataDog/integrations-extras/blob/master/vespa/metadata.csv

A couple of good things with this:
1. We avoid using almost the same unit, e.g. "document" vs "documents"
2. The unit is automatically added to the description.

Note that it's usually with the 'rate' suffix/aggregator that we use "per second", so most metrics should be declared with unit 'document', 'operation' etc.


kkraune (approved):
👍

PR #25482, opened 2023-01-10
Define in ContainerMetrics.java

kkraune:
... or @olaaun

### DIFF
--- a/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java
+++ b/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java
@@ -66,13 +66,13 @@ class MetricDefinitions {
     static final String SSL_HANDSHAKE_FAILURE_UNKNOWN = "jdisc.http.ssl.handshake.failure.unknown";
     static final String SSL_HANDSHAKE_FAILURE_CONNECTION_CLOSED = "jdisc.http.ssl.handshake.failure.connection_closed";
 
-    static final String JETTY_THREADPOOL_MAX_THREADS = "jdisc.http.jetty.threadpool.thread.max";
-    static final String JETTY_THREADPOOL_MIN_THREADS = "jdisc.http.jetty.threadpool.thread.min";
-    static final String JETTY_THREADPOOL_RESERVED_THREADS = "jdisc.http.jetty.threadpool.thread.reserved";
-    static final String JETTY_THREADPOOL_BUSY_THREADS = "jdisc.http.jetty.threadpool.thread.busy";
-    static final String JETTY_THREADPOOL_IDLE_THREADS = "jdisc.http.jetty.threadpool.thread.idle";
-    static final String JETTY_THREADPOOL_TOTAL_THREADS = "jdisc.http.jetty.threadpool.thread.total";
-    static final String JETTY_THREADPOOL_QUEUE_SIZE = "jdisc.http.jetty.threadpool.queue.size";
+    static final String JETTY_THREADPOOL_MAX_THREADS = ContainerMetrics.JETTY_THREADPOOL_MAX_THREADS.baseName();
+    static final String JETTY_THREADPOOL_MIN_THREADS = ContainerMetrics.JETTY_THREADPOOL_MIN_THREADS.baseName();
+    static final String JETTY_THREADPOOL_RESERVED_THREADS = ContainerMetrics.JETTY_THREADPOOL_RESERVED_THREADS.baseName();
+    static final String JETTY_THREADPOOL_BUSY_THREADS = ContainerMetrics.JETTY_THREADPOOL_BUSY_THREADS.baseName();
+    static final String JETTY_THREADPOOL_IDLE_THREADS = ContainerMetrics.JETTY_THREADPOOL_IDLE_THREADS.baseName();
+    static final String JETTY_THREADPOOL_TOTAL_THREADS = ContainerMetrics.JETTY_THREADPOOL_TOTAL_THREADS.baseName();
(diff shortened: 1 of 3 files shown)


container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
```diff
@@ -10,7 +10,22 @@ public enum ContainerMetrics {
...
+    JETTY_THREADPOOL_RESERVED_THREADS("jdisc.http.jetty.threadpool.thread.reserved", Unit.RESPONSE,
+            "Configured number of reserved threads or -1 for heuristic"),
+    JETTY_THREADPOOL_BUSY_THREADS("jdisc.http.jetty.threadpool.thread.busy", Unit.RESPONSE,
+            "Number of threads executing internal and transient jobs"),
+    JETTY_THREADPOOL_IDLE_THREADS("jdisc.http.jetty.threadpool.thread.idle", Unit.RESPONSE,
+            "Number of idle threads"),
+    JETTY_THREADPOOL_TOTAL_THREADS("jdisc.http.jetty.threadpool.thread.total", Unit.RESPONSE,
+            "Current number of threads"),
+    JETTY_THREADPOOL_QUEUE_SIZE("jdisc.http.jetty.threadpool.queue.size", Unit.RESPONSE,
+            "Current size of the job queue");
```
gjoranv, line 28: All these should probably use `Unit.THREAD`

gjoranv (changes requested):

container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
kkraune: all right, I was pondering on that - hang on ...

container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
```diff
@@ -10,7 +10,22 @@ public enum ContainerMetrics {
     HTTP_STATUS_3XX("http.status.3xx", Unit.RESPONSE, "Number of responses with a 3xx status"),
     HTTP_STATUS_4XX("http.status.4xx", Unit.RESPONSE, "Number of responses with a 4xx status"),
     HTTP_STATUS_5XX("http.status.5xx", Unit.RESPONSE, "Number of responses with a 5xx status"),
-    JDISC_GC_MS("jdisc.gc.ms", Unit.MILLISECOND, "Time spent in garbage collection");
+    JDISC_GC_MS("jdisc.gc.ms", Unit.MILLISECOND, "Time spent in garbage collection"),
+
+    JETTY_THREADPOOL_MAX_THREADS("jdisc.http.jetty.threadpool.thread.max", Unit.RESPONSE,
```
gjoranv, line 15: The way this was designed, the 'max' suffix should not be included in the definition.

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java
```diff
@@ -66,13 +66,13 @@ class MetricDefinitions {
...
     static final String SSL_HANDSHAKE_FAILURE_CONNECTION_CLOSED = "jdisc.http.ssl.handshake.failure.connection_closed";
 
-    static final String JETTY_THREADPOOL_MAX_THREADS = "jdisc.http.jetty.threadpool.thread.max";
-    static final String JETTY_THREADPOOL_MIN_THREADS = "jdisc.http.jetty.threadpool.thread.min";
-    static final String JETTY_THREADPOOL_RESERVED_THREADS = "jdisc.http.jetty.threadpool.thread.reserved";
-    static final String JETTY_THREADPOOL_BUSY_THREADS = "jdisc.http.jetty.threadpool.thread.busy";
-    static final String JETTY_THREADPOOL_IDLE_THREADS = "jdisc.http.jetty.threadpool.thread.idle";
-    static final String JETTY_THREADPOOL_TOTAL_THREADS = "jdisc.http.jetty.threadpool.thread.total";
-    static final String JETTY_THREADPOOL_QUEUE_SIZE = "jdisc.http.jetty.threadpool.queue.size";
+    static final String JETTY_THREADPOOL_MAX_THREADS = ContainerMetrics.JETTY_THREADPOOL_MAX_THREADS.baseName();
```
gjoranv, line 69: Instead of `basename()`, use `max()`.

kkraune: Not sure, see https://github.com/vespa-engine/vespa/blob/master/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java#L150 - for the "max" metric, we add all sorts of suffixes. That is maybe not right, but I did not want to change behavior in these PRs

container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
kkraune: see comment above

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java
gjoranv: Sorry, I didn't notice that 'max' was part of the actual metric name, and not a suffix.

gjoranv (approved):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java
kkraune: np - this is confusing :-)

PR #25495, opened 2023-01-11
Update abi MERGEOK

baldersheim:
@jonmv or @hmusum PR
Do we need classes in config-model-api to be @PublicApi ?
They are in com.yahoo.config.application.api, but not in com.yahoo.config.model.api ?


jonmv:
👍

jonmv:
Thanks for fixing.

PR #25496, opened 2023-01-11
Revert "Read tags from deployment spec in package"

hmusum:
Reverts vespa-engine/vespa#25492

Breaks build (basic system test):

java.lang.NoClassDefFoundError: com/yahoo/config/application/api/xml/DeploymentSpecXmlReader


hmusum:
Seems like this has been fixed in other ways

PR #25498, opened 2023-01-11
Balder/no publicapi needed

baldersheim:
@hmusum or @jonmv PR


jonmv:
@bratseth any objections?

jonmv:
The only reason I can think of is to remind ourselves that these changes can break the configserver ... which is perhaps a good enough reason?

baldersheim:
But then it should be the same on all packages there, not only 20%.
I agree that catching changes here is good, but won't we catch that in CD ?
And it is not a public api ........

bratseth:
No objections.

hmusum (approved):

PR #25500, opened 2023-01-11
Don't attempt to lookup class in internal bundles if a bundle was explicitly specified

gjoranv:


hmusum (approved):

baldersheim (approved):

hmusum:
I guess this PR depends on https://git.ouryahoo.com/vespa/vespa-yahoo/pull/10542?

PR #25501, opened 2023-01-11
Jonmv/remove tags wiring

jonmv:
@bratseth

### DIFF
--- a/config-application-package/pom.xml
+++ b/config-application-package/pom.xml
@@ -109,12 +109,6 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-          <groupId>xmlunit</groupId>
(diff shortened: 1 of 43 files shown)


bratseth (approved):
Great! Now we _really_ can never use it for something else, which i quite like ...

bratseth:
Failed to execute goal com.yahoo.vespa:abi-check-plugin:8-SNAPSHOT:abicheck (default) on project config-model-api: ABI spec mismatch.

jonmv:
Yes, fixed. Removed method is in use in older models, just not on my branching point ...

bratseth (approved):

jonmv:
Sorry, accidentally removed a line of xml unit test helper. Resulted in killing the dependency instead. PTAL.

mpolden (approved):

PR #25506, opened 2023-01-11
Kkraune/metrics

kkraune:
- Review only - I want to move more connectiopn-metrics, see comment

### DIFF
--- a/container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
+++ b/container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
@@ -10,7 +10,11 @@ public enum ContainerMetrics {
     HTTP_STATUS_3XX("http.status.3xx", Unit.RESPONSE, "Number of responses with a 3xx status"),
     HTTP_STATUS_4XX("http.status.4xx", Unit.RESPONSE, "Number of responses with a 4xx status"),
     HTTP_STATUS_5XX("http.status.5xx", Unit.RESPONSE, "Number of responses with a 5xx status"),
+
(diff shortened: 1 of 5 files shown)


container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
```diff
@@ -40,7 +44,10 @@ public enum ContainerMetrics {
     JETTY_THREADPOOL_TOTAL_THREADS("jdisc.http.jetty.threadpool.thread.total", Unit.THREAD,
             "Current number of threads"),
     JETTY_THREADPOOL_QUEUE_SIZE("jdisc.http.jetty.threadpool.queue.size", Unit.THREAD,
-            "Current size of the job queue");
+            "Current size of the job queue"),
+
+    SERVER_NUM_CONNECTIONS("serverNumConnections", Unit.RESPONSE,
```
kkraune, line 49: @gjoranv : Do we need a new unit for connections, or should we add something generic? Not sure what this will be used for

gjoranv: Yes, I suggest to add new units for `connection` and `request`. Units are, at least currently, only for documentation, so no need to be restrictive.

kkraune:
@gjoranv PTAL - easier to review per commit, the last is just for readability

container-core/src/main/java/com/yahoo/metrics/ContainerMetrics.java
kkraune: Added. Thanks

gjoranv:
Great!

PR #25507, opened 2023-01-11
Add utility functionality for trapping memory accesses

vekterli:
@baldersheim please review
@havardpe FYI

At its core this is a way to wrap a range of memory in a way where all non-zero writes are eventually detected and, if supported by the underlying OS and hardware, instantly detects both reads and writes to as much as possible of the underlying memory range. Implemented using `mprotect` of page-aligned regions under the hood.

### DIFF
--- a/vespalib/src/vespa/vespalib/util/memory_trap.cpp
+++ b/vespalib/src/vespa/vespalib/util/memory_trap.cpp
@@ -8,11 +8,8 @@
 #include <cstdlib>
 #include <cstring>
 #include <malloc.h>
-#ifdef __linux__
-#  define VESPA_HAS_MPROTECT
-#  include <unistd.h>
-#  include <sys/mman.h>
-#endif
+#include <unistd.h>
+#include <sys/mman.h>
 
 #include <vespa/log/log.h>
 LOG_SETUP(".vespalib.util.memory_trap");
@@ -75,7 +72,7 @@ void MemoryRangeTrapper::verify_buffer_is_all_zeros() {
     }
 }
 
-#ifdef VESPA_HAS_MPROTECT
+#ifdef __linux__
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


havardpe (commented):
possible improvements:
* check protected area before checking unprotected are to prefer worse error case
* protect with read and check zeroes on init to avoid before-protected race

vespalib/src/vespa/vespalib/util/memory_trap.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <vespa/log/log.h>
+LOG_SETUP(".vespalib.util.memory_trap");
+
+using namespace std::string_view_literals;
+
+namespace vespalib {
+
+namespace {
+
+#ifdef VESPA_HAS_MPROTECT
```
baldersheim, line 29: I am not a fan of #ifdefs. Is it possible to gather all 5 into just a single one. That makes it far more readable and maintainable.
You get clean method definitions. One dummy set and one for mprotect.

baldersheim (changes requested):
Prefer a single ifdef.
I also think that @havardpe or @toregge should take a look. I have a better eye for the gory details.

vekterli:
> Prefer a single ifdef.

Reduced from 5 to 2. I don't want to mix includes in with the code body section, nor do I want to include Linux-specific headers on other OSes. PTAL

vekterli:
> possible improvements:
> 
> * check protected area before checking unprotected are to prefer worse error case
> * protect with read and check zeroes on init to avoid before-protected race

Will consider these in a follow-up

vekterli:
Actually, the relevant headers appear to be POSIX, so I can remove the `ifdef` around those, causing much happiness across the land

vespalib/src/vespa/vespalib/util/memory_trap.cpp
vekterli: Fixed

baldersheim (approved):

PR #25519, opened 2023-01-11
Yngveaasheim/define container metrics [skip-systemtest]

yngveaasheim:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
@@ -164,21 +164,17 @@ public class VespaMetricSet {
             addMetric(metrics, ContainerMetrics.JETTY_THREADPOOL_QUEUE_SIZE.baseName(), suffixes);
         }
 
-        metrics.add(new Metric("httpapi_latency.max"));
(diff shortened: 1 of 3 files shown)


yngveaasheim:
@gjoranv or @olaaun Please review and merge.
@kkraune FYI

gjoranv (approved):
👍  Merge if green build.

kkraune:
I suggest we also make the changes in the other files to use the definition, too, so easier to keep track of what needs to be done:
```
git grep httpapi_latency
config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java:        metrics.add(new Metric("httpapi_latency.max"));
config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java:        metrics.add(new Metric("httpapi_latency.sum"));
config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java:        metrics.add(new Metric("httpapi_latency.count"));
C02DR9A6MD6R:vespa kraune$ 
```
I.e. change VespaMetricSet.java

gjoranv:
It seems that is done in this branch already, or am I missing something?

kkraune:
> It seems that is done in this branch already, or am I missing something?

Yes you are right :-) my bad

PR #25531, opened 2023-01-12
Use enum for mem metrics

yngveaasheim:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
@@ -176,21 +176,19 @@ public class VespaMetricSet {
         addMetric(metrics, ContainerMetrics.HTTPAPI_CONDITION_NOT_MET.baseName(), List.of("rate"));
         addMetric(metrics, ContainerMetrics.HTTPAPI_NOT_FOUND.baseName(), List.of("rate"));
 
-        metrics.add(new Metric("mem.heap.total.average"));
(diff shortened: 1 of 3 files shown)


yngveaasheim:
@olaaun or @gjoranv Please review and merge.

kkraune (approved):
👍

config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
```diff
@@ -176,21 +176,19 @@ private static Set<Metric> getContainerMetrics() {
         addMetric(metrics, ContainerMetrics.HTTPAPI_CONDITION_NOT_MET.baseName(), List.of("rate"));
         addMetric(metrics, ContainerMetrics.HTTPAPI_NOT_FOUND.baseName(), List.of("rate"));
 
-        metrics.add(new Metric("mem.heap.total.average"));
-        metrics.add(new Metric("mem.heap.free.average"));
-        metrics.add(new Metric("mem.heap.used.average"));
-        metrics.add(new Metric("mem.heap.used.max"));
+        addMetric(metrics, ContainerMetrics.MEM_HEAP_TOTAL.baseName(), List.of("average"));
```
kkraune, line 179: consider using private static void addMetric(Set<Metric> metrics, String nameWithSuffix) instead of list with one item (applies to the other below, too, with only one suffix)

yngveaasheim: Thanks. Changed. I agree this is better for readability and also prevents possible typos from being merged.

kkraune (approved):

PR #25545, opened 2023-01-12
Revert "No need to notify if bucket is already locked"

baldersheim:
Reverts vespa-engine/vespa#25537
@vekterli I think this only holds when the message added is an async message.


vekterli (approved):
:+1:

PR #25549, opened 2023-01-12
Parse tensor JSON values at root

bratseth:
Our current tensor JSON formats require a "blocks", "cells" or "values" key at the root, containing values in various forms.

This adds support for skipping that extra level and adding content at the root, where the permissible content format depends on the tensor type, and matches the formats below "blocks", "cells" or "values" for the corresponding tensor types.


lesters (approved):
Great!

PR #25550, opened 2023-01-12
Drop the lock before notifying

baldersheim:
@vekterli PR


vekterli (commented):
:+1: did unlocked notify turn out to be faster after all?

PR #25551, opened 2023-01-12
No need to notify anyone when you grab a bucket lock. Nobody will be …

baldersheim:
…allowed to do anything anyway.

@vekterli PR


vekterli (commented):
:+1:

PR #25558, opened 2023-01-13
Throw invalid application if hitting hosted code with self-hosted syntax

jonmv:
@bratseth please review.


bratseth (approved):

jonmv:
No longer nedded.

PR #25565, opened 2023-01-13
Move parent-child cache to NodeList

mpolden:
...so you don't have to remember to use a special list for fast lookups.

NodeListMicroBenchmarkTest before:
Calling NodeList.childrenOf took PT5.015094S (50 microseconds per invocation)

and after:
Calling NodeList.childrenOf took PT0.021818S (0 microseconds per invocation)

@baldersheim


mpolden:
or @bratseth

bratseth (approved):
Very nice

PR #25566, opened 2023-01-13
Move from spec.CloudAccount.isEnclave(Zone) to enclave()

hakonhall:


config-provisioning/src/main/java/com/yahoo/config/provision/CloudAccount.java
```diff
@@ -14,11 +15,25 @@ public class CloudAccount extends PatternedStringWrapper<CloudAccount> {
...
     public static final CloudAccount empty = new CloudAccount("", EMPTY, "cloud account");
 
+    /** Verifies accountId is a valid AWS account ID and return it unaltered, or throw an IllegalArgumentException. */
+    public static String requireAwsAccountId(String accountId) {
+        Validation.requireMatch(accountId, "AWS account ID", AWS_ACCOUNT_ID_PATTERN);
+        return accountId;
+    }
+
+    /** Verifies accountId is a valid GCP project ID and return it unaltered, or throw an IllegalArgumentException. */
+    public static String requireGcpProjectId(String projectId) {
```
freva, line 32: Maybe this should return a `CloudAccount` instead?

freva (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/CloudAccount.java
hakonhall: Let me remove the return value as it is not used ATM.  The idea was that it would follow Objects.requireNonNull() and friends that returns its argument.

freva (approved):

PR #25572, opened 2023-01-13
Only attribute fields can represent virtual fields.

geirst:
@vekterli or @toregge or @baldersheim please review
@havardpe FYI


vekterli (approved):
:+1:

PR #25579, opened 2023-01-16
simplify: use current target

arnej27959:
@toregge please review

### DIFF
--- a/vespalog/src/vespa/log/llparser.cpp
+++ b/vespalog/src/vespa/log/llparser.cpp
@@ -6,6 +6,7 @@
 #include <cstdlib>
 #include <unistd.h>
 #include <sys/time.h>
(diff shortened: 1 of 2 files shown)


vespalog/src/vespa/log/llparser.cpp
```diff
@@ -18,31 +19,19 @@ LLParser::LLParser()
...
-	    // If we catch an exception, use default target
-	    target = LogTarget::defaultTarget();
-	}
-        delete _target;
-        _target = target;
-    }
     snprintf(_defPid, 10, "%d", (int)getpid());
 }
 
 LLParser::~LLParser()
```
baldersheim, line 33: Can just `use = default;`

baldersheim (approved):

PR #25584, opened 2023-01-16
Add enums for remaining JDISC metrics

yngveaasheim:


yngveaasheim:
@gjoranv or @olaaun or @kkraune Please review and merge.

kkraune (approved):
Isn't it better if we also update the code updating these metrics in the same PR, so we know that ContainerMetrics.java is a list of metrics we have completed so far? E.g.

```
$ git grep jdisc.http.ssl.handshake.failure.missing_client_cert
config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java:        metrics.add(new Metric("jdisc.http.ssl.handshake.failure.missing_client_cert.rate"));
container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/MetricDefinitions.java:    static final String SSL_HANDSHAKE_FAILURE_MISSING_CLIENT_CERT = "jdisc.http.ssl.handshake.failure.missing_client_cert";
```

yngveaasheim:
With adding enums I will also make follow-up PR(s) for using the corresponding enums. Separating like this reduces the probability of merge conflicts between PRs.

PR #25597, opened 2023-01-17
- Remove the explicit close method in SessionManager. Just do it in d…

baldersheim:
…estructor instead.

- Destruct instead of close. Works even if it has not been constructed.
- Minor code cleanup.

@geirst or @toregge PR


geirst (approved):
👍

PR #25609, opened 2023-01-17
profile source blender children

havardpe:
more testing (termwise)
extend simplesearch with strictness and initRange

@geirst please review
@baldersheim FYI


geirst (approved):
👍

PR #25613, opened 2023-01-18
Move cloud.def (CloudConfig) to configdefinitions module

hmusum:
We want users to be able to use this config, so need to be in a module that is provided.


gjoranv:
The package cannot be exported from configdefinitions, because it's already exported from config-provisioning. 

I suggest to use a different approach. Create a new config in e.g. `configdefinitions` that is populated by the config-model. I don't see how CloudConfig gets its values except for special applications like the configserver-app. We could also consider adding the cloud native region name to the `SystemInfo` api as well?

hmusum:
Thanks @gjoranv , will try another approach, as you suggest

PR #25587, opened 2023-01-16
Parse, validate and use new zone endpoint syntax

jonmv:
@mpolden please review.

PR #25614, opened 2023-01-18
Revert "Parse, validate and use new zone endpoint syntax"

jonmv:
Reverts vespa-engine/vespa#25587


hmusum (approved):

jonmv:
Also fixed with https://github.com/vespa-engine/vespa/pull/25615

PR #25630, opened 2023-01-19
simplify document deserialize

arnej27959:
after some looking and thinking this buffer shuffling doesn't seem to serve any purpose.
I am doing a full rewrite of this code now, so this is just for review.
@baldersheim please review
@vekterli FYI


baldersheim (approved):

arnej27959:
please take another look now after refactoring.

baldersheim (approved):

PR #25631, opened 2023-01-19
Use string ID for WeightedAliasTarget

mpolden:
See internal PR.

@jonmv


jonmv (approved):
So we no longer require that ID to be a zone ID, ok ... and

PR #25634, opened 2023-01-19
Stop looking for DNS records by data only

mpolden:
Merge with internal PR.

@jonmv

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/dns/RemoveRecords.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/dns/RemoveRecords.java
@@ -8,7 +8,6 @@ import com.yahoo.vespa.hosted.controller.api.integration.dns.Record;
 import com.yahoo.vespa.hosted.controller.api.integration.dns.RecordData;
 import com.yahoo.vespa.hosted.controller.api.integration.dns.RecordName;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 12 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/dns/RemoveRecords.java
```diff
@@ -64,32 +55,18 @@ public Optional<RecordData> data() {
...
-            name.ifPresent(n -> records.addAll(nameService.findRecords(type, n)));
-            data.ifPresent(d -> records.addAll(nameService.findRecords(type, d)));
-        }
-        nameService.removeRecords(records);
+        // Deletions require all records fields to match exactly, data may be incomplete even if present. To ensure
+        // completeness we search for the record(s) first
+        List<Record> completeRecords = nameService.findRecords(type, name).stream()
+                                                  .filter(record -> data.isEmpty() || matchingFqdnIn(data.get(), record))
+                                                  .toList();
+        nameService.removeRecords(completeRecords);
```
jonmv, line 63: So we just drop deletion of unmatched records? Should we log a warning when this happens?

jonmv (approved):
One comment, looks good.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/dns/RemoveRecords.java
mpolden: The logic should be the same as before this change. Not sure what you mean by unmatched here? If you specify both name+data you specifically want to only delete records matching both fields.

jonmv: Aha, that sounds right.

PR #25642, opened 2023-01-19
Move failed DNS requests behind those of other apps, but keep interna…

jonmv:
…l ordering

@mpolden please review.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/NameServiceDispatcher.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/NameServiceDispatcher.java
@@ -1,22 +1,15 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.maintenance;
 
-import com.yahoo.collections.Iterables;
 import com.yahoo.vespa.hosted.controller.Controller;
 import com.yahoo.vespa.hosted.controller.api.integration.dns.NameService;
 import com.yahoo.vespa.hosted.controller.dns.NameServiceQueue;
-import com.yahoo.vespa.hosted.controller.dns.NameServiceRequest;
 import com.yahoo.vespa.hosted.controller.persistence.CuratorDb;
 
 import java.time.Clock;
 import java.time.Duration;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.Map;
-import java.util.function.Function;
+import java.util.ArrayList;
 import java.util.logging.Level;
-import java.util.stream.Collectors;
 
 /**
  * This dispatches requests from {@link NameServiceQueue} to a {@link NameService}. Successfully dispatched requests are
@@ -30,11 +23,15 @@ public class NameServiceDispatcher extends ControllerMaintainer {
     private final CuratorDb db;
     private final NameService nameService;
 
-    public NameServiceDispatcher(Controller controller, Duration interval) {
+    NameServiceDispatcher(Controller controller, NameService nameService, Duration interval) {
         super(controller, interval);
         this.clock = controller.clock();
         this.db = controller.curator();
-        this.nameService = controller.serviceRegistry().nameService();
+        this.nameService = nameService;
+    }
+
+    public NameServiceDispatcher(Controller controller, Duration interval) {
+        this(controller, controller.serviceRegistry().nameService(), interval);
     }
 
     @Override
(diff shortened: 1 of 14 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/routing/RoutingPolicies.java
```diff
@@ -133,7 +134,7 @@ public void setRoutingStatus(ZoneId zone, RoutingStatus.Value value) {
                                                                                        controller.clock().instant())));
             Map<ApplicationId, RoutingPolicyList> allPolicies = readAll().groupingBy(policy -> policy.id().owner());
             for (var instancePolicies : allPolicies.values()) {
-                updateGlobalDnsOf(instancePolicies, Set.of(), lock);
+                updateGlobalDnsOf(instancePolicies, Set.of(), Optional.empty(), lock);
```
jonmv, line 137: This is the only one I identified as an operator/cross-application event. I'm not even sure we need this distinction, though, because this also happens per application ... so we could probably just do all of these like usual, too? That would simplify a bit.

jonmv: OTOH, this _is_ the right treatment, should any owner-less request appear in the queue, so perhaps leave it like this?

jonmv: (Could alternatively assign an owner to each of the requests generated by taking zones in/out.)

mpolden: Agree that we don't need this. Preserving the order per application is sufficient.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/NameServiceDispatcher.java
```diff
@@ -36,20 +44,24 @@ protected double maintain() {
         int requestCount = trueIntervalInSeconds();
         try (var lock = db.lockNameServiceQueue()) {
             var queue = db.readNameServiceQueue();
+            if (queue.requests().isEmpty() || requestCount == 0) return 1.0;
+
             var instant = clock.instant();
             var remaining = queue.dispatchTo(nameService, requestCount);
-            if (queue.equals(remaining)) return 1.0; // Queue unchanged
+            var dispatched = queue.requests().stream()
+                                  .filter(new HashSet<>(remaining.requests())::remove)
```
jonmv, line 52: missing negation, fixed in next commit

jonmv:
Last commit has NSD release the NSQ lock while dispatching, to avoid hogging the RP lock. Since only NSD should remove from the queue, and only one NSD should run at a time (locked on maintainer level), the assumption that the queue has only been prepended or appended to should hold. I think logging a warning, and doing a best-effort amendment when this assumption fails, is OK.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/maintenance/NameServiceDispatcherTest.java
```diff
@@ -0,0 +1,122 @@
...
+        tester.curator().writeNameServiceQueue(new NameServiceQueue(base));
+        expectations.clear();
+        expectations.add(name -> {
+            assertEquals(rec1.name(), name);
+            tester.curator().writeNameServiceQueue(new NameServiceQueue(List.of(req1, req2, req3, req4, req6, req7, req8)));
+        });
+        expectations.add(failOwner1.get()); // Recursively adds itself to the tail of the expectation queue.
+        expectations.add(name -> assertEquals(rec3.name(), name));
+        assertEquals(0.5, dispatcher.maintain()); // 2 of 4 requests were ok (the fourth was never attempted).
+        assertEquals(List.of(req2, req3, req4, req6, req7, req8), tester.curator().readNameServiceQueue().requests());
```
jonmv, line 119: Note the "error" of removing req1 instead of req3, which is really the same request. If req2 was req1's inverse, this would be significant. The odds of this happening seem _very_ low.

mpolden:
> Last commit has NSD release the NSQ lock while dispatching, to avoid hogging the RP lock. Since only NSD should remove from the queue, and only one NSD should run at a time (locked on maintainer level), the assumption that the queue has only been prepended or appended to, should hold. I think logging a warning, and doing a best-effort amendment when this assumption fails, is OK.

I don't think this is worth the extra complexity and assumptions about other locks. Current locking semantics has worked fine for years. If there is lock contention due to a bug in the dispatching process it's better if everything blows up rather than to keep pushing more things onto the queue. Deployments don't work anyway if we can't do DNS updates in a timely manner.

jonmv:
The problem is that not just deployment are stalled, but _all_ modifications to application metadata are blocked, once a deployment is attempted, fails and blocks due to this, because the application lock is also held while holding the routing policies lock. I think it's worthwhile to avoid this complete deadlocking. We could look for all places where we take both locks, and see if we can eliminate them, instead/as well.

jonmv:
Ah, the NSQ lock is also taken directly other places, while hold application locks. Not sure how hard it is to remove all those, or if it's feasible at all.

mpolden (approved):

PR #25656, opened 2023-01-20
Inline apache http client artifacts in container-apache-http-client-bundle

bjorncs:
Allows container-apache-http-client-bundle to be used on classpath for classic fatjars. Since the bundle is now built with Felix's bundle plugin, there is no need to depend on jdisc_core or manually export through `@PublicApi` annotations.

### DIFF
--- a/container-apache-http-client-bundle/pom.xml
+++ b/container-apache-http-client-bundle/pom.xml
@@ -6,7 +6,7 @@
    http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <artifactId>container-apache-http-client-bundle</artifactId>
-    <packaging>container-plugin</packaging>
+    <packaging>jar</packaging>
     <version>8-SNAPSHOT</version>
 
     <parent>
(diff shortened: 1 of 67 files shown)


baldersheim (approved):

container-apache-http-client-bundle/pom.xml
```diff
@@ -68,9 +48,72 @@
...
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
                 <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Bundle-Version>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}.${parsedVersion.incrementalVersion}</Bundle-Version>
+                        <Export-Package>org.apache.hc.*;version=1.0.0;-noimport:=true,org.apache.http.*;version=1.0.0;-noimport:=true</Export-Package>
+                        <Embed-Dependency>*;scope=compile;type=!pom;inline=true</Embed-Dependency>
+                        <Import-Package>!javax.servlet,javax.*,org.slf4j,org.ietf.jgss</Import-Package>
```
gjoranv, line 72: javax.* seems a bit risky? How future proof is this?

bjorncs: It's importing only `javax` packages part of the standard library atm.

PR #25663, opened 2023-01-20
Add query trace parameters for profiling backend query evaluation.

geirst:
With this change, profiling of matching, first-phase ranking, and second-phase ranking can be tuned separately.

@bratseth please review
@havardpe FYI


bratseth (approved):
👍  great!

PR #25665, opened 2023-01-20
Just do nothing on-prem, as no access types are supported, but do not…

jonmv:
… kill wildcard use

This allows someone with deployments in Yahoo and AWS to specify private endpoints with a wildcard setting, without killing new deployments to Yahoo.

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
@@ -504,7 +504,6 @@ public class ApplicationController {
                 instance = application.get().require(job.application().instance());
 
                 containerEndpoints = controller.routing().of(deployment).prepare(application);
-
             } // Release application lock while doing the deployment, which is a lengthy task.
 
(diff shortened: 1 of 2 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java
```diff
@@ -867,9 +866,11 @@ public void setSuspension(DeploymentId deploymentId, boolean suspend) {
     /** Deactivate application in the given zone. Even if the application itself does not exist, deactivation of the deployment will still be attempted */
     public void deactivate(ApplicationId instanceId, ZoneId zone) {
         TenantAndApplicationId applicationId = TenantAndApplicationId.from(instanceId);
-        try (Mutex lock = lock(applicationId)) {
-            Optional<LockedApplication> application = getApplication(applicationId).map(app -> new LockedApplication(app, lock));
-            deactivate(instanceId, zone, application).ifPresent(this::store);
+        try (Mutex deploymentLock = lockForDeployment(instanceId, zone)) {
```
jonmv, line 869: Not related, but a tiny bug that should be fixed anyway. I doubt it often matters.

freva (approved):

PR #25667, opened 2023-01-21
Support a group size constraint in content clusters

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
```diff
@@ -39,8 +39,11 @@ public CapacityPolicies(NodeRepository nodeRepository) {
     }
 
     public Capacity applyOn(Capacity capacity, ApplicationId application, boolean exclusive) {
-        return capacity.withLimits(applyOn(capacity.minResources(), capacity, application, exclusive),
-                                   applyOn(capacity.maxResources(), capacity, application, exclusive));
+        var min = applyOn(capacity.minResources(), capacity, application, exclusive);
+        var max = applyOn(capacity.maxResources(), capacity, application, exclusive);
+        var groupSize = capacity.groupSize().fromAtMost(max.nodes() / max.groups())
```
freva, line 44: Shouldn't this be `max.nodes() / min.groups()` ?

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
bratseth: Good question! I think that is slightly worse in the cases where the other limits are moved, which is what this is for (e.g we have configured that we want multiple large groups, this will hard prioritize _large_ over _multiple_ in perf), but otoh, it avoids ever moving this constraint in prod, which we don't want, so the best alternative is something a bit more complicated, so: Yes.

PR #25671, opened 2023-01-22
Move IntRange out of vespajlib

bratseth:
vespajlib classes in the interface between config model and config server causes LinkageErrors.


hmusum (commented):
👍

PR #25681, opened 2023-01-23
Revert "switch to "go" version of vespa-start-services"

hmusum:
Reverts vespa-engine/vespa#25676

Core dump test fails, could be related to setting of core dump size

FYI @tokle


arnej27959 (approved):
👍

PR #25685, opened 2023-01-23
Exclude all org.slf4j artifacts from jetty in container-test.

gjoranv:
- For unknown reasons, user projects using a base pom could get slf4j-api in compile scope when the test scoped dep from container-test was nearer the dep tree root than the provided scoped dep from 'container' (via container-dev and jdisc_core).


baldersheim (approved):

bjorncs (commented):
👍

PR #25687, opened 2023-01-23
Bratseth/redundancy validation

bratseth:


config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/RedundancyChangeValidatorTest.java
```diff
@@ -0,0 +1,65 @@
+package com.yahoo.vespa.model.application.validation.change;
```
geirst, line 1: Please add copyright.

geirst (approved):
👍

config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/RedundancyChangeValidatorTest.java
bratseth: I have a script doing that for all files missing it and I hereby commit to doing another round of it soon.

PR #25688, opened 2023-01-23
Balder/also handle non overlapping bitvectors

baldersheim:
@toregge @havardpe @geirst @vekterli

### DIFF
--- a/searchlib/src/tests/common/bitvector/bitvector_test.cpp
+++ b/searchlib/src/tests/common/bitvector/bitvector_test.cpp
@@ -304,9 +304,6 @@ TEST("requireThatSequentialOperationsOnPartialWorks")
 
     PartialBitVector after(1000, 1100);
     after.setInterval(1000, 1100);
-    EXPECT_EXCEPTION(p2.orWith(after), vespalib::IllegalArgumentException,
-                     "IllegalArgumentException: [717, 919] starts before which is not allowed currently [1000, 1100] at "
-                     "verifyInclusiveStart in /home/balder/git/vespa/searchlib/src/vespa/searchlib/common/bitvector.cpp:33");
     EXPECT_EQUAL(202u, p2.countTrueBits());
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/common/bitvector.cpp
```diff
@@ -90,10 +90,15 @@ void
 BitVector::clearInterval(Index start, Index end)
 {
     clearIntervalNoInvalidation(Range(start, end));
-
     invalidateCachedCount();
 }
 
+void
+BitVector::store(Word &word, Word value) {
+    assert(!_enable_range_check && (&word >= getActiveStart()));
```
arnej27959, line 98: Looks like this should be || also should we check end in addition to start?

arnej27959:
merging this now as it's likely to be an improvement at least; please review after the fact.

searchlib/src/tests/common/bitvector/bitvector_test.cpp
```diff
@@ -365,9 +362,13 @@ verifyNonOverlappingWorksAsZeroPadded(bool clear, Func func) {
     BitVector::UP right = createEveryNthBitSet(3, 2000, 100);
     EXPECT_EQUAL(CNT, left->countTrueBits());
     EXPECT_EQUAL(CNT, right->countTrueBits());
-    EXPECT_EXCEPTION(func(*left, *right), vespalib::IllegalArgumentException,
-                     "IllegalArgumentException: [1000, 1100] starts before which is not allowed currently [2000, 2100] at "
-                     "verifyInclusiveStart in /home/balder/git/vespa/searchlib/src/vespa/searchlib/common/bitvector.cpp:33");
```
baldersheim: Thanks, did not notice the full path :)

havardpe (commented):
bitvector create probably has the same issue as well...

PR #25708, opened 2023-01-24
Do not filter on error code

mpolden:
`addCUDA` can return other error codes, coming from the native code, e.g. if you
pass an invalid device number.

@arnej27959


arnej27959 (approved):
👍

PR #25709, opened 2023-01-24
ignore ._foo.sd and other "resource fork" like names

arnej27959:
@bratseth please review
@kkraune FYI


config-application-package/src/main/java/com/yahoo/config/model/application/AbstractApplicationPackage.java
```diff
@@ -41,4 +41,18 @@ public Map<String,String> legacyOverrides() {
         return result;
     }
 
+    public static boolean validSchemaFilename(String fn) {
```
geirst, line 44: We should add unit testing of this helper function.

bratseth (approved):
👍

PR #25712, opened 2023-01-24
Add new KeywordExtractor with two factories (one each for indexed search

toregge:
and streaming search).

@geirst : please review


geirst (approved):
👍

PR #25722, opened 2023-01-25
Hmusum/file registry refactoring

hmusum:
Minor refactoring


configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDBRegistry.java
```diff
@@ -105,7 +104,7 @@ public synchronized FileReference addUri(String uri) {
     }
 
     @Override
-    public FileReference addBlob(String blobName, ByteBuffer blob) {
+    public synchronized FileReference addBlob(String blobName, ByteBuffer blob) {
         String relativePath = blobToRelativeFile(blobName);
         synchronized (this) {
```
baldersheim, line 112: Not strictly necessary as the blobToRelativeFile is a static method, but more in line with the other synchronized methods.
But you should remove synchronized (this).

baldersheim (changes requested):
Remove the redundant synchronized (this) in addBlob.

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileDBRegistry.java
hmusum: Ah, yes, will fix

baldersheim (approved):

PR #25731, opened 2023-01-25
- It is a bit of an overkill to use java Stack that is built on java …

baldersheim:
…Vector which is synchronized.

  Especially when you only need to store a boolean.
- Use a Deque instead.

@bratseth or @bjorncs PR


bratseth (approved):
👍

bjorncs (approved):
🎖️

PR #25735, opened 2023-01-25
Validate rank profiles early

bratseth:
This is to warn about missing rank profiles before that leads to consequences which emits harder to understand errors.


arnej27959 (approved):
looks good to me

PR #25742, opened 2023-01-26
Remove unstable test that tests a class that is only used in unit tests

hmusum:


baldersheim (approved):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/rpc/RpcServer.java
```diff
@@ -35,6 +35,7 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+// TODO: RpcServer is only used in unit tests, should be removed
```
baldersheim, line 38: If only use in test code locally in this module then it should move from main => test

hmusum: `RpcServer` is injected into `FleetController` (but it is null except from in tests), so getting rid of RpcServer or moving it to test requires rewriting almost all unit tests and test helper classes to use something else than RpcServer. This is a lot of work, so not done now.

vekterli (approved):
:+1: away it goes!

PR #25744, opened 2023-01-26
Fix unstable test

hmusum:
Seems like the logic was flipped in https://github.com/vespa-engine/vespa/pull/21372
(`isClosed()` is not equal `isValid()`)


jrt/tests/com/yahoo/jrt/ConnectTest.java
```diff
@@ -25,12 +24,12 @@ public void testConnect() throws ListenFailedException {
         target.close();
 
         for (int i = 0; i < 100; i++) {
-            if (!target.isClosed()) {
+            if (target.isClosed()) {
```
baldersheim, line 27: Ouch, good catch...

baldersheim (approved):

PR #25746, opened 2023-01-26
Skip CUDA entirely if GPU device is optional

mpolden:
CUDA may fail after its library is loaded, e.g. when the session is created.

@arnej27959


arnej27959 (approved):
looks correct to me

PR #25761, opened 2023-01-27
Support embedding an array to a mixed 2d tensor

bratseth:
@geirst and @jobergum fyi

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
@@ -88,6 +88,7 @@ public class TensorType {
     private final List<Dimension> dimensions;
 
     private final TensorType mappedSubtype;
+    private final TensorType indexedSubtype;
 
     public TensorType(Value valueType, Collection<Dimension> dimensions) {
         this.valueType = valueType;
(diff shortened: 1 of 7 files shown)


lesters (approved):
Great!

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
```diff
@@ -135,6 +142,9 @@ public static TensorType fromSpec(String specString) {
     /** The type representing the mapped subset of dimensions of this. */
     public TensorType mappedSubtype() { return mappedSubtype; }
 
+    /** The type representing the indexed subset of dimensions of this. */
+    public TensorType indexedSubtype() { return indexedSubtype; }
```
geirst, line 146: @bratseth Consider adding test for this (and mappedSubtype()) in TensorTypeTestCase.java.

geirst (commented):
Great 👍

vespajlib/src/main/java/com/yahoo/tensor/TensorType.java
bratseth: will do

jobergum:
This is awesome 💯 🏅

PR #25763, opened 2023-01-27
Use original source for dynamic summary transform.

toregge:
@geirst : please review


config-model/src/main/java/com/yahoo/schema/processing/DynamicSummaryTransformUtils.java
```diff
@@ -50,10 +52,20 @@ public static boolean summaryFieldIsRequiredInDocumentType(SummaryField summaryF
         return summaryFieldIsPopulatedBySourceField(summaryField.getDataType());
     }
 
-    public static String getSource(SummaryField summaryField) {
+    public static String getSource(SummaryField summaryField, Schema schema) {
         // Summary fields with the original supported type is always present in the document type,
         // and we must use that field as source at run-time.
-        return isOriginalSupportedType(summaryField.getDataType()) ?
-                summaryField.getName() : summaryField.getSingleSource();
+        if (isOriginalSupportedType(summaryField.getDataType())) {
```
geirst, line 58: The comment must be updated. Suggestion: 
_Summary fields with the original supported type is always present in the document type.
However, if the source of that summary field is a single explicit source that exists in the schema we
use that as source instead as this is handled by the backend code.
This is a move in the right direction to avoid adding some summary fields as extra document fields._

geirst (approved):
👍

config-model/src/main/java/com/yahoo/schema/processing/DynamicSummaryTransformUtils.java
toregge: Comment has been updated.

geirst (approved):
👍

PR #25780, opened 2023-01-28
Disable dangling-reference and overloaded-virtual warnings when using gcc 13

toregge:
@baldersheim : please review

### DIFF
--- a/build_settings.cmake
+++ b/build_settings.cmake
@@ -80,6 +80,13 @@ else()
   set(VESPA_XXHASH_DEFINE "-DXXH_INLINE_ALL")
 endif()
 
+# Disable dangling reference and overloaded virtual warnings when using gcc 13
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "13")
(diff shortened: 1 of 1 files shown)


baldersheim (approved):

baldersheim:
I assume we must to a code cleanup before going to gcc 13.

build_settings.cmake
```diff
@@ -80,6 +80,13 @@ else()
   set(VESPA_XXHASH_DEFINE "-DXXH_INLINE_ALL")
 endif()
 
+# Disable dangling reference and overloaded virtual warnings when using gcc 13
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "13")
+    set(CXX_SPECIFIC_WARN_OPTS "${CXX_SPECIFIC_WARN_OPTS} -Wno-dangling-reference -Wno-overloaded-virtual")
```
baldersheim, line 86: What is the issue here ?
Are these new warnings, or are they interpreted differently ?

toregge: With gcc 13, the `-Wall` option implies both the new `-Wdangling-reference` option and the `-Woverloaded-virtual=1` option. gcc 12 had `-Woverloaded-virtual`, but without levels, and the `-Wall` option did not enable that warning.

PR #25797, opened 2023-01-30
Let json be default explicit

baldersheim:
@vekterli PR


vekterli (commented):
Looks like a unit test has to be updated first, otherwise LGTM

vekterli (approved):

PR #25805, opened 2023-01-30
General metrics

gjoranv:
Enums in Java cannot be extended, as they are most commonly used to encapsulate a finite set of known values. Using an ordinary class would require adding boilerplate code to each declaration. The downside of the chosen solution is that the name/description/unit fields and methods must be copy/pasted to each enum.


kkraune (approved):
Great!

PR #25808, opened 2023-01-30
Various improvements to avoid excessive scaling

bratseth:
- Don't scale if there are provisioned nodes
- Don't scale if less than 5 minutes since last time
- Metrics are stable unless the cluster is actually retiring


freva (approved):
Nodes are never in `provisioned`? Hosts are created in `provisioned` and nodes are created in `reserved`.

bratseth:
Yes, but it's the same category of states.

PR #25817, opened 2023-01-31
Revert "Use aligned_alloc instead of memalign."

baldersheim:
Reverts vespa-engine/vespa#25813


baldersheim:
@toregge or @vekterli PR
Need vespamalloc support first.

vekterli (commented):
:+1:

PR #25826, opened 2023-01-31
Revert "Revert "Use aligned_alloc instead of memalign.""

baldersheim:
Reverts vespa-engine/vespa#25817
@toregge or @vekterli PR


baldersheim:
Should be good to go now.

toregge (commented):
👍

PR #25832, opened 2023-02-01
add parsing (only) of container-phase in schema

arnej27959:
@bratseth please review

(merge delayed until after architect meeting)


bratseth (approved):
👍

PR #25836, opened 2023-02-01
Fail closed when no core dump encryption public key is found

vekterli:
@freva or @hakonhall please review

### DIFF
--- a/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
+++ b/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
@@ -289,7 +289,7 @@ public class CoredumpHandler {
                 .filter(k -> !k.isEmpty())
                 .map(KeyId::ofString)
                 .flatMap(secretSharedKeySupplier::create)
-                .orElseThrow(() -> new IllegalStateException("No core dump encryption key provided"));
+                .orElseThrow(() -> ConvergenceException.ofError("No core dump encryption key provided"));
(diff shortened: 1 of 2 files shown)


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/maintenance/coredump/CoredumpHandler.java
```diff
@@ -287,11 +285,12 @@ void processAndReportSingleCoreDump(NodeAgentContext context, ContainerPath core
...
         dockerImage.flatMap(DockerImage::tag).ifPresent(metadata::setVespaVersion);
         dockerImage.ifPresent(metadata::setDockerImage);
-        Optional<SecretSharedKey> sharedCoreKey = Optional.of(corePublicKeyFlagValue(context))
+        SecretSharedKey sharedCoreKey = Optional.of(corePublicKeyFlagValue(context))
                 .filter(k -> !k.isEmpty())
                 .map(KeyId::ofString)
-                .flatMap(secretSharedKeySupplier::create);
-        sharedCoreKey.map(key -> key.sealedSharedKey().toTokenString()).ifPresent(metadata::setDecryptionToken);
+                .flatMap(secretSharedKeySupplier::create)
+                .orElseThrow(() -> new IllegalStateException("No core dump encryption key provided"));
```
freva, line 292: Please use `ConvergenceException.ofError()` (this exception is specially handled in host-admin to not print stack trace, used when the error is obvious from the message alone)

vekterli: Fixed

freva (approved):

hakonhall (approved):

baldersheim (approved):

PR #25849, opened 2023-02-02
Collect GPU metrics

mpolden:


mpolden:
Merge with internal PR.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/ContainerStatsCollector.java
```diff
@@ -1,54 +1,87 @@
...
+                                                       "nvidia-smi",
+                                                       "--query-gpu=index,utilization.gpu,memory.total,memory.free",
+                                                       "--format=csv,noheader,nounits")
+                                              .getOutputLinesStream();
+        return lines.map(ContainerStatsCollector::parseGpuStats).toList();
+    }
+
+    private static ContainerStats.GpuStats parseGpuStats(String s) {
+        String[] fields = fields(s, ",\\s*");
+        if (fields.length < 4) throw new IllegalArgumentException("Could not parse GPU stats from '" + s + "'");
```
freva, line 75: Throw `ConvergenceException.ofError()` instead

freva (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/ContainerStatsCollector.java
mpolden: This is in the "should never happen" class of exceptions, so leaving it as-is.

PR #25851, opened 2023-02-02
make clang happy

havardpe:
Avoid checking thread id more than once in coroutines, since clang will assume it cannot change. Also set return values from the outside of the coroutine instead of trying to infer it from the inside.

Fixed a race in async_run fuctions ('this' is not safe to use after giving the coroutine away). This goes for all compilers.

Minor fix to io_uring connect to avoid returning valid non-connected socket if connect was called after AsyncIo shutdown.

Stop using std::move when co_returning local move-only values. clang 15 likes this, but it might not make all compilers more happy...

@geirst please review


geirst (approved):
👍

PR #25852, opened 2023-02-02
Handle negative numeric labels.

baldersheim:
@bratseth or @bjormel or @bjorncs PR


bratseth (approved):
I don't think we have ever claimed to support negative labels, but ok.

PR #25854, opened 2023-02-02
make gcc 10 happy

havardpe:
avoid using atomic notify/wait

do not put co_await inside other constructs (for/if)

@toregge please review


toregge (approved):

baldersheim:
Why do we want to support gcc 10.
I think 11 should do just fine.

PR #25857, opened 2023-02-03
- Use T && f() && to avoid moving temporaries.

baldersheim:
- std::make_unique/make_shared

@vekterli PR
This addresses the comments in #25844

### DIFF
--- a/messagebus/src/vespa/messagebus/testlib/custompolicy.cpp
+++ b/messagebus/src/vespa/messagebus/testlib/custompolicy.cpp
@@ -12,16 +12,15 @@ LOG_SETUP(".custompolicy");
 namespace mbus {
 
 CustomPolicy::CustomPolicy(bool selectOnRetry,
-                           const std::vector<uint32_t> consumableErrors,
-                           const std::vector<Route> &routes) :
+                           std::vector<uint32_t> consumableErrors,
+                           std::vector<Route> routes) :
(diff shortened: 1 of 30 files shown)


messagebus/src/vespa/messagebus/testlib/custompolicy.cpp
```diff
@@ -38,10 +37,8 @@ CustomPolicy::select(RoutingContext &context)
     str.append(" }.");
     context.trace(1, str);
     context.setSelectOnRetry(_selectOnRetry);
-    for (std::vector<uint32_t>::iterator it = _consumableErrors.begin();
-         it != _consumableErrors.end(); ++it)
-    {
-        context.addConsumableError(*it);
+    for (unsigned int & _consumableError : _consumableErrors) {
```
vekterli, line 40: Would it be better to leave this as `for (uint32_t ...` to ensure same type as the underlying vector (or just `auto` it)? Also consider removing ref to primitive and leading underscore on temporary name

vekterli (approved):
:+1:

messagebus/src/vespa/messagebus/testlib/custompolicy.cpp
baldersheim: clion automatic fix takes a few short cuts...

PR #25866, opened 2023-02-03
Re-organize Go code

mpolden:
As discussed. New structure:

```
$ tree -L 2 client/go
client/go
├── Makefile
├── README.md
├── cond_make.go
├── go.mod
├── go.sum
└── internal
    ├── admin
    ├── cli
    ├── curl
    ├── mock
    ├── util
    ├── version
    └── vespa
```

Putting everything in `internal` prevents other projects from pulling in on our
code as a dependency. If we want to have a public API at some point, we can pull
the relevant things up one level, next to `internal`.

`admin` holds the code for internal Vespa tools. `cli` holds Vespa CLI. Other
packages below `internal` holds common code that is not specific to either
`admin` or `cli`.

@arnej27959


arnej27959:
this looks good to me; after I get a better grasp of the "cli" code I will probably try to make more modules common (e.g. envvars), but it's best to separate it first.

kkraune:
hehe - maybe - I cannot say as I don't yet fully understand the vespa-cli codebase. but it would make me very happy if I was able to reuse (as in import) some of its functions by just building the TLS options myself (e.g. in a google function / aws lambda get pem secrets from the environment), provide an endpoint, and call a function (e.g. visit), with minimal dependencies. my code would be tiny and based on the Vespa Team's collective wisdom. but not sure if feasible and/or if this PR moves us closer to that. thanks for looping me in, @arnej27959

mpolden:
I agree that we should have a public Go API at some point. However, right now all packages are public by default and we definitely don't want that.

kkraune:
All right, lets try this :-)

kkraune (approved):

PR #25868, opened 2023-02-03
Replace dimension allowlisting with blocklist in /metrics/v2

olaaun:


metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/dimensions/BlocklistDimensions.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class BlocklistDimensions {
+
+    private BlocklistDimensions() {}
+
+    /**
+     * Deployment related metrics - most of which are redundant
+     * E.g. app/applicationName/tenantName/instanceName is already included in applicationId
+     */
+    public static final String APP = "app";
```
gjoranv, line 17: You could avoid "public static final" for each constant if you use an enum with a constructor that takes a String. I think you could also avoid repeating all the constants by using EnumSet.all().

gjoranv (approved):
Great! Please consider my comment about using enum.

gjoranv (commented):
❤️

PR #25870, opened 2023-02-03
Use the simple tokenizer, no need to pull in boost

baldersheim:
@vekterli or @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
+++ b/searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
@@ -8,8 +8,7 @@
 #include <vespa/vespalib/geo/zcurve.h>
 #include <vespa/vespalib/util/issue.h>
 #include <vespa/vespalib/util/stash.h>
-#include <boost/algorithm/string/split.hpp>
-#include <boost/algorithm/string/classification.hpp>
+#include <vespa/vespalib/text/stringtokenizer.h>
 #include <cmath>
 
 #include <vespa/log/log.h>
@@ -50,9 +49,9 @@ DistanceToPathExecutor::execute(uint32_t docId)
             double len = std::sqrt(len2);
 
             // For each document location, do
-            for (uint32_t loc = 0; loc < _intBuf.size(); ++loc) {
+            for (long loc : _intBuf) {
                 int32_t x = 0, y = 0;
-                vespalib::geo::ZCurve::decode(_intBuf[loc], &x, &y);
+                vespalib::geo::ZCurve::decode(loc, &x, &y);
 
                 double u = 0, dx, dy;
                 if (len < 1e-6) {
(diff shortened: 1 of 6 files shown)


toregge (approved):

searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
```diff
@@ -50,9 +49,9 @@ DistanceToPathExecutor::execute(uint32_t docId)
             double len = std::sqrt(len2);
 
             // For each document location, do
-            for (uint32_t loc = 0; loc < _intBuf.size(); ++loc) {
+            for (long loc : _intBuf) {
```
vekterli, line 52: The underlying integer type is determined via transitive typedefs; should this instead be `auto` to always be in sync with any changes (unlikely but possible)?

searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
```diff
@@ -132,17 +131,16 @@ DistanceToPathBlueprint::createExecutor(const search::fef::IQueryEnvironment &en
...
-        vespalib::string str = pro.getAt(0);
+        vespalib::stringref str = pro.getAt(0);
         uint32_t len = str.size();
         if (str[0] == '(' && len > 1 && str[len - 1] == ')') {
             str = str.substr(1, len - 1); // remove braces
-            std::vector<vespalib::string> arr;
-            boost::split(arr, str, boost::is_any_of(","));
-            len = arr.size() - 1;
+            vespalib::StringTokenizer tokenizer(str);
+            len = tokenizer.size() - 1;
```
vekterli, line 139: Can the contents of the `()` ever be empty/just whitespace? `vespalib::StringTokenizer` implicitly prunes away a token list consisting of a single empty token, so it seems that an input of `str == "()"` would cause `tokenizer.size() == 0` which in turn would cause an integer underflow. Consider checking for empty tokens case, or `assert` if it should never happen.

vespalib/src/vespa/vespalib/text/stringtokenizer.h
```diff
@@ -41,25 +41,34 @@ class StringTokenizer {
      * @param separators The characters to be used as token separators
      * @param strip      Characters to be stripped from both ends of each token
      **/
+    explicit StringTokenizer(vespalib::stringref source)
+        : StringTokenizer(source, ",")
+    {}
     StringTokenizer(vespalib::stringref source,
-                    vespalib::stringref separators = ",",
-                    vespalib::stringref strip = " \t\f\r\n");
+                             vespalib::stringref separators)
```
vekterli, line 48: Nit: indentation

searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
baldersheim: Fixed in #25883

searchlib/src/vespa/searchlib/features/distancetopathfeature.cpp
baldersheim: Fixed in #25883

vespalib/src/vespa/vespalib/text/stringtokenizer.h
baldersheim: Fixed in #25883

PR #25872, opened 2023-02-04
Avoid boost operators.

baldersheim:
@toregge or @vekterli PR

### DIFF
--- a/storage/src/vespa/storageframework/generic/clock/time.h
+++ b/storage/src/vespa/storageframework/generic/clock/time.h
@@ -1,7 +1,6 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #pragma once
 
-#include <boost/operators.hpp>
 #include <vespa/vespalib/stllike/string.h>
 #include <vespa/vespalib/util/time.h>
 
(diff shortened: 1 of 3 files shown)


toregge (approved):

storage/src/vespa/storageframework/generic/clock/time.h
```diff
@@ -45,37 +44,34 @@ uint64_t getRawMicroTime(const Clock&);
  * automatic conversion between time types.
  */
 template<typename Type, int MicrosPerUnit>
-class Time : public boost::operators<Type>
+class Time
```
vekterli, line 47: Looks like the majority of member functions here could safely be tagged as `noexcept`

storage/src/vespa/storageframework/generic/clock/time.h
```diff
@@ -170,4 +166,39 @@ inline MicroSecTime MilliSecTime::getMicros() const {
...
+    result -= b;
+    return result;
+}
+
+inline SecondTime
+operator - (SecondTime a, SecondTime b) {
+    SecondTime result(a);
+    result -= b;
+    return result;
+}
```
vekterli, line 202: These could probably be generalized as a single overload per operator, using the "micros per unit" template arg to properly normalize values. But that's basically what `std::chrono` already does, so no point if we're already replacing with that.

baldersheim: That happens further down the PR pipeline.

storage/src/vespa/storageframework/generic/clock/time.h
baldersheim: That happens further down the PR pipeline.

PR #25877, opened 2023-02-06
Rename script-utils -> vespa-wrapper

mpolden:
To make it more descriptive when installed locally.

@arnej27959


arnej27959 (approved):
Looks good to me

PR #25881, opened 2023-02-06
Add a `vespa.none` sentinel capability to C++ capability code

vekterli:
@bjorncs please review

Allows for configuring rules for peers that should pass mTLS handshakes but should not be able to access any capability-restricted APIs.

### DIFF
--- a/vespalib/src/vespa/vespalib/net/tls/policy_checking_certificate_verifier.cpp
+++ b/vespalib/src/vespa/vespalib/net/tls/policy_checking_certificate_verifier.cpp
@@ -61,7 +61,7 @@ public:
 
     ~PolicyConfiguredCertificateVerifier() override;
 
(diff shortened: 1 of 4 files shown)


vespalib/src/vespa/vespalib/net/tls/policy_checking_certificate_verifier.cpp
```diff
@@ -61,7 +61,7 @@ class PolicyConfiguredCertificateVerifier : public CertificateVerificationCallba
 
     ~PolicyConfiguredCertificateVerifier() override;
 
-    VerificationResult verify(const PeerCredentials& peer_creds) const override;
+    [[nodiscard]] VerificationResult verify(const PeerCredentials& peer_creds) const override;
```
bjorncs, line 64: I wish this annotation was available in Java...

bjorncs (approved):

PR #25884, opened 2023-02-06
Make attribute collection spec serial num optional. Pass reconfig

toregge:
serial num explicitly to various reconfigure member functions since it might not be available from attribute collection spec.

@geirst : please review


geirst (approved):
👍

PR #25886, opened 2023-02-06
Revert "Make attribute collection spec serial num optional. Pass reconfig"

baldersheim:
Reverts vespa-engine/vespa#25884
@geirst or @toregge PR
Breaks some system tests.


toregge (commented):
:+1:

PR #25874, opened 2023-02-05
Øess use of getSeconds/getMicroSeconds/getMilliSeconds.

baldersheim:
@vekterli PR

PR #25893, opened 2023-02-07
Revert "Øess use of getSeconds/getMicroSeconds/getMilliSeconds."

baldersheim:
Reverts vespa-engine/vespa#25874
@vekterli PR
Must be splitt up to identify reason for performance regression.


vekterli (commented):
:+1:

PR #25898, opened 2023-02-07
Use steady_time for vdslib::NodeState MERGEOK

baldersheim:
@vekterli PR

### DIFF
--- a/vdslib/src/vespa/vdslib/state/nodestate.cpp
+++ b/vdslib/src/vespa/vdslib/state/nodestate.cpp
@@ -8,9 +8,8 @@
 #include <vespa/vespalib/util/exceptions.h>
 #include <vespa/vespalib/stllike/asciistream.h>
 #include <sstream>
-#include <cmath>
-#include <vespa/log/log.h>
 
+#include <vespa/log/log.h>
 LOG_SETUP(".vdslib.nodestate");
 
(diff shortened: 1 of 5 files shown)


vdslib/src/vespa/vdslib/state/nodestate.cpp
```diff
@@ -56,25 +55,24 @@ NodeState::NodeState(vespalib::stringref serialized, const NodeType* type)
...
       _minUsedBits(16),
-      _startTimestamp(0)
+      _startTimestamp()
 {
 
     vespalib::StringTokenizer st(serialized, " \t\f\r\n");
     st.removeEmptyTokens();
-    for (vespalib::StringTokenizer::Iterator it = st.begin();
-         it != st.end(); ++it)
+    for (auto it : st)
```
vekterli, line 63: Consider using a non-generic name for loop variable (`token`?)

Additional nit: consider moving brace on next line up onto this line since this is now a single-line statement

vekterli (commented):
:+1:

vdslib/src/vespa/vdslib/state/nodestate.cpp
baldersheim: Thanks, will fix when reapplying

PR #25904, opened 2023-02-07
Use std::chrono based time for bucket manager time-to-work check.

baldersheim:
@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/bucketdb/bucketmanager.cpp
+++ b/storage/src/vespa/storage/bucketdb/bucketmanager.cpp
@@ -279,8 +279,8 @@ void BucketManager::updateMinUsedBits()
 // Responsible for sending on messages that was previously queued
 void BucketManager::run(framework::ThreadHandle& thread)
 {
-    const int64_t CHECK_MINUSEDBITS_INTERVAL = 1000*30;
-    framework::MilliSecTime timeToCheckMinUsedBits(0);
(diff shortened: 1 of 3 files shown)


storage/src/vespa/storage/bucketdb/bucketmanager.cpp
```diff
@@ -305,10 +305,9 @@ void BucketManager::run(framework::ThreadHandle& thread)
                 thread.registerTick(framework::PROCESS_CYCLE);
             }
         }
-        if (timeToCheckMinUsedBits < _component.getClock().getTimeInMillis()) {
+        if (timeToCheckMinUsedBits < _component.getClock().getMonotonicTime()) {
             updateMinUsedBits();
-            timeToCheckMinUsedBits = _component.getClock().getTimeInMillis();
-            timeToCheckMinUsedBits += framework::MilliSecTime(CHECK_MINUSEDBITS_INTERVAL);
+            timeToCheckMinUsedBits = _component.getClock().getMonotonicTime() + CHECK_MINUSEDBITS_INTERVAL;
```
vekterli, line 310: Should we change this to only read the clock once?

baldersheim: As long as duration of updateMinUsedBits is sane (much less than CHECK_MINUSEDBITS_INTERVAL).

PR #25905, opened 2023-02-07
Revert "Use steady_time for vdslib::NodeState MERGEOK"

baldersheim:
Reverts vespa-engine/vespa#25898
@vekterli PR
This is the culprit of performance regression.


vekterli (commented):
:+1:

PR #25910, opened 2023-02-07
Support configuring BCP structure

bratseth:
Not added to deployment schema spec yet on purpose.


bratseth:
Some kind of incorrect merging happening, I'll rebase.

PR #25912, opened 2023-02-07
Bratseth/bcp rebased

bratseth:


config-model-api/src/main/java/com/yahoo/config/application/api/Bcp.java
```diff
@@ -0,0 +1,122 @@
...
+            return "BCP group of " + members;
+        }
+
+    }
+
+    public record RegionMember(RegionName region, double fraction) {
+
+        public RegionMember {
+            if (fraction < 0 || fraction > 1)
+                throw new IllegalArgumentException("Fraction must be a number between 0.0 and 1.0, but got " + fraction);
```
mpolden, line 116: Consider using `Validation.requireInRange`.

config-model-api/src/test/java/com/yahoo/config/application/api/DeploymentSpecWithBcpTest.java
```diff
@@ -0,0 +1,254 @@
...
+                         <region>us-west2</region>
+                         <region>eu-east1</region>
+                         <region>eu-west1</region>
+                      </prod>
+                   </instance>
+                   <bcp>
+                     <group>
+                        <region>us-east1</region>
+                        <region>us-east2</region>
+                        <region fraction="0.3">us-central1</region>
```
mpolden, line 89: I assume we'll discuss the syntax later. Maybe `traffic-share` or simply `share` would be more descriptive?

mpolden (approved):
Great!

config-model-api/src/main/java/com/yahoo/config/application/api/Bcp.java
bratseth: Thanks for the tip. I'll do it in the next pr.

config-model-api/src/test/java/com/yahoo/config/application/api/DeploymentSpecWithBcpTest.java
bratseth: Yes, on Thursday. This isn't traffic share though, but fractional membership ...

PR #25927, opened 2023-02-07
add InputRecorder for global-phase expressions

arnej27959:
@bratseth @lesters for discussion


config-model/src/main/java/com/yahoo/schema/expressiontransforms/InputRecorder.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            neededInputs.add(feature.toString());
+            return;
+        }
+        if (args.size() == 1) {
+            if ("value".equals(name)) {
+                transform(args.expressions().get(0), context);
+                return;
+            }
+            if ("attribute".equals(name) || "query".equals(name)) {
```
bratseth, line 57: You might want to use FeatureNames.isAttributeFeature(feature.reference()) etc. here.

bratseth (approved):
Looks good to me! Quite simple ...

PR #25936, opened 2023-02-08
Balder/reduce use of microsecondtime

baldersheim:
@vekterli PR


storage/src/tests/visiting/visitormanagertest.cpp
```diff
@@ -47,7 +47,7 @@ struct VisitorManagerTest : Test {
     std::unique_ptr<TestVisitorMessageSessionFactory> _messageSessionFactory;
     std::unique_ptr<TestServiceLayerApp> _node;
     std::unique_ptr<DummyStorageLink> _top;
-    VisitorManager* _manager;
+    VisitorManager* _manager{};
```
vekterli, line 50: This is redundant with the explicit initialization of `_manager` to `nullptr` in the constructor

storage/src/vespa/storage/distributor/pendingclusterstate.cpp
```diff
@@ -306,7 +306,7 @@ PendingClusterState::getSummary() const
 {
     return Summary(getPrevClusterStateBundleString(),
                    getNewClusterStateBundleString(),
-                   (_clock.getTimeInMicros().getTime() - _creationTimestamp));
+                   std::chrono::microseconds(vespalib::count_us(_clock.getSystemTime().time_since_epoch()) - _creationTimestamp));
```
vekterli, line 309: Is there any point in explicitly truncating this down to microseconds here when the underlying type passed and stored as `duration`?

storage/src/vespa/storageframework/defaultimplementation/clock/fakeclock.h
```diff
@@ -22,13 +22,12 @@ struct FakeClock : public framework::Clock {
 
 private:
     Mode _mode;
-    framework::MicroSecTime _absoluteTime;
+    vespalib::duration      _absoluteTime;
```
vekterli, line 25: Semantically it would make more sense to let this be a `time_point`, but it would require more wrapping and unwrapping...

vekterli (approved):

storage/src/tests/visiting/visitormanagertest.cpp
baldersheim: Thanks, fixed

storage/src/vespa/storage/distributor/pendingclusterstate.cpp
baldersheim: Hardly

storage/src/vespa/storageframework/defaultimplementation/clock/fakeclock.h
baldersheim: Agree, leaving as is.

PR #25944, opened 2023-02-08
- Disconnect clock and MicroSecTime.

baldersheim:
- GC unused functionality from clock/time.h
- Unify on vespalib::steady_time/system_time/duration.

@vekterli PR
This should be the final one....


storage/src/vespa/storage/bucketdb/bucketinfo.h
```diff
@@ -21,6 +21,7 @@ template <typename NodeSeq>
 class BucketInfoBase
 {
 protected:
+    //TODO: Should we use a chrono timepoint to ensure we are using same clock everywhere ?
```
vekterli, line 24: It would be good to unify this, but I think we'd have to be careful to not expose an API that makes it seem we can store and later retrieve a higher resolution time point than seconds. We're limited to u32 seconds after epoch, since the GC timestamp is stored as the 32 MSBs of the 64-bit bucket DB B-tree KV value (with a DataStore entry ref taking up the 32 LSBs).

storage/src/vespa/storage/distributor/stripe_bucket_db_updater.h
```diff
@@ -144,12 +144,12 @@ class StripeBucketDBUpdater final
                                        const BucketRequest& req);
     void mergeBucketInfoWithDatabase(const std::shared_ptr<api::RequestBucketInfoReply>& repl,
                                      const BucketRequest& req);
-    void convertBucketInfoToBucketList(const std::shared_ptr<api::RequestBucketInfoReply>& repl,
+    static void convertBucketInfoToBucketList(const std::shared_ptr<api::RequestBucketInfoReply>& repl,
```
vekterli, line 147: Nit: missing indent update on next line

vekterli (approved):

storage/src/vespa/storage/bucketdb/bucketinfo.h
baldersheim: Agree

storage/src/vespa/storage/distributor/stripe_bucket_db_updater.h
baldersheim: Fixed

PR #25956, opened 2023-02-09
Add errata entry for bit vector memory corruption bug

vekterli:
@kkraune @geirst please review

Primary change from the draft is an added "affected version range" tl;dr to the top, as that's what people will tend to want to know before reading anything else.


kkraune (approved):
👍

geirst (approved):
👍

PR #25967, opened 2023-02-09
trace global filter iterator tree

havardpe:
also simplify making and joining sub-traces

@geirst please review
@baldersheim @vekterli FYI


geirst (approved):
👍

PR #25975, opened 2023-02-10
Drop explicit active config snapshot generation in document db. It is

toregge:
already part of the active config snapshot.

@geirst : please review


geirst (approved):
👍

PR #25979, opened 2023-02-10
Add a controller handler for resealing decryption tokens

vekterli:
@bjorncs please review. The controller is not my usual hunting grounds, so please verify that I'm doing things in the Correct Way(tm).

Handles an _already authenticated and authorized request_, using a config-provided secret private key decrypt the original token and reseal it towards the requested public key.

Key IDs are expected to be on the format "name.version" where version is an unsigned integer. The name must exactly match the key name used to look up the secret private key, or the request will be failed.


bjorncs (approved):

vekterli:
(the force push is only a rebase onto master to see if test failure is related to any concurrent changes)

bjorncs (approved):

PR #25988, opened 2023-02-10
Avoid std::ranges::reverse_view until clang supports it properly

baldersheim:
@toregge PR


toregge (approved):

toregge:
The `ranges` header file is present in libc++14 and libc++15, but the content is commented out unless the `-fexperimental-library` option has been passed to clang++.

Note that `std::ranges_reverse_view` doesn't work with clang++ on linux (before version 16 which is not yet released) even though the header files for `libstdc++` are used.

PR #25989, opened 2023-02-10
Revert "handle that the component graph contains multiple ModelsEvaluator instances"

baldersheim:
Reverts vespa-engine/vespa#25985
@arnej27959 or @bratseth PR
Ssupect this is reason for failing system test


arnej27959 (approved):
yes - this seemed too easy...

PR #26006, opened 2023-02-13
Reenable default cpu arch option for clang on linux.

toregge:
@baldersheim : please review


default_build_settings.cmake
```diff
@@ -130,6 +130,8 @@ function(vespa_use_default_build_settings)
     message("-- CMAKE_SYSTEM_PROCESSOR = ${CMAKE_SYSTEM_PROCESSOR}")
     if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
       if(APPLE AND (("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang") OR ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")))
+      elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+        set(DEFAULT_VESPA_CPU_ARCH_FLAGS "-march=westmere -mtune=haswell")
```
baldersheim, line 134: Foreslår at du sette skylake som default der. Finnes i praksis ikke noe eldre for apple.
Also add a comment why.

baldersheim (changes requested):
"-march=skylake"  and a comment.

default_build_settings.cmake
```diff
@@ -130,6 +130,8 @@ function(vespa_use_default_build_settings)
...
       else()
         if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
           set(DEFAULT_VESPA_CPU_ARCH_FLAGS "-march=ivybridge")
```
baldersheim, line 137: I think you should add a comment here with reference to the gcc bugzilla ticket too.

default_build_settings.cmake
toregge: Added comment.

default_build_settings.cmake
toregge: Added comment.

baldersheim (approved):

PR #26008, opened 2023-02-13
Revert "Add debug logging for successfactor triggering alert"

bjormel:
Reverts vespa-engine/vespa#25939


bjormel:
Did not work :(

PR #26014, opened 2023-02-13
Use config id with sentinel suffix for config sentinel

hmusum:
This will allow us to simplify config model code


hmusum:
@gjoranv what do you think?

gjoranv (approved):

gjoranv:
This should be safe, as the config cascading ensures same config as before.

PR #26015, opened 2023-02-13
split AbstractConfigProducer

arnej27959:
After trying (again) to figure out how AbstractConfigProducer is supposed to be used,
versus how it's actually mis-used, I've embarked on a quest to make things more sane.

As a first. step, this splits the class into a superclass "AnyConfigProducer" (without children)
and a subclass "TreeConfigProducer" (with just the support for having children).

@gjoranv please review
@bratseth please review
@hmusum FYI

### DIFF
--- a/config-model/src/main/java/com/yahoo/config/model/producer/TreeConfigProducer.java
+++ b/config-model/src/main/java/com/yahoo/config/model/producer/TreeConfigProducer.java
@@ -2,21 +2,10 @@
 package com.yahoo.config.model.producer;
 
 import com.yahoo.api.annotations.Beta;
-import com.yahoo.config.ConfigInstance;
 import com.yahoo.config.model.ApplicationConfigProducerRoot;
-import com.yahoo.config.model.deploy.DeployState;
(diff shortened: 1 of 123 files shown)


bratseth (approved):
It's great you have a plan for improving this!

And this step looks good to me.

config-model/src/main/java/com/yahoo/config/model/producer/TreeConfigProducer.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Superclass for all config producers with children.
+ * Config producers constructs and returns config instances on request.
+ *
+ * @author gjoranv
```
gjoranv, line 35: I suggest you set or add yourself as author.

gjoranv (approved):
Great, this is a step in the right direction!

gjoranv:
👍

PR #26019, opened 2023-02-13
Balder/gc timedout optimization

baldersheim:


dist/patch.stl_vector.h.diff
```diff
@@ -0,0 +1,30 @@
+--- /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/stl_vector.h	2022-09-14 17:00:12.000000000 +0000
```
toregge, line 1: Where is this patch used ?

baldersheim: Ading the patch now, using it in separate PR soon to come.

toregge (approved):

PR #26022, opened 2023-02-14
When lock was closed, unbeknownst to us, release it internally

jonmv:
@hakonhall please review.

### DIFF
--- a/zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
+++ b/zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
@@ -128,18 +128,19 @@ class SingletonManager {
 
         }
 
-        private static final Instant INVALID = Instant.ofEpochMilli(-1);
+        private static final Instant PENDING = Instant.ofEpochMilli(-1);
+        private static final Instant INVALID = Instant.ofEpochMilli(-2);
+        private static final Instant CLOSING = Instant.ofEpochMilli(-3);
 
         final BlockingDeque<Task> tasks = new LinkedBlockingDeque<>();
         final Deque<SingletonWorker> singletons = new ArrayDeque<>(2);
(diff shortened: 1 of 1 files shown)


zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
```diff
@@ -139,6 +139,7 @@ private Task(Type type, SingletonWorker singleton) {
         final Path path;
         final MetricHelper metrics;
         Lock lock = null;
+        Lock toClose = null;
```
jonmv, line 142: Would it be clearer to have an addition state on `doom` (say `CLOSING`) to signal this closing state, since that is already used as a state field?

jonmv: (The point is to be able to retry closing the lock, but at the same time not activating, because "we have the lock" (which we don't really have after invalidation).

hakonhall (approved):

zkfacade/src/main/java/com/yahoo/vespa/curator/SingletonManager.java
```diff
@@ -272,13 +273,12 @@ private void doUnregister(SingletonWorker singleton) {
...
-            if (doom.get() == INVALID) {
+            // Witness value to detect if invalidation occurs between here and successful ping.
+            Instant ourDoom = doom.get();
+            if (ourDoom == INVALID) {
                 logger.log(INFO, "Lease invalidated");
-                doom.set(null);
                 return; // Skip to updateStatus, deactivation, and release the lock.
             }
-            // Witness value to detect if invalidation occurs between here and successful ping.
-            Instant ourDoom = doom.get();
```
jonmv, line 281: Witness should be taken before we check for `INVALID`, lest we risk taking an `INVALID` witness, which means we _should_ nullify the lock, but we miss that.

PR #26029, opened 2023-02-14
Arnej/less wildcards for config producers 1

arnej27959:
@gjoranv please review
@bratseth @hmusum FYI


gjoranv (approved):
Big improvement!

PR #26041, opened 2023-02-15
Put independent function to copy memory in separate compilation unit.

baldersheim:
@toregge PR

### DIFF
--- a/vespamalloc/src/vespamalloc/malloc/CMakeLists.txt
+++ b/vespamalloc/src/vespamalloc/malloc/CMakeLists.txt
@@ -72,5 +72,6 @@ vespa_add_library(vespamalloc_mmap OBJECT
 vespa_add_library(vespamalloc_load_as_huge OBJECT
     SOURCES
     load_as_huge.cpp
+    independent_non_inlined_memcpy.cpp
     DEPENDS
 )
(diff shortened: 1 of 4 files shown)


vespamalloc/src/vespamalloc/malloc/CMakeLists.txt
```diff
@@ -72,5 +72,6 @@ vespa_add_library(vespamalloc_mmap OBJECT
 vespa_add_library(vespamalloc_load_as_huge OBJECT
     SOURCES
     load_as_huge.cpp
+    independent_non_inlined_memcpy.cpp
```
toregge, line 75: Consider setting extra compiler flags to avoid generating call to memcpy.

toregge (approved):

vespamalloc/src/vespamalloc/malloc/CMakeLists.txt
baldersheim: I tried compiling with many combinations, never did it try to add call to memcpy.
I doubt that it ever will try that. Normally it is the other way around, it will optimize away calls to memcpy.
So until it has been observed that it does, I will not add any special handling.

PR #26046, opened 2023-02-15
docproc metrics as gauges instead of counters

olaaun:
When using `metric.add`, the metrics are interpreted as counters - which are only available with `.count` and `.rate` suffixes.
 

yngveaasheim (approved):
Metrics are not updated per document, but per micro-batch of documents. This may cause distorted metrics, at least with variable batch size. But let us use this for now.

PR #26048, opened 2023-02-15
Add string command to index info config.

toregge:
Use index info to validate fuzzy query terms.

@bratseth or @geirst  : please review


bratseth (approved):
👍

geirst (commented):
👍

PR #26061, opened 2023-02-16
cosmetic changes

arnej27959:
@havardpe please review


searchlib/src/vespa/searchlib/fef/blueprintresolver.cpp
```diff
@@ -139,6 +139,9 @@ struct Compiler : public Blueprint::DependencyHandler {
             } else {
                 msg = fmt("invalid %s: %s\n%s", describe(feature_name).c_str(), reason.c_str(), trace.c_str());
             }
+            if (msg.rbegin()[0] == '\n') {
```
havardpe, line 142: need to check size as well

havardpe (approved):

PR #26098, opened 2023-02-18
Add initial text generator component

lesters:
@bratseth Please review.
@jobergum FYI.

This only implements a greedy search and doesn't really generate satisfactory results. Will be adding beam search and sampling methods later.


bratseth (approved):
Great!

PR #26099, opened 2023-02-18
Add model ids for generator models

lesters:
@bratseth Please review
@jobergum FYI

For reference, the large decoder is around 1.8Gb, the XL is more than 3x that again.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ModelIdResolver.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ModelIdResolver.java
@@ -3,6 +3,9 @@ package com.yahoo.vespa.model.container.xml;
 
 import com.yahoo.text.XML;
 import org.w3c.dom.Element;
+
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


lesters:
Removed XL models because they consisted of 2 parts: a description and a data file which I believe we do not support yet.

bratseth (approved):

bratseth:
```
[ERROR] Failures: 
13:24:08 [ERROR]   EmbedderTestCase.testUnknownModelId_hosted:93->assertTransformThrows:197 expected: <Unknown model id 'my_model_id' on 'transformerModel'. Available models are [bert-base-uncased, minilm-l6-v2, mpnet-base-v2]> but was: <Unknown model id 'my_model_id' on 'transformerModel'. Available models are [bert-base-uncased, flan-t5-base-decoder, flan-t5-base-encoder, flan-t5-large-decoder, flan-t5-large-encoder, flan-t5-small-decoder, flan-t5-small-encoder, flan-t5-vocab, minilm-l6-v2, mpnet-base-v2]>

```
Perhaps we should make this test less specific ...

lesters:
Thanks, fixed

PR #26100, opened 2023-02-18
Predict best case overhead during autoscaling

bratseth:
Predicting worst case has the unavoidable consequence that we will conclude we cannot fulfill requested resources in the case where there are no resource ranges (only node count ranges), or where they are too narrow, and thus refrain from scaling down when we should.

Instead, mostly predict the best case and let the ideal < 1 headroom absorb the deficiency in what we are actually allocated.

This is as complex as this logic will ever become and there will definitely never be any need to refine further :relaxed:


freva (approved):
:open_mouth:

PR #26107, opened 2023-02-19
add metric for current config generation

arnej27959:
@geirst please review
@yngveaasheim @bratseth FYI

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/metrics/content_proton_metrics.cpp
+++ b/searchcore/src/vespa/searchcore/proton/metrics/content_proton_metrics.cpp
@@ -29,7 +29,7 @@ ContentProtonMetrics::ProtonExecutorMetrics::~ProtonExecutorMetrics() = default;
 
 ContentProtonMetrics::ContentProtonMetrics()
     : metrics::MetricSet("content.proton", {}, "Search engine metrics", nullptr),
-      configGeneration("config_generation", {}, "The oldest config generation used by this process", this),
+      configGeneration("config.generation", {}, "The oldest config generation used by this process", this),
       transactionLog(this),
       resourceUsage(this),
       executor(this),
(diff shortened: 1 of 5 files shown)


searchcore/src/vespa/searchcore/proton/metrics/content_proton_metrics.cpp
```diff
@@ -29,6 +29,7 @@ ContentProtonMetrics::ProtonExecutorMetrics::~ProtonExecutorMetrics() = default;
 
 ContentProtonMetrics::ContentProtonMetrics()
     : metrics::MetricSet("content.proton", {}, "Search engine metrics", nullptr),
+      configGeneration("config_generation", {}, "The oldest config generation used by this process", this),
```
geirst, line 32: I suggest renaming this to _config.generation_ in case we need other config related metrics later.

bratseth: Sensible, but we already have this metric.

bratseth: Correction: The existing container metric is actually called "application_generation". 
But since this will have a proton prefix it won't be the same here in any case.

I agree that "config.generation" is better, and perhaps we should move to that both places, and then without any prefix - conceptually this is a config metric, not a proton/container one. It's not a big deal to handle different names though, so not sure it is worth prioritizing.

arnej27959:
I've changed the name to "content.proton.config.generation" now.  It's possible to change it to "content.config.generation" with some restructuring, but getting to "config.generation" would need rewriting the proton metrics engine to have multiple root metric-sets.

bratseth:
👍 thanks!

geirst (approved):
👍

PR #26108, opened 2023-02-20
Allow patching wireguardPubkey

freva:


gjoranv (approved):
Thanks, missed that one!

PR #26116, opened 2023-02-20
Read content config generation

bratseth:


bratseth:
Waiting for https://github.com/vespa-engine/vespa/pull/26107 to merge

container-core/pom.xml
```diff
@@ -453,6 +453,8 @@
             <arg>-Xlint:-try</arg>
             <arg>-Werror</arg>
           </compilerArgs>
+            <source>17</source>
```
freva, line 456: Why is this needed?

freva (approved):

container-core/pom.xml
bratseth: Sorry, it's not, just more IntelliJ inserts - deleted.

PR #26124, opened 2023-02-21
Bratseth/admin provisioning

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
```diff
@@ -88,10 +88,13 @@ private long reservedDiskSpaceGb(NodeResources.StorageType storageType, boolean
...
 
-    private double minRealVcpu(ApplicationId applicationId, ClusterSpec cluster) { return minAdvertisedVcpu(applicationId, cluster); }
+    private double minRealVcpu(ApplicationId applicationId, ClusterSpec cluster) {
+        return minAdvertisedVcpu(applicationId, cluster);
+    }
 
     private double minRealMemoryGb(ClusterSpec cluster) {
-        return minAdvertisedMemoryGb(cluster) - 1.7;
+        if (cluster.type() == ClusterSpec.Type.admin) return 0.95; // TODO: Increase to 1.05 after March 2023
+        return 2.3;
```
hmusum, line 97: Some explanation on how these numbers have been found would be nice

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
```diff
@@ -115,10 +115,12 @@ private NodeResources clusterControllerResources(ClusterSpec clusterSpec, Archit
         if (nodeRepository.exclusiveAllocation(clusterSpec)) {
             return versioned(clusterSpec, Map.of(new Version(0), smallestExclusiveResources()));
         }
-        // TODO (hmusum): Go back to 1.14 Gb memory when bug in resource limits for admin nodes
-        // has been fixed
+
+        // 1.32 fits floor(8/1.32) = 6 cluster controllers on each 8Gb host, and each will have
+        // 1.32-(0.7+0.6)*(1.32/8) = 1.1 Gb real memory given current taxes.
```
hmusum, line 120: Assumes shared hosts are at least 8 Gb and that host-memory feature flag is 0.6. First issue we should fix, for the second issue it seems like we should use 1.2 for AWS/public (cd and publiccd uses 1.2 already). But this will still be better that it was, so I think this is OK

hmusum (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
bratseth: I added an explanation where they are specified, in CapacityPolicies. The numbers here just lead to a provision failure if they are not met, so they matter less.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
bratseth: Yes, if the host need more we'll have to change, but hopefully not.

PR #26128, opened 2023-02-21
Fix C++ document selection regression where supertype in field expression would not match as expected

vekterli:
@bratseth please review

A change between Vespa major versions 7 and 8 to use exact-matching for document types was implemented a tad too aggressively, as it should still be possible to match against supertypes in a document selection _field expression_.

Let type `B` inherit `A`, `A` have a field `foo`, and matching a document of type `B` with field `foo` present:

Prior to this fix:
 * `B.foo` _would_ match (exact type match and implicit supertype field resolving)
 * `A.foo` would _not_ match due to exact type mismatch

After this fix, both will match (as would generally be expected). This brings behavior in sync with the Java implementation.


geirst (approved):
👍

bratseth:
👍

PR #26146, opened 2023-02-22
add configurable input/output mappings

arnej27959:
@lesters or @bratseth please review


bratseth (dismissed):
👍

model-evaluation/src/main/java/ai/vespa/models/evaluation/OnnxModel.java
```diff
@@ -48,9 +48,11 @@ static class OutputSpec {
     final List<OutputSpec> outputSpecs = new ArrayList<>();
 
     void addInputMapping(String onnxName, String source) {
+        assert(referencedEvaluator == null);
```
bratseth, line 51: Do we really want to kill the container on this?

bratseth (approved):

PR #26148, opened 2023-02-22
Emit type inheritance statements for concrete document types

### LINKED ISSUE #26126: DocumentSelector doesn't work when having a concrete type that inherits from another
**Describe the bug**
We are working with the [DocumentSelector](https://github.com/vespa-engine/vespa/blob/master/document/src/main/java/com/yahoo/document/select/DocumentSelector.java) class and we are encountering the following problem:

We have document type called "Caselaw" that inherits from another document type called "Docs", so the fields of both of them are the following ones:

- Docs
   - **recordId**
- Caselaw
   - **recordId**
   - otherField

We set up two different conditions for a document of type Caselaw with the recordId set to "example", that should be true:

- docs.recordId == "example"
- caselaw.recordId == "example"

However, the first one returns false and the second one returns true. I have been checking your source code and I have identified this bug at [DocumentGenMojo](https://github.com/vespa-engine/vespa/blob/master/vespa-documentgen-plugin/src/main/java/com/yahoo/vespa/DocumentGenMojo.java#L451), when generating the document type for the concrete type. As you can see, it does not generate the field [_inherits_](https://github.com/vespa-engine/vespa/blob/07afcf4b4ad1034601b8b438ff4b13ed63e1b0c1/document/src/main/java/com/yahoo/document/DocumentType.java#L39) using the function [_inherit_](https://github.com/vespa-engine/vespa/blob/07afcf4b4ad1034601b8b438ff4b13ed63e1b0c1/document/src/main/java/com/yahoo/document/DocumentType.java#L267).

When evaluating the expression at https://github.com/vespa-engine/vespa/blob/07afcf4b4ad1034601b8b438ff4b13ed63e1b0c1/document/src/main/java/com/yahoo/document/select/rule/DocumentNode.java#L56

it calls [isA](https://github.com/vespa-engine/vespa/blob/master/document/src/main/java/com/yahoo/document/DocumentType.java#L200-L202), that checks if it belongs to the inherited document types.

**To Reproduce**
1. Define two schemas A and B, where B inherits from A. A defines a field that is inherited by B.
2. Use vespa-documentgen-plugin to generate concrete types for both A and B
3. Create an object of the class of concrete type B
4. Define the following conditions for the parser: "A.field == 'value'" and "B.field == 'value'"
5. The first condition will fail but not the second one

**Expected behavior**
Both expressions should be true

**Screenshots**

Screenshot 2023-02-21 at 12 38 46

**Environment (please complete the following information):**
 - OS: macOS
 - Infrastructure: self-hosted
 - Versions: Ventura 13.1

**Vespa version**
8.127.1

**Additional context**
Nothing

vekterli:
@bratseth please review.

This relates to issue #26126 where inheritance specified in the schema was not propagated to the generated meta type, causing `is-a` type checks to not give the expected results.

### DIFF
--- a/documentgen-test/src/test/java/com/yahoo/vespa/config/DocumentGenPluginTest.java
+++ b/documentgen-test/src/test/java/com/yahoo/vespa/config/DocumentGenPluginTest.java
@@ -47,8 +47,10 @@ import com.yahoo.vespa.documentgen.test.Book;
 import com.yahoo.vespa.documentgen.test.Book.Ss0;
 import com.yahoo.vespa.documentgen.test.Book.Ss1;
 import com.yahoo.vespa.documentgen.test.Common;
+import com.yahoo.vespa.documentgen.test.Common2;
 import com.yahoo.vespa.documentgen.test.ConcreteDocumentFactory;
 import com.yahoo.vespa.documentgen.test.Music;
+import com.yahoo.vespa.documentgen.test.Music2;
 import com.yahoo.vespa.documentgen.test.Music3;
 import com.yahoo.vespa.documentgen.test.Music4;
 import com.yahoo.vespa.documentgen.test.Parent;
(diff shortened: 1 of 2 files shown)


bratseth (approved):
👍

PR #26155, opened 2023-02-23
Bjorncs/onnx model reuse

bjorncs:


bratseth (dismissed):
LGTM, but I wonder if the "cache" abstraction is right on an application API level.
Does the user of an evaluator really need to decide whether or not to use a cache?
If not I think it would be better to hide the caching aspect inside the OnnxEvaluator interface, for example by making that just a delegator to an implementation, and have an implementation that caches and forwards to another implementation which contains most of the current OnnEvaluator stuff.

arnej27959:
I can't see a good way of hiding the cache, we would need to add a finalizer to do close().
AFAIK we have decided to avoid finalizers and release resources in destroy() instead.
My initial guess was to use the jdisc SharedResource, but it doesn't look like that would help anything?

bratseth:
Isn't it always better to cache? If so it should just be a built-in feature of the evaluator and not something the user of it has to think about and choose explicitly, is what I mean (even if we then need it to be closable/destructable).

Components can be made destructable by extending AbstractComponent.

bjorncs:
My tentative plan is to introduce a `OnnxRuntime` component available for injection. This component will act as factory/provider for evaluator instances. The actual cache will be hidden inside `OnnxRuntime` and performed at a lower abstraction layer (`ai.onnxruntime.OrtSession`).

bjorncs:
@arnej27959 @bratseth PTAL

bratseth (commented):
Great! 

I think injectable and destructable components which may be replaced by a new config generation (HuggingFaceEmbedder, BertBaseEmbedder) must extend AbstractComponent not AbstractResource though?

bjorncs:
@bratseth Thanks, fixed in [4117dff](https://github.com/vespa-engine/vespa/pull/26155/commits/4117dff58df12182298efdbd6cc5f2f5f93733cd).

bratseth (approved):

PR #26166, opened 2023-02-23
assert that there are no pending resolves when deleting

havardpe:
@geirst please review

if there is a double subref somewhere, this will hopefully catch it


geirst (approved):
👍

PR #26167, opened 2023-02-23
Revert "Revert "Store original capability (set) names from JSON config in PeerPolicy" MERGEOK"

bjorncs:
Reverts vespa-engine/vespa#26153


vekterli (approved):
:+1:

PR #26173, opened 2023-02-24
New billing dimensions

oyving:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/resource/ResourceSnapshot.java
```diff
@@ -41,37 +44,26 @@ public static ResourceSnapshot from(List<Node> nodes, Instant timestamp, ZoneId
                                                  .map(node -> node.owner().get())
                                                  .collect(Collectors.toSet());
 
+        Set<Version> versions = nodes.stream()
+                .map(Node::currentVersion)
+                .collect(Collectors.toSet());
+
         if (applicationIds.size() != 1) throw new IllegalArgumentException("List of nodes can only represent one application");
+        if (versions.size() != 1) throw new IllegalArgumentException("List of nodes can only represent one version");
```
olaaun, line 52: This constraint is correct, but the node list needs to be grouped by version beforehand. Otherwise the maintainer will fail whenever an application is upgrading, and nodes in the same cluster are on different Vespa versions.

oyving: Good catch. I created a grouping on major version in addition to architecture.

olaaun (approved):

PR #26178, opened 2023-02-24
Allow patching account archive URIs in node-repo

freva:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/CuratorDb.java
```diff
@@ -70,7 +70,7 @@ public class CuratorDb {
     private static final Path infrastructureVersionsPath = root.append("infrastructureVersions");
     private static final Path osVersionsPath = root.append("osVersions");
     private static final Path firmwareCheckPath = root.append("firmwareCheck");
-    private static final Path archiveUrisPath = root.append("archiveUris");
+    private static final Path archiveUrisPath = root.append("archiveUri");
```
freva, line 73: Changed to new path since the new serialization is incompatible with the previous. It's OK to forget the previous data since it'll quickly be updated by the controller.

hakonhall (approved):

PR #26179, opened 2023-02-24
add new components for global-phase handling

arnej27959:
still work-in-progress, but should be possible to review some of this now.

my biggest open questions are in rerankHits():
 - how to iterate over hits
 - what to do if rescoring fails
 - how to adjust scores of not-reranked hits

@bratseth please review
@bjorncs please review
@havardpe FYI


container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseHelper.java
```diff
@@ -0,0 +1,152 @@
...
+                double oldScore = hit.getRelevance().getScore();
+                boolean didRerank = hitRescorer.rescoreHit(hit);
+                if (didRerank) {
+                    double newScore = hit.getRelevance().getScore();
+                    if (oldScore < worstRerankedOldScore) worstRerankedOldScore = oldScore;
+                    if (newScore < worstRerankedScore) worstRerankedScore = newScore;
+                    --rerank;
+                } else {
+                    // failed to rescore this hit, what should we do?
+                    hit.setRelevance(-Double.MAX_VALUE);
```
bjorncs, line 109: When can this happen?

container-search/src/main/java/com/yahoo/search/ranking/SimpleEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public Collection<String> needInputs() { return List.copyOf(neededInputs); }
+
+    public SimpleEvaluator bind(String name, Tensor value) {
+        if (value != null) evaluator.bind(name, value);
+        neededInputs.remove(name);
+        return this;
+    }
+
+    public double evaluateScore() {
```
bjorncs, line 33: Consider using `@Override` when implementing/overriding members from inherited class/interface

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseHelper.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.function.Supplier;
+
+public class GlobalPhaseHelper {
+
+    private static final Logger logger = Logger.getLogger(GlobalPhaseHelper.class.getName());
+    private final RankingExpressionEvaluatorFactory factory;
+    private final Set<String> skipProcessing = new HashSet<>();
```
bjorncs, line 26: This should ideally be deduced from config

container-search/src/main/java/com/yahoo/search/ranking/SimpleEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class SimpleEvaluator implements Evaluator {
+
+    private final FunctionEvaluator evaluator;
+    private final Set<String> neededInputs;
+
+    
+    public SimpleEvaluator(FunctionEvaluator prototype) {
+        this.evaluator = prototype;
+        this.neededInputs = new HashSet<String>(prototype.function().arguments());
```
bjorncs, line 22: The convention in Java is to perform defence copying in constructor for immutable types. That will remove the need for `List.copyOf` for each `needInputs()` invocation.

bjorncs (commented):
See comments

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseHelper.java
```diff
@@ -0,0 +1,152 @@
...
+        Supplier<Evaluator> supplier = () -> {
+            var evaluator = functionEvaluatorSource.get();
+            var simple = new SimpleEvaluator(evaluator);
+            for (var entry : prepared) {
+                simple.bind(entry.name(), entry.value());
+            }
+            return simple;
+        };
+        // TODO need to get rerank-count somehow
+        int rerank = 7;
```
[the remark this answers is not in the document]
arnej27959: yes - we can pick it up from rank-profile config there.

container-search/src/main/java/com/yahoo/prelude/cluster/ClusterSearcher.java
```diff
@@ -228,8 +234,23 @@ private Result doSearch(Searcher searcher, Query query, Execution execution) {
...
+        }
+    }
+
+    private Result perSchemaSearch(Searcher searcher, Query query, Execution execution) {
+        Set<String> restrict = query.getModel().getRestrict();
+        if (restrict.size() != 1) {
+            throw new IllegalStateException("perSchemaSearch must always be called with 1 schema, got: " + restrict.size());
+        }
+        for (String schema : restrict) {
+            Result result = searcher.search(query, execution);
```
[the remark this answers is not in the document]
arnej27959: here it's invoked exactly 1 time, that's checked immediately previous.  I could use restrict.iterator().next() instead of a loop, to make it clearer that it only happens once?

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseHelper.java
arnej27959: this should only fail if something is missing from match-features.  after looking at how score adjustment happens in C++ after second-phase I am implementing the same algorithm here.

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseHelper.java
arnej27959: yes - we can deduce it and wire it same way as rerank-count

container-search/src/main/java/com/yahoo/prelude/cluster/ClusterSearcher.java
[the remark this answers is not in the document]
bjorncs: You're right; my bad. I didn't read the preceding lines carefully enough 🤦 I agree with your proposal.

bjorncs (dismissed):

bratseth (commented):
See my suggestions, otherwise LGTM!

bjorncs (approved):

PR #26183, opened 2023-02-24
handle other Value subtypes holding tensors

arnej27959:

The implementations in TensorValue did not handle if any other Value sub-type actually contained a Tensor.
In particular, LazyValue in model-evaluation could only be used if the lazy value was a double.


lesters (approved):
LGTM!

bratseth (approved):

PR #26188, opened 2023-02-26
pick up declared type of query/attribute features,

arnej27959:
and handle globalphase not wrapped in rankExpression().


bratseth (approved):
👍

PR #26194, opened 2023-02-27
Add details on why an autoscaling is empty

bratseth:
Add temporary details to see why the autoscaling test rarely make it to rescaling before timeout.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterNodesTimeseries.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterNodesTimeseries.java
@@ -23,6 +23,8 @@ public class ClusterNodesTimeseries {
     /** The measurements for all nodes in this snapshot */
     private final List<NodeTimeseries> timeseries;
 
+    private int initialCount, afterWarmupCount, afterStableCount, finalCount;
+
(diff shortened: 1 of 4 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterNodesTimeseries.java
```diff
@@ -45,15 +51,25 @@ private ClusterNodesTimeseries(NodeList clusterNodes, List<NodeTimeseries> times
...
     public boolean isEmpty() {
         return measurementsPerNode() == 0;
     }
 
     /** Returns the average number of measurements per node */
     public double measurementsPerNode() {
         if (clusterNodes.size() == 0) return 0;
-        int measurementCount = timeseries.stream().mapToInt(m -> m.size()).sum();
-        return (double)measurementCount / clusterNodes.size();
+        return (double) totalMeasurementsIn(timeseries) / clusterNodes.size();
```
freva, line 68:
```suggestion
        return (double) finalCount / clusterNodes.size();
```

freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/ClusterNodesTimeseries.java
bratseth: I'm planning to remove the counters later.

PR #26195, opened 2023-02-27
Balder/x sys v0.5.0

baldersheim:
@arnej27959 PR
@bjorncs @gjoranv @tokle FYI
How should we maintain our dependencies with go.
Check for updates regularly, or just wait for dependabot ?
Can dependabot be more agile ?


bjorncs (approved):

gjoranv (approved):

gjoranv:
If possible, we should reduce the number of dependencies.
This is probably well known, I just found it by googling:
https://go.dev/doc/modules/managing-dependencies#:~:text=You%20can%20check%20to%20see,version%20available%20for%20that%20module.

PR #26213, opened 2023-02-28
GC xml output of metrics.

baldersheim:
@vekterli PR
There should be no semantic changes here


metrics/src/vespa/metrics/metricmanager.cpp
```diff
@@ -165,9 +169,8 @@ void
 MetricManager::init(const config::ConfigUri & uri, bool startThread)
 {
     if (isInitialized()) {
-        throw vespalib::IllegalStateException(
-                "The metric manager have already been initialized. "
-                "It can only be initialized once.", VESPA_STRLOC);
+        throw IllegalStateException("The metric manager have already been initialized. "
+                                    "It can only be initialized once.", VESPA_STRLOC);
```
vekterli, line 173: Also looks like something that should be an `assert`. `IllegalStateException` always sounded like a bad idea since it implies the program execution has gone awry...

vekterli (approved):
:+1:

metrics/src/vespa/metrics/metricmanager.cpp
```diff
@@ -34,13 +37,12 @@ MetricManager::Timer::getTime() const {
 void
 MetricManager::assertMetricLockLocked(const MetricLockGuard& g) const {
     if ( ! g.owns(_waiter)) {
-        throw vespalib::IllegalArgumentException("Given lock does not lock the metric lock.", VESPA_STRLOC);
+        throw IllegalArgumentException("Given lock does not lock the metric lock.", VESPA_STRLOC);
```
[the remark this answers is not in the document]
baldersheim: I will do a cleanup of this afterwards.

metrics/src/vespa/metrics/metricmanager.cpp
baldersheim: I will do a cleanup of this afterwards.

PR #26215, opened 2023-02-28
Revert "Deopend on component to get AbstractComponent"

bjorncs:
Reverts vespa-engine/vespa#26211

The PR duplicates existing `component` dependency


bratseth:
I guess somebody else also did this ..

PR #26218, opened 2023-02-28
Use vespalib::steady_time for getMilliSecTime to ensure no wraps arou…

baldersheim:
…nd and safer code.

@vekterli PR


vekterli (approved):
:+1:

PR #26234, opened 2023-02-28
Balder/enforce system time for snapshot

baldersheim:
@vekterli PR


metrics/src/vespa/metrics/metricsnapshot.cpp
```diff
@@ -6,13 +6,15 @@
...
+
 namespace metrics {
 
 MetricSnapshot::MetricSnapshot(const Metric::String& name)
     : _name(name),
       _period(0),
-      _fromTime(0),
-      _toTime(0),
+      _fromTime(default_system_time),
+      _toTime(default_system_time),
```
vekterli, line 17: Should this be `system_time_epoch` instead? Or maybe `zero_system_time`? "default" is a bit ambiguous

vekterli (dismissed):
:+1:

metrics/src/vespa/metrics/metricsnapshot.cpp
baldersheim: system_time_epoch it is.

PR #26246, opened 2023-03-01
Pass ClusterInfo

bratseth:
Passing ClusterInfo to node-repo containing bcp deadline, but not using it for anything yet.
This is a bit more elaborate than needed for that particular piece of data but I plan to use it for other things too.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/ApplicationPackage.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/ApplicationPackage.java
@@ -72,6 +72,9 @@ public interface ApplicationPackage {
     String RULES_NAME_SUFFIX = ".sr";
     String EXT_DIR = "ext";
 
+    // TODO: Remove when oldest model version is 8.95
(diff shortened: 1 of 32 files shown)


config-model-api/src/main/java/com/yahoo/config/application/api/ApplicationPackage.java
```diff
@@ -71,9 +72,6 @@ public interface ApplicationPackage {
     String RULES_NAME_SUFFIX = ".sr";
     String EXT_DIR = "ext";
 
-    // TODO: Remove when oldest model version is 8.95
```
freva, line 74: Needed until Vespa 7 is gone?

freva (dismissed):

config-model-api/src/main/java/com/yahoo/config/application/api/ApplicationPackage.java
bratseth: I don't think so, but since I'm not 100% sure I put it back.

PR #26247, opened 2023-03-01
Preserve original exception when deployments fail to read from S3

jonmv:
@hmusum or @freva please review and merge


hmusum (approved):

hmusum (changes requested):
Build fails with:

[ERROR] /sd/workspace/src/github.com/vespa-engine/vespa/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/ApplicationController.java:[638,18] cannot infer type for local variable postDeployment

jonmv:
😢

hmusum (approved):

jonmv:
Wonder if it was JDK18 locally that made this pass?

PR #26250, opened 2023-03-01
Record UP events also when not having been down

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeHealthTracker.java
```diff
@@ -42,14 +42,13 @@ public NodeHealthTracker(ServiceMonitor serviceMonitor, NodeRepository nodeRepos
...
     @Override
     protected double maintain() {
-        return updateActiveNodeDownState();
+        return updateNodeHealth();
     }
 
     /**
-     * If the node is down (see {@link #allDown}), and there is no "down" history record, we add it.
-     * Otherwise we remove any "down" history record.
+     * Update UO and DOWN node records for each node as they change.
```
freva, line 49:
```suggestion
     * Update UP and DOWN node records for each node as they change.
```

freva (dismissed):

PR #26253, opened 2023-03-01
Use typesafe time in vespalog

baldersheim:
@vekterli PR

### DIFF
--- a/vespalog/src/vespa/log/bufferedlogger.cpp
+++ b/vespalog/src/vespa/log/bufferedlogger.cpp
@@ -193,9 +193,7 @@ BackingBuffer::BackingBuffer()
 {
 }
 
-BackingBuffer::~BackingBuffer()
-{
-}
+BackingBuffer::~BackingBuffer() = default;
 
 BufferedLogger::BufferedLogger()
 {
@@ -279,16 +277,12 @@ void
 BackingBuffer::flush()
 {
     std::lock_guard<std::mutex> guard(_mutex);
-    for (LogCacheBack::const_iterator it = _cacheBack.begin();
-         it != _cacheBack.end(); ++it)
-    {
-        log(*it);
+    for (const auto & entry : _cacheBack) {
+        log(entry);
     }
     _cacheBack.clear();
-    for (LogCacheFront::const_iterator it = _cacheFront.begin();
-         it != _cacheFront.end(); ++it)
-    {
-        log(*it);
+    for (const auto & entry : _cacheFront) {
+        log(entry);
     }
     _cacheFront.clear();
 }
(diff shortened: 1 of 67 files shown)


vespalog/src/vespa/log/log.h
```diff
@@ -144,19 +147,24 @@ namespace ns_log {
...
+
+constexpr int64_t
+count_s(duration d) {
+    return std::chrono::duration_cast<std::chrono::seconds>(d).count();
+}
+
+constexpr int64_t
+count_us(duration d) {
+    return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
+}
```
vekterli, line 161: These can probably both be `noexcept`

vespalog/src/vespa/log/log.h
```diff
@@ -219,7 +227,7 @@ class Logger {
      *
      * @param timestamp Time in microseconds.
      */
-    void doLogCore(uint64_t timestamp, LogLevel level,
+    void doLogCore(system_time timestamp, LogLevel level,
```
vekterli, line 230: Is there any way we can refactor things so that this method doesn't have to have its signature be visible in the main `log.h` file? The `Timer` should be easy to move away and just forward declare. I'm not sure how much heft there is in `<chrono>` but since C++20 adds chrono printing to both both `std::ostream` and `std::formatter` (not yet in GCC's STL), it might grow in transitive size over time. It might be worth reducing the number of transitive includes for a file included by so many others.

vekterli (dismissed):
:+1:

vespalog/src/vespa/log/log.h
baldersheim: Fixed

vespalog/src/vespa/log/log.h
```diff
@@ -144,19 +147,24 @@ namespace ns_log {
 
 class LogTarget;
 class ControlFile;
+using system_time  = std::chrono::system_clock::time_point;
+using duration  = std::chrono::nanoseconds;
```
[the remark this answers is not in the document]
baldersheim: Fixed

vespalog/src/vespa/log/bufferedlogger.cpp
```diff
@@ -330,10 +341,10 @@ BackingBuffer::log(const Entry& e) const
     if (e._count > 1) {
         std::ostringstream ost;
         ost << e._message << " (Repeated " << (e._count - 1)
-            << " times since " << (e._timestamp / 1000000) << "."
-            << std::setw(6) << std::setfill('0') << (e._timestamp % 1000000)
+            << " times since " << count_s(e._timestamp.time_since_epoch()) << "."
```
vekterli, line 344: Since this is to be read by humans, could use this as an opportunity to emit as a date/time instead of a UNIX timestamp

vekterli (dismissed):

vespalog/src/vespa/log/bufferedlogger.cpp
baldersheim: Wait for std::format on gcc

PR #26256, opened 2023-03-01
Arnej/handle more rank profiles

arnej27959:
@bratseth please review
@bjorncs FYI


bratseth (approved):

bjorncs (commented):
Looks like the `expressions-as-arguments` test configuration is not used in any tests?

PR #26262, opened 2023-03-01
Always propagate the document type to the internal StructFieldValue.

geirst:
After deserialization of an empty document the internal StructFieldValue would not have the document type. Then, if a StringFieldValue of that empty document was updated (e.g. by an AssignValueUpdate), later lazy deserialization of its annotations would crash as the document type was not present.

@toregge please review
@baldersheim FYI


toregge (approved):
👍

baldersheim (approved):

PR #26264, opened 2023-03-01
Unify on system_time from chrono for file modification time.

baldersheim:
@havardpe PR

### DIFF
--- a/vespalib/src/vespa/fastos/unix_file.cpp
+++ b/vespalib/src/vespa/fastos/unix_file.cpp
@@ -94,19 +94,13 @@ FastOS_UNIX_File::Stat(const char *filename, FastOS_StatInfo *statInfo)
         statInfo->_isRegular = S_ISREG(stbuf.st_mode);
         statInfo->_isDirectory = S_ISDIR(stbuf.st_mode);
         statInfo->_size = static_cast<int64_t>(stbuf.st_size);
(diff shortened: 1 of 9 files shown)


vespalib/src/vespa/fastos/unix_file.cpp
```diff
@@ -94,19 +94,13 @@ FastOS_UNIX_File::Stat(const char *filename, FastOS_StatInfo *statInfo)
         statInfo->_isRegular = S_ISREG(stbuf.st_mode);
         statInfo->_isDirectory = S_ISDIR(stbuf.st_mode);
         statInfo->_size = static_cast<int64_t>(stbuf.st_size);
-        statInfo->_modifiedTime = stbuf.st_mtime;
+        uint64_t modTimeNS = stbuf.st_mtime * 1000 * 1000 * 1000lu;
```
toregge, line 97: Consider replacing `1000 * 1000 * 1000lu` with `1000lu * 1000 * 1000` or just `1000 * 1000 * 1000` (since stbuf.st_mtime should already be 64-bit on both 64-bit linux and 64-bit darwin).

havardpe (dismissed):

vespalib/src/vespa/fastos/unix_file.cpp
baldersheim: Fixed with a constexpr constant.

PR #26271, opened 2023-03-02
Correctly classify transient/retriable errors server side for legacy feed API

bjorncs:


bjorncs:
FYI @vekterli

baldersheim (dismissed):
but I think @vekterli should let his voice be heard too.

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/ErrorCode.java
```diff
@@ -30,4 +35,16 @@ public boolean isTransient() {
         return _transient;
     }
 
+    static ErrorCode fromBusError(Error mbusError) {
+        Collection<Integer> fatalsHandledAsTransient = Set.of(
```
vekterli, line 39: This could probably be moved out to a `private static final` since it is immutable

vespaclient-container-plugin/src/main/java/com/yahoo/vespa/http/server/ErrorCode.java
```diff
@@ -30,4 +35,16 @@ public boolean isTransient() {
...
+                com.yahoo.messagebus.ErrorCode.SEND_QUEUE_CLOSED,
+                com.yahoo.messagebus.ErrorCode.ILLEGAL_ROUTE,
+                com.yahoo.messagebus.ErrorCode.NO_SERVICES_FOR_ROUTE,
+                com.yahoo.messagebus.ErrorCode.NETWORK_ERROR,
+                com.yahoo.messagebus.ErrorCode.SEQUENCE_ERROR,
+                com.yahoo.messagebus.ErrorCode.NETWORK_SHUTDOWN,
+                com.yahoo.messagebus.ErrorCode.TIMEOUT);
```
vekterli, line 46: Can we risk that a client expects timeouts to eventually be fatal to avoid resending "forever" if the target cluster is having issues?

vekterli (dismissed):
:+1:

PR #26272, opened 2023-03-02
Revert "Run ArtifactExpirer more often in cd systems"

hmusum:
Reverts vespa-engine/vespa#26269

Please review only, will revert when fixes to artifact expiry have been verified


hmusum:
Artifact expiry works again now, not needed

hmusum:
@aressem PTAL

aressem (approved):

PR #26274, opened 2023-03-02
smart intrusive reference counting

havardpe:
@vekterli please review
@baldersheim @toregge @geirst @arnej27959 FYI

### DIFF
--- a/vespalib/src/tests/ref_counted/ref_counted_test.cpp
+++ b/vespalib/src/tests/ref_counted/ref_counted_test.cpp
@@ -1,0 +1,218 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include <vespa/vespalib/util/ref_counted.h>
+#include <vespa/vespalib/util/gate.h>
+#include <vespa/vespalib/util/thread.h>
+#include <vespa/vespalib/gtest/gtest.h>
+
+using namespace vespalib;
+
+struct Base : enable_ref_counted {
+    static std::atomic<int> ctor_cnt;
+    static std::atomic<int> dtor_cnt;
+    int val;
+    Base(int val_in) : val(val_in) {
+        ctor_cnt.fetch_add(1, std::memory_order_relaxed);
+    }
+    ~Base() {
+        dtor_cnt.fetch_add(1, std::memory_order_relaxed);
+    }
+};
+std::atomic<int> Base::ctor_cnt = 0;
+std::atomic<int> Base::dtor_cnt = 0;
+
+struct Leaf : Base {
+    static std::atomic<int> ctor_cnt;
+    static std::atomic<int> dtor_cnt;
+    Leaf(int val_in) : Base(val_in) {
+        ctor_cnt.fetch_add(1, std::memory_order_relaxed);
+    }
+    ~Leaf() {
+        dtor_cnt.fetch_add(1, std::memory_order_relaxed);
+    }
+};
+std::atomic<int> Leaf::ctor_cnt = 0;
+std::atomic<int> Leaf::dtor_cnt = 0;
+
+// check that the expected number of objects have been created and
+// destroyed while this object was in scope.
+struct CheckObjects {
+    int expect_base;
+    int expect_leaf;
+    int old_base_ctor;
+    int old_base_dtor;
+    int old_leaf_ctor;
+    int old_leaf_dtor;
+    CheckObjects(int expect_base_in = 0, int expect_leaf_in = 0)
+      : expect_base(expect_base_in),
+        expect_leaf(expect_leaf_in)
+    {
+        old_base_ctor = Base::ctor_cnt.load(std::memory_order_relaxed);
+        old_base_dtor = Base::dtor_cnt.load(std::memory_order_relaxed);
+        old_leaf_ctor = Leaf::ctor_cnt.load(std::memory_order_relaxed);
+        old_leaf_dtor = Leaf::dtor_cnt.load(std::memory_order_relaxed);
+    }
+    ~CheckObjects() {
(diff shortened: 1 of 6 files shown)


vespalib/src/vespa/vespalib/util/ref_counted.h
```diff
@@ -0,0 +1,117 @@
...
+private:
+    T *_ptr;
+    ref_counted(T *ptr) noexcept : _ptr(ptr) {}
+    void maybe_subref() noexcept {
+        if (_ptr) {
+            _ptr->internal_subref();
+        }
+    }
+    static T *maybe_addref(T *ptr) noexcept {
+        if (ptr) {
```
vekterli, line 60: Would it be beneficial with `[[likely]]` on these branches in `maybe_subref`/`maybe_addref` (assuming the common case is that `_ptr` is present)?

vespalib/src/vespa/vespalib/util/ref_counted.h
```diff
@@ -0,0 +1,117 @@
...
+            ptr->internal_addref();
+        }
+        return ptr;
+    }
+public:
+    ref_counted() noexcept : _ptr(nullptr) {}
+    ref_counted(ref_counted &&rhs) noexcept : _ptr(std::exchange(rhs._ptr, nullptr)) {}
+    ref_counted(const ref_counted &rhs) noexcept : _ptr(maybe_addref(rhs._ptr)) {}
+    ref_counted &operator=(ref_counted &&rhs) noexcept {
+        maybe_subref();
```
vekterli, line 70: Do these assignment ops need a self-assignment guard? I.e. in the case of `a = a` or `a = std::move(a)` where `a` is the only live ref, do we risk subref'ing ourselves into oblivion _before_ bumping (or stealing) the refcount again?

vespalib/src/vespa/vespalib/util/ref_counted.h
```diff
@@ -0,0 +1,117 @@
...
+        _ptr = std::exchange(rhs._ptr, nullptr);
+        return *this;
+    }
+    ref_counted &operator=(const ref_counted &rhs) noexcept {
+        maybe_subref();
+        _ptr = maybe_addref(rhs._ptr);
+        return *this;
+    }
+    template <typename X> ref_counted(ref_counted<X> &&rhs) noexcept : _ptr(std::exchange(rhs._ptr, nullptr)) {}
```
vekterli, line 79: Could be pretty neat to have a concept requirement here that `X*` is convertible to `T*` for added compiler helpfulness 🦄 But can be done as a followup, if we want it

vespalib/src/vespa/vespalib/util/ref_counted.h
```diff
@@ -0,0 +1,117 @@
...
+template <typename T, typename ... Args>
+ref_counted<T> make_ref_counted(Args && ... args) {
+    return ref_counted<T>::internal_attach(new T(std::forward<Args>(args)...));
+}
+
+// similar to shared_from_this;
+// create a new handle to a reference counted object
+// (NB: object must still be valid)
+template <typename T>
+ref_counted<T> ref_counted_from(T &t) {
```
vekterli, line 112: Could probably also be `noexcept`

vespalib/src/tests/ref_counted/ref_counted_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+    for (bool has_src: {true, false}) {
+        for (bool has_dst: {true, false}) {
+            for (bool same: {true, false}) {
+                if (same) {
+                    CheckObjects check(2);
+                    ref_counted<Base> empty;
+                    ref_counted<Base> obj1 = make_ref_counted<Base>(10);
+                    ref_counted<Base> obj2 = make_ref_counted<Base>(20);
+                    ref_counted<Base> src = has_src ? obj1 : empty;
+                    ref_counted<Base> dst = has_dst ? obj2 : empty;
```
vekterli, line 158: Do we want `src` and `dst` to be _refs_ to `ref_counted` instead of values to avoid any implicit reference count bumping here?

vekterli (dismissed):
:+1: mostly one question about whether self-assignment guarding is needed

vespalib/src/vespa/vespalib/util/ref_counted.h
vekterli: Disregard this. SFINAE-removal makes things strictly worse when adding a concept requirement here...

vespalib/src/vespa/vespalib/util/ref_counted.h
havardpe: will do

vespalib/src/vespa/vespalib/util/ref_counted.h
havardpe: The concept requirement makes the error message worse since it removes function form the overload set and you end up with an error saying ref_counted<Foo> cannot be converted to ref_counted<Bar> instead of an error saying that Foo* cannot be converted to Bar*.

vespalib/src/vespa/vespalib/util/ref_counted.h
havardpe: yes

vespalib/src/tests/ref_counted/ref_counted_test.cpp
havardpe: No; calling the copy constructor with an empty/non-empty ref_counted is an intended part of the test

vespalib/src/vespa/vespalib/util/ref_counted.h
havardpe: Will re-structure the code to make self-assignment safe.

havardpe:
@vekterli PTAL

vekterli (approved):
:+1:

PR #26282, opened 2023-03-03
Only add each deployment as account URI or tenant URI

freva:


freva:
Will do this differently

PR #26287, opened 2023-03-03
Store account in deployment

freva:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArchiveUriUpdater.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArchiveUriUpdater.java
@@ -6,10 +6,10 @@ import com.yahoo.config.provision.TenantName;
 import com.yahoo.config.provision.zone.ZoneId;
 import com.yahoo.vespa.hosted.controller.ApplicationController;
 import com.yahoo.vespa.hosted.controller.Controller;
-import com.yahoo.vespa.hosted.controller.api.identifiers.DeploymentId;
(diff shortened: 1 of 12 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArchiveUriUpdater.java
```diff
@@ -55,9 +57,8 @@ protected double maintain() {
         for (var application : applications.asList()) {
             for (var instance : application.instances().values()) {
                 for (var deployment : instance.deployments().values()) {
-                    tenantsByZone.get(deployment.zone()).add(instance.id().tenant());
-                    applications.decideCloudAccountOf(new DeploymentId(instance.id(), deployment.zone()), application.deploymentSpec())
-                            .ifPresent(account -> accountsByZone.get(deployment.zone()).add(account));
+                    if (zoneRegistry.isEnclave(deployment.cloudAccount())) accountsByZone.get(deployment.zone()).add(deployment.cloudAccount());
+                    else tenantsByZone.get(deployment.zone()).add(instance.id().tenant());
```
jonmv, line 61: SO this won't work before all deployemnts are re-done, but I guess that's OK?

jonmv (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArchiveUriUpdater.java
freva: Yes, there are no exclave deployments currently and that method will return false since the default cloud account is `CloudAccount.empty`, which will result in the old behavior

PR #26288, opened 2023-03-03
Resolve config when we discover a new config generation is active

hmusum:
Calling configActivated() will end up in going through all delayed responses, putting them on a queue and wait for configs to be resolved. This happens with the same executor that call configActivated() so it might end up in a deadlock when all threads have been used. Resolving config immediately instead should avoid this.

@arnej27959 I think you were involved making the original change, please review


arnej27959 (dismissed):
This change looks good to me 👍

arnej27959 (commented):
I don’t remember details of this code but after staring at it I think it’s possible that we should avoid doing rpcserver.delayresponse in this case?

hmusum:
Yes, I think that's right, will fix

PR #26289, opened 2023-03-03
Make it possible to specify billId when adding line item

oyving:


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/billing/MockBillingController.java
```diff
@@ -2,6 +2,7 @@
 package com.yahoo.vespa.hosted.controller.api.integration.billing;
 
 import com.yahoo.config.provision.TenantName;
+import org.apache.hc.core5.http.NotImplementedException;
```
olaaun, line 5: Unused

olaaun (dismissed):

olaaun (approved):

PR #26291, opened 2023-03-03
Use the metric provider returned after configuration

bratseth:


application/src/test/java/com/yahoo/application/ApplicationTest.java
```diff
@@ -262,6 +265,25 @@ void component() throws Exception {
...
+            var metrics = (SimpleMetricConsumer)component.metrics().newInstance(); // not actually a new instance
+            assertNotNull(metrics);
+            int maxWaitMs = 10000;
+            Bucket snapshot = null;
+            while (maxWaitMs-- > 0 && ( snapshot = metrics.receiver().getSnapshot() ) == null) {
+                Thread.sleep(1);
+            }
+            assertNotNull(snapshot);
+            assertEquals(1, snapshot.getValuesForMetric("application_generation").size());
+            assertEquals(0, snapshot.getValuesForMetric("application_generation").iterator().next().getValue().getLast());
```
gjoranv, line 283: Out of curiosity, where did the test fail?

gjoranv (commented):
👍

application/src/test/java/com/yahoo/application/ApplicationTest.java
bratseth: On 282 - it never got the metric. 

In a real setup we still never get the metric any more though, so this didn't solve the problem.

PR #26312, opened 2023-03-06
Add raw as attribute vector basic type.

toregge:
@geirst : please review


baldersheim (approved):

baldersheim:
Also add isRawType to IAttributeVector similar to isStringType()

PR #26318, opened 2023-03-06
start wireguard earlier

andreer:


andreer:
let's begin with this to quickly improve the issue before starting any rewrite to systemd

gjoranv (approved):

PR #26322, opened 2023-03-06
Use java 17 during %check build stage.

toregge:
@aressem : please review

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -466,6 +466,15 @@ VERSION=%{version} CI=true make -C client/go install-all
 
 %check
 %if ! 0%{?installdir:1}
+%if 0%{?_java_home:1}
+export JAVA_HOME=%{?_java_home}
+%else
+export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+%endif
+export PATH="$JAVA_HOME/bin:$PATH"
+%if 0%{?el8}
+python3.9 -m pip install --user pytest
+%endif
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -466,6 +466,15 @@ VERSION=%{version} CI=true make -C client/go install-all
...
 %check
 %if ! 0%{?installdir:1}
+%if 0%{?_java_home:1}
+export JAVA_HOME=%{?_java_home}
+%else
+export JAVA_HOME=/usr/lib/jvm/java-17-openjdk
+%endif
+export PATH="$JAVA_HOME/bin:$PATH"
+%if 0%{?el8}
+python3.9 -m pip install --user pytest
```
aressem, line 476: Not a good idea to install stuff as part of the rpmbuild ?

aressem (changes requested):

dist/vespa.spec
toregge: Not installing pytest causes a failure due to pytest not being installed.

aressem: Ok. I do wish we could express this dependency with `BuildRequires`.

aressem (approved):

dist/vespa.spec
baldersheim: Is this due to ANN testing ?

aressem: It is for the ANN benchmark unit test.

PR #26325, opened 2023-03-07
- Handle single value raw attributes in grouping.

baldersheim:
- Let stringattributes implement get_raw interface.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/stringbase.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/stringbase.cpp
@@ -86,9 +86,10 @@ StringAttribute::getFloat(DocId doc) const {
 }
 
 vespalib::ConstArrayRef<char>
-StringAttribute::get_raw(DocId) const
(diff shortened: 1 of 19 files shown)


toregge (dismissed):

toregge:
@geirst : FYI

searchlib/src/vespa/searchlib/attribute/stringbase.cpp
```diff
@@ -86,9 +86,10 @@ StringAttribute::getFloat(DocId doc) const {
 }
 
 vespalib::ConstArrayRef<char>
-StringAttribute::get_raw(DocId) const
+StringAttribute::get_raw(DocId doc) const
```
geirst, line 89: Some of the StringAttribute unit tests should be extended to use this API as well.

baldersheim: Done.

PR #26345, opened 2023-03-07
In order to detect faster that nodes are up, ping failed nodes more o…

baldersheim:
…ften (5x).

@bratseth and @hmusum PR


bratseth (approved):
I guess you think this is worth it for some specific real-world reason ...

baldersheim:
Unfortunately, yes.
When testing a new cluster, feed is allowed once a node is up, but search interface will be up some time after that and at different times. This will bring the difference down from 0.5-2.0s to below 0.5. That is sufficient to make tests stable.

An alternative would be to have have special config for tests preventing search until all nodes were up.
But special config for tests are most of the time a bad idea.
Will hold to see if I get a better idea.

baldersheim:
The whole ping mechanism is suboptimal and needs a redesign.
Should use longpoll instead to have faster detection of nodes up/down.

PR #26347, opened 2023-03-07
Only setup `RankProfilesEvaluator` for schemas with 'global-phase'

bjorncs:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static com.yahoo.vespa.model.container.PlatformBundles.SEARCH_AND_DOCPROC_BUNDLE;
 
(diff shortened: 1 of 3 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
```diff
@@ -65,6 +67,11 @@ public ContainerSearch(DeployState deployState, ApplicationContainerCluster clus
         cluster.addSearchAndDocprocBundles();
     }
 
+    private static Collection<String> getSchemasWithGlobalPhase(DeployState state) {
+        return state.rankProfileRegistry().all().stream()
+                .filter(rp -> rp.getGlobalPhase() != null).map(rp -> rp.schema().getName()).toList();
```
arnej27959, line 72: consider something like `.collect(Collectors.toSet())` instead of `.toList()`

arnej27959 (dismissed):
looks good 👍

config-model/src/main/java/com/yahoo/vespa/model/container/search/ContainerSearch.java
bjorncs: Addressed in [4b55e59](https://github.com/vespa-engine/vespa/commit/4b55e59b0bdda6559a40addf0ede434ab955dc07)

PR #26359, opened 2023-03-08
Arnej/fix tinybert test

arnej27959:
@bratseth please review
note: the caching introduced here was needed to make the TinyBert setup (added as unit test) complete in a reasonable time.


bratseth (approved):
Nice

PR #26362, opened 2023-03-08
add ExpressionNode computing an output from an ONNX model

arnej27959:
@bratseth please review
@bjorncs @lesters FYI

### DIFF
--- a/model-evaluation/src/main/java/ai/vespa/models/evaluation/OnnxExpressionNode.java
+++ b/model-evaluation/src/main/java/ai/vespa/models/evaluation/OnnxExpressionNode.java
@@ -1,0 +1,102 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package ai.vespa.models.evaluation;
+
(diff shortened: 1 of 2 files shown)


bratseth (approved):
👍

model-evaluation/src/main/java/ai/vespa/models/evaluation/OnnxExpressionNode.java
```diff
@@ -0,0 +1,102 @@
...
+    static Optional<Reference> parseOnnxInput(String input) {
+        var optRef = Reference.simple(input);
+        if (optRef.isPresent()) {
+            return optRef;
+        }
+        try {
+            var ref = Reference.fromIdentifier(input);
+            return Optional.of(ref);
+        } catch (Exception e) {
+            // fallthrough
```
bjorncs, line 60: When can this happen?

arnej27959: > When can this happen?

only on unexpected configuration; it's there only to make it possible to give a better error message just above.

PR #26365, opened 2023-03-08
Add workaround to restart wg process early after host-admin restart.

gjoranv:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -498,7 +505,7 @@ void doConverge(NodeAgentContext context) {
                 }
 
                 aclMaintainer.ifPresent(maintainer -> maintainer.converge(context));
-                wireguardTasks.forEach(task -> task.converge(context));
+                if (! wireguardTasksRun) wireguardTasks.forEach(task -> task.converge(context));
```
hakonhall, line 508: wireguardTasksRun must be set to true if the container is restarted.

andreer (dismissed):
Hacky but should be a good improvement for now.

hakonhall (approved):

PR #26368, opened 2023-03-08
Log at lower level when installing bundles

hmusum:
@gjoranv please review
@kkraune FYI

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
+++ b/container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
@@ -70,7 +70,7 @@ public class ApplicationBundleLoader {
 
         bundlesFromNewGeneration = installBundles(newFileReferences);
         BundleStarter.startBundles(activeBundles.values());
-        log.info(installedBundlesMessage());
+        log.fine(installedBundlesMessage());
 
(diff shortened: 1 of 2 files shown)


kkraune:
👍

baldersheim (approved):

container-core/src/main/java/com/yahoo/container/core/config/ApplicationBundleLoader.java
```diff
@@ -70,7 +70,7 @@ public synchronized void useBundles(List<FileReference> newFileReferences) {
 
         bundlesFromNewGeneration = installBundles(newFileReferences);
         BundleStarter.startBundles(activeBundles.values());
-        log.info(installedBundlesMessage());
+        log.fine(installedBundlesMessage());
```
gjoranv, line 73: Why don't we want to see this?

hmusum:
For me it's just noise. I would guess it's useful if a customer has issues with bundle loading or we want to debug bundle loading issues? It's shown both for application containers and other containers (config server, cluster controller, metrics proxy), so if it's possible to log less for some types of services that could be a thing to look into.

If you think this we need this log message I could add it back, of course.

gjoranv:
Yes, I'll add a Jira task to myself.

PR #26369, opened 2023-03-09
Use lower log level when waiting for zookeeper server to come up

hmusum:


kkraune (approved):
All right! I was puzzeled about the "dummy-node-" name ;-)

PR #26373, opened 2023-03-09
Reduce log level for container env and args

hmusum:
Trying to reduce log lines at startup, I guess this was used when developing startup script and is not needed anymore @arnej27959  please review
@kkraune FYI


hmusum:
Good, closing this PR then

PR #26375, opened 2023-03-09
Don't use NodeResources::add to create ResourceSnapshot

oyving:

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/resource/ResourceSnapshot.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/resource/ResourceSnapshot.java
@@ -53,7 +53,7 @@ public class ResourceSnapshot {
 
         var resources = nodes.stream()
                 .map(Node::resources)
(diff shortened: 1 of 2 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/resource/ResourceSnapshot.java
```diff
@@ -95,4 +95,22 @@ public boolean equals(Object o) {
...
         return Objects.hash(applicationId, resources, timestamp, zoneId, version);
     }
+
+    /* This function does pretty much the same thing as NodeResources::add, but it allows adding resources
+     * where some dimensions that are not relevant for billing (yet).
+     *
+     * TODO: Make this code respect all dimensions.
+     */
+    private static NodeResources addResources(NodeResources a, NodeResources b) {
+        assert a.architecture() == b.architecture() || a.architecture() == NodeResources.Architecture.any || b.architecture() == NodeResources.Architecture.any;
```
bratseth, line 105: Do we really want the controller to die over this?

olaaun (approved):

PR #26392, opened 2023-03-10
Remove high-level locking of maintenance deployments

mpolden:
@bratseth @freva

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
@@ -6,7 +6,6 @@ import com.yahoo.config.provision.ApplicationLockException;
 import com.yahoo.config.provision.ClusterResources;
 import com.yahoo.config.provision.ClusterSpec;
 import com.yahoo.config.provision.Deployer;
-import com.yahoo.config.provision.Environment;
 import com.yahoo.jdisc.Metric;
 import com.yahoo.vespa.hosted.provision.Node;
 import com.yahoo.vespa.hosted.provision.NodeList;
@@ -19,6 +18,7 @@ import com.yahoo.vespa.hosted.provision.autoscale.Autoscaler;
 import com.yahoo.vespa.hosted.provision.autoscale.Autoscaling;
 import com.yahoo.vespa.hosted.provision.autoscale.NodeMetricSnapshot;
 import com.yahoo.vespa.hosted.provision.node.History;
+
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Map;
@@ -68,6 +68,7 @@ public class AutoscalingMaintainer extends NodeRepositoryMaintainer {
      * @return true if an autoscaling decision was made or nothing should be done, false if there was an error
      */
     private boolean autoscale(ApplicationId applicationId, ClusterSpec.Id clusterId) {
+        boolean redeploy = false;
         try (var lock = nodeRepository().applications().lock(applicationId)) {
             Optional<Application> application = nodeRepository().applications().get(applicationId);
             if (application.isEmpty()) return true;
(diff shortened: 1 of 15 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MaintenanceDeployment.java
```diff
@@ -161,39 +118,37 @@ public boolean execute(boolean verifyTarget,
...
-                                              .first();
-                        if (expectedNewNode.isEmpty()) return false;
-                        if (!expectedNewNode.get().hasParent(toHost.hostname())) return false;
-                    }
-                    if ( deployment.activate().isEmpty()) return false;
-
-                    log.info(agent + " redeployed " + application + " to " +
-                             ( verifyTarget ? this : "move " + (node + " from " + fromHost.hostname())));
-                    return true;
+            MaintenanceDeployment deployment = new MaintenanceDeployment(application, deployer, metric, nodeRepository);
```
bratseth, line 121: Are you sure we don't need to take the lock in this method?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/AutoscalingMaintainer.java
```diff
@@ -55,23 +55,32 @@ protected double maintain() {
         for (var applicationNodes : activeNodesByApplication().entrySet()) {
             for (var clusterNodes : nodesByCluster(applicationNodes.getValue()).entrySet()) {
                 attempts++;
-                if ( ! autoscale(applicationNodes.getKey(), clusterNodes.getKey()))
-                    failures++;
+                Result result = autoscale(applicationNodes.getKey(), clusterNodes.getKey());
```
bratseth, line 58: I think this is fine, but why not do it simply by doing the deploy attempt outside the try-with block in autoscale?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
```diff
@@ -62,27 +61,32 @@ protected double maintain() {
...
-                    log.info("Redeployed " + application + " at session " + session.get() + " to deactivate retired nodes: " + nodeList);
+                if (removeNodes(application, removal, nodes)) {
                     successes++;
                 }
             }
         }
         return attempts == 0 ? 1.0 : ((double)successes / attempts);
     }
 
+    /** Mark nodes as removable and redeploy application */
```
bratseth, line 72: I think there was missing locking here since the node list determining what to remove was determined before locking ... and the new code still has that problem.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MaintenanceDeployment.java
mpolden: I suppose since we have multiple node movers, there's a higher chance their deployment's prepare+active interleaving. Maybe we should have a separate node mover lock?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirer.java
mpolden: True, will fix.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MaintenanceDeployment.java
bratseth: Yes, that sounds like a good idea.

mpolden:
PTAL

freva (approved):
Great, now we hold this maintenance lock for all deployments except external ones, I guess that's OK since they aren't as frequent?

bratseth (approved):
Great!

PR #26395, opened 2023-03-10
Expose from/to visitor timestamps in Document V1 API

vekterli:
@jonmv please review.

This directly mirrors `vespa-visit` (and the underlying `VisitorParameters`) in being in microseconds from UTC epoch. It can be argued whether this should be in _seconds_, as Vespa timestamps operations with wall-clock second time combined with a synthetic microsecond counter. But since the existing APIs use microseconds, I'm leaving this as micros as well to keep things consistent.


jonmv (approved):

jonmv:
Nice :)

PR #26401, opened 2023-03-10
- Just start from the beginning to locate a free buffer.

baldersheim:
- Assert that the requested buffer id is legal.

@toregge and @geirst PR

### DIFF
--- a/vespalib/src/vespa/vespalib/datastore/datastorebase.cpp
+++ b/vespalib/src/vespa/vespalib/datastore/datastorebase.cpp
@@ -106,15 +106,8 @@ DataStoreBase::~DataStoreBase()
 void
 DataStoreBase::switch_primary_buffer(uint32_t typeId, size_t elemsNeeded)
 {
-    size_t buffer_id = primary_buffer_id(typeId);
-    for (size_t i = 0; i < getNumBuffers(); ++i) {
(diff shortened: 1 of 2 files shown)


vespalib/src/vespa/vespalib/datastore/datastorebase.cpp
```diff
@@ -164,6 +157,23 @@ DataStoreBase::consider_grow_active_buffer(uint32_t type_id, size_t elems_needed
...
 
+uint32_t
+DataStoreBase::getFirstFreeBufferId() {
+    for (uint32_t buffer_id = 0; buffer_id < _states.size(); buffer_id++) {
+        if (getBufferState(buffer_id).isFree()) {
+            return buffer_id;
+        }
+    }
+    // Need next(new) buffer
+    return _states.size();
```
toregge, line 168: DataStoreBase::onActive() needs to assert that buffer id is reasonable to avoid a buffer overflow.

toregge (approved):

vespalib/src/vespa/vespalib/datastore/datastorebase.cpp
baldersheim: Correct, that is part of upcomming PR.

PR #26413, opened 2023-03-13
Be explicit about lbound/ubound for bucket DB iteration and add lbound variant

vekterli:
@geirst please review

The DB API was rather coy about whether `forEach` had lower or upper bound semantics with regards to the bucket ID passed in as a starting point. Be explicit and add a lower-bound variant.


geirst (approved):
👍

PR #26424, opened 2023-03-13
Add cloud account to node spec

freva:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/NodeSpec.java
```diff
@@ -125,6 +128,7 @@ public NodeSpec(
         this.currentDockerImage = Objects.requireNonNull(currentDockerImage);
         this.state = Objects.requireNonNull(state);
         this.type = Objects.requireNonNull(type);
+        this.cloudAccount = Objects.requireNonNull(cloudAccount);
```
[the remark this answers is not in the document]
freva: The ones that don't have will get `CloudAccount.empty`
https://github.com/vespa-engine/vespa/blob/3556fbbb81b9785902341f02478e4a038845a163/node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/RealNodeRepository.java#L206

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/configserver/noderepository/NodeSpec.java
```diff
@@ -526,6 +538,11 @@ public Builder type(NodeType nodeType) {
             return this;
         }
 
+        public Builder cloudAccount(CloudAccount cloudAccount) {
+            this.cloudAccount = cloudAccount;
```
hakonhall, line 542: requireNonNull here?

freva: We don't have any validation in any other builder methods in this class and instead rely on the validation in the main constructor

hakonhall (approved):

PR #26429, opened 2023-03-14
Add and use percentage unit

bratseth:
Maybe there are more?


yngveaasheim:
Thanks! Those are the ones I found while going through our dashboards for this.

PR #26437, opened 2023-03-14
add special handling of TensorFunctionNode containing Generate function

arnej27959:
@bratseth please review
@lesters @jobergum @bjorncs FYI


bratseth (approved):
This became fairly nice imo

PR #26445, opened 2023-03-15
Turn off logging for containers for some classes

hmusum:
Make it possible to set logctl specs for all types of containers and set default logctl specs for some external classes that we want to avoid in vespa log

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/Admin.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/Admin.java
@@ -94,7 +94,7 @@ public class Admin extends TreeConfigProducer<AnyConfigProducer> implements Seri
         this.multitenant = multitenant;
         this.fileDistribution = new FileDistributionConfigProducer(parent);
         this.applicationType = applicationType;
-        this.logctlSpecs.addAll(setDefaultLogctlSpecs());
+        this.logctlSpecs.addAll(defaultLogctlSpecs());
     }
 
     public Configserver getConfigserver() { return defaultConfigserver; }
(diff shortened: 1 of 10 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/Admin.java
```diff
@@ -331,4 +325,25 @@ public boolean multitenant() {
...
     public ApplicationType getApplicationType() { return applicationType; }
 
+    public List<LogctlSpec> getLogctlSpecs() {
+        return logctlSpecs;
+    }
+    public void addLogctlCommand(String componentSpec, String levelsModSpec) {
+        logctlSpecs.add(new LogctlSpec(componentSpec,  levelsModSpec));
+    }
+
+    private static Set<LogctlSpec> setDefaultLogctlSpecs() {
```
arnej27959, line 335: consider renaming this method to "alwaysLogctlSpecs()" or just "defaultLogctlSpecs()"

config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
```diff
@@ -124,19 +124,12 @@ public ApplicationContainerCluster(TreeConfigProducer<?> parent, String configSu
...
-            }});
-    }
-
     @Override
     protected void doPrepare(DeployState deployState) {
         addAndSendApplicationBundles(deployState);
         sendUserConfiguredFiles(deployState);
         createEndpointList(deployState);
-        wireLogctlSpecs();
+        super.doPrepare(deployState);
```
arnej27959, line 132: super() should be first in method

arnej27959 (dismissed):
looks good; my LogctlSpec seems clunky in retrospect, I will try to come up with a small improvement.

config-model/src/main/java/com/yahoo/vespa/model/container/ApplicationContainerCluster.java
hmusum: Fixed

config-model/src/main/java/com/yahoo/vespa/model/admin/Admin.java
hmusum: Yup, that's better, fixed.

gjoranv (commented):
👍

PR #26453, opened 2023-03-15
Count more of the memory used by the document metastore.

baldersheim:
@toregge or @geirst PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/documentmetastore/lid_allocator.cpp
+++ b/searchcore/src/vespa/searchcore/proton/documentmetastore/lid_allocator.cpp
@@ -27,11 +27,19 @@ LidAllocator::LidAllocator(uint32_t size,
       _freeLids(size, capacity, genHolder, true, false),
       _usedLids(size, capacity, genHolder, false, true),
       _pendingHoldLids(size, capacity, genHolder, false, false),
-      _lidFreeListConstructed(false),
       _activeLids(size, capacity, genHolder, false, false),
-      _numActiveLids(0u)
-{
+      _numActiveLids(0u),
+      _lidFreeListConstructed(false)
+{ }
 
+vespalib::MemoryUsage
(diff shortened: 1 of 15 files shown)


searchcore/src/vespa/searchcore/proton/documentmetastore/lid_allocator.cpp
```diff
@@ -27,11 +27,19 @@ LidAllocator::LidAllocator(uint32_t size,
...
-      _numActiveLids(0u)
-{
+      _numActiveLids(0u),
+      _lidFreeListConstructed(false)
+{ }
 
+vespalib::MemoryUsage
+LidAllocator::getMemoryUsage() const {
+    vespalib::MemoryUsage usage;
+    size_t allocated = sizeof(*this) + _freeLids.byteSize() + _usedLids.byteSize() +
```
toregge, line 38: Calling byteSize() on a LidStateVector is unsafe unless the proper generation guard is held if called from a reader thread.

toregge (dismissed):

searchcore/src/vespa/searchcore/proton/documentmetastore/lid_hold_list.h
```diff
@@ -37,6 +37,8 @@ class LidHoldList
      **/
     size_t size() const { return _holdList.size(); }
 
+    size_t byteSize() const { return _holdList.size() * sizeof(Element); }
```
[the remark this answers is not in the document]
baldersheim: Good catch added the call to it.

searchcore/src/vespa/searchcore/proton/documentmetastore/lid_allocator.cpp
baldersheim: It is called from the same place as used to call getUsedLidsSize(), updateStat().
That should be fine as it happens as part of writer thread.

PR #26457, opened 2023-03-16
join dimensions must equal common dimensions for optimization

arnej27959:
@lesters please review; also: can you add proper unit tests for ReduceJoin later?
@havardpe could you quickly compare with C++ version?
@bratseth FYI


havardpe:
the C++ version looks for dot product specifically

lesters (approved):

PR #26460, opened 2023-03-16
Revert "Turn off logging for containers for some classes"

hmusum:
Reverts vespa-engine/vespa#26445

Logging is turned off for containers, but seeing lines like this in log (I haven't found where they are created):

[2023-03-16 11:02:20.860] INFO    container        stdout	.com.yahoo.vespa.spifly.repackaged.spifly.BaseActivator   ON  ON  ON  ON OFF  ON OFF OFF

Also seeing

1678964771.464928	ip-10-47-142-7.ec2.internal	361/16725	config-sentinel	sentinel.sentinel.logctl	error	waitpid() failed: Interrupted system call

and some tests that check output from some config verification script started failing, probably due to one of the above issues


hmusum:
@arnej27959 please review

baldersheim (approved):

arnej27959:
the INFO logging is because we run vespa-logctl and this is what it prints on stdout.
waitpid() message was unexpected, will look at it.
config verification fails because of different ordering; will try using List.of() instead of Set.of() as a fix.

hmusum:
Great, thanks for looking into it @arnej27959

PR #26462, opened 2023-03-16
* make InputRecorder handle slice with embedded expressions

arnej27959:
* special handling for Generate and Slice in BindingExtractor

@bratseth please review
@lesters @bjorncs FYI


bratseth (approved):
👍

PR #26466, opened 2023-03-16
Make VespaMetrics a public interface. DefaultMetrics uses suffix enum sets for metric creation

olaaun:
Reference documentation generation will be easier when `VespaMetricSet.java` and `DefaultMetrics.java` are more alike


yngveaasheim (approved):
👍

gjoranv (commented):
👍

PR #26469, opened 2023-03-16
Include mutex to get definition of std::mutex.

toregge:
@vekterli : please review

Use of std::mutex was added in #26463


vekterli (approved):
:+1:

PR #26472, opened 2023-03-16
auto-detect arrays of long or double

havardpe:
@bratseth @baldersheim please review

Note that due to the API, we will create wrappers when native values are added as well as each time they are accessed

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
+++ b/vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
@@ -6,9 +6,93 @@ package com.yahoo.slime;
  */
 final class ArrayValue extends Value {
 
-    private int capacity = 16;
+    static final int initial_capacity = 16;
+
+    private interface Impl {
+        public void prepareFor(ArrayValue self, Type type);
+        public Value add(Value value, int used);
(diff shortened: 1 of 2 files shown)


vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
```diff
@@ -6,43 +6,116 @@
...
-    private int capacity = 16;
+    static final int initial_capacity = 16;
+
+    private interface Impl {
+        public void prepareFor(ArrayValue self, Type type);
+        public Value add(Value value, int used);
+        public Value get(int index);
+    }
+
+    private static final class EmptyImpl implements Impl {
```
baldersheim, line 17: This has no state and could be a singleton. Then you avoid creating a dummy every time.

baldersheim (dismissed):

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
havardpe: good point

havardpe:
@baldersheim PTAL

baldersheim (approved):

PR #26494, opened 2023-03-20
Add utility wrapper around RE2 possible regex prefix match range

vekterli:
@geirst please review

For a strictly _start-anchored_ regex, this provides a lower/upper bound pair that constrains the possible prefix range that may contain a string matching the regex.

### DIFF
--- a/vespalib/src/vespa/vespalib/regex/regex.cpp
+++ b/vespalib/src/vespa/vespalib/regex/regex.cpp
@@ -49,6 +49,16 @@ public:
         }
         return RE2::FullMatch(StringPiece(input.data(), input.size()), _regex);
     }
+
+    std::pair<std::string, std::string> possible_anchored_match_prefix_range() const {
+        constexpr int max_len = 128; // TODO determine a "reasonable" value. RE2 docs are not clear on this.
+        std::string min_prefix, max_prefix;
+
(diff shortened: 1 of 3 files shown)


vespalib/src/vespa/vespalib/regex/regex.cpp
```diff
@@ -49,6 +49,16 @@ class Regex::Impl {
         }
         return RE2::FullMatch(StringPiece(input.data(), input.size()), _regex);
     }
+
+    std::pair<std::string, std::string> possible_anchored_match_prefix_range() const {
+        constexpr int max_len = 128; // TODO determine a "reasonable" value. RE2 docs are not clear on this.
+        std::string min_prefix, max_prefix;
+
+        if (!_regex.PossibleMatchRange(&min_prefix, &max_prefix, max_len)) {
+            return {};
```
geirst, line 58: Consider adding a test case for this code path as well.

geirst (dismissed):
👍

vespalib/src/vespa/vespalib/regex/regex.cpp
vekterli: The easiest way to deterministically get this to actually fail is by supplying a bad regex, so I'll add a test with that.

geirst (approved):
👍

PR #26498, opened 2023-03-20
Limit use of [[maybe_unused]] attribute when using gcc 11.

toregge:
@baldersheim : please review

This is the first part of #26495


toregge:
This pr is obsoleted by #26499

PR #26500, opened 2023-03-20
Reduce log level for some loggers

hmusum:
This is needed to reduce log level for standalone containers, where setting log level cannot be done by config sentinel (since these containers are not started by config sentinel).

### DIFF
--- a/vespalog/src/main/java/com/yahoo/log/VespaLogHandler.java
+++ b/vespalog/src/main/java/com/yahoo/log/VespaLogHandler.java
@@ -2,6 +2,7 @@
 package com.yahoo.log;
 
 import java.io.UnsupportedEncodingException;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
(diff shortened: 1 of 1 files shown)


vespalog/src/main/java/com/yahoo/log/VespaLogHandler.java
```diff
@@ -73,12 +74,28 @@ public synchronized void publish(LogRecord record) {
         closeFileTarget();
     }
 
+    // Reduce log level from info level to fine level for some loggers
+    private Level possiblyReduceLogLevel(String loggerName, Level level) {
+        if (loggerName == null)
+            return level;
+
+        if (Set.of("com.yahoo.vespa.spifly.repackaged.spifly.BaseActivator",
```
arnej27959, line 82: I guess this set should be in a "private static final" member variable, and the method should be static as well

hmusum: Agreed, done

PR #26507, opened 2023-03-21
Basic CLI feed client

mpolden:
This is a working client, but still lacks ordering on doc ID, retrying,
circuit-breaking etc. However, I think that functionality should fit neatly in
this structure.

Also needs more tuning of connection/stream/goroutine count.

@jonmv


client/go/internal/vespa/feed/document.go
```diff
@@ -0,0 +1,96 @@
...
+			}
+			number = &n
+		} else if group == "" && strings.HasPrefix(part, "g=") {
+			group = part[2:]
+			if len(group) == 0 {
+				return DocumentId{}, parseError(serialized)
+			}
+		} else {
+			userSpecific = part
+		}
```
jonmv: Won't this fail on `id:ns:type::n=user-specific` (error, weird but legal) and `id:ns:type:user-specific:foo-bar` (illegal, but "group" is ignored)?

client/go/internal/vespa/feed/document.go
```diff
@@ -1,11 +1,123 @@
...
+	Condition string          `json:"condition"`
+	Create    bool            `json:"create"`
+	Fields    json.RawMessage `json:"fields"`
+}
+
+// Operation returns the operator to perform and the target document ID of the operation.
+func (d Document) Operation() (string, DocumentId, error) {
+	operation := ""
+	id := ""
+	if id != "" {
```
jonmv: `if d.ID != ""`?

client/go/internal/vespa/feed/document.go
```diff
@@ -21,11 +22,10 @@ type Document struct {
 	Fields    json.RawMessage `json:"fields"`
 }
 
-// Operation returns the operator to perform and the target document ID of the operation.
-func (d Document) Operation() (string, DocumentId, error) {
+func (d Document) operation() (string, DocumentId, error) {
 	operation := ""
 	id := ""
-	if id != "" {
+	if d.Id != "" {
```
jonmv: Aha OK.

client/go/internal/vespa/feed/document.go
```diff
@@ -47,6 +47,46 @@ func (d Document) Operation() (string, DocumentId, error) {
...
+	}
+	u.Path = docId.URLPath()
+	u.RawQuery = queryParams.Encode()
+	return httpMethod, u, nil
+}
+
+// Body returns the body part of this document, suitable for sending to the /document/v1 API.
+func (d Document) Body() []byte {
+	jsonObject := `{"fields":`
+	body := make([]byte, 0, len(jsonObject)+len(d.Fields)+1)
```
jonmv: Best JSON generation :p

jonmv (dismissed):
LGTM, but not a go(-routine) expert :)

jonmv:
I guess you need an async http client at some point? Just for curiosity, I understand it as currently allowing up to a fixed (concurrency) number of inflight documents?

client/go/internal/vespa/feed/document.go
mpolden: Thanks! Should be fixed now.

geirst:
> LGTM, but not a go(-routine) expert :)

Perhaps @arnej27959 should have a look as well?

mpolden:
Rebased to fix conflicts.

PR #26508, opened 2023-03-21
Add utility wrapper around RE2 possible regex prefix match range (take 2)

vekterli:
@geirst please review

* ed1b065de64874981b4cdb23c239e7e07dc50d28 is the same as #26494, but squashed
* fe50c5095c7e6616571d36270e0b2189f7ec8730 adds the necessary Valgrind suppressions for RE2's (intentional) dancing around in the flowers of uninitialized memory


geirst (approved):
👍

PR #26509, opened 2023-03-21
Disallow incremental non-exclusive container allocation

bratseth:
Disallow shared allocation of containers also when there is an existing non-exclusive host which can fit the requested node.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
@@ -215,7 +215,7 @@ class NodeAllocation {
             // If any of the nodes on the host requires exclusivity to another application, allocating on it is a violation
             for (Node nodeOnHost : allNodes.childrenOf(candidate.parentHostname().get())) {
                 if (nodeOnHost.allocation().isEmpty()) continue;
-                if (requestedNodes.isExclusive() || nodeOnHost.allocation().get().membership().cluster().isExclusive()) {
+                if (nodeOnHost.allocation().get().membership().cluster().isExclusive()) {
                     if (!nodeOnHost.allocation().get().owner().equals(application))
                         return true;
                 }
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
@@ -194,27 +194,34 @@ class NodeAllocation {
         return false;
     }
 
+    /**
+     * Returns whether allocating the candidate on this host would violate exclusivity constraints.
+     * Note that while we currently require that exclusive allocations uses the entire host,
+     * this method also handles the case where smaller exclusive nodes are allocated on it.
+     */
     private boolean violatesExclusivity(NodeCandidate candidate) {
-        if (candidate.parentHostname().isEmpty()) return false;
+        if (candidate.parent.isEmpty()) return false;
 
-        // In nodes which does not allow host sharing, exclusivity is violated if...
-        if ( ! nodeRepository.zone().cloud().allowHostSharing()) {
-            // TODO: Write this in a way that is simple to read
-            // If either the parent is dedicated to a cluster type different from this cluster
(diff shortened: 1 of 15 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -194,27 +194,34 @@ private boolean offeredNodeHasParentHostnameAlreadyAccepted(NodeCandidate candid
...
+            return false;
         }
-
-        // In zones with shared hosts we require that if either of the nodes on the host requires exclusivity,
-        // then all the nodes on the host must have the same owner
-        for (Node nodeOnHost : allNodes.childrenOf(candidate.parentHostname().get())) {
-            if (nodeOnHost.allocation().isEmpty()) continue;
-            if (requestedNodes.isExclusive() || nodeOnHost.allocation().get().membership().cluster().isExclusive()) {
-                if ( ! nodeOnHost.allocation().get().owner().equals(application)) return true;
+        else {
```
freva, line 214: If we require that the node takes up the entire host, this should not be possible?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -194,27 +194,34 @@ private boolean offeredNodeHasParentHostnameAlreadyAccepted(NodeCandidate candid
...
-                    // or this cluster is requiring exclusivity, but the host is exclusive to a different owner
-                    (requestedNodes.isExclusive() && !candidate.parent.flatMap(Node::exclusiveToApplicationId).map(application::equals).orElse(false));
+        if (candidate.parent.isEmpty()) return false;
+
+        if (nodeRepository.exclusiveAllocation(cluster)) {
+            // Node must allocate the host entirely and not violate application or cluster type constraints
+            var parent = candidate.parent.get();
+            if (!candidate.resources().isUnspecified() &&
+                ! nodeRepository.resourcesCalculator().advertisedResourcesOf(parent.flavor()).compatibleWith(candidate.resources())) return true;
+            if (parent.exclusiveToApplicationId().isPresent() && !parent.exclusiveToApplicationId().get().equals(application)) return true;
```
freva, line 210: If we require exclusivity, our host should have exclusiveToApplicationId set, so this should be 
```suggestion
            if (parent.exclusiveToApplicationId().isEmpty() || !parent.exclusiveToApplicationId().get().equals(application)) return true;
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
bratseth: Yes, I've tried to explain that in the javadoc. It doesn't hurt that this method is a bit more generic than strictly necessary with today's policy, and anyway we might have this at the moment this is deployed.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
bratseth: That we need this PR at all is proof that we are offered nodes on hosts which are not exclusive. Perhaps that should be fixed elsewhere.

With the "fills completely" policy there is no need for exclusiveToApplicationId, though, so perhaps we should remove that property (although not in _this_ PR). In any case, I think this method should enforce the actual constraints regardless of what might obtain at some point in time given the rest of the code (and live pre-existing state). And I think it does with these three constraints?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
bratseth: Well, then it will always be false so we can just remove the first condition.
(We can also remove the second, after a while, if we are sure we want to keep this policy).

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
bratseth: Is there more to discuss here?

freva (approved):

PR #26511, opened 2023-03-21
Implement overloads of native types

baldersheim:
@havardpe PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
+++ b/vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
@@ -10,9 +10,11 @@ final class ArrayValue extends Value {
     static final Impl initial_impl = new EmptyImpl();
 
     private interface Impl {
-        public void prepareFor(ArrayValue self, Type type);
-        public Value add(Value value, int used);
-        public Value get(int index);
+        void prepareFor(ArrayValue self, Type type);
+        Value add(Value value, int used);
+        Value add(long value, int used);
+        Value add(double value, int used);
+        Value get(int index);
     }
 
     private static final class EmptyImpl implements Impl {
@@ -26,6 +28,8 @@ final class ArrayValue extends Value {
             }
         }
         public Value add(Value value, int used) { return NixValue.invalid(); }
+        public Value add(long value, int used) { return NixValue.invalid(); }
+        public Value add(double value, int used) { return NixValue.invalid(); }
         public Value get(int index) { return NixValue.invalid(); }
     }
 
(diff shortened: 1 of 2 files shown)


vespajlib/src/main/java/com/yahoo/slime/Value.java
```diff
@@ -48,8 +48,8 @@ public void traverse(ObjectTraverser ot) {}
 
     public final Cursor addNix() { return addLeaf(NixValue.instance()); }
     public final Cursor addBool(boolean bit) { return addLeaf(BoolValue.instance(bit)); }
-    public final Cursor addLong(long l) { return addLeaf(new LongValue(l)); }
-    public final Cursor addDouble(double d) { return addLeaf(new DoubleValue(d)); }
+    public Cursor addLong(long l) { return addLeaf(new LongValue(l)); }
```
havardpe, line 51: these should return NixValue.invalid() the same way addArray/addObject does

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
```diff
@@ -10,9 +10,11 @@ final class ArrayValue extends Value {
     static final Impl initial_impl = new EmptyImpl();
 
     private interface Impl {
-        public void prepareFor(ArrayValue self, Type type);
-        public Value add(Value value, int used);
-        public Value get(int index);
+        void prepareFor(ArrayValue self, Type type);
```
havardpe, line 13: as long as the implementations are public, these should be public as well

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
```diff
@@ -36,15 +40,22 @@ public void prepareFor(ArrayValue self, Type type) {
                 self.impl = new GenericImpl(this, self.used);
             }
         }
+        private static long [] grow(long [] arr) {
+            long [] v = new long[arr.length << 1];
+            System.arraycopy(arr, 0, v, 0, arr.length);
+            return v;
+        }
         public Value add(Value value, int used) {
+            return add(value.asLong(), used);
```
havardpe, line 49: this could also return NixValue.invalid as it will never be called

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
```diff
@@ -55,15 +66,22 @@ public void prepareFor(ArrayValue self, Type type) {
                 self.impl = new GenericImpl(this, self.used);
             }
         }
+        private static double [] grow(double [] arr) {
+            double [] v = new double[arr.length << 1];
+            System.arraycopy(arr, 0, v, 0, arr.length);
+            return v;
+        }
         public Value add(Value value, int used) {
+            return add(value.asDouble(), used);
```
havardpe, line 75: this could also return NixValue.invalid as it will never be called

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
```diff
@@ -36,15 +40,22 @@ public void prepareFor(ArrayValue self, Type type) {
                 self.impl = new GenericImpl(this, self.used);
             }
         }
+        private static long [] grow(long [] arr) {
```
havardpe, line 43: I think the type should not contain space: long[] (similar for other grow functions)

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
baldersheim: Interface are always public.

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
baldersheim: Fixed

vespajlib/src/main/java/com/yahoo/slime/ArrayValue.java
baldersheim: Fixed

PR #26536, opened 2023-03-22
Wait less before redeploying second time

freva:
With our custom AMIs we are able to provision an instance in less than 4 minutes, so controller should wait less before trying to redeploy again.


jonmv (approved):
Nice!

PR #26539, opened 2023-03-22
It is enough to have a single document to verify that the corpus is OK.

### LINKED ISSUE #26490: Streaming queries always return degraded true if you only have a single document in your index....
**Describe the bug**
When using Vespa with [mode streaming](https://docs.vespa.ai/en/streaming-search.html), I always get a coverage degraded in the response, with `"non-ideal-state": true`. It also seems like metadata like nodes, results, and resultsFull is incomplete (0). 

```
curl "$endpoint/search/?yql=select%20*%20from%20sources%20*%20where%20text%20contains%20%22april%22&streaming.userid=1&" -q -s |python3 -m json.tool
{
    "root": {
        "id": "toplevel",
        "relevance": 1.0,
        "fields": {
            "totalCount": 1
        },
        "coverage": {
            "coverage": 0,
            "documents": 1,
            "degraded": {
                "match-phase": false,
                "timeout": false,
                "adaptive-timeout": false,
                "non-ideal-state": true
            },
            "full": true,
            "nodes": 0,
            "results": 0,
            "resultsFull": 0
        },
        "children": [
            {
                "id": "id:doc:doc:n=1:1",
                "relevance": 0.22853769543241514,
                "source": "text.doc",
                "fields": {
                    "sddocname": "doc",
                    "documentid": "id:doc:doc:n=1:1",
                    "subject": "Instant Rebates on Leupold BX-4 Pro Guide & SX-4 Pro Guide Spotters!",
                    "text": "Through April 2nd, 2023, save $100 on select Leupold BX-4 Pro Guide HD Binoculars or save $200 on SX-4 Pro Guide HD Spotters!",
                    "embedding_vector": [
                        1.0,
                        2.0,
                        3.0,
                        4.0
                    ]
                }
            }
        ]
    }
}
```

baldersheim:
@vekterli or @bratseth PR
Fixes #26490

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/result/Coverage.java
+++ b/container-search/src/main/java/com/yahoo/search/result/Coverage.java
@@ -10,7 +10,7 @@ package com.yahoo.search.result;
 public class Coverage extends com.yahoo.container.handler.Coverage {
 
     public Coverage(long docs, long active) {
-        this(docs, active, docs > 1 ? 1 : 0, docs > 1 ? 1: 0);
+        this(docs, active, (docs > 0) ? 1 : 0, (docs > 0) ? 1: 0);
     }
 
     public Coverage(long docs, long active, int nodes) {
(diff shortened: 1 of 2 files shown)


vekterli (dismissed):
:+1: do we have any tests for the zero doc count edge case?

PR #26546, opened 2023-03-23
Mock auth clients in tests

mpolden:
Tests were making real requests to `/auth0/v1/device-flow-config`... And Auth0
client wasn't reusing connections since it used its own client instead of the
default CLI one.

@jonmv


jonmv (approved):
Hehe, nice ...

PR #26547, opened 2023-03-23
improve error for cycle in ranking expression functions

arnej27959:
@bratseth please review
@bjormel FYI

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/derived/DerivedConfiguration.java
+++ b/config-model/src/main/java/com/yahoo/schema/derived/DerivedConfiguration.java
@@ -88,8 +88,8 @@ public class DerivedConfiguration implements AttributesConfig.Producer {
             }
             Validation.validate(this, schema);
         }
-        catch (IllegalArgumentException e) {
-            throw new IllegalArgumentException("Invalid " + schema, e);
+        catch (IllegalArgumentException|IllegalStateException e) {
+            throw new IllegalArgumentException("Invalid " + schema + " -> " + e.getMessage(), e);
         }
     }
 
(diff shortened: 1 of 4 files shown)


searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/rule/ReferenceNode.java
```diff
@@ -79,7 +79,7 @@ public StringBuilder toString(StringBuilder string, SerializationContext context
                 path = new ArrayDeque<>();
             String myPath = getName() + getArguments().expressions();
             if (path.contains(myPath))
-                throw new IllegalStateException("Cycle in ranking expression function: " + path);
+                throw new IllegalStateException("Cycle in ranking expression function '" + getName() + "' called from: " + path);
```
baldersheim, line 82: Should this be IllegalArgumentException too, it is error in the input ?

bratseth: Yes.

config-model/src/main/java/com/yahoo/schema/derived/DerivedConfiguration.java
```diff
@@ -88,8 +88,8 @@ public DerivedConfiguration(Schema schema, DeployState deployState) {
             }
             Validation.validate(this, schema);
         }
-        catch (IllegalArgumentException e) {
-            throw new IllegalArgumentException("Invalid " + schema, e);
+        catch (IllegalArgumentException|IllegalStateException e) {
+            throw new IllegalArgumentException("Invalid " + schema + " -> " + e.getMessage(), e);
```
bratseth, line 92:
```suggestion
            throw new IllegalArgumentException("Invalid " + schema, e);
```

bratseth (commented):
Otherwise LGTM

PR #26556, opened 2023-03-23
Implement RFC 9180 HPKE sender asymmetric key authentication mode

vekterli:
@bjorncs please review. Might come in handy! See [RFC 9180—Authentication Using an Asymmetric Key](https://www.rfc-editor.org/rfc/rfc9180.html#name-authentication-using-an-asy) for mode details.

We already have support for the `base` unauthenticated mode, so this just adds the `auth` mode where the sender's key pair is added to the ECDH shared key derivation mix. This ensures that a message may only be successfully opened if the sender was in possession of the private key (`skS`) corresponding to the expected public key (`pkS`).


bjorncs (approved):
Cool! 😎 👍

PR #26560, opened 2023-03-23
Revert "Cleanup: Remove RemoveApplicationWaiter, use CuratorCompletionWaiter …"

aressem:
Reverts vespa-engine/vespa#26525

Suspect that this broke a lot of system/performance tests.


hmusum:
Thanks!

PR #26565, opened 2023-03-24
Dispatch documents with common ID in order

mpolden:
@jonmv

### DIFF
--- a/client/go/internal/vespa/document/dispatcher.go
+++ b/client/go/internal/vespa/document/dispatcher.go
@@ -5,59 +5,125 @@ import (
 	"sync"
 )
 
+const maxAttempts = 10
+
 // Dispatcher dispatches documents from a queue to a Feeder.
 type Dispatcher struct {
-	concurrencyLevel int
-	feeder           Feeder
-	pending          chan Document
-	closed           bool
-	mu               sync.RWMutex
-	wg               sync.WaitGroup
+	workers  int
+	feeder   Feeder
+	ready    chan Id
+	inflight map[string]*documentGroup
+	mu       sync.RWMutex
+	wg       sync.WaitGroup
+	closed   bool
 }
 
-func NewDispatcher(feeder Feeder, concurrencyLevel int) *Dispatcher {
-	if concurrencyLevel < 1 {
-		concurrencyLevel = 1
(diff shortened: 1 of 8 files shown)


client/go/internal/vespa/document/dispatcher.go
```diff
@@ -5,59 +5,125 @@ import (
...
-	d.readPending()
+	d.start()
 	return d
 }
 
-func (d *Dispatcher) readPending() {
-	for i := 0; i < d.concurrencyLevel; i++ {
+func (d *Dispatcher) dispatchAll(g *documentGroup) int {
+	g.mu.Lock()
+	defer g.mu.Unlock()
```
jonmv, line 55: Hmm, won't this block the client thread if it tries to add another document with this ID, until all operations for that ID have completed? I guess it's not likely to be a problem ... but is the locking necessary?

client/go/internal/vespa/document/dispatcher.go
```diff
@@ -5,59 +5,125 @@ import (
...
+	failCount := len(g.operations)
+	for i := 0; !g.failed && i < len(g.operations); i++ {
```
jonmv, line 57: I'm not sure what we'd like to do when a single operation against a particular document ID fails. In the Java feed client, we just keep going.

jonmv (approved):
It'll be interesting to see the results of a performance test, but I guess that's quite a while away? :)

client/go/internal/vespa/document/dispatcher.go
```diff
@@ -5,59 +5,125 @@ import (
...
+	for i := 0; i < d.workers; i++ {
 		d.wg.Add(1)
-		go func(n int) {
+		go func() {
 			defer d.wg.Done()
-			for doc := range d.pending {
-				d.feeder.Send(doc)
+			for id := range d.ready {
+				d.mu.RLock()
+				group := d.inflight[id.String()]
```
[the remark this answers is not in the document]
mpolden: Not currently. Can look into that if memory use becomes a problem with really large document batches.

client/go/internal/vespa/document/dispatcher.go
mpolden: I'm unsure what is right as well. I can change it to match vespa-feed-client behavior in later PR.

PR #26571, opened 2023-03-24
Retrieve execution value explicitly by '_'

bratseth:
With multiple embeddings per document it might be useful to be able to prepend each string to be embedded by some context, e.g the title. Currently there is no way to do that (as far as I know) because it requires referring to the execution value (passed from the previous expression in the indexing statement) explicitly.

This adds a new expression which is simply '_' which returns the current context value, such that one can say e.g

`input myTextArray | for_each { input title . " " . _ } | embed | attribute 'mySparseTensor' `

I think this is needed at the top level in any case, but for nested expressions it would also be nice to be able to name the expression value explicitly to be able to say

`input myTextArray | for_each text { input title . " " . text } | embed | attribute 'mySparseTensor' `

but I'm not doing that right now.


geirst (approved):
👍

PR #26590, opened 2023-03-27
Support bucketSpace doc/v1 argument in vespa cli

kkraune:
... or @mpolden

### DIFF
--- a/client/go/internal/cli/cmd/visit.go
+++ b/client/go/internal/cli/cmd/visit.go
@@ -34,6 +34,7 @@ type visitArgs struct {
 	slices         int
 	sliceId        int
 	bucketSpace    string
+	bucketSpaces   []string
 	cli            *CLI
 }
 
@@ -133,7 +134,7 @@ $ vespa visit --field-set "[id]" # list document IDs
 	cmd.Flags().StringVar(&vArgs.to, "to", "", `Timestamp to visit up to, in seconds`)
 	cmd.Flags().IntVar(&vArgs.sliceId, "slice-id", -1, `The number of the slice this visit invocation should fetch`)
 	cmd.Flags().IntVar(&vArgs.slices, "slices", -1, `Split the document corpus into this number of independent slices`)
-	cmd.Flags().StringVar(&vArgs.bucketSpace, "bucket-space", "default", `"default" or "global" bucket space`)
+	cmd.Flags().StringSliceVar(&vArgs.bucketSpaces, "bucket-space", []string{"global", "default"}, `"default" or "global" bucket space`)
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


arnej27959 (dismissed):
this looks good, but even better would be if it visited both as default

kkraune:
I don't know why the default is not including global buckets - @vekterli ? We can consider making this change in the vespa CLI

vekterli:
> I don't know why the default is not including global buckets - @vekterli ? We can consider making this change in the vespa CLI

Global buckets (and thus global documents) exist in their own logically separate _bucket space_, which is akin to existing in a separate database table. Including global buckets by default would require a full iteration over this bucket space in addition to the default space. Certainly possible to do, but will inherently have extra overhead.

kkraune:
Thanks. The default is to iterate over all schemas already. We have had a few questions about this, but maybe easiest to keep the current default ...

vekterli:
Iterating over all schemes (within—not across—global and non-global document types) is the default because it is effectively "free" due to document colocation within a particular bucket space. I.e. you only pay the cost of iteration across the bucket space once. This is an artefact of the content layer originally having one singular, shared data space. 

Now with separate bucket spaces, it's not just a matter of sending a visitor operation to the backend saying "visit both spaces instead of just the one"—they are inherently in their own little realms of existence (separate bucket sets, bucket databases, maintenance management etc). Consequently, transparently supporting multi-space iteration would need to add a new indirection that also tracks which bucket space is being currently iterated over. Or the progress into each individual space, if parallell iteration is to be supported.

client/go/internal/cli/cmd/visit.go
```diff
@@ -133,7 +134,7 @@ $ vespa visit --field-set "[id]" # list document IDs
 	cmd.Flags().StringVar(&vArgs.to, "to", "", `Timestamp to visit up to, in seconds`)
 	cmd.Flags().IntVar(&vArgs.sliceId, "slice-id", -1, `The number of the slice this visit invocation should fetch`)
 	cmd.Flags().IntVar(&vArgs.slices, "slices", -1, `Split the document corpus into this number of independent slices`)
-	cmd.Flags().StringVar(&vArgs.bucketSpace, "bucket-space", "default", `"default" or "global" bucket space`)
+	cmd.Flags().StringSliceVar(&vArgs.bucketSpaces, "bucket-space", []string{"global", "default"}, `"default" or "global" bucket space`)
```
kkraune, line 137: ... or "default,global" ?

client/go/internal/cli/cmd/visit_test.go
```diff
@@ -105,6 +105,7 @@ func TestVisitCommand(t *testing.T) {
 	assertVisitResults(
 		[]string{
 			"visit",
+			"--bucket-space", "default",
```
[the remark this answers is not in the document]
arnej27959: the test only expects one round of visiting, so this change would need more work

client/go/internal/cli/cmd/visit.go
arnej27959: my thinking was that "global" should be visited first.

kkraune: ok. I am just confused, there are three options "global", "default" and "global,default", right? (and the last one is default). I think the help text should reflect that

kkraune:
I think this is good to go, @arnej27959 ?

arnej27959 (approved):
👍

PR #26595, opened 2023-03-27
Ensure proper memory visibility on distributor stripe flush

vekterli:
@geirst please review
@havardpe @toregge FYI

Flushing the distributor stripes happens as part of process shutdown and therefore takes place in the main thread. At this point we should no longer receive any messages from the RPC/messaging subsystems, but there may still be lingering replies to be purged. At this point, the stripe threads are joined but the communication manager thread is not (happens slightly after in the shutdown sequence).

We therefore have to explicitly form a mutex acquire/release pair to ensure proper memory visibility between anything touched by the comm. manager thread (within the mutex) and our main thread.

This should resolve a ThreadSanitizer error.


geirst (approved):
👍

baldersheim (approved):

PR #26603, opened 2023-03-28
- Count number of items in the tree that has a unique id.

baldersheim:
- Use this to either exit early, or presize the array to collect them in.

@jonmv or @bratseth PR

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/query/CompositeItem.java
+++ b/container-search/src/main/java/com/yahoo/prelude/query/CompositeItem.java
@@ -239,22 +239,20 @@ public abstract class CompositeItem extends Item {
     }
 
     private static int countUniqueId(Item tree) {
-        if (tree.hasUniqueID()) {
-            return 1;
-        } else if (tree instanceof CompositeItem asComposite) {
-            int sum = 0;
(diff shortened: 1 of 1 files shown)


container-search/src/main/java/com/yahoo/prelude/query/CompositeItem.java
```diff
@@ -221,33 +221,47 @@ public CompositeItem clone() {
...
+            int connectedTo = find(orig.connectedItem, flat);
             if (connectedTo >= 0) {
                 TaggableItem tagged = (TaggableItem) flatCopy.get(i);
                 tagged.setConnectivity(flatCopy.get(connectedTo), orig.connectivity);
             }
         }
     }
 
-    private void taggingFlatten(Item tree, List<Item> container) {
+    private static int countUniqueId(Item tree) {
```
bratseth, line 241: This assumes that no item that has an unique id can have any children that has an unique id. Is that guaranteed by anything? I couldn't find it.

baldersheim: Good point. I  assumed that the logic in taggingFlatten was correct.
If not this is an issue in current code too.
I will see what happens if I change it.

bratseth (approved):

PR #26611, opened 2023-03-28
Return ID when adding line item

oyving:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/billing/BillingApiHandler.java
```diff
@@ -225,13 +225,18 @@ private HttpResponse addLineItem(HttpRequest request, String tenant, String user
...
                 TenantName.from(tenant),
                 getInspectorFieldOrThrow(inspector, "description"),
                 new BigDecimal(getInspectorFieldOrThrow(inspector, "amount")),
                 billId,
                 userId);
 
+        var message = new Slime();
+        var root = message.setObject();
+        root.setString("message", "Added line item for tenant " + tenant);
+        root.setString("id", lineItemId);
```
olaaun, line 238: Unused?

PR #26612, opened 2023-03-28
make nicer

kkraune:
- imho this could be removed - I don't think this is a good place to lecture the user
- users deploying to prod for the first time using the guide will see this, it should not warn, it is normal to do this


mpolden:
Just remove it instead. It's not a debug message.

kkraune:
@mpolden PTAL

mpolden (approved):

PR #26614, opened 2023-03-28
Deprecate distributor_auto_ownership_transfer_on_whole_group_down

hmusum:
Stop using the config value (same as currrent default value, which is true). Also: distribution config does not have this field, only stor-distribution config has.


vekterli (approved):
:+1:

PR #26621, opened 2023-03-28
Balder/add general compound name cache

baldersheim:
@bratseth PR


container-core/src/main/java/com/yahoo/processing/request/CompoundName.java
```diff
@@ -298,7 +302,21 @@ private static String toCompoundString(String [] compounds) {
         return b.length()==0 ? "" : b.substring(0, b.length()-1);
     }
 
-    public static CompoundName from(String name) { return new CompoundName(name); }
+    /** @deprecated Use of() instead */
```
bratseth, line 305: why? "of" is for when you lift some information from an object and return it as another type, while "from" is when you convert the entire object imo.

PR #26624, opened 2023-03-29
Reduce log level, no connection is normal in some cases

kkraune:


baldersheim:
When is this normal ? It has not been able to get contact for quite some time.
I think that there should be some traces of that.
An improvement would be to have it edge triggered. First time, and then silence util good again.

havardpe (approved):

havardpe:
the config agent also logs about this, with a more informative message

baldersheim:
K, then we are good.

PR #26637, opened 2023-03-29
maintainer success factor baseline deviation

bjormel:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/BcpGroupUpdater.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/BcpGroupUpdater.java
@@ -42,11 +42,13 @@ public class BcpGroupUpdater extends ControllerMaintainer {
 
     private final ApplicationController applications;
     private final NodeRepository nodeRepository;
+    private final Double successFactorBaseline;
 
     public BcpGroupUpdater(Controller controller, Duration duration, Double successFactorBaseline) {
         super(controller, duration, successFactorBaseline);
         this.applications = controller.applications();
         this.nodeRepository = controller.serviceRegistry().configServer().nodeRepository();
+        this.successFactorBaseline = successFactorBaseline;
     }
 
     @Override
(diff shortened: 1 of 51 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/BcpGroupUpdater.java
```diff
@@ -75,12 +75,12 @@ protected double maintain() {
...
             }
         }
-        double successFactor = asSuccessFactor(attempts, failures);
-        if ( successFactor == 0 )
+        double successFactorDeviation = asSuccessFactorDeviation(attempts, failures);
+        if ( successFactorDeviation == 1.0 )
             log.log(Level.WARNING, "Could not update traffic share on any applications", lastException);
-        else if ( successFactor < 0.9 )
+        else if ( successFactorDeviation > 0.1 )
             log.log(Level.FINE, "Could not update traffic share on all applications", lastException);
```
olaaun, line 82: `successFactorDeviation` is in the range [-baseline, 1 - baseline]

olaaun (changes requested):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/BcpGroupUpdater.java
bjormel: @olaaun , question is if I also should set deviation to 0, when there has been no attempts, or if it makes sense to distinguish this case?
```
    protected final double asSuccessFactorDeviation(int attempts, int failures) {
        double factor = attempts == 0 ? 1.0 : 1 - (double)failures / attempts;
        return new BigDecimal(factor - successFactorBaseline).setScale(2, RoundingMode.HALF_UP).doubleValue();
    }
```

olaaun: This still applies. Baseline is set to .7, so deviation can never be -1

bjormel: :(
Waiting for a brain reboot

olaaun (approved):

PR #26642, opened 2023-03-30
Balder/more statically generated compounds in cache 2

baldersheim:
@bratseth PR


bratseth (approved):
Perhaps we should have fromComponents construct a string and try the cache lookup.

baldersheim:
Yes, but will take this in small steps, by first cleaning up our own usage. and then evaluate what functionality we should keep and what we should drop. Currently there are too many ways to Rome.
There are more in the pipeline.

PR #26643, opened 2023-03-30
Use DomainName instead of the stricter HostName in duper-model

jonmv:
@andreer


andreer (dismissed):
LGTM but 
```
00:11:23 [ERROR] /sd/workspace/src/github.com/vespa-engine/vespa/node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/RetiredExpirerTest.java:[271,22] incompatible types: invalid method reference
00:11:23 [ERROR]     method value in class ai.vespa.validation.StringWrapper<T> cannot be applied to given types
00:11:23 [ERROR]       required: no arguments
00:11:23 [ERROR]       found:    ai.vespa.http.DomainName
00:11:23 [ERROR]       reason: actual and formal argument lists differ in length```

PR #26650, opened 2023-03-30
Arnej/cleanup identifier with dash

arnej27959:
@bratseth please review carefully; I am guessing on the intent here.
@bjorncs FYI


bratseth (approved):
👍 👍

PR #26660, opened 2023-03-31
Ignore input also listed in initializers

bjorncs:
Ignore inputs which is already initialized when determining required features as input to ONNX models in global-phase.


bratseth (approved):

arnej27959 (commented):
great!

PR #26669, opened 2023-04-03
Improve error message

kkraune:
- We need to improve the error message for full disk or memory.
- The information is already available in the message, but hard for users to translate into what it means: (0.767 > 0.750)
- Ideally, it should say something like "Resource _disk_ on node 0 is at _76.7%_, which is over the feed-block limit at _75%_, refer to https://docs.vespa.ai/en/operations/feed-block.html "

If we are OK with this, I can update code and tests (the change below is a dummy for now) - OK @geirst ?


geirst (commented):
👍

kkraune:
lets merge this for now

PR #26671, opened 2023-04-03
Wire in TensorExtAttribute.

toregge:
@vekterli : please review


vekterli (approved):
:+1: @geirst also should probably take a look when back

geirst (commented):
👍

PR #26674, opened 2023-04-03
Use return value from std::unique.

toregge:
@vekterli : please review

### DIFF
--- a/vdslib/src/tests/distribution/distributiontest.cpp
+++ b/vdslib/src/tests/distribution/distributiontest.cpp
@@ -17,8 +17,9 @@
 #include <vespa/vespalib/util/size_literals.h>
 #include <gmock/gmock.h>
 #include <chrono>
-#include <thread>
 #include <fstream>
+#include <iterator>
+#include <thread>
 
 using namespace ::testing;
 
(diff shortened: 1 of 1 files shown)


vdslib/src/tests/distribution/distributiontest.cpp
```diff
@@ -434,7 +434,8 @@ TEST(DistributionTest, test_distribution)
                 _distribution[i].second = distr.getIdealStorageNodes(
                         systemState, document::BucketId(26, i));
                 sort(_distribution[i].second.begin(), _distribution[i].second.end());
-                unique(_distribution[i].second.begin(), _distribution[i].second.end());
+                auto unique_nodes = unique(_distribution[i].second.begin(), _distribution[i].second.end()) - _distribution[i].second.begin();
```
vekterli, line 437: This might benefit from using `std::distance` instead, to make the semantics more clear

vekterli (dismissed):
:+1:

vdslib/src/tests/distribution/distributiontest.cpp
toregge: Done.

vekterli (approved):
:+1:

PR #26675, opened 2023-04-03
Read local config when in an application package sub-directory

mpolden:
A few fixes for bugs I found related to local config.

@jonmv

### DIFF
--- a/client/go/internal/cli/cmd/config.go
+++ b/client/go/internal/cli/cmd/config.go
@@ -544,12 +544,12 @@ func (c *Config) list(includeUnset bool) []string {
 }
 
 // flagValue returns the set value and default value of the named flag.
-func (c *Config) flagValue(name string) (string, string) {
+func (c *Config) flagValue(name string) (string, string, bool) {
 	f, ok := c.flags[name]
 	if !ok {
-		return "", ""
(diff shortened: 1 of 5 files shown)


bratseth (approved):

client/go/internal/cli/cmd/config.go
```diff
@@ -307,18 +336,30 @@ func athenzPath(filename string) (string, error) {
 	return filepath.Join(userHome, ".athenz", filename), nil
 }
 
-func (c *Config) loadLocalConfigFrom(parent string) error {
-	home := filepath.Join(parent, ".vespa")
-	_, err := os.Stat(home)
-	if err != nil && !os.IsNotExist(err) {
-		return err
+func (c *Config) findLocalConfigDir(dir string) (string, error) {
+	userHome, err := os.UserHomeDir()
```
jonmv, line 340: This broke the use of Vespa CLI to run automated system and staging tests. I guess it's simple to fix that, though.

jonmv: @mpolden should the CLI handle no `$HOME` on UNIX, or is that a user error that shouldn't really happen?

PR #26678, opened 2023-04-04
Switch from elements to entries for raw alloator alloc parameter.

toregge:
@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1: I'll casually defer to @geirst for more in-depth reviewing

geirst (commented):
👍

PR #26698, opened 2023-04-11
Set user.home as HOME for Vespa CLI tests

jonmv:
@hmusum or @mpolden


hmusum (approved):

jonmv:
Perhaps the CLI should instead handle no `$HOME` variable?

jonmv:
https://github.com/vespa-engine/vespa/pull/26699

PR #26708, opened 2023-04-11
Allow start end sequence tokens as args bertbaseembedder

xgboosting:
This PR is to allow passing seperator and cls token ids as arguments for loading models that don't use the default bert base uncased vocab.


lesters (approved):
👍

PR #26712, opened 2023-04-12
When a canary fails after other apps upgrade that is not broken confi…

jonmv:
…dence


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/versions/VersionStatusTest.java
```diff
@@ -349,6 +349,13 @@ public void testVersionConfidence() {
         assertEquals(Confidence.high, confidence(tester.controller(), version0), "Confidence remains unchanged for version0: High");
         assertEquals(VespaVersion.Confidence.high, confidence(tester.controller(), version2), "90% of defaults deployed successfully: High");
 
+        // Canary failing a new revision does not affect confidence
```
jonmv, line 352: This was always true

freva (approved):

PR #26725, opened 2023-04-13
Jonmv/improve job status insight

jonmv:
@freva


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/deployment/DeploymentStatus.java
```diff
@@ -1181,28 +1202,69 @@ public Versions versions() {
...
+                          }
+                        : "not ready" + switch (cause) {
+                              case unverified -> ": waiting for verification test to complete";
+                              case notReady -> ": waiting for dependencies to complete";
+                              default -> throw new IllegalStateException(cause + " should have an instant at which it is ready");
+                          };
+        }
+    }
+
+    static <T extends Comparable<T>> T min(T a, T b) {
```
freva, line 1262: FYI we have this in `vespajlib`: https://github.com/vespa-engine/vespa/blob/7e1b7baba3f2f723405985d636089650a521f5d7/vespajlib/src/main/java/com/yahoo/collections/Comparables.java

freva (approved):

PR #26726, opened 2023-04-13
andreer/relax cert logging

andreer:


bratseth:
Will we log something sensible somewhere if it never becomes available?

andreer:
After 20 min we'll fail the deployment with 
```if (startTime.plus(timeouts.endpointCertificate()).isBefore(controller.clock().instant())) {
        logger.log(WARNING, "CA signed certificate for app not available within " +
                   timeouts.endpointCertificate() + ": " + Exceptions.toMessageString(e));
        return Optional.of(RunStatus.endpointCertificateTimeout);```
and the failed deployment triggers an alert to on-call.

PR #26739, opened 2023-04-14
Automatically approve CMR impacting proxy node

olaaun:
Also adds status field to report, and check to avoid rewriting manually removed reports

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VcmrReport.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VcmrReport.java
@@ -11,7 +11,6 @@ import java.time.ZonedDateTime;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 
 import static com.yahoo.yolean.Exceptions.uncheck;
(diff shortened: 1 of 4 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VcmrReport.java
```diff
@@ -96,15 +96,18 @@ public String toString() {
     public static class Vcmr {
```
andreer, line 96: This is exactly the kind of class that should become a record, even if IntelliJ doesn't want to suggest it. Or maybe there's some object-mapper related reason it can't be?

```java
    public record Vcmr (@JsonProperty("id") String id,
                        @JsonProperty("plannedStartTime") ZonedDateTime plannedStartTime,
                        @JsonProperty("plannedEndTime") ZonedDateTime plannedEndTime,
                        @JsonProperty("status") String status ) {}
```

andreer (dismissed):

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/vcmr/VcmrReport.java
andreer: (IntelliJ doesn't suggest it because the constructor is not public and the instance variables are not final. But they can be.)

olaaun: >Or maybe there's some object-mapper related reason it can't be?

Nope, no reason

andreer (approved):

PR #26740, opened 2023-04-14
Set reserved doc values after attribute vector has been loaded.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #26745, opened 2023-04-14
Do not fail if no capacity due to retirement

freva:
Fixes VESPA-25275


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupPreparer.java
```diff
@@ -124,6 +123,14 @@ private List<Node> prepareWithLocks(ApplicationId application, ClusterSpec clust
                     hosts.forEach(host -> nodeRepository.nodes().deprovision(host.hostname(), Agent.system, nodeRepository.clock().instant()));
                     throw e;
                 }
+            } else if (allocation.hostDeficit().isPresent() && requestedNodes.canFail() &&
```
bratseth, line 126: 😭

bratseth (approved):

PR #26749, opened 2023-04-17
Add flag for allowing more than one content group down

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -250,6 +250,7 @@ public FeatureFlags(FlagSource source, ApplicationId appId, Version version) {
             this.heapPercentage = flagValue(source, appId, version, PermanentFlags.HEAP_SIZE_PERCENTAGE);
             this.enableGlobalPhase = flagValue(source, appId, version, Flags.ENABLE_GLOBAL_PHASE);
             this.summaryDecodePolicy = flagValue(source, appId, version, Flags.SUMMARY_DECODE_POLICY);
+            this.allowMoreThanOneContentGroupDown = clusterId -> flagValueAsInt(source, appId, version, clusterId, Flags.ALLOW_MORE_THAN_ONE_CONTENT_GROUP_DOWN);
```
hakonhall, line 253: Why not make allowMoreThanOneContentGroupDown a `Predicate<ClusterSpec.Id>`, and then invoke `flagValue(...)` here?  Why go via `int` and `flagValueAsInt`?

hmusum: Yeah, why not. Fixed

hakonhall (approved):

PR #26751, opened 2023-04-17
Support TLS in custom target

mpolden:
Performance tests require mTLS, but CLI didn't support TLS for local/custom
targets. Also had to refactor a bit to better support all the various
combinations of HTTP/2 or not, HTTP or HTTPS, certificate or not, Auth0, ZTS,
API key......!

@jonmv

### DIFF
--- a/client/go/internal/vespa/document/dispatcher.go
+++ b/client/go/internal/vespa/document/dispatcher.go
@@ -184,10 +184,11 @@ func (d *Dispatcher) enqueue(op documentOp) error {
 	if !d.started {
 		return fmt.Errorf("dispatcher is closed")
 	}
-	group, ok := d.inflight[op.document.Id.String()]
+	key := op.document.Id.String()
+	group, ok := d.inflight[key]
 	if !ok {
 		group = &documentGroup{}
-		d.inflight[op.document.Id.String()] = group
+		d.inflight[key] = group
 	}
 	d.mu.Unlock()
 	group.add(op, op.attempts > 0)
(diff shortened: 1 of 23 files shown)


client/go/internal/vespa/document/dispatcher.go
```diff
@@ -84,6 +82,8 @@ func (d *Dispatcher) sendDocumentIn(group *documentGroup) {
 	op.attempts++
 	result := d.feeder.Send(op.document)
 	d.results <- result
+	d.releaseSlot()
+	group.mu.Unlock()
```
jonmv, line 86: Slots in the Java implementation are per-ID, so a retry uses the same slot as the failed operation, and makes in unnecessary to acquire a new lease for both retries and for operations which are blocked due to an inflight operation with the same ID. 
Is this change because you'd otherwise risk blocking your thread trying to acquire a slot?

jonmv (approved):
One question about slots.
Also, the TLS refactoring is hopefully tested somewhere before we release it?

client/go/internal/vespa/document/dispatcher.go
mpolden: > Slots in the Java implementation are per-ID

Right, I can look into doing that in a later PR.

>  Is this change because you'd otherwise risk blocking your thread trying to acquire a slot?

No, it's to prevent a deadlock on retry since we previously held the lock while enqueuing the retry (this attempts to take the same lock and locks are not re-entrant).

mpolden:
> Also, the TLS refactoring is hopefully tested somewhere before we release it?

Only in public CD integration tests. I've verified it manually on local, cloud and hosted targets.

PR #26758, opened 2023-04-17
Render raw values in grouping as base64.

baldersheim:
@bjorncs or @bratseth PR


bjorncs (dismissed):
Ok assuming format is changed to not use padding

bratseth:
We did already conclude this change was ok, didn't we?

baldersheim:
Yes, that was my conclusion too when trying to remember the discussion before easter.

PR #26766, opened 2023-04-18
Jonmv/revision rollbacks

jonmv:
@freva


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
```diff
@@ -2231,7 +2237,7 @@ private HttpResponse cancelDeploy(String tenantName, String applicationName, Str
                 return;
             }
 
-            ChangesToCancel cancel = ChangesToCancel.valueOf(choice.toUpperCase());
+            ChangesToCancel cancel = ChangesToCancel.valueOf(choice.replaceAll("-", "_").toUpperCase());
```
freva, line 2240: :raised_eyebrow:

freva (dismissed):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/ApplicationApiHandler.java
jonmv: We use hyphenated rest api paths, afaik, and that can't be an enum name.

PR #26767, opened 2023-04-18
Consolidate rendering of raw data by using the same representation as…

baldersheim:
… in normal hits by using RawBase64.

@bratseth and @bjorncs PR
Note the subtle change of the generic return value of RawId.getValue() implemented in the base class ValueGroupId.

This is probably breaking more than it seems, but I think is better to do this now as we assume raw has not been used yet as there is attribute support came now.

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/hitfield/RawBase64.java
+++ b/container-search/src/main/java/com/yahoo/prelude/hitfield/RawBase64.java
@@ -3,8 +3,10 @@ package com.yahoo.prelude.hitfield;
 
 import java.util.Arrays;
 import java.util.Base64;
+import java.util.Objects;
 
 /**
+ * Wraps a byte [] and renders it as base64 encoded string
  * @author baldersheim
  */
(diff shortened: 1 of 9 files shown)


bratseth (approved):

container-search/src/main/java/com/yahoo/prelude/hitfield/RawBase64.java
```diff
@@ -29,4 +34,17 @@ public String toString() {
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        RawBase64 rawBase64 = (RawBase64) o;
+        return Arrays.equals(content, rawBase64.content);
+    }
+
+    @Override
+    public int hashCode() {
+        return Arrays.hashCode(content);
+    }
```
bjorncs, line 49: Assuming not including `withoutPadding` is intentional?

bjorncs (commented):
👍

PR #26769, opened 2023-04-18
The reserved document for a multivalue attribute vector should have an

toregge:
empty vector.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #26772, opened 2023-04-18
Add support for node-agent dropping all container docs

freva:
This implements steps 4-8 of VESPA-25966


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -494,6 +517,7 @@ void doConverge(NodeAgentContext context) {
                     context.log(logger, "Waiting for image to download " + context.node().wantedDockerImage().get().asString());
                     return;
                 }
+                dropDocsIfNeeded(context, container);
```
hakonhall, line 520: Shouldn't this be after the container has been stopped?

freva: Container is stopped in `dropDocsIfNeeded()` and will then continue throwing exception until node-repo moves us to next stage. Reason for this is that we want to stop without orchestration (since we cant suspend all nodes at once anyway) and it'd create more mess trying to handle this further down

hakonhall (approved):

PR #26775, opened 2023-04-19
Balder/allow chunks larger than 1k

baldersheim:
@jonmv or @bjorncs or @vekterli PR

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
@@ -526,23 +526,18 @@ public class DocumentV1ApiHandler extends AbstractRequestHandler {
         parameters = getProperty(request, TIMEOUT, timeoutMillisParser).map(clock.instant()::plusMillis)
                                                                        .map(parameters::withDeadline)
                                                                        .orElse(parameters);
-        for (String name : names) switch (name) {
-            case CLUSTER:
-                parameters = getProperty(request, CLUSTER).map(cluster -> resolveCluster(Optional.of(cluster), clusters).name())
-                                                          .map(parameters::withRoute)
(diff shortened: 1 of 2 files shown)


jonmv (approved):

vekterli (commented):
Removing the upper bound for the amount of memory that can be buffered during visiting disables the safety of a rather large foot-gun—but at least this is a case where the caller explicitly asks for it to happen.

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -1168,9 +1148,8 @@ private void updateRemoveMetrics(Outcome outcome, double latency) {
     // ------------------------------------------------- Visits ------------------------------------------------
 
     private VisitorParameters parseGetParameters(HttpRequest request, DocumentPath path, boolean streamed) {
-        int wantedDocumentCount = Math.min(streamed ? Integer.MAX_VALUE : 1 << 10,
-                                           getProperty(request, WANTED_DOCUMENT_COUNT, integerParser)
-                                                   .orElse(streamed ? Integer.MAX_VALUE : 1));
+        int wantedDocumentCount = getProperty(request, WANTED_DOCUMENT_COUNT, integerParser)
+                .orElse(streamed ? Integer.MAX_VALUE : 1 << 10);
```
jonmv, line 1152: You've also changed the default from 1 to 1024 here. That's probably OK? But we need to update some system test if we want this change.

baldersheim: I noticed that too, will reconsider.

baldersheim:
@vekterli I removed the limit, as I wanted to see impact on performance stability.
Will see what makes sense afterwards.

bjorncs (commented):
👍

PR #26784, opened 2023-04-19
Forbid applications without a container cluster in hosted Vespa

jonmv:
@bratseth


bratseth (commented):
👍  but i think we should validate this for self-hosted too

bratseth (commented):
Rename validator too I guess ...

bratseth (dismissed):

jonmv:
Hehe yes ... :)

jonmv:
Hmm, this causes lots of other unit tests to fail, though :)

PR #26801, opened 2023-04-21
Add bucketSpace parameter to Visit command

bartek:
Previously, global documents could not be visited via the CLI as they exist in a non-default space ("global"). By adding the bucket-space flag argument, the visit CLI can now access global documents.

To use:

```
cd vespa/client/go
make install
./bin/vespa visit --selection="some_global_document" --bucket-space "global"
```


kkraune:
Thanks! This overlaps with #26590 that is in progress, we will look into completing that shortly

PR #26802, opened 2023-04-21
Do not retry SSLException for feed operations

jonmv:
@bjorncs or @bratseth please review and merge.


bratseth (approved):

bjorncs (commented):
I/O failure such as read timeout can be represented as a `SSLException` with `IOException` as cause. This is the case for `SSLSocket`. The underlying client is likely using `SSLEngine` instead, so this is probably safe.

jonmv:
Hmm... Ok, we _can_ catch specific SSLExceptions, then, if that's better.

jonmv:
This shouldn't be necessary, come to think of it, since we added a handshake when starting the client, to detect just this kind of error.

PR #26812, opened 2023-04-21
for document-summary blocks:

arnej27959:
* for existing fields, allow skipping the type declaration
* if type is same except for weighted-set flags, adjust type to match
* for this to work with inheritance, wire in inherited schema early

@bratseth please review
@jonmv please take a look, and check if this helps with your ticket


jonmv:
Yes, this solves the document gen issue. Let me just push that unit test here ...

bratseth (approved):
Great!

PR #26815, opened 2023-04-21
Start wireguard in a wireguard.scope cgroup

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/CGroupV2.java
```diff
@@ -27,12 +28,43 @@ public class CGroupV2 {
...
+        fullCommand[2] = cgroup;
+        System.arraycopy(command, 0, fullCommand, 3, command.length);
+        return fullCommand;
+    }
+
+    /**
+     * Returns the cgroup directory of the Podman container relative the cgroup file system mount point,
+     * and which appears as the root cgroup within the container.
+     */
+    public String cgroupOf(ContainerId containerId) {
```
freva, line 66: Add this as a getter to `PartialContainer`?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/CGroupV2.java
```diff
@@ -27,12 +28,43 @@ public class CGroupV2 {
...
+     *
+     * <p>WARNING: This method must be called only after vespa-cgexec has been installed.</p>
+     *
+     * @param cgroup  The cgroup to execute the command in, e.g. system.slice/wireguard.scope.
+     * @param command The command to execute in the cgroup.
+     * @see #cgroupOf(ContainerId)
+     */
+    public String[] wrapForExecutionIn(String cgroup, String... command) {
+        UnixPath path = new UnixPath(rootCgroupPath.resolve(cgroup).normalize());
+        if (!path.toString().startsWith(rootCgroupPath + "/"))
```
freva, line 49: Shouldn't `vespa-cgexec` check this?

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/CGroupV2.java
hakonhall: I'm thinking it's better to collect the cgroup hierarchy in this file.

hakonhall: Changed to `static`.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/container/CGroupV2.java
hakonhall: OK.  It will add another dependency of the script: `realpath`, which is installed on a sample host in AWS.  It would be very strange if that is not installed (to my knowledge only MacOS doesn't have it).

hakonhall: And, I added a `mkdir` if cgroup does not already exist, and also allows passing a /sys/fs/cgroup path.

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
```diff
@@ -530,7 +531,10 @@ void doConverge(NodeAgentContext context) {
                 }
 
                 aclMaintainer.ifPresent(maintainer -> maintainer.converge(context));
-                if (runOrdinaryWireguardTasks) wireguardTasks.forEach(task -> task.converge(context));
+                if (runOrdinaryWireguardTasks) {
+                    Optional<Container> finalContainer = container;
```
gjoranv, line 535: This seems redundant, as the name does not add any value(?)

gjoranv (approved):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/nodeagent/NodeAgentImpl.java
hakonhall: Unfortunately, Java requires variables referenced in lambdas to be effectively final :-(

PR #26818, opened 2023-04-21
Revert "use value density to estimate needed decode index size"

baldersheim:
Reverts vespa-engine/vespa#26813
Broke some performance tests


hmusum:
👍

PR #26844, opened 2023-04-25
We no longer need to size the shared pool based on number of document…

baldersheim:
… types.

@geirst PR


geirst (approved):
👍  @toregge FYI

PR #26851, opened 2023-04-25
Put the openai client in a separate component

bratseth:
Too many inconvenient dependencies in this openai client.


lesters (approved):
LGTM!

PR #26883, opened 2023-04-27
Reduce allocations when creating result

mpolden:
...and some other fine-tuning.

@jonmv


jonmv (approved):
Yes, that's equal to the updated client timeouts of yesterday :)

PR #26900, opened 2023-04-27
move MatchFeatureData class to vespajlib

arnej27959:
@geirst this is a simple way to solve your use case (just move the class to vespajlib)
we could consider making it more general-purpose but I think that can wait until we have some actual needs.
@gjoranv FYI


geirst (approved):
👍

PR #26901, opened 2023-04-28
Include roles from logfwarder config in identity document

tokle:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -407,7 +407,7 @@ public class Flags {
             "new_iddoc_layout", false, List.of("tokle", "bjorncs", "olaa"), "2023-04-24", "2023-05-31",
             "Whether to use new identity document layout",
             "Takes effect on node reboot",
-            HOSTNAME);
+            HOSTNAME, APPLICATION_ID);
```
olaaun, line 410: Should this also include `VESPA_VERSION`?

olaaun (approved):

PR #26909, opened 2023-04-28
Ensure process-internal message ID uniqueness

vekterli:
@geirst @havardpe please review and enjoy this commit message thesis.
@baldersheim FYI (and feel free to review as well)

When a storage API command is created internally on a node it is always assigned a strictly increasing message ID that is guaranteed to be unique within the process. Some parts of the code use this as a way to distinguish messages from another. However, uniqueness (prior to this commit) did not necessarily hold, as the underlying wire protocol would inherit message IDs _from other nodes_ and explicitly override the generated ID with this.

This had exciting consequences when the stars aligned and a remote node sent the same ID as one generated at the same time internally on the receiver node. Luckily, in practice this would only be used in a potentially ambiguous context when sanity checking shared read lock sets for the _same bucket_ in the persistence threads. Invariant checks would detect this is as an attempted duplicate lock acquisition and abort the process. This has been latent for many, many years, but we've seen it happen exactly once.

This commit introduces an explicit domain separation between the node-internal (locally unique) IDs and the ID used by the originator. The originator ID is maintained and returned over the wire to the caller when sending a response to the incoming request.

Curiously, we don't actually need this originator ID at all since the caller maintains explicit state containing the sender command. Unfortunately we can't simply remove it, since versions prior to this commit will still use whatever's on the wire. Will have to wait until we bump the storage API wire protocol version.


geirst (approved):
👍

havardpe (approved):

PR #26915, opened 2023-04-28
Bjorncs/jdisc timeout

bjorncs:
FYI @havardpe


baldersheim (approved):
Then we will see how many times timeRemaining/getTiemout is called.....

Issue #26925, opened 2023-05-01
Assertion Failure in HnswIndex<type>::shrink_lid_space


lukasmoellerch:
**Describe the bug**
We observe an assertion failure after content nodes are being restarted. The concise error message can be found below.

**To Reproduce**
I can't guarantee that this reproduces the issue and I have not yet spent time to come up with a minimal reproduction case. Approximately the prior history of the content node was the following:
1. Set up multi-node cluster (I have had this issue both with a 3 node cluster and with a 5 node cluster)
2. Create a schema with a mixed-tensor hnsw index
3. Insert a couple million documents 
4. Restart content node after failure

The search node then fails to come up again with the following error message.

```
INFO    searchnode       proton.proton.server.proton	Done initializing components
INFO    searchnode       proton.searchcorespi.index.indexmaintainer	compactLidSpace(6086062, 7572237)
WARNING searchnode       stderr	vespa-proton-bin: /builddir/build/BUILD/vespa-8.158.10/searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp:827: void search::tensor::HnswIndex<type>::shrink_lid_space(uint32_t) [with search::tensor::HnswIndexType type = search::tensor::HnswIndexType::MULTI; uint32_t = unsigned int]: Assertion `doc_id_limit >= _graph.nodes_size.load(std::memory_order_relaxed)' failed.
```

**Expected behavior**

**Environment (please complete the following information):**
 - OS: Linux
 - Infrastructure: self-hosted
 - Versions: official docker image from hub.docker.io

**Vespa version**
Include Vespa version(s) used. happens both with 8.157.14 and 8.158.10

**Additional context**

geirst:
This issue has been fixed in Vespa 8.160.3.

@lukasmoellerch Thanks for reporting!

lukasmoellerch:
Yes, we can confirm that this solved the problem - thanks a lot for the fix!

PR #26926, opened 2023-05-01
Only shrink hnsw graph as part of shrinking lid space on tensor attri…

toregge:
…bute

when using identity mapping between lid and nodeid.

@geirst : please review

This addresses the assertion failure reported in issue #26925


baldersheim (approved):

geirst (approved):
👍

PR #26933, opened 2023-05-02
Add some more details when config is not converging

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
```diff
@@ -188,28 +186,44 @@ private void restartServicesIfNeeded(ApplicationId applicationId) {
...
+        }
+        return true;
+    }
+
+    private String servicesNotConvergedFormatted(ServiceListResponse response) {
+        return response.services().stream()
+                .filter(service -> service.currentGeneration != response.wantedGeneration)
+                .map(service -> service.serviceInfo.getHostName() + ":" + service.serviceInfo.getServiceName() +
+                        " on generation " + service.currentGeneration)
+                .collect(Collectors.joining(","));
```
jonmv, line 224:
```suggestion
                .collect(Collectors.joining(", "));
```

jonmv (dismissed):
Nice!

Issue #26947, opened 2023-05-02
Raw field data seems to be arbitrarily trimmed


bartek:
**Describe the bug**

We recently upgraded to `8.157.14` and noticed a regression in the data stored within our `raw` fields. We have a plain and simple raw type defined in our schema:

```
field source_data type raw {}
```

When we PUT a new document to Vespa, we assign a base64 encoded string as the value for this field. The base64-decoded data within this field is a JSON blob, which we store as a representation of data from the "source" (original API we consumed from). The field is for internal replay/testing, mostly. We don't represent it in summaries but we do have some applications bootstrapped from the data within this raw field.

Since the version upgrade, we are seeing a case where some source_data representations are partial. I speculate, but in my (non exhaustive) tests I did see base64 strings *without* padding get serialized correctly, however those without, not so.

I also found this commit, which exists within the range where I've noticed the error: https://github.com/vespa-engine/vespa/commit/1d818e96e1d6002726f6b89967e0835116251f15

When I downgrade to `8.153.19` I no longer see the issue.

**To Reproduce**

I wrote a simple bash script which replicates the issue. Note you'll need to swap Vespa versions between runs as my script simply queries the document API.

Versions to consider:

* 8.155.19 (the first version I see the issue on)
* 8.153.19 (the newest tag where the issue is not present)

Here's the script:

```shell
#!/bin/bash

set -e

docid=test-b64-plain
[...]

PR #26949, opened 2023-05-02
Use padded base64 encoding as that seems to be the most common format.

baldersheim:
@bjorncs PR
This is required to fix issue #26947


bratseth (approved):

bjorncs (commented):
👍

PR #26951, opened 2023-05-03
Revert "Reapply "Remove access logging for container clusters that are internal""

hmusum:
Reverts vespa-engine/vespa#26917

Merge if this breaks something


hmusum:
Seems to break proxy nodes in cd

aressem (approved):

PR #26964, opened 2023-05-03
Move access log setup for non-application containers to cluster

hmusum:
Having some setup per container and some per container is confusing, want to do this before continuing work on https://github.com/vespa-engine/vespa/pull/26917

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainer.java
@@ -7,9 +7,6 @@ import com.yahoo.config.model.deploy.DeployState;
 import com.yahoo.config.model.producer.TreeConfigProducer;
 import com.yahoo.config.provision.ClusterSpec;
 import com.yahoo.vespa.model.container.Container;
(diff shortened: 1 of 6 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/LogserverContainer.java
```diff
@@ -19,11 +16,8 @@ public class LogserverContainer extends Container {
 
     public LogserverContainer(TreeConfigProducer<?> parent, DeployState deployState) {
         super(parent, "" + 0, 0, deployState);
-        if (deployState.isHosted() && deployState.getProperties().applicationId().instance().isTester()) useDynamicPorts();
-        LogserverContainerCluster cluster = (LogserverContainerCluster) parent;
-        addComponent(new AccessLogComponent(cluster, AccessLogType.jsonAccessLog,
-                                            deployState.featureFlags().logFileCompressionAlgorithm("zstd"),
-                                            Optional.of(cluster.getName()), true));
+        if (deployState.isHosted() && deployState.getProperties().applicationId().instance().isTester())
```
bjorncs, line 19: Consider using `DeployState.isHostedTenantApplication`

bjorncs (approved):

PR #26968, opened 2023-05-03
Revert "Disable automatic compression"

baldersheim:
Reverts vespa-engine/vespa#26961


baldersheim:
Performance degradation seems to have been a fluke...

PR #26972, opened 2023-05-03
Use pointer to const wordfolder in juniper.

toregge:
Use reference to const wordfolder in vsm adapter.

@geirst : please review


geirst (approved):
👍

PR #26977, opened 2023-05-04
Change compression library

mpolden:
Gives a decent boost over stdlib:

```
$ benchstat /tmp/old.txt /tmp/new.txt
goos: darwin
goarch: arm64
pkg: github.com/vespa-engine/vespa/client/go/internal/vespa/document
                        │ /tmp/old.txt │            /tmp/new.txt             │
                        │    sec/op    │   sec/op     vs base                │
ClientSendCompressed-10   13.243µ ± 1%   4.701µ ± 1%  -64.50% (p=0.000 n=10)
```

This library also provides a pure-Go implementation of zstd for when we want to
add that.

Review, but hold merge.

@jonmv

### DIFF
--- a/client/go/go.mod
+++ b/client/go/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/briandowns/spinner v1.16.0
 	github.com/fatih/color v1.10.0
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
+	github.com/klauspost/compress v1.16.5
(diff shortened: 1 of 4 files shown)


jonmv (approved):

mpolden:
Please don’t merge at will. I wanted a seperate run with only #26976. That’s why it says ”hold merge”.

geirst:
For better visibility in such cases, consider using the label "DO NOT MERGE".

PR #26993, opened 2023-05-05
Ensure file config is read before generation is updated

jonmv:
@bjorncs


jonmv:
Urhm, failed again, just after 134 runs instead of 14 ...

jonmv:
Maybe I'll just make a proper testable config source then.

PR #26994, opened 2023-05-05
Avoid race between config thread and updating file config source

jonmv:
@bjorncs or @baldersheim this is dirty, but should at least improve the current situation while I work on a proper test config source 🙈


jonmv:
The problem is that the file config source and its reload mechanism is broken, w.r.t. how it's used by the config subscription, and waiting exactly 1s for the graph, before writing the new config, makes a race with the config thread likely, as that _also_ sleeps a second between attempts at getting new config.

bjorncs (approved):

PR #26998, opened 2023-05-05
Change return type for VisitorFactory::makeVisitorEnvironment member …

toregge:
…function from unique pointer to shared pointer.

@geirst : please review


geirst (approved):
👍

PR #27006, opened 2023-05-08
Increase reserved memory based on observed usage

bratseth:


config-model/src/test/java/com/yahoo/vespa/model/search/NodeResourcesTuningTest.java
```diff
@@ -39,7 +39,7 @@ void require_that_hwinfo_memory_size_is_set() {
 
     @Test
     void reserved_memory_on_content_node_is_0_5_gb() {
```
baldersheim, line 41: Matching test method name

baldersheim (changes requested):
Update method name and you are good to go.

baldersheim (approved):

PR #27027, opened 2023-05-08
Default-enable fast path updates when document versions are consistent

vekterli:
@havardpe please review
@baldersheim FYI

This feature is always _implicitly_ enabled when 3-phase updates are enabled (which is the case by default). Flip the config default to `true` as well to be more in line with what's actually the case in production.

### DIFF
--- a/storage/src/vespa/storage/config/stor-distributormanager.def
+++ b/storage/src/vespa/storage/config/stor-distributormanager.def
@@ -219,7 +219,7 @@ use_btree_database bool default=true restart
 ## content nodes, the two-phase update path reverts back to the regular fast path.
 ## Since all replicas of the document were in sync, applying the update in-place
 ## shall be considered safe.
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/config/stor-distributormanager.def
```diff
@@ -219,7 +219,7 @@ use_btree_database bool default=true restart
 ## content nodes, the two-phase update path reverts back to the regular fast path.
 ## Since all replicas of the document were in sync, applying the update in-place
 ## shall be considered safe.
-restart_with_fast_update_path_if_all_get_timestamps_are_consistent bool default=false
+restart_with_fast_update_path_if_all_get_timestamps_are_consistent bool default=true
```
baldersheim, line 222: Add deprecated comment, I assume this is was just to introduce the feature.

baldersheim (dismissed):

havardpe (dismissed):

baldersheim (approved):

PR #27040, opened 2023-05-09
Config servers are never inactive

hakonhall:
inactive_config_server_expires_according_to_custom_timeout() passes only because *all* config servers become inactive, as the config server cluster is set up wrong (like a tenant application jdisc cluster).  This PR fixes ProvisioningTester to set it up correctly, which ought to make exactly one config server move to inactive in the test.

However https://github.com/vespa-engine/vespa/pull/23264 changed cfgs to transition directly from active to dirty without the intervening inactivate state.  I don't think cfgs can naturally end in inactive anymore, and some of the changes in https://github.com/vespa-engine/vespa/pull/16752 can then be undone.  This also means that the entire inactive_config_server_expires_according_to_custom_timeout() test is no longer relevant, and the transition retiring -->  dirty is handled by unit tests in RetiredExpirerTest.


mpolden (approved):
👍

PR #27049, opened 2023-05-09
Disallow removing the cfg lb as cfg will then replace it with a tenant lb

hakonhall:


freva (approved):

hakonhall:
@mpolden can you confirm the lb should only be created by Terraform in AWS too?

mpolden (approved):
"cfg will then replace it with a tenant lb" shouldn't we just fix that?

hakonhall:
At least in GCP the lb is managed by Terraform, and it looked like that in AWS too.  So fixing `createPublic()` would be to add code that should never be exercised.  An alternative is to throw in `createPublic()` on a cfg-like application ID.

PR #27055, opened 2023-05-09
Balder/propagate metastorereadguard

baldersheim:
@geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
@@ -46,6 +46,7 @@
 LOG_SETUP(".searchlib.attribute.attribute_blueprint_factory");
 
 using search::attribute::BasicType;
+using search::attribute::SearchContextParams;
 using search::attribute::CollectionType;
 using search::attribute::IAttributeVector;
 using search::attribute::ISearchContext;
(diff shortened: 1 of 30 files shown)


searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
```diff
@@ -720,13 +700,19 @@ class CreateBlueprintVisitor : public CreateBlueprintVisitorHelper
             NodeAsKey key(n, _scratchPad);
             setResult(std::make_unique<DirectAttributeBlueprint>(_field, _attr.getName(), _attr, *_dwa, key));
         } else {
+            attribute::SearchContextParams scParams;
+            scParams.useBitVector(_field.isFilter())
+                    .metaStoreReadGuard(getRequestContext().getMetaStoreReadGuard());
```
geirst, line 705: To avoid forgetting to propagate the meta store read guard, consider adding a helper function that creates the baseline SearchContextParams (with meta store read guard set), and use this function all places where _setResult()_ is called.

geirst (changes requested):
Some changes are needed in ImportedSearchContext.

searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
baldersheim: Done

geirst (approved):
👍

PR #27059, opened 2023-05-10
Pass transport and file distributor connection spec to SearchEnvironment

toregge:
in preparation for using RankingAssetsBuilder when handling config in streaming search.

@geirst : please review


geirst (approved):
👍

PR #27060, opened 2023-05-10
Always complete non-retriable operations

mpolden:
@kkraune


kkraune (approved):

kkraune:
maybe look into test coverage for this case, too

mpolden:
It's covered by existing tests now since this removed special-casing on status code.

PR #27065, opened 2023-05-10
Depend on application-model instead [MERGEOK]

hakonhall:
ConfigServerApplication is not from an exported package.


hakonhall:
`install -amd` passed.

freva (approved):

PR #27086, opened 2023-05-12
Impose job pauses over other delay causes for UI

jonmv:
@freva


jonmv:
If a job is paused, that's an explicit thing we always want to show. Other statuses are changed by the system, but this is only changed by humans (and enough time).

jonmv:
@ldalves

freva (approved):

PR #27089, opened 2023-05-12
wait for ping OK to majority of configserver hosts

arnej27959:
@hakonhall please review
note: do not merge before Tuesday (May 16).

### DIFF
--- a/client/go/internal/admin/envvars/env_vars.go
+++ b/client/go/internal/admin/envvars/env_vars.go
@@ -66,6 +66,7 @@ const (
 	VESPA_ONLY_IP_V6_NETWORKING            = "VESPA_ONLY_IP_V6_NETWORKING"
 	VESPA_PORT_BASE                        = "VESPA_PORT_BASE"
 	VESPA_SERVICE_NAME                     = "VESPA_SERVICE_NAME"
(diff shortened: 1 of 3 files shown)


hakonhall (commented):
We should do the same check for the tenant nodes, but perhaps it is better to extend the network connectivity checks unless it supports it already?

IPv6 complicates the situation, the cfgs may be reachable only via IPv6.

Internally we set VESPA_NETWORK_STACK to one of ipv4, ipv6, or dualStack.  It is assumed the local machine has the same network stack as the cfgs.  It is then used to ping/ping6 to self and cfgs.  Perhaps this could be used to (A) demand DNS resolves to IP addresses according to VESPA_NETWORK_STACK, and (B) ping/ping6 works accordingly.

Perhaps there could be an environment variable specifying which IP protocol versions that are supposed to work, and then check accordingly.  A special value of "none" could serve as skipping the checks.

arnej27959:
> We should do the same check for the tenant nodes

they will wait until they get config anyway; I don't see how it's better to wait here instead?

hakonhall:
We added this connectivity test on the tenant containers because the hostname was not set up correctly before the network came up, and which caused secondary failures in the remaining startup scripts.  This may no longer be a problem.

hakonhall (approved):

PR #27095, opened 2023-05-12
Add attribute access recorder for streaming search mode. Use it to

toregge:
determine which attributes to populate during a streaming search.

@geirst : please review


geirst (approved):
👍

PR #27100, opened 2023-05-12
Add cluster.cost metric

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
```diff
@@ -77,7 +78,8 @@ public double maintain() {
         updateTenantUsageMetrics(nodes);
         updateRepairTicketMetrics(nodes);
         updateAllocationMetrics(nodes);
-        updateExclusiveSwitchMetrics(nodes);
+        updateClusterMetrics(nodes);
+        updateClusterMetrics(nodes);
```
mpolden, line 82: Once should be enough?

bratseth: fixed

mpolden (approved):

PR #27102, opened 2023-05-12
Reprovision in enclave compatible clouds

freva:


config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
```diff
@@ -27,6 +27,7 @@ public Zone(ConfigserverConfig configserverConfig, CloudConfig cloudConfig) {
                   .name(CloudName.from(configserverConfig.cloud()))
                   .dynamicProvisioning(cloudConfig.dynamicProvisioning())
                   .allowHostSharing(cloudConfig.allowHostSharing())
+                  .allowEnclave(cloudConfig.dynamicProvisioning())
```
hakonhall, line 30: If so, do we even need to set it since it's given by dynamicProvisioning()?  In all clouds up to now, the allowEnclave lags dynamicProvisioning by years, so I think it is better with a whitelist of clouds.

hakonhall (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
freva: I just want to have this check in one place so it's easy to change if needed in the future. I could change this line to `configserverConfig.cloud().equals("aws") || configserverConfig.cloud().equals("gcp")` if you think that's better?

hakonhall: Yes, that would be more intuitive.

freva: In #27108

PR #27112, opened 2023-05-15
Ensure thread safety of handlA.

baldersheim:
Use a loop with short wait to speed up test.

@geirst or @havardpe PR


havardpe (commented):
consider adding a max wait time or time bomb to avoid deadlock on test failure

baldersheim:
Done

havardpe (approved):

PR #27115, opened 2023-05-15
Support replacing root disk for arm64 again

hmusum:
This works again with vespa-almalinux-8 AMIs created after 2023-05-13. @aressem FYI


aressem (approved):

aressem (approved):

mpolden (approved):
👍

PR #27118, opened 2023-05-15
Initiate connections shutdown uniform-randomly up to 10% before conn TTL expiry

jonmv:
@bjorncs


jonmv:
Hmm ... or is this too coarse, perhaps? I guess there could be multiple connections made within the same millisecond, although I'd be surprised if that was consistently true, to the degree where it was a problem ... no ?

bjorncs (approved):

PR #27125, opened 2023-05-15
To avoid an inherent race in testing flush versus prune instead test … MERGEOK

baldersheim:
…for what should certainly not be present, and for

a smaller subset of what will normally be present in the prune history.

@geirst PR


geirst (approved):
👍

PR #27127, opened 2023-05-16
Verify plan on deploy calls as well

oyving:


jonmv (approved):

jonmv:
https://github.com/vespa-engine/vespa/pull/27130 also moves this earlier for submit, to before we parse the payload.

PR #27129, opened 2023-05-16
Use longer interval for RetiredExpirer

hmusum:
When doing OS upgrades and hosts are retiring we might end up doing a lot of deployments in a short time for some zones, causing e.g full disk (especially when having few apps in the zone). Increase interval for expirer in such zones.


aressem (approved):

hmusum:
@bjormel FYI

PR #27152, opened 2023-05-18
Add cluster load metrics MERGEOK

bratseth:
or @gjoranv


metrics/src/main/java/ai/vespa/metrics/ConfigServerMetrics.java
```diff
@@ -45,8 +44,13 @@ public enum ConfigServerMetrics implements VespaMetrics {
...
     ORCHESTRATOR_LOCK_ACQUIRE("orchestrator.lock.acquire", Unit.OPERATION, "Number of attempts to acquire zookeeper lock"),
     ORCHESTRATOR_LOCK_ACQUIRED("orchestrator.lock.acquired", Unit.OPERATION, "Number of times zookeeper lock was acquired"),
-    ORCHESTRATOR_LOCK_HOLD_LATENCY("orchestrator.lock.hold-latency", Unit.SECOND, "Time zookeeper lock was held before it was released");
+    ORCHESTRATOR_LOCK_HOLD_LATENCY("orchestrator.lock.hold-latency", Unit.SECOND, "Time zookeeper lock was held before it was released"),
 
+    // Node repository metrics
+    CLUSTER_COST("cluster.cost", Unit.DOLLAR_PER_HOUR, "The cost of the nodes allocated to a certain cluster, in $/hr"),
+    CLUSTER_LOAD_IDEAL_CPU("cluster.load.ideal.cpu", Unit.FRACTION, "The ideal cpu load of a certain cluster"),
+    CLUSTER_LOAD_IDEAL_MEMORY("cluster.load.ideal.memory", Unit.FRACTION, "The ideal memory load of a certain cluster"),
+    CLUSTER_LOAD_IDEAL_DISK("cluster.load.ideal.disk", Unit.FRACTION, "The ideal disk load of a certain cluster");
```
gjoranv, line 53: None of the other node-repo metrics have been defined here, but we should probably do it in another PR.

gjoranv, line 53: Should these metrics be added to the doc? None of the other node-repo metrics have been defined here, but that could just be an oversight.

config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
```diff
@@ -131,6 +131,11 @@ private static Set<Metric> getConfigServerMetrics() {
         addMetric(metrics, ConfigServerMetrics.ZK_CONNECTIONS.last());
         addMetric(metrics, ConfigServerMetrics.ZK_OUTSTANDING_REQUESTS.last());
 
+        addMetric(metrics, ConfigServerMetrics.CLUSTER_COST.last());
+        addMetric(metrics, ConfigServerMetrics.CLUSTER_LOAD_IDEAL_CPU.last());
+        addMetric(metrics, ConfigServerMetrics.CLUSTER_LOAD_IDEAL_MEMORY.last());
+        addMetric(metrics, ConfigServerMetrics.CLUSTER_LOAD_IDEAL_DISK.last());
```
gjoranv, line 137: The node-repo metrics are not reported via metrics-proxy, so there is no need to add them here.

metrics/src/main/java/ai/vespa/metrics/Unit.java
```diff
@@ -12,6 +12,7 @@ public enum Unit {
     CONNECTION(BaseUnit.CONNECTION, "A link used for communication between a client and a server"),
     DOCUMENT(BaseUnit.DOCUMENT, "Vespa document, a collection of fields defined in a schema file"),
     DOCUMENTID(BaseUnit.DOCUMENTID, "A unique document identifier"),
+    DOLLAR_PER_HOUR(BaseUnit.MAGNITUDE, "$/hr"),
```
gjoranv, line 15: If we need documentation for this metric, please add BaseUnits 'dollar' and 'hour' and declare the unit with per_unit hour.

gjoranv (changes requested):

metrics/src/main/java/ai/vespa/metrics/ConfigServerMetrics.java
bratseth: Yes, I don't know why but don't we want all metrics defined?

metrics/src/main/java/ai/vespa/metrics/Unit.java
bratseth: Done. It would be nice with some explanation of what baseunit is for.

bratseth:
@gjoranv PTAL

metrics/src/main/java/ai/vespa/metrics/Unit.java
```diff
@@ -81,13 +81,14 @@ private enum BaseUnit {
         CONNECTION("connection"),
         DOCUMENT("document"),
         DOCUMENTID("documentid"),
+        DOLLAR("dollar"),
         FAILURE("failure"),
         FILE("file"),
         FRACTION("fraction"),
         HIT("hit"),
+        HOUR("hout"),
```
gjoranv, line 89: typo

metrics/src/main/java/ai/vespa/metrics/Unit.java
```diff
@@ -81,13 +81,14 @@ private enum BaseUnit {
         CONNECTION("connection"),
         DOCUMENT("document"),
         DOCUMENTID("documentid"),
+        DOLLAR("dollar"),
```
gjoranv, line 84: I'm wondering if "dollar" is ambiguous (AUSD, CAD), so I suggest to use 'USD' instead.

gjoranv (dismissed):
Please fix typo and consider comment about USD

config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/VespaMetricSet.java
gjoranv: These sets are only allow-lists for the metrics-proxy, so adding them is a NOP. But I guess they could be processed by the proxy some time in the future.

metrics/src/main/java/ai/vespa/metrics/Unit.java
bratseth: I think it's ok ...

metrics/src/main/java/ai/vespa/metrics/Unit.java
bratseth: done

PR #27156, opened 2023-05-19
Enable explicit indexing docproc in streaming search mode.

toregge:
@geirst : please review


geirst (dismissed):
👍  We should have a unit test in config-model that sets up document processing (indexing docproc) when using streaming mode.

geirst (approved):
👍

PR #27158, opened 2023-05-19
Revert "Add DNS TTL flag"

jonmv:
Reverts vespa-engine/vespa#27155

@freva or @mpolden


jonmv:
Actually, this doesn' tneed to be reverted, I guess.

PR #27162, opened 2023-05-22
Model memory overhead

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
```diff
@@ -27,9 +27,9 @@ public class NodeResourcesTuning implements ProtonConfig.Producer {
     private final int threadsPerSearch;
     private final double fractionOfMemoryReserved;
 
-    // "Reserve" 0.5GB of memory for other processes running on the content node (config-proxy, metrics-proxy).
-    // And 0.2GB for distributor and proton basics.
-    public static final double reservedMemoryGb = 0.7;
+    // Memory for other processes running on the node (config-proxy, metrics-proxy).
+    // Keep in sync with node-repository/ClusterModel
+    public static final double nodeMemoryOverheadGb = 0.7;
```
baldersheim, line 32: Now this is just used for content nodes.
We should also have this for application container nodes as they have the same basic footprint is the same there.
Then we could allow more than 70% of memory for heap and get better utilization, specially for larger hosts.
I think we could then push the limit closer to somewhere between 80-90% without any issue.

bratseth: Agreed, I can follow that up in a separate PR.

freva (approved):

config-model/src/main/java/com/yahoo/vespa/model/search/NodeResourcesTuning.java
bratseth: https://github.com/vespa-engine/vespa/pull/27194

PR #27163, opened 2023-05-22
Use another exception and error in response when quota is exceed

hmusum:


oyving (approved):

oyving:
I think this must be handled in `ConfigServerException` on the client side as well to be understood.

oyving (changes requested):
see previous comment

hmusum:
Yes, planned to do that in a later PR, but as you suggest it's probably better to do it now. Will look into it.

oyving (approved):

PR #27164, opened 2023-05-22
Threat 'other symbols' as letters

bratseth:
The unicode class 'other symbols' contains emojis, math symbols, etc. Treat these as letter characters to support searching for them.


arnej27959 (dismissed):
what happens with two emojis in a row - is it tokenised as one or two tokens?
consider extending unit tests with `"abc\uD83D\uDD2A\uD83D\uDE00xyz"` also.

bratseth:
Yes, they should always be separate tokens, and so must be treated as a separate class, which required some additional changes.

PR #27165, opened 2023-05-22
Always print documentCount field when visiting was started

jonmv:
@vekterli please review. Ref. https://factory.vespa.oath.cloud/testrun/62558846/test/Visiting::test_process_visit_throughput


vekterli (approved):
:+1: consistency is king

PR #27172, opened 2023-05-22
Only check suspended status for new allocations

freva:
During a regular internal redeploy we should not exclude hosts where a node is already allocated if the host is suspended in orchestrator. This check was added for new allocations to avoid having deployments timeout in case the host was suspended for longer periods (e.g. due to OS upgrade).

I believe this is the cause for why sometimes, during rollout of new Vespa version, config server would reallocate some nodes.


freva:
Ideally we would still consider suspended hosts and only prioritize them lower, but would have to propagate the suspended status a bit then

bratseth (approved):

PR #27175, opened 2023-05-22
Use a smallvector to avoid indirection in most common case, and reser…

baldersheim:
…ve some space upfront

@geirst or @havardpe PR


geirst (approved):
👍

PR #27180, opened 2023-05-23
Less lecturing of user

kkraune:
- and yes, I got this in a GitHub Action output ;-)

### DIFF
--- a/client/go/internal/cli/cmd/config.go
+++ b/client/go/internal/cli/cmd/config.go
@@ -515,7 +515,7 @@ func (c *Config) readAPIKey(cli *CLI, system vespa.System, tenantName string) ([
 		if _, err := os.Stat(c.authConfigPath()); err == nil {
 			return nil, nil // We have auth config, so we should prefer Auth0 over API key
 		}
-		cli.printWarning("Authenticating with API key. This is discouraged in non-CI environments", "Authenticate with 'vespa auth login' instead")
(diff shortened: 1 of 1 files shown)


mpolden (changes requested):
> I got this in a GitHub Action output 

We should fix that instead. This warning is only shown when `CI` is unset. It shouldn't be shown at all in CI environments.

```
$ vespa status -t cloud
Warning: Authenticating with API key. This is discouraged in non-CI environments
Hint: Authenticate with 'vespa auth login' instead
Container (query API) at xxx is ready
$ env CI=true vespa status -t cloud
Container (query API) at xxx is ready
```

`CI` should always be set according to https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables, so not sure what's up with your action.

Is there a way to show all the variables set in the action?

kkraune:
I agree we should fix the root cause, but regardless of that I do not need the "This is discouraged in non-CI environments" string :-)

I can do a run to print the environment, hang on ...

kkraune:
I think we found the reason why CI was not set in the env.

mpolden:
Yes. Instead of removing the explaination, how about rephrasing it? "Authenticating with API key, which is intended for use in CI environments". That way we still explain the purpose of an API key.

kkraune:
yes, that is better!

mpolden (approved):

PR #27182, opened 2023-05-23
Store max squared norm in file header during hnsw index save when using

toregge:
dotproduct distance metric. Adjust hnsw index load.

@geirst : please review


geirst (approved):
👍

PR #27189, opened 2023-05-23
Use getCurrentCertificateWithKey

olaaun:


tokle (approved):

bjorncs (approved):
There is a theoretic race here if invocations to `getPrivateKey()` and `getIdentityService()` happens while the underlying certificate and private key is updated.

PR #27193, opened 2023-05-23
Avoid PTR lookup in GCP

hakonhall:
Allocation to shared hosts currently fails with:

```
Failed allocating address on active host h2583.prod.gcp-us-central1-f.gcp-cd.vespa.yahoo.cloud allocated to hosted-vespa.tenant-host as 'container/tenant-host/0/8': Could not resolve IP address: 2600:1900:4000:b633:0:10:1:0 (NodeCandidate)
```

Which happens because findAllocation() tries to resolve PTR records for IPv6 addresses that are assigned to newly provisioned Public CD GCP hosts (flag).


freva (approved):

hakonhall:
This should not be needed once we get PTR records for IPv6 in GCP noclave.

PR #27204, opened 2023-05-24
Factor out conversion between strings and distance metric to

toregge:
DistanceMetricUtils.

@geirst : please review


geirst (approved):
👍

PR #27215, opened 2023-05-25
add minimal port of Slime API

arnej27959:
@havardpe please take a look


havardpe:
We ended up with a simplified data model for go (https://github.com/vespa-engine/vespa/pull/32301)

PR #27217, opened 2023-05-25
- Use a rwlock as reading surpasses writing by a very large factor.

baldersheim:
- size() does not need a lock.

@geirst or @toregge PR


toregge (approved):

geirst (approved):
👍

PR #27225, opened 2023-05-26
keep previous ordering of members for correct teardown order

arnej27959:
@baldersheim or @toregge please review and merge


toregge:
This looks to be a mostly the same as the change in #27224

PR #27235, opened 2023-05-31
Less memory for default node resources in dev cd

hmusum:


freva (approved):

bratseth:
We just got rid of 4Gb nodes to make tests work reliably with the new increased vespa tax - so shouldn't this be 8Gb?

hmusum:
This is for an on-prem zone in cd only where DeployProdAppsTest has started taking a very long time (was 20-25 minutes a week ago, now takes 50-60 minutes) due to requiring more node resources when they are not specified in services.xml (maybe this is just because some apps don't specify node resources for dev environments) . We don't have any more available HW for this zone, so need to do something about this.

bratseth:
Set 4Gb + a comment in the servioces.xml file of the app instead?

PR #27256, opened 2023-06-01
Prefer latest generation with fallback to older

mpolden:
Merge together with internal PR. Moving existing nodes to a newer generation
will be addressed in a future PR.

@bratseth


bratseth (approved):
Great!

PR #27265, opened 2023-06-01
Firu a dummy query at end of search handler constructor

jonmv:
@bratseth is this too naïve, or just perfect? :>

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/handler/SearchHandler.java
+++ b/container-search/src/main/java/com/yahoo/search/handler/SearchHandler.java
@@ -139,6 +139,8 @@ public class SearchHandler extends LoggingRequestHandler {
         this.numRequestsLeftToTrace = new AtomicLong(numQueriesToTraceOnDebugAfterStartup);
         metric.set(SEARCH_CONNECTIONS, 0.0d, null);
         this.zoneInfo = zoneInfo;
+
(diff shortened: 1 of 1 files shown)


container-search/src/main/java/com/yahoo/search/handler/SearchHandler.java
```diff
@@ -150,6 +152,24 @@ private static int examineExecutor(Executor executor) {
         return Integer.MAX_VALUE; // assume unbound
     }
 
+    private void warmup() {
+        try {
+            String uri = "/search/?yql=select+*+from+sources+*+where+true+limit+1;";
+            HttpRequest request = HttpRequest.createTestRequest(uri, com.yahoo.jdisc.http.HttpRequest.Method.GET);
+            Query query = new Query.Builder().setRequest(request)
```
jonmv, line 159: Hmm, perhaps this can be simpler ...

bratseth (approved):

bratseth:
LGTM!

aressem:
This broke a handful of system tests : https://factory.vespa.oath.cloud/testrun/63143128

jonmv:
Yes, mostly metrics expecting 0 queries, but there was also one instance where the warmup query failed. I'll have a look at all of it.

PR #27275, opened 2023-06-02
Avoid blocking CommunicationManager thread during cluster state transitions

vekterli:
@havardpe please review
@baldersheim FYI

Incoming cluster state versions are not applied locally on a content node until all potentially conflicting operations running in the persistence threads have completed and all potentially conflicting operations in the persistence queues have been aborted.

This can take a relatively long time when running LID space compactions etc via the persistence threads, and we'd risk blocking the main `CommunicationManager` thread (which handles all cluster controller communication) for prolonged periods of time.

This commit moves operation abortion/draining blocking and internal state propagation to a dedicated task executor. The executor only has 1 thread, effectively turning it into an asynchronous FIFO executor. Semantics should otherwise be the same as with the old behavior.

### DIFF
--- a/storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
+++ b/storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
@@ -69,7 +69,7 @@ struct ChangedBucketOwnershipHandlerTest : Test {
 
     std::shared_ptr<AbortBucketOperationsCommand> fetch_dispatched_abort_operations_command() {
         _bottom->waitForMessages(2, 60); // abort cmd + set cluster state cmd
-        EXPECT_EQ(2, _bottom->getNumCommands()); // always _at least_ 2
+        EXPECT_EQ(2, _bottom->getNumCommands());
         return std::dynamic_pointer_cast<AbortBucketOperationsCommand>(_bottom->getCommand(0));
     }
 
--- a/storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
+++ b/storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
@@ -40,15 +40,11 @@ struct ChangedBucketOwnershipHandlerTest : Test {
             uint16_t wantedOwner,
             const lib::ClusterState& state);
 
-    std::shared_ptr<api::SetSystemStateCommand> createStateCmd(
-            const lib::ClusterState& state) const
-    {
+    std::shared_ptr<api::SetSystemStateCommand> createStateCmd(const lib::ClusterState& state) const {
         return std::make_shared<api::SetSystemStateCommand>(state);
     }
 
-    std::shared_ptr<api::SetSystemStateCommand> createStateCmd(
-            const std::string& stateStr) const
-    {
+    std::shared_ptr<api::SetSystemStateCommand> createStateCmd(const std::string& stateStr) const {
         return createStateCmd(lib::ClusterState(stateStr));
     }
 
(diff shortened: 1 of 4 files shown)


storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
```diff
@@ -71,11 +67,17 @@ struct ChangedBucketOwnershipHandlerTest : Test {
     template <typename MsgType, typename... MsgParams>
     void expectDownAbortsMessage(bool expected, MsgParams&& ... params);
 
-    lib::ClusterState getDefaultTestClusterState() const {
+    std::shared_ptr<AbortBucketOperationsCommand> fetch_dispatched_abort_operations_command() {
+        _bottom->waitForMessages(2, 60); // abort cmd + set cluster state cmd
+        EXPECT_EQ(2, _bottom->getNumCommands()); // always _at least_ 2
```
havardpe, line 72: at least? GE?

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
```diff
@@ -235,18 +236,77 @@ ChangedBucketOwnershipHandler::makeLazyAbortPredicate(
...
+     * 4) Some nodes down or up
+     *      - abort ops for buckets that have changed ownership between
+     *        current and new cluster state.
+     */
+    void run() override {
+        OwnershipState::CSP old_ownership;
+        OwnershipState::CSP new_ownership;
+        // Get old state and update own current cluster state _before_ it is
+        // applied to the rest of the system. This helps ensure that no message
+        // can get through in the off-case that the lower level storage links
```
havardpe, line 268: This comment seems off; we should start rejecting now, but the other components will not known about the state change until after conflicting operations have been successfully aborted.

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
```diff
@@ -255,47 +315,11 @@ ChangedBucketOwnershipHandler::onSetSystemState(
...
-    // on the lower level links, at which point it is safe to send down
-    // the SetSystemStateCommand.
-    sendDown(cmd);
-
-    durationTimer.stop(_metrics.averageAbortProcessingTime);
-    return false;
+    // Dispatch to background worker. This indirection is because operations such as lid-space compaction
+    // may cause the implicit operation abort waiting step to block the caller for a relatively long time.
+    // It is very important that the executor only has 1 thread, which means this has FIFO behavior.
+    _state_sync_executor.execute(std::make_unique<ClusterStateSyncAndApplyTask>(*this, stateCmd));
```
havardpe, line 321: should check result since we do shutdown in onClose

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
```diff
@@ -451,4 +474,10 @@ ChangedBucketOwnershipHandler::onInternalReply(
     return (reply->getType() == AbortBucketOperationsReply::ID);
 }
 
+void
+ChangedBucketOwnershipHandler::onClose()
+{
+    _state_sync_executor.shutdown();
```
havardpe, line 480: should do sync somewhere, either here (to wait for pending tasks) or in the destructor of ChangedBucketOwnershipHandler

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
vekterli: The expectation is that we should not receive any further messages through the pipeline after `onClose` has completed, as the `CommunicationManager` is closed _before_ this component and that explicitly shuts down and syncs the content node RPC subsystems that can deliver messages unto our doorstep.

Since this should not happen, I'll add an explicit assertion for this.

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
vekterli: Comment is indeed old—it belonged to the days of yore when the `CommunicationManager` message thread was the only thread allowed to dispatch messages. I have rewritten it.

storage/src/tests/storageserver/changedbucketownershiphandlertest.cpp
vekterli: Exact equality is the intended semantics. I've removed this particular comment to avoid confusion.

storage/src/vespa/storage/storageserver/changedbucketownershiphandler.cpp
vekterli: Added an explicit `sync()` after `shutdown()`

vekterli:
@havardpe PTAL

havardpe (approved):

PR #27276, opened 2023-06-02
Pass two int64_t values to std::min

toregge:
@baldersheim : please review


baldersheim (approved):

vespalib/src/vespa/vespalib/testkit/time_bomb.cpp
```diff
@@ -14,7 +15,7 @@ void bomb(Gate &gate, vespalib::duration timeout) {
             return;
         }
     }
-    size_t countdown = std::min(count_s(timeout), 5l);
+    size_t countdown = std::min(count_s(timeout), INT64_C(5));
```
baldersheim, line 18: Which compiler combo is unhappy here ?

toregge: Any compiler on Darwin, where int64_t is `long long int`.

PR #27279, opened 2023-06-02
Allow parallell cloud accounts in dep-spec, and simply some usages

jonmv:
@freva please review.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
@@ -4,6 +4,7 @@ package com.yahoo.config.application.api;
 import ai.vespa.validation.Validation;
 import com.yahoo.config.provision.AthenzService;
 import com.yahoo.config.provision.CloudAccount;
+import com.yahoo.config.provision.CloudName;
 import com.yahoo.config.provision.ClusterSpec;
 import com.yahoo.config.provision.Environment;
 import com.yahoo.config.provision.InstanceName;
@@ -59,7 +60,7 @@ public class DeploymentInstanceSpec extends DeploymentSpec.Steps {
     private final List<DeploymentSpec.ChangeBlocker> changeBlockers;
     private final Optional<String> globalServiceId;
     private final Optional<AthenzService> athenzService;
-    private final Optional<CloudAccount> cloudAccount;
+    private final Map<CloudName, CloudAccount> cloudAccounts;
     private final Optional<Duration> hostTTL;
     private final Notifications notifications;
     private final List<Endpoint> endpoints;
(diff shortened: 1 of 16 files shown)


config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentInstanceSpec.java
```diff
@@ -266,22 +264,22 @@ public Optional<AthenzService> athenzService(Environment environment, RegionName
...
                       .findFirst()
-                      .flatMap(DeploymentSpec.DeclaredZone::cloudAccount)
-                      .or(() -> cloudAccount);
+                      .map(DeploymentSpec.DeclaredZone::cloudAccounts)
+                      .orElse(cloudAccounts);
     }
 
     /** Returns the host TTL to use for given environment and region, if any */
-    public Optional<Duration> hostTTL(Environment environment, Optional<RegionName> region) {
+    public Optional<Duration> hostTTL(Environment environment, RegionName region) {
```
jonmv, line 280: Oops, this is in used in config model, lemme fix.

config-model-api/src/test/java/com/yahoo/config/application/api/DeploymentSpecTest.java
```diff
@@ -1782,18 +1787,24 @@ public void cloudAccount() {
...
-        assertCloudAccount("300000000000", spec.requireInstance("main"), prod, "us-east-1");
-        assertCloudAccount("100000000000", spec.requireInstance("main"), prod, "eu-west-1");
-        assertCloudAccount("400000000000", spec.requireInstance("main"), dev, "");
-        assertCloudAccount("500000000000", spec.requireInstance("main"), test, "");
-        assertCloudAccount("100000000000", spec.requireInstance("main"), staging, "");
-        assertCloudAccount("default", spec.requireInstance("beta"), prod, "us-west-2");
+        assertEquals(Map.of(AWS, CloudAccount.from("100000000000"),
+                            GCP, CloudAccount.from("gcp:foobar")), spec.cloudAccounts());
+        assertCloudAccount("800000000000", spec, AWS, "alpha", prod,    "us-east-1");
+        assertCloudAccount("",             spec, GCP, "alpha", prod,    "us-east-1");
```
freva, line 1793: Empty string is so ugly though.. Why not 'default'?

freva (commented):
Unit test failure

config-model-api/src/test/java/com/yahoo/config/application/api/DeploymentSpecTest.java
jonmv: Both work ... but empty string here means custom accounts for _no_ clouds, because the semantics of the string is that each cloud with a custom account is listed, comma separated, in the format accepted by `CloudAccount.from`, which is documentable. If we want, we can document that `default` is an alias for that.

freva (approved):

PR #27292, opened 2023-06-05
Setup distance metrics for streaming search.

toregge:
Add range checks when converting to internal distance threshold.

@geirst : please review


geirst (approved):
👍

PR #27302, opened 2023-06-06
Upgrade host flavor

mpolden:
Merge with internal PR.

@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
@@ -1,0 +1,104 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.provision.maintenance;
+
+import com.yahoo.config.provision.Deployer;
+import com.yahoo.config.provision.NodeAllocationException;
+import com.yahoo.config.provision.NodeType;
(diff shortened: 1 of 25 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
```diff
@@ -0,0 +1,104 @@
...
+                deploymentValid = deployment.isValid();
+                if (!deploymentValid) continue;
+
+                attempts++;
+                log.log(Level.INFO, () -> "Redeploying " + node.allocation().get().owner() + " to upgrade flavor (" +
+                                          parent.get().flavor().name() + ") of " + parent.get());
+                upgradeFlavor(parent.get(), true);
+                deployment.activate();
+                redeployed = true;
+                return asSuccessFactorDeviation(attempts, failures);
```
bratseth, line 83: Is it really a failure that no capacity of the newer flavor is available at the moment? If we assume that is quite normal (i.e not something ops should look into) I think it would be better to return 1 here.

bratseth (dismissed):
Great!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
mpolden: Good point, will change it.

mpolden:
Rebased to fix conflict.

bratseth (approved):

PR #27309, opened 2023-06-06
Add a simple token primitive to security utils 

vekterli:
@bjorncs please review
@tokle FYI

### DIFF
--- a/security-utils/src/main/java/com/yahoo/security/token/Token.java
+++ b/security-utils/src/main/java/com/yahoo/security/token/Token.java
@@ -1,0 +1,85 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.security.token;
+
+import com.yahoo.security.HKDF;
+
+import java.util.Objects;
+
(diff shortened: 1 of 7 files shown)


security-utils/src/main/java/com/yahoo/security/token/Token.java
```diff
@@ -0,0 +1,85 @@
...
+    // of the hash code.
+    @Override
+    public int hashCode() {
+        return Objects.hash(domain, fingerprint);
+    }
+
+    @Override
+    public String toString() {
+        // Avoid leaking raw token secret as part of toString() output
+        return "Token(fingerprint: %s)".formatted(fingerprint.toHexString());
```
bjorncs, line 70: Why is `domain` excluded from `toString()`?

bjorncs (dismissed):

security-utils/src/main/java/com/yahoo/security/token/Token.java
vekterli: No reason in particular—the test's primary purpose is to ensure that the secret itself is not printed. I'll update the test name to reflect this and add the domain to the output as well.

bjorncs (approved):

PR #27314, opened 2023-06-06
Avoid maintaining routing policies for non-active load balancers

mpolden:
When deactivating a deployment, the config server moves the load balancer to
inactive. Since the LB was still present, we kept its routing policy (and DNS
record) even though both should've been removed.


mpolden:
@tokle

tokle (approved):

PR #27318, opened 2023-06-06
Disable inline warnings when compiling with gcc 13.

toregge:
@baldersheim : please review


baldersheim (commented):
I suggest we try to avoid this, unless it hits very many compilation units.
It is normally a sign to excessive inlining.

baldersheim:
Not needed anymore .....

PR #27319, opened 2023-06-06
Revert "Validate prefix matching"

hmusum:
Reverts vespa-engine/vespa#27285

Broke 13 system tests, not sure if this was expected and tests should be fixed or if this needs to be reverted.


arnej27959 (approved):
Ok

PR #27338, opened 2023-06-08
No CPU capping in test zones

hmusum:
Try without CPU capping to see if it helps with first query after start taking a long time.


bratseth (approved):

baldersheim (approved):

bratseth:
If this works we need to followup somehow since staging tests should support performance validation.

freva:
This value is already controlled by the `container-cpu-cap` feature flag, except for the method changed in this PR which bypasses the flag entirely. In this case I'd recommend just using the flag, then change the code if you want it to be permanent.

Though ideally we should remove the method and always use the feature flag.

hmusum:
Nice, then we can control this per app in staging and this will have no impact on other apps. I'll try to remove the method and always use the flag

hmusum:
https://github.com/vespa-engine/vespa/pull/27343 changes this to only use the feature flag

PR #27340, opened 2023-06-08
avoid having "index" command for fields without an index

arnej27959:
@bratseth please review
@hmusum @baldersheim FYI
NOTE: this change has large implications and will likely require changes various places in container-search


bratseth (approved):
I think this is safe because I couldn't find any use of the "index" command when looking into this earlier. There's also a different place that adds the "index" command for fieldsets which should be changed the same way.

Since I want to deprecate index-commands and IndexFacts and friends in favor of the the SchemaInfo mechanism I opted to fix the validator by rewriting to that (and adding the missing stuff needed to SchemaInfo) instead - you'll get a PR on that soon.

PR #27342, opened 2023-06-08
Skip logging when rethrowing

mpolden:
Avoids logging stack trace when HostFlavorUpgrader redeployment fails due to
lack of capacity.

@bratseth


bratseth (commented):
Don't we want to increment the metric in both cases?

bratseth (approved):

PR #27345, opened 2023-06-08
Allow importing user yahoo packages

gjoranv:


gjoranv:
Replaced by simpler and more robust #27364

PR #27346, opened 2023-06-08
Remove unneeded atomic thread fences.

toregge:
@geirst : please review


toregge:
The thread fences were added as replacement for memory barriers implemented with volatile asm statements a long
time ago when x86_64 was tne only target platform. We now use release type atomic store and aquire type atomic load
to ensure that writes peformed by the write thread are properly visible for the read threads.

geirst (approved):
👍

PR #27348, opened 2023-06-08
Make failure be the default when reporting success factor

oyving:


oyving:
It looks like the semantics of the return value has changed. The default should be the failure state such that exceptions are counted as failures.

bjormel (approved):

PR #27352, opened 2023-06-08
Keep only meta data in truncated package by default

jonmv:
@freva

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageTest.java
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageTest.java
@@ -229,7 +229,7 @@ public class ApplicationPackageTest {
         assertEquals(content, unzip(zip));
         AngryStreams angry = new AngryStreams(zip);
 
-        ApplicationPackageStream identity = new ApplicationPackageStream(angry::stream);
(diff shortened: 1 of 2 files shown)


jonmv:
Looks like the whole package was accidentally buffered again, some time in December last year, when a constructor call was changed to the single-argument one. Now that keeps only the metadata.

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageTest.java
```diff
@@ -229,7 +229,7 @@ void testApplicationPackageStream() throws Exception {
         assertEquals(content, unzip(zip));
         AngryStreams angry = new AngryStreams(zip);
 
-        ApplicationPackageStream identity = new ApplicationPackageStream(angry::stream);
+        ApplicationPackageStream identity = new ApplicationPackageStream(angry::stream, () -> __ -> true);
```
freva, line 232: (☞ﾟヮﾟ)☞

freva (approved):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/application/pkg/ApplicationPackageTest.java
jonmv: :D

PR #27366, opened 2023-06-10
Revert "Validate semantic rules when building config model" MERGEOK

aressem:
Reverts vespa-engine/vespa#27356

Factory broken


aressem:
Fix in https://github.com/vespa-engine/vespa/pull/27365

PR #27374, opened 2023-06-12
- Implement array lookup by using the attribute node and setting curr…

baldersheim:
…ent index.

- Unify code layout.

@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/expression/arrayatlookupfunctionnode.cpp
+++ b/searchlib/src/vespa/searchlib/expression/arrayatlookupfunctionnode.cpp
@@ -1,157 +1,97 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 #include "arrayatlookupfunctionnode.h"
-#include "floatresultnode.h"
-#include "integerresultnode.h"
-#include "stringresultnode.h"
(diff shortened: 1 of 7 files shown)


searchlib/src/vespa/searchlib/expression/arrayatlookupfunctionnode.cpp
```diff
@@ -1,157 +1,97 @@
...
 
-void ArrayAtLookup::wireAttributes(const search::attribute::IAttributeContext & attrCtx)
+Deserializer &
+ArrayAtLookup::onDeserialize(Deserializer & is)
 {
-    _attribute = attrCtx.getAttribute(_attributeName);
-    if (_attribute == nullptr) {
-        throw std::runtime_error(vespalib::make_string("Failed locating attribute vector '%s'", _attributeName.c_str()));
-    }
+    FunctionNode::onDeserialize(is);
```
havardpe, line 73: need comment, this is not intuitive

havardpe (dismissed):

searchlib/src/vespa/searchlib/expression/arrayatlookupfunctionnode.cpp
baldersheim: Fixed

PR #27383, opened 2023-06-12
rw spin lock

havardpe:
still only experimental; both the lock itself and its benchmarking

@vekterli please review


vekterli (approved):
:+1:

PR #27384, opened 2023-06-12
Balder/unify more attribute node related code

baldersheim:
@havardpe PR


baldersheim:
@havardpe I think is better to review commits in order.

havardpe (dismissed):
consider making onPrepare final in AttributeNode

baldersheim:
Final it is

PR #27386, opened 2023-06-12
Ignore GCP internal IPv6 addresses

hakonhall:


node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/network/IPAddresses.java
```diff
@@ -56,12 +57,14 @@ default List<String> getAddresses(String hostname, IPVersion ipVersion) {
      * @throws ConvergenceException if multiple addresses are found
      */
     default Optional<Inet6Address> getIPv6Address(String hostname) {
+        CidrBlock gcpInternalBlock = CidrBlock.fromString("2600:2d00::/32");
```
freva, line 60: Make static?

freva (dismissed):

node-admin/src/main/java/com/yahoo/vespa/hosted/node/admin/task/util/network/IPAddresses.java
hakonhall: Done.  Unfortunately it cannot be made private.

freva (approved):

PR #27388, opened 2023-06-12
For USER bundles, add Vespa build version to manifest.

gjoranv:


bundle-plugin/src/main/java/com/yahoo/container/plugin/mojo/GenerateOsgiManifestMojo.java
```diff
@@ -148,6 +152,7 @@ private String wantedProvidedDependency() {
     }
 
     private BundleType effectiveBundleType() {
+        getLog().warn("XXXXXXX Bundle " + project.getArtifactId() + " has bundle type " + bundleType);
```
bjorncs, line 155: Leftover from debugging?

bjorncs (dismissed):

PR #27390, opened 2023-06-12
Change message when service is not up and running with new config

hmusum:


hmusum:
@bratseth changed the message as you suggested, please take a look

bratseth (approved):
👍

Issue #23395, opened 2022-07-06
Fail config server start-up if "upgrade interval" is too large


frodelu:
Per [our documentation](https://vespa.ai/releases.html) we expect Vespa versions to be less than two months apart to maintain our semantic versioning commitment. Since this is the stated policy, we occasionally see cases where a too large "jump" in Vespa versions – even within a major version – can cause problems (e.g. issue #23223).

When the config server starts up, it knows what version of Vespa was previously run on this node. Should we add a sanity check that would fail to start (and leave ZK untouched) if the "upgrade interval" is more than the supported two months and instruct the user to upgrade via a safe path?

PR #27394, opened 2023-06-13
Log or fail config server startup when upgrading between 2 versions t…

hmusum:
…oo far apart

https://github.com/vespa-engine/vespa/issues/23395
Fail on hosted, log on self serve Vespa (can be changed to fail for both when we have tried it out)


hakonhall (dismissed):
If a user gets this error - it may even be no such intermediate versions exists anymore.  Is there a way for him to acknowledge and ignore the error in the hope it may still work?

hmusum:
Environment variable VESPA_SKIP_UPGRADE_CHECK can be used for this now

hakonhall (approved):

PR #27381, opened 2023-06-12
Pass array size to allocArray member function.

toregge:
@geirst : please review

PR #27402, opened 2023-06-13
Revert "Pass array size to allocArray member function."

toregge:
Reverts vespa-engine/vespa#27381

Methods with changed signatures are not used by array store.


geirst (approved):
👍

PR #27412, opened 2023-06-13
Add CurrentIndex to GroupingLevel

baldersheim:
@havardpe PR


havardpe (dismissed):
consider specifying copy constructor as default since it is used (MultiGrouper)

PR #27415, opened 2023-06-13
Add get_entry_size member function in array store type mappers that maps

toregge:
from type id to entry size.

@geirst : please review


geirst (approved):
👍

PR #27417, opened 2023-06-14
Warn on low disk compared to memory

mpolden:
@bratseth


bratseth (approved):

bjormel:
@mpolden, seems like we are warning on this when using shared host, was that intended?

mpolden:
@bjormel Yes, I think warning in all scenarios makes sense.

PR #27434, opened 2023-06-15
Add cloud account details in /deployment/v1 as well

jonmv:
@freva please review, @ldalves FYI :)

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelper.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelper.java
@@ -15,6 +15,7 @@ import com.yahoo.text.Text;
 import com.yahoo.vespa.hosted.controller.Application;
 import com.yahoo.vespa.hosted.controller.Controller;
 import com.yahoo.vespa.hosted.controller.NotExistsException;
+import com.yahoo.vespa.hosted.controller.api.identifiers.DeploymentId;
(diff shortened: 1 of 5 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelper.java
```diff
@@ -395,6 +391,12 @@ static HttpResponse overviewResponse(Controller controller, TenantAndApplication
                     toSlime(runObject.setObject("versions"), versions.versions(), application);
                 }
 
+                if ( ! jobStatus.runs().isEmpty())
+                    controller.applications().decideCloudAccountOf(new DeploymentId(job.application(),
```
freva, line 395: I was hoping we could make `decideCloudAccountOf()` private and always use `Deployment::cloudAccount()`

freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/application/JobControllerApiHandlerHelper.java
jonmv: Doesn't work for test jobs which have only temporary deployments, though.

jonmv: We'd have to add it to each run in that case.

PR #27436, opened 2023-06-15
Stricter validation when creating ApplicationId from a string

hmusum:


config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
```diff
@@ -54,18 +59,23 @@ public static ApplicationId from(String tenant, String application, String insta
...
-        String[] parts = idString.split("\\.");
-        if (parts.length < 3)
-            throw new IllegalArgumentException("Application ids must be on the form tenant.application.instance, but was " + idString);
+        return fromIdString(idString, "\\.");
+    }
 
+    private static ApplicationId fromIdString(String idString, String splitCharacter) {
+        String[] parts = idString.split(splitCharacter);
+        String errorMessage = "Application ids must be on the form tenant" +
+                splitCharacter + "application" + splitCharacter + "instance, but was " + idString;
```
hakonhall, line 72: Note that this will print a slightly confusing application ID: "... the form tenant\\.application\\.instance, but..."

config-provisioning/src/test/java/com/yahoo/config/provision/ApplicationIdTest.java
```diff
@@ -81,9 +82,34 @@ void require_that_idstring_can_be_parsed() {
 
     @Test
     void require_that_invalid_idstring_throws_exception() {
-        assertThrows(IllegalArgumentException.class, () -> {
+        try {
             ApplicationId.fromSerializedForm("foo:baz");
-        });
+            fail("Expected to fail");
```
hakonhall, line 87: The code with assertThrows is perhaps even more readable:

```
        var e = assertThrows(IllegalArgumentException.class, () -> {
            ApplicationId.fromSerializedForm("foo:baz");
        });
        assertEquals("Application ids must be on the form tenant:application:instance, but was foo:baz", e.getMessage());
```

hakonhall (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
hmusum: Yes, good point, fixed and made sure to actually test both forms

hakonhall (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/ApplicationId.java
```diff
@@ -54,18 +59,24 @@ public static ApplicationId from(String tenant, String application, String insta
...
     public static ApplicationId fromFullString(String idString) {
-        String[] parts = idString.split("\\.");
-        if (parts.length < 3)
-            throw new IllegalArgumentException("Application ids must be on the form tenant.application.instance, but was " + idString);
+        return fromIdString(idString, "\\.");
+    }
 
+    private static ApplicationId fromIdString(String idString, String splitCharacter) {
+        String[] parts = idString.split(splitCharacter);
+        String unescapedSplitCharacter = splitCharacter.equals("\\.") ? "." : splitCharacter;
```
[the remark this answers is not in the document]
hmusum: Yep, nicer, thanks, done

freva (approved):

PR #27442, opened 2023-06-15
Add token endpoint

tokle:
@olaaun fyi


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/routing/RoutingPoliciesTest.java
```diff
@@ -322,6 +325,103 @@ private void zone_routing_policies(boolean sharedRoutingLayer) {
         assertEquals(4, tester.routingPolicies().read(context1.instanceId()).size(), "Keeps routing policies for " + context1.application());
     }
 
+    @Test
+    void zone_token_endpoints() {
```
mpolden, line 329: I think this can be simplified a bit. This can just do a single deployment and check whether the token endpoint was created.

mpolden (approved):

PR #27443, opened 2023-06-16
Limit DNS updates to endpoints targeting current deployment

mpolden:
This necessitated some changes to semantics around what happens when taking out
the last deployment in an endpoint. We now refuse this action instead of putting
all targets back in.

@jonmv


jonmv (commented):
Nice!

PR #27444, opened 2023-06-16
Make sure to shutdown maintainer properly

hmusum:
Call deconstruct on metrics fetcher

Log message "NodeMetricsDbMaintainer failed to shutdown within PT30S" seen in the wild


bratseth (approved):
Thanks!

PR #27463, opened 2023-06-16
- Add explicit test that onInsert/onRemove is called correctly when c…

baldersheim:
…ache is full,

  do not rely on monitoring cache size.
- Call correct method for properly erasing an element, even if it is old :)

@geirst PR


geirst (approved):
👍

PR #27469, opened 2023-06-19
Avoid long (possibly infinite) pauses in maintainer when few apps

hmusum:
Seeing "Maintainer NodeMetricsDbMaintainer failed to shutdown within PT30S" regularly in cd zones with few apps


jonmv (approved):

jonmv:
I guess there are a lot of maintainers that possibly run for a long time, inside a loop. Perhaps we should check for whether the maintainer executor is shutdown in (some of) those loops?

PR #27483, opened 2023-06-19
Check for IPv4 conflicts within each cloud account, not across

jonmv:
@freva please review, @hakonhall FYI :)


freva (dismissed):
Ideally we want to validate that private/local addresses (regardless of version) are unique per icon, while public/global addresses (https://github.com/vespa-engine/vespa/blob/ba83ad3508f3a591bb433c4160d60d3e45c0955b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/WireguardResponse.java#L47) are unique across all accounts.

jonmv:
PTAL.

freva (approved):
:100:

PR #27488, opened 2023-06-20
randomized endpoint cert pool

andreer:

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CertPoolMaintainer.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CertPoolMaintainer.java
@@ -1,0 +1,154 @@
+// Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.vespa.hosted.controller.maintenance;
+
+import com.yahoo.component.annotation.Inject;
+import com.yahoo.config.provision.ApplicationId;
+import com.yahoo.container.jdisc.secretstore.SecretNotFoundException;
+import com.yahoo.container.jdisc.secretstore.SecretStore;
+import com.yahoo.jdisc.Metric;
+import com.yahoo.transaction.Mutex;
+import com.yahoo.vespa.flags.Flags;
+import com.yahoo.vespa.flags.IntFlag;
+import com.yahoo.vespa.hosted.controller.Controller;
+import com.yahoo.vespa.hosted.controller.api.integration.certificates.EndpointCertificateMetadata;
+import com.yahoo.vespa.hosted.controller.api.integration.certificates.EndpointCertificateProvider;
+import com.yahoo.vespa.hosted.controller.application.Endpoint;
+import com.yahoo.vespa.hosted.controller.persistence.CuratorDb;
+
+import java.security.SecureRandom;
+import java.time.Clock;
+import java.time.Duration;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import static com.yahoo.vespa.hosted.controller.maintenance.CertPoolMaintainer.CertificatePool.ready_to_use;
+import static com.yahoo.vespa.hosted.controller.maintenance.CertPoolMaintainer.CertificatePool.requested;
+
+/**
+ * Manages pool of ready-to-use randomized endpoint certificates
+ *
+ * @author andreer
+ */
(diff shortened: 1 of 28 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CertPoolMaintainer.java
```diff
@@ -0,0 +1,152 @@
...
+
+import static com.yahoo.vespa.hosted.controller.maintenance.CertPoolMaintainer.CertificatePool.ready_to_use;
+import static com.yahoo.vespa.hosted.controller.maintenance.CertPoolMaintainer.CertificatePool.requested;
+
+/**
+ * Manages pool of ready-to-use randomized endpoint certificates
+ *
+ * @author andreer
+ */
+public class CertPoolMaintainer extends ControllerMaintainer {
```
mpolden, line 37: CertPoolMaintainer -> CertificatePoolMaintainer.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -739,6 +744,25 @@ public List<DataplaneTokenVersions> readDataplaneTokens(TenantName tenantName) {
...
+
+    public void addToCertificatePool(String s, EndpointCertificateMetadata metadata, String pool) {
+        curator.set(certificatePoolRoot.append(pool).append(s), asJson(EndpointCertificateMetadataSerializer.toSlime(metadata)));
+    }
+
+    public void removeFromCertificatePool(String s, String pool) {
+        curator.delete(certificatePoolRoot.append(pool).append(s));
+    }
+
+    public Map<String, EndpointCertificateMetadata> readCertificatePool(String pool) {
```
mpolden, line 757: I think it would be better to create a concrete pair-type to represent a certificate and it's state, e.g. `record PooledCertificate(EndpointCertificateMetadata, PooledCertificate.State)` and have this method return `List<PooledCertificate>`.

(I also think `EndpointCertificateMetadata` could be simplified to `EndpointCertificate`, but we can do that later)

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CertPoolMaintainer.java
```diff
@@ -0,0 +1,152 @@
...
+                    }
+                }
+            } catch (SecretNotFoundException s) {
+                // Likely because the certificate is very recently provisioned - ignore till next time - should we log?
+                log.log(Level.INFO, "Could not yet read secrets for randomized endpoint cert %s - maybe next time ...".formatted(cert.getKey()));
+            }
+        }
+    }
+
+    enum CertificatePool {
```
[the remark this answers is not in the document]
mpolden: I assume there will be another state here eventuelly, `active` (in use)?

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/CuratorDb.java
```diff
@@ -739,6 +744,25 @@ public List<DataplaneTokenVersions> readDataplaneTokens(TenantName tenantName) {
         return readSlime(dataplaneTokenPath(tenantName)).map(DataplaneTokenSerializer::fromSlime).orElse(List.of());
     }
 
+    // -------------- Endpoint certificate pool -------------------------------
+
+    public void addToCertificatePool(String s, EndpointCertificateMetadata metadata, String pool) {
+        curator.set(certificatePoolRoot.append(pool).append(s), asJson(EndpointCertificateMetadataSerializer.toSlime(metadata)));
```
mpolden, line 750: If we model it like this we need to use use transactions to ensure atomicity when certificate state changes. However, if we instead store a `PooledCertificate` as a single object (see below), it can only ever have one state, so I suggest we do that.

Example:
`/controller/v1/certificatePool/<generated-identifier>`, which contains:
```
{
  "state": <pooled-certificate-state>,
  "certificate": <endpoint-certificate-metadata>,
}
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/CertPoolMaintainer.java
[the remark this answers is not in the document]
andreer: No, my plan was to just remove them, when they are assigned to an application the metadata will be written to where it is stored today.

andreer:
This batch of commits should resolve most of it, a PooledCertificate class managed to sneak into the last one. Will refactor to List<PooledCertificate> tomorrow when temperatures are lower

mpolden (approved):

PR #27495, opened 2023-06-20
Add constexpr specifiers to aligner member functions that can be

toregge:
evaluated at compile time.

@geirst : please review


geirst (approved):
👍

PR #27499, opened 2023-06-21
Store dynamic array size size at start of entry.

toregge:
@geirst : please review


geirst (approved):
👍

aressem:
Seems to have shifted the performance a bit on factory.

PR #27522, opened 2023-06-22
Hmusum/node repo cleanup 1

hmusum:
Mostly cleanup, last commit takes disk speed, storage type and architecture into consideration when provisioning hosts according to PREPROVISION_CAPACITY feature flag.


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -203,7 +191,9 @@ private List<Node> provisionUntilNoDeficit(NodeList nodeList) {
                 throw new IllegalStateException("Have provisioned " + numProvisions + " times but there's still deficit: aborting");
             }
 
+            System.out.println("nodesPlusProvisioned before: " + nodesPlusProvisioned);
```
freva, line 194:
```suggestion
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -203,7 +191,9 @@ private List<Node> provisionUntilNoDeficit(NodeList nodeList) {
...
             nodesPlusProvisioned.addAll(provisionHosts(deficit.get().count(), toNodeResources(deficit.get())));
+            System.out.println("nodesPlusProvisioned after: " + nodesPlusProvisioned);
```
freva, line 196:
```suggestion
```

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -282,11 +272,42 @@ true, nodeRepository().nameResolver(), nodeRepository().nodes(), nodeRepository(
...
+    private static List<Node> getChildren(Map<Optional<String>, List<Node>> nodesByParent, Node host) {
+        return nodesByParent.getOrDefault(Optional.of(host.hostname()), List.of());
+    }
+
+    private static boolean allChildrenCanBeDeprovisioned(Map<Optional<String>, List<Node>> nodesByParent, Node host) {
+        return getChildren(nodesByParent, host).stream().allMatch(HostCapacityMaintainer::canDeprovision);
+    }
+
+    private static boolean someChildrenCannotBeDeprovisioned(Map<Optional<String>, List<Node>> nodesByParent, Node host) {
+        return  ! allChildrenCanBeDeprovisioned(nodesByParent, host);
```
freva, line 310: This will always return `false` if the given host has no children?

hmusum: Good catch, thanks, fixed

freva (approved):

PR #27529, opened 2023-06-23
Do not pass children to HostProvisioner::provision

freva:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -373,6 +373,8 @@ public class Nodes {
 
     /** Update IP config for nodes in given config */
     public void setIpConfig(HostIpConfig hostIpConfig) {
+        // Ideally this should hold the unallocated lock over the entire method, but unallocated lock must be taken
+        // after the application lock, making this impossible
         Predicate<Node> nodeInConfig = (node) -> hostIpConfig.contains(node.hostname());
         performOn(nodeInConfig, (node, lock) -> {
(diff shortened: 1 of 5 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -373,6 +373,8 @@ private Node failOrMark(Node node, Agent agent, String reason, Mutex lock) {
 
     /** Update IP config for nodes in given config */
     public void setIpConfig(HostIpConfig hostIpConfig) {
+        // Ideally this should hold the unallocated lock over the entire method, but unallocated lock must be taken
+        // after the application lock, making this impossible
```
jonmv, line 377: Conflict alert beep beep beep. Maybe no need for this comment if we decide to take that super-lock in the other PR?

freva: Yes, I dont know if your super lock handles this?

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
jonmv: Takes all the node locks first, then the host lock, then the unallocated, so it does hold all of them at the same time, which is what you want here, but also the potential downside, as one slow lock may cause others to be held longer. It _achieves_ this state by trying a few times to read all nodes after getting the unallocated lock last. No guarantee it'll succeed, if new nodes are added all the time. There is a practical limit to the number of nodes on a host, though.

PR #27534, opened 2023-06-23
Set the recommended searchable copies with streaming

bratseth:
With streaming it is better to make all copies ready. This does it automatically so we can avoid documenting that users need to make sure of this by themselves.


config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
```diff
@@ -470,20 +471,23 @@ void testRedundancyRequired() {
             new VespaModelCreatorWithMockPkg(getHosts(), xml, sds).create();
             fail("Deploying without redundancy should fail");
         } catch (IllegalArgumentException e) {
-            assertTrue(e.getMessage().contains("Either <redundancy> or <min-redundancy> must be set"), e.getMessage());
+            System.out.println("Ya, the message is '" + Exceptions.toMessageString(e) + "'");
```
geirst, line 474: I guess this should be removed?

geirst (dismissed):
👍

PR #27540, opened 2023-06-23
Balder/add query time control of rank score drop limit

baldersheim:
@bratseth or @geirst or @havardpe PR


baldersheim:
Documented in https://github.com/vespa-engine/documentation/pull/2736

bratseth (approved):
👍

PR #27542, opened 2023-06-24
Limit 64-byte dynamic array buffer type alignment based on element type.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #27544, opened 2023-06-26
add euclidean_distance

arnej27959:
@havardpe please review
@lesters @bratseth FYI


havardpe (dismissed):
consider adding test of actual primitive expression

PR #27548, opened 2023-06-26
Implement command to generate markdown files from --help pages

tmaregge:
This PR implements a new CLI command which generates markdown files from the --help pages in the Vespa CLI.
The plan is to use this to automatically generate documentation for the CLI (see https://jira.ouryahoo.com/browse/VESPA-26867).


kkraune:
This is great - with this, we can automate a copy into Vespa Docs, so the help texts can be found is search! I will look into this once the Vespa CLI is released with this

mpolden (approved):
👍

PR #27549, opened 2023-06-26
Cap number of entries in a dense tensor store buffer to avoid very

toregge:
large buffers.

@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #27552, opened 2023-06-26
Avoid double negation to improve readability.

baldersheim:
@havardpe PR


havardpe (commented):
I still find it confusing. Maybe it is wrong?

If grouping needs ranking but we ask for 0 hits and sort on something other than ranking, we end up with 'willNeedRanking' false. This takes away access to the rank score of results from the grouping request.

PR #27558, opened 2023-06-27
Allow patching wantToUpgradeFlavor

mpolden:
@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
@@ -80,6 +80,8 @@ public class HostFlavorUpgrader extends NodeRepositoryMaintainer {
                 return 1.0;
             } catch (NodeAllocationException e) {
                // Fine, no capacity for upgrade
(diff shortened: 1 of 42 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
```diff
@@ -80,6 +80,8 @@ private double upgradeHostFlavor(NodeList allNodes) {
                 return 1.0;
             } catch (NodeAllocationException e) {
                // Fine, no capacity for upgrade
+                log.log(Level.WARNING, "Failed to redeploy " + node.allocation().get().owner() + " to upgrade flavor (" +
```
bratseth, line 83: I don't think we should warn here? We'll emit a warning each time we try but can't upgrade to a newer flavor ...

bratseth (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostFlavorUpgrader.java
mpolden: Yes, but I need it for debugging an issue. I can remove it later.

Issue #26723, opened 2023-04-12
YQL parsing fails when the bitwise AND operator is supplied Long.MIN_VALUE


wsandeepd:
**Describe the bug**
We’re running into an issue with one of our YQL queries involving bitwise operators. the YQL for our use-case is something like below -

```
"select * from my_schema where (version contains \"20230411150312\") limit 0 | all(group(predefined(or(0, and(-9223372036854775808, array.at(my_multivalued_attr, 10))), bucket[0, 0])) each(output(sum(my_agg_value))))"
```
The number in the `and` bitwise operator clause can be any `Long` value which is incidentally `Long.MIN_VALUE` in this case. While for all other values (including `Long.MAX_VALUE`) the query seemingly works as expected, this particular query fails with the following error -

`"Could not create query from YQL: For input string: \"9223372036854775808\""`

```
the stack trace indicates a number format exception -
java.lang.NumberFormatException: For input string: \"9223372036854775808\"\n\tat java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)\n\tat java.base/java.lang.Long.parseLong(Long.java:715)\n\tat java.base/java.lang.Long.valueOf(Long.java:1163)\n\tat com.yahoo.search.grouping.request.parser.GroupingParser.resolveNumber
....
....
```

**To Reproduce**
Run above query with `Long.MIN_VALUE` as the first argument to the bitwise `and` operator.

**Expected behavior**
Since `Long.MIN_VALUE` qualifies as a valid 64-bit signed integer, the query should work as expected without any parsing failures.

**Environment (please complete the following information):**
- OS: Linux
- Infrastructure: Kubernetes

**Vespa version**
8.128.22

jonmv:
It looks like our grammar for parsing grouping queries reads the `and` operands as expressions, which in this case is read as a  `-`, followed by a number; this number should be transformed to its negative, due to the leading `-`, which obviously fails in the reported case (when `Long.MIN_VALUE` is represented). 
Perhaps we can change the grammar to read negative numbers directly in expressions?

bjorncs:
8.185.14 includes a fix that will allow `Long.MIN_VALUE`/`-9223372036854775808` in grouping expressions. There are still some quirks regarding operator precedence as demonstrated by @jonmv. The parser may interpret unary minus operator as binary subtraction. The correct interpretation can be forced by wrapping the negative number with parentheses, e.g `(-3) + foo`.

PR #27559, opened 2023-06-27
Add workaround for handling of unary minus operator in combination wi…

bjorncs:
…th `Long.MIN_VALUE`

### DIFF
--- a/container-search/src/main/javacc/com/yahoo/search/grouping/request/parser/GroupingParser.jj
+++ b/container-search/src/main/javacc/com/yahoo/search/grouping/request/parser/GroupingParser.jj
@@ -35,7 +35,13 @@ public class GroupingParser {
         if (str.indexOf('.') >= 0) {
             return Double.valueOf(str);
         }
+        if (str.equals(Long.toString(Long.MIN_VALUE).substring(1))) return Double.NaN;
         return Long.valueOf(str);
+    }
+
+    private static GroupingExpression resolveNegation(GroupingExpression exp) {
+        if (exp instanceof DoubleValue && ((DoubleValue)exp).getValue().isNaN()) return new LongValue(Long.MIN_VALUE);
+        return new NegFunction(exp);
     }
 }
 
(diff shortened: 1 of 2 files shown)


bjorncs:
This is a workaround for usage of smallest signed 64-bit integer in grouping expressions (#26723) The grouping parser cannot handle negative numbers in expressions - they are represented as unsigned integers wrapped with `neg()`. An unary `-` operator can be forced by wrapping the operator and the single operand with parentheses. Otherwise it might be interpreted as a binary `-`.

container-search/src/main/javacc/com/yahoo/search/grouping/request/parser/GroupingParser.jj
```diff
@@ -35,8 +35,14 @@ public class GroupingParser {
         if (str.indexOf('.') >= 0) {
             return Double.valueOf(str);
         }
+        if (str.equals(Long.toString(Long.MIN_VALUE).substring(1))) return Double.NaN;
         return Long.valueOf(str);
     }
+
+    private static GroupingExpression resolveNegation(GroupingExpression exp) {
+        if (exp instanceof DoubleValue && ((DoubleValue)exp).getValue().isNaN()) return new LongValue(Long.MIN_VALUE);
```
jonmv, line 43: `-NaN`? 🙈

jonmv: Hehe I guess we don't parse `NaN` ... looking at the above. Ok, this is the dirtiest PR I'll ever approve 🤣

jonmv (approved):
Cringe

container-search/src/main/javacc/com/yahoo/search/grouping/request/parser/GroupingParser.jj
bjorncs: My original efforts all stranded with left-recursion which JavaCC does not support. So it ended up with this little hack sadly.

baldersheim (approved):

PR #27561, opened 2023-06-27
Greatly reduce lock acquisitions MERGEOK

jonmv:
@freva

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/FailedExpirer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/FailedExpirer.java
@@ -76,7 +76,7 @@ public class FailedExpirer extends NodeRepositoryMaintainer {
                                          isExpired,
                                          (node, lock) -> recycle(node, List.of(), allNodes).get());
 
-        nodeRepository.nodes().performOnRecursively(allNodes.nodeType(NodeType.host),
+        nodeRepository.nodes().performOnRecursively(allNodes.nodeType(NodeType.host).matching(isExpired),
                                                     nodes -> isExpired.test(nodes.parent().node()),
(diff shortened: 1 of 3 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/FailedExpirer.java
```diff
@@ -76,7 +76,7 @@ protected double maintain() {
                                          isExpired,
                                          (node, lock) -> recycle(node, List.of(), allNodes).get());
 
-        nodeRepository.nodes().performOnRecursively(allNodes.nodeType(NodeType.host),
+        nodeRepository.nodes().performOnRecursively(allNodes.nodeType(NodeType.host).matching(isExpired),
```
jonmv, line 79: Otherwise takes the tenant-host and unallocated-lock for each deprovisioned host ...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -745,7 +745,7 @@ public List<Node> performOn(NodeList nodes, BiFunction<Node, Mutex, Node> action
 
     public List<Node> performOn(NodeList nodes, Predicate<Node> filter, BiFunction<Node, Mutex, Node> action) {
         List<Node> resultingNodes = new ArrayList<>();
-        nodes.stream().collect(groupingBy(Nodes::applicationIdForLock))
+        nodes.matching(filter).stream().collect(groupingBy(Nodes::applicationIdForLock))
```
jonmv, line 748: There are no guarantees as to when the filter is applied, so might as well apply it outside the lock. Whatever we read should be self-consisten.

freva (approved):

PR #27571, opened 2023-06-28
Use 2-8-50 as tester flavor in GCP

freva:


freva:
Done in #27572 instead

PR #27589, opened 2023-06-29
Use real InfraApplication in node-repository tests

freva:
Make the unit tests more realistic and simplify slightly.


mpolden (approved):
Great!

PR #27595, opened 2023-06-30
Delay using released versions as compile version for three hours

jonmv:
@mpolden please review.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/maven/Metadata.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/maven/Metadata.java
@@ -5,9 +5,11 @@ import com.yahoo.component.Version;
 import com.yahoo.text.XML;
 import org.w3c.dom.Element;
 
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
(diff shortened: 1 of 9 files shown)


mpolden (approved):
👍

baldersheim:
This sounds like .... in your pants to keep warm

jonmv:
@baldersheim, the rationale is that users of other mirrors may see artefacts later than we do, and we can't possibly check all mirrors. I'd be open to other suggestions. I don't like this either, but I don't think there's a real cost with delaying our "detecting a version was released" artificially by a few hours.

jonmv:
Hmm ... So I suppose we could check the available maven metadata in the compile-version plugin, find the most up-to-date version there, and request a version not newer than that from the controller ...

mpolden:
Maybe the plugin could include its version from the metadata in the compile version request? That way the controller remains the decision maker, but respects the versions available to the plugin.

jonmv:
That's a simple solution, but we would no longer nudge users into compiling against the latest and greatest Vespa, like we want to. If they were to pin the parent version, it will currently fail after some time, but with this change, it wouldn't. 

There may some means of getting the newest available version, in their current mirror, of the parent artefact. If we could do that, that version would be perfect to use. 

We could also just accept that users may be building against old Vespa artefacts, as it's both not that likely to happen a lot, and we'd also be able to see it in our console, but I'm afraid it'd add an extra failure mode, that costs us extra support in the long run. Maybe not.

PR #27599, opened 2023-06-30
Need larger stack when using address santizer

baldersheim:
@toregge or @vekterli or @geirst PR


vekterli (approved):
:+1: we should probably also check if this applies when using TSan

toregge (approved):

PR #27602, opened 2023-06-30
Do not drop rank if rank-score-drop-limit has been specified explicit.

baldersheim:
@geirst PR


geirst (approved):
👍  Consider adding a system test for this.

baldersheim:
In progress

PR #27613, opened 2023-07-03
Generate endpoint IDs

mpolden:
Will be used to provide "randomized" endpoints. Merge with internal PR.

@jonmv

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerEndpoint.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerEndpoint.java
@@ -1,0 +1,61 @@
+package com.yahoo.vespa.hosted.provision.lb;
+
+import ai.vespa.validation.Validation;
+import com.google.common.hash.Hashing;
+import com.yahoo.config.provision.Environment;
+import com.yahoo.config.provision.RegionName;
+
+import java.nio.charset.StandardCharsets;
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.random.RandomGenerator;
+import java.util.regex.Pattern;
+
+/**
+ * An endpoint for a load balancer, with a pre-generated ID. This ID may be included in the endpoint DNS name.
+ *
+ * @author mpolden
+ */
+public record LoadBalancerEndpoint(String id, AuthMethod authMethod) {
+
+    private static final Pattern ID_PATTERN = Pattern.compile("^[a-f][a-f0-9]{7}$");
+
+    public LoadBalancerEndpoint {
+        Validation.requireMatch(id, "Endpoint ID", ID_PATTERN);
+        Objects.requireNonNull(authMethod);
+    }
+
+    public static List<LoadBalancerEndpoint> createAll(LoadBalancerId loadBalancer, Environment environment, RegionName region) {
+        SecureRandom random = new SecureRandom();
+        return Arrays.stream(AuthMethod.values())
+                     .map(method -> create(loadBalancer, environment, region, method, random))
+                     .toList();
+    }
+
+    static LoadBalancerEndpoint create(LoadBalancerId loadBalancer, Environment environment, RegionName region, AuthMethod method, RandomGenerator random) {
+        String hexLetters = "abcdef";
+        StringBuilder id = new StringBuilder();
+        id.append(hexLetters.charAt(random.nextInt(hexLetters.length()))); // First character must be a letter
+        String hash = Hashing.sha256().newHasher()
+                             .putString(loadBalancer.cluster().value(), StandardCharsets.UTF_8)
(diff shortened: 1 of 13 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerEndpoint.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * An endpoint for a load balancer, with a pre-generated ID. This ID may be included in the endpoint DNS name.
+ *
+ * @author mpolden
+ */
+public record LoadBalancerEndpoint(String id, AuthMethod authMethod) {
+
+    private static final Pattern ID_PATTERN = Pattern.compile("^[a-f][a-f0-9]{7}$");
```
jonmv, line 23: Is that a hash, or a counted value?

jonmv: Right, I found the answer below. Do we retry on collisions, then?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerEndpoint.java
```diff
@@ -0,0 +1,61 @@
...
+                             .putString(loadBalancer.cluster().value(), StandardCharsets.UTF_8)
+                             .putString(loadBalancer.application().serializedForm(), StandardCharsets.UTF_8)
+                             .putString(environment.value(), StandardCharsets.UTF_8)
+                             .putString(region.value(), StandardCharsets.UTF_8)
+                             .putString(method.name(), StandardCharsets.UTF_8)
+                             .putLong(random.nextLong()) // Salt
+                             .hash()
+                             .toString();
+        id.append(hash, 0, 7);
+        return new LoadBalancerEndpoint(id.toString(), method);
```
jonmv, line 51: Not sure I understand what we need the hashing for when we have a random in there that essentially determines the whole value by itself. Since we don't store the random values we drew, we can't brute-force-inverse-map either.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/LoadBalancerSerializer.java
```diff
@@ -135,8 +140,17 @@ public static LoadBalancer fromJson(byte[] data) {
...
 
-        return new LoadBalancer(LoadBalancerId.fromSerializedForm(object.field(idField).asString()),
-                                instance,
+        LoadBalancerId id = LoadBalancerId.fromSerializedForm(object.field(idField).asString());
+        final List<LoadBalancerEndpoint> endpoints;
+        Cursor endpointsArray = object.field(endpointsField);
+        if (endpointsArray.valid()) { // TODO(mpolden): Require this field after 2023-08-01
+            endpoints = new ArrayList<>();
+            endpointsArray.traverse((ArrayTraverser) (idx, endpointObject) -> endpoints.add(new LoadBalancerEndpoint(endpointObject.field(idField).asString(),
+                                                                                                                     LoadBalancerEndpoint.AuthMethod.valueOf(endpointObject.field(authMethodField).toString()))));
```
jonmv, line 149: I'd prefer a custom enum <==> string translator, just in case.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerEndpoint.java
mpolden: Sure, I was just implementing the proposed design. If we rely on purely random numbers alone, I think we also need to check for collisions.

jonmv: I don't think I see the difference between a pure random, and a hash including a pure random. @andreer do you have anything to share on this?

jonmv: Well, there's no such thing as a pure random, I guess ... but this just feels a bit weird :)

mpolden: I assume the motivation for using a one-way hash was that you can't easily guess valid endpoints. If we use random numbers alone, IDs will be predictable, i.e. always a number between 0 and 2^32-1 since we only use 8 characters (32 bits).

mpolden: Hmm, I guess hashing doesn't change this fact since we truncate it to 32 bits.

mpolden: Ok, if we skip hashing we don't need to know the cluster name either. I think we should be able to generate all identifiers in the controller then.

PR #27617, opened 2023-07-03
Avoid allocating non-exclusive children to exclusive hosts of differe…

freva:
…nt application

VESPA-26974: Fixes a bug where we allocated a non-exclusive node to an exclusive (to an application) host.


mpolden (approved):
👍

PR #27621, opened 2023-07-03
Warn if dispatch config is used in services.xml

hmusum:
Log a warning if config for old dispatch process removed in Vespa 7 is used. Add deprecation TODOs and remove code only used in tests


baldersheim (approved):

baldersheim:
@geirst FYI

PR #27626, opened 2023-07-04
Support groups-allowed-down-ratio

hmusum:


hmusum:
@hakonhall for discussion

config-model/src/main/java/com/yahoo/vespa/model/content/ClusterControllerConfig.java
```diff
@@ -122,79 +112,83 @@ public void getConfig(FleetcontrollerConfig.Builder builder) {
...
-                                           minNodeRatioPerGroup,
-                                           minSplitBits);
-    }
 
-}
+        private static Optional<Integer> maxGroupsAllowedDown(ModelElement tuning, boolean allowMoreThanOneContentGroupDown, int numberOfLeafGroups) {
+            var minGroupsUpRatio = tuning.childAsDouble("min-group-up-ratio");
+
+            if (minGroupsUpRatio != null) {
+                if (minGroupsUpRatio < 0.01 || minGroupsUpRatio > 1)
```
hakonhall, line 159: Why is  `0 < minGroupsUpRatio < 0.01`  disallowed?

config-model/src/main/resources/schema/content.rnc
```diff
@@ -82,7 +82,7 @@ ClusterControllerTuning = element cluster-controller {
     element stable-state-period {  xsd:string { pattern = "([0-9\.]+)\s*([a-z]+)?" }  }? &
     element min-distributor-up-ratio { xsd:double }? &
     element min-storage-up-ratio { xsd:double }? &
-    element max-groups-allowed-down { xsd:nonNegativeInteger }?
+    element min-group-up-ratio { xsd:double }?
```
hakonhall, line 85: Hm, I checked with Henning:  Both he and I intuitively preferred the negated field, i.e. something like `groups-allowed-down-ratio`.  Henning preferred to leave out a `max-` prefix, and that seems slightly better to me too.

hakonhall: But now we see min-group-up-ratio matches the two other min-X-up-ratio names :thinking: .  The semantics of those may be sufficiently different that it confuse more than help?   I think the current min-X-up-ratio fields cause the cluster to go down, and is unrelated to orchestration?

hmusum: Naming was done in same style as the other `min-X-up-ratio`, but I agree that `groups-allowed-down-ratio` is easier to understand and work with, I will change the name

config-model/src/main/java/com/yahoo/vespa/model/content/ClusterControllerConfig.java
hmusum: Will fix at the same time as changing the name

hakonhall (approved):

PR #27639, opened 2023-07-05
Support anonymized endpoints

mpolden:
@jonmv

FYI @andreer


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/routing/RoutingPolicies.java
```diff
@@ -402,7 +409,11 @@ var record = policy.canonicalName().isPresent() ?
 
     private void setPrivateDns(Endpoint endpoint, LoadBalancer loadBalancer, DeploymentId deploymentId) {
         if (loadBalancer.service().isEmpty()) return;
-        if (endpoint.isTokenEndpoint()) return;
+        boolean skipBasedOnAuthMethod = switch (endpoint.authMethod()) {
+            case token -> true;
+            case mtls -> false;
+        };
+        if (skipBasedOnAuthMethod) return;
```
jonmv, line 416: Do you know why this is done?

jonmv (approved):
Nice!

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/routing/RoutingPolicies.java
mpolden: No, but I can add a TODO to consider enabling all methods.

PR #27643, opened 2023-07-05
- Consolidate on isFilter.

baldersheim:
- Add has_weight_iterator to IDocumentWeightAttribute to allow fallback to bitvector.
- Allow filter attributes to enjoy IDirectWeightedSet optimization.

@geirst or @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
@@ -633,7 +633,11 @@ public:
                 return bitvector_iterator;
             }
         }
-        return std::make_unique<queryeval::DocumentWeightSearchIterator>(*tfmda[0], _attr, _dict_entry);
+        if (_attr.has_weight_iterator(_dict_entry.posting_idx)) {
+            return std::make_unique<queryeval::DocumentWeightSearchIterator>(*tfmda[0], _attr, _dict_entry);
+        } else {
+            return _attr.make_bitvector_iterator(_dict_entry.posting_idx, get_docid_limit(), *tfmda[0], strict);
+        }
     }
 
     SearchIteratorUP createFilterSearch(bool strict, FilterConstraint constraint) const override {
--- a/searchlib/src/vespa/searchlib/attribute/multinumericpostattribute.hpp
+++ b/searchlib/src/vespa/searchlib/attribute/multinumericpostattribute.hpp
@@ -146,12 +146,17 @@ MultiValueNumericPostingAttribute<B, M>::DocumentWeightAttributeAdapter::make_bi
 }
 
 template <typename B, typename M>
+bool
+MultiValueNumericPostingAttribute<B, M>::DocumentWeightAttributeAdapter::has_weight_iterator(vespalib::datastore::EntryRef idx) const noexcept
+{
+    return self.getPostingList().has_btree(idx);
+}
+
+template <typename B, typename M>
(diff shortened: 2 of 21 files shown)


toregge (approved):

geirst (approved):
👍

searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
```diff
@@ -633,7 +633,11 @@ class DirectAttributeBlueprint : public queryeval::SimpleLeafBlueprint
                 return bitvector_iterator;
             }
         }
-        return std::make_unique<queryeval::DocumentWeightSearchIterator>(*tfmda[0], _attr, _dict_entry);
+        if (_attr.has_weight_iterator(_dict_entry.posting_idx)) {
+            return std::make_unique<queryeval::DocumentWeightSearchIterator>(*tfmda[0], _attr, _dict_entry);
+        } else {
+            return _attr.make_bitvector_iterator(_dict_entry.posting_idx, get_docid_limit(), *tfmda[0], strict);
```
toregge, line 639: `tfmda[0]->populate_fixed()` is not called when using bitvector iterator
but it is called when using `DocumentWeightSearchIterator`. For a filter field that is used by ranking, this might matter, depending on what is used by the feature executor.

searchlib/src/vespa/searchlib/attribute/multinumericpostattribute.hpp
```diff
@@ -145,13 +145,18 @@ MultiValueNumericPostingAttribute<B, M>::DocumentWeightAttributeAdapter::make_bi
...
+}
+
 template <typename B, typename M>
 const IDocumentWeightAttribute *
 MultiValueNumericPostingAttribute<B, M>::asDocumentWeightAttribute() const
 {
-    if (this->hasWeightedSetType() &&
-        this->getBasicType() == AttributeVector::BasicType::INT64 &&
-        !this->getConfig().getIsFilter()) {
+    if (this->hasWeightedSetType() && (this->getBasicType() == AttributeVector::BasicType::INT64)) {
```
toregge, line 159: The `create()` member functions in `MultiValueNumericPostingAttribute` and `MultiValueStringPostingAttributeT` only operate with `DocumentWeightIterator` instances. This means that a posting list that only has a bitvector representation and not a btree representation will incorrectly be considered as an empty posting list. This is likely the cause for the observed regression.

searchlib/src/vespa/searchlib/attribute/multistringpostattribute.hpp
```diff
@@ -169,9 +176,7 @@ template <typename B, typename T>
 const IDocumentWeightAttribute *
 MultiValueStringPostingAttributeT<B, T>::asDocumentWeightAttribute() const
 {
-    if (this->hasWeightedSetType() &&
-        this->getBasicType() == AttributeVector::BasicType::STRING &&
-        !this->getConfig().getIsFilter()) {
+    if (this->hasWeightedSetType() && (this->getBasicType() == AttributeVector::BasicType::STRING)) {
```
toregge, line 179: Same issue here as with `MultiValueNumericPostingAttribute`

searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
baldersheim: Good point.

PR #27654, opened 2023-07-06
Restart also on internal redeployment if nodes change

jonmv:
@bratseth please review.


jonmv:
Could this be excessive, though?

jonmv:
I wonder if perhaps we should add this as a separate thing instead.

baldersheim (commented):
making a the rabbit hole more complicated is not the way to go.
Avoid it instead.

bratseth (approved):

PR #27662, opened 2023-07-06
Revert "- Use a single templated fixture to avoid copy-paste errors when exte…"

baldersheim:
Reverts vespa-engine/vespa#27653
@geirst PR


geirst (commented):
👍

PR #27666, opened 2023-07-06
Field name is used to lookup attribute and is hence the same.

baldersheim:
The attribute name already exist as long as the attribute, so hence no need to make a copy.

@geirst and @toregge PR


geirst (approved):
👍

PR #27667, opened 2023-07-06
Query Blueprints have a longer lifetime, hence there is no need to ta…

baldersheim:
…ke a copy of the readonly weights vector.

@toregge or @geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/dot_product_search.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/dot_product_search.cpp
@@ -4,30 +4,29 @@
 #include "iterator_pack.h"
 #include <vespa/vespalib/objects/visit.h>
 
-
 using search::fef::TermFieldMatchData;
 using search::fef::MatchData;
 using vespalib::ObjectVisitor;
 
 namespace search::queryeval {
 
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/queryeval/dot_product_search.cpp
```diff
@@ -27,7 +27,7 @@ class DotProductSearchImpl : public DotProductSearch
     };
 
     TermFieldMatchData     &_tmd;
-    std::vector<int32_t>    _weights;
+    const std::vector<int32_t> &_weights;
```
toregge, line 30: consider `vespalib::ConstArrayRef<int32_t>`

toregge (dismissed):

searchlib/src/vespa/searchlib/queryeval/dot_product_search.cpp
baldersheim: Any particular advantage ?

toregge: Fewer indirections when accessing the weights.

PR #27697, opened 2023-07-07
Add API for searching deployments by endpoint

mpolden:
Not fast, but better than nothing. I think we want this limited to operators
anyway.

@jonmv

### DIFF
--- a/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/search-deployments-single.json
+++ b/controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/search-deployments-single.json
@@ -6,6 +6,7 @@
       "instance": "instance1",
       "environment": "prod",
       "region": "us-central-1",
+      "cluster": "default",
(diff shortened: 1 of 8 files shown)


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/application/responses/search-deployments-single.json
```diff
@@ -0,0 +1,12 @@
+{
+  "deployments": [
+    {
+      "tenant": "tenant1",
+      "application": "application1",
+      "instance": "instance1",
+      "environment": "prod",
+      "region": "us-central-1",
```
mpolden, line 8: Maybe we should include the cluster name here as well.

jonmv: Yes. 

Also, tenant and application are always the same, but keeping all the attributes together may be fine just for simplicity anyway.

jonmv (commented):
Yeah, this will be nice, but you're right we should include cluster as well.

PR #27701, opened 2023-07-07
Bratseth/disk limit

bratseth:
The first two commits moves this to where it can be used also when making recommendations, and the third uses it for that.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
@@ -38,15 +38,10 @@ public class NodeResourceLimits {
             illegal(type, "diskGb", "Gb", cluster, requested.diskGb(), minAdvertisedDiskGb(requested, cluster.isExclusive()));
     }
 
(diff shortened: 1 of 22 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
```diff
@@ -35,6 +38,17 @@ public void ensureWithinAdvertisedLimits(String type, NodeResources requested, A
             illegal(type, "diskGb", "Gb", cluster, requested.diskGb(), minAdvertisedDiskGb(requested, cluster.isExclusive()));
     }
 
+    // TODO: Move this check into the above when we are ready to fail, not just warn on this. */
```
mpolden: Consider mentioning "Vespa 9" in this comment so that we don't forget it.

mpolden (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeResourceLimits.java
bratseth: I think we can do it before ...

PR #27702, opened 2023-07-07
Ignore json library

mpolden:
Not compatible with Go version (1.19) currently used by build systems.

Closes #27683

@bjorncs


bjorncs (approved):
Thanks 👍

PR #27708, opened 2023-07-07
Reapply "Less memory for default node resources in dev cd zone"

hmusum:
Reverts vespa-engine/vespa#27706

Please review only, will merge if this does not work out


bratseth (approved):

hmusum:
Not needed, closing

PR #27716, opened 2023-07-09
Allocate all groups in one go

bratseth:
With many groups and dynamic allocation allocating group by group is too slow.

This leaves a number of simplification opportunities untaken, but I wanted to keep this diff as small as possible.

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
@@ -22,16 +22,28 @@ public class ClusterMembership {
     private ClusterMembership(String stringValue, Version vespaVersion, Optional<DockerImage> dockerImageRepo,
                               ZoneEndpoint zoneEndpoint) {
         String[] components = stringValue.split("/");
-        if (components.length < 4)
+        if (components.length < 3)
             throw new RuntimeException("Could not parse '" + stringValue + "' to a cluster membership. " +
                                        "Expected 'clusterType/clusterId/groupId/index[/retired][/exclusive][/stateful][/combinedId]'");
+
+        Integer groupIndex = components[2].isEmpty() ? null : Integer.parseInt(components[2]);
+        Integer nodeIndex;
+        int missingElements = 0;
+        try {
+            nodeIndex = Integer.parseInt(components[3]);
+        } catch (NumberFormatException | ArrayIndexOutOfBoundsException e) {
+            // Legacy form missing the group component
+            nodeIndex = groupIndex;
+            groupIndex = null;
+            missingElements = 1;
+        }
 
         boolean exclusive = false;
         boolean stateful = false;
         var combinedId = Optional.<String>empty();
         boolean retired = false;
-        if (components.length > 4) {
-            for (int i = 4; i < components.length; i++) {
+        if (components.length > (4 - missingElements)) {
+            for (int i = (4 - missingElements); i < components.length; i++) {
                 String component = components[i];
                 switch (component) {
                     case "exclusive" -> exclusive = true;
@@ -44,7 +56,7 @@ public class ClusterMembership {
 
         this.cluster = ClusterSpec.specification(ClusterSpec.Type.valueOf(components[0]),
                                                  ClusterSpec.Id.from(components[1]))
-                                  .group(components[2].isEmpty() ? null : ClusterSpec.Group.from(Integer.parseInt(components[2])))
+                                  .group(groupIndex == null ? null : ClusterSpec.Group.from(groupIndex))
                                   .vespaVersion(vespaVersion)
                                   .exclusive(exclusive)
                                   .combinedId(combinedId.map(ClusterSpec.Id::from))
@@ -52,7 +64,7 @@ public class ClusterMembership {
                                   .loadBalancerSettings(zoneEndpoint)
                                   .stateful(stateful)
                                   .build();
-        this.index = Integer.parseInt(components[3]);
+        this.index = nodeIndex;
         this.retired = retired;
         this.stringValue = toStringValue();
     }
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
@@ -44,7 +44,7 @@ public class ClusterMembership {
 
         this.cluster = ClusterSpec.specification(ClusterSpec.Type.valueOf(components[0]),
                                                  ClusterSpec.Id.from(components[1]))
-                                  .group(ClusterSpec.Group.from(Integer.parseInt(components[2])))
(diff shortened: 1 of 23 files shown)


hmusum:
I am unable to review all of this today, so just a comment on of the changes: 

Cluster membership format is changed with this PR. Won't there be problems when upgrading, since the old code won't be able to read the new format? In that case, code for supporting the new cluster membership format needs to go in first, then the rest of the changes can be done.

bratseth:
I'm pretty sure it's ok since the specs without group were never returned out of the node repo before this PR - and could not be parsed at all since the constructor started with

```
        if (components.length < 4)
            throw new RuntimeException("Could not parse '" + stringValue + "' to a cluster membership. " +
                                       "Expected 'clusterType/clusterId/groupId/index[/retired][/exclusive][/stateful][/combinedId]'");

```

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
```diff
@@ -22,16 +22,28 @@ public class ClusterMembership {
     private ClusterMembership(String stringValue, Version vespaVersion, Optional<DockerImage> dockerImageRepo,
                               ZoneEndpoint zoneEndpoint) {
         String[] components = stringValue.split("/");
-        if (components.length < 4)
+        if (components.length < 3)
```
hmusum, line 25:
```suggestion
        if (components.length < 4)
```

hmusum:
Yes, right, that looks good

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
bratseth: I wanted to support everything this can output now ...

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -289,15 +286,28 @@ private Node acceptNode(NodeCandidate candidate, Retirement retirement, boolean
...
         return node;
     }
 
+    private boolean canBeUsedInGroupWithDeficiency(Node node) {
+        if (requestedNodes.count().isEmpty()) return true;
+        if (node.allocation().isEmpty()) return true;
+        var group = node.allocation().get().membership().cluster().group();
+        if (group.isEmpty()) return true;
+        long nodesInGroup = nodes.values().stream().filter(n -> groupOf(n).equals(group)).count();
+        return nodesInGroup < requestedNodes.count().get() / requestedNodes.groups();
```
freva, line 304: Not sure I understand this... Seems like we want to return true if the group the given `node` is in already has more nodes than what is requested, then this node can instead by reallocated to a different group? If so, then shouldn't the sign be reversed?

bratseth: The premise is: We cannot _change_ the group of a node we will keep (unless that group is to be removed), because that will lead to temporary coverage loss.

This method returns whether - given that - this node can be used in a group which has to few nodes. If the group the node belongs to has fewer nodes than it should, it can.

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/NodeListMicroBenchmarkTest.java
```diff
@@ -58,7 +58,6 @@ public void testChildrenOf() {
             nodeList.childrenOf(nodes.get(indexes.get(i)));
         }
         Duration duration = Duration.between(start, Instant.now());
-        System.out.println("Calling NodeList.childrenOf took " + duration + " (" + duration.toNanos() / iterations / 1000 + " microseconds per invocation)");
```
mpolden, line 61: This is the whole point of this (ignored) test, so it shouldn't be removed.

hmusum (dismissed):
Looks good to me, but a lot of changes and high risk, so maybe someone else should have a look as well

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
jonmv: So ideally, this should roll out first, since upgrades aren't atomic.

jonmv: Ah, well, no, the converse to this should roll out first, i.e, when an upgraded server writes a spec with `//`, the other servers should still be able to handle that meaningfully. They won't now, will they?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupIndices.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /** Attempt to unretire the given node if it is retired. */
+    private NodeCandidate unretire(NodeCandidate node) {
+        if (node.retiredNow()) return node;
+        if ( ! node.allocation().get().membership().retired()) return node;
+        if ( ! hasCompatibleResources(node) ) return node;
+        var parent = node.parentHostname().flatMap(hostname -> allNodes.node(hostname));
+        if (parent.isPresent() && (parent.get().status().wantToRetire() || parent.get().status().preferToRetire())) return node;
+        node = node.withNode();
+        if ( ! requested.isCompatible(node.resources()))
```
jonmv, line 145: Isn't it necessary to check the free parent capacity here, as well?

jonmv: I see that wasn't done before, so could do separately, if needed.

bratseth: We don't create candidates on hosts where there isn't capacity.

bratseth:
@jonmv, @freva, @mpolden let me know if any of you want to look more at this before I merge.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/GroupIndices.java
jonmv: Ah, that's good.

jonmv:
I don't want to look more at it, unless you need to let old configservers handle the format written with `//` by upgraded servers with the new cluster membership format ... ?

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterMembership.java
jonmv: Here.

bratseth:
... turns out it was easy to move storing of the reserved nodes to after assigning groups so I just did that.

mpolden (approved):
LGTM. Hopefully there are integration tests that use multiple groups?

bratseth:
There are.

PR #27722, opened 2023-07-10
Rewrite to iterative lock closing :'( MERGEOK

jonmv:
@bjorncs

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -1011,7 +1011,9 @@ public class Nodes {
         private static void close(Collection<NodeMutex> nodes) {
             RuntimeException thrown = null;
             for (NodeMutex node : reversed(List.copyOf(nodes))) {
-                try (node) { }
+                try {
+                    node.close();
+                }
                 catch (RuntimeException e) {
                     if (thrown == null) thrown = e;
                     else thrown.addSuppressed(e);
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
@@ -35,7 +35,6 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.NavigableSet;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -1002,15 +1001,23 @@ private NodeMutexes lockAndGetAll(Collection<Node> nodes, Optional<Duration> tim
...
 
     /** A node with their locks, acquired in a universal order. */
     public record NodeMutexes(List<NodeMutex> nodes) implements AutoCloseable {
-        @Override public void close() { close(nodes.iterator()); }
-        private static void close(Iterator<NodeMutex> nodes) {
-            if (nodes.hasNext()) try (NodeMutex node = nodes.next()) { close(nodes); }
+        @Override public void close() { close(nodes); }
+        private static void close(Iterable<NodeMutex> nodes) {
+            RuntimeException thrown = null;
+            for (NodeMutex node : nodes) {
```
jonmv, line 1013: This changes lock release order, but that shouldn't matter. 

Actually ... come to think of it, there were some problems with instrumentation some time ago, when this was the case. I'm pretty sure that fixed, but can reverse the order here, just in case.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -1002,15 +1001,23 @@ private NodeMutexes lockAndGetAll(Collection<Node> nodes, Optional<Duration> tim
...
+                try (node) { }
```
mpolden, line 1014: `node.close()` would be clearer I think.

mpolden (dismissed):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Nodes.java
```diff
@@ -1002,15 +1001,23 @@ private NodeMutexes lockAndGetAll(Collection<Node> nodes, Optional<Duration> tim
...
     /** A node with their locks, acquired in a universal order. */
     public record NodeMutexes(List<NodeMutex> nodes) implements AutoCloseable {
-        @Override public void close() { close(nodes.iterator()); }
-        private static void close(Iterator<NodeMutex> nodes) {
-            if (nodes.hasNext()) try (NodeMutex node = nodes.next()) { close(nodes); }
+        @Override public void close() { close(nodes); }
+        private static void close(Collection<NodeMutex> nodes) {
+            RuntimeException thrown = null;
+            for (NodeMutex node : reversed(List.copyOf(nodes))) {
+                try (node) { }
```
jonmv, line 1014:
```suggestion
                try {
                    node.close();
                }
```

mpolden (approved):

PR #27723, opened 2023-07-10
Fix memory leak in vespa-index-inspect.

toregge:
This fixes a regression introduced in #27704


vekterli (approved):
:+1:

toregge:
Obsoleted by #27726

PR #27726, opened 2023-07-10
Change createIterator member functions to return unique ptr to

toregge:
search iterator.

@vekterli : please review


vekterli (approved):
:+1:

PR #27739, opened 2023-07-11
Revert "- Enable use of optimized DirectWeightedAttribute for all integer att…"

toregge:
Reverts vespa-engine/vespa#27645

Broke range search on weightedset fields.


vekterli (approved):
:+1:

PR #27741, opened 2023-07-11
Jonmv/reconfigure dispatch

jonmv:
@bjorncs please review, but do not merge! I'll add some tests, and possibly change things a bit more before I'm done. 

I think it would have been easier to refactor all of this if the code style was more that of returning immutable snapshots of state, rather than passing around and using stateful objects—in particular the `SearchCluster` would benefit from updating a snapshot on every ping completion, and passing that out to users, instead of passing out something that possibly updates mid-flight. I'm not sure it's worth the effort to make these changes. The only observable changes are related to coverage, and the dispatcher looks robust w.r.t. that. 

I opted for closing down the cluster monitor for the above reason, as well as the load balancer—both hold only auxiliary state, that is quickly re-computed. The state held in the `SearchGroups`, however, is more valuable to keep. I also chose not to wait for all new nodes to have known status in the reconfiguration update. I believe the behaviour of the cluster with or without known state for these is identical, assuming the nodes are actually new, and not yet loaded with documents—furthermore, once the node set of the search cluster is updated, the observable changes (coverage) may be effected immediately anyway, again because of the above, so waiting doesn't really make sense.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
@@ -21,8 +21,11 @@ public abstract class CloseableInvoker implements Closeable {
     private RequestDuration duration;
 
     public void teardown(BiConsumer<Boolean, RequestDuration> teardown) {
-        this.teardown = teardown;
-        this.duration = new RequestDuration();
+        this.teardown = this.teardown == null ? teardown : (success, duration) -> {
+            this.teardown.accept(success, duration);
+            teardown.accept(success, duration);
+        };
+        this.duration = this.duration == null ? new RequestDuration() : this.duration;
     }
 
     protected void setFinalStatus(boolean success) {
--- a/container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
@@ -2,7 +2,6 @@
 package com.yahoo.search.dispatch;
 
 import java.io.Closeable;
-import java.time.Duration;
 import java.util.function.BiConsumer;
 
 /**
@@ -21,10 +20,7 @@ public abstract class CloseableInvoker implements Closeable {
     private RequestDuration duration;
 
     public void teardown(BiConsumer<Boolean, RequestDuration> teardown) {
-        this.teardown = this.teardown == null ? teardown : (success, duration) -> {
-            this.teardown.accept(success, duration);
-            teardown.accept(success, duration);
-        };
+        this.teardown = this.teardown == null ? teardown : this.teardown.andThen(teardown);
         this.duration = this.duration == null ? new RequestDuration() : this.duration;
     }
 
--- a/container-search/src/main/java/com/yahoo/search/dispatch/Dispatcher.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/Dispatcher.java
@@ -12,6 +12,7 @@ import com.yahoo.search.Query;
 import com.yahoo.search.Result;
 import com.yahoo.search.cluster.ClusterMonitor;
 import com.yahoo.search.dispatch.SearchPath.InvalidSearchPathException;
+import com.yahoo.search.dispatch.rpc.RpcConnectionPool;
 import com.yahoo.search.dispatch.rpc.RpcInvokerFactory;
 import com.yahoo.search.dispatch.rpc.RpcPingFactory;
 import com.yahoo.search.dispatch.rpc.RpcResourcePool;
@@ -55,8 +56,9 @@ public class Dispatcher extends AbstractComponent {
     /** If set will control computation of how many hits will be fetched from each partition.*/
     public static final CompoundName topKProbability = CompoundName.from(DISPATCH + "." + TOP_K_PROBABILITY);
 
+    private final InvokerFactoryFactory invokerFactories;
     private final DispatchConfig dispatchConfig;
-    private final RpcResourcePool rpcResourcePool;
+    private final RpcConnectionPool rpcResourcePool;
     private final SearchCluster searchCluster;
     private volatile VolatileItems volatileItems;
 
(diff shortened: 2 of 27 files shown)


jonmv:
@baldersheim you're very welcome to have a look as well!

container-search/src/main/java/com/yahoo/search/dispatch/Dispatcher.java
```diff
@@ -103,6 +127,38 @@ public Dispatcher(ComponentId clusterId, DispatchConfig dispatchConfig,
...
+     * 3. When inflight operations against the old snapshot are done, all obsolete resources are cleaned up.
+     */
+    void updateWithNewConfig(DispatchNodesConfig nodesConfig) {
+        try (var items = volatileItems()) { // Replace the volatile items snapshot, marking a reference to the old one.
+            items.get().countDown();        // Decrement for the initial creation reference.
+            this.volatileItems = update(null);
+
+            // Set up the cleanup that we need to do.
+            Collection<? extends AutoCloseable> connectionPoolsToClose = rpcResourcePool.updateNodes(nodesConfig);
+            items.get().cleanup = () -> {
```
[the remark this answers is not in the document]
jonmv: That can't happen, because we're holding one reference ourselves here, so the snapshot can't get to zero references.

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -55,13 +58,18 @@ public SearchCluster(String clusterId, SearchGroupsImpl groups, VipStatus vipSta
...
     @Override
     public String name() { return clusterId; }
-    public VipStatus getVipStatus() { return vipStatus; }
+
+    public void updateNodes(Collection<Node> newNodes, double minActivedocsPercentage) {
+        Collection<Node> retainedNodes = groups.nodes();
+        Collection<Node> currentNodes = new HashSet<>(newNodes);
+        retainedNodes.retainAll(currentNodes);
+        currentNodes.removeIf(retainedNodes::contains);
+        currentNodes.addAll(retainedNodes);
```
bjorncs, line 67: What does this do?

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
```diff
@@ -107,14 +115,14 @@ public int groupsWithSufficientCoverage() {
...
-        if ( localCorpusDispatchTarget == null) return Optional.empty();
+        if (localCorpusDispatchTarget == null) return Optional.empty();
 
         // Only use direct dispatch if the local group has sufficient coverage
         Group localSearchGroup = groups.get(localCorpusDispatchTarget.group());
         if ( ! localSearchGroup.hasSufficientCoverage()) return Optional.empty();
 
         // Only use direct dispatch if the local search node is not down
-        if ( localCorpusDispatchTarget.isWorking() == Boolean.FALSE) return Optional.empty();
+        if (localCorpusDispatchTarget.isWorking() == Boolean.FALSE) return Optional.empty();
```
bjorncs, line 125: No idea why a boxed boolean is necessary here... 😄

container-search/src/main/java/com/yahoo/search/dispatch/Dispatcher.java
```diff
@@ -57,16 +58,39 @@ public class Dispatcher extends AbstractComponent {
...
             this.loadBalancer = loadBalancer;
             this.invokerFactory = invokerFactory;
+            this.clusterMonitor = clusterMonitor;
+        }
+
+        private void countDown() {
+            if (inflight.decrementAndGet() == 0) cleanup.run();
+        }
+
+        private class Ref implements AutoCloseable {
```
bjorncs, line 81: Consider using resouce/reference concepts from jdisc_core. Take a look at `OnnxRuntime` and `ReferencedOrtSession`

container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
```diff
@@ -21,8 +20,8 @@ public abstract class CloseableInvoker implements Closeable {
     private RequestDuration duration;
 
     public void teardown(BiConsumer<Boolean, RequestDuration> teardown) {
-        this.teardown = teardown;
-        this.duration = new RequestDuration();
+        this.teardown = this.teardown == null ? teardown : this.teardown.andThen(teardown);
```
bjorncs, line 23: Consider using `CompletableFuture`. Does the exception semantics for `andThen()` fit here?

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
jonmv: Replaces the newly created, information-less nodes, with the ones that we have updated with information from the previous monitor.

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchCluster.java
jonmv: yes / no / dunno :)

container-search/src/main/java/com/yahoo/search/dispatch/CloseableInvoker.java
jonmv: Yes, trying not to change too much at once. There shouldn't be any exceptions here, all we do is update a few numbers, but ... yes, we could handle a potential exception, indeed.

container-search/src/main/java/com/yahoo/search/dispatch/Dispatcher.java
jonmv: Yes, I just wanted this to be as simple as possible, but I'll take a look at that and see if it's better.

bjorncs (dismissed):

jonmv:
Superseded by #27770

PR #27742, opened 2023-07-11
Remove unused config from dependency list

jonmv:
This removes the false dependency on config that was at one time intended to be injected and used, but this was forgotten. I'm not sure we want to do that anyway, now, but before we have that discussion, let's just remove the unnecessary config, and graph reconstructions this causes.


jonmv:
@bratseth feel free to review.

bratseth (approved):

vekterli (commented):
:+1: I can't even remember seeing that particular config before... Must have been Very Important™

PR #27745, opened 2023-07-12
Bratseth/provisioning cleanup

bratseth:
No functional changes. 
Reviewing each commit should make the changes easy to follow.


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -54,34 +53,19 @@ public List<Node> prepare(ApplicationId application, ClusterSpec cluster, NodeSp
...
-        List<Node> accepted = groupPreparer.prepare(application, cluster, requested, surplusNodes, allNodes);
+        List<Node> accepted = groupPreparer.prepare(application, cluster, requested, allNodes);
         if (requested.rejectNonActiveParent()) { // TODO: Move to NodeAllocation
             NodeList activeHosts = allNodes.state(Node.State.active).parents().nodeType(requested.type().hostType());
             accepted = accepted.stream()
                                .filter(node -> node.parentHostname().isEmpty() || activeHosts.parentOf(node).isPresent())
                                .toList();
         }
-        return accepted.stream().filter(node -> ! surplusNodes.contains(node)).collect(Collectors.toList());
+        return new ArrayList<>(accepted);
```
mpolden: Ok, but in general I think it's cleaner to pass around immutable collections and create mutable ones in the same place it's mutated.

mpolden (approved):
Nice!

PR #27749, opened 2023-07-12
Expire deprovisioned nodes in test zones earlier

hmusum:


jonmv (dismissed):

jonmv:
Perhaps 1 day is a bit short, though? Should it be 3?

mpolden (approved):

PR #27750, opened 2023-07-12
Wait by default in deploy commands

mpolden:
This also makes `--wait` non-global and it is now only declared for commands
where it is useful (activate, curl, deploy, document, feed, query, status, test,
visit).

@bratseth


mpolden:
Needs some more work.

mpolden:
PTAL. A side-effect of this change is that "wait" cannot be stored in config since it's no longer a global option.

bratseth (approved):

PR #27755, opened 2023-07-13
Replace hash based node cache with ZK node versino based

jonmv:
@mpolden please review and merge. This is the same construct as in the controller.


mpolden (approved):
👍 

We tried this once before, but a lot of the surrounding parts have changed since then. Let's try it again.

jonmv:
Yes, that was when nodes didn't have a stable path. That change should allow this :)

PR #27756, opened 2023-07-13
Remove DNS records for zones that use A records

mpolden:
GCP uses A records instead of CNAME...

@jonmv


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/routing/RoutingPolicies.java
```diff
@@ -395,8 +394,8 @@ private void updateZoneDnsOf(RoutingPolicy policy, LoadBalancer loadBalancer, De
         RoutingMethod routingMethod = controller.zoneRegistry().routingMethod(deploymentId.zoneId());
         boolean addTokenEndpoint = controller.routing().tokenEndpointEnabled(deploymentId.applicationId());
         for (var endpoint : policy.zoneEndpointsIn(controller.system(), routingMethod, addTokenEndpoint)) {
-            var name = RecordName.from(endpoint.dnsName());
-            var record = policy.canonicalName().isPresent() ?
+            RecordName name = RecordName.from(endpoint.dnsName());
+            Record record = policy.canonicalName().isPresent() ?
```
jonmv, line 398: Hehe. Yes, I've decided I usually prefer the types as well :)

jonmv (approved):

jonmv:
Good.

PR #27765, opened 2023-07-13
Balder/refactor query building 1

baldersheim:
@toregge PR
The last commit here prevents incorrectly assuming that a string term can not be a range term.
This is the reason why #27645 had to reverted. Even with the revert it is incorrect for ws<long>.
This should fix the root cause.


toregge (approved):
👍

baldersheim:
The issue with + and whitespace must be fixed before adding #27645 back.
The latter was the main intention, and should be fine.

baldersheim:
@toregge Any reason for the no merge label still ?

toregge:
No. @vekterli: FYI

Issue #27768, opened 2023-07-13
Vespa  container crash during startup


107dipan:
**Describe the bug**
We have a single node  vespa environment which failed to come up after restart. 
Could you us understanding what the issue could be

[2023-07-13 14:45:13.697] INFO    container-clustercontroller Container.com.yahoo.vespa.clustercontroller.core.EventLog Added node only event: Event: storage.0: Node got back into slobrok with same address as before: tcp/d190259-356.dc.gs.com:19106
[2023-07-13 14:45:13.701] INFO    searchnode       proton.status        Starting status web server on port 19107.
[2023-07-13 14:45:13.999] WARNING searchnode       stderr       /builddir/build/BUILD/vespa-8.137.25/searchlib/src/vespa/searchlib/diskindex/pagedict4file.cpp:28: Abort called. Reason: should not be reached
[2023-07-13 14:45:14.000] WARNING searchnode       stderr       terminate called after throwing an instance of 'std::runtime_error'
[2023-07-13 14:45:14.000] WARNING searchnode       stderr         what():  Fatal: Reading 18446744073709551615 bytes, got 0 from '/opt/vespa/var/db/vespa/search/cluster.vespadoc/n0/documents/vespaDocumentType/0.ready/index/index.fusion.12/docStorageMode/boolocc.idx' failed: short read
[2023-07-13 14:45:14.001] WARNING searchnode       stderr       terminate called recursively
[2023-07-13 14:45:14.001] WARNING searchnode       stderr       terminate called recursively
[2023-07-13 14:45:14.001] WARNING searchnode       stderr       terminate called recursively
[2023-07-13 14:45:14.000] ERROR   searchnode       proton.diskindex.pagedict4file       Could not open /opt/vespa/var/db/vespa/search/cluster.vespadoc/n0/documents/`vespaDocumentType/0.ready/index/index.fusion.12/accountNumber_s/dictionary.ssdat for write: error=24("Too many open files") fileStat[name=/opt/vespa/var/db/vespa/search/cluster.vespadoc/n0/documents/vespaDocumentType/0.ready/index/index.fusion.12/accountNumber_s/dictionary.ssdat errno=2("No such file or directory")] dirStat[name=/opt/vespa/var/db/vespa/search/cluster.vespadoc/n0/documents/vespaDocumentType/0.ready/index/index.fusion.12/accountNumber_s mode=40755 uid=1000 gid=1000 size=57 mtime=1689259512]
[...]

PR #27769, opened 2023-07-13
Fail when unable to open file.

toregge:
The confusing error message (`Fatal: Reading 18446744073709551615 bytes, got 0`) shown in #27768 was caused
by missing error checking when writing bitvector files.


vekterli (approved):
:+1:

PR #27772, opened 2023-07-14
Revert "- Pack data closer to let config fit in 2 cache lines instead of 4."

toregge:
Reverts vespa-engine/vespa#27646

This revert is needed in order to revert #27643


vekterli (approved):
:+1:

PR #27773, opened 2023-07-14
Revert "- Consolidate on isFilter."

toregge:
Reverts vespa-engine/vespa#27643

Observed regression in 8.191.40 was not fixed by reverting #27645.


vekterli (approved):
:+1:

PR #27777, opened 2023-07-14
Use config self-subscription also for session-cache which powers docproc [run-systemtest]

jonmv:
@vekterli please review.


jonmv:
This makes the setup here similar to that used by the `/document/v1` handler, and more similar to what it was earlier. I wonder if we should also try to reconfigure the RPC layer when the config for that changes, but that's a different PR.

jonmv:
@bratseth @baldersheim FYI.

vekterli (approved):
:+1: _the reconfig is coming from inside the house_

PR #27790, opened 2023-07-14
Use std::filesystem::remove in transaction log.

toregge:
@vekterli : please review


vekterli (approved):
:+1: I see some code uses `_fileName` and some uses `_transLog->GetFileName()`, but I presume the latter is mostly as a convenience to get a raw pointer

PR #27792, opened 2023-07-16
Check redundancy also for groups that are up

hmusum:


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -396,26 +421,55 @@ private Result checkIfNodesAreUpOrRetired(ClusterState clusterState) {
...
-                            + ", but we require at least " + requiredRedundancy);
-                } else {
-                    return allow();
-                }
+        Integer minReplication = minReplication(distributorNodeInfo).get(node.getIndex());
+        return verifyRedundancy(distributorNodeInfo, minReplication, node.getIndex());
+    }
+
+    private Result checkRedundancySeenFromDistributor(DistributorNodeInfo distributorNodeInfo, Set<Integer> indexesToCheck) {
+        Map<Integer, Integer> replication = new LinkedHashMap<>(minReplication(distributorNodeInfo));
```
vekterli, line 429: Do we need to copy the returned map here?

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -196,86 +209,98 @@ private static Result checkIfStateSetWithDifferentDescription(NodeInfo nodeInfo,
...
+
+    private Optional<Result> checkRedundancyForGroupsThatAreUp(Set<Integer> retiredAndNotUpGroups) {
+        Set<Integer> groupsThatAreUp = groupsThatAreUp(retiredAndNotUpGroups);
+        log.log(FINE, "Check min replication for groups " + groupsThatAreUp);
+        for (int group : groupsThatAreUp) {
+            List<ConfiguredNode> nodesInGroup = getNodesInGroup(group);
+            for (var n : nodesInGroup) {
+                log.log(FINE, "Check min replication for index " + n.index() + " in group " + group);
+                Set<Integer> indexesToCheck = nodesInGroup.stream().map(ConfiguredNode::index).collect(Collectors.toSet());
+                var r = checkRedundancySeenFromDistributor(clusterInfo.getDistributorNodeInfo(n.index()), indexesToCheck);
```
vekterli, line 290: Are we guaranteed to check _all_ possible distributors that can have knowledge of buckets in the `indexesToCheck` set? Since we prune nodes from groups that are retired/not up. Bucket ownership and management happens _across_ groups and retired-status only applies to content nodes and not distributors, so I'm wondering if we'll risk skipping looking at the corresponding distributors on retired nodes. E.g. if we have 4 groups and 1 has all retired content nodes, 25% of the buckets will still be owned by (non-retired) distributors in that group.

Perhaps we should instead loop over all distributors that are in an available state rather than doing this by group?

vekterli (dismissed):
:+1: primarily have a question related to aggregation of replication stats and whether we will always check all relevant distributors

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -396,26 +421,55 @@ private Result checkIfNodesAreUpOrRetired(ClusterState clusterState) {
...
+            return disallow("Distributor " + distributorNodeInfo.getNodeIndex()
+                                    + " says storage node " + minReplicationIndex
+                                    + " has buckets with redundancy as low as "
+                                    + minReplication + ", but we require at least " + requiredRedundancy);
+        }
+
         return allow();
     }
 
+    private Map<Integer, Integer> minReplication(DistributorNodeInfo distributorNodeInfo) {
```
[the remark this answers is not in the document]
hmusum: Yes, that would be nice

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
hmusum: No, can't see why that's needed

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
hmusum: Yes, you might right about that, will look into it

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -153,16 +156,22 @@ private Result canSetStateMaintenanceTemporarily(StorageNodeInfo nodeInfo, Clust
         if (result.notAllowed())
             return result;
 
-        if (maxNumberOfGroupsAllowedToBeDown == -1) {
-            result = checkIfAnotherNodeInAnotherGroupHasWantedState(nodeInfo);
+        if (isGroupedSetup()) {
+            if (maxNumberOfGroupsAllowedToBeDown == - 1) {
```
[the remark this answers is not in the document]
hmusum: Fixed

clustercontroller-core/src/test/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeCheckerTest.java
```diff
@@ -215,6 +216,8 @@ void testMaintenanceAllowedFor2Of4Groups() {
 
     @Test
     void testMaintenanceAllowedFor2Of4Groups8Nodes() {
+        // LogSetup.initVespaLogging("foo");
```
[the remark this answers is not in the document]
hmusum: Yup, removed

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
hmusum: @vekterli Gave it a try, please have a look again

vekterli (approved):
:+1:

PR #27803, opened 2023-07-17
Require certificate files specified in environment

mpolden:
I suspect the configured files do not exist when the test runs in CD. This will
make the CLI print something useful if that's the case.

@bjorncs or @hmusum


hmusum (dismissed):

bjorncs (dismissed):

bjorncs (commented):
👍

PR #27810, opened 2023-07-18
Implement Levenshtein DFA with successor string generation

vekterli:
@toregge please review. A lot of the line count is taken up by tests and comments.

This implements code for building and evaluating Levenshtein Deterministic Finite Automata, where the resulting DFA efficiently matches all possible source strings that can be transformed to the target string within _k_ max edits. This allows for _O(n)_ matching of strings. We currently support _k_ in {1, 2} (extensible to higher numbers if we want it).

Additionally, when matching using a DFA, in the case where the source string does _not_ match, we can generate the _successor_ string; the next matching string that is lexicographically _greater_ than the source string. This string has the invariant that there are no possibly matching strings within _k_ edits ordered after the source string but before the successor.

This lets us do possibly massive leaps forward in an ordered dictionary, turning a scan for matches into a sub-linear operation.

Matching and successor generation is fully Unicode-aware. All input strings are expected to be in UTF-8 (without nulls), and the generated successor is also encoded as UTF-8. Internally, matching is done on UTF-32 code points and the DFA itself is built around UTF-32. UTF-8 decoding of source strings is done in a streaming fashion and does not require any allocations.

This commit includes a templated core Levenshtein DFA matching (and successor generation) algorithm and two separate DFA implementations that can be used; one explicit and one implicit.

The explicit DFA is an immutable DAG built up-front that represents all DFA states and transitions as explicit nodes and edges in a graph. This is currently the fastest to evaluate, but the build time and memory usage means its usage should be preferred for shorter strings (up to a few hundred chars).

The implicit DFA does not build any graph up-front, but rather evaluates state transitions on-demand for any given source string. This is currently slower than the explicit DFA, but its _O(1)_ memory usage (aside from the memory used by the target string itself) means that it can be used for arbitrary string lengths.

This code currently exists as a freestanding vespalib utility, and is not yet wired to any production code (fuzzy matching or similar).

Future optimizations:
 * Redesign sparse state representation and stepping logic to require much less branching, in turn making the code much less likely to stall the CPU pipeline.
 * Emit as much as possible of the successor string suffix by copying directly from the target string UTF-8 representation instead of following the DFA and encoding UTF-32 to UTF-8 chars.


toregge (commented):
👍

PR #27812, opened 2023-07-18
Fix credentials paths for hosted

mpolden:
@bjorncs

### DIFF
--- a/vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
+++ b/vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
@@ -6,6 +6,8 @@ import com.yahoo.component.annotation.Inject;
 import com.yahoo.slime.Cursor;
 import com.yahoo.slime.Slime;
 import com.yahoo.slime.SlimeUtils;
+import com.yahoo.vespa.athenz.api.AthenzIdentity;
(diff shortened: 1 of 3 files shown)


vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/VespaCliTestRunner.java
```diff
@@ -126,14 +132,35 @@ ProcessBuilder testRunProcessBuilder(Suite suite, TestConfig config) throws IOEx
...
+        final Path privateKeyFile;
+        final Path certificateFile;
+        if (config.system().isPublic()) {
+            privateKeyFile = artifactsPath.resolve("key");
+            certificateFile = artifactsPath.resolve("cert");
+        } else {
+            Path siaRoot = vespaHome.resolve("var/vespa/sia");
+            List<AthenzIdentity> services = SiaUtils.findSiaServices(siaRoot);
+            if (services.isEmpty()) {
+                throw new IllegalArgumentException("No service credentials in " + siaRoot + ". Application has no " +
```
bjorncs, line 153: Will exceptions thrown here be visible for end-user, or should failure be added to log/test report?

mpolden: This is called inside `runTests`, which catches `Exception` and creates a log report for it.

bjorncs (approved):

PR #27819, opened 2023-07-19
Don't throw InvalidApplicationException when there are node allocatio…

hmusum:
…n issues


bratseth:
👍 but could you use Exceptions.toMessageString(e)?

hmusum:
Yes, fixed

bratseth (approved):

PR #27823, opened 2023-07-19
- Add noexcept and some constexpr.

baldersheim:
- Use BitWord as helper class instead of inheriting in many static methods.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/common/bitword.cpp
+++ b/searchlib/src/vespa/searchlib/common/bitword.cpp
@@ -22,6 +22,6 @@ BitWord::Init::Init()
     fillUp(BitWord::_checkTab, std::numeric_limits<BitWord::Word>::max());
 }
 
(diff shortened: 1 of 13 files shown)


searchlib/src/vespa/searchlib/common/bitword.cpp
```diff
@@ -22,6 +22,6 @@ BitWord::Init::Init()
     fillUp(BitWord::_checkTab, std::numeric_limits<BitWord::Word>::max());
 }
 
-BitWord::Word BitWord::_checkTab[BitWord::WordLen];
+constexpr BitWord::Word BitWord::_checkTab[BitWord::WordLen];
```
toregge, line 25: Consider removing this added constexpr. `BitWord::_checkTab` is modified by `BitWord::Init::Init()`.

toregge (dismissed):

toregge (approved):

PR #27830, opened 2023-07-19
We only need the FieldSpecBase. No need to carry and copy the name ar…

baldersheim:
…ound.

@toregge PR


baldersheim:
@toregge If you see something risky, we hold it until holiday is over.

toregge (approved):
:+1: @vekterli : FYI

PR #27840, opened 2023-07-20
OS downgrades at system level

mpolden:
With this we will support changing the target version for a system to a past
version, and optionally pinning the system to that version. This is to simplify
manually stopping upgrades/rolling back.

Currently, a given zone won't downgrade existing active hosts to a lower
version, but any pending upgrades will be canceled and new hosts will use the
lower version. I couldn't think of a scenario where we'd want to downgrade
working hosts, so leaving it as-is for now.

No functional changes in the last commit.

Merge with internal PR.

@bratseth


bratseth (approved):
👍

PR #27846, opened 2023-07-20
Remove vespalib::unlink.

toregge:
@vekterli : please review

### DIFF
--- a/vespalib/src/vespa/vespalib/io/fileutil.cpp
+++ b/vespalib/src/vespa/vespalib/io/fileutil.cpp
@@ -433,7 +433,7 @@ bool
 File::unlink()
 {
     close();
-    return vespalib::unlink(_filename);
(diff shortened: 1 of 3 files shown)


vespalib/src/vespa/vespalib/io/fileutil.cpp
```diff
@@ -433,7 +433,7 @@ bool
 File::unlink()
 {
     close();
-    return vespalib::unlink(_filename);
+    return std::filesystem::remove(std::filesystem::path(_filename));
```
vekterli, line 436: Since the legacy code throws our internal `IoException` exception type, we might want to do a sweep of our code to see if we have any explicit `catch (vespalib::IoException)` or `catch (vespalib::Exception)` blocks for graceful IO error handling anywhere that should be upgraded to use the stdlib exception types.

vekterli (approved):
:+1:

PR #27852, opened 2023-07-20
Rename fsa library to vespafsa.

toregge:
@aressem : please review


aressem (approved):

aressem (changes requested):
I just discovered that we have the same issue with `libfsamanagers.so` so we should rename that as well.

aressem (approved):

PR #27854, opened 2023-07-20
Don't scale down if we are likely to scale back up

bratseth:
When we decide to scale number of nodes down we'll see a lower ideal load because we need to be able to handle one node going down. This may lead us to be closer to ideal (at current peak load) than otherwise anticipated, such that we are quite likely to soon scale back up. This checks for that and avoids scaling down dimensions where this is the case.


mpolden (approved):
👍

PR #27866, opened 2023-07-21
Avoid dereferencing empty unique pointer.

toregge:
@vekterli : please review


vekterli (approved):
:+1:

bjormel:
@toregge , can you share some details? 
* What caused the dereferencing of empty unique pointer?
* Is it safe to reapply the application change that triggered this, after this fix has been applied?

PR #27883, opened 2023-07-25
Prefer std::filesystem::exists over FastOS_StatInfo

baldersheim:
@vekterli or @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/diskindex/fieldreader.cpp
+++ b/searchlib/src/vespa/searchlib/diskindex/fieldreader.cpp
@@ -5,7 +5,6 @@
 #include "extposocc.h"
 #include "pagedict4file.h"
 #include "field_length_scanner.h"
-#include <vespa/vespalib/util/error.h>
 #include <filesystem>
 
 #include <vespa/log/log.h>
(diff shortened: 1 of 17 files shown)


searchlib/src/vespa/searchlib/diskindex/fieldreader.cpp
```diff
@@ -124,11 +126,9 @@ FieldReader::open(const vespalib::string &prefix,
                   const TuneFileSeqRead &tuneFileRead)
 {
     vespalib::string name = prefix + "posocc.dat.compressed";
-    FastOS_StatInfo statInfo;
 
-    bool statres = FastOS_File::Stat(name.c_str(), &statInfo);
-    if (!statres) {
-        LOG(error, "Could not stat compressed posocc file %s: %s", name.c_str(), getLastErrorString().c_str());
+    if (!fs::exists(fs::path(name))) {
+        LOG(error, "Compressed posocc file %s does not exist: %s", name.c_str(), getLastErrorString().c_str());
```
vekterli, line 131: Will the output of `getLastErrorString()` (i.e. `errno`) make sense here, or should we just skip it? Since this is the non-`std::error_code` version of `exists()` it will throw on OS errors.

vekterli (dismissed):
:+1:

searchlib/src/vespa/searchlib/diskindex/fieldreader.cpp
baldersheim: Fixed

vekterli (approved):

PR #27886, opened 2023-07-25
Require certification of scheduled OS upgrades

mpolden:
Factory will notify the controller about certified versions based on status in
CD systems.

@bratseth


bratseth (approved):
Nice

PR #27887, opened 2023-07-25
Bratseth/scale down less 2

bratseth:
Just a nicer solution to the problem solved by https://github.com/vespa-engine/vespa/pull/27854


mpolden (approved):
👍

PR #27898, opened 2023-07-26
Make generated automatic feed block error messages more user-friendly

vekterli:
@hakonhall please review
@kkraune please review quality™️ of new message syntax

Feed block messages are generated centrally by the cluster controller and pushed to content nodes as part of a cluster state bundle; the distributors nodes merely repeat back to the clients what they have been told. This changes the cluster controller feed block error message code to be less ambiguous and to include a URL to our public documentation about feed blocks.

Example of _old_ message:
```
disk on node 1 [storage.1.local] (0.510 > 0.500)
```
Same feed block with _new_ message:
```
disk on node 1 [storage.1.local] is 51.0% full (the configured limit is 50.0%).
See https://docs.vespa.ai/en/operations/feed-block.html
```

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeResourceExhaustion.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeResourceExhaustion.java
@@ -47,18 +47,19 @@ public class NodeResourceExhaustion {
     }
 
     public String toExhaustionAddedDescription() {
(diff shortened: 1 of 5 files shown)


hakonhall (approved):

hakonhall:
Mention which content cluster this is in the message?

vekterli:
> Mention which content cluster this is in the message?

Will add this in a follow-up since this information isn't part of the message today

PR #27903, opened 2023-07-26
controller tls always use latest cert/key pair

andreer:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/tls/ControllerSslContextFactoryProvider.java
```diff
@@ -65,17 +72,34 @@ private TlsContext createTlsContext(int port) {
                 port != 443 ? PeerAuthentication.WANT : PeerAuthentication.DISABLED);
     }
 
+    record TlsCredentials(List<X509Certificate> certificates, PrivateKey privateKey){}
+
+    private static TlsCredentials latestValidCredentials(SecretStore secretStore, TlsConfig tlsConfig) {
+        int version = latestVersionInSecretStore(secretStore, tlsConfig).orElseThrow();
```
mpolden, line 78: It's better to just let the `SecretNotFoundException` be thrown as it's more descriptive than `orElseThrow()`.

mpolden (approved):

bjorncs (commented):
👍

PR #27924, opened 2023-07-28
Allow cloud, environment, and zone in flag data conditions

hakonhall:
This PR adds the cloud and environment dimensions and allows setting the cloud, environment, and zone dimensions in flag data.  The typical use-case will be for users to change the system-wide flag data file in hosted-feature-flag, and either
 - set a whitelist of zones, and/or
 - set a whitelist of environment, and/or
 - set a whitelist of clouds
as the condition of a flag value.

These dimensions are resolved in the controller and just before forwarding the flag data to a particular zone.  This is a "partial" resolution, since only the conditions with those dimensions are resolved.  This leads to conditions being removed (if true), or an entire rule being removed (if false).  The dry-run checks in hosted-feature-flags will work as expected, and show diffs if the partially resolved flag data diffs against the current zone's flag data.

The controller zone is special.  Because these dimensions may be useful in the controller, there is a way to avoid resolving these dimensions for the controller's zone (always prod.controller), environment (always prod), and cloud (whatever cloud the controller itself happens to be in).
 - The partial resolution of zone and/or cloud is DISABLED if that dimension is DECLARED on the flag.
Up to now, all dimensions declared on a flag has been informational, so this is the first semantic use of them.  In particular, flags should no longer declare ZONE_ID - almost all flags have it.  I will follow up with a PR that removes these.

When using the hv command, these dimensions are resolved just before hv pushes the flag data to a zone.  This may in fact turn a "put" of a flag data into a delete if all rules were removed.  Which is also what would happen if the same flag data is pushed through hosted-feature-flags.

This PR also adds the system dimension, and it is partially resolved like the zone and cloud as described above.  But I'm currently thinking that developers should not use it and I may document only the above two and/or discourage using the system dimension.


hakonhall:
FYI @bjorncs

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
```diff
@@ -217,41 +230,62 @@ private static void addFile(Builder builder, String rawData, Path filePath, Set<
...
-                    String applicationIdString = conditionValue.asString()
-                            .orElseThrow(() -> new IllegalArgumentException("Non-string application ID: " + conditionValue));
-                    // Throws exception if not recognized
-                    ApplicationId.fromSerializedForm(applicationIdString);
+            FetchVector.Dimension dimension = DimensionHelper
+                    .fromWire(condition.get("dimension")
+                                       .asString()
+                                       .orElseThrow(() -> new IllegalArgumentException("Invalid dimension in condition: " + condition)));
+            switch (dimension) {
+                case APPLICATION_ID -> validateStringValues(condition, ApplicationId::fromSerializedForm);
```
freva, line 248: :1st_place_medal:

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/systemflags/SystemFlagsDeployerTest.java
```diff
@@ -74,7 +81,81 @@ void deploys_flag_data_to_targets() throws IOException {
...
 
     @Test
-    void dryrun_should_not_change_flags() throws IOException {
+    void deploys_partial_flag_data_to_targets() {
+        // default.json contains one rule with 2 conditions, one of which has a condition on the aws cloud.
+        // This condition IS resolved for a config server target, but NOT for a controller target, because FLAG_ID
+        // has the CLOUD dimension set.
+        deployFlags(Optional.empty(), "partial/default.json", Optional.of("partial/put-controller.json"), true, PutType.CREATE, FetchVector.Dimension.CLOUD);
+        deployFlags(Optional.empty(), "partial/default.json", Optional.empty(), false, PutType.NONE, FetchVector.Dimension.CLOUD);
+        deployFlags(Optional.of("partial/initial.json"), "partial/default.json", Optional.of("partial/put-controller-2.json"), true, PutType.UPDATE, FetchVector.Dimension.CLOUD);
```
freva, line 90: `put-controller-2.json` is identical to `put-controller.json`?

flags/src/main/java/com/yahoo/vespa/flags/FetchVector.java
```diff
@@ -20,20 +23,25 @@ public class FetchVector {
...
 
-        /** Node type from com.yahoo.config.provision.NodeType::name, e.g. tenant, host, confighost, controller, etc. */
-        NODE_TYPE,
+        /**
+         * Cloud from com.yahoo.config.provision.CloudName::value, e.g. yahoo, aws, gcp.
+         *
+         * <p><em>Eager resolution</em>:  This dimension is resolved before putting the flag data to the config server
+         * or controller, unless controller and the flag has declared this dimension.
+         */
+        CLOUD,
```
freva, line 35: Also add `ENVIRONMENT`?

freva (dismissed):

controller-server/src/test/java/com/yahoo/vespa/hosted/controller/restapi/systemflags/SystemFlagsDeployerTest.java
hakonhall: Right, fixed.

flags/src/main/java/com/yahoo/vespa/flags/FetchVector.java
hakonhall: Good idea.

bjorncs (dismissed):
🥇

freva (approved):

PR #27932, opened 2023-07-31
Bratseth/lucene linguistics

bratseth:
@dainiusjocas feel free to review too


dainiusjocas (approved):
thanks for improving the code!

lucene-linguistics/src/main/java/com/yahoo/language/lucene/LuceneLinguistics.java
```diff
@@ -18,65 +18,29 @@
  *
  * TODO: docs for all available analysis components.
  * TODO: some registry for available language Analyzers.
```
dainiusjocas, line 20: This can be removed, [this](https://github.com/vespa-engine/vespa/pull/27932/files#diff-d81b01205e097cd1c0e99fc62533da4e4dc962003cf36b3e00e503711b3a089c) class serves as such a register

PR #27939, opened 2023-08-01
Add cluster name to service convergence response

mpolden:
To make it easier for CLI to discover deployed clusters.

@hmusum


mpolden:
or @bratseth

hmusum (approved):

PR #27940, opened 2023-08-01
Infrastructure metric sets

olaaun:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/MetricsConsumer.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/MetricsConsumer.java
@@ -1,21 +1,21 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.admin.monitoring;
 
+import ai.vespa.metrics.set.Metric;
+import ai.vespa.metrics.set.MetricSet;
 import ai.vespa.metricsproxy.core.VespaMetrics;
 import ai.vespa.metricsproxy.http.ValuesFetcher;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-import static com.yahoo.vespa.model.admin.monitoring.AutoscalingMetrics.autoscalingMetricSet;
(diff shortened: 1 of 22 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/MetricsConsumer.java
```diff
@@ -1,21 +1,21 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.admin.monitoring;
 
+import ai.vespa.metrics.set.Metric;
+import ai.vespa.metrics.set.MetricSet;
```
yngveaasheim, line 5: Needed?

container-core/src/main/java/com/yahoo/container/jdisc/state/MetricsPacketsHandler.java
```diff
@@ -208,11 +243,57 @@ private void addMetrics(MetricSet metricSet, ObjectNode packet) {
...
+            } else if (value instanceof GaugeMetric) {
+                GaugeMetric gauge = (GaugeMetric) value;
+                metrics.put(name + ".average", sanitizeDouble(gauge.getAverage()));
+                metrics.put(name + ".last", sanitizeDouble(gauge.getLast()));
+                metrics.put(name + ".max", sanitizeDouble(gauge.getMax()));
+                if (gauge.getPercentiles().isPresent()) {
+                    for (Tuple2<String, Double> prefixAndValue : gauge.getPercentiles().get()) {
+                        metrics.put(name + "." + prefixAndValue.first + "percentile", prefixAndValue.second.doubleValue());
+                    }
+                }
```
yngveaasheim, line 262: Should we include 'min', 'sum', 'count' suffixes here?

yngveaasheim: And 'rate'?

config-model/src/main/java/com/yahoo/vespa/model/admin/monitoring/MetricsConsumer.java
olaaun: Yes, used further down

PR #27946, opened 2023-08-02
- Put a lambda for getting predefined properties in a hash map.

baldersheim:
- Lookup the correct getter by a single lookup instead of navigating a large complicated if tree with quite a few branch mispredictions.
- Will do the same for setters afterwards.

@bratseth or @bjorncs PR


bratseth (approved):

bjorncs (commented):
👍

PR #27947, opened 2023-08-02
Revert "Revert "- initial document count must be number of unique documents per node." MERGEOK"

baldersheim:
Reverts vespa-engine/vespa#27944


baldersheim:
Fixed in #27953

PR #27952, opened 2023-08-02
Don't limit requests per destination

bjorncs:


baldersheim (approved):
What was default ?

bjorncs:
> What was default ?

1024. I've only been able to hit the limit when feeding through a secure HTTP/2 proxy though.

PR #27956, opened 2023-08-02
Limit max configserver payload accepted

bjorncs:


hmusum (approved):

hmusum (changes requested):
Build fails with:

Caused by: java.lang.IllegalArgumentException: Invalid XML according to XML schema, error in services.xml: element "config" not allowed here; expected the element end-tag or element "filtering" or "server" [143:43], input:

hmusum (approved):

PR #27957, opened 2023-08-02
Balder/use common map for getter and setters

baldersheim:
@bratseth or @bjorncs PR


bjorncs (approved):

bratseth:
//TODO Why is there error handling in set path and not in get path ?

Because it is illegal to set some things but not illegal to ask for them.

PR #27959, opened 2023-08-03
Hmusum/handle failure when sending file

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileServer.java
```diff
@@ -114,29 +118,28 @@ private boolean hasFile(FileReference reference) {
     FileDirectory getRootDir() { return fileDirectory; }
 
     void startFileServing(FileReference reference, Receiver target, Set<CompressionType> acceptedCompressionTypes) {
-        if ( ! fileDirectory.getFile(reference).exists()) return;
+        File file = fileDirectory.getFile(reference);
+        if ( ! file.exists()) return;
+
+        FileReferenceData fileData =
```
baldersheim, line 124: Should this be in a try-w-r clause ?

baldersheim (dismissed):
Only one comment

configserver/src/main/java/com/yahoo/vespa/config/server/filedistribution/FileServer.java
hmusum: Yes, that will be better, fixed.

baldersheim (approved):

PR #27965, opened 2023-08-03
Avoid taking a reference to a nullptr

baldersheim:
@arnej27959 or @aressem or @hakonhall PR

### DIFF
--- a/searchsummary/src/vespa/juniper/sumdesc.cpp
+++ b/searchsummary/src/vespa/juniper/sumdesc.cpp
@@ -623,9 +623,8 @@ int SummaryDesc::complete_extended_token(unsigned char* start, ssize_t length,
 /* Return a highlight tagged summary string from this summary
  * description
  */
-std::string SummaryDesc::get_summary(const char* buffer, size_t bytes,
-                                     const SummaryConfig* sumconf,
-                                     size_t& char_size)
+std::string
+SummaryDesc::get_summary(const char* buffer, size_t bytes, const SummaryConfig* sumconf, size_t& char_size)
 {
     std::vector<char> s;
     ssize_t prev_end = 0;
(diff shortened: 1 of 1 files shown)


searchsummary/src/vespa/juniper/sumdesc.cpp
```diff
@@ -759,7 +758,7 @@ std::string SummaryDesc::get_summary(const char* buffer, size_t bytes,
                s.size(), a.charLen());
     _sumconf = NULL; // Not valid after this call.
     char_size = a.charLen();
-    return std::string(&s[0], s.size());
```
hakonhall, line 762: I don't understand this could end up taking a reference to nullptr, but if size() is 0 then s[0] is certainly illegal.  Are you saying string's internal char buffer is initialized to nullptr?  Because c_str() could be called on a newly initialized string I thought the internal buffer always had at least length 1 to include the '\0'.

hakonhall (commented):
:+1:

searchsummary/src/vespa/juniper/sumdesc.cpp
baldersheim: Yes, it is &s[0] which is the problem. The vector's buffer is nullptr when nothing has been added, or space reserved.

hakonhall: What will c_str() do then if nothing has been added?  Return a pointer to a '\0' byte stored outside the buffer or will a buffer be allocated?

baldersheim: std::string has a buffer on the stack for up to 16 bytes which it uses for small strings before allocating any memory. The zero termination will be there if it is empty.

PR #27972, opened 2023-08-06
Hosted feature flags validation

hakonhall:
Specifying the zone as a dimension for a condition always fails due to a bug, and is fixed in this PR.  The bug is that the controller validates any and all the zones are members of ... an empty set, which obviously always fails.

This PR changes validation:
 - On the client-side (the `hv` command), avoid testing for the existence of a zone, whether it is a zone inferred from the filename or in a zone condition.  Validate the existence of these in the controller.  This ensures any Vespa version mismatch between the hv tool and the controller, only the controller decides existence of a zone.
 - Only files in the flags directory are zipped.  Files beginning with '.' are ignored.  All other files must have the correct filename format, and either they must be destined for the current system (including default.json), or another system, otherwise an exception is thrown.  However as mentioned above, the existence of a zone is not checked, but e.g. the region name must be valid.   In the controller, all files in the zip archive must be destined for the implied system.  If the controller detects an unknown file in the zip archive, an error is returned WITHOUT trying to deploy against all targets.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
@@ -171,13 +171,11 @@ public class SystemFlagsDataArchive {
     private static void addFile(Builder builder, String rawData, Path filePath, ZoneRegistry zoneRegistry, boolean force) {
         String filename = filePath.getFileName().toString();
 
-        if (!force) {
-            if (filename.startsWith("."))
-                return; // Ignore files starting with '.'
+        if (filename.startsWith("."))
+            return; // Ignore files starting with '.'
 
-            if (!FlagsTarget.filenameForSystem(filename, zoneRegistry.system()))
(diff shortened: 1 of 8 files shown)


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/systemflags/v1/SystemFlagsDataArchive.java
```diff
@@ -153,67 +153,49 @@ public List<FlagData> flagData(FlagsTarget target) {
...
-                throw new IllegalArgumentException(String.format(
-                        "Environment or zone in filename '%s' does not exist", filename));
-            }
-            return; // Ignore files irrelevant for system
-        }
-        if (!filename.endsWith(".json")) {
-            throw new IllegalArgumentException(String.format("Only JSON files are allowed in 'flags/' directory (found '%s')", filePath.toString()));
+
+        if (!force) {
+            if (filename.startsWith("."))
```
freva, line 175: Dont we want to always ignore these? (i.e. move outside of the `if (!force)`

freva (dismissed):

freva (approved):

PR #27976, opened 2023-08-07
Fade out round-robin dispatch policy.

baldersheim:
@geirst or @yngveaasheim PR
Removed from doc in https://github.com/vespa-engine/documentation/pull/2789


geirst (approved):
👍

yngveaasheim:
Do we need round-robin at times to be better able to reason over other differences in the searchnode request volumes, etc?

PR #27988, opened 2023-08-08
Balder/prepare for better stats reset code

baldersheim:
@geirst PR
@vekterli FYI
Here should be no semantic changes, except changing from std::vector to std::array for GlobalStats

### DIFF
--- a/storage/src/vespa/storage/distributor/idealstatemetricsset.cpp
+++ b/storage/src/vespa/storage/distributor/idealstatemetricsset.cpp
@@ -134,7 +134,7 @@ IdealStateMetricSet::IdealStateMetricSet()
 
 IdealStateMetricSet::~IdealStateMetricSet() = default;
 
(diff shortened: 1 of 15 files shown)


geirst (approved):
👍

storage/src/vespa/storage/distributor/idealstatemetricsset.cpp
```diff
@@ -134,7 +134,7 @@ IdealStateMetricSet::IdealStateMetricSet()
 
 IdealStateMetricSet::~IdealStateMetricSet() = default;
 
-void IdealStateMetricSet::setPendingOperations(const std::vector<uint64_t>& newMetrics) {
+void IdealStateMetricSet::setPendingOperations(vespalib::ConstArrayRef<uint64_t> newMetrics) {
```
vekterli, line 137: For new code I think we should prefer `std::span<const T>` over `ConstArrayRef<T>`

PR #28001, opened 2023-08-09
- Use a hashmap for string Node/NodeState in ClusterState for fast lo…

baldersheim:
…okup.

- Do explicit sort when serializing.

@geirst PR


geirst (approved):
👍

PR #28005, opened 2023-08-09
- Use hash_map instead of std::unordered_set dur to performance and m…

baldersheim:
…emory fragmentation.

- Move code from .h to .cpp that uses a lot of templated code.

@geirst PR


geirst (approved):
👍

PR #28006, opened 2023-08-09
- Since the vespalib::hash_map iterators are invalidated on modificat…

baldersheim:
…ion store a unique_ptr in the map and deref the pointer instead.

- Also expose vectors as const &.

@geirst PR


geirst (approved):
👍

PR #28008, opened 2023-08-09
Use faster and lighter hash_map for message tracking in distributor.

baldersheim:
@geirst PR

### DIFF
--- a/storage/src/vespa/storage/distributor/messagetracker.cpp
+++ b/storage/src/vespa/storage/distributor/messagetracker.cpp
@@ -33,21 +33,14 @@ MessageTracker::flushQueue(MessageSender& sender)
 uint16_t
 MessageTracker::handleReply(api::BucketReply& reply)
 {
-    std::map<uint64_t, uint16_t>::iterator found = _sentMessages.find(reply.getMsgId());
-    if (found == _sentMessages.end()) {
+    const auto found = _sentMessages.find(reply.getMsgId());
+    if (found == _sentMessages.end()) [[unlikely]] {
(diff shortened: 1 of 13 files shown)


geirst (approved):
👍

storage/src/vespa/storage/distributor/messagetracker.cpp
```diff
@@ -19,10 +20,11 @@ MessageTracker::~MessageTracker() = default;
 void
 MessageTracker::flushQueue(MessageSender& sender)
 {
-    for (uint32_t i = 0; i < _commandQueue.size(); i++) {
-        _commandQueue[i]._msg->setAddress(api::StorageMessageAddress::create(_cluster_ctx.cluster_name_ptr(), lib::NodeType::STORAGE, _commandQueue[i]._target));
-        _sentMessages[_commandQueue[i]._msg->getMsgId()] = _commandQueue[i]._target;
-        sender.sendCommand(_commandQueue[i]._msg);
+    _sentMessages.resize(_commandQueue.size());
```
vekterli, line 23: In the case `flushQueue()` is called multiple times per tracker (doesn't seem to currently be the case), this might be more optimal as `_sentMessages.resize(_sentMessages.size() + _commandQueue.size())` to avoid explicitly sizing the table down before implicitly sizing it up again upon reinsertion.

I've got a small branch going on the side, I can sneak this change in there.

baldersheim: Thanks, I be ready to review.

PR #28013, opened 2023-08-10
keep rpc target alive until pending rpc invocations complete

havardpe:
@baldersheim please review
@arnej27959 FYI


havardpe:
There might still be issues related to the async version callback though...

havardpe:
Version callback code seems reasonable (ref. counted)

baldersheim (approved):

PR #28022, opened 2023-08-10
Balder/cheaper split nodes into leaf groups

baldersheim:
@geirst PR
This is a rather large PR, but should be just cleanup.


geirst (approved):
👍

PR #28025, opened 2023-08-11
When splitting a large number of nodes into their leaf groups you end…

baldersheim:
… up with many vectors with only a single node in them.

Then it is more efficient to use a small_vector that keeps the 4 first entries within. Reduces # allocations and avoids indirection.

@geirst PR


geirst (approved):
👍

PR #28045, opened 2023-08-15
Hmusum/session data in one place

hmusum:
Controlled by a feature flag for now

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionData.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionData.java
@@ -59,7 +59,7 @@ public record SessionData(ApplicationId applicationId,
             return SlimeUtils.toJsonBytes(slime);
         }
         catch (IOException e) {
-            throw new RuntimeException("Serialization of " + this + " to json failed", e);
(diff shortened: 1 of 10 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionData.java
```diff
@@ -0,0 +1,87 @@
...
+    static final String SESSION_DATA_PATH = "sessionData";
+
+    public byte[] toJson() {
+        try {
+            Slime slime = new Slime();
+            toSlime(slime.setObject());
+            return SlimeUtils.toJsonBytes(slime);
+        }
+        catch (IOException e) {
+            throw new RuntimeException("Serialization of " + this + " to json failed", e);
```
hakonhall, line 62: This will print the dataplane tokens, certificates, and more.  The tokens are sensitive and too much is output?

hakonhall (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionData.java
hmusum: Yes, good point, will remove that

hakonhall (approved):

PR #28054, opened 2023-08-15
Control the auto-adjustment of targetHits in ANN using post-filtering.

geirst:
When searching the HNSW index in a post-filtering case, targetHits is auto-adjusted in an effort to still expose targetHits hits to first-phase ranking after post-filtering. The following formula is now used to ensure an upper bound of adjustedTargetHits, avoiding that the search in the HNSW index takes too long.

adjustedTargetHits = min(targetHits / estimatedHitRatio, targetHits * targetHitsMaxAdjustmentFactor).

The target-hits-max-adjustment-factor can be set in a rank profile and overridden per query. The value is in the range [1.0,inf], with the default being 20.0.

When setting this to 1.0, auto-adjustment of targetHits is effectively disabled.

@arnej27959 please review
@jobergum FYI


geirst:
@arnej27959 ping

arnej27959 (approved):
looks good to me

PR #28064, opened 2023-08-16
Add hint for empty API key

kkraune:


mpolden (commented):
I think this will only generate confusion as this hint may also show up when Auth0 is being used. We don't document that `VESPA_CLI_API_KEY` can be empty anywhere and I think we should keep it that way.

kkraune:
I am open to suggestions. We will point users to an example GitHub Action with Dataplane credentials only, that writes to Vespa Cloud with no warnings - the current version is at https://github.com/vespa-engine/documentation/blob/master/.github/workflows/feed.yml

What do you suggest we do to get to this?

baldersheim:
Complete or close ?

mpolden:
There's a ticket tracking the solution to this.

PR #28100, opened 2023-08-21
Make sure we use a free port when creating RpcServer

hmusum:


configserver/src/test/java/com/yahoo/vespa/config/server/rpc/RpcTester.java
```diff
@@ -69,20 +67,38 @@ public class RpcTester implements AutoCloseable {
...
+            tempRpcServer = createRpcServer(tempConfig);
+            tempTenantRepository = new TestTenantRepository.Builder()
+                    .withHostRegistry(hostRegistry)
+                    .withConfigserverConfig(tempConfig)
+                    .build();
+            tempTenantRepository.addTenant(tenantName);
+            startRpcServer(tempRpcServer, tempTenantRepository, tempSpec);
+            iterations++;
+            Thread.sleep(1);
+        } while (!tempRpcServer.isRunning() && iterations < 10);
```
jonmv, line 94: Perhaps assert here that the server is running?

jonmv (dismissed):
One minor comment, otherwise LGTM :)

configserver/src/test/java/com/yahoo/vespa/config/server/rpc/RpcTester.java
hmusum: Yes, good point, done

PR #28108, opened 2023-08-22
Add modify operation that creates non-existing sub-spaces with defaul…

geirst:
…t cell values.

@arnej27959 please review


arnej27959 (approved):
this looks good to me

PR #28116, opened 2023-08-22
Avoid dynamic_cast by adding an interface to get allocated size

baldersheim:
@toregge or @havardpe PR


havardpe (commented):
consider using type_index and static cast instead

baldersheim:
if (std::type_index(typeid(*cleanup)) == std::type_index(typeid(const stash::DeleteMemory)))
with static_cast was slower, it used string compare.

PR #28123, opened 2023-08-23
Hmusum/cleanup configgen

hmusum:


configgen/src/main/java/com/yahoo/config/codegen/CNode.java
```diff
@@ -69,9 +68,8 @@ void setMd5(String md5) {
         defMd5 = md5;
     }
 
-    public String getVersion() {
-        return defVersion;
-    }
+    @Deprecated // TODO: Vespa 9: Remove
```
gjoranv, line 71: Can't we just remove this, or is it considered a public api?

gjoranv (dismissed):

configgen/src/main/java/com/yahoo/config/codegen/CNode.java
hmusum: Yeah, I was considering that, but wasn't sure. Looking more at it I think this is only used internally. Removing it

gjoranv (approved):
🏆

PR #28133, opened 2023-08-23
Extend modify update operation in Java to match the new C++ behavior.

geirst:
This creates non-existing sub-spaces with default cell value first.

@arnej27959 please review


arnej27959 (approved):
👍

PR #28141, opened 2023-08-24
Avoid writing new hosts twice when retrying in GCP

jonmv:
@freva


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -16,15 +16,18 @@
...
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import static java.util.function.Predicate.not;
+
```
jonmv, line 30:
```suggestion
```

freva (dismissed):
Nice

Issue #28012, opened 2023-08-10
component url model is downloaded each time services.xml is deployed


eostis:
Here a capture of the Vespa's logs after a redeployment of services.xml.

The lack of caching could be due to the redirection of the Dropbox url?

![image](https://github.com/vespa-engine/vespa/assets/1785759/94e48c4b-a6f2-4a8e-8367-54dfb5521de1)

jobergum:
A URL file is expected to be downloaded if the deployed configuration caused a change so components need to be re-created. There is no caching mechanism.

eostis:
I notice that other urls are cached:
<img width="1328" alt="image" src="https://github.com/vespa-engine/vespa/assets/1785759/79e0f890-22e7-4646-a307-8c16cfedb42b">

I thought it was the norm to cache urls.

hmusum:
The HTTP request adds the `If-Modified-Since` header when downloading. If the server does not honor this (as [this Dropbox forum post](https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/If-Modified-Since-and-Last-Modifed-headers/td-p/217613) suggests, it will be downloaded every time. This would also explains why you are seeing this for only some URLs.

That said, the plan is to change this to never download an url if we already have cached it. Which means that you need to use a new/changed url if new content is wanted.

eostis:
> That said, the plan is to change this to never download an url if we already have cached it

Does that mean the cloud storage urls like Dropbox will be cached then?

Currently, each redeploy takes minutes and downloads GBs. And cloud storage urls are the only scalable solution.

hmusum:
Yes, that's the plan, but I haven't looked at the details and I have no ETA for this yet.

eostis:
Thanks.

hmusum:
Cached URLs will always be used if already download, available in Vespa 8.217.31

eostis:
@hmusum Works great with 8.217.46: the Dropbox model is cached as expected.

PR #28145, opened 2023-08-24
Use cached file if an url has already been downloaded

hmusum:
Stop using If-Modified-Since header when downloading, as most cloud provider solutions don't support it. Instead, always use cached download if it exists, which in the case where If-Modified-Since header is not supported will avoid downloading url for every deployment. This also means that there will be no changes to content used for an url unless the url configured in the application package is explicitly changed is explicitly changed to use another url

https://github.com/vespa-engine/vespa/issues/28012


vekterli:
Are [ETags](https://en.wikipedia.org/wiki/HTTP_ETag) more widely supported? If so, would that be an option?

hmusum:
I have not researched that.

We want to always use the cached version if it exists, as this means the content will not change unless a deployment is done (i.e. it does not happen at some arbitrary time due to someone changing the content of the file on the server it is downloaded from).

baldersheim (approved):

PR #28152, opened 2023-08-25
Add "create non-existing cells" flag to TensorModifyUpdate.

geirst:
When this is true, non-existing cells in the input tensor is created before applying the modify update.
The default cell value is 0.0 for REPLACE and ADD operations, and 1.0 for MULTIPLY operation.

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #28154, opened 2023-08-25
Make paged attribute test (checking growth of swap file) more robust.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #28157, opened 2023-08-25
Set up maven wrapper in bootstrap

gjoranv:
Not used for building vespa yet.

FYI: @arnej27959, @bjorncs, @baldersheim


bootstrap.sh
```diff
@@ -36,7 +36,10 @@ get_env_var_with_optional_default() {
     fi
 }
 
+# TODO: use maven-wrapper after changing mvn command for vespa in factory/build-java.sh
```
aressem, line 39: I think we can remove these two comments now.

aressem (dismissed):

aressem (dismissed):

aressem (approved):

PR #28185, opened 2023-08-28
Issue rebuild commands in parallel, since lock is shared

jonmv:
@hakonhall

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
@@ -1,6 +1,7 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.maintenance;
 
+import com.yahoo.concurrent.DaemonThreadFactory;
 import com.yahoo.jdisc.Metric;
 import com.yahoo.vespa.hosted.provision.Node;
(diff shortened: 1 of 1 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
```diff
@@ -33,26 +42,41 @@ public class DiskReplacer extends NodeRepositoryMaintainer {
...
         int failures = 0;
-        for (var host : nodes) {
-            Optional<NodeMutex> optionalMutex = nodeRepository().nodes().lockAndGet(host, Duration.ofSeconds(10));
-            if (optionalMutex.isEmpty()) continue;
-            try (NodeMutex mutex = optionalMutex.get()) {
-                // Re-check flag while holding lock
-                host = mutex.node();
-                if (!host.status().wantToRebuild()) {
-                    continue;
+        try (var locked = nodeRepository().nodes().lockAndGetAll(nodes.asList(), Optional.of(Duration.ofSeconds(10)))) {
```
hakonhall, line 47: Cap the list to e.g. 10?

hakonhall (dismissed):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
jonmv: But I guess something else stops these instances ... and attempting the first 10 over and over due to some failure would keep all later hosts waiting indefinitely.

PR #28197, opened 2023-08-28
added pop_back function to SmallVector

havardpe:
follow std::vector by making it undefined for empty vectors

@arnej27959 please review


arnej27959 (approved):
👍

PR #28241, opened 2023-08-29
Add cluster specific settings for model evaluation

lesters:
@bratseth or @hmusum: Please review

Suggestion, if you have any other proposals to fix this please let me know.

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/OnnxModel.java
+++ b/config-model/src/main/java/com/yahoo/schema/OnnxModel.java
@@ -20,11 +20,13 @@ import java.util.Set;
  */
 public class OnnxModel extends DistributableResource implements Cloneable {
 
+    // Model information
     private OnnxModelInfo modelInfo = null;
     private final Map<String, String> inputMap = new HashMap<>();
     private final Map<String, String> outputMap = new HashMap<>();
     private final Set<String> initializers = new HashSet<>();
 
+    // Runtime options
     private String  statelessExecutionMode = null;
     private Integer statelessInterOpThreads = null;
     private Integer statelessIntraOpThreads = null;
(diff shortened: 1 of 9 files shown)


config-model/src/main/java/com/yahoo/schema/OnnxModel.java
```diff
@@ -39,6 +39,19 @@ public OnnxModel(String name, String fileName) {
         validate();
     }
 
+    @Override
+    public OnnxModel clone() {
+        try {
+            OnnxModel clone = (OnnxModel) super.clone();
```
bratseth, line 45: This will be a shallow clone but I think you want a deep one?
So, you should initialize inputMap, outputMap and initializers to new, empty collections before put/addAll, and deep clone modelInfo.

bratseth (changes requested):

config-model/src/main/java/com/yahoo/schema/OnnxModel.java
lesters: Actually, a shallow clone would work here, because the `inputMap`, `outputMap`, `initializers` and `modelInfo` are model-specific parameters, while only the last 4 are configurable via `services.xml` and thus cluster-specific. However, that would probably break the clone contract?

bratseth: It's ok then but then you shouldn't add anything to those existing collections in clone.

And consider if this may lead to bugs later, maybe either deep clone anyway, or separate the general model and use-specific values into two classes.

lesters: Removed the part of adding to the existing collections. Looking at the code, the fields are effectively immutable so specified that shallow clone is sufficient for now.

bratseth (approved):

PR #28246, opened 2023-08-29
added universal dot product

havardpe:
note that optimization is not yet active in production

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #28250, opened 2023-08-29
Let node info for cluster controller be explicit, and not a metric co…

baldersheim:
…nsumer.

@vekterli PR

### DIFF
--- a/storage/src/vespa/storage/storageserver/servicelayernode.cpp
+++ b/storage/src/vespa/storage/storageserver/servicelayernode.cpp
@@ -39,6 +39,15 @@ ServiceLayerNode::ServiceLayerNode(const config::ConfigUri & configUri, ServiceL
 {
 }
 
+void
+ServiceLayerNode::report(vespalib::JsonStream &writer) const
+{
+    (void) writer;
+    if (_bucket_manager) {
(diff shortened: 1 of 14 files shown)


storage/src/vespa/storage/storageserver/servicelayernode.cpp
```diff
@@ -39,6 +39,15 @@ ServiceLayerNode::ServiceLayerNode(const config::ConfigUri & configUri, ServiceL
 {
 }
 
+void
+ServiceLayerNode::report(vespalib::JsonStream &writer) const
+{
+    (void) writer;
```
vekterli, line 45: This seems unnecessary since the variable is referenced within the `if`

vekterli (dismissed):
:+1: host info reporting will now do an on-demand DB scan per `GetNodeState` response in the context of the state manager thread instead of using cached metrics, but I don't think this should matter much in practice since the scan is on a lock-free snapshot.

storage/src/vespa/storage/storageserver/servicelayernode.cpp
baldersheim: Fixed

PR #28257, opened 2023-08-29
Balder/unify code

baldersheim:
@vekterli PR
Best review commit by commit,
All but last 2 are just cleanup.


vekterli (approved):
👍🧹 blast from the past

PR #28272, opened 2023-08-30
Keep load balancer provision failure cause, but retain logic otherwise

jonmv:
@mpolden please review.

I believe I managed to retain the control flow here; in cases where there were exceptions, we'd always return an `empty()` instance, and write that, which should be the case now as well. The difference should really just be that we propagate out an exception with the original cause.


mpolden (approved):
👍

PR #28281, opened 2023-08-30
Use std::filesystem::directory_iterator in disk index cleaner unit test.

toregge:
@baldersheim : please review


baldersheim (dismissed):
Consider reversing logic to avoid continue

baldersheim (approved):

PR #28286, opened 2023-08-30
Use std::filesystem::directory_iterator in index manager unit test.

toregge:
@baldersheim : please review


baldersheim (approved):

baldersheim (approved):
This code could be shared with the fusion test...

PR #28298, opened 2023-08-31
Print file content if assert fails

hmusum:


hmusum:
Not needed, fixed in https://github.com/vespa-engine/vespa/pull/28299

PR #28310, opened 2023-08-31
Add back shutdown(), breaks classes that extend AbstractSpoolingLogger

hmusum:


baldersheim (approved):

jonmv:
Is this missing a `@PublicApi` then?

jonmv (approved):

hmusum:
Looking more into it users should use KinesisLogger now, so it does not have to be a public API (but we have tried to make it possible to use as a public API, not documented yet). So I think we can just update the classes that extend AbstractSpoolingLogger (I'll make PRs)

PR #28314, opened 2023-08-31
Upgrade questdb 6.2 -> 6.3.1

baldersheim:
@bjorncs PR


bjorncs:
Newest minor version is 6.7 (https://central.sonatype.com/artifact/org.questdb/questdb/6.7/overview). I believe the non `jdk8` suffixed version should be used instead as it's targeting JDK11.

bjorncs (approved):

PR #28336, opened 2023-08-31
Have SystemInfo injected, which actually knows our cluster name

jonmv:
@freva


jonmv:
`QrConfig` wasn't configured on the container cluster level, just on the container level, obviously ... and since the `Dispatcher` component was added to the cluster, not the container, it would just have an empty cluster name.

freva (approved):
¯\\\_(ツ)_/¯

baldersheim (approved):

PR #28339, opened 2023-08-31
benchmark universal dot product vs other options

havardpe:
@arnej27959 please review

we probably want to add more cases in the future and also factor out the benchmarking code for easy reuse


arnej27959 (approved):
👍

PR #28369, opened 2023-09-04
Remove unused executor

mpolden:
@jonmv

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
@@ -1,7 +1,6 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.provision.maintenance;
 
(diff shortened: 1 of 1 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
```diff
@@ -39,10 +34,13 @@ public class DiskReplacer extends NodeRepositoryMaintainer {
 
     @Override
     protected double maintain() {
-        NodeList nodes = nodeRepository().nodes().list().rebuilding(true);
+        NodeList candidates = nodeRepository().nodes().list().rebuilding(true);
+        if (candidates.isEmpty()) {
+            return 0;
```
jonmv, line 39: Not 1?

jonmv (approved):
Consider return value when no work to do.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/DiskReplacer.java
mpolden: The definition has changed. 0 now indicates there is no work to do.

jonmv:
How do we then differentiate between complete failure, and no work to do?

PR #28378, opened 2023-09-04
When pruning unavailable replicas, use intersection of cluster state and distribution config

vekterli:
@geirst please review

For a node to be considered available, it needs to be present and available in the current cluster state, _as well_ as be part of a valid group in the distribution config itself.

If either of these conditions are not satisfied, the node will be considered unavailable. The distributor will remove all knowledge of replicas on such nodes from its internal bucket DB.


geirst (approved):
👍

PR #28395, opened 2023-09-05
Keep and reconfigure ClusterMonitor when reconfiguring dispatcher

jonmv:
@bjorncs or @baldersheim please review.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/cluster/ClusterMonitor.java
+++ b/container-search/src/main/java/com/yahoo/search/cluster/ClusterMonitor.java
@@ -3,16 +3,22 @@ package com.yahoo.search.cluster;
 
 import com.yahoo.concurrent.ThreadFactoryFactory;
 import com.yahoo.search.result.ErrorMessage;
+import com.yahoo.yolean.UncheckedInterruptedException;
 
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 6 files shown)


bjorncs (changes requested):
```

10:57:42 [INFO] --- abi-check-plugin:8-SNAPSHOT:abicheck (default) @ container-search ---
10:57:43 [ERROR] Class com.yahoo.search.cluster.ClusterMonitor: Missing method public void ping(java.util.concurrent.Executor)
10:57:43 [ERROR] Class com.yahoo.search.cluster.ClusterMonitor: Extra method public synchronized void reconfigure(java.util.Collection)
10:57:43 [ERROR] Class com.yahoo.search.cluster.ClusterMonitor: Extra method public synchronized void ping(java.util.concurrent.Executor)
```

container-search/src/main/java/com/yahoo/search/cluster/ClusterMonitor.java
```diff
@@ -74,30 +105,48 @@ public void add(T node, boolean internal) {
...
+        if (nodesToRemove.remove(node)) {
+            nodeMonitors.remove(node);
+            monitor = null;
+        }
+
+        // Update monitor state only when it actually changes.
+        if (monitor != null) {
+            Boolean wasWorking = monitor.isKnownWorking();
+            monitorUpdate.accept(monitor);
+            if (wasWorking != monitor.isKnownWorking())
```
bjorncs, line 132: `monitorUpdate.accept()` changes result of `isKnownWorking()`?

bjorncs (approved):

container-search/src/main/java/com/yahoo/search/cluster/ClusterMonitor.java
jonmv: Yes, I kept this code as it was. The cluster monitor is used elsewhere, and I suppose we want to avoid unnecessary updates to state there.

PR #28400, opened 2023-09-05
Never clear load balancer instance

mpolden:
Once a load balancer instance has been provisioned, we shouldn't attempt to
clear it if future provision calls (reconfiguration) fails. The model already
disallows this and throws if state is active and instance is empty.

@jonmv


oyving:
FYI @andreer

jonmv (approved):
Yes, good, I like this better :)

PR #28401, opened 2023-09-05
Proposal for new DNS load balancing option in vespa feed client

pfrybar:
Hello, this PR proposes adding an option to the feed client to use DNS load balancing. It doesn't implement any of the code (yet) but I wanted to get an opinion if this is something you would find useful.

The code changes should be relatively minimal - the endpoint passed into the `FeedClient` would be periodically resolved in the background to come up with a new set of feed container endpoints. These are the actual endpoints which would be used for feeding.

### DIFF
--- a/vespa-feed-client-api/abi-spec.json
+++ b/vespa-feed-client-api/abi-spec.json
@@ -165,6 +165,7 @@
       "public abstract ai.vespa.feed.client.FeedClientBuilder setProxyCaCertificates(java.util.Collection)",
       "public abstract ai.vespa.feed.client.FeedClientBuilder setEndpointUris(java.util.List)",
       "public abstract ai.vespa.feed.client.FeedClientBuilder setProxy(java.net.URI)",
+      "public abstract ai.vespa.feed.client.FeedClientBuilder setDnsLoadBalancing(boolean)",
       "public abstract ai.vespa.feed.client.FeedClientBuilder setCompression(ai.vespa.feed.client.FeedClientBuilder$Compression)",
       "public abstract ai.vespa.feed.client.FeedClient build()"
     ],
--- a/vespa-feed-client-api/src/main/java/ai/vespa/feed/client/FeedClientBuilder.java
+++ b/vespa-feed-client-api/src/main/java/ai/vespa/feed/client/FeedClientBuilder.java
@@ -144,6 +144,14 @@ public interface FeedClientBuilder {
     /** Specify HTTP(S) proxy for all endpoints */
     FeedClientBuilder setProxy(URI uri);
 
+    /**
+     * Turns on DNS load balancing for the cluster.
+     *
+     * Use this option if the given endpoint resolves to a list of feed container endpoints.
+     * This list will be periodically refreshed (resolved) to get the latest container endpoints.
+     */
+    FeedClientBuilder setDnsLoadBalancing(boolean enabled);
+
     /** What compression to use for request bodies; default {@code auto}. */
     FeedClientBuilder setCompression(Compression compression);
 
--- a/vespa-feed-client-api/src/main/java/ai/vespa/feed/client/FeedClientBuilder.java
+++ b/vespa-feed-client-api/src/main/java/ai/vespa/feed/client/FeedClientBuilder.java
@@ -145,10 +145,10 @@ public interface FeedClientBuilder {
     FeedClientBuilder setProxy(URI uri);
 
     /**
-     * Turns on DNS load balancing for the cluster.
+     * Turns on DNS load balancing to the feed cluster.
      *
      * Use this option if the given endpoint resolves to a list of feed container endpoints.
-     * This list will be periodically refreshed (resolved) to get the latest container endpoints.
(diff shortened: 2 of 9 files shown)


jonmv:
This would be a welcome contribution! Please note that we recently replaced the Apache client library with the Jetty one. 
I'm also not sure we'd need a feed client that _doesn't_ support DNS based load balancing, but having this flag at least during development is a good idea.

pfrybar:
Great! I pushed up a pretty simple implementation. If enabled, it will do a DNS A record lookup on each of the original endpoints every 30 seconds, gather the results into a new set of endpoints, and atomically replace the previous ones.

You can see it in action in this screenshot. I have a DNS record (`vespa-feed.docker.home`) which is used to load balance a set of feed containers. I scale from 1 feed container (`10.0.23.129`) to 3 containers (`10.0.23.129`, `10.0.16.72`, `10.0.13.21`). I then scale back down again. You can see that the feed client picks up these new endpoints. I don't show it here, but I also verified that the containers were actually receiving the requests.

Screenshot 2023-09-07 at 23 07 28

If the endpoint is unresolveable then the behaviour is the same as it was previously. For example, here is what would happen before this change with an unknown hostname:

Screenshot 2023-09-07 at 16 26 39

And here's what happens with this change:

Screenshot 2023-09-07 at 22 32 27

Note: I only implemented this for A records and not SRV, which are sometimes used for service discovery via DNS. It wouldn't be a big change to add this in the future.

Please let me know what you think.

jonmv:
Thanks for the contribution!

One immediate concern here is that this breaks feeding where the feed endpoints present the client with a standard web certificate, as the hostname the HTTP client expects is now simply an IP address, but that is not likely to be present in the web certificate. There are surely ways to work around this, but a lower-level implementation would avoid this problem altogether.

It looks to me like the Jetty client is easily customisable to support this behaviour by manipulating the `SocketAddressResolver` you configure the client with. You can see the `Ipv4PreferringResolver` inner class of `JettyCluster` as a starting point here, and also how that is used. 
What I think we want is to extend that class with the option to shuffle or shift the addresses before returning them to the Jetty HTTP client that uses the resolved addresses. We saw the client iterates through the returned list, and picks the first address it successfully connects to, so returning the full list is the right thing to do, but ordered differently each time. By far the easiest seems to be to just call `Collections.shuffle(...)` on the list, before returning it, whenever DNS load balancing is turned on. 

Additionally, you need to ensure the pooled connections are eventually closed, so the client re-resolves the hostname(s). This can be done by setting a finite connection TTL on the connection pool, see https://github.com/vespa-engine/vespa/pull/28401/files#diff-96378c5b3a9553a69e25a6d6449f2ec569c82c468940314e613a9c6d15868b29R200-R203, with `pool.setMaxDuration(...)`.

pfrybar:
Thanks for the quick review!

I hadn't thought about the certificates 🤔

Regarding your suggestion to push the logic lower, I thought about that approach but saw some limitations. The time it takes to add/remove a feed container becomes limited by the connection pool TTL. We can set this to a lower value, like 30 seconds, so that changes get propagated faster, but that means connections will only last for 30 seconds and there will be a performance penalty. I haven't explored how big this penalty is, but my gut feeling is that it would be quite large. We could try raising the TTL to something like 5 minutes, but that slows down our scaling abilities and might still have some negative performance implications.

The other thing I'm not sure about is how well the load will be balanced between the feed endpoints. I'm not too familiar with Jetty, but I assume by default that connections will be re-used and with the addition of mutiplexing that means a single connection to a single feed container can probably serve many requests in parallel. We could try to force more connections by setting a low value for `maxStreamPerConnection` but then that defeats the purpose of HTTP/2.

I can keep investigating further. Another potential option would be to push the implementation even further down the stack and create a `DNSLoadBalancingHttpClient` Jetty client which does proper balancing across the available DNS records while refreshing the list in the background. That might be an interesting avenue.

jonmv:
I haven't looked closely at the internals of the Jetty HTTP client, its connection pooling, and so on, but I wouldn't be surprised if it was relatively easy to change the connection pool to balance the load pretty well at that level?

I'm not too concerned about performance hits, at least not until demonstrated. I don't think we have any clear indications that a few more connections, or a reasonable connection TTL, hurts much, in our performance tests. With multiple open connections, the feed client would also still have other connections to use while reconnecting one of the others. 

I could also add that in Vespa Cloud, the  containers by default drop connections roughly every minute, to facilitate load balancing. We also add some randomness to the TTL, to spread reconnect attempts across the TTL interval. 

If you have the time and the means to experiment with this, it would be great!

Ah, and before I forget, do you know what we can expect from the DNS resolver, in terms of getting cached results from, e.g., the host machine?

pfrybar:
Hey again. I tried it out your initial suggestion of shuffling the records returned from `Ipv4PreferringResolver` and it does indeed work well when combined with a low TTL on the connection pool. I didn't see any negative performance effects either.

There's a chance that we may end up implementing this at a higher level in our indexing jobs which use the feed client since it may still be beneficial to have more control over the DNS refreshing behaviour. If we decide that we want to try and put it into the feed client, I'll open another PR at that time.

Thanks for the help!

PR #28403, opened 2023-09-05
Add parameter for listing all endpoints

mpolden:
@ean


mpolden:
or @tokle. This will list all endpoints if `includeAllEndpoints=true`.

tokle (approved):

PR #28406, opened 2023-09-05
Don't spam log with stack trace if application did not request GPU

bjorncs:


mpolden:
I don't understand this. If the application does not have a GPU device, we don't try to load CUDA support at all: https://github.com/vespa-engine/vespa/pull/28406/files#diff-4d521eb135ccd5e000c93aea500daf7c213e06652d38b482d82f0df8fa8fcd31R143

bjorncs:
That's not the case for the Huggingface and Bert embedders, which optimistically tries to load CUDA.

mpolden:
How? OrtSession::addCUDA needs a GPU device number.

bjorncs:
By using device number `0` as default.

mpolden:
In that case I would make it:
```
if (loadCuda) {
  // log same as before as the exception is useful when you explicitly request GPU and it fails
} else {
  log.info("Ignoring optimistic attempt to load CUDA blabla")
}

bjorncs:
Replaced by https://github.com/vespa-engine/vespa/pull/29107

PR #28413, opened 2023-09-05
detect not having to make a new value index

havardpe:
- more benchmarking
- drop expression descriptions
- inline sparse plan execution

@arnej27959 please review


arnej27959 (approved):
👍

arnej27959 (approved):
👍

havardpe:
Note that there is currently a bug with how expanding reduce is handled. This will be fixed in an upcoming PR once the merge pipeline is re-opened.

PR #28414, opened 2023-09-05
Enumerate pending message IDs on a bucket predicate basis

vekterli:
@geirst please review

Lets a caller enumerate all IDs of messages pending towards buckets that match a caller-provided predicate function. A separate message ID callback is invoked per distinct message.

Intended to be used as part of cancellation of pending messages where bucket ownership has been lost (with the predicate being "do I no longer own this bucket?").

Also remove hard-coded multi-index numeric indices in favor of named constants.


geirst (approved):
👍

PR #28431, opened 2023-09-07
Consider config models in use when expiring artifacts

hmusum:
Keep artifacts with the same version as a config model version we support. In addition, in cd systems we don't want to expire any artifacts with the same version as a config model version in use in the correspondin main/public system

This change should only end up expiring _fewer_ artifacts than before.


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArtifactExpirer.java
```diff
@@ -3,21 +3,32 @@
...
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import static com.yahoo.yolean.Exceptions.uncheck;
+import static java.util.logging.Level.INFO;
 
 /**
- * Periodically expire unused artifacts, e.g. container images and RPMs.
+ * Periodically expire unused artifacts, e.g. container images and RPMs. Artifacts with a version that is
+ * present in config-models-*.xml are never expired (in cd we .
```
hakonhall, line 31: Unfinished sentence.

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArtifactExpirer.java
```diff
@@ -73,4 +92,35 @@ private boolean isExpired(Artifact artifact, Instant now, VersionStatus versionS
         return true;
     }
 
+    /** Model versions in use in this system, and, if this is a CD system, in the main/public system */
+    private Set<Version> modelVersionsInUse() {
+        var system = controller().system();
+        var versions = versionsForSystem(system);
+
+        if (controller().system().isCd()) {
```
hakonhall, line 100: This `if` is unnecessary as it is covered by the below `if (system == SystemName.PublicCd)` etc

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/ArtifactExpirer.java
```diff
@@ -73,4 +92,35 @@ private boolean isExpired(Artifact artifact, Instant now, VersionStatus versionS
...
+        var versions = readConfigModelVersionsForSystem(systemName.name());
+        log.log(INFO, "versions for system " + systemName.name() + ": " + versions);
+        return versions;
+    }
+
+    private Set<Version> readConfigModelVersionsForSystem(String systemName) {
+        List<String> lines = uncheck(() -> Files.readAllLines(configModelPath.resolve("config-models-" + systemName + ".xml")));
+        return lines.stream()
+                .filter(line -> line.contains("VespaModelFactory."))
+                .map(line -> line.substring(line.indexOf("id='VespaModelFactory") + 22, line.indexOf("' class")))
```
hakonhall, line 121: `+ 22` comes from `"id='VespaModelFactory".size()` ?  Extract `"id='VespaModelFactory"` as a constant?

With `startIndex = line.indexOf("id='VespaModelFactory") + 22`, the end index of the version would be `lineIndex.indexOf("'", startIndex)`, which avoids relying on the presence of the following `class` tag.

hakonhall (dismissed):
Approves because the code looks correct, but have a few suggestions for this or a follow-up PR.

hmusum:
Thanks, all should be fixed now

hakonhall (approved):

mpolden:
Great!

PR #28438, opened 2023-09-07
Support optional 'path' config type

hmusum:
Initial work needed for this to work (config definition and config library)

### DIFF
--- a/configgen/src/test/java/com/yahoo/config/codegen/DefLineParsingTest.java
+++ b/configgen/src/test/java/com/yahoo/config/codegen/DefLineParsingTest.java
@@ -1,6 +1,7 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.config.codegen;
 
+import java.util.Optional;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.*;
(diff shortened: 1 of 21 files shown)


configgen/src/test/java/com/yahoo/config/codegen/DefLineParsingTest.java
```diff
@@ -246,4 +247,24 @@ void testRestartSpecification() {
...
+        DefLine l = new DefLine("pathWithDef path optional");
+
+        assertEquals("pathWithDef", l.getName());
+        assertNull(l.getDefault());
+        assertTrue(l.isOptional());
+        assertEquals("optionalPath", l.getType().getName());
+    }
+
+    @Test
+    void testParsPathWithDefault() {
```
gjoranv, line 261: There seems to be a mismatch between the test name and the actual def lines. We probably want to test optional path both with and without default?

gjoranv (dismissed):

configgen/src/test/java/com/yahoo/config/codegen/DefLineParsingTest.java
hmusum: Right, thanks, leftover from when this was implemented differently, rewritten and renamed now

gjoranv (approved):

PR #28441, opened 2023-09-07
Wire distributor operation cancelling to cluster state/config change edges

vekterli:
@havardpe please review

This wires cancellation of pending operations/messages to content nodes in the following scenarios:

 * One or more content nodes become unavailable in a newly received cluster state version (triggered when first received, i.e. at the pending state start edge).
 * One or more nodes are removed from the distribution config.
 * The set of available distributors changes, which in turn changes the ownership of a fraction of the set of super buckets. Pending operations to buckets that were owned by the current distributor in the previous state, but not in the new state, are all cancelled.

Introduce cancellation support for internal maintenance operations. As part of this, move `CancelScope` tracking out into the parent `Operation` class to unify cancellation tracking across both client and maintenance operations.

Remove interface vs. impl indirection for `PersistenceMessageTracker` since it's only ever had a single implementation and it likely never will have another.

Add config for enabling cancellation (defaults to false).

### DIFF
--- a/storage/src/vespa/storage/distributor/persistencemessagetracker.cpp
+++ b/storage/src/vespa/storage/distributor/persistencemessagetracker.cpp
@@ -39,7 +39,7 @@ PersistenceMessageTracker::PersistenceMessageTracker(
 
 PersistenceMessageTracker::~PersistenceMessageTracker() = default;
 
-bool
+PersistenceMessageTracker::PostPruningStatus
 PersistenceMessageTracker::prune_cancelled_nodes_if_present(
         BucketInfoMap& bucket_and_replicas,
         const CancelScope& cancel_scope)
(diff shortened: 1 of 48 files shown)


storage/src/vespa/storage/distributor/persistencemessagetracker.cpp
```diff
@@ -28,41 +29,42 @@ PersistenceMessageTrackerImpl::PersistenceMessageTrackerImpl(
...
 
-void
-PersistenceMessageTrackerImpl::cancel(const CancelScope& cancel_scope)
-{
-    _cancel_scope.merge(cancel_scope);
-}
-
-void
-PersistenceMessageTrackerImpl::prune_cancelled_nodes_if_present(
+bool
```
havardpe, line 42: consider enum here to better communicate the meaning of true and false

havardpe (dismissed):

storage/src/vespa/storage/distributor/persistencemessagetracker.cpp
vekterli: Done

vekterli:
@havardpe PTAL

havardpe (approved):

PR #28454, opened 2023-09-08
Ensure that application-test is included when it exists

mpolden:
Also removes use of `target/application{,-test}.zip` as this will always contain
the same as `target/application{,-test}`.

@bratseth


bratseth (dismissed):

mpolden:
Some tests need to be updated.

PR #28456, opened 2023-09-08
Wg timestamp in node repo

gjoranv:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodePatcher.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodePatcher.java
@@ -109,8 +109,7 @@ public class NodePatcher {
                                               "reports",
                                               "trustStore",
                                               "vespaVersion",
-                                              "wireguardPubkey",
-                                              "wireguardKeyTimestamp"));
+                                              "wireguardPubkey"));
             if (!disallowedFields.isEmpty()) {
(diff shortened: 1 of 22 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodePatcher.java
```diff
@@ -274,6 +275,8 @@ private Node applyField(Node node, String name, Inspector value, Inspector root,
                 return nodeWithTrustStore(node, value);
             case "wireguardPubkey":
                 return node.withWireguardPubkey(SlimeUtils.optionalString(value).map(WireguardKey::new).orElse(null));
+            case "wireguardKeyTimestamp":
```
freva, line 278: Why is this patchable? Take the current timestamp in node-repo when the key is updated instead?

gjoranv: It did not make sense to patch this separately. Updated to setting it along with the key.

freva: Why should it be patchable at all? Generally we don't take timestamps for client for things that are meant to be 'now'.

gjoranv: I misunderstood your comment. Will fix.

freva (dismissed):

freva (approved):

PR #28461, opened 2023-09-08
Remove deprecated "revert" functionality from distributor code

vekterli:
@geirst please review
@baldersheim FYI

This is a remnant from the VDS days and can only work when the backend is a multi-version store. Code has been explicitly config model-disabled for Proton since day one.

This PR removes the distributor-specific bits. Tearing out the rest (persistence thread processing code, serialization logic etc) will come as a follow-up.


baldersheim (approved):

geirst (approved):
👍

PR #28488, opened 2023-09-12
Make normal log message less scary

kkraune:
- Note: I don't have a C++ build env, so have not run tests

CC @jobergum


jobergum:
Small step forward!

baldersheim (approved):

PR #28525, opened 2023-09-14
allow universal dot product optimization

havardpe:
only used for cases where both inputs have mapped dimensions

@arnej27959 please review


arnej27959 (approved):
👍

PR #28538, opened 2023-09-15
- Add utility to do substring  extraction by codepoints, instead of j…

baldersheim:
…ava char index.

- Test and use it in SubstringExpression in indexing language.

@bratseth or @bjorncs PR


bratseth (approved):
:+1

PR #28540, opened 2023-09-15
Add support for case-insensitive matching to Levenshtein DFAs

vekterli:
@geirst please review
@toregge @havardpe FYI

Adds matching modes `Cased` and `Uncased`.

`Cased` requires UTF-32 code points to match exactly, and successor strings are guaranteed to be strictly higher than the source (candidate) string in `memcmp` order. This mirrors the behavior of the current DFA implementation.

`Uncased` treats all characters as if they were lowercased, both for the target and source strings. The target (query) string is explicitly lowercased at DFA build-time to avoid duplicate work. Source strings are implicitly lowercased character by character on-demand during matching.

Important ordering note: Successor strings for `Uncased` are generated _as if_ the source string was originally all in lowercase form. This requires some extra added handling when emitting successor prefixes, as we can't just blindly copy UTF-8 bytes from the source string as we do when matching in `Cased` mode.

A new casing-dimension has been added to most parameterized unit tests.


toregge (approved):

geirst (approved):
👍

PR #28541, opened 2023-09-15
Hmusum/always write application data as json

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/application/ApplicationCuratorDatabase.java
```diff
@@ -74,34 +74,56 @@ public boolean exists(ApplicationId id) {
...
+            curator.set(applicationPath(id), applicationData.toJson());
+            modifyReindexing(id, ApplicationReindexing.empty(), UnaryOperator.identity());
+        }
+    }
+
+    /**
+     * Creates a node for the given application, marking its existence.
+     */
+    // TODO: Remove in Vespa 9
+    public void createApplicationInOldFormat(ApplicationId id) {
```
hakonhall, line 94: Why cannot this method be removed now?  The only reference to this method is in the test below.

hakonhall (approved):

configserver/src/main/java/com/yahoo/vespa/config/server/application/ApplicationCuratorDatabase.java
hmusum: There are config server out there (self-hosted) that have data in old format and need to be able to read it when upgrading

PR #28555, opened 2023-09-18
Hmusum/support download of files from s3

hmusum:


config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloadRpcServer.java
```diff
@@ -72,60 +68,40 @@ private void download(Request req) {
...
-    private static void downloadFile(Request req, HttpURLConnection connection, File downloadDir) throws IOException {
-        long start = System.currentTimeMillis();
-        String url = connection.getURL().toString();
-        Files.createDirectories(downloadDir.toPath());
-        File contentsPath = new File(downloadDir, CONTENTS_FILE_NAME);
-        try (ReadableByteChannel rbc = Channels.newChannel(connection.getInputStream())) {
-            try (FileOutputStream fos = new FileOutputStream((contentsPath.getAbsolutePath()))) {
-                fos.getChannel().transferFrom(rbc, 0, Long.MAX_VALUE);
+    private static Downloader downloader(String url) {
+        return url.startsWith("s3://") ? new S3Downloader() : new UrlDownloader();
```
freva, line 94: Consider converting this to `URI`, then this could be a `switch` on `uri.getScheme()`. I guess `UrlDownloader` only supports `http`/`https`? Rest should fail.

In `S3Downloader` the bucket name will be `url.getHost()`.

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/S3Downloader.java
```diff
@@ -0,0 +1,76 @@
...
+                .withCredentials(new CredentialsProvider())
+                .build();
+    }
+
+    @Override
+    public Optional<File> downloadFile(String url, File targetDir) throws IOException {
+        AmazonS3URI s3URI = new AmazonS3URI(url);
+        S3Object s3Object = s3Client.getObject(s3URI.getBucket(), s3URI.getKey());
+        File file = new File(targetDir, fileName());
+        Files.copy(s3Object.getObjectContent(), file.toPath());
```
freva, line 40: `UrlDownloader` also does `Files.createDirectories(downloadDir.toPath());`. Is this needed here/not needed there?

freva (approved):

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloadRpcServer.java
hmusum: Nice, fixed

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/S3Downloader.java
hmusum: Fixed

PR #28558, opened 2023-09-18
Levenshtein DFA successor generation optimization and UTF-32 output

vekterli:
@geirst please review
@toregge @havardpe FYI

This contains two separate optimizations.

99fd8f3e9ca9677ce25401ef606506aeecd824e6 optimizes successor suffix string generation:

Avoids explicitly stepping the DFA states and handling each transition character separately by detecting the case where the only way the generated suffix can possibly match is for it to _exactly_ match the remaining target string suffix.

This is the case when the sparse cost matrix row only has 1 column remaining, and this column is equal to the max edit distance. I.e. no more edits can possibly be done.

In local synthetic benchmarks this speeds up successor generation 4x when the target string is 64 characters.

This optimization is currently only enabled for the _implicit_ DFA type.

6ad887593b100962cf4c7186deaa598720b8f555 adds support for raw UTF-32 successor string output:

Avoids needing to encode UTF-32 characters to UTF-8 internally, as this is likely to be subsequently reversed by a caller that itself operates on UTF-32 code points.

**API change:** the `match()` function is split into one single-argument overload that does not produce a successor, and two overloads for UTF-8 and UTF-32 successor generation (now by _ref_, not pointer). This removes `nullptr` usage from the API (always a good thing).

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/dfa_fuzzy_matcher.h
+++ b/searchlib/src/vespa/searchlib/attribute/dfa_fuzzy_matcher.h
@@ -25,7 +25,7 @@ public:
 
     template <typename DictionaryConstIteratorType>
     bool is_match(const char* word, DictionaryConstIteratorType& itr, const DfaStringComparator::DataStoreType& data_store) {
-        auto match = _dfa.match(word, &_successor);
+        auto match = _dfa.match(word, _successor);
         if (match.matches()) {
             return true;
         } else {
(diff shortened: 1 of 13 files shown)


vekterli:
Note: the API change will conflict with #28557, so whichever PR ends up last in line should be modified to update the affected `match()` call-site.

vekterli:
Rebased onto master and resolved API conflict in 4a004d2e8d472c660cc9c2b73ee637c5247968da

geirst (approved):
👍

PR #28560, opened 2023-09-18
Use make_for_lookup() member function on existing comparator

toregge:
to make a new comparator which is used for lookup.

@geirst : please review


geirst (approved):
👍

PR #28567, opened 2023-09-18
Handle the exception that will come when codepoints < 'from' < len. MERGEOK

baldersheim:
Already done and tested for 'to'.

@bjorncs or @bratseth PR


bjorncs (approved):
Consider using `String.codePointCount(int,int)` to validate `fromCP`

baldersheim:
Considered, but doing so would double the work as you would have to iterate twice.

PR #28568, opened 2023-09-18
Add a configserver with wg key, but missing timetamp in node repo

gjoranv:


freva (approved):
Now that this has rolled out, it shouldn't be possible to have wg timestamp missing if the key is there, so maybe just enforce that? 
Ideally the REST API and the NodeSpec should expose a single Optional<Wireguard> that has the key and the timestamp

PR #28572, opened 2023-09-19
Stop overriding equal member function in EnumStoreStringComparator.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #28573, opened 2023-09-20
Ignore count during GPU accumulation

olaaun:

FYI @oyving

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
@@ -150,7 +150,9 @@ public class NodeResources {
         }
 
         public GpuResources minus(GpuResources other) {
-            return new GpuResources(this.count - other.count, this.memoryGb - other.memoryGb);
+            var thisMem = this.count() * this.memoryGb();
+            var otherMem = other.count() * other.memoryGb();
+            return new NodeResources.GpuResources(1, thisMem - otherMem);
         }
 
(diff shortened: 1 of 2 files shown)


bratseth:
Shouldn't we fix the plus method in NodeResources instead?

olaaun:
> Shouldn't we fix the plus method in NodeResources instead?

Seems like it's used for several things, was unsure if other method callers expect count > 1. It'd be great if we can just change `NodeResources::plus` instead

bratseth:
@mpolden should review this ...

config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
```diff
@@ -144,7 +144,9 @@ private boolean lessThan(GpuResources other) {
         public static GpuResources getDefault() { return zero; }
 
         public GpuResources plus(GpuResources other) {
-            return new GpuResources(this.count + other.count, this.memoryGb + other.memoryGb);
+            var thisMem = this.count() * this.memoryGb();
+            var otherMem = other.count() * other.memoryGb();
+            return new NodeResources.GpuResources(1, thisMem + otherMem);
```
mpolden, line 149: Shouldn't `minus` have the same change?

mpolden (dismissed):

bratseth:
We're missing some unit tests here.

olaaun:
Added test in https://github.com/vespa-engine/vespa/pull/28587

PR #28574, opened 2023-09-20
Avoid incorrect match for brute force case sensitive prefix search.

toregge:
@geirst : please review
@baldersheim : FYI


geirst (approved):
👍

PR #28586, opened 2023-09-20
Kkraune/fix links

kkraune:
- based on htmlchecker output + spellchecker


kkraune:
sorry for the big PR, some autoformatting from intellij here. the gist of it is, add a screwdriver job for auto link check of .html and .md, and fix most of the broken links/html files + spellcheck

aressem (approved):

PR #28588, opened 2023-09-20
Posting list search contexts:

toregge:
Rename useThis() to use_dictionary_entry() and step iterator one or more steps when not using dictionary entry.

@geirst : please review


geirst (approved):
👍

PR #28598, opened 2023-09-21
Split core DFA match loop into match-only and successor-emitting specializations

vekterli:
@geirst please review
@havardpe FYI

This allows for "hybrid" schemes where raw matching (without successor generation) is done with a dedicated matcher implementation that is faster for that particular purpose.

Also gives a much tighter loop for the match-only case and removes some branches from the successor-emitting case.


geirst (approved):
👍

PR #28602, opened 2023-09-21
Check for soft doom before doing potentially expensive initRange

baldersheim:
@havardpe PR


baldersheim:
Ah, this explain the strange name of #28668. Push it was indeed too many branches...

PR #28608, opened 2023-09-21
Add IPv6 to public zonal endpoints in GCP

hakonhall:
Adds an IPv6 address to load balancers in front of an application cluster in a zone (zone lb), if the endpoint is public. In particular:
 - The endpoint may or may not have been specified as a private service connect endpoint.  It looks like service connect requires an private regional IPv4 address at the moment and may be disjoint from the public IPv4 endpoint.  The public IPv4 endpoint may be disabled separately in deployment.xml.  Only the public IPv4 gets a sibling IPv6 address, both pointing to a target TCP proxy, etc.
 - The private service connect endpoint is associated with a private IPv4 address.  I have not added a private IPv6 address at this point.  That may or may not be something we want.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerInstance.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerInstance.java
@@ -21,7 +21,8 @@ import java.util.Set;
 public class LoadBalancerInstance {
 
     private final Optional<DomainName> hostname;
-    private final Optional<String> ipAddress;
+    private final Optional<String> ip4Address;
+    private final Optional<String> ip6Address;
     private final Optional<DnsZone> dnsZone;
     private final Set<Integer> ports;
     private final Set<String> networks;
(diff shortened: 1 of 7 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/lb/LoadBalancerInstance.java
```diff
@@ -43,9 +45,9 @@ public LoadBalancerInstance(Optional<DomainName> hostname, Optional<String> ipAd
         this.serviceIds = List.copyOf(Objects.requireNonNull(serviceIds, "private service id must be non-null"));
         this.cloudAccount = Objects.requireNonNull(cloudAccount, "cloudAccount must be non-null");
 
-        if (hostname.isEmpty() == ipAddress.isEmpty()) {
+        if (hostname.isEmpty() == ip4Address.isEmpty()) {
             throw new IllegalArgumentException("Exactly 1 of hostname=%s and ipAddress=%s must be set".formatted(
```
freva, line 49:
```suggestion
            throw new IllegalArgumentException("Exactly 1 of hostname=%s and ip4Address=%s must be set".formatted(
```

freva (dismissed):

freva (approved):

PR #28615, opened 2023-09-22
Feature flag allows specifying metric set

olaaun:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
@@ -147,9 +147,9 @@ public class PermanentFlags {
             "Takes effect on next deployment from controller",
             APPLICATION_ID);
 
-    public static final UnboundBooleanFlag SEND_LIMITED_METRIC_SET = defineFeatureFlag(
-            "send-limited-metric-set", true,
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
```diff
@@ -147,9 +147,9 @@ public class PermanentFlags {
             "Takes effect on next deployment from controller",
             APPLICATION_ID);
 
-    public static final UnboundBooleanFlag SEND_LIMITED_METRIC_SET = defineFeatureFlag(
-            "send-limited-metric-set", true,
-            "Whether a limited metric set should be fetched from metrics-proxy (CD systems only)",
+    public static final UnboundStringFlag METRIC_SET = defineStringFlag(
+            "metric-set", "",
```
yngveaasheim, line 151: Don't we want a default value here?

yngveaasheim (dismissed):

flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
olaaun: An empty default would lead to the same behavior as before, but I'll add `Vespa` as default to make things clear

freva (approved):

PR #28631, opened 2023-09-23
- Single filter terms can be lifted out from weighted sets.

baldersheim:
@havardpe PR


searchlib/src/vespa/searchlib/attribute/attribute_weighted_set_blueprint.cpp
```diff
@@ -157,6 +157,10 @@ AttributeWeightedSetBlueprint::createLeafSearch(const fef::TermFieldMatchDataArr
     assert(tfmda.size() == 1);
     assert(getState().numFields() == 1);
     fef::TermFieldMatchData &tfmd = *tfmda[0];
+    bool field_is_filter = getState().fields()[0].isFilter();
+    if (field_is_filter && (_contexts.size() == 1)) {
+        return _contexts[0]->createIterator(&tfmd, strict);
```
havardpe, line 162: might produce different match information than what would be expected from the weighted set, if the inner iterator is not itself a filter iterator.

havardpe (approved):

searchlib/src/vespa/searchlib/attribute/attribute_weighted_set_blueprint.cpp
baldersheim: I think it will produce at least what is required.

PR #28633, opened 2023-09-24
Avoid random lookup to determine if attribute has undefined value for…

baldersheim:
… a document.

Instead of comparing value against undefined value, compare enum reference.

@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/singleenumattribute.h
+++ b/searchlib/src/vespa/searchlib/attribute/singleenumattribute.h
@@ -110,6 +110,9 @@ public:
     void onUpdateStat() override;
     void reclaim_memory(generation_t oldest_used_gen) override;
     void before_inc_generation(generation_t current_gen) override;
+    bool isUndefined(DocId doc) const override {
+        return acquire_enum_entry_ref(doc) == this->getEnumStore().get_default_value_ref().load_relaxed();
+    }
     EnumHandle getEnum(DocId doc) const override {
        return getE(doc);
     }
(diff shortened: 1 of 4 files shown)


toregge (approved):

searchlib/src/vespa/searchlib/attribute/singleenumattribute.h
```diff
@@ -110,6 +110,9 @@ class SingleValueEnumAttribute : public B, public SingleValueEnumAttributeBase {
     void onUpdateStat() override;
     void reclaim_memory(generation_t oldest_used_gen) override;
     void before_inc_generation(generation_t current_gen) override;
+    bool isUndefined(DocId doc) const override {
+        return acquire_enum_entry_ref(doc) == this->getEnumStore().get_default_value_ref().load_relaxed();
```
toregge, line 114: Note that during enum store compaction, isUndefined() might return false when it should return true, due to acquire_enum_entry_ref(doc) returning an EntryRef referencing the moved default value, while get_default_value_ref() returns an EntryRef referencing the original default value.

toregge (changes requested):
I do not think we should accept IsUndefined() returning false when it should return true.

searchlib/src/vespa/searchlib/attribute/singleenumattribute.h
baldersheim: hmm, that is probably something we do not want to allow.
Lets discuss options tomorrow.
Could the default_value_ref be fixed for the application lifetime ?

baldersheim:
Will not pursue this one until further proof is found as to its cost.

PR #28643, opened 2023-09-25
Reapply "Validate use of s3 urls in config"

hmusum:


config-model/src/main/java/com/yahoo/vespa/model/application/validation/UrlConfigValidator.java
```diff
@@ -0,0 +1,49 @@
...
+                .stream()
+                .flatMap(hostResource -> hostResource.spec().membership().stream())
+                .filter(membership -> membership.cluster().id().equals(cluster.id()))
+                .anyMatch(membership -> membership.cluster().isExclusive());
+    }
+
+    private static void validateS3UlsInConfig(DeployState state, ApplicationContainerCluster cluster, boolean isExclusive) {
+        if (hasUrlInConfig(cluster)) {
+            // TODO: Would be even better if we could add which config/field the url is set for in the error message
+            String message = "Found s3:// urls in config for container cluster " + cluster.getName();
```
freva, line 36: Now that `hasUrlInConfig()` does not check for scheme, we should probably make the error message scheme-neutral too.

freva (dismissed):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/UrlConfigValidator.java
hmusum: Changed this to only consider s3 urls in validator

freva (approved):

PR #28649, opened 2023-09-25
Add another is_match member function to dfa fuzzy matcher that doesn't

toregge:
try to update directory iterator.

@geirst : please review


geirst (approved):
👍

PR #28651, opened 2023-09-25
Generate Levenshtein successor prefix "as we go" during match loop

vekterli:
@geirst please review
@toregge @havardpe FYI

Moving away from on-demand/deferred generation means that we only decode (and possibly case-normalize) UTF-8 characters _once_ instead of twice in the case of UTF-32 output or non-normalized input.

These changes make it much more easy to add future support for preserving a caller-supplied successor prefix, which would be used for prefix-locked dictionary matching. It also makes the successor logic itself significantly less hairy.

Note: this subtly changes the API to potentially _always_ mutate the input successor string. The API documentation has been updated to reflect this. No current users of the API should be affected.


geirst (approved):
👍

PR #28654, opened 2023-09-25
- Return early in doSeek if docId found.

baldersheim:
- get_hits should not match anything before its current position.
- Cache bitvector size as a atomic does not optimize that well.

@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/document_weight_or_filter_search.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/document_weight_or_filter_search.cpp
@@ -32,7 +32,7 @@ public:
     }
 
     std::unique_ptr<BitVector> get_hits(uint32_t begin_id) override {
-        return _children.get_hits(begin_id, getEndId());
+        return _children.get_hits(std::min(getEndId(), std::max(begin_id, getDocId())), getEndId());
     }
 
     Trinary is_strict() const override { return Trinary::True; }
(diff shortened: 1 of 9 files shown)


searchlib/src/vespa/searchlib/attribute/document_weight_or_filter_search.cpp
```diff
@@ -32,7 +32,7 @@ class DocumentWeightOrFilterSearchImpl : public DocumentWeightOrFilterSearch
     }
 
     std::unique_ptr<BitVector> get_hits(uint32_t begin_id) override {
-        return _children.get_hits(begin_id, getEndId());
+        return _children.get_hits(std::min(getEndId(), std::max(begin_id, getDocId())), getEndId());
```
havardpe, line 35: this gives me bad memories

havardpe (commented):
are we sure we want different sized bitvectors for the same docid range?

searchlib/src/vespa/searchlib/attribute/document_weight_or_filter_search.cpp
baldersheim: It failed a conformance test without. I will check for alternative options.

baldersheim: Fixed

PR #28667, opened 2023-09-26
Report customer status periodically

oyving:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/BillingReportMaintainer.java
```diff
@@ -2,23 +2,83 @@
...
+
+    private List<TenantName> billableTenants(List<TenantName> tenants) {
+        return billablePlans().stream()
+                .collect(Collectors.toMap(
+                        p -> p,
+                        p -> billing.tenantsWithPlan(tenants, p.id())))
+                .values()
+                .stream()
+                .flatMap(Collection::stream)
+                .toList();
```
gjoranv, line 81: Can't this be simplified to:
```
        return billablePlans().stream()
                .flatMap(p -> billing.tenantsWithPlan(tenants, p.id()).stream())
                .toList();
```

gjoranv (dismissed):

gjoranv (approved):

PR #28529, opened 2023-09-14
LuceneLinguistics optional `configDir`

dainiusjocas:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

This PR makes the previously required `configDir` parameter optional.
Making it optional simplifies setups that just want to use LuceneLinguistics without any analyzers created via XML configuration.

A required Vespa version is 8.224.19.

For the application packages created with earlier versions it should not introduce any breaking changes.
However, the tests might break if you're building the `LuceneAnalysisConfig` class, the `configDir` type has changed to optional.

Related [issue](https://github.com/vespa-engine/vespa/issues/27912).

PR #28682, opened 2023-09-27
LuceneLinguistics: optionaly consider StemMode as an analysis key

dainiusjocas:

Considering a `StemMode` as a part of analysis key allows to have more than one analysis configuration per language. There are 5 stem modes. So, 6 analysis configurations per language are possible. E.g. for English language there are 6 valid keys: `['en', 'en/ALL', 'en/BEST', 'en/DEFAULT', 'en/NONE', 'en/SHORTEST']` (note case sensitivity). 

The `/` is used as a separator so that it doesn't clash with the `ComponentId` specification with versions.

This is not a breaking change.


geirst:
@hmusum can you review this (as you also looked at the last PR https://github.com/vespa-engine/vespa/pull/28529).

hmusum (approved):
Nice! 👍

PR #28700, opened 2023-09-27
Bratseth/pricing

bratseth:


hmusum:
Adapted and added to internal repo

PR #28701, opened 2023-09-28
Add public APIs and abi-spec.json to config-model-api

hmusum:


config-model-api/src/main/java/com/yahoo/config/application/api/package-info.java
```diff
@@ -1,5 +1,7 @@
 // Copyright Yahoo. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 @ExportPackage
+@PublicApi
```
baldersheim, line 3: Add a comment that this is not a real public api, just an internal one

baldersheim (dismissed):
Add comment  in all places there are public api annotations

hmusum:
Yes, good point, will do

baldersheim (approved):

PR #28706, opened 2023-09-28
use equals to compare OptionalInt instances MERGEOK

arnej27959:
@kkraune please review
@jonmv @bratseth FYI


kkraune (approved):
👍

baldersheim (approved):

baldersheim:
One reason why optional should be used with care...

bratseth:
Thanks!

PR #28710, opened 2023-09-28
Add comments about stuff that are mandatory for all threads in a thre…

baldersheim:
…ad bundle to do.

Either all do it or none.

@havardpe PR


searchcore/src/vespa/searchcore/proton/matching/match_thread.cpp
```diff
@@ -480,6 +498,7 @@ MatchThread::run()
                         result->getNumHits(),
                         resultContext->sort->hasSortData(),
                         bool(resultContext->grouping)));
+        (void) processToken;  // Avoid unused warning
```
havardpe, line 501: we should not have to ignore guards

PR #28713, opened 2023-09-28
Metric doc generator

olaaun:
Intended to replace [metric_reference_generator.py](https://github.com/vespa-engine/documentation/blob/master/metric_reference_generator.py), which is prone to breakage from refactoring. E.g. when extracting common metrics to separate classes 95c2e6060d431e9b281d7d69242d4779a267baa3


gjoranv (approved):
If possible, we should have unit tests that fail if we break doc generation somehow.

kkraune (commented):
👍

PR #28714, opened 2023-09-28
dump table_dfa as actual dfa in graphviz

havardpe:
enumerate states based on best-edge-first
'*' means any character without its own edge

@vekterli please review


vekterli (approved):
:+1:

PR #28715, opened 2023-09-28
Propagate GPU resources when determining current resources for autoscaling

freva:
Fixes VESPA-27581: If a cluster is scaled ideally or as good as possible within the configured limits, and has GPU configured, the API will always return that the cluster is currently being rescaled, even though no resource change is scheduled.

* `/nodes/v2/application/<app>` will return identical "target" and "current". The later being the `NodeList::toResources` on `active`, non-retired nodes in the cluster.
* In `Activator::rememberResourceChange`, "current" resources are defined the same way, so because "target" is equal to "current", the target is reset every time the application is activated.
* On the next iteration of `AutoscalingMaintainer`, `Autoscaling::autoscale` will set the "target" back to the same resources as "current" (since no better configuration exists).
  * In this case "current" resources are determined using `AllocatableResources::averageRealResourcesOf`, which did not set GPU resources. I'm not sure how to calculate the average GPU resources, for now I'm just copying over from requested resources which should be OK since there is no autoscaling over GPU resources and GPU real resources being equal to advertised?
  * Because the "current" did not include GPU,  `Autoscaling::worthRescaling` will return true as the cost difference becomes significant.

Either way, why are we comparing real resources (current) to advertised resources (target) when determining if a change is worth rescaling? Something is wrong here.


mpolden (commented):
> I'm not sure how to calculate the average GPU resources, for now I'm just copying over from requested resources which should be OK since there is no autoscaling over GPU resources and GPU real resources being equal to advertised?

I think this should be fine, but wouldn't average simply be total GPU memory GB / node count?

> Either way, why are we comparing real resources (current) to advertised resources (target) when determining if a change is worth rescaling? Something is wrong here.

Where do you see this?

freva:
> wouldn't average simply be total GPU memory GB / node count?

Assuming GPU count is 1 or that we don't care that it is normalized to 1, yes.

> Where do you see this?

Never mind, both are real. Now I'm not sure why we calculate real resources using average here
https://github.com/vespa-engine/vespa/blob/5b3c08a08be7984fc7bceaf6e56619c2c81da34d/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/AllocatableResources.java#L50
meanwhile here 
https://github.com/vespa-engine/vespa/blob/5b3c08a08be7984fc7bceaf6e56619c2c81da34d/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/AllocatableResources.java#L41
we get real by using calculator on advertised. Shouldn't they both do the same?

freva:
Changing from average to calculator.requestToReal(advertised) causes small change in allocation that breaks 29 tests... :tired_face:

mpolden (approved):

bratseth:
> Now I'm not sure why we calculate real resources using average here

The reasoning is that this is used as input to the model which determines if the cluster is properly loaded, using average metrics from the nodes, and so it should reflect the average resources actually available, not what we get in theory from translating the requested resources to real.

PR #28723, opened 2023-09-29
Lift out single iterators if they are leafs and tfmd is not needed.

baldersheim:
@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
@@ -96,8 +96,13 @@ SearchIterator::UP
 WeightedSetTermBlueprint::createLeafSearch(const fef::TermFieldMatchDataArray &tfmda, bool) const
 {
     assert(tfmda.size() == 1);
-    if ((_terms.size() == 1) && tfmda[0]->isNotNeeded() && _terms[0]->isLeaf()) {
-        return static_cast<LeafBlueprint &>(*_terms[0]).createLeafSearch(tfmda, true);
+    if ((_terms.size() == 1) && tfmda[0]->isNotNeeded()) {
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
```diff
@@ -96,11 +96,15 @@ SearchIterator::UP
 WeightedSetTermBlueprint::createLeafSearch(const fef::TermFieldMatchDataArray &tfmda, bool) const
 {
     assert(tfmda.size() == 1);
+    if ((_terms.size() == 1) && tfmda[0]->isNotNeeded() && _terms[0]->isLeaf()) {
```
havardpe, line 99: consider doing this as an actual optimization

searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
```diff
@@ -96,11 +96,15 @@ SearchIterator::UP
...
+        return static_cast<LeafBlueprint &>(*_terms[0]).createLeafSearch(tfmda, true);
```
havardpe, line 100: need a good comment here to explain why we cannot preserve the iterator strictness

searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
baldersheim: Turned out that was not possible due to mismatch of handle vs matchdata mismatch.

searchlib/src/vespa/searchlib/queryeval/weighted_set_term_blueprint.cpp
baldersheim: Comment added. PTAL

havardpe (dismissed):

PR #28732, opened 2023-09-29
Jonmv/token tell handler

jonmv:
@bjorncs or @tokle please review. 
@ldalves FYI, token fingerprints will get an additional `"state"` field, and _may_ have only the `"fingerprint"` and `"state"` fields, if deleted, but still present on some nodes. See the unit test for an example payload:
```json 
                                  {
                                    "tokens": [
                                      {
                                        "id": "myTokenId",
                                        "versions": [
                                          {
                                            "fingerprint": "%s",
                                            "created": "2020-09-13T12:26:40Z",
                                            "author": "user@test",
                                            "expiration": "2020-10-13T12:26:40Z",
                                            "state": "active"
                                          },
                                          {
                                            "fingerprint": "ff:01",
                                            "state": "deactivating"
                                          }
                                        ]
                                      }
                                    ]
                                  }
```
The possible `"state"` values are `"active"`, `"deploying"`, "`deactivating"`.

### DIFF
--- a/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/application/v4/model/DeploymentData.java
+++ b/controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/application/v4/model/DeploymentData.java
@@ -39,7 +39,7 @@ public class DeploymentData {
     private final List<TenantSecretStore> tenantSecretStores;
     private final List<X509Certificate> operatorCertificates;
     private final Supplier<Optional<CloudAccount>> cloudAccount;
-    private final List<DataplaneTokenVersions> dataPlaneTokens;
+    private final Supplier<List<DataplaneTokenVersions>> dataPlaneTokens;
     private final boolean dryRun;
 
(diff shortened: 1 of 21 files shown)


jonmv:
Hmm, come to think of it, we _do_ need to store the _tokens_ in use by each deployment, for retriggering. That'll let us differentiate between tokens which aren't referenced in any deployment (`"unused"`), and those that are, but just aren't present anywhere yet (`"deploying"`).

jonmv:
If someone wants different names for these states, just holler, btw. Perhaps "inactive", "activating", "active", "deactivating"?

bjorncs:
My suggestion is `revoking` and `revoked` which common nomenclature for invalidating credentials

bjorncs (dismissed):

jonmv:
Hmm. A token could be created, but not in use anywhere, though. Not yet. We need a name for that as well. Versions that are completely revoked are not shown, because they are, indeed, irrelevant.

tokle (dismissed):

bjorncs (approved):

PR #28736, opened 2023-09-29
- Use asBitVectorIterator instead of isBitVector + casting to present…

baldersheim:
… a BitVectorIterator interface.

- Allow Filter wrapper to expose underlying BitVector.
   - This ensures that the bitvectors are handled first during termwise evaluation, as they have a constant cost and will reduce the cost for the ones coming later on.

@havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
+++ b/searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
@@ -55,6 +55,9 @@ public:
     BitVector::UP get_hits(uint32_t begin_id) override {
         return _wrapped_search->get_hits(begin_id);
     }
+    BitVectorIterator * asBitVector() noexcept override {
+        return _wrapped_search->asBitVector();
+    }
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/queryeval/filter_wrapper.h
```diff
@@ -55,6 +55,9 @@ class FilterWrapper : public SearchIterator {
     BitVector::UP get_hits(uint32_t begin_id) override {
         return _wrapped_search->get_hits(begin_id);
     }
+    BitVectorIterator * asBitVector() noexcept override {
+        return _wrapped_search->asBitVector();
```
havardpe, line 59: using the inner iterator will invalidate the outer one. I do not think this is a good idea.

baldersheim: Very good point, fixed PTAL

havardpe (approved):

PR #28740, opened 2023-09-29
Require at least 50% of all canaries for normal, of defaults for high

jonmv:
@mpolden please review.


mpolden (commented):
LGTM, but needs a test.

jonmv:
Yes, probably a good idea :)

PR #28758, opened 2023-10-02
Fail if class files are built for JDK version > 17.

gjoranv:


bjorncs (dismissed):
Approved assuming this does not break class analysis on multi-release JARs

gjoranv:
PTAL

bjorncs (approved):

PR #28759, opened 2023-10-02
Read application id through SessionData

hmusum:
Read from old or new path (session data in one blob) based on feature flag, fallback to reading from old path if it fails

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
@@ -63,9 +63,8 @@ public class SessionSerializer {
             try {
                 return zooKeeperClient.readSessionData();
             } catch (Exception e) {
-                log.log(WARNING, "Unable to read session dato for session " + zooKeeperClient.sessionId() +
+                log.log(WARNING, "Unable to read session data for session " + zooKeeperClient.sessionId() +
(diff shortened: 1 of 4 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
```diff
@@ -53,20 +59,30 @@ void write(SessionZooKeeperClient zooKeeperClient, ApplicationId applicationId,
...
-                                   zooKeeperClient.readOperatorCertificates(),
-                                   zooKeeperClient.readCloudAccount(),
-                                   zooKeeperClient.readDataplaneTokens());
+        if (readSessionData.value() && zooKeeperClient.sessionDataExists())
+            try {
+                return zooKeeperClient.readSessionData();
+            } catch (Exception e) {
+                log.log(WARNING, "Unable to read session dato for session " + zooKeeperClient.sessionId() +
+                        ": " + Exceptions.toMessageString(e));
+                readSessionDataFromLegacyPaths(zooKeeperClient);
```
hakonhall, line 68: Unnecessary read?

hmusum: Yeah, I guess I can just let if fall through, will remove it

hakonhall (approved):

PR #28760, opened 2023-10-02
Check query timeout in SearchHandler

jonmv:
@bratseth please review.


jonmv:
As far as I understand, this will also respect timeouts set in query profiles.

bratseth (approved):

PR #28785, opened 2023-10-04
Add support for persisting plan in ZooKeeper for a tenant

hmusum:


controller-server/src/test/java/com/yahoo/vespa/hosted/controller/notification/NotificationsDbTest.java
```diff
@@ -27,6 +27,7 @@
 import com.yahoo.vespa.hosted.controller.tenant.LastLoginInfo;
 import com.yahoo.vespa.hosted.controller.tenant.TenantContacts;
 import com.yahoo.vespa.hosted.controller.tenant.TenantInfo;
+import org.apache.zookeeper.Op;
```
freva, line 30:
```suggestion
```

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/persistence/TenantSerializer.java
```diff
@@ -215,7 +217,10 @@ private CloudTenant cloudTenantFrom(Inspector tenantObject) {
         Instant tenantRolesLastMaintained = SlimeUtils.instant(tenantObject.field(tenantRolesLastMaintainedField));
         List<CloudAccountInfo> cloudAccountInfos = cloudAccountsFromSlime(tenantObject.field(cloudAccountsField));
         Optional<BillingReference> billingReference = billingReferenceFrom(tenantObject.field(billingReferenceField));
-        return new CloudTenant(name, createdAt, lastLoginInfo, creator, developerKeys, info, tenantSecretStores, archiveAccess, invalidateUserSessionsBefore, tenantRolesLastMaintained, cloudAccountInfos, billingReference);
+        Optional<String> planId = planId(tenantObject.field(planIdField));
```
freva, line 220: Consider simplifying with something like
```suggestion
        Optional<String> planId = SlimeUtils.optionalString(tenantObject.field(planIdField));
```

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/tenant/CloudTenant.java
```diff
@@ -28,13 +28,15 @@ public class CloudTenant extends Tenant {
     private final ArchiveAccess archiveAccess;
     private final Optional<Instant> invalidateUserSessionsBefore;
     private final Optional<BillingReference> billingReference;
+    private final Optional<String> planId;
```
freva, line 31: Use `PlanId` type?

freva (dismissed):
Maybe we should default to `none` type in ZK deserialization code so we can avoid the `Optional`? The future maintainer will update to the actual plan before we start using this anyway.

controller-api/src/main/java/com/yahoo/vespa/hosted/controller/tenant/CloudTenant.java
hmusum: Yep, done

freva (approved):

PR #28786, opened 2023-10-04
Ensure endpoint is generated for all requested auth methods

mpolden:
@tokle


controller-api/src/main/java/com/yahoo/vespa/hosted/controller/api/integration/certificates/EndpointCertificate.java
```diff
@@ -13,7 +13,7 @@ public record EndpointCertificate(String keyName, String certName, int version,
                                   String rootRequestId, // The id of the first request made for this certificate. Should not change.
                                   Optional<String> leafRequestId, // The id of the last known request made for this certificate. Changes on refresh, may be outdated!
                                   List<String> requestedDnsSans, String issuer, Optional<Long> expiry,
-                                  Optional<Long> lastRefreshed, Optional<String> randomizedId) {
+                                  Optional<Long> lastRefreshed, Optional<String> generatedId) {
 
     public EndpointCertificate withRandomizedId(String randomizedId) {
```
tokle, line 18: change this as well?

PR #28796, opened 2023-10-04
- Instead of keeping a map of bucketId => lids, just append everythin…

baldersheim:
…g to a vector and sort when complete.

- This significantly improves memory usage during compaction. Instead of many heap allocations
- You now get fewer mmapped allocations that are dropped when done.

@geirst or @toregge PR


geirst (approved):
👍

PR #28800, opened 2023-10-05
- Reduce max lids per file and max file size to 4M and 256M during un…

baldersheim:
…it testing.

- Reduce max lids from 40M to 8M as default configuration.

@geirst PR


geirst (approved):
👍  Before merge, validate that the compaction code handles that `maxnumlids` is adjusted down.

PR #28802, opened 2023-10-05
- Number of partitions is fixed compile time => use std::array.

baldersheim:
- Use unique_ptr on outer object instead of unique_ptr on multiple non-movable inner objects.

@geirst PR


geirst (dismissed):
👍

PR #28808, opened 2023-10-05
Isolate inner jetty server component from wrapper with filters

jonmv:
@bjorncs please review.


container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/testutils/TestDriver.java
```diff
@@ -83,9 +88,10 @@ private static Module createBaseModule(ServerConfig serverConfig, ConnectorConfi
                 new AbstractModule() {
                     @Override
                     protected void configure() {
+                        bind(JettyHttpServer.class).in(Singleton.class);
```
bjorncs, line 91: 🎖️

bjorncs (dismissed):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/testutils/TestDriver.java
jonmv: 😜

PR #28838, opened 2023-10-09
andreer/add routing config diff to vespa log

andreer:


routing-generator/src/main/java/com/yahoo/vespa/hosted/routing/nginx/Nginx.java
```diff
@@ -104,12 +106,17 @@ private void loadConfig(int upstreamCount) throws IOException {
         Files.move(tempConfigPath, configPath, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
         metric.add(CONFIG_RELOADS_METRIC, 1, null);
         // Retry reload. Same rationale for retrying as in testConfig()
-        LOG.info("Loading new configuration file from " + configPath);
+        LOG.info("Loading new configuration file from " + configPath + (outputRoutingDiff ? " with diff:\n" + getDiff(configPath, tempConfigPath) : ""));
```
mpolden, line 109: `tempConfigPath` no longer exists at this point. Diff must be computed before the file is moved above. I also don't think `configPath` is guaranteed to exist the first time, e.g. on a new proxy node, so `getDiff` should confirm its existence before running the command.

routing-generator/src/main/java/com/yahoo/vespa/hosted/routing/RoutingGenerator.java
```diff
@@ -67,7 +66,8 @@ public RoutingGenerator(ZoneConfig zoneConfig, RoutingStatus routingStatus, Metr
                                                                        Sleeper.DEFAULT,
                                                                        Clock.systemUTC(),
                                                                        routingStatus,
-                                                                       metric),
+                                                                       metric,
+                                                                       zoneConfig.outputRoutingDiff()),
```
mpolden, line 70: I think this can simply inject `SystemInfo` to know the zone and environment.

mpolden (approved):

PR #28840, opened 2023-10-09
Flush memory indexes to disk then fusion disk indexes as soon as

toregge:
possible when enabling interleaved features.

@geirst : please review


geirst (dismissed):
👍

geirst (approved):
👍

PR #28846, opened 2023-10-10
Balder/deliver reply before sending next in sequence

baldersheim:
@vekterli or @bjorncs or @jonmv PR

### DIFF
--- a/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
+++ b/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
@@ -33,6 +33,9 @@ public class Sequencer implements MessageHandler, ReplyHandler {
         this.sender = sender;
         this.msn = msn;
     }
+    public Sequencer(MessageHandler sender) {
+        this(sender, null);
+    }
 
     /**
      * Sets the destroyed flag to true. The very first time this method is called, it cleans up all its dependencies.
--- a/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
+++ b/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
@@ -21,8 +21,7 @@ public class Sequencer implements MessageHandler, ReplyHandler {
     private final MessageHandler sender;
     private final Map<Long, Queue<Message>> seqMap = new HashMap<>();
     private final Messenger msn;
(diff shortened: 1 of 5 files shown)


jonmv (dismissed):

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
```diff
@@ -20,14 +20,21 @@ public class Sequencer implements MessageHandler, ReplyHandler {
     private final AtomicBoolean destroyed = new AtomicBoolean(false);
     private final MessageHandler sender;
     private final Map<Long, Queue<Message>> seqMap = new HashMap<>();
+    private final Messenger msn;
+    private final static Object BUSY = new Object();
+    private final static ThreadLocal<Object> isSending = new ThreadLocal<>();
```
vekterli, line 25: Would it suffice with `ThreadLocal<Boolean>` here without a separate sentinel object?

vekterli (dismissed):
:+1:

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
baldersheim: Yes, it would. Fixed

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
```diff
@@ -147,10 +162,16 @@ public void handleReply(Reply reply) {
             }
         }
         if (msg != null) {
-            sequencedSend(msg);
+            Boolean alreadySending = isSending.get();
+            if ((alreadySending == Boolean.TRUE) && (msn != null)) {
```
vekterli, line 166: Could probably use auto-(un)boxing  of bools for these if desired, i.e. `if (alreadySending && ...)`, `isSending.set(true)`. etc

vekterli (dismissed):

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
baldersheim: It will be null first time, unless .withInitial

baldersheim: Done

PR #28869, opened 2023-10-11
Jonmv/job runner thread metrics

jonmv:
@hakonhall

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
@@ -1,7 +1,9 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.hosted.controller.maintenance;
 
+import ai.vespa.metrics.ControllerMetrics;
 import com.yahoo.concurrent.DaemonThreadFactory;
+import com.yahoo.jdisc.Metric;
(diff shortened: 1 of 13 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
```diff
@@ -32,29 +37,37 @@ public class JobRunner extends ControllerMaintainer {
...
+    private final Metrics metrics;
 
     public JobRunner(Controller controller, Duration duration) {
-        this(controller, duration, Executors.newFixedThreadPool(32, new DaemonThreadFactory("job-runner-")), new InternalStepRunner(controller));
+        this(controller, duration, Executors.newFixedThreadPool(32, new DaemonThreadFactory("job-runner-")),
+             new InternalStepRunner(controller));
     }
 
     public JobRunner(Controller controller, Duration duration, ExecutorService executors, StepRunner runner) {
+        this(controller, duration, executors, runner, new Metrics(controller.metric(), Duration.ofMillis(100)));
```
hakonhall, line 48: Metrics gathering is done every 1 minute so the interval is typically set just shy of 60 seconds.  Why 100 ms?

hakonhall (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/maintenance/JobRunner.java
hakonhall: Metric.set has count, last, max, min, and sum.

hakonhall (approved):

PR #28870, opened 2023-10-11
Require zone-scoped endpoint in config-model

mpolden:
A container cluster in a hosted system should always have a zone endpoint. This
change adds a safe-guard (in `ApplicationContainerCluster`) to prevent
generating a routing config not containing required endpoints.

Large diff is due to the high number of tests that violated this new constraint.

@hmusum

FYI @tokle


hmusum (approved):
~Looks good, but I think we need a feature flag to avoid breaking upgrade of infrastructure if there exist apps that don't have zone-scoped endpoints~
(all apps have zone endpoints)

PR #28876, opened 2023-10-11
Revert "Ensure internal messages are flushed before shutting down RPC subsystem, take 2"

hmusum:
Reverts vespa-engine/vespa#28871

Some assertions in system tests still seen


vekterli (approved):
_sound of the saddest violin_

vekterli:
The underlying problem is symmetrical to the one solved by the reverted PR, and was caused by one of the storage chain components internally transitioning to a "closed" state one step earlier than I thought it would. Will be fixed in much the same way as the other issue.

PR #28878, opened 2023-10-11
- Use a Supplier<Identifable> to construct objects from id.

baldersheim:
- Provide lambda for default constructing frequently used objects to avoid reflection.

@bjorncs PR


baldersheim:
or @bratseth

bratseth (approved):
💯

vespajlib/src/main/java/com/yahoo/vespa/objects/Identifiable.java
```diff
@@ -186,6 +185,14 @@ protected static int registerClass(int id, Class<? extends Identifiable> spec) {
         registry.add(id, spec);
         return id;
     }
```
bjorncs, line 187: Consider deprecating

bjorncs (commented):
👍

PR #28880, opened 2023-10-11
Silly hack to close HTTP server before mbus server

jonmv:
@bjorncs please review.


bjorncs (approved):
Nice 👍 Consider adding some kind of priority concept to the `Server` interface

PR #28882, opened 2023-10-11
Info log message when MbusServer is finally destroyed

jonmv:
@baldersheim or @hmusum please review and merge.

### DIFF
--- a/messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
+++ b/messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
@@ -6,6 +6,7 @@ import com.yahoo.messagebus.Protocol;
 import com.yahoo.text.Utf8Array;
 
 import java.util.Deque;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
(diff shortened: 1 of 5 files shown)


messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
```diff
@@ -68,7 +69,7 @@ else if (owners.contains(owner))
 
     public void unregisterSession(String session, NetworkOwner owner, boolean broadcast) {
         sessions.computeIfPresent(session, (name, owners) -> {
-            if (owners.size() == 1 && owners.contains(owner)) {
+            if (owners.equals(List.of(owner))) {
```
hmusum, line 72: Will this equals() really do what's intended? (owners is a `Deque`)

Otherwise LGTM

hmusum (dismissed):

messagebus/src/main/java/com/yahoo/messagebus/network/NetworkMultiplexer.java
jonmv: Ah, thanks, that's why it was like it was in the first place!

jonmv:
```suggestion
            if (owners.size() == 1 && owners.contains(owner)) {
```

hmusum (approved):

PR #28895, opened 2023-10-12
Ensure messages are flushed prior to RPC shutdown, chapter III

vekterli:
@havardpe please review. One additional commit stacked onto the pile, the rest are unchanged.

Rejection of requests that happen concurrent with `CommunicationManager` shutdown is now explicitly allowed for the `Bouncer` component instead of triggering an invariant violation (unwanted in this particular case).

All previously failing system tests now pass locally, the incense has been lit and I have performed the secret rituals of the moose. This should mean success is all but guaranteed.

### DIFF
--- a/storage/src/tests/storageserver/bouncertest.cpp
+++ b/storage/src/tests/storageserver/bouncertest.cpp
@@ -51,9 +51,10 @@ struct BouncerTest : public Test {
             api::Timestamp timestamp,
             document::BucketSpace bucketSpace);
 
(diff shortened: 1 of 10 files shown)


havardpe (dismissed):
when stacking magic bools, a way to make things more readable is to use binary enums to add names to the different true/false values. This also has the added benefit of having different types for different toggles.

vekterli:
Sneaky bools of temptation have been replaced with type-safe enums, PTAL

havardpe (approved):

PR #28898, opened 2023-10-12
add global phase rerank count

arnej27959:
this follow the pattern from MatchPhase properties.
@bratseth please review
@havardpe FYI

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseRanker.java
+++ b/container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseRanker.java
@@ -107,7 +107,10 @@ public class GlobalPhaseRanker {
     }
 
     private int resolveRerankCount(GlobalPhaseSetup setup, Query query) {
-        if (setup == null) return 0;
+        if (setup == null) {
+            // there is no global-phase at all (ignore override)
+            return 0;
+        }
         Integer override = query.getRanking().getGlobalPhase().getRerankCount();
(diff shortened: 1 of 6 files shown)


geirst (commented):
Consider adding a test in container-search/src/test/java/com/yahoo/search/test/QueryTestCase.java, similar to how "trace.profiling..." is tested.

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseRanker.java
```diff
@@ -101,4 +106,12 @@ private Optional<GlobalPhaseSetup> globalPhaseSetupFor(Query query, String schem
                 .flatMap(evaluator -> evaluator.getGlobalPhaseSetup(query.getRanking().getProfile()));
     }
 
+    private int resolveRerankCount(GlobalPhaseSetup setup, Query query) {
+        if (setup == null) return 0;
```
bratseth, line 110: Should this be after the next line?

bratseth (commented):
LGTM, and yes I think we should have a test.

container-search/src/main/java/com/yahoo/search/ranking/GlobalPhaseRanker.java
arnej27959: no (added comment to clarify)

bratseth (approved):

PR #28899, opened 2023-10-12
Support price calculations with more than 1 app

hmusum:
Keeping the old API for now, to avoid breaking exiting usage, we can remove when not in use anymore

### DIFF
--- a/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/pricing/PricingApiHandler.java
+++ b/controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/pricing/PricingApiHandler.java
@@ -112,12 +112,12 @@ public class PricingApiHandler extends ThreadedHttpRequestHandler {
         List<ClusterResources> clusterResources = new ArrayList<>();
 
         for (Pair<String, String> entry : keysAndValues(elements)) {
-            switch (entry.getFirst()) {
-                case "committedSpend" -> committedSpend = parseDouble(entry.getSecond());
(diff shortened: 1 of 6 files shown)


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/pricing/PricingApiHandler.java
```diff
@@ -138,8 +179,44 @@ private ClusterResources clusterResources(String resourcesString) {
...
+        var enclaveVcpu = 0d;
+        var enclaveMemoryGb = 0d;
+        var enclaveDiskGb = 0d;
+        var enclaveGpuMemoryGb = 0d;
+
+        for (var element : keysAndValues(elements)) {
+            switch (element.getFirst().toLowerCase()) {
+                case "name" -> applicationName = element.getSecond();
+
+                case "vcpu" -> vcpu = parseDouble(element.getSecond());
```
freva, line 199: Do we need to go via `double` here` or can this be parsed straight to `BigDecimal`?

freva (dismissed):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/pricing/PricingApiHandler.java
hmusum: Yeah, I thought about that as well earlier today, will look into it in a later PR

freva (approved):

PR #28901, opened 2023-10-12
CLI documentation improvements

mpolden:
Mainly `vespa auth`, but I also tried to clarify some other things that might be
confusing.

@oyving


oyving (approved):

kkraune (approved):
Super! This is helpful for our users!

PR #28911, opened 2023-10-12
Jonmv/respect exclusivity in hcm

jonmv:
@freva and/or @hakonhall please review

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
@@ -4,11 +4,13 @@ package com.yahoo.vespa.hosted.provision.provisioning;
 import com.yahoo.component.Version;
 import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.ClusterMembership;
+import com.yahoo.config.provision.ClusterSpec;
 import com.yahoo.config.provision.Flavor;
 import com.yahoo.config.provision.NodeResources;
 import com.yahoo.config.provision.NodeType;
 import com.yahoo.vespa.hosted.provision.LockedNodeList;
 import com.yahoo.vespa.hosted.provision.Node;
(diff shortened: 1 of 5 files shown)


freva (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
```diff
@@ -559,4 +563,32 @@ public String toString() {
 
     }
 
+    public boolean violatesExclusivity(ClusterSpec cluster, ApplicationId application,
+                                       boolean exclusiveCluster, boolean hostSharing, NodeList allNodes) {
+        if (parentHostname().isEmpty()) return false;
+        if (type() != NodeType.tenant) return false;
+
+        // We always violate exclusivity if the parent is exclusive to someone else that the requesting application.
+        if ( ! emptyOrEqual(parent.flatMap(Node::exclusiveToApplicationId), application)) return true;
```
hakonhall, line 572: Why was this moved out from within the !hostSharing condition ?

jonmv: This was to ensure hosts with `exclusiveTo` set would _never_ be reused by someone else (including in zones where host-sharing is not allowed, but where the host is empty).

hakonhall: Moving it out of the !hostSharing condition doesn't affect zones where host-sharing is not allowed.  Host-sharing here means YAHOO.

This is likely a no-op because hosts do not have exclusiveToApplicationId in YAHOO.  Same with exclusiveToClusterType.

My point is that the 3 conditions under !hostSharing belong together, so either we should keep them under !hostSharing, or move all out.

jonmv: Then I think we should move them all out. It had me confused, at least.

PR #28919, opened 2023-10-13
allow configuring normalizers

arnej27959:
@havardpe for discussion


havardpe (dismissed):
Seems reasonable to me (disclaimer; not an expert on 'ranking expression transforms' or 'seeing through onnx models to verify its inputs')

bratseth (dismissed):
Great!

On exceptions, what we show to users is Exceptions.toMessageString(exception).
It might be better to assert on that to see the exact full message.
And since that method does 
e.getMessage() + ": " + e.getCause().getMessage() ...
we usually want each message to be capitalized.

PR #28921, opened 2023-10-13
adjust offset/hits if necessary to get enough input to global-phase

arnej27959:
@bratseth please review
@havardpe @geirst FYI


bratseth (approved):
👍

PR #28924, opened 2023-10-13
Jonmv/reapply balder break recusrion

jonmv:
@vekterli or @baldersheim please review and merge.

@baldersheim FYI.

### DIFF
--- a/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
+++ b/messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
@@ -20,14 +20,20 @@ public class Sequencer implements MessageHandler, ReplyHandler {
     private final AtomicBoolean destroyed = new AtomicBoolean(false);
     private final MessageHandler sender;
     private final Map<Long, Queue<Message>> seqMap = new HashMap<>();
+    private final Messenger msn;
+    private final static ThreadLocal<Boolean> isSending = ThreadLocal.withInitial(() -> Boolean.FALSE);
 
     /**
      * Constructs a new sequencer on top of the given async sender.
      *
      * @param sender The underlying sender.
      */
-    public Sequencer(MessageHandler sender) {
+    public Sequencer(MessageHandler sender, Messenger msn) {
         this.sender = sender;
+        this.msn = msn;
+    }
+    public Sequencer(MessageHandler sender) {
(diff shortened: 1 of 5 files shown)


jonmv:
No, hang on, this test doesn't fail if I add back 'discard()'.

jonmv:
There ...

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
```diff
@@ -148,7 +148,7 @@ private class SequencedSendTask implements Messenger.Task {
         private final Message msg;
         SequencedSendTask(Message msg) { this.msg = msg; }
         @Override public void run() { sequencedSend(msg); }
-        @Override public void destroy() { msg.discard(); }
```
jonmv: This is the fix.

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
```diff
@@ -148,7 +148,7 @@ private class SequencedSendTask implements Messenger.Task {
...
+        @Override public void destroy() { }
```
jonmv, line 151: This is the fix.

jonmv: `ResenderTask` does the same. I find it a bit weird that all tasks are destroyed. It would make more sense to me to just destroy anything left in the messenger queue upon _its_ destruction ...

jonmv: (Also, that wouldn't generate any replies, violating the claim I read somewhere that "message has guaranteed reply". This clearly isn't true 😂)

bratseth (approved):

messagebus/src/main/java/com/yahoo/messagebus/Sequencer.java
baldersheim: I wonder if we should remember if run has been called, and call discard if not.

jonmv: Yes, I wondered about that, too. I'm not too fond of worshipping false gods, but perhaps this one is real, and the sacrifice seems minor.

PR #28930, opened 2023-10-14
Balder/go directly to onnxruntime 1.16.1 MERGEOK

baldersheim:
@arnej27959 PR


searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
```diff
@@ -191,7 +191,7 @@ createPostingIterator(fef::TermFieldMatchData *matchData, bool strict)
         }
     }
     // returning nullptr will trigger fallback to filter iterator
-    return SearchIterator::UP();
+    return {};
```
hmusum, line 194:
```suggestion
    return SearchIterator::UP();
```

hmusum (dismissed):

hmusum:
Factory is broken, so let's try

PR #28945, opened 2023-10-16
Wire `CommunicationManager` config from its owner rather than self-subscribing

vekterli:
@baldersheim please review. This is the start of an effort to reduce the number of freestanding config (self-)subscribers in the content layer in favor of wiring in config changes from a more centralized location.

This moves the responsibility for bootstrapping and updating config for the `CommunicationManager` component to its owner. By doing this, a dedicated `ConfigFetcher` can be removed. Since this is a component used by both the distributor and storage nodes, this reduces total thread count by 2 on a host.

We probably want to pull config subscribing even further up and out with time, but this is a start.

### DIFF
--- a/storage/src/vespa/storage/storageserver/storagenode.h
+++ b/storage/src/vespa/storage/storageserver/storagenode.h
@@ -1,9 +1,6 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 /**
- * @class storage::StorageNode
- * @ingroup storageserver
- *
- * @brief Main storage server class.
+ * Main storage server class.
  *
  * This class sets up the entire storage server.
  *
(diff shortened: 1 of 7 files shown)


storage/src/vespa/storage/storageserver/storagenode.h
```diff
@@ -137,25 +133,30 @@ class StorageNode : private config::IFetcherCallback<vespa::config::content::cor
...
+
     // Current running config. Kept, such that we can see what has been
     // changed in live config updates.
     std::unique_ptr<StorServerConfig> _serverConfig;
     std::unique_ptr<UpgradingConfig> _clusterConfig;
     std::unique_ptr<StorDistributionConfig> _distributionConfig;
     std::unique_ptr<document::config::DocumenttypesConfig> _doctypesConfig;
     std::unique_ptr<BucketspacesConfig> _bucketSpacesConfig;
+    std::unique_ptr<CommunicationManagerConfig> _comm_mgr_config;
+
```
baldersheim, line 152: Seems to be a pattern here, Perhaps a small utility class with _current and _new config at some point in time.

baldersheim (approved):

storage/src/vespa/storage/storageserver/storagenode.h
vekterli: Agreed, many more configs to go... 🥹

PR #28946, opened 2023-10-16
Set provisioned-for if enabled and application has exclusive="true"

hakonhall:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/CapacityPolicies.java
```diff
@@ -174,7 +174,7 @@ private NodeResources smallestSharedResources() {
     public ClusterSpec decideExclusivity(Capacity capacity, ClusterSpec requestedCluster) {
         if (capacity.cloudAccount().isPresent()) return requestedCluster.withExclusivity(true); // Implicit exclusive
         boolean exclusive = requestedCluster.isExclusive() && (capacity.isRequired() || zone.environment() == Environment.prod);
-        return requestedCluster.withExclusivity(exclusive);
+        return requestedCluster.withExclusivity(exclusive).withProvisionForApplication(exclusive);
```
jonmv, line 177: There are other uses of the `Builder.exclusive(...)` as well. Did you check those?

jonmv: 💯

jonmv (dismissed):

jonmv (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterSpec.java
```diff
@@ -79,10 +77,8 @@ public Optional<Id> combinedId() {
         return combinedId;
     }
 
-    /**
-     * Returns whether the physical hosts running the nodes of this application can
-     * also run nodes of other applications. Using exclusive nodes for containers increases security and cost.
-     */
+
+    /** Returns whether the physical hosts must be provisioned specifically for this application. */
```
freva, line 81: The previous documentation also covered zones with host sharing (Yahoo)

freva (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/ClusterSpec.java
hakonhall: Let me restore the previous docstring.

PR #28953, opened 2023-10-16
Lesters/openai async client

lesters:
@bratseth Please review. For discussion.

Here, I re-used the `Completion` to also be a partial completion, e.g. a token. Perhaps this should be another concept?


bratseth (approved):
LGTM!

PR #28954, opened 2023-10-16
Avoid incorrect gcc warning compiling inlined XXH3 code. Also stick t…

baldersheim:
…o including official interface.

@vekterli PR


vekterli (approved):
:+1:

PR #28964, opened 2023-10-17
Remove unused configurability of operation priorities

vekterli:
@baldersheim please review

As far as I know, this config has not been used by anyone for at least a decade (if it ever was used for anything truly useful). Additionally, operation priorities are a foot-gun at the best of times. The ability to dynamically change the meaning of priority enums even more so.

This commit entirely removes configuration of Document API priority mappings in favor of a fixed mapping that is equal to the default config, i.e. what everyone's been using anyway.

This removes a thread per distributor/storage node process as well as 1 mutex and 1 (presumably entirely unneeded `seq_cst`) atomic load in the message hot path (primarily on the distributor since it serves doc API requests). Also precomputes a LUT for the priority reverse mapping to avoid needing to lower-bound seek an explicit map.

### DIFF
--- a/storage/src/vespa/storage/storageserver/communicationmanager.cpp
+++ b/storage/src/vespa/storage/storageserver/communicationmanager.cpp
@@ -230,7 +230,7 @@ CommunicationManager::CommunicationManager(StorageComponentRegister& compReg,
       _mbus(),
       _configUri(configUri),
       _closed(false),
-      _docApiConverter(configUri, std::make_shared<PlaceHolderBucketResolver>()), // TODO wire config from outside
+      _docApiConverter(std::make_shared<PlaceHolderBucketResolver>()),
       _thread()
 {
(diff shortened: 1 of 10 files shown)


baldersheim (approved):

storage/src/vespa/storage/storageserver/communicationmanager.cpp
```diff
@@ -230,7 +230,7 @@ CommunicationManager::CommunicationManager(StorageComponentRegister& compReg,
       _mbus(),
       _configUri(configUri),
```
baldersheim, line 231: If you bring the configUri in the on_configure call, you do not need it in the constructor, nor remember it here.

vekterli: `on_configure` only takes in the actual concrete config object. The `ConfigUri` is used when setting up MessageBus and its merry friends, so looks like it survives for now

PR #28969, opened 2023-10-17
Remove the moved ann_bencmark. No more python dependencies.

aressem:


toregge (dismissed):
👍

baldersheim (approved):
Yeeeesssss !!! :)

baldersheim (dismissed):
What about pybind ? Seem to remember that was introduced as dependency by ann-benchmark.

CMakeLists.txt
```diff
@@ -52,7 +52,6 @@ find_package(JNI REQUIRED)
 find_package(GTest REQUIRED)
 
 find_package(Python 3.6 COMPONENTS Interpreter Development REQUIRED)
```
baldersheim, line 54: We require python too, but that should no longer be necessary if we do not need it for anything else.

aressem: Testing removal of this dependency in https://github.com/vespa-engine/vespa/pull/28997

PR #28984, opened 2023-10-17
Since the cached size can be updated by many threads, it must be an a…

baldersheim:
…tomic since there can be many readers.

@toregge or @vekterli PR

### DIFF
--- a/vespalib/src/vespa/fastos/linux_file.cpp
+++ b/vespalib/src/vespa/fastos/linux_file.cpp
@@ -239,8 +239,8 @@ FastOS_Linux_File::internalWrite2(const void *buffer, size_t length)
         }
         if (writeRes > 0) {
             _filePointer += writeRes;
-            if (_filePointer > _cachedSize) {
-                _cachedSize = _filePointer;
+            if (_filePointer > _cachedSize.load(std::memory_order_relaxed)) {
+                _cachedSize.store(_filePointer, std::memory_order_release);
             }
         }
     } else {
@@ -277,7 +277,7 @@ FastOS_Linux_File::SetSize(int64_t newSize)
     bool rc = FastOS_UNIX_File::SetSize(newSize);
 
     if (rc) {
-        _cachedSize = newSize;
+        _cachedSize.store(newSize, std::memory_order_release);
     }
     return rc;
 }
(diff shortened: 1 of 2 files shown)


toregge (dismissed):

vespalib/src/vespa/fastos/linux_file.cpp
```diff
@@ -239,8 +239,8 @@ FastOS_Linux_File::internalWrite2(const void *buffer, size_t length)
         }
         if (writeRes > 0) {
             _filePointer += writeRes;
-            if (_filePointer > _cachedSize) {
-                _cachedSize = _filePointer;
+            if (_filePointer > _cachedSize.load(std::memory_order_relaxed)) {
+                _cachedSize.store(_filePointer, std::memory_order_release);
```
vekterli, line 243: If this can happen from multiple threads more or less at the same time, should this be a (relaxed) `compare_exchange` loop where we retry until the cached size is at least as high as `_filePointer`? This would be quite a bit more expensive, so maybe it's not required to be that up-to-date?

vekterli: Are there any transitive memory visibility dependencies that require this to be `release` rather than `relaxed`?

vespalib/src/vespa/fastos/linux_file.cpp
```diff
@@ -334,19 +334,21 @@ FastOS_Linux_File::DirectIOPadding (int64_t offset, size_t length, size_t &padBe
...
             padAfter = 0;
         }
-        if (int64_t(offset+length+padAfter) > _cachedSize) {
+        int64_t fileSize = _cachedSize.load(std::memory_order_relaxed);
+        if (int64_t(offset+length+padAfter) > fileSize) {
             // _cachedSize is not really trustworthy, so if we suspect it is not correct, we correct it.
             // The main reason is that it will not reflect the file being extended by another filedescriptor.
-            _cachedSize = getSize();
+            fileSize = getSize();
+            _cachedSize.store(fileSize, std::memory_order_release);
```
vekterli, line 342: Same question re: `release` vs. `relaxed`

vespalib/src/vespa/fastos/linux_file.cpp
```diff
@@ -277,7 +277,7 @@ FastOS_Linux_File::SetSize(int64_t newSize)
     bool rc = FastOS_UNIX_File::SetSize(newSize);
 
     if (rc) {
-        _cachedSize = newSize;
+        _cachedSize.store(newSize, std::memory_order_release);
```
vekterli, line 280: Same question re: `release` vs. `relaxed`

vekterli (dismissed):
:+1:

vespalib/src/vespa/fastos/linux_file.cpp
baldersheim: relaxed will do.

vespalib/src/vespa/fastos/linux_file.cpp
baldersheim: Relaxed will do.

PR #28985, opened 2023-10-17
Improve thread safety of MessageBus (DocumentAPI) ContentPolicy

vekterli:
@baldersheim please review. Observed on internal TSan testing. Of course, `ContentPolicy` also has its own personal `ConfigFetcher`... Cleanup for another day 😑

Updates of distribution config and cached cluster state are now both thread safe. Move to `shared_ptr` to allow for taking immutable strong refs.

Also remove pointless two-phased config switch-over in favor of directly updating value inside lock.

### DIFF
--- a/documentapi/src/vespa/documentapi/messagebus/policies/contentpolicy.h
+++ b/documentapi/src/vespa/documentapi/messagebus/policies/contentpolicy.h
@@ -6,7 +6,7 @@
 #include <vespa/vdslib/distribution/distribution.h>
 #include <vespa/document/bucket/bucketidfactory.h>
 #include <vespa/messagebus/routing/hop.h>
-#include <mutex>
+#include <shared_mutex>
 
 namespace config {
     class ICallback;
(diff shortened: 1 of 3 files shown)


documentapi/src/vespa/documentapi/messagebus/policies/contentpolicy.h
```diff
@@ -6,55 +6,62 @@
...
-    document::BucketIdFactory _bucketIdFactory;
-    std::unique_ptr<storage::lib::ClusterState> _state;
-    string _clusterName;
-    string _clusterConfigId;
-    std::unique_ptr<config::ICallback>          _callBack;
-    std::unique_ptr<config::ConfigFetcher>      _configFetcher;
-    std::unique_ptr<storage::lib::Distribution> _distribution;
-    std::unique_ptr<storage::lib::Distribution> _nextDistribution;
+    document::BucketIdFactory                         _bucketIdFactory;
+    mutable std::mutex                                _lock;
```
baldersheim, line 27: Could this be a read/write lock ?

baldersheim (dismissed):
Just small note about possible RW lock.

documentapi/src/vespa/documentapi/messagebus/policies/contentpolicy.h
vekterli: The critical section in the expected code path only spans two atomic up-refs, but write-locking should be very rare so it might be beneficial anyway. Will replace with `shared_mutex`.

baldersheim (approved):

PR #28987, opened 2023-10-17
- Modifications are guarded by external lock.

baldersheim:
- Use relaxed atomics to ensure reader visibility.

@vekterli or @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/docstore/filechunk.cpp
+++ b/searchlib/src/vespa/searchlib/docstore/filechunk.cpp
@@ -250,7 +250,7 @@ void
 FileChunk::remove(uint32_t lid, uint32_t size)
 {
      (void) lid;
-     _erasedCount.store(getBloatCount() + 1, std::memory_order_relaxed);
+     _erasedCount.store(getErasedCount() + 1, std::memory_order_relaxed);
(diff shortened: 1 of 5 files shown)


searchlib/src/vespa/searchlib/docstore/filechunk.cpp
```diff
@@ -250,8 +250,8 @@ void
 FileChunk::remove(uint32_t lid, uint32_t size)
 {
      (void) lid;
-     _erasedCount++;
-     _erasedBytes += adjustSize(size);
+     _erasedCount.store(getBloatCount() + 1, std::memory_order_relaxed);
```
vekterli, line 253: At first sight it's a bit confusing that `_erasedCount` is updated with the result of `getBloatCount()` (even though that just relaxed-loads `_erasedCount`). Should the function (or variable) be renamed so that they match?

vekterli (dismissed):
:+1:

searchlib/src/vespa/searchlib/docstore/filechunk.cpp
baldersheim: Yes, they should. Fixed

vekterli (approved):

PR #28990, opened 2023-10-17
Move xxh3_64 methods to vespalib. That also removes the need for work…

baldersheim:
…arounds for GCC false positives.

@vekterli PR


vekterli (approved):
:+1:

PR #28994, opened 2023-10-18
Add TestNG as banned dependency

bjorncs:
TestNG breaks our junit5 based test framework for Vespa Cloud


bjormel (approved):
Have we checked if this affects any apps?

PR #29000, opened 2023-10-18
Use exclusive host sharing, without exclusiveTo set, when pre-provisioning containers

jonmv:
@hakonhall

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
@@ -7,6 +7,7 @@ import com.yahoo.concurrent.UncheckedTimeoutException;
 import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.ClusterMembership;
 import com.yahoo.config.provision.ClusterSpec;
+import com.yahoo.config.provision.ClusterSpec.Type;
 import com.yahoo.config.provision.NodeAllocationException;
 import com.yahoo.config.provision.NodeResources;
 import com.yahoo.config.provision.NodeType;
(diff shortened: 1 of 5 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -219,15 +220,19 @@ private List<Node> provisionHosts(int count, NodeResources nodeResources, Option
             }
             Version osVersion = nodeRepository().osVersions().targetFor(NodeType.host).orElse(Version.emptyVersion);
             List<Integer> provisionIndices = nodeRepository().database().readProvisionIndices(count);
+            HostSharing sharingMode = clusterType.map(ClusterSpec.Type::isContainer).orElse(false) ? HostSharing.exclusive : HostSharing.shared;
```
hakonhall, line 223: It would be nice to reuse Preparer::hostSharing, but with HostSharing.shared instead of HostSharing.any.  (:thinking: I wonder if it is really necessary to have both `any` and `shared`?)
For instance, we want HostSharing.shared for public containers in test and staging.  And HostSharing.shared in main systems.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -219,15 +220,19 @@ private List<Node> provisionHosts(int count, NodeResources nodeResources, Option
...
+                                                                    sharingMode, clusterType, Optional.empty(),
                                                                     nodeRepository().zone().cloud().account(), false);
             List<Node> hosts = new ArrayList<>();
             hostProvisioner.provisionHosts(request,
                                            resources -> true,
                                            provisionedHosts -> {
-                                               hosts.addAll(provisionedHosts.stream().map(host -> host.generateHost(Duration.ZERO)).toList());
+                                               hosts.addAll(provisionedHosts.stream()
+                                                                            .map(host -> host.generateHost(Duration.ZERO))
+                                                                            .map(host -> host.withExclusiveToApplicationId(null))
```
hakonhall, line 234: Throw an exception here or in withExclusiveToApplicationId() if provisionedForApplicationId is present?

jonmv: Yes, could just remove the nulling of provisionedFor in this setter. I originally wanted this as well.

hakonhall (approved):

PR #29003, opened 2023-10-18
Remove unused document config update logic

vekterli:
@baldersheim please review

Actual document config changes are propagated in from the top-level `Process` via an entirely different call chain. Having the unused one around is just confusing, so remove it.


baldersheim (approved):
What have we not....

PR #29010, opened 2023-10-18
add defaults extraction and unit test

arnej27959:
@havardpe please review


havardpe (approved):
looks good

We should extend the test to verify actual values and also test with normalizers

PR #29011, opened 2023-10-18
Add reactive infra-application redeployer, triggered when hosts compl…

jonmv:
…ete provisioning

@hakonhall please review.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
@@ -85,7 +85,9 @@ public class InfraApplicationRedeployer implements AutoCloseable {
         return switch (type) {
             case host -> Optional.of(InfrastructureApplication.TENANT_HOST);
             case confighost -> Optional.of(InfrastructureApplication.CONFIG_SERVER_HOST);
+            case config -> Optional.of(InfrastructureApplication.CONFIG_SERVER);
             case controllerhost -> Optional.of(InfrastructureApplication.CONTROLLER_HOST);
+            case controller -> Optional.of(InfrastructureApplication.CONTROLLER);
             case proxyhost -> Optional.of(InfrastructureApplication.PROXY_HOST);
             default -> Optional.empty();
         };
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
@@ -72,7 +72,7 @@ public class InfraApplicationRedeployer extends AbstractComponent {
             log.log(INFO, () -> "Redeploying " + application.id() + " after completing provisioning of " + application.name());
             try {
                 deployer.getDeployment(application.id()).ifPresent(Deployment::activate);
-                if (childOf(application) != null) readied(childOf(application));
+                readied(childOf(application));
             }
             catch (RuntimeException e) {
                 log.log(WARNING, "Failed redeploying " + application.id() + ", will be retried by maintainer", e);
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
@@ -1,6 +1,5 @@
 package com.yahoo.vespa.hosted.provision.maintenance;
 
-import com.yahoo.component.AbstractComponent;
 import com.yahoo.component.annotation.Inject;
 import com.yahoo.concurrent.DaemonThreadFactory;
 import com.yahoo.concurrent.UncheckedTimeoutException;
@@ -13,8 +12,8 @@ import com.yahoo.vespa.applicationmodel.InfrastructureApplication;
 import com.yahoo.vespa.hosted.provision.Node.State;
 import com.yahoo.vespa.hosted.provision.NodeList;
 import com.yahoo.vespa.hosted.provision.NodeRepository;
-import com.yahoo.vespa.service.duper.TenantHostApplication;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentSkipListSet;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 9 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
```diff
@@ -0,0 +1,117 @@
...
+        readied(applicationOf(type));
+    }
+
+    private void readied(InfrastructureApplication application) {
+        if (application == null) return;
+        if (readiedTypes.add(application)) executor.execute(() -> checkAndRedeploy(application));
+    }
+
+    private void checkAndRedeploy(InfrastructureApplication application) {
+        log.log(INFO, () -> "Checking if " + application.name() + " should be redeployed");
```
hakonhall, line 67: To be demoted or removed later?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        catch (UncheckedTimeoutException collision) {
+            readied(application);
+        }
+    }
+
+    private static InfrastructureApplication applicationOf(NodeType type) {
+        return switch (type) {
+            case host -> InfrastructureApplication.TENANT_HOST;
+            case confighost -> InfrastructureApplication.CONFIG_SERVER_HOST;
```
hakonhall, line 89: Add config server and controller apps?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
```diff
@@ -0,0 +1,117 @@
...
+        if (application == null) return;
+        if (readiedTypes.add(application)) executor.execute(() -> checkAndRedeploy(application));
+    }
+
+    private void checkAndRedeploy(InfrastructureApplication application) {
+        log.log(INFO, () -> "Checking if " + application.name() + " should be redeployed");
+        if ( ! readiedTypes.remove(application)) return;
+        log.log(INFO, () -> "Trying to redeploy " + application.id() + " after completing provisioning of " + application.name());
+        try (Mutex lock = locks.apply(application.id())) {
+            if (application.nodeType().isHost() && nodes.get().state(State.ready).nodeType(application.nodeType()).isEmpty()) return;
```
hakonhall, line 71: Test for `application.nodeType().isHost()` is wrong since this code should work for child types too.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesV2ApiHandler.java
```diff
@@ -75,13 +76,16 @@ public class NodesV2ApiHandler extends ThreadedHttpRequestHandler {
     private final Orchestrator orchestrator;
     private final NodeRepository nodeRepository;
     private final NodeFlavors nodeFlavors;
+    private final InfraApplicationRedeployer infraApplicationRedeployer;
 
     @Inject
-    public NodesV2ApiHandler(Context parentCtx, Orchestrator orchestrator, NodeRepository nodeRepository, NodeFlavors flavors) {
+    public NodesV2ApiHandler(Context parentCtx, Orchestrator orchestrator, NodeRepository nodeRepository,
+                             NodeFlavors flavors, InfraApplicationRedeployer infraApplicationRedeployer) {
```
hakonhall, line 83: You need to declare this component in services for open-source and hosted?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
jonmv: Ah, forgot to remove these.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
jonmv: Those are never set ready? We should only ever translate from host type nodes, I believe?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
jonmv: It skips deployment if it's a host type without ready nodes. I don't think containers have an equivalent ready state? So they're never skipped, but only triggered (unconditionally) after we actually redeploy their host applications.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/restapi/NodesV2ApiHandler.java
jonmv: Ah, yes, probably. Thanks.

jonmv: A lot simpler to just create it in the handler, then, possibly based on some other config.

jonmv: Ah, I don't think the open-source configserver comes with a node repo, does it? The system test of the PR went fine, which strongly suggests it doesn't.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
jonmv: Fixed, thanks!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
hakonhall: I think cfg/ctl can be parked/failed, and then recycled with --> dirty --> ready --> reserved/active. I agree it's not on the common path.

jonmv: Handling it now.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/InfraApplicationRedeployer.java
hakonhall: isHost() can now be removed?

hakonhall (approved):

PR #29014, opened 2023-10-18
Report peak memory usage.

baldersheim:
@vekterli @havardpe @toregge PR
@bjorncs FYI. You can try this branch and see what differs. No java changes necessary.

### DIFF
--- a/vespamalloc/src/vespamalloc/malloc/mmappool.cpp
+++ b/vespamalloc/src/vespamalloc/malloc/mmappool.cpp
@@ -10,6 +10,7 @@ namespace vespamalloc {
 MMapPool::MMapPool()
     : _page_size(getpagesize()),
       _huge_flags((getenv("VESPA_USE_HUGEPAGES") != nullptr) ? MAP_HUGETLB : 0),
+      _peakBytes(0ul),
       _count(0),
       _mutex(),
       _mappings()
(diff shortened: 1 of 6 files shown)


bjorncs (dismissed):

baldersheim:
I will add these as extra numbers so we can get a feel for what is a better estimate.

havardpe (dismissed):
how much slower will this make mmap?

vespamalloc/src/vespamalloc/malloc/mmappool.cpp
```diff
@@ -76,9 +83,10 @@ MMapPool::mmap(size_t sz) {
         std::lock_guard guard(_mutex);
         auto [it, inserted] = _mappings.insert(std::make_pair(buf, MMapInfo(mmapId, sz)));
         ASSERT_STACKTRACE(inserted);
+        size_t sum(0);
+        std::for_each(_mappings.begin(), _mappings.end(), [&sum](const auto & e){ sum += e.second._sz; });
```
vekterli, line 87: Since this is tracking the peak usage, could we instead increase a "current usage" sum on mmap and decrease it on unmap, then maintain the max value observed of this? The O(n) CPU time per mapping seems like it could add up over time...

havardpe (approved):

vekterli (approved):
:+1:

PR #29020, opened 2023-10-19
PricingController has moved, injected as a component now

hmusum:
Must be merged with PR in internal repo


freva (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/restapi/pricing/PricingApiHandler.java
```diff
@@ -48,11 +49,13 @@ public class PricingApiHandler extends ThreadedHttpRequestHandler {
     private static final Logger log = Logger.getLogger(PricingApiHandler.class.getName());
 
     private final Controller controller;
+    private final PricingController pricingController;
 
     @Inject
-    public PricingApiHandler(Context parentCtx, Controller controller) {
+    public PricingApiHandler(Context parentCtx, Controller controller, PricingController pricingController) {
```
mpolden, line 55: It's also possible to move the handler itself. That way the `PricingController` interface won't be needed.

hmusum: Yeah, thanks, I'll look into that

PR #29032, opened 2023-10-19
Improve modelling of match strategies to use in numeric range search.

geirst:
This should improve the performance by choosing the strategy that is most optimal in different scenarios: lookup-based filter matching vs posting lists merging.

The modelling is based on results from the range search performance test: https://github.com/vespa-engine/system-test/blob/master/tests/performance/range_search/.

@baldersheim please review
@toregge FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
+++ b/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
@@ -451,17 +451,15 @@ NumericPostingSearchContext<BaseSC, AttrT, DataT>::calc_estimated_hits_in_range(
 {
     size_t exact_sum = 0;
     size_t estimated_sum = 0;
-    uint32_t count = 0;
     constexpr uint32_t max_posting_lists_to_count = 1000;
-    for (auto it = this->_lowerDictItr; it != this->_upperDictItr; ++it) {
-        if (count >= max_posting_lists_to_count) {
-            uint32_t remaining_posting_lists = this->_upperDictItr - it;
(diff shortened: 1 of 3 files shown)


searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
```diff
@@ -455,7 +401,70 @@ getIterators(bool shouldApplyRangeLimit)
...
+template <typename BaseSC, typename AttrT, typename DataT>
+size_t
+NumericPostingSearchContext<BaseSC, AttrT, DataT>::calc_estimated_hits_in_range() const
+{
+    size_t exact_sum = 0;
+    size_t estimated_sum = 0;
+    uint32_t count = 0;
+    constexpr uint32_t max_posting_lists_to_count = 1000;
+    for (auto it = this->_lowerDictItr; it != this->_upperDictItr; ++it) {
+        if (count >= max_posting_lists_to_count) {
```
baldersheim, line 457: I find it easier to read and understand if you add (count < max_posting_lists_to_count) as criteria in the for loop,
and have if branch after the for loop. Then you don't need the break. either. It makes it more clear what happens in the for loop.

baldersheim (dismissed):
See comment about readability.

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.h
geirst: Fixed

geirst:
@baldersheim PTAL

PR #29045, opened 2023-10-20
Avoid cutting surrogate pairs when tokenising

jonmv:
@baldersheim or @bratseth please review and merge.

### DIFF
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
@@ -72,7 +72,7 @@ public class LinguisticsAnnotator {
         Tokenizer tokenizer = factory.getTokenizer();
         String input = (text.getString().length() <= config.getMaxTokenizeLength())
                 ? text.getString()
-                : Text.safeSubstring(text.getString(), config.getMaxTokenizeLength());
+                : Text.substringByCodepoints(text.getString(), 0, config.getMaxTokenizeLength());
         Iterable<Token> tokens = tokenizer.tokenize(input, config.getLanguage(), config.getStemMode(),
                                                     config.getRemoveAccents());
         TermOccurrences termOccurrences = new TermOccurrences(config.getMaxTermOccurrences());
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
@@ -12,6 +12,7 @@ import com.yahoo.language.Linguistics;
 import com.yahoo.language.process.StemMode;
 import com.yahoo.language.process.Token;
 import com.yahoo.language.process.Tokenizer;
+import com.yahoo.text.Text;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
```diff
@@ -71,7 +72,7 @@ public boolean annotate(StringFieldValue text) {
         Tokenizer tokenizer = factory.getTokenizer();
         String input = (text.getString().length() <= config.getMaxTokenizeLength())
                 ? text.getString()
-                : text.getString().substring(0, config.getMaxTokenizeLength());
+                : Text.safeSubstring(text.getString(), config.getMaxTokenizeLength());
```
baldersheim, line 75: Why not use Text.substringByCodepoints ?
That is much easier to reason over for the user.

baldersheim (changes requested):

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
jonmv: It's a larger change, and slower for long fields, although probably not by much. I agree it's a better end state.

jonmv: Not sure I dare do that with `truncate(...)`, though. It specifically states "characters" in its documentation. WDYT?

baldersheim: For every body, except the java developers who came up with the high/low surrogate trick, a character is the whole character and not only the part of a character that fits inside a java character.

jonmv: I agree with that, but this is a public-API method that specifies characters, and it could potentially be used in code which assumes the returned length in `char` is not greater than the argument. I'm not saying it's _very_ likely, but we'd break anyone who made that assumption.

baldersheim: Check usages of truncate, and see what they think they expect....

baldersheim:
You also need to update ABI.

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
baldersheim: The only usage I found you can safely treat a character as a human, and not an well aged java core developer...
`ParseException p = new ParseException("Could not parse '" + Text.truncate(expression, 50) + "'");`

baldersheim: If you look at the comments you see this can not be used for anything machine readable.
     * If length is 4 or less the string will be truncated to length.
     * If length is longer than 4, it will be truncated at length-4 with " ..." added at the end.

We have a lot of public methods that we have only added for our own convenience, and this looks like one of them.

And the change we are making here will already change change something, that is just a theoretical difference.
The only value it add over s.substring(), is the optional padding with "...".

jonmv: Sold!

baldersheim (approved):

PR #29055, opened 2023-10-20
Support exclusive allocation on non-exclusive host

hakonhall:
This PR is a no-op.  But when enabling the new make-exclusive flag:
- An exclusive allocation will be allowed on a non-exclusive host
- The host is then made exclusive

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Applications.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Applications.java
@@ -63,19 +63,24 @@ public class Applications {
         db.deleteApplication(transaction);
     }
 
+    public record Lock(Mutex mutex, ApplicationId application) implements Mutex {
+        @Override
+        public void close() { mutex.close(); }
+    }
+
     /** Create a lock which provides exclusive rights to making changes to the given application */
-    public Mutex lock(ApplicationId application) {
-        return db.lock(application);
+    public Lock lock(ApplicationId application) {
+        return new Lock(db.lock(application), application);
     }
 
     /** Create a lock with a timeout which provides exclusive rights to making changes to the given application */
-    public Mutex lock(ApplicationId application, Duration timeout) {
-        return db.lock(application, timeout);
+    public Lock lock(ApplicationId application, Duration timeout) {
+        return new Lock(db.lock(application, timeout), application);
     }
 
     /** Create a lock which provides exclusive rights to perform a maintenance deployment */
-    public Mutex lockMaintenance(ApplicationId application) {
-        return db.lockMaintenance(application);
+    public Lock lockMaintenance(ApplicationId application) {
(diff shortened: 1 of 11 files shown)


hakonhall:
FYI @jonmv

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeAllocation.java
```diff
@@ -186,18 +192,15 @@ private boolean checkForClashingParentHost() {
...
     private boolean offeredNodeHasParentHostnameAlreadyAccepted(NodeCandidate candidate) {
-        for (NodeCandidate acceptedNode : nodes.values()) {
-            if (acceptedNode.parentHostname().isPresent() && candidate.parentHostname().isPresent() &&
-                    acceptedNode.parentHostname().get().equals(candidate.parentHostname().get())) {
-                return true;
-            }
-        }
-        return false;
+        if (candidate.parentHostname().isEmpty()) return false;
+        return nodes.values().stream().anyMatch(acceptedNode -> acceptedNode.parentHostname().equals(candidate.parentHostname()));
```
jonmv, line 196: 👍

jonmv (approved):
LGTM!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Applications.java
```diff
@@ -63,19 +63,24 @@ public void remove(ApplicationTransaction transaction) {
         db.deleteApplication(transaction);
     }
 
+    public record Lock(Mutex mutex, ApplicationId application) implements Mutex {
```
freva, line 66: FYI: There's already a type for this: `com.yahoo.config.provision.ProvisionLock`

node-repository/src/test/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainerTest.java
```diff
@@ -316,6 +320,44 @@ public void respects_exclusive_allocation() {
         tester.assertNodesUnchanged();
     }
 
+    @Test
+    public void works_as_before_without_make_exclusive() {
```
freva, line 324: What about a test with `make-exclusive`?

jonmv: The above test has that.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/applications/Applications.java
hakonhall: Thanks, I failed to find such a class even though I vaguely remembered there was one.  Fixed in https://github.com/vespa-engine/vespa/pull/29070

PR #29071, opened 2023-10-23
Set title for existing notifications

freva:


controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/Notification.java
```diff
@@ -32,18 +32,18 @@ public Notification(Instant at, Type type, Level level, NotificationSource sourc
...
-        level = Objects.requireNonNull(level, "level cannot be null");
-        source = Objects.requireNonNull(source, "source cannot be null");
-        title = Objects.requireNonNull(title, "title cannot be null");
-        messages = List.copyOf(Objects.requireNonNull(messages, "messages cannot be null"));
+        Objects.requireNonNull(at, "at cannot be null");
+        Objects.requireNonNull(type, "type cannot be null");
+        Objects.requireNonNull(level, "level cannot be null");
+        Objects.requireNonNull(source, "source cannot be null");
+        Objects.requireNonNull(title, "title cannot be null");
+        List.copyOf(Objects.requireNonNull(messages, "messages cannot be null"));
```
bjorncs, line 40: Missing assignment for list copy?

bjorncs (approved):

controller-server/src/main/java/com/yahoo/vespa/hosted/controller/notification/Notification.java
freva: Yes, thanks, fixed in #29074

PR #29073, opened 2023-10-23
Propagate existing `StorageNode` config from main `Process` reconfig loop

vekterli:
@baldersheim please review. There are more changes brewing in the pipeline (tearing out config fetchers from individual storage chain components in `ServiceLayerNode`), but I would like the main wiring changes to go out first.

### DIFF
--- a/storage/src/vespa/storage/storageserver/distributornode.cpp
+++ b/storage/src/vespa/storage/storageserver/distributornode.cpp
@@ -18,12 +18,13 @@ namespace storage {
 DistributorNode::DistributorNode(
         const config::ConfigUri& configUri,
         DistributorNodeContext& context,
+        BootstrapConfigs bootstrap_configs,
         ApplicationGenerationFetcher& generationFetcher,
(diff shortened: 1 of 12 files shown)


baldersheim (dismissed):

baldersheim:
Needs deinline
[2023-10-21 03:13:17.845] WARNING distributor      vds.distributor.operations.external.update	Update operation for 'id:ecdata_legolas_personal_offer_score:ecdata_legolas_personal_offer_score::shp_8344200' in bucket Bucket(BucketSpace(0x0000000000000001), BucketId(0x4000000000003232)) updated documents with different timestamps. This should not happen and may indicate undetected replica divergence. Found low ts=0 on node 8 (pending GC: no), highest ts=1697854316000166 on node 3 (pending GC: no)

vekterli:
Will back-port some dtor de-inlining that I already had in a followup-commit...

baldersheim (approved):

PR #29076, opened 2023-10-24
Add tax code/purchase order/invoice email

olaaun:


gjoranv (dismissed):
We should consider creating custom types for these strings, perhaps using PatternedStringWrapper

olaaun:
> We should consider creating custom types for these strings, perhaps using PatternedStringWrapper

Used StringWrapper instead, want to avoid assumptions regarding pattern

gjoranv (dismissed):

gjoranv (approved):

PR #29085, opened 2023-10-24
Prepare for emitting warning if summary field type is specified.

toregge:
Improve resolving of summary field type from source field type.

@geirst : please review


config-model/src/main/java/com/yahoo/schema/parser/ConvertParsedSchemas.java
```diff
@@ -137,7 +141,82 @@ private void convertDocument(Schema schema, ParsedDocument parsed,
         schema.addDocument(document);
     }
 
-    private void convertDocumentSummary(Schema schema, ParsedDocumentSummary parsed, TypeResolver typeContext) {
+    /*
+     * Helper class for resolving data type for a document summary. Summary type is still
+     * used internally in config model when generating and processing indexing scripts.
+     * See DynamicSummaryTransformUtils class comment for more details.
```
geirst, line 147: Please also add that this kind of resolving is a temporary measurement, and that the final goal is to get rid of using summary fields in indexing scripts and as extra fields in document types.

geirst (dismissed):
👍

config-model/src/main/java/com/yahoo/schema/parser/ConvertParsedSchemas.java
geirst: And in that case the data type of a summary field is no longer relevant.

geirst (approved):
👍

PR #29095, opened 2023-10-25
Take parent lock only for tenant app in GCP/AWS

hakonhall:


config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
```diff
@@ -88,6 +88,14 @@ public boolean canRun(NodeType type) {
         return childNodeTypes.contains(type);
     }
 
+    /** Returns the parent host type. */
+    public NodeType parentNodeType() {
```
freva, line 92: What's wrong with `NodeType::hostType`?

freva (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/NodeType.java
hakonhall: It's pretty good.  It returns `this` if the type is already a host.  In our case I need the parent host, and it's wrong if the parent host type doesn't exist. :shrug:

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
```diff
@@ -628,7 +628,9 @@ public ExclusivityViolation violatesExclusivity(ClusterSpec cluster, Application
 
             // this cluster requires exclusivity, but the parent is not exclusive
             if (exclusiveAllocation && parent.flatMap(Node::exclusiveToApplicationId).isEmpty())
-                return makeExclusive ? ExclusivityViolation.PARENT_HOST_NOT_EXCLUSIVE : ExclusivityViolation.YES;
+                return Preparer.requireParentHostLock(makeExclusive, type(), hostSharing) ?
```
jonmv, line 631: It's a bit weird asking whether you need a lock, to determine if exclusivity is violated, but I suppose it's otherwise code duplication. Perhaps rename the method, though?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
```diff
@@ -92,10 +92,24 @@ public List<Node> prepare(ApplicationId application, ClusterSpec cluster, NodeSp
         }
     }
 
+    /**
+     * Whether the preparation of an application MAY require changes to the parent hosts, and therefore the parent host lock is required.
+     * See {@link NodeCandidate#withExclusiveParent(boolean)}.
+     */
+    public static boolean requireParentHostLock(boolean makeExclusive, NodeType type, boolean allowHostSharing) {
```
jonmv, line 99: Something like: `canMakeEmptyHostExclusive`, perhaps? I'm sure there are better variations than that, too.

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeCandidate.java
hakonhall: :+1:

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Preparer.java
hakonhall: Good suggestion

PR #29102, opened 2023-10-25
Use new method

hmusum:
Make sure to use the new method introduced in https://github.com/vespa-engine/vespa/pull/29090


andreer (approved):

bjorncs (commented):
👍

PR #29119, opened 2023-10-26
keep the implicit (hidden) match features in a separate set in RankProfile

arnej27959:
@havardpe maybe this is more clean


havardpe (approved):

havardpe:
yes :-)

PR #29121, opened 2023-10-26
switch to io.airlift:aircompressor

arnej27959:
@vekterli please look over the java code changes
@gjoranv see if the changes to dependency management looks OK
@bjorncs @baldersheim FYI

### DIFF
--- a/container/pom.xml
+++ b/container/pom.xml
@@ -35,8 +35,8 @@
           <artifactId>asm</artifactId>
         </exclusion>
         <exclusion>
-          <groupId>com.yahoo.vespa</groupId>
-          <artifactId>airlift-zstd</artifactId>
+          <groupId>io.airlift</groupId>
+          <artifactId>aircompress</artifactId>
         </exclusion>
         <exclusion>
           <groupId>net.java.dev.jna</groupId>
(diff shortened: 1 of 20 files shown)


vekterli:
In `com.yahoo.compress.Compressor` there is a field

```java
private final ZstdCompressor zstdCompressor = new ZstdCompressor();
```

and `Compressor` has the class comment

> Compressor which can compress and decompress in various formats.
> This class is thread safe. Creating a reusable instance is faster than creating instances as needed.

😬

Looks like a broader sweep of API usages might be required. Alternatively, we can create `ZstdCompressor` and `ZstdDecompressor` instances on-demand in the individual `compress`/`decompress` methods to avoid any chance of cross-thread usage.

container/pom.xml
```diff
@@ -35,8 +35,8 @@
           <artifactId>asm</artifactId>
         </exclusion>
         <exclusion>
-          <groupId>com.yahoo.vespa</groupId>
-          <artifactId>airlift-zstd</artifactId>
+          <groupId>io.airlift</groupId>
+          <artifactId>aircompress</artifactId>
```
gjoranv, line 39: Should be 'aircompressor'.

gjoranv (approved):

vekterli (approved):
:+1:

PR #29147, opened 2023-10-30
Delete zookeeper session as well when deployment fails

hmusum:


jonmv (dismissed):

jonmv (dismissed):
The log message on the previous line is not right with this change. But this looks right ... I think?

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/Deployment.java
```diff
@@ -136,7 +137,6 @@ public long activate() {
                 waitForActivation(applicationId, timeoutBudget, activation);
             } catch (Exception e) {
                 log.log(Level.FINE, "Activating session " + session.getSessionId() + " failed, deleting it");
```
jonmv, line 139:
```suggestion
```

hmusum (commented):
Thanks!

PR #29177, opened 2023-11-01
Avoid clinit deadlock between ApplicationId and its constituent field…

jonmv:
… classes


bjorncs (approved):
Good catch! 👏

PR #29182, opened 2023-11-01
Move `HwInfo` from searchcore to vespalib

vekterli:
@baldersheim please review. Moving this out so that we can propagate this information from the Proton bootstrap config into the storage-specific modules during process init.
@geirst @toregge FYI

This is information that is valuable to many different components, not just the search core internals.

Will initially be used when auto-deducing merge memory limits, but we should also change existing non-cgroups-aware thread pool setups in the persistence code to use this information when no explicit config is provided.


geirst (approved):
👍

baldersheim (approved):

PR #29191, opened 2023-11-01
validate for array/wset attributes, take 2:

arnej27959:
* ensure attribute(arrayfield).count does not trigger validation
* log warning via deploylogger instead of throwing exception

@geirst please review
@hmusum @bratseth FYI

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/RankProfile.java
+++ b/config-model/src/main/java/com/yahoo/schema/RankProfile.java
@@ -1265,14 +1265,45 @@ public class RankProfile implements Cloneable {
         return Optional.empty();  // if this context does not contain this input
     }
 
+    private static class AttributeErrorType extends TensorType {
+        private final DeployLogger deployLogger;
+        private final String attr;
+        private final Attribute.CollectionType collType;
+        private boolean shouldWarn = true;
+        AttributeErrorType(DeployLogger deployLogger, String attr, Attribute.CollectionType collType) {
+            super(TensorType.Value.DOUBLE, List.of());
+            this.deployLogger = deployLogger;
+            this.attr = attr;
+            this.collType = collType;
+        }
+        private void warnOnce() {
+            if (shouldWarn) {
+                deployLogger.log(Level.WARNING, "Using attribute(" + attr +") " + collType + " as ranking expression input");
+                shouldWarn = false;
+            }
+        }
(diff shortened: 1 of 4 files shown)


config-model/src/main/java/com/yahoo/schema/RankProfile.java
```diff
@@ -1265,14 +1265,45 @@ private Optional<TensorType> resolveOnnxInputType(String onnxInputName, OnnxMode
...
+        private boolean shouldWarn = true;
+        AttributeErrorType(DeployLogger deployLogger, String attr, Attribute.CollectionType collType) {
+            super(TensorType.Value.DOUBLE, List.of());
+            this.deployLogger = deployLogger;
+            this.attr = attr;
+            this.collType = collType;
+        }
+        private void warnOnce() {
+            if (shouldWarn) {
+                deployLogger.log(Level.WARNING, "Using attribute(" + attr +") " + collType + " as ranking expression input");
```
geirst, line 1281: Perhaps the message should say something about what this implies, so that the user can do some changes.

config-model/src/test/java/com/yahoo/schema/processing/RankingExpressionTypeResolverTestCase.java
```diff
@@ -458,6 +458,54 @@ void undeclaredQueryFeaturesAreAcceptedWithWarningWhenUsingTensors() throws Exce
...
+        builder.addSchema(joinLines(
+                "search test {",
+                "  document test { ",
+                "    field foo type weightedset<int> {",
+                "      indexing: attribute",
+                "    }",
+                "  }",
+                "  rank-profile my_rank_profile {",
+                "    first-phase {",
+                "      expression: attribute(foo)",
```
geirst, line 498: What will happen if you use `attribute(foo).count`? This is a supported feature for both array and weighted set. From https://docs.vespa.ai/en/reference/rank-features.html#document-features:

_attribute(name).count - The number of elements in the attribute with the given name_

geirst (dismissed):
👍

config-model/src/test/java/com/yahoo/schema/processing/RankingExpressionTypeResolverTestCase.java
geirst: Discussed with Arne, and this is handled.

geirst (approved):
👍

PR #29198, opened 2023-11-02
Test that OpenSSL mTLS integration is not vulnerable to certificate stuffing

vekterli:
@bjorncs please review. Inspired by https://github.blog/2023-08-17-mtls-when-certificate-authentication-is-done-wrong/

This adds a test that our OpenSSL mTLS integration is not vulnerable to CVE-2023-2422-style certificate credential stuffing.

Spoiler alert: we're not—and never have been—vulnerable. But this test shall help to ensure we also never accidentally will be in the future.

If a server is vulnerable to certificate stuffing, a sneaky client may include both a valid certificate chain (containing credential set A) as well as a self-signed peer certificate (containing credential set B). The vulnerable server thinks the latter cert has been verified, even though the mTLS implementation only verifies the first (actual) client cert as being signed by the CA. The server may then wrongfully choose to include set B as the client's credentials.

We explicitly only consider certificates in the chain at OpenSSL "error depth zero", which means the "end entity certificate", i.e. the client peer. In addition, when testing locally, the bad cert is never actually presented to the certificate verify callback at all.

### DIFF
--- a/vespalib/src/tests/net/tls/openssl_impl/openssl_impl_test.cpp
+++ b/vespalib/src/tests/net/tls/openssl_impl/openssl_impl_test.cpp
@@ -455,8 +455,8 @@ struct CertFixture : Fixture {
     {}
     ~CertFixture();
 
-    CertKeyWrapper create_ca_issued_peer_cert(const std::vector<vespalib::string>& common_names,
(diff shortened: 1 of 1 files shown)


bjorncs (approved):
👍

PR #29199, opened 2023-11-02
Exclude airlift from linkcheck

kkraune:


aressem (approved):

baldersheim (commented):
Note that airlift code will soon disappear from our repo.

PR #29202, opened 2023-11-02
Don't autoscale resources in single node clusters

bjormel:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -78,6 +78,9 @@ private Autoscaling autoscale(Application application, Cluster cluster, NodeList
         if (target.isEmpty())
             return Autoscaling.dontScale(Status.insufficient, "No allocations are possible within configured limits", model);
 
+        if (target.get().nodes() == 1 && model.current().nodes() == 1)
+            return Autoscaling.dontScale(Status.unavailable, "Autoscaling of resources is not enabled for single node clusters", model);
+
```
bratseth, line 83:
```suggestion
        if (target.get().nodes() == 1)
            return Autoscaling.dontScale(Status.unavailable, "Autoscaling is disabled in single node clusters", model);

```

bratseth (dismissed):

bjormel:
Thanks @bratseth!

PR #29207, opened 2023-11-02
Retire if there are shared hosts dedicated to cluster type

hakonhall:
Violate exclusivity if shared-host contains a shared host for the cluster type being allocated.  When an "admin" shared host is defined for a zone, this ensures that all admin nodes will be put on these "admin" shared hosts.  As it is now, this should only affect CD prod.aws-us-east-1a, main AWS test and staging, and public CD AWS. 

This is a no-op if make-exclusive is disabled, which it is in main and public systems, currently.


freva (approved):

hakonhall:
FYI @jonmv

jonmv (commented):
Wohoo!

PR #29223, opened 2023-11-03
Stop adding and populating extra fields for dynamic summary.

toregge:
They have not been used for a long time (summary field source field is used instead).

@geirst : please review


baldersheim (approved):

geirst (approved):
👍

baldersheim:
Should this go in now, or does it need to wait for a later version. If so it should be anoted with do not merge.

PR #29228, opened 2023-11-04
Evaluate flags referenced multiple times at the start of prepare

hakonhall:
Avoid exclusiveAllocation flipping in the middle of a cluster prepare, or any other expressions depending on the make-exclusive or shared-host flags.  This also touched various parts of the autoscaler which I hope I got right.


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
```diff
@@ -62,7 +61,7 @@ public class HostCapacityMaintainer extends NodeRepositoryMaintainer {
 
     private final HostProvisioner hostProvisioner;
     private final ListFlag<ClusterCapacity> preprovisionCapacityFlag;
-    private final BooleanFlag makeExclusiveFlag;
+    private final FlagSource flagSource;
```
freva, line 64: Not used?

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Limits.java
```diff
@@ -65,12 +66,12 @@ public NodeResources cap(NodeResources resources) {
         return resources;
     }
 
-    public Limits fullySpecified(ClusterSpec clusterSpec, NodeRepository nodeRepository, ApplicationId applicationId) {
+    public Limits fullySpecified(AllocationParams params, NodeRepository nodeRepository) {
```
freva, line 69: Use `NodeRepository` in `params`?

freva (dismissed):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostCapacityMaintainer.java
hakonhall: Removed

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Limits.java
hakonhall: Done

freva (approved):

PR #29229, opened 2023-11-04
Add and test correct implementation for undefined const char values

baldersheim:
@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchcommon/common/undefinedvalues.h
+++ b/searchlib/src/vespa/searchcommon/common/undefinedvalues.h
@@ -31,43 +31,42 @@ inline constexpr const char* getUndefined<const char*>() {
 
 // for all signed integers
 template <typename T>
-bool isUndefined(const T & value) {
+bool isUndefined(T value) {
     return value == getUndefined<T>();
 }
 
 template <>
-inline bool isUndefined<uint8_t>(const uint8_t &) {
+inline bool isUndefined<uint8_t>(uint8_t) {
     return false;
 }
 
 template <>
-inline bool isUndefined<uint16_t>(const uint16_t &) {
+inline bool isUndefined<uint16_t>(uint16_t) {
     return false;
 }
 
 template <>
-inline bool isUndefined<uint32_t>(const uint32_t &) {
+inline bool isUndefined<uint32_t>(uint32_t) {
     return false;
 }
 
 template <>
-inline bool isUndefined<uint64_t>(const uint64_t &) {
+inline bool isUndefined<uint64_t>(uint64_t) {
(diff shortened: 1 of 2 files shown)


toregge (dismissed):

searchlib/src/vespa/searchcommon/common/undefinedvalues.h
```diff
@@ -31,43 +31,42 @@ inline constexpr const char* getUndefined<const char*>() {
...
-inline bool isUndefined<double>(const double & value) {
+inline bool isUndefined<double>(double value) {
     return std::isnan(value);
 }
 
-template <>
-inline bool isUndefined<vespalib::string>(const vespalib::string & value) {
-    return value.empty();
+inline bool isUndefined(const char * value) {
+    return (value == nullptr) || (value[0] == '\0');
```
toregge, line 69: Note that value should never be nullptr here.

baldersheim: So should probably just remove it  then...

toregge: There are issues with multi value string attributes for documents with zero values
that might cause a crash if the check for nullptr is removed:

https://github.com/vespa-engine/vespa/blob/df7854e4c340ea814ca5cdf128fc31828a4b543b/searchlib/src/vespa/searchlib/attribute/multistringattribute.h#L62

this is also a problem for numeric multivalue attributes (wrong return values):

https://github.com/vespa-engine/vespa/blob/df7854e4c340ea814ca5cdf128fc31828a4b543b/searchlib/src/vespa/searchlib/attribute/multinumericattribute.h#L85
https://github.com/vespa-engine/vespa/blob/df7854e4c340ea814ca5cdf128fc31828a4b543b/searchlib/src/vespa/searchlib/attribute/multinumericattribute.h#L89
https://github.com/vespa-engine/vespa/blob/df7854e4c340ea814ca5cdf128fc31828a4b543b/searchlib/src/vespa/searchlib/attribute/multinumericattribute.h#L93
https://github.com/vespa-engine/vespa/blob/df7854e4c340ea814ca5cdf128fc31828a4b543b/searchlib/src/vespa/searchlib/attribute/multinumericenumattribute.h#L63

toregge (dismissed):

searchlib/src/vespa/searchcommon/common/undefinedvalues.h
baldersheim: Sigh..what have we not. The singlevalue methods should throw instead...

toregge (approved):

PR #29248, opened 2023-11-06
Not possible to do safe cast here. Must leav as is and wait until we …

baldersheim:
…clean up with proper get interfaces.

@toregge PR


baldersheim:
or @vekterli

toregge (commented):
👍

PR #29267, opened 2023-11-07
If match-phase limiting has concluded that a post filter is most effi… MERGEOK

baldersheim:
…cient,

we must only generate posting lists if it is actually benefiscal. If not the fixed cost is too high.

@geirst PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/attribute_limiter.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/attribute_limiter.cpp
@@ -99,7 +99,7 @@ AttributeLimiter::create_search(size_t want_hits, size_t max_group_size, double 
         FieldSpecList field; // single field API is protected
         field.add(FieldSpec(_attribute_name, my_field_id, my_handle));
         _blueprint = _searchable_attributes.createBlueprint(_requestContext, field, node);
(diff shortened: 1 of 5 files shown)


geirst (dismissed):
👍

baldersheim:
@havardpe PR
@geirst note last commit

searchcore/src/vespa/searchcore/proton/matching/attribute_limiter.cpp
```diff
@@ -99,7 +99,7 @@ AttributeLimiter::create_search(size_t want_hits, size_t max_group_size, double
         FieldSpecList field; // single field API is protected
         field.add(FieldSpec(_attribute_name, my_field_id, my_handle));
         _blueprint = _searchable_attributes.createBlueprint(_requestContext, field, node);
-        _blueprint->fetchPostings(ExecuteInfo::create(strictSearch, hit_rate, &_requestContext.getDoom()));
+        _blueprint->fetchPostings(ExecuteInfo::create(strictSearch, strictSearch ? 1.0F : hit_rate, &_requestContext.getDoom()));
```
havardpe, line 102: the hit_rate here represents fewer hits than we are going to observe, which might lead to regret later on

PR #29271, opened 2023-11-07
Sameelement behaves like an and with extra constraints.

baldersheim:
So it should behave the sameway during fetchPostings too.

@havardpe @geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/executeinfo.h
+++ b/searchlib/src/vespa/searchlib/queryeval/executeinfo.h
@@ -22,13 +22,10 @@ public:
     static ExecuteInfo create(bool strict, const ExecuteInfo & org) noexcept {
         return {strict, org._hitRate, org.getDoom()};
     }
-    static ExecuteInfo create(bool strict, const vespalib::Doom * doom) noexcept {
-        return create(strict, 1.0F, doom);
-    }
     static ExecuteInfo create(bool strict, float hitRate, const vespalib::Doom * doom) noexcept {
         return {strict, hitRate, doom};
     }
-    static ExecuteInfo create(bool strict) noexcept {
+    static ExecuteInfo createForTest(bool strict) noexcept {
         return create(strict, 1.0F);
     }
     static ExecuteInfo create(bool strict, float hitRate) noexcept {
--- a/searchlib/src/vespa/searchlib/queryeval/executeinfo.h
+++ b/searchlib/src/vespa/searchlib/queryeval/executeinfo.h
@@ -26,9 +26,9 @@ public:
         return {strict, hitRate, doom};
     }
     static ExecuteInfo createForTest(bool strict) noexcept {
-        return create(strict, 1.0F);
+        return createForTest(strict, 1.0F);
     }
-    static ExecuteInfo create(bool strict, float hitRate) noexcept {
+    static ExecuteInfo createForTest(bool strict, float hitRate) noexcept {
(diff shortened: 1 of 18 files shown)


searchlib/src/vespa/searchlib/queryeval/executeinfo.h
```diff
@@ -22,13 +22,10 @@ class ExecuteInfo {
...
-        return create(strict, 1.0F, doom);
-    }
     static ExecuteInfo create(bool strict, float hitRate, const vespalib::Doom * doom) noexcept {
         return {strict, hitRate, doom};
     }
-    static ExecuteInfo create(bool strict) noexcept {
+    static ExecuteInfo createForTest(bool strict) noexcept {
         return create(strict, 1.0F);
     }
     static ExecuteInfo create(bool strict, float hitRate) noexcept {
```
havardpe, line 31: maybe this also is 'for test'?

searchlib/src/vespa/searchlib/queryeval/same_element_blueprint.cpp
```diff
@@ -56,8 +56,15 @@ SameElementBlueprint::optimize_self()
 void
 SameElementBlueprint::fetchPostings(const ExecuteInfo &execInfo)
 {
-    for (size_t i = 0; i < _terms.size(); ++i) {
-        _terms[i]->fetchPostings(ExecuteInfo::create(execInfo.isStrict() && (i == 0), execInfo));
+    if (_terms.empty()) return;
```
havardpe, line 59: I do not think 2x code is worth the one call to hit_ratio

havardpe (dismissed):

searchlib/src/vespa/searchlib/queryeval/same_element_blueprint.cpp
```diff
@@ -56,8 +56,15 @@ SameElementBlueprint::optimize_self()
...
+    _terms[0]->fetchPostings(execInfo);
+    double hit_rate = execInfo.hitRate();
+    const Blueprint * prev = _terms[0].get();
+    for (size_t i = 1; i < _terms.size(); ++i) {
+        hit_rate = hit_rate * prev->hit_ratio();
+        Blueprint * current = _terms[i].get();
+        current->fetchPostings(ExecuteInfo::create(false, execInfo));
```
geirst, line 66: Shouldn't `hit_rate` be used here?

searchlib/src/vespa/searchlib/queryeval/executeinfo.h
baldersheim: It is, fixed.

searchlib/src/vespa/searchlib/queryeval/same_element_blueprint.cpp
baldersheim: Yes, indeed :)

searchlib/src/vespa/searchlib/queryeval/same_element_blueprint.cpp
baldersheim: I did it for readability. I find it better to handle the first outside when it is different than the next. Here it is strictness that differs.

havardpe: I see, but you could still get rid of the current/prev shuffle if you called hit_ratio one extra time...

geirst (dismissed):
👍

PR #29282, opened 2023-11-08
Stop subscribing to config that is static

hmusum:
File acquirer will always use localhost:19090 for getting files now, remove need for subscribing to qr config.
Will also remove restart log message when e.g. moving logserver to a new node

Please review only, will merge after release Thursday


hmusum:
@hakonhall FYI

hakonhall (commented):
:+1:

hmusum:
I think there is an opportunity for cleaning up even more, but I want to verify that this is OK first

baldersheim (approved):

PR #29284, opened 2023-11-08
Include doc type name and GID in metadata iteration results

vekterli:
@geirst @toregge please review. In particular pay close attention to the expected semantics of handler lifetime and handler → retriever relationships.
@baldersheim FYI

Document type is fetched from the associated `IPersistenceHandler` on-demand; it is assumed the lifetime of the pointer must be valid for the entire lifetime of the iterator itself, as the latter holds a valid handler snapshot.

For simplicity, it's possible to _not_ pass in a handler, in which case the doc type name will be implicitly empty.

Some expected `DocEntry` sizes have been adjusted, as we now report the size of the document type and GID alongside the base type size.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/persistenceengine/document_iterator.cpp
+++ b/searchcore/src/vespa/searchcore/proton/persistenceengine/document_iterator.cpp
@@ -1,8 +1,10 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "document_iterator.h"
+#include "ipersistencehandler.h"
 #include <vespa/persistence/spi/docentry.h>
 #include <vespa/searchcore/proton/common/cachedselect.h>
+#include <vespa/searchcore/proton/common/doctypename.h>
 #include <vespa/searchcore/proton/common/selectcontext.h>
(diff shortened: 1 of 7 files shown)


toregge (dismissed):

searchcore/src/vespa/searchcore/proton/persistenceengine/document_iterator.cpp
```diff
@@ -91,18 +100,24 @@ DocumentIterator::DocumentIterator(const storage::spi::Bucket &bucket,
 
 DocumentIterator::~DocumentIterator() = default;
 
+void
+DocumentIterator::add(const IPersistenceHandler *handler, IDocumentRetriever::SP retriever)
```
geirst, line 104: Consider passing the document type name directly, to avoid adding a coupling between DocumentIterator and IPersistenceHandler. Otherwise LGTM.

geirst (dismissed):
👍

searchcore/src/vespa/searchcore/proton/persistenceengine/document_iterator.cpp
vekterli: Done

vekterli:
@geirst PTAL

geirst (approved):
👍

PR #29292, opened 2023-11-08
store cells in blocks for MixedTensor

arnej27959:
@bratseth or @lesters or @jonmv please review

note: two internal classes are no longer public;
I'm taking the bet that they aren't actually used outside the class itself.

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
@@ -5,10 +5,12 @@ package com.yahoo.tensor;
 import com.google.common.collect.ImmutableMap;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -28,12 +30,21 @@ public class MixedTensor implements Tensor {
     private final TensorType type;
     private final int blockSize; // aka dense subspace size
 
-    static class DenseBlock {
+    static final class DenseBlock {
         final TensorAddress sparseAddr;
         final double[] cells;
         DenseBlock(TensorAddress sparseAddr, double[] cells) {
             this.sparseAddr = sparseAddr;
             this.cells = cells;
+        }
+        @Override public int hashCode() {
+            return Objects.hash(sparseAddr, cells);
+        }
+        @Override public boolean equals(Object other) {
+            if (other instanceof DenseBlock o) {
+                return sparseAddr.equals(o.sparseAddr) && Arrays.equals(cells, o.cells);
+            }
+            return false;
         }
     }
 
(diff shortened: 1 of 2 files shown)


vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
```diff
@@ -26,17 +26,44 @@ public class MixedTensor implements Tensor {
 
     /** The dimension specification for this tensor */
     private final TensorType type;
+    private final int blockSize; // aka dense subspace size
 
-    /** The list of cells in the tensor */
-    private final List<Cell> cells;
+    static class DenseBlock {
+        final TensorAddress sparseAddr;
+        final double[] cells;
```
jonmv, line 35: Do we want coarser numbers here too, later?

jonmv (dismissed):
Nice!

baldersheim (dismissed):
👍

vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
arnej27959: yes; I would like to have float and byte as options at least.

vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
```diff
@@ -26,17 +28,53 @@ public class MixedTensor implements Tensor {
...
+        if (this.blockSize < 1) {
+            throw new IllegalStateException("invalid dense subspace size " + blockSize);
+        }
+        long count = 0;
+        for (var block : this.cellBlocks) {
+            if (index.sparseMap.get(block.sparseAddr) != count) {
+                throw new IllegalStateException("map vs list mismatch");
+            }
+            if (block.cells.length != blockSize) {
+                throw new IllegalStateException("block length mismatch");
```
geirst, line 71: Consider outputting expected vs actual values so its easier to pinpoint the issue if this happens.

vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
```diff
@@ -26,17 +28,53 @@ public class MixedTensor implements Tensor {
...
+            }
+            ++count;
+        }
+        if (count != index.sparseMap.size()) {
+            throw new IllegalStateException("map vs list size mismatch");
```
geirst, line 76: Consider outputting expected vs actual values so its easier to pinpoint the issue if this happens.

vespajlib/src/main/java/com/yahoo/tensor/MixedTensor.java
```diff
@@ -26,17 +28,53 @@ public class MixedTensor implements Tensor {
...
+        this.blockSize = index.denseSubspaceSize();
+        this.cellBlocks = List.copyOf(cellBlocks);
         this.index = index;
+        if (this.blockSize < 1) {
+            throw new IllegalStateException("invalid dense subspace size " + blockSize);
+        }
+        long count = 0;
+        for (var block : this.cellBlocks) {
+            if (index.sparseMap.get(block.sparseAddr) != count) {
+                throw new IllegalStateException("map vs list mismatch");
```
geirst, line 68: Consider outputting the sparse address so its easier to pinpoint the issue if this happens.

PR #29299, opened 2023-11-09
expose dense subspace blocks for serializing

arnej27959:
this simplifies serialization by exposing the internal representation;
we probably want to do more redesign (to allow float cells etc at least)
so this API should only be used internally.

@lesters or @bratseth please review
@jonmv @geirst  @havardpe FYI


bratseth (commented):
Otherwise LGTM!

baldersheim (approved):
👍

PR #29300, opened 2023-11-09
Ban JAX-RS API and warn on Guice/javax.inject.

bjorncs:
Improve UX with custom error message


gjoranv (approved):
Will our build fail if one of these artifacts are removed as provided deps from container in the future?

bjorncs:
> Will our build fail if one of these artifacts are removed as provided deps from container in the future?

No, we'll have to manually update the enforcer rules.

PR #29301, opened 2023-11-09
detect unpack bits in more cases

havardpe:
- with multiple dimensions
- inside map_subspaces

@arnej27959 please review


arnej27959 (approved):
great!

PR #29307, opened 2023-11-10
Limit max payload size based on heap size

bjorncs:


container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/ServletRequestReader.java
```diff
@@ -105,14 +107,28 @@ private enum State {
...
     }
 
+    private static String resolveMaxContentSizeErrorMessage(ConnectorConfig cfg) {
+        return cfg.maxContentSizeErrorMessageTemplate().strip();
+    }
+
+    private static long resolveMaxContentSize(ConnectorConfig cfg) {
+        // Scale based on max heap size if 0
+        long maxContentSize = cfg.maxContentSize() != 0
+                ? cfg.maxContentSize() : Math.min(Runtime.getRuntime().maxMemory() / 2, Integer.MAX_VALUE);
```
jonmv, line 127:
```suggestion
                ? cfg.maxContentSize() : Math.min(Runtime.getRuntime().maxMemory() / 2, Integer.MAX_VALUE - 8);
```
If you want a limit of 2GB here, should probably set it to what can be held in an array as well.

jonmv (dismissed):
See comment.

PR #29332, opened 2023-11-14
Add env var for setting `no_new_privs` process bit on Linux

vekterli:
@arnej27959 please review

If set, this will apply to all processes launched by the config sentinel, directly or transitively. This is a one-way toggle.

From https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt:

> The `no_new_privs` bit (since Linux 3.5) is a new, generic mechanism to make it safe for a process to modify its
execution environment in a manner that persists across `execve`. Any task can set `no_new_privs`. Once the bit is set, it is inherited across `fork`, `clone`, and `execve` and cannot be unset. With `no_new_privs` set, `execve` promises not to grant the privilege to do anything that could not have been done without the `execve` call. For example, the `setuid` and `setgid` bits will no longer change the uid or gid; file capabilities will not add to the permitted set, and LSMs will not relax constraints after `execve`.

### DIFF
--- a/configd/src/apps/sentinel/CMakeLists.txt
+++ b/configd/src/apps/sentinel/CMakeLists.txt
@@ -14,6 +14,7 @@ vespa_add_executable(configd_config-sentinel_app
     output-connection.cpp
     outward-check.cpp
     peer-check.cpp
+    platform-specific.cpp
     report-connectivity.cpp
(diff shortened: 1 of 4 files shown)


arnej27959 (approved):
👍

PR #29339, opened 2023-11-15
Explicitly print backtrace on bucket space invariant violation

vekterli:
@geirst please review

This is to help catch an unknown edge case that can happen if distributor operation cancellation is enabled (note: not enabled by default). Will be removed once the edge case is found+fixed.


geirst (approved):
👍

PR #29342, opened 2023-11-15
GC old and unused persistence-level metrics

vekterli:
These metrics are from the age of VDS and a dozen spinning disks per node 🦕


vekterli:
@baldersheim please review

baldersheim (approved):

PR #29349, opened 2023-11-16
Hmusum/onnx model options

hmusum:
Not sure if the 2nd commit made things better or not, what do you think?

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/component/BertEmbedder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/component/BertEmbedder.java
@@ -1,12 +1,11 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.container.component;
 
+import com.yahoo.config.ModelReference;
 import com.yahoo.config.model.deploy.DeployState;
 import com.yahoo.embedding.BertBaseEmbedderConfig;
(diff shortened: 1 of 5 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/component/BertEmbedder.java
```diff
@@ -16,56 +18,45 @@
...
-    private final String transformerAttentionMask;
-    private final String transformerTokenTypeIds;
-    private final String transformerOutput;
-    private final Integer tranformerStartSequenceToken;
-    private final Integer transformerEndSequenceToken;
-    private final String poolingStrategy;
-    private final String onnxExecutionMode;
-    private final Integer onnxInteropThreads;
-    private final Integer onnxIntraopThreads;
-    private final Integer onnxGpuDevice;
```
bjorncs, line 32: Only the parameters `onnxExecutionMode`, `onnxInteropThreads`, `onnxIntraopThreads` and `onnxGpuDevice` should be included in `OnnxModelOptions`. The other parameters to not control how to the Onnx model is loaded, only its input/output.

bjorncs:
I'm open for both with a slight preference for `Optional` :)

config-model/src/main/java/com/yahoo/vespa/model/container/component/BertEmbedder.java
hmusum: Fixed

bjorncs (approved):

PR #29350, opened 2023-11-16
Balder/handle executeinfo in or

baldersheim:
@geirst @havardpe PR


havardpe (approved):
it is a bit confusing that we assume stuff without fastsearch matches both everything and nothing at the same time

baldersheim:
Yes, indeed. We should probably find a better way to signal that in the HitEstimate.

PR #29353, opened 2023-11-16
always process with SerializationContext (needed by unpack_bits)

arnej27959:
@bratseth please review
@jobergum FYI


bratseth (approved):
LGTM

PR #29372, opened 2023-11-17
- Or and SourceBlender inherits max cost tier of children.

baldersheim:
- Rank and AndNot inherits cost tier for first child.
- And and the others inherits minimum cost tier of children as it already does.

Tests are missing.


baldersheim:
@havardpe PR

havardpe (approved):

PR #29380, opened 2023-11-20
atoi => std::from_chars

baldersheim:
@vekterli PR

### DIFF
--- a/searchlib/src/vespa/searchlib/fef/indexproperties.cpp
+++ b/searchlib/src/vespa/searchlib/fef/indexproperties.cpp
@@ -4,6 +4,7 @@
 #include "properties.h"
 #include <vespa/vespalib/locale/c.h>
 #include <limits>
+#include <charconv>
 
 namespace search::fef::indexproperties {
 
@@ -49,10 +50,12 @@ uint32_t
 lookupUint32(const Properties &props, const vespalib::string &name, uint32_t defaultValue)
 {
     Property p = props.lookup(name);
+    uint32_t value(defaultValue);
     if (p.found()) {
-        return atoi(p.get().c_str());
+        const auto & s = p.get();
+        std::from_chars(s.c_str(), s.c_str() + s.size(), value);
     }
-    return defaultValue;
+    return value;
 }
 
 bool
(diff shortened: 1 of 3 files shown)


searchlib/src/vespa/searchlib/fef/indexproperties.cpp
```diff
@@ -49,10 +50,12 @@ uint32_t
 lookupUint32(const Properties &props, const vespalib::string &name, uint32_t defaultValue)
 {
     Property p = props.lookup(name);
+    uint32_t value(defaultValue);
     if (p.found()) {
-        return atoi(p.get().c_str());
+        const auto & s = p.get();
+        std::from_chars(s.c_str(), s.c_str() + s.size(), value);
```
vekterli, line 56: Is there a chance that these properties may have leading whitespace? `atoi` implicitly consumes these while `from_chars` does not.

vekterli, line 56: Do we want to check the `error_code` returned as part of the result tuple, or is the fallback to `defaultValue` considered good enough?

vekterli (dismissed):
:+1:

searchlib/src/vespa/searchlib/fef/indexproperties.cpp
baldersheim: We have no way of handling any error, so fallback to default value is the best we can do I think.
One step better than atoi.

baldersheim: I expect no leading whitespace, but I will make it atoi compliant anyway to reduce risk of regression.

baldersheim: Fixed and tested explicit

searchlib/src/vespa/searchlib/fef/indexproperties.cpp
```diff
@@ -52,8 +52,11 @@ lookupUint32(const Properties &props, const vespalib::string &name, uint32_t def
     Property p = props.lookup(name);
     uint32_t value(defaultValue);
     if (p.found()) {
-        const auto & s = p.get();
-        std::from_chars(s.c_str(), s.c_str() + s.size(), value);
+        const auto & valS = p.get();
+        const char * start = valS.c_str();
+        const char * end = start + valS.size();
+        while (isspace(start[0]) && start != end) { start++; }
```
vekterli, line 58: Consider swapping comparison order so that `start` is never dereferenced when `start == end`

baldersheim: Ouch, good catch.

vekterli (approved):

PR #29382, opened 2023-11-20
Avoid reconfiguring TLS when a custom client is used

mpolden:
Thanks for the workaround! Always reconfiguring TLS had a negative impact on
feed performance though, which should be fixed by this change. `vespa feed`
brings a pre-configured client through `SetClient`.

@arnej27959


arnej27959 (approved):
looks good to me

PR #29384, opened 2023-11-20
Balder/always nesting multivalue grouping for indexed search now

baldersheim:
@vekterli PR

### DIFF
--- a/searchcore/src/tests/grouping/grouping_test.cpp
+++ b/searchcore/src/tests/grouping/grouping_test.cpp
@@ -45,7 +45,7 @@ struct MyWorld {
         bv.setInterval(0, NUM_DOCS);
         // attribute context
         {
-            SingleInt32ExtAttribute *attr = new SingleInt32ExtAttribute("attr0");
+            auto *attr = new SingleInt32ExtAttribute("attr0");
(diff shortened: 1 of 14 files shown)


searchcore/src/tests/grouping/grouping_test.cpp
```diff
@@ -94,16 +94,17 @@ using GroupingList = GroupingContext::GroupingList;
 
 SessionId createSessionId(const std::string & s) {
     std::vector<char> vec;
-    for (size_t i = 0; i < s.size(); i++) {
-        vec.push_back(s[i]);
+    for (char c : s) {
+        vec.push_back(c);
     }
-    return SessionId(&vec[0], vec.size());
+    return {&vec[0], vec.size()};
```
vekterli, line 100: Not anything changed by this PR, but: should this be `vec.data()` instead of `vec[0]` to avoid undefined behavior if the input string is empty?

vekterli (dismissed):
:+1:

searchcore/src/tests/grouping/grouping_test.cpp
baldersheim: Yes, it should. Done

PR #29385, opened 2023-11-20
GC no longer used useReconfigurableDispatcher and adminClusterArchite…

baldersheim:
…cture

@jonmv and @hmusum PR

### DIFF
--- a/config-model-api/abi-spec.json
+++ b/config-model-api/abi-spec.json
@@ -1282,7 +1282,6 @@
       "public boolean enableGlobalPhase()",
       "public java.lang.String summaryDecodePolicy()",
       "public boolean enableNestedMultivalueGrouping()",
-      "public boolean useReconfigurableDispatcher()",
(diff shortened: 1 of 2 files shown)


hmusum (changes requested):
adminClusterArchitecture is still in use and cannot be removed (the "removeAfter" annotation should be removed as well).

jonmv:
The admin architecture _flag_ is in use, but just directly—not through model.

jonmv (dismissed):

baldersheim:
@hmusum PTAL

jonmv:
Either that, or wire it through the model so it follows Vespa version ... if that was the intention?

baldersheim:
Then we wait with admin architecture flag and take that separately.
PTAL

hmusum (approved):
@jonmv you are right, `adminClusterArchitecture` is not used through the model, I was fooled by IntelliJ not showing the method in ModelContextImpl as unused. I'll take care of removing it in another PR

PR #29390, opened 2023-11-20
Jonmv/pre provisioned lbs cleanup

jonmv:
@freva don't merge yet.


freva (dismissed):
`id` argument no longer used?

jonmv:
Ah, thanks, that's right. Should remove that, too.

freva (dismissed):

PR #29397, opened 2023-11-21
Count operations passed to feeder

mpolden:
@jonmv


jonmv (approved):

jonmv:
Do we have a way to prevent drift between the two implementations of the feed client?

mpolden:
No, the clients are already very different though. You have anything specific in mind?

PR #29410, opened 2023-11-21
Add command for downloading application package

### LINKED ISSUE #26487: New API to download application packages
From #26104

Build an API to get all files from an active application to redeply it incrementally

(I noticed vespa-deploy fetch, but could not find the same fetch in the APIs)

Use case:  an application package is already deployed with some schemas on a remote server or in the cloud. You want to add/remove/update a schema from it, without erasing other schemas. But you do not have access to the source of the last deployed files, nor to the server filesystem.

mpolden:
Fixes #26487.

@jonmv

### DIFF
--- a/client/go/internal/vespa/deploy.go
+++ b/client/go/internal/vespa/deploy.go
@@ -151,14 +151,15 @@ func fetchFromController(deployment DeploymentOptions, path string) error {
 		pkgURL *url.URL
 		err    error
 	)
-	if deployment.Target.Deployment().Zone.Environment == "dev" {
+	switch deployment.Target.Deployment().Zone.Environment {
+	case "dev", "perf":
 		pkgURL, err = deployment.url(fmt.Sprintf("/application/v4/tenant/%s/application/%s/instance/%s/job/%s/package",
 			deployment.Target.Deployment().Application.Tenant,
 			deployment.Target.Deployment().Application.Application,
 			deployment.Target.Deployment().Application.Instance,
 			deployment.Target.Deployment().Zone.Environment+"-"+deployment.Target.Deployment().Zone.Region,
 		))
-	} else {
+	default:
 		pkgURL, err = deployment.url(fmt.Sprintf("/application/v4/tenant/%s/application/%s/package",
 			deployment.Target.Deployment().Application.Tenant,
 			deployment.Target.Deployment().Application.Application),
--- a/client/go/internal/vespa/deploy.go
+++ b/client/go/internal/vespa/deploy.go
@@ -12,6 +12,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"net/url"
+	"os"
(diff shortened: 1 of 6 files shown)


client/go/internal/vespa/deploy.go
```diff
@@ -119,6 +119,126 @@ func ZoneFromString(s string) (ZoneID, error) {
...
+	_, err = io.Copy(w, response.Body)
+	return err
+}
+
+func fetchFromController(deployment DeploymentOptions, path string) error {
+	var (
+		pkgURL *url.URL
+		err    error
+	)
+	if deployment.Target.Deployment().Zone.Environment == "dev" {
```
jonmv, line 154: Or `perf`?

client/go/internal/vespa/deploy.go
```diff
@@ -119,6 +119,126 @@ func ZoneFromString(s string) (ZoneID, error) {
...
+		pkgURL, err = deployment.url(fmt.Sprintf("/application/v4/tenant/%s/application/%s/instance/%s/job/%s/package",
+			deployment.Target.Deployment().Application.Tenant,
+			deployment.Target.Deployment().Application.Application,
+			deployment.Target.Deployment().Application.Instance,
+			deployment.Target.Deployment().Zone.Environment+"-"+deployment.Target.Deployment().Zone.Region,
+		))
+	} else {
+		pkgURL, err = deployment.url(fmt.Sprintf("/application/v4/tenant/%s/application/%s/package",
+			deployment.Target.Deployment().Application.Tenant,
+			deployment.Target.Deployment().Application.Application),
```
jonmv, line 164: Not including the `build` option is fine with me here.

mpolden: Agree, we can add if/when someone wants it.

client/go/internal/vespa/deploy.go
```diff
@@ -119,6 +119,126 @@ func ZoneFromString(s string) (ZoneID, error) {
...
+			}
+		} else {
+			if err := os.MkdirAll(filepath.Dir(entryName), 0755); err != nil {
+				return err
+			}
+			f, err := os.Create(entryName)
+			if err != nil {
+				return err
+			}
+			if err := deployServiceGet(fu, deployment, f); err != nil {
```
jonmv, line 232: Any reason not to `defer f.close()`?

jonmv (dismissed):

client/go/internal/vespa/deploy.go
mpolden: This is called in a loop. If you defer closing, all files are kept open until the loop is done and the function returns. If the package has many files you could run into the open file limit.

jonmv: Ah, thanks for educating me :)

jonmv (approved):

PR #29421, opened 2023-11-22
Reduce timer frequency when QEMU virtualization is used

mpolden:
Podman Desktop/podman machine uses QEMU to run the Linux VM where containers
live on macOS. With this setup, an idle Vespa application causes `qemu` to use
constantly use ~250% CPU. Dropping timer frequency reduced this to around ~80%
for me.

@arnej27959

FYI @baldersheim


baldersheim (approved):
🥇

PR #29425, opened 2023-11-22
Check for valid successor.

toregge:
@geirst : please review
@vekterli : FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/dfa_fuzzy_matcher.h
+++ b/searchlib/src/vespa/searchlib/attribute/dfa_fuzzy_matcher.h
@@ -5,7 +5,6 @@
 #include "dfa_string_comparator.h"
 #include <vespa/vespalib/datastore/atomic_entry_ref.h>
 #include <vespa/vespalib/fuzzy/levenshtein_dfa.h>
-#include <iostream>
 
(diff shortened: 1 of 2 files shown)


searchlib/src/vespa/searchlib/attribute/dfa_fuzzy_matcher.h
```diff
@@ -47,7 +49,7 @@ class DfaFuzzyMatcher {
                     return true;
                 }
                 _successor.resize(_prefix.size());
-                _successor.emplace_back(1);
+                _successor.emplace_back(beyond_unicode);
```
baldersheim, line 52: Belong in a hpp file due to size.
hardly needs iostream

baldersheim (approved):

geirst (approved):
👍

geirst:
We already test the corner case where prefix length > target length (exact match) in _fuzzy_match_in_dictionary_with_prefix_size_ (https://github.com/vespa-engine/vespa/blob/master/searchlib/src/tests/attribute/dfa_fuzzy_matcher/dfa_fuzzy_matcher_test.cpp). With the assert added in _is_match()_ this test starts failing without the successor fix.

PR #29434, opened 2023-11-23
Set more SSL properties

jonmv:
@bjorncs


bjorncs (approved):

jonmv:
> ssl.authProvider : (Java system property: zookeeper.ssl.authProvider) Specifies a subclass of org.apache.zookeeper.auth.X509AuthenticationProvider to use for secure client authentication. This is useful in certificate key infrastructures that do not use JKS. It may be necessary to extend javax.net.ssl.X509KeyManager and javax.net.ssl.X509TrustManager to get the desired behavior from the SSL stack. To configure the ZooKeeper server to use the custom provider for authentication, choose a scheme name for the custom AuthenticationProvider and set the property zookeeper.authProvider.[scheme] to the fully-qualified class name of the custom implementation. This will load the provider into the ProviderRegistry. Then set this property zookeeper.ssl.authProvider=[scheme] and that provider will be used for secure authentication.

PR #29442, opened 2023-11-23
Do not setup API authenticator when using concrete URL

mpolden:
Avoids the need to setup API key or `vespa auth login` when you simply want to
interact directly with a deployment.

@jonmv


jonmv (approved):

kkraune:
Thanks! So we can remove this line then?

https://github.com/vespa-engine/documentation/blob/master/.github/workflows/feed.yml#L52

        export VESPA_CLI_API_KEY=

mpolden:
Yes.

PR #29444, opened 2023-11-23
Support headers in basic HTTP tests

### LINKED ISSUE #28092: Support for headers in system tests executed by vespa test
Hello Vespa team!

I have come across with the need of creating system tests with http requests with headers.

I would usually execute them with `vespa test` and a json file describing the test configuration.
However, it seems that http requests with headers are not supported with `vespa test` and I have found that the only way to test them are with Java API tests, as stated in https://docs.vespa.ai/en/testing.html.
The documentation suggests the Java API tests are for complex tests in mind, but I think just adding a header to an http request doesn't count as a complex situation at all and just adds too much complexity to simple system tests that just require adding an additional header.

Would it be possible to add the ability of executing http requests with headers for `vespa test` system tests?

What I had in mind was a configuration similar to this to execute an http request with headers:

```json
"steps": [
        {
            "name": "get documents",
            "request": {
                "method": "GET",
                "uri": "/document/v1",
                "parameters": {
                    "cluster": "music",
                    "selection": "true"
                },
                "headers": {
                    "context": "{\"key\": \"value\"}"       
                }
            }
        }
]
```

Thank you.

Kind regards,

Sergio Moreno López.

mpolden:
Fixes #28092.

@jonmv

### DIFF
--- a/client/go/internal/cli/cmd/test.go
+++ b/client/go/internal/cli/cmd/test.go
@@ -198,6 +198,14 @@ func verify(step step, defaultCluster string, defaultParameters map[string]strin
 		method = "GET"
 	}
 
+	header := http.Header{}
+	for k, v := range step.Request.Headers {
+		header.Set(k, v)
+	}
+	if header.Get("Content-Type") == "" { // Set default if not specified by test
+		header.Set("Content-Type", "application/json") // TODO: Not guaranteed to be true ...
+	}
+
 	var service *vespa.Service
 	requestUri := step.Request.URI
 	if requestUri == "" {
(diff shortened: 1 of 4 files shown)


client/go/internal/cli/cmd/test.go
```diff
@@ -198,6 +198,14 @@ func verify(step step, defaultCluster string, defaultParameters map[string]strin
 		method = "GET"
 	}
 
+	header := http.Header{}
+	for k, v := range step.Request.Headers {
+		header.Set(k, v)
+	}
+	if header.Get("Content-Type") == "" { // Set default if not specified by test
+		header.Set("Content-Type", "application/json") // TODO: Not guaranteed to be true ...
```
jonmv, line 206:
```suggestion
		header.Set("Content-Type", "application/json")
```
Remove TODO that was really resolved now :)

jonmv (dismissed):

PR #29449, opened 2023-11-23
jackson 2.16 changes some of its default settings so we consolidate o…

baldersheim:
…ur use of the ObjectMapper.

Unless special options are used, use a common instance, or create via factory metod.

@bjorncs PR


bjorncs (commented):
👍

PR #29453, opened 2023-11-24
Jonmv/zk 3.9.1

jonmv:
@bjorncs

### DIFF
--- a/zookeeper-server/zookeeper-server-3.9.1/src/main/java/org/apache/zookeeper/common/ClientX509Util.java
+++ b/zookeeper-server/zookeeper-server-3.9.1/src/main/java/org/apache/zookeeper/common/ClientX509Util.java
@@ -71,17 +71,17 @@ public class ClientX509Util extends X509Util {
     public SslContext createNettySslContextForClient(ZKConfig config)
         throws X509Exception.KeyManagerException, X509Exception.TrustManagerException, SSLException {
 
-        KeyManager km = null;
-        TrustManager tm = null;
+        KeyManager km;
+        TrustManager tm;
         String authProviderProp = System.getProperty(getSslAuthProviderProperty());
         if (authProviderProp == null) {
             String keyStoreLocation = config.getProperty(getSslKeystoreLocationProperty(), "");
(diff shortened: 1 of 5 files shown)


zookeeper-server/zookeeper-server-3.9.1/src/main/java/org/apache/zookeeper/common/ClientX509Util.java
```diff
@@ -0,0 +1,220 @@
...
+
+            if (authProvider == null) {
+                LOG.error("Auth provider not found: {}", authProviderProp);
+                throw new SSLException("Could not create SSLContext with specified auth provider: " + authProviderProp);
+            }
+            LOG.info("Using auth provider for client: {}", authProviderProp);
+            km = authProvider.getKeyManager();
+            tm = authProvider.getTrustManager();
+        }
+
```
jonmv: This switch is copied from `NettyServerCnxnFactory`; it tries the authProvider, when that is set, instead of jumping straight to file-based key and trust stores. There should be no other changes here, except for a refactoring to avoid some code duplication.

zookeeper-server/zookeeper-server-3.9.1/src/main/java/org/apache/zookeeper/common/ClientX509Util.java
```diff
@@ -0,0 +1,220 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ *
+ * <em>NOTE: Overridden because ZK 3.9 completely broke the SSL setup APIs; for clients, key and trust stores are
+ * now mandatory, unlike for servers, where it's still possible to provide a custom authProvider. This patch fixes that.</em>
+ * <p>
+ * X509 utilities specific for client-server communication framework.
+ */
+public class ClientX509Util extends X509Util {
```
bjorncs, line 46: Consider adding gitref of the original ZK source code so it's easier to update our "fork" in the future.

bjorncs (dismissed):

bjorncs (approved):

PR #29485, opened 2023-11-28
collapse co-nested and/andnot in first optimize pass

havardpe:
do not mention specific optimize passes in test
collapse second optimize pass into first

@baldersheim please review


baldersheim (approved):

baldersheim:
Build failure is unrelated

PR #29487, opened 2023-11-29
Move share dir to vespa-base

hmusum:
The goal is to get config definitions in vespa-base, so that they are available for node-admin as well. Not sure if this is the right way to do it, though. @arnej27959 could you have a look


arnej27959 (approved):
looks good to me

PR #29497, opened 2023-11-29
Define nessus-agent-group flag and make Dimension top-level class

hakonhall:


flags/src/test/java/com/yahoo/vespa/flags/DimensionTest.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author hakonhall
+ */
+class DimensionTest {
+    /**
+     * A compile-time test:  If this breaks you have most likely added (or removed) a dimension?
+     * If so you need to update the following:</p>
+     *
+     * <ul>
+     *     <li>Dimension validation in SystemFlagsDataArchive</li>
+     *     <li>Flag edit documenting dimension list in EditFlagDataCommandTest</li>
```
freva, line 14: This was removed in the internal PR

freva (approved):

flags/src/test/java/com/yahoo/vespa/flags/DimensionTest.java
hakonhall: right, good catch, I have fixed this in the next PR

PR #29513, opened 2023-11-30
Ensure mbus is not shut down while processing mbus requests

jonmv:
@bjorncs please review. @vekterli @baldersheim FYI.

### DIFF
--- a/jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java
+++ b/jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java
@@ -186,7 +186,7 @@ public class ApplicationLoader implements BootstrapLoader, ContainerActivator, C
         synchronized (appLock) {
             application = null;
         }
-        activateContainer(null);
+        try (DeactivatedContainer deactivated = activateContainer(null)) { }
         synchronized (appLock) {
             this.applicationInUseTracker = null;
         }
--- a/jdisc_core/src/main/java/com/yahoo/jdisc/test/TestDriver.java
+++ b/jdisc_core/src/main/java/com/yahoo/jdisc/test/TestDriver.java
@@ -73,9 +73,10 @@ public class TestDriver implements ContainerActivator, CurrentContainer {
         return loader.newContainerBuilder();
     }
 
+    /** Returns the deactivated container, with its container reference already released. */
     @Override
     public DeactivatedContainer activateContainer(ContainerBuilder builder) {
-        return loader.activateContainer(builder);
+        try (DeactivatedContainer deactivated = loader.activateContainer(buionf
     }
 
     @Override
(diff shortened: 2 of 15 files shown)


jonmv:
The idea here is that a `Request` that is created by an `MbusServer` while its owning `Container` is about to be switched out, will instead reference the newly created `Container`—this potentially leads to destruction of the `MessageBus` that is used by the `MbusServer`, e.g., when sessions change as a result of a document type being removed. As a first step when handling a `Message`, `MessageBus` will push itself as a `ReplyHandler`; when handling the reply that is eventually returned from the `MbusServer`, though handling in the new `Container`, it may have been `destroyed()` together with its owning `Container`, and then the `Reply` will simply be `discard()`ed, leaving the client on the other side of the RPC connection to wait for ever.

baldersheim:
Is this observed to be a problem currently => will this solve a real problem.
What is the risk that it will create a new set of problems ?

jonmv:
We have an unstable system test that feeds while removing a document type. Messages sometimes go missing in the container precisely when it reconfigures. I believe this may be the cause. There's also another race when we switch containers: we should take care to not release the creation reference to the previous container before we have stopped its servers, synced with them to ensure any in-progress `Request` creation is complete, and activated the new servers. We don't do that, but that is the next commit.

jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java
```diff
@@ -186,7 +186,7 @@ public void stop() throws Exception {
         synchronized (appLock) {
             application = null;
         }
-        activateContainer(null);
+        try (DeactivatedContainer deactivated = activateContainer(null)) { }
```
bjorncs, line 189: Can be simplified to `activacteContainer(null).close()`

jdisc_core/src/main/java/com/yahoo/jdisc/test/TestDriver.java
```diff
@@ -73,9 +73,10 @@ public ContainerBuilder newContainerBuilder() {
         return loader.newContainerBuilder();
     }
 
+    /** Returns the deactivated container, with its container reference already released. */
     @Override
     public DeactivatedContainer activateContainer(ContainerBuilder builder) {
-        return loader.activateContainer(builder);
+        try (DeactivatedContainer deactivated = loader.activateContainer(buionf
```
bjorncs, line 79: Typo

bjorncs (changes requested):

jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java
jonmv: `activateContainer(...)` returns `null` when there is no previous container.

jdisc_core/src/main/java/com/yahoo/jdisc/test/TestDriver.java
jonmv: What!?

bjorncs (approved):

jdisc_core/src/main/java/com/yahoo/jdisc/test/TestDriver.java
jonmv: Thanks.

jdisc_core/src/main/java/com/yahoo/jdisc/core/ApplicationLoader.java
bjorncs: Clever!

PR #29515, opened 2023-11-30
add parsing of special strings for inf/nan cell values

arnej27959:
@havardpe please review


havardpe:
we should also consider null for NaN, 1e999 for +inf and -1e999 for -inf (maybe even change json rendering to render +-inf as very big numbers to obtain a 'stable' solution)

havardpe (approved):

PR #29546, opened 2023-12-04
Use fake ZooKeeper database implementation for subset of CC tests

vekterli:
@hmusum please review
@jonmv FYI

The fake impl acts "as if" a single-node ZK quorum is present, so it cannot be directly used with most multi-node tests that require multiple nodes to actually participate in leader elections. Though we could probably expand the fake to handle this if desired.


hmusum (approved):
👍

PR #29565, opened 2023-12-06
Balder/add use estimate for fetch postings flag

baldersheim:
@havardpe PR


havardpe (commented):
looks good, but all explicit defaults should be false, not true

baldersheim:
Currently it is true in test code and false else where. Intention is to ensure new behavior gets good test coverage.

PR #29571, opened 2023-12-06
Let distributor metric decide cluster stability

mpolden:
@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
@@ -2,6 +2,7 @@
 package com.yahoo.vespa.hosted.provision.autoscale;
 
 import ai.vespa.metrics.ContainerMetrics;
+import ai.vespa.metrics.DistributorMetrics;
 import ai.vespa.metrics.HostedNodeAdminMetrics;
 import ai.vespa.metrics.SearchNodeMetrics;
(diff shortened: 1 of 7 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
```diff
@@ -107,7 +108,10 @@ private void consumeMetricsItem(Inspector item, ListMap<String, Double> values)
         item.field("values").traverse((ObjectTraverser)(name, value) -> values.put(name, value.asDouble()));
     }
 
-    private boolean clusterIsStable(Node node, NodeList applicationNodes) {
+    private boolean clusterIsStable(Node node, NodeList applicationNodes, ListMap<String, Double> nodeValues) {
+        if (Metric.redistributing.from(nodeValues) > 0) {
```
bratseth, line 112: We should probably rely only on this?

bratseth (approved):
We can probably remove the previous method - otherwise LGTM!

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/MetricsResponse.java
mpolden: Yes, but I don't think the merge pending metric will be exposed in the "autoscale" set until the application upgrades?

bratseth: Good point!

PR #29590, opened 2023-12-08
Bjorncs/commons logging

bjorncs:


jonmv (commented):
You mentioned we "export" 1.2 through that bridge. Could we have a problem now where users import ≥1.3?

bjorncs:
> You mentioned we "export" 1.2 through that bridge. Could we have a problem now where users import ≥1.3?

Yes, although unlikely as 1) bundle-plugin may have generated imports based on manifest from jcl-over-slf4j instead and 2) commons-logging is typically embedded in applications.

gjoranv (commented):
👍

PR #29602, opened 2023-12-09
Fix comparison of options

hmusum:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/RestartOnDeployForOnnxModelChangesValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/RestartOnDeployForOnnxModelChangesValidator.java
@@ -22,7 +22,7 @@ import static java.util.logging.Level.FINE;
  */
 public class RestartOnDeployForOnnxModelChangesValidator implements ChangeValidator {
 
-    private static final Logger log = Logger.getLogger(RestartOnDeployForOnnxModelChangesValidator.class.getSimpleName());
+    private static final Logger log = Logger.getLogger(RestartOnDeployForOnnxModelChangesValidator.class.getName());
 
     @Override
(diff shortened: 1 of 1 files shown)


hmusum:
or @balder or @valerijf or anyone

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/RestartOnDeployForOnnxModelChangesValidator.java
```diff
@@ -60,7 +60,9 @@ public List<ConfigChangeAction> validate(VespaModel currentModel, VespaModel nex
     private Optional<String> modelChanged(OnnxModelCost.ModelInfo a, OnnxModelCost.ModelInfo b) {
         if (a.estimatedCost() != b.estimatedCost()) return Optional.of("estimated cost");
         if (a.hash() != b.hash()) return Optional.of("model hash");
-        if (! a.onnxModelOptions().equals(b.onnxModelOptions())) return Optional.of("model option(s)");
+        if (a.onnxModelOptions().isPresent() && b.onnxModelOptions().isPresent()
```
baldersheim, line 63: So only if both exist, can they be different.. ?

baldersheim (dismissed):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/RestartOnDeployForOnnxModelChangesValidator.java
hmusum: We only care about them if they are both present, really. Not having options is legacy (apps on newer versions always have them, will be simplified when all apps have upgraded to a newer version). But I will fix it

baldersheim (approved):

bjorncs (approved):

PR #29605, opened 2023-12-11
fix default cycle time in fbench-usage to 0 MERGEOK

chgzm:
Hi,

The default fbench cycle time was set to 0 in [this PR](https://github.com/vespa-engine/vespa/pull/23876). However, the corresponding usage was not updated, so fixed it.


baldersheim (approved):
👍

PR #29647, opened 2023-12-13
Amdahl's law needs due respect. By allocating bitvector and zero init…

baldersheim:
…ializing it in the producing thread we achieve:

  - Shift work from sequential path to parallell path.
  - Avoid filling master threads cache during bitvector creation.
  - Pull directly into correct workers cache.
  - And increase the chance the memory is allocated in a numa region close to you.

@geirst @havardpe PR


baldersheim:
Closing for now

PR #29649, opened 2023-12-14
Amdahl's law needs due respect. By allocating bitvector and zero init…

baldersheim:
…ializing it in the producing thread we achieve:

  - Shift work from sequential path to parallell path.
  - Avoid filling master threads cache during bitvector creation.
  - Pull directly into correct workers cache.
  - And increase the chance the memory is allocated in a numa region close to you.

@havardpe @geirst PR
This contains the first half of #29647. I want to see effect them separately.

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
+++ b/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
@@ -68,9 +68,12 @@ PostingListSearchContextT<DataT>::fillArray()
 
 template <typename DataT>
 struct PostingListSearchContextT<DataT>::FillPart : public vespalib::Runnable {
-    FillPart(const PostingStore& posting_store, const DictionaryConstIterator & from, size_t count,
-             BitVector & bv, uint32_t limit)
-        : _posting_store(posting_store), _bv(bv),
+    FillPart(const PostingStore& posting_store, const DictionaryConstIterator & from, size_t count, uint32_t limit)
+        : FillPart(posting_store, from, count, nullptr, limit)
+    { }
+    FillPart(const PostingStore& posting_store, const DictionaryConstIterator & from, size_t count, BitVector * bv, uint32_t limit)
+        : _posting_store(posting_store),
+          _bv(bv),
           _docIdLimit(limit),
           _from(from),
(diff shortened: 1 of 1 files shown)


searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
```diff
@@ -68,31 +68,38 @@ PostingListSearchContextT<DataT>::fillArray()
...
-    BitVector               &_bv;
-    uint32_t                 _docIdLimit;
-    DictionaryConstIterator  _from;
-    DictionaryConstIterator  _to;
+    const PostingStore        &_posting_store;
+    BitVector                 *_bv;
+    uint32_t                   _docIdLimit;
+    DictionaryConstIterator    _from;
+    DictionaryConstIterator    _to;
+    std::unique_ptr<BitVector> _owned_bv;
```
havardpe, line 102: it is polite to initialize this in the constructor as well

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
```diff
@@ -68,31 +68,38 @@ PostingListSearchContextT<DataT>::fillArray()
...
+          _bv(bv),
           _docIdLimit(limit),
           _from(from),
           _to(from)
     {
         _to += count;
     }
     void run() override {
+        if (_bv == nullptr) {
+            _owned_bv = BitVector::create(_docIdLimit);
```
havardpe, line 85: are the bitvectors the same size? does it matter?

havardpe (dismissed):

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
baldersheim: Agree, fixed

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
baldersheim: They are the same.  And it would not matter, except that it would indicate bad flow of docIdLimit :)

PR #29656, opened 2023-12-14
Recursively merge bitvectors with thread bundle.

baldersheim:
@geirst @havardpe PR


baldersheim:
Focus on #29660 instead as it has higher potential.

PR #29552, opened 2023-12-05
Update ZK client version to 3.9.1

jonmv:
@bjorncs

PR #29658, opened 2023-12-14
Jonmv/zk 3.9.1 clients 2

jonmv:
This is a variant of #29552 where we depend on our modified `zookeeper-server` instead of the raw ZK artefact, to avoid having to patch it in multiple places. I think this is better.


jonmv:
@hmusum please review. @baldersheim @bjorncs FYI.

jonmv:
The changes, summarised, are:

1. Depend on modified `zookeeper-server-3.9.1` etc. in the client setup (`zkfacade`, `zookeeper-command-line-client`), instead of the raw ZK artefact, so we have a single place we modify that.
2. Get `zookeeper-client-common` through `zkfacade` for `clustercontroller-core`, to avoid false imports of code referenced by our modified ZK (specifically, netty packages). 
3. Modify `ClientX509Util` to directly use the Vespa custom TLS setup, instead of using the auth provider indirection, which wasn't intended for clients anyway.

hmusum (approved):
👍

PR #29670, opened 2023-12-15
- Add a doom that will not expire until it does not matter anymore.

baldersheim:
- Doom can never be null.
- Wire doom into fillBitVector. Using it will be a separate PR.

@havardpe or @geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
+++ b/searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
@@ -68,11 +68,14 @@ PostingListSearchContextT<DataT>::fillArray()
 
 template <typename DataT>
 struct PostingListSearchContextT<DataT>::FillPart : public vespalib::Runnable {
-    FillPart(const PostingStore& posting_store, const DictionaryConstIterator & from, size_t count, uint32_t limit)
-        : FillPart(posting_store, from, count, nullptr, limit)
+    FillPart(const vespalib::Doom & doom, const PostingStore& posting_store, const DictionaryConstIterator & from,
+             size_t count, uint32_t limit)
+        : FillPart(doom, posting_store, from, count, nullptr, limit)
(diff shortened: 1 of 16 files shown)


searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
```diff
@@ -95,6 +99,7 @@ struct PostingListSearchContextT<DataT>::FillPart : public vespalib::Runnable {
             if (__builtin_expect(key < _docIdLimit, true)) { _bv->setBit(key); }
         });
     }
+    const vespalib::Doom       _doom;
```
geirst, line 102: Was it intentional to take a copy here?

geirst (dismissed):
👍

vespalib/src/vespa/vespalib/util/doom.h
```diff
@@ -19,6 +19,7 @@ class Doom {
     duration soft_left() const noexcept { return _softDoom - _clock.getTimeNS(); }
     duration hard_left() const noexcept { return _hardDoom - _clock.getTimeNS(); }
     bool isExplicitSoftDoom() const noexcept { return _isExplicitSoftDoom; }
+    static const Doom & armageddon() noexcept;
```
havardpe, line 22: what about 'never' ?

havardpe (dismissed):

searchlib/src/vespa/searchlib/attribute/postinglistsearchcontext.hpp
baldersheim: Yes, you avoid an indirection as it wraps a reference.

vespalib/src/vespa/vespalib/util/doom.h
baldersheim: Fixed, and simplified as discussed.

PR #29676, opened 2023-12-15
Revert "Revert "Balder/separate hot cold path tomake fast path faster""

baldersheim:
Reverts vespa-engine/vespa#29673


baldersheim:
Superceeded by #29679

PR #29687, opened 2023-12-17
- When estimating number of hits in range measure at both ends of ran…

baldersheim:
…ge, as they might be very different.

- Reduce the effect of outliers in the measured range by considering the global rate.

@geirst PR


geirst (approved):
👍

PR #29695, opened 2023-12-18
Fix semgrep warnings

tokle:


searchlib/src/main/java/com/yahoo/searchlib/gbdt/XmlHelper.java
```diff
@@ -27,38 +27,25 @@ abstract class XmlHelper {
...
 
-    private static DocumentBuilderFactory createDocumentBuilderFactory() throws ParserConfigurationException {
-        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
-        factory.setNamespaceAware(true);
-        factory.setXIncludeAware(false);
-
-        // XXE prevention
-        factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
-        factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
-        factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```
hmusum, line 58: Seems like this feature is not set in XML.getDocumentBuilder methods. Intended?

tokle: This PR is a bit more strict, need to verify that we can indeed do this. I'll split this in 2 PRs, and we can merge the first part now.

baldersheim:
@tokle State of this one, close or rebase and complete ?

PR #29713, opened 2023-12-19
Consider the nordic variations over A and O as the other european var…

baldersheim:
…iations.

@havardpe or @geirst PR


havardpe (dismissed):
seems legit...

PR #29719, opened 2023-12-20
- Having expensive asserts during shutdown is not worthwhile. It does…

baldersheim:
… not prevent persisting any errors,

  nor does it provide any information about what went wrong.

@toregge @geirst PR


toregge (commented):
👍

PR #29730, opened 2023-12-21
Revert "Jonmv/keep config change actions in dev" MERGEOK

hmusum:
Reverts vespa-engine/vespa#29692

Getting invalid application package for manual deployments (e.g. in dev) now, e.g. when changing certificate. Looks like it fails instead of auto-overriding

### DIFF
--- a/config-model-api/abi-spec.json
+++ b/config-model-api/abi-spec.json
@@ -771,9 +771,7 @@
     "attributes" : [
       "public"
     ],
-    "methods" : [
-      "public java.util.Map messagesById()"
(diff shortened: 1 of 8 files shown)


hmusum:
Seems to me that validators in `com.yahoo.vespa.model.application.validation.Validation.validateChanges()` are run directly (not using `exexution.run()`), so exceptions are not handled in the same way as for the other validators.
This means that the exception will propagate up and an exception will be thrown for manually deployed applications where validations should be auto-overridden.

PR #29732, opened 2023-12-21
- Separate methods for lowercasing, and lowercasing and folding.

baldersheim:
- Hide implementations and use accessors.
- Minor code cleanup.

@bratseth PR


bratseth (approved):
👍

PR #29733, opened 2023-12-21
Add support for collapsing on multiple fields

tomglk:
The FieldCollapsingSearcher is currently only able to collapse on a single field.

This PR adds support to collapse on an arbitrary amount of fields.

**State of the PR:**
~~The PR shows a possible solution, which needs feedback for the implemented approach.
Test Cases are missing, Code Review is needed, but later.~~

~~**First goal is to clarify if we agree on the general idea.**~~

We agree on the general idea.
Code Review was done.
Test Cases are in progress.

I think that multi-field collapsing can be understood in multiple ways:
- collapse by the fields **one by one**, looking at their values & collapsesizes separately after each other
- collapse by the **combination** of the fields, meaning the value-combinations of the fields all at once

This PR implements the first understanding.

---
**Changes in FieldCollapsingSearcher:**
- Allow passing a comma-separated list of fields in collapsefield
- Use collapsesize to override the default for all fields
- Use collapsesize.fieldname to override the size for a specific field
- collapse the result multiple times by iterating through the collapsefields

This means that the collapsesize for field `foo` is determined like this:
- IF collapsesize.foo is set: use that
- ELSE IF collapsesize is set: use that
- ELSE: use default collapsesize

---
**Closed topics:**
- The order of fields is important (see testCollapsingWithMultipleFields), opinions towards that?
    - This is fine.
- The parameter is still called `collapsefield` (to keep backwards-compatibility), is that a deal-breaker?
    - It is not.
- Regarding the different understandings from the top: Do you agree on those? Which should we support, both?
    - We support the first understanding - use the fields sequentially.

**Thoughts about the field-order topic:**
If the order of the fields would not matter, how should we determine which hits are kept? Should we try to keep as many hits per collapsefield as possible (reaching as close as possible to collapsesize)?
What happens if there is no clear "winner", meaning we can only get n hits, no matter how we combine the collapsefields?

In short: I think that the collapse-logic can become very complex if we don't simply take the fields in the order they are configured.

You are welcome to challenge that. :)

---


vespa-headless (commented):
Great! Collapsing the fields one by one (or, locally not globally) sounds good to me, and I agree it should then depend on ordering.

Reusing the parameter names like this is also good.

(I agree some more testing is needed but otherwise the code LGTM.)

bratseth (commented):
Great! Collapsing the fields one by one (or, locally not globally) sounds good to me, and I agree it should then depend on ordering.

Reusing the parameter names like this is also good.

(I agree some more testing is needed but otherwise the code LGTM.)

container-search/src/main/java/com/yahoo/prelude/searcher/FieldCollapsingSearcher.java
```diff
@@ -159,19 +181,35 @@ private void collapse(Result result, Map<String, Integer> knownCollapses,
                 continue;
             }
 
-            if (knownCollapses.containsKey(collapseId)) {
-                int numHitsThisField = knownCollapses.get(collapseId);
+            // prepending the fieldname is necessary to distinguish between values in the different collapsefields
+            String collapseKey = collapseField + "_" + collapseId;
```
bratseth, line 185: Underscore is legal in field names so I suggest using e.g @.

tomglk:
Update for the documentation: https://github.com/vespa-engine/documentation/pull/3051

bratseth (approved):
👍  Ready to merge?

tomglk:
yes

PR #29753, opened 2023-12-29
Skip upgrade when OS version is unavailable

mpolden:

Previously we assumed that the requested OS version was always available, but
that may not be true for a number of reasons:

- Image publishing is broken
- Image is GC-ed due to age
- Cloud account is incorrectly setup and lacks the expected images

This adds a check that verifies whether the wanted OS version is available,
before triggering an upgrade. In case an image disappears immediately after
checking, the host provisioner will still fall back to a different image, but
we'll avoid retrying the upgrade in a loop.

@bratseth


bratseth (approved):
👍

PR #29760, opened 2024-01-02
Use a static cache

jonmv:
@mpolden please review.


jonmv:
Probably fails some unit tests. Blargh. I'll do the DI by hand, then.

PR #29762, opened 2024-01-02
Use std::from_chars instead of strtoll.

baldersheim:
@vekterli PR


vekterli (approved):
:+1: had to stare at the documentation for these functions for a while to cross-check pointer semantics... 🙂

For pondering: should we add _"strto(u)ll-ish functionality, but not built around locale-dependent legacy stuff"_ as a vespalib util, as this is not the first and presumably not last time we replace usages of it with something based on `from_chars`?

baldersheim:
Yes, we should probably do some unification. @arnej27959 has made some common float/double stuff. But since the devil hides in the details that requires great care. I will be happy to integrate if you make it.

PR #29766, opened 2024-01-02
Jonmv/reapply keep config changes in dev

jonmv:
@bratseth or @hmusum please review and merge.


hmusum (dismissed):

bratseth (commented):
This is becoming pretty hard to understand - it indicates there's something wrong with the model, probably should change the validator API? But maybe it's better to merge this and do it later. @hmusum please review as well.

And in any case, I think we need a unit test of the thing that wen't wrong with the first try.

jonmv:
Yes, I had those thoughts about the API as well. I may as well change it right away, while it's fresh in my mind.

hmusum (dismissed):

PR #29770, opened 2024-01-02
Avoid interface key MERGEOK

mpolden:
@bjorncs


mpolden:
Never mind. This `CloudAccount` is a concrete class.

PR #29774, opened 2024-01-02
Use std::from_chars instead of strtod. Also improve handling of float…

baldersheim:
… vs double.

@vekterli PR

### DIFF
--- a/searchlib/src/vespa/searchlib/query/query_term_simple.cpp
+++ b/searchlib/src/vespa/searchlib/query/query_term_simple.cpp
@@ -32,7 +32,7 @@ bool isFullRange(vespalib::stringref s) noexcept {
 struct IntDecoder {
     static int64_t fromstr(const char * q, const char * qend, const char ** end) noexcept {
         int64_t v(0);
-        for (;q < qend && isspace(*q); q++);
+        for (;q < qend && (isspace(*q) || (*q == '+')); q++);
         std::from_chars_result err = std::from_chars(q, qend, v, 10);
         if (err.ec == std::errc::result_out_of_range) {
             v = (*q == '-') ? std::numeric_limits<int64_t>::min() : std::numeric_limits<int64_t>::max();
@@ -48,7 +48,7 @@ template <typename T>
 struct FloatDecoder {
     static T fromstr(const char * q, const char * qend, const char ** end) noexcept {
         T v(0);
-        for (;q < qend && isspace(*q); q++);
+        for (;q < qend && (isspace(*q) || (*q == '+')); q++);
         std::from_chars_result err = std::from_chars(q, qend, v);
         if (err.ec == std::errc::result_out_of_range) {
             v = (*q == '-') ? -std::numeric_limits<T>::infinity() : std::numeric_limits<T>::infinity();
--- a/searchlib/src/vespa/searchlib/query/query_term_simple.cpp
+++ b/searchlib/src/vespa/searchlib/query/query_term_simple.cpp
@@ -4,7 +4,6 @@
 #include "base.h"
 #include <vespa/vespalib/objects/visit.h>
 #include <vespa/vespalib/util/classname.h>
-#include <vespa/vespalib/locale/c.h>
 #include <cmath>
 #include <limits>
 #include <charconv>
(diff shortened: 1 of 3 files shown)


searchlib/src/vespa/searchlib/query/query_term_simple.cpp
```diff
@@ -122,94 +160,72 @@ QueryTermSimple::getIntegerRange() const
...
 {
     lower = - std::numeric_limits<double>::max();
     upper =   std::numeric_limits<double>::max();
-    return getAsNumericTerm(lower, upper, DoubleDecoder());
+    return getAsNumericTerm(lower, upper, FloatDecoder<double>());
+}
+
+bool QueryTermSimple::getAsFloatTerm(float & lower, float & upper) const noexcept
+{
+    lower = - std::numeric_limits<float>::max();
```
vekterli, line 219: Entirely optional: looks like C++11 has [`std::numeric_limits<T>::lowest()`](https://en.cppreference.com/w/cpp/types/numeric_limits/lowest) which is equal to `- max()` for floating point values, but might be more immediately obvious when reading the code

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
```diff
@@ -4,19 +4,66 @@
...
+    static int64_t nearestDownwd(int64_t n, int64_t min) noexcept { return (n > min ? n - 1 : n); }
+    static int64_t nearestUpward(int64_t n, int64_t max) noexcept { return (n < max ? n + 1 : n); }
+};
+
+template <typename T>
+struct FloatDecoder {
+    static T fromstr(const char * q, const char * qend, const char ** end) noexcept {
+        T v(0);
+        for (;q < qend && isspace(*q); q++);
+        std::from_chars_result err = std::from_chars(q, qend, v);
```
vekterli, line 52: Do we support having a leading (no-op) `+` character in floating point terms? If so, it looks like we have to explicitly consume it ourselves (a-la leading whitespace), as `std::from_chars` [does not implicitly consume it](https://en.cppreference.com/w/cpp/utility/from_chars):
> Floating-point parsers: Expects the pattern identical to the one used by [std::strtod](https://en.cppreference.com/w/cpp/string/byte/strtof) in the default ("C") locale, except that
> * the plus sign is not recognized outside of the exponent (only the minus sign is permitted at the beginning)

If we do support this, we should probably add a unit test for it as well.

vekterli, line 52: Additionally, do we support hex floating point values? If so, it looks like we might have to detect+consume any leading `0x` and invoke `from_chars` with [std::chars_format::hex](https://en.cppreference.com/w/cpp/utility/chars_format)

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
```diff
@@ -4,19 +4,66 @@
...
+        if (err.ec == std::errc::result_out_of_range) {
+            v = (*q == '-') ? std::numeric_limits<T>::min() : std::numeric_limits<T>::max();
```
vekterli, line 54: It looks like the prior usage of `strtod` didn't check for `errno == ERANGE`, which would leave the end value as either `-inf` or `+inf`, depending on the sign of the input.

1. Is this an intended behavior?
2. If not, should the lower bound be `-max()` (i.e. `std::numeric_limits<T>::lowest()`) to mirror `getAsFloatTerm()`?

vekterli (dismissed):
:+1: a few question on parsing semantics before/after

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
baldersheim: We do not support hexencoded integers, so we should not support hex encoded floats either.

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
baldersheim: Using lowest.

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
baldersheim: +/- infinity it is.

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
baldersheim: Concluded that +/- infinity is even better.

searchlib/src/vespa/searchlib/query/query_term_simple.cpp
baldersheim: There were no tests for + prefixed numbers, but it was supported. So keeping that support.

PR #29782, opened 2024-01-03
Keep stack trace no execution exception

bratseth:


baldersheim (changes requested):
Looks good, but you need to update the ABI.

baldersheim (approved):

PR #29797, opened 2024-01-04
Jonmv/refactor validation

jonmv:
@bratseth please review. Not sure how happy I am with this, but at least it ensures most errors are seen at once, and that a few more actions are triggered when deploying changes to `dev` that would trigger these actions elsewhere, with an override.


bratseth (approved):
Great!

PR #29812, opened 2024-01-05
Reset downtime at resume

hakonhall:
* Defines 2 new node history events: suspended and resumed.  AFAIK this should be handled transparently by any/all clients.
* The NodeFailer measures the time a node has been down from the latest of the following events: the node went down, the node was activated, the node was resumed.  The last one (resumed) is new with this PR.
* Whether a node is down or not is now based on the node's history instead of asking the Orchestrator directly.  It is the NodeHealthTracker maintainer that queries the Orchestrator and adds up and down events.


mpolden (approved):
Great!

PR #29824, opened 2024-01-05
No normalization should take place for streaming search. This is hand…

baldersheim:
…led symmetrically in the backend during query time.

@geirst or @bratseth PR


geirst (dismissed):
👍  But I would have liked to see a unit test for this

bratseth (dismissed):
👍

baldersheim:
Unit test added for IndexInfo.

PR #29845, opened 2024-01-10
Increase alignment for bit vectors 64 => 256 bytes

baldersheim:
@toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/common/bitvector.h
+++ b/searchlib/src/vespa/searchlib/common/bitvector.h
@@ -281,6 +281,8 @@ public:
     static UP create(Index numberOfElements);
     static UP create(const BitVector & rhs);
     static void consider_enable_range_check();
+    static Index numWords(Index bits) noexcept { return wordNum(bits + 1 + (WordLen - 1)); }
+    static Index numBytes(Index bits) noexcept { return numWords(bits) * sizeof(Word); }
 protected:
     using Alloc = vespalib::alloc::Alloc;
     VESPA_DLL_LOCAL BitVector(void * buf, Index start, Index end) noexcept;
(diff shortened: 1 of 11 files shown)


searchlib/src/vespa/searchlib/common/bitvector.h
```diff
@@ -295,7 +295,7 @@ class BitVector : protected BitWord
     static Index numWords(Index bits) noexcept { return wordNum(bits + 1 + (WordLen - 1)); }
     static Index numBytes(Index bits) noexcept { return numWords(bits) * sizeof(Word); }
     size_t numWords() const noexcept { return numWords(size()); }
-    static constexpr size_t getAlignment() noexcept { return 0x40u; }
+    static constexpr size_t getAlignment() noexcept { return 0x100u; }
```
toregge, line 298: Unfortunately, https://github.com/vespa-engine/vespa/blob/c4e33003e5ce3f385951c107714ede8556ef8083/searchlib/src/vespa/searchlib/diskindex/bitvectordictionary.cpp#L61
indicates that we currently don't store the aligned bitvector byte size  in the file header, thus old bitvector files will be misread when changing the alignment. We likely need a getLegacyFileBytes() method that returns old vector size based on old alignment, used when bit the vector file lacks a file header tag for aligned bitvector byte size.
We should add a file header tag for new bitvector files, and use that value if present instead of the implicit legacy value.

toregge (changes requested):

searchlib/src/vespa/searchlib/common/bitvector.h
baldersheim: Solved as suggested.

baldersheim:
@toregge PTAL

toregge (approved):

searchlib/src/vespa/searchlib/common/fileheadertags.cpp
```diff
@@ -0,0 +1,16 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+#include "fileheadertags.h"
+
+namespace search::tags {
+// Do not change these constants, they are persisted in many file headers.
+vespalib::string FREEZE_TIME("freezeTime");
+vespalib::string CREATE_TIME("createTime");
+vespalib::string FROZEN("frozen");
+vespalib::string DOCID_LIMIT("docidLimit");
```
toregge, line 10: Should be "docIdLimit"

baldersheim: #29848

PR #29847, opened 2024-01-10
Don't fill async rendering executor queue with tasks that blocks and prefer continue rendering in same thread

bjorncs:


baldersheim:
I suggest we discuss this tomorrow. The devil hides in the details here.

bjorncs:
@baldersheim PTAL

baldersheim (approved):

container-core/src/main/java/com/yahoo/processing/rendering/AsynchronousSectionedRenderer.java
```diff
@@ -334,26 +336,46 @@ private class DataListListener extends RendererListener {
...
+                RuntimeException exception = null;
+                while (!syncTasks.isEmpty()) {
+                    try {
+                        syncTasks.poll().run();
+                    } catch (RuntimeException e) {
+                        if (exception == null) exception = e;
+                        else exception.addSuppressed(e);
+                    }
+                }
+                if (exception != null) throw exception;
```
jonmv, line 377: This will hide exceptions thrown in the `try { }` block. Consider re-throwing those, with any additional exceptions here as suppressed.

bjorncs: The exception will only be hidden if `finally` throws a new exception right? That is acceptable imho.

jonmv: `try-with-resources` automatically handles this, btw., when the `close()` there throws; then, that is added as a suppressed exception to whatever was thrown inside the `try { }` block; or it is thrown as the main exception if nothing else was thrown.

PR #29849, opened 2024-01-11
- Do not reset correctly set _vectorSize to an incorrect value.

baldersheim:
@toregge PR


toregge (approved):

baldersheim:
Fixes a regression introduced in #29845

PR #29850, opened 2024-01-11
Revert "Revert "Update tenant test runner JUnit version as well""

jonmv:
@bjorncs or @baldersheim

### DIFF
--- a/vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/TestReport.java
+++ b/vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/TestReport.java
@@ -4,6 +4,7 @@ package com.yahoo.vespa.testrunner;
 import ai.vespa.hosted.cd.InconclusiveTestException;
 import com.yahoo.collections.Comparables;
 import com.yahoo.vespa.testrunner.TestRunner.Suite;
+import org.junit.platform.commons.util.ReflectionUtils;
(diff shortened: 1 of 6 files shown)


bjorncs (dismissed):

vespa-osgi-testrunner/src/main/java/com/yahoo/vespa/testrunner/TestReport.java
```diff
@@ -353,12 +356,18 @@ static void trimStackTraces(Throwable thrown, String testFrameworkRootClass) {
                 break;
             }
         }
+        if (cutoff == 0) {
+            while (--i >= 0 && (   stack[i].isNativeMethod()
+                                || stack[i].getClassName().startsWith("java.lang.")
+                                || stack[i].getClassName().startsWith("java.util.")));
+            cutoff = i + 1;
```
jonmv, line 363: JUnit seems to now do a stack trim by default, but a really poor one, leaving a tail of stack frames that we don't want, but which also doesn't contain the info we expected in _our_ trim code, so that also breaks 😢 
This is a best-effort attempt to remedy that. 
But, hmm ... maybe `java.lang.Thread` should be exempt here. That's actually a valid bottom.

bjorncs: 🤞

bjorncs (approved):

PR #29851, opened 2024-01-11
Avoid generic reduce to reduce gc pressure

jobergum:
Merge the two tensor.reduce calls to a single custom reduce for the exact reduce function we need for this, significantly reduces memory allocations. 

### DIFF
--- a/model-integration/src/main/java/ai/vespa/embedding/SpladeEmbedder.java
+++ b/model-integration/src/main/java/ai/vespa/embedding/SpladeEmbedder.java
@@ -13,7 +13,6 @@ import com.yahoo.language.process.Embedder;
 import com.yahoo.tensor.IndexedTensor;
 import com.yahoo.tensor.Tensor;
 import com.yahoo.tensor.TensorType;
-import com.yahoo.tensor.functions.Reduce;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


model-integration/src/main/java/ai/vespa/embedding/SpladeEmbedder.java
```diff
@@ -167,4 +158,42 @@ public void deconstruct() {
...
+
+    /**
+     * Custom max reduce over the sequence dimension of the output tensor. This
+     * to reduce GC pressure from generic Tensor.reduce operation.
+     *
+     * @param tensor the model output tensor of shape d1,dim where d1 is the sequence length and dim is size
+     *               of the vocabulary
+     * @return A tensor of shape d1,1 where each value is the max of the log of the relu of the input tensor
+     */
+    private static IndexedTensor customMaxReduceOverLogOfRelu(IndexedTensor tensor) {
```
bjorncs, line 170: Consider inlining this logic with for loop in `sparsify()` - the temporary tensor is not really needed.

model-integration/src/test/java/ai/vespa/embedding/SpladeEmbedderTest.java
```diff
@@ -43,6 +43,19 @@ public void testZeroTokens() {
         assertEquals(0, result.size());
     }
 
+    @Test
```
bjorncs, line 46: Either add a timeout or ignore the test by default

bjorncs (dismissed):

model-integration/src/main/java/ai/vespa/embedding/SpladeEmbedder.java
jobergum: Great feedback!

bjorncs (approved):
Nice! 👍

PR #29856, opened 2024-01-11
cloud deployment variant

andreer:

### DIFF
--- a/config-application-package/src/test/java/com/yahoo/config/application/HostedOverrideProcessorComplexTest.java
+++ b/config-application-package/src/test/java/com/yahoo/config/application/HostedOverrideProcessorComplexTest.java
@@ -3,8 +3,8 @@ package com.yahoo.config.application;
 
 import com.yahoo.config.application.api.ApplicationPackage;
 import com.yahoo.config.application.api.DeploymentInstanceSpec;
-import com.yahoo.config.application.api.xml.DeploymentSpecXmlReader;
 import com.yahoo.config.model.application.provider.FilesApplicationPackage;
+import com.yahoo.config.provision.CloudName;
 import com.yahoo.config.provision.Environment;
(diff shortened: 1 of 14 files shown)


config-application-package/src/test/java/com/yahoo/config/application/HostedOverrideProcessorComplexTest.java
```diff
@@ -52,8 +51,8 @@ public void testProdBetaUsWest2a() throws TransformerException {
                         <redundancy>1</redundancy>
                     </content>
                     <content id="filedocument" version="1.0">
-                        <nodes count="2" groups="2">
-                            <resources disk="32Gb" memory="8Gb" vcpu="4"/>
+                        <nodes count="2" groups="2" required="true">
```
jonmv, line 54: Why `required="true"` all of a sudden?

jonmv (dismissed):
LGTM, assuming you have an answer for the one question :)

config-application-package/src/test/java/com/yahoo/config/application/HostedOverrideProcessorComplexTest.java
andreer: It is special cased for the node element in https://github.com/vespa-engine/vespa/blob/master/config-application-package/src/main/java/com/yahoo/config/application/OverrideProcessor.java#L230-L238
I don't know exactly why this is done, but it would not happen if I had tested the override on another element and I'm pretty sure this PR doesn't change the behaviour, only makes it happen in the unit test.

andreer: Thanks, that makes perfect sense.

jonmv (approved):
Nice :)

PR #29857, opened 2024-01-11
Store config change actions on prepare, trigger on activate

jonmv:
@hmusum please review. 

This will enable automatic reindexing in self-hosted Vespa as well, whenever a schema change is activated. FYI @bratseth.


hmusum (approved):

jonmv:
I suppose this will break the system tests that assume reindexing must be triggered by the user.
It should also be accompanied by an update to the reindexing documentation.
I want to get started on those before I merge this.

jonmv:
Ah, I forgot, the maintainer that _redeploys_ is in the node repository, so this will only make reindexing _ready_ automatically, but to start it, a new, manual deployment still has to be made in self-hosted.

PR #29873, opened 2024-01-12
Revert "Revert "Balder/unify attributes over streaming indexed""

baldersheim:
Reverts vespa-engine/vespa#29872
@geirst PR


streamingvisitors/src/vespa/vsm/searcher/utf8exactstringfieldsearcher.cpp
```diff
@@ -7,6 +7,13 @@ using search::streaming::QueryTermList;
 
 namespace vsm {
 
+UTF8ExactStringFieldSearcher::UTF8ExactStringFieldSearcher(FieldIdT fId)
+    : UTF8StringFieldSearcherBase(fId)
+{
+    match_type(EXACT);
+    normalize_mode(Normalizing::LOWERCASE);
```
geirst, line 14: Do we have a unit test for this behavior? If not, I think it should be added.

geirst (dismissed):
👍

streamingvisitors/src/vespa/vsm/searcher/utf8exactstringfieldsearcher.cpp
baldersheim: We are currently testing only exact/lowercase.
I will add tests for the missing combinations in a separate PR. This is not the only missing combination.

PR #29875, opened 2024-01-12
Always delay config until after restart when loading models, for rest…

jonmv:
…arting clusters

@hmusum


jonmv:
Hmm, this still isn't waterproof, though: if another deployment is made shortly after one that triggers a restart, but before the nodes have a chance to restart, that new config won't be delayed after all, right?

configserver/src/main/java/com/yahoo/vespa/config/server/modelfactory/ActivatedModelsBuilder.java
```diff
@@ -118,7 +123,7 @@ protected Application buildModelVersion(ModelFactory modelFactory,
                 wantedNodeVespaVersion);
         MetricUpdater applicationMetricUpdater = metrics.getOrCreateMetricUpdater(Metrics.createDimensions(applicationId));
         ServerCache serverCache = new ServerCache(configDefinitionRepo, zkClient.getUserConfigDefinitions());
-        return new Application(modelFactory.createModel(modelContext),
+        return new Application(withDeferredConfigForRestartingClusters(modelFactory.createModel(modelContext)),
```
jonmv, line 126: As far as I could see, this builder is the only one that is used for models that end up being used for serving config. Please correct me if I'm wrong, and the prepared-models-builder also does that.

hmusum: Yes, that's right models built by this will serve config

hmusum (approved):
Looks good to me. Consider adding feature flag to enable this, if testing is considered good enough (system testing, integration tests) I think it can go as is.

jonmv:
Yes, we have tests specifically for this.

PR #29878, opened 2024-01-12
Fail if missing access control filter for all clouds

hmusum:
Found when trying to add a new cloud provider. I assume this should fail for all "public" clouds.


tokle:
This validator only runs in hosted non-public systems. But probably good anyway.

tokle (approved):

PR #29882, opened 2024-01-12
Store multiple resource suggestions

olaaun:

Will remove the old suggestion object after API clients migrates to suggestion list


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/ApplicationSerializer.java
```diff
@@ -140,6 +142,7 @@ private static void toSlime(Cluster cluster, Cursor clusterObject) {
         toSlime(cluster.groupSize(), clusterObject.setObject(groupSizeKey));
         clusterObject.setBool(requiredKey, cluster.required());
         toSlime(cluster.suggested(), clusterObject.setObject(suggestedKey));
+        toSlime(cluster.suggestions(), clusterObject.setArray(suggestionsKey));
```
mpolden, line 145: Looks like the same information is stored twice now, since `suggested` is simply the first element of `suggestions`. Consider adding a TODO for removing `suggested` from serialization (needs to happen after this rolls out).

mpolden (approved):
Great!

PR #29883, opened 2024-01-12
Drop tokenize expressions from ilscript for streaming mode.

baldersheim:
@geirst PR


config-model/src/test/java/com/yahoo/schema/derived/IndexingScriptTestCase.java
```diff
@@ -0,0 +1,40 @@
+package com.yahoo.schema.derived;
```
geirst, line 1: Please add copyright.

geirst (dismissed):
👍

config-model/src/test/java/com/yahoo/schema/derived/IndexingScriptTestCase.java
baldersheim: Fixed

PR #29897, opened 2024-01-14
Always add an indexing docproc, also for streaming search.

baldersheim:
@geirst PR

### DIFF
--- a/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
+++ b/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
@@ -1440,48 +1440,34 @@ public class ContentClusterTest extends ContentBaseTest {
         assertGroupsAllowedDown(2, 1, 2);
     }
 
-    private void assertIndexingDocprocEnabled(boolean indexed, boolean force) {
+    private void assertIndexingDocprocEnabled(boolean indexed) {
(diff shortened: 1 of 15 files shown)


config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
```diff
@@ -1456,25 +1456,32 @@ private void assertIndexingDocprocEnabled(boolean indexed, boolean force, boolea
                 "</services>";
         VespaModel model = createEnd2EndOneNode(new TestProperties(), services);
         var searchCluster = model.getContentClusters().get("search").getSearch();
-        assertEquals(expEnabled, searchCluster.getIndexingDocproc().isPresent());
+        if (force) {
+            assertEquals("default", searchCluster.getIndexingDocproc().getClusterName("search"));
```
geirst, line 1460: Do we need to handle the tests on lines 1459-1466 differently? They all seem the same.

geirst (dismissed):
👍

config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
baldersheim: Simplified

jonmv:
This is fine, but it broke a test that relies on an additional field not being populated with streaming mode.

PR #29902, opened 2024-01-15
Delay expiry of `dynamic-heap-size`

bjorncs:


hmusum (approved):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -386,7 +386,7 @@ public class Flags {
 
     public static final UnboundBooleanFlag DYNAMIC_HEAP_SIZE = defineFeatureFlag(
             "dynamic-heap-size", false,
```
baldersheim, line 388: I think this has rolled out with a default of true. Update default here too .... ?

PR #29903, opened 2024-01-15
Move cluster restarting -> defer config inside model

jonmv:
@hmusum


hmusum (changes requested):
LGTM, just need to update ABI spec

hmusum (approved):

PR #29910, opened 2024-01-15
Prepare removal of feature flag 'dynamic-heap-size'

bjorncs:

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
+++ b/config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
@@ -109,7 +109,7 @@ public interface ModelContext {
         @ModelFeatureFlag(owners = {"arnej"}) default String logFileCompressionAlgorithm(String defVal) { return defVal; }
         @ModelFeatureFlag(owners = {"baldersheim"}, comment = "Select summary decode type") default String summaryDecodePolicy() { return "eager"; }
         @ModelFeatureFlag(owners = {"vekterli"}) default int contentLayerMetadataFeatureLevel() { return 0; }
-        @ModelFeatureFlag(owners = {"bjorncs"}) default boolean dynamicHeapSize() { return false; }
+        @ModelFeatureFlag(owners = {"bjorncs"}, removeAfter = "8.288") default boolean dynamicHeapSize() { return true; }
         @ModelFeatureFlag(owners = {"hmusum"}) default String unknownConfigDefinition() { return "warn"; }
         @ModelFeatureFlag(owners = {"hmusum"}) default int searchHandlerThreadpool() { return 2; }
         @ModelFeatureFlag(owners = {"vekterli"}) default long mergingMaxMemoryUsagePerNode() { return -1; }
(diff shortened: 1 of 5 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
```diff
@@ -311,7 +309,6 @@ public FeatureFlags(FlagSource source, ApplicationId appId, Version version) {
         }
         @Override public boolean alwaysMarkPhraseExpensive() { return alwaysMarkPhraseExpensive; }
         @Override public int contentLayerMetadataFeatureLevel() { return contentLayerMetadataFeatureLevel; }
-        @Override public boolean dynamicHeapSize() { return dynamicHeapSize; }
```
baldersheim, line 314: Keep this and the the flag with updated default.
Then GC flag override this version is rolled out.
Then you can remove this and the flag definition afterwards

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -385,8 +385,8 @@ public class Flags {
             INSTANCE_ID);
 
     public static final UnboundBooleanFlag DYNAMIC_HEAP_SIZE = defineFeatureFlag(
-            "dynamic-heap-size", false,
-            List.of("bjorncs"), "2023-09-21", "2024-02-15",
+            "dynamic-heap-size", true,
+            List.of("bjorncs"), "2023-09-21", "2024-01-15",
```
baldersheim, line 389: I think you need to keep the future date of February 1st

config-model-api/src/main/java/com/yahoo/config/model/api/ModelContext.java
```diff
@@ -109,7 +109,7 @@ interface FeatureFlags {
         @ModelFeatureFlag(owners = {"arnej"}) default String logFileCompressionAlgorithm(String defVal) { return defVal; }
         @ModelFeatureFlag(owners = {"baldersheim"}, comment = "Select summary decode type") default String summaryDecodePolicy() { return "eager"; }
         @ModelFeatureFlag(owners = {"vekterli"}) default int contentLayerMetadataFeatureLevel() { return 0; }
-        @ModelFeatureFlag(owners = {"bjorncs"}) default boolean dynamicHeapSize() { return false; }
+        @ModelFeatureFlag(owners = {"bjorncs"}, removeAfter = "8.288") default boolean dynamicHeapSize() { return true; }
```
baldersheim, line 112: Given the below comment, you probably need 289 here.

baldersheim (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/deploy/ModelContextImpl.java
bjorncs: The interface's default implementation defines the new default value. There is no need for keeping the wiring anymore.

baldersheim: Yes, that is correct. But you now leave a flagdefinition file that has no effect, other that than preventing an error messeage. That is why I like to keep the wiring one version longer and remove flag def and wiring in same PR.

hmusum (approved):

PR #29913, opened 2024-01-15
Add regular expression support to streaming search

vekterli:
@geirst and @toregge please review
@baldersheim FYI

Introduces an explicit regex query term node (which wraps an RE2 regex instance internally) and extends the existing UTF-8 flexible string searcher to use this query node.

Regex matching is optionally case (in)sensitive depending on the normalization mode used.

Note on `searcher/searcher_test.cpp`: this adds a magic sentinel `#` char prefix to query term parsing in the test to let a query term be interpreted as a regex rather than exact/prefix/suffix/substring match.


geirst (approved):
👍

baldersheim (approved):

PR #29914, opened 2024-01-15
Avoid generic reduce and keep PAD token embedding

jobergum:
- Approx 30% speedup of the evaluation on CPU by avoiding generic reduce 
- Retain the PAD token as have observed small degradation in accuracy that I cannot explain other than the pad removal optimization. 

Updated test in https://github.com/vespa-engine/system-test/pull/3193
  

bjorncs (approved):
👍

PR #29915, opened 2024-01-15
take strictness into account for flow/cost/sorting

havardpe:
use common code with adapters
make cost calculation const (sort index, not children) empty AND/ANDNOT estimate now equals input flow

@baldersheim please review


havardpe:
I will change the estimates to always return zero for empty collections to be conservative

baldersheim (approved):

PR #29916, opened 2024-01-15
Revert "Move cluster restarting -> defer config inside model"

jonmv:
Reverts vespa-engine/vespa#29903

@hmusum @baldersheim broke some tests.


hmusum (approved):

jonmv:
Seems this worked. And we had a few performance tests which triggered this unintentionally (change or RPC port in container), and didn't handle that.

PR #29931, opened 2024-01-16
Reapply defer container config changes when restarting

jonmv:
@hmusum


hmusum (approved):

baldersheim:
I am a bit reluctant to this change. Historically it has been a lot worse to delay config changes. You will then leave your system in a inconsistent state for a long time while doing orchestrated restart. That is why you must explicit set restartOnDeploy. All config changes should be handled ASAP, and that is true for most. Some are annotated as requiring restart. But that does not mean that the other config should not be be applied immediately.
It only means that a restart is required for everything to be perfect.

jonmv:
I assumed these considerations had already been made, as what I've done is just move some existing code, and actually make it work (the deferring of config activation). I also have the same concerns you offer here, and I'm easily swayed in this case; since this code never worked before, it's probably not that important, and making it work now could potentially cause other problems. Let's revert it, then, and clean up, instead.

PR #29933, opened 2024-01-16
Revert "Reapply defer container config changes when restarting"

jonmv:
Reverts vespa-engine/vespa#29931

@baldersheim @hmusum


jonmv:
Let me add the cleanup here as well, before we merge.

hmusum (approved):

PR #29936, opened 2024-01-16
- Only call getKey once.

baldersheim:
- Just use a presized array of native ints, instead of LinkedList.

@vekterli or @bjorncs PR


bjorncs (approved):

vekterli (commented):
:+1:

PR #29954, opened 2024-01-17
Arnej/extend verify ranksetup

arnej27959:
@havardpe please review
@baldersheim FYI


havardpe (approved):

havardpe:
we should also extend the unit test for verify rank setup to include streaming cases

PR #29959, opened 2024-01-17
Balder/create indexes to remove once

baldersheim:
@bratseth or @bjorncs or @jonmv PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
@@ -62,7 +62,11 @@ public interface Tensor {
     /** Returns whether this have any cells */
     default boolean isEmpty() { return size() == 0; }
 
-    /** Returns the number of cells in this */
+    /**
+     *  Returns the number of cells in this.
+     *  TODO Figure how to best return an int instead of a long
+     *  An int is large enough, and java is far better at int base loops than long
+     **/
     long size();
 
     /** Returns the value of a cell, or 0.0 if this cell does not exist */
--- a/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
@@ -146,7 +146,7 @@ public class Reduce<NAMETYPE extends Name> extends PrimitiveTensorFunction<NAMET
 
         return reducedBuilder.build();
     }
-    private static int [] createIndexesToKeep(TensorType argumentType, List<String> dimensions) {
+    private static int[] createIndexesToKeep(TensorType argumentType, List<String> dimensions) {
         Set<Integer> indexesToRemove = new HashSet<>(dimensions.size()*2);
         for (String dimensionToRemove : dimensions)
             indexesToRemove.add(argumentType.indexOfDimension(dimensionToRemove).get());
--- a/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
@@ -159,7 +159,7 @@ public class Reduce<NAMETYPE extends Name> extends PrimitiveTensorFunction<NAMET
         return indexesToKeep;
     }
 
-    private static TensorAddress reduceDimensions(int [] indexesToKeep, TensorAddress address) {
+    private static TensorAddress reduceDimensions(int[] indexesToKeep, TensorAddress address) {
         String[] reducedLabels = new String[indexesToKeep.length];
         int reducedLabelIndex = 0;
         for (int toKeep : indexesToKeep)
--- a/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
@@ -152,7 +152,7 @@ public class Reduce<NAMETYPE extends Name> extends PrimitiveTensorFunction<NAMET
             indexesToRemove.add(argumentType.indexOfDimension(dimensionToRemove).get());
         int [] indexesToKeep = new int[argumentType.rank() - indexesToRemove.size()];
         int toKeepIndex = 0;
-        for (int i = 0;i < argumentType.rank(); i++) {
+        for (int i = 0; i < argumentType.rank(); i++) {
             if ( ! indexesToRemove.contains(i))
                 indexesToKeep[toKeepIndex++] = i;
         }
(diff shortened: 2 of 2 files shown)


vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
```diff
@@ -62,7 +62,11 @@ public interface Tensor {
     /** Returns whether this have any cells */
     default boolean isEmpty() { return size() == 0; }
 
-    /** Returns the number of cells in this */
+    /**
+     *  Returns the number of cells in this.
+     *  TODO Figure how to best return an int instead of a long
```
bratseth, line 67: I don't think an int is enough. We could have a sizeAsInt() which would fail if the size doesn't fit, but this could also be done on the receiving side?

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
         TensorType reducedType = outputType(argument.type(), dimensions);
 
         // Reduce cells
-        Map<TensorAddress, ValueAggregator> aggregatingCells = new HashMap<>();
+        int [] indexesToKeep = createIndexesToKeep(argument.type(), dimensions);
```
bratseth, line 131:
```suggestion
        int[] indexesToKeep = createIndexesToKeep(argument.type(), dimensions);
```

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
...
         Tensor.Builder reducedBuilder = Tensor.Builder.of(reducedType);
         for (Map.Entry<TensorAddress, ValueAggregator> aggregatingCell : aggregatingCells.entrySet())
             reducedBuilder.cell(aggregatingCell.getKey(), aggregatingCell.getValue().aggregatedValue());
 
         return reducedBuilder.build();
-
     }
-
-    private static TensorAddress reduceDimensions(TensorAddress address, TensorType argumentType, TensorType reducedType, List<String> dimensions) {
+    private static int [] createIndexesToKeep(TensorType argumentType, List<String> dimensions) {
```
bratseth, line 149:
```suggestion
    private static int[] createIndexesToKeep(TensorType argumentType, List<String> dimensions) {
```

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
...
         Set<Integer> indexesToRemove = new HashSet<>(dimensions.size()*2);
         for (String dimensionToRemove : dimensions)
             indexesToRemove.add(argumentType.indexOfDimension(dimensionToRemove).get());
+        int [] indexesToKeep = new int[argumentType.rank() - indexesToRemove.size()];
```
bratseth, line 153:
```suggestion
        int[] indexesToKeep = new int[argumentType.rank() - indexesToRemove.size()];
```

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
...
+        int toKeepIndex = 0;
+        for (int i = 0;i < argumentType.rank(); i++) {
+            if ( ! indexesToRemove.contains(i))
+                indexesToKeep[toKeepIndex++] = i;
+        }
+        return indexesToKeep;
+    }
 
-        String[] reducedLabels = new String[reducedType.dimensions().size()];
+    private static TensorAddress reduceDimensions(int [] indexesToKeep, TensorAddress address) {
```
bratseth, line 162:
```suggestion
    private static TensorAddress reduceDimensions(int[] indexesToKeep, TensorAddress address) {
```

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
...
     }
-
-    private static TensorAddress reduceDimensions(TensorAddress address, TensorType argumentType, TensorType reducedType, List<String> dimensions) {
+    private static int [] createIndexesToKeep(TensorType argumentType, List<String> dimensions) {
         Set<Integer> indexesToRemove = new HashSet<>(dimensions.size()*2);
         for (String dimensionToRemove : dimensions)
             indexesToRemove.add(argumentType.indexOfDimension(dimensionToRemove).get());
+        int [] indexesToKeep = new int[argumentType.rank() - indexesToRemove.size()];
+        int toKeepIndex = 0;
+        for (int i = 0;i < argumentType.rank(); i++) {
```
bratseth, line 155:
```suggestion
        for (int i = 0; i < argumentType.rank(); i++) {
```

bratseth (dismissed):

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
```diff
@@ -128,31 +128,42 @@ else if (argument.type().dimensions().size() == 1 && argument instanceof Indexed
         TensorType reducedType = outputType(argument.type(), dimensions);
 
         // Reduce cells
-        Map<TensorAddress, ValueAggregator> aggregatingCells = new HashMap<>();
+        int[] indexesToKeep = createIndexesToKeep(argument.type(), dimensions);
+        // TODO cells.size() is most likely an overestimate, and might need a better heuristic
+        // But the upside is larger than the downside.
+        Map<TensorAddress, ValueAggregator> aggregatingCells = new HashMap<>((int)argument.size());
```
jonmv, line 134: Consider checking for truncation here, if we support (ever) support large enough tensors for that to happen.

vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
baldersheim: There is already a cast to int in quite a few places (not checked) already, so making a sizeAsInt might be worth it.

I think that 2G values in a single tensor will not be reached in a very long time. And when that is reached you will have to overcome all the limits in java and all its libraries. All collections operate with int sizes and arrays only accepts int as size and index.

vespajlib/src/main/java/com/yahoo/tensor/functions/Reduce.java
baldersheim: See comment above

jonmv (approved):

vespajlib/src/main/java/com/yahoo/tensor/Tensor.java
bratseth: Yes, but this is an interface that could be backed by anything ...

PR #29960, opened 2024-01-18
run verify for streaming clusters as well

arnej27959:
@baldersheim please review
@havardpe FYI


baldersheim (approved):

baldersheim (changes requested):
Factory requires some StreamingValidator tests to be updated

baldersheim (approved):

PR #29961, opened 2024-01-18
- Add sizeAsInt to allow for safe cast from long to int of the size o…

baldersheim:
…f a tensor.

@bratseth @jonmv @bjorncs PR


baldersheim:
Followup from #29959

bratseth (dismissed):

PR #29964, opened 2024-01-18
Use `string_view` for standalone DFA fuzzy match function

vekterli:

This is to allow for matching non-zero terminated strings, which will be needed if we are to support fuzzy matching of streaming tokenized terms.

Note that the underlying DFA API already takes in a `string_view`, so this is just moving an implicit `strlen` one step higher in the stack. This will pull the string data into the CPU cache ever so slightly earlier than before, but this is not expected to matter in practice since the data will be used immediately after anyway.

Added explicit `string_view` instantiations at the call sites to make implicit `strlen` more obviously visible than they currently are with implicit type conversions from `const char*`.

Also adds a utility constructor to hide default underlying DFA implementation.


geirst (approved):
👍

toregge (approved):

PR #29968, opened 2024-01-18
Jonmv/two phase deploy

jonmv:
@hmusum please review, but don't merge yet.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/response/SessionPrepareAndActivateResponse.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/response/SessionPrepareAndActivateResponse.java
@@ -8,6 +8,7 @@ import com.yahoo.container.jdisc.HttpRequest;
 import com.yahoo.restapi.SlimeJsonResponse;
 import com.yahoo.slime.Cursor;
 import com.yahoo.vespa.config.server.configchange.ConfigChangeActionsSlimeConverter;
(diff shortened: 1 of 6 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/http/v2/response/SessionPrepareAndActivateResponse.java
```diff
@@ -17,25 +18,19 @@
...
+        root.setBool("activated", result.activationFailure() == null);
 
-        // TODO: remove unused fields, but add whether activation was successful.
-        root.setString("tenant", tenantName.value());
-        root.setString("url", "http://" + request.getHost() + ":" + request.getPort() +
-                              "/application/v2/tenant/" + tenantName +
-                              "/application/" + applicationId.application().value() +
-                              "/environment/" + zone.environment().value() +
-                              "/region/" + zone.region().value() +
-                              "/instance/" + applicationId.instance().value());
```
hmusum, line 37: I don't think you can change field names and remove something here, this is a documented API:
https://docs.vespa.ai/en/reference/deploy-rest-api-v2.html#prepareandactivate

hmusum (changes requested):
Looks good, except changing/removing some stuff in response

jonmv:
Ah, I didn't think this one was public. I'll re-add those fields, then.

hmusum (approved):

PR #29969, opened 2024-01-18
Support fuzzy term matching in streaming search

vekterli:
@geirst please review
@toregge FYI

Uses a DFA-based matcher for max edits $`k \in \{ 1, 2 \}`$ and falls back to the legacy non-DFA matcher for all other values (including 0).

Currently only supports fuzzy matching across the full field string, i.e. there's no implicit tokenization or whitespace removal. This matches the semantics we currently have for fuzzy search over attributes in a non-streaming case


geirst (approved):
👍

PR #29976, opened 2024-01-19
write vsmfields.cfg and add smoke test

arnej27959:
@havardpe please review


havardpe (dismissed):
consider having BOTH as the default option to simplify re-use of old tests while also making it clear which tests only makes sense for indexed mode.

havardpe (approved):

PR #29980, opened 2024-01-19
fix semantics for empty feature lists

### LINKED ISSUE #29949: Ignore inherited `match-features`
**Is your feature request related to a problem? Please describe.**
I have a parent rank-profile that defines `match-features` but in the child rank-profile I need not to return them. What also would be great is to ignore inherited `second-phase` and `global-phase`. 

**Describe the solution you'd like**
I'd like to specify that in the child rank-profile `match-features` inherited from the parent should not be included. Or maybe just providing `match-features {}` but currently the app deployment complaints.

**Describe alternatives you've considered**
Doing `match-features: firstPhase` but in the Vespa response I still have `matchfeatures` field present.

**Additional context**
I'm experimenting with multiple ways to do the ranking. I have a base ranking profile that contains all the function definitions and `match-features` for convenient debugging. It is very close to [this sample app ](https://github.com/vespa-engine/sample-apps/blob/master/commerce-product-ranking/application/schemas/product.sd). I've tried to create a separate rank-profile that would include only `match-features` and then inherit it into the "debug" ranking profiles that would inherit the ranking profile. But double inheritance doesn't work as the function names are defined twice.

arnej27959:
* allow FeatureList to parse empty input and return empty list
* if an empty feature list block is specified in a rank-profile, trigger that we no longer get the implicit inheritance

@bratseth please review
this should fix https://github.com/vespa-engine/vespa/issues/29949


bratseth:
I think we also need to separate null from empty in ParsedRankProfile?
And we should add a test for this ...

arnej27959:
unit test added. ParsedRankProfile is OK, it will receive and send through an empty FeatureList.

PR #29981, opened 2024-01-19
Iterator from correct place

jonmv:
@baldersheim


jonmv:
This bug was introduced in https://github.com/vespa-engine/vespa/commit/cb10bc9b8d66abdfb9b8c5d92314d3b6afce9164 where `tokens` was changed from a `Deque` to a `List`.

jonmv:
Seems weird this hasn't been noticed, though ... so perhaps that code path is never run?

baldersheim (approved):
zombies are everywhere ... :)

PR #29983, opened 2024-01-19
Add a class for assist efficient traversal of dimensions in an Indexe…

baldersheim:
…dTensor.

@bratseth @jobergum PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/tensor/DirectIndexedAddress.java
+++ b/vespajlib/src/main/java/com/yahoo/tensor/DirectIndexedAddress.java
@@ -3,7 +3,7 @@ package com.yahoo.tensor;
 /**
  * Utility class for efficient access and iteration along dimensions in Indexed tensors.
  * Usage: Use setIndex to lock the indexes of the dimensions that don't change in this iteration.
- *        long base = addr.getIndex();
+ *        long base = addr.getDirectIndex();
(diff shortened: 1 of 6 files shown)


baldersheim:
Now runtime of unit test is down from around 40s to below 8s.

vespajlib/src/main/java/com/yahoo/tensor/DirectIndexedAddress.java
```diff
@@ -0,0 +1,38 @@
...
+    public void setIndex(int dimension, int index) {
+        if (index < 0 || index >= sizes.size(dimension)) {
+            throw new IndexOutOfBoundsException("Index " + index + " outside of [0," + sizes.size(dimension) + ">");
+        }
+        int diff = index - indexes[dimension];
+        directIndex += getStride(dimension) * diff;
+        indexes[dimension] = index;
+    }
+    /** Retrieve the index that can be used for direct lookup in an indexed tensor. */
+    public long getIndex() { return directIndex; }
```
bratseth, line 33: getDirectIndex?

bratseth (dismissed):
Nice!

vespajlib/src/main/java/com/yahoo/tensor/DirectIndexedAddress.java
baldersheim: Fixed

PR #29991, opened 2024-01-20
- Put the inner loops in separate methods. This improves ability to i…

baldersheim:
…nline.

- Use Buffer.get(int index) instead of Buffer.get(). That avoids a write.
- Use int as loop variable.
- This brings the splade perfoamnce test down from 8s to 7s
- TensorConverter.toVespaTensor more than doubled speed.

@bratseth @jobergum PR


bratseth (approved):
Great!

PR #29992, opened 2024-01-21
Balder/cached xxh3 hash

baldersheim:
@bratseth PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/compress/Hasher.java
+++ b/vespajlib/src/main/java/com/yahoo/compress/Hasher.java
@@ -20,7 +20,7 @@ public class Hasher {
     private Hasher(LongHashFunction hasher) {
         this.hasher = hasher;
     }
-    public static Hasher of(long seed) {
+    public static Hasher withSeed(long seed) {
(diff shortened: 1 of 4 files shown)


vespajlib/src/main/java/com/yahoo/compress/Hasher.java
```diff
@@ -8,8 +8,25 @@
...
         return LongHashFunction.xx3().hashBytes(data);
     }
+    public static long xxh3(byte [] data, long seed) {
+        return LongHashFunction.xx3(seed).hashBytes(data);
+    }
+
+    private Hasher(LongHashFunction hasher) {
+        this.hasher = hasher;
+    }
+    public static Hasher of(long seed) {
```
bratseth, line 23: "withSeed"?

bratseth (dismissed):

vespajlib/src/main/java/com/yahoo/compress/Hasher.java
baldersheim: Good advice :)

PR #29993, opened 2024-01-21
- Extract dimension names in a set to avoid recomputing it in dimensi…

baldersheim:
…onNames() mathod.

- Also extract some information about what kind of dimensions a tensor has construcion time. This avoids streamin through all diemsion later on, and keeps this local in the TensorType.

@bratseth PR


bratseth (dismissed):
👍

PR #29995, opened 2024-01-21
Revert "Balder/cached xxh3 hash"

baldersheim:
Reverts vespa-engine/vespa#29992
See if this this is causing any issues for system test runs. Killing configserver .... perhaps


hmusum:
👍

PR #29996, opened 2024-01-21
- Avoid expensive iteration with hash lookups just for sanity checkin…

baldersheim:
…g in private MixedTensor constructor.

- Refactor so there is less need for sanity checks, and do them in the buiulder instead.

@bratseth PR


bratseth (approved):
LGTM!

PR #30004, opened 2024-01-21
Allow simple queries with negative terms only

bratseth:


container-search/src/main/java/com/yahoo/prelude/query/parser/SimpleParser.java
```diff
@@ -79,6 +79,7 @@ private Item anyItemsBody(boolean topLevel) {
                         not = new NotItem();
                         not.addNegativeItem(item);
                         topLevelItem = combineItems(topLevelItem, not);
+                        System.out.println("not : " + not + " not poisitve: " + not.getPositiveItem());
```
baldersheim, line 82: Leftover debugging ... ?

baldersheim (changes requested):
Looks good, just GC the system.println

arnej27959 (approved):
👍

baldersheim (approved):

PR #30008, opened 2024-01-22
Add `noexcept` to Unicode util functions

vekterli:
@baldersheim please review. No idea why this had a virtual destructor, so removed it and made the thing `final`.


baldersheim (approved):
Good, thanks

PR #30011, opened 2024-01-22
Consider group membership when retiring host for OS upgrade

mpolden:
@bratseth


bratseth (approved):
Getting pretty complicated but I think I got it :-)

mpolden:
Yes, but not as bad as I expected. 🙂

PR #30016, opened 2024-01-22
Treat regex and fuzzy whole-field matching as 1 logical word

vekterli:
@geirst please review
@toregge FYI

We have concluded that this is the most semantically correct way of reporting the count, and as a bonus it avoids having to do a separate pass over the string buffer.


baldersheim (approved):

geirst (commented):
👍

PR #30017, opened 2024-01-22
wire in strict flow analysis and strict-aware sorting

havardpe:
strict_cost added to all blueprints
separate top-down sort step after optimize
move relative estimate out of blueprint state
optimize all children; to calculate flow stats
leaf defaults:
  matching>0.9: est: 0.5, cost: 1.0, strict_cost: 1.0
  matching<=0.9: est: rel_est, cost: 1.0, strict_cost: rel_est

@baldersheim please review
@geirst FYI

strict OR cost is currently calculated as OR flow with short-circuiting using children strict_cost, which might be underestimating the actual cost.


baldersheim (approved):
Looks very good, easy to understand code  and well tested. 🥇

PR #30020, opened 2024-01-22
Some use cases needs a lower limit

baldersheim:
@geirst PR


baldersheim:
@geirst The limit introduced in #29355 was too strict. Th original issue that was intended to remedy has since been solved at the root.

geirst (commented):
👍

PR #30022, opened 2024-01-23
Allow nested "match" update syntax (with arrays)

jonmv:
@vekterli

### DIFF
--- a/document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
+++ b/document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
@@ -31,6 +31,7 @@ import com.yahoo.document.datatypes.StringFieldValue;
 import com.yahoo.document.datatypes.Struct;
 import com.yahoo.document.datatypes.TensorFieldValue;
 import com.yahoo.document.datatypes.WeightedSet;
+import com.yahoo.document.fieldpathupdate.FieldPathUpdate;
 import com.yahoo.document.internal.GeoPosType;
 import com.yahoo.document.json.readers.DocumentParseInfo;
 import com.yahoo.document.json.readers.VespaJsonDocumentReader;
@@ -62,6 +63,7 @@ import org.junit.Test;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
```diff
@@ -410,11 +420,20 @@ public void testUpdateWeighted() throws IOException {
...
+                                                   "        'element': 'person',",
+                                                   "        'increment': 13 }}}}"));
+
+        DocumentUpdate otherDoc = parseUpdate(inputJson("{ 'update': 'id:unittest:testset::whee',",
+                                                        "  'fields': {",
+                                                        "    'actualset': {",
+                                                        "      'match': {",
+                                                        "        'increment': 13,",
+                                                        "        'element': 'person' }}}}"));
+
```
jonmv, line 435: This fails now, so don't merge :)

jonmv:
Just for discussions, so far! I'll make some actual changes, too, and I guess the question is whether preserving the current behaviour is desirable.

jonmv:
Feel free to review now, @vekterli

document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
```diff
@@ -719,15 +885,16 @@ public void testMapStringToArrayOfInt() throws IOException {
...
-        DocumentType docType = r.readDocumentType(parseInfo.documentId);
-        DocumentPut put = new DocumentPut(new Document(docType, parseInfo.documentId));
-        new VespaJsonDocumentReader(false).readPut(parseInfo.fieldsBuffer, put);
-        Document doc = put.getDocument();
+        Document doc = docFromJson("""
+                                   {
+                                     "put": "id:unittest:testMapStringToArrayOfInt::whee",
+                                     "fields": {
+                                       "actualMapStringToArrayOfInt": {
+                                         "bamse": [1, 2, 3]
```
vekterli, line 893: This previously appears to have tested a (presumed legacy) update syntax with explicit `"key": foo, "value": bar` mappings. With these changes it's identical to `testMapStringToArrayOfInt()`.

Consider either testing the explicit KV syntax (if we want to test it) or removing this test entirely (if we don't).

vekterli (dismissed):
:+1:

document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
jonmv: Ah, my mistake. I tried to verify all the AI generated JSON, but evidently I failed here.

vekterli (approved):

PR #30023, opened 2024-01-23
Update new defaults for use-per-document-throttled-delete-bucket(true…

baldersheim:
…) and merging-max-memory-usage-per-node to auto(0)

@vekterli or @hmusum PR


hmusum (approved):

vekterli (commented):
:+1:

PR #30024, opened 2024-01-23
GC control of use-per-document-delete and max-merge-memory from confi…

baldersheim:
…g production side in java.

@vekterli PR
Hold merge until 8.292 is released

### DIFF
--- a/config-model/src/test/java/com/yahoo/vespa/model/content/StorageClusterTest.java
+++ b/config-model/src/test/java/com/yahoo/vespa/model/content/StorageClusterTest.java
@@ -189,7 +189,7 @@ public class StorageClusterTest {
         var config = configFromProperties(new TestProperties());
         var limit = config.merge_throttling_memory_limit();
 
-        assertEquals(0L, limit.max_usage_bytes()); // TODO change default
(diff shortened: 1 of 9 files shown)


config-model/src/test/java/com/yahoo/vespa/model/content/StorageClusterTest.java
```diff
@@ -189,7 +189,7 @@ void merge_throttler_memory_limit_config_has_expected_defaults() {
         var config = configFromProperties(new TestProperties());
         var limit = config.merge_throttling_memory_limit();
 
-        assertEquals(-1L, limit.max_usage_bytes()); // TODO change default
+        assertEquals(0L, limit.max_usage_bytes()); // TODO change default
```
vekterli, line 192: The `TODO` can probably be removed with this change

vekterli (dismissed):
:+1:

config-model/src/test/java/com/yahoo/vespa/model/content/StorageClusterTest.java
baldersheim: Fixed

PR #30029, opened 2024-01-23
Add enum skeleton for labels

yngveaasheim:

@gjoranv , @olaaun Please take a look and see if you agree with this approach for enum for labels.


gjoranv (approved):
LGTM 👍

PR #30038, opened 2024-01-24
Jonmv/leaner token buffer

jonmv:
@vekterli please review.

I tried to reuse as much as feasible of the existing code, and without modifying it too much either. That doesn't necessarily produce the sweetest of new code, but this is what you get to review today :>

### DIFF
--- a/document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
+++ b/document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
@@ -39,7 +39,10 @@ public class LazyTokenBuffer extends TokenBuffer {
                     return null;
 
                 Token token = buffered.get();
-                if (token == null) tokens.add(token = nextToken());
+                if (token == null) {
+                    token = nextToken();
+                    tokens.add(token);
+                }
                 localNesting += nestingOffset(token.token);
                 return token;
             }
--- a/document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
+++ b/document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
@@ -1,0 +1,61 @@
+package com.yahoo.document.json;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+
+import java.io.IOException;
+import java.util.function.Supplier;
+
+/**
+ * A {@link TokenBuffer} which only buffers tokens when needed, i.e., when peeking.
+ *
+ * @author jonmv
+ */
+public class LazyTokenBuffer extends TokenBuffer {
+
+    private final JsonParser parser;
+
+    public LazyTokenBuffer(JsonParser parser) {
+        this.parser = parser;
+        try { addFromParser(parser); }
+        catch (IOException e) { throw new IllegalArgumentException("failed parsing document JSON", e); }
+        if (JsonToken.START_OBJECT != current())
+            throw new IllegalArgumentException("expected start of JSON object, but got " + current());
+        updateNesting(current());
+    }
+
+    void advance() {
+        super.advance();
+        if (tokens.isEmpty() && nesting() > 0) tokens.add(nextToken()); // Fill current token if needed and possible.
+    }
+
+    @Override
+    public Supplier<Token> lookahead() {
+        return new Supplier<>() {
+            int localNesting = nesting();
+            Supplier<Token> buffered = LazyTokenBuffer.super.lookahead();
+            @Override public Token get() {
+                if (localNesting == 0)
+                    return null;
+
+                Token token = buffered.get();
+                if (token == null) tokens.add(token = nextToken());
+                localNesting += nestingOffset(token.token);
+                return token;
+            }
+        };
+    }
+
+    private Token nextToken() {
+        try {
+            JsonToken token = parser.nextValue();
+            if (token == null)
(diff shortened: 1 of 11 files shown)


document/src/main/java/com/yahoo/document/json/readers/TensorReader.java
```diff
@@ -67,6 +67,9 @@ else if (buffer.nesting() == initNesting && JsonToken.END_OBJECT == next.token)
                 break;
             }
         }
+        if (buffer.nesting() + 1 != initNesting)
+            throw new IllegalArgumentException("incomplete JSON structure for " + tensorFieldValue);
```
jonmv, line 71: Actually, the JsonParser would detect this ...

jonmv: Reverted.

document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java
```diff
@@ -2120,69 +2120,93 @@ public void testAssignUpdateOfTensorDenseShortForm() {
...
-                        "  'operation': 'replace',",
-                        "  'cells': [",
-                        "    { 'address': { 'x': 'a', 'y': 'b' }, 'value': 2.0 } ]}"));
+                                 """
+                                 {
+                                   "operation": "replace",
+                                   "cells": [
+                                     { "address": { "x": "a", "y": "b" }, "value": 2.0 }
+                                   ]
+                                 }""");
```
jonmv, line 2129: "te*x*t" blocks, not "test" ... 🤦

document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.IOException;
+import java.util.function.Supplier;
+
+/**
+ * A {@link TokenBuffer} which only buffers tokens when needed, i.e., when peeking.
+ *
+ * @author jonmv
+ */
+public class LazyTokenBuffer extends TokenBuffer {
```
jonmv, line 14: Could the roles be reversed? Yes, certainly—the old `TokenBuffer` could indeed be replicated by simply peeking at a lazy one until it produced no more tokens. But there was so much unrelated wiring to change, so in the end, I opted for just getting the job done. I look forward to removing lots of this code, immediately when we're allowed to remove the old feeder support.

vekterli: I think ideally we'd remove the non-lazy token buffer entirely. This would also remove the need for the current  intimate (and not necessarily obvious) member variable mutations into the super-type.

vekterli (dismissed):
:+1: Ideally I'd prefer to have syntax-only changes as a separate, dedicated PR to avoid being spooked into thinking I've received a PR with 1K+ lines worth of changes to low-level buffer management and parsing code 🙈

document/src/main/java/com/yahoo/document/json/readers/TensorReader.java
```diff
@@ -196,10 +205,12 @@ else if (hasMapped && hasIndexed)
...
 
-    private static boolean isArrayOfObjects(TokenBuffer buffer, int ahead) {
-        if (buffer.peek(ahead++) != JsonToken.START_ARRAY) return false;
-        if (buffer.peek(ahead) == JsonToken.START_ARRAY) return isArrayOfObjects(buffer, ahead); // nested array
-        return buffer.peek(ahead) == JsonToken.START_OBJECT;
+    private static boolean isArrayOfObjects(TokenBuffer buffer) {
+        if (buffer.current() != JsonToken.START_ARRAY) return false;
+        Supplier<Token> lookahead = buffer.lookahead();
+        Token next;
+        while ((next = lookahead.get()).token == JsonToken.START_ARRAY);
```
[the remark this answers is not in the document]
jonmv:
```suggestion
        while ((next = lookahead.get()).token == JsonToken.START_ARRAY) { }
```

document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
```diff
@@ -0,0 +1,61 @@
...
+    public Supplier<Token> lookahead() {
+        return new Supplier<>() {
+            int localNesting = nesting();
+            Supplier<Token> buffered = LazyTokenBuffer.super.lookahead();
+            @Override public Token get() {
+                if (localNesting == 0)
+                    return null;
+
+                Token token = buffered.get();
+                if (token == null) tokens.add(token = nextToken());
```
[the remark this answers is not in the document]
jonmv:
```suggestion
                if (token == null) {
                    token = nextToken();
                    tokens.add(token);
                }
```

document/src/main/java/com/yahoo/document/json/LazyTokenBuffer.java
jonmv: Yes ... and that also requires a silly `consume()` or similar, called on the token buffer prior to reading any _other_ field from the JSON parser, after handing the parser to said buffer; it's doable, but I considered it to be even more confusing than this, where the different buffers at least have completely separate uses.

vekterli (approved):
:+1:

jonmv:
Hmm ... on second thought, it's perhaps not a good idea to suddenly forbid stray fields in the payload?

PR #30049, opened 2024-01-24
Revert "Remove unused dependency"

freva:
Reverts vespa-engine/vespa#30048


hakonhall (approved):

freva:
FYI: @hmusum several dependency enforcers started failed, don't know how to update them or whether it is safe to do, so reverting for now

PR #30076, opened 2024-01-26
Use stored entry count rather than bucket count for (dis-)allowing permanent node down edge

vekterli:
@hakonhall please review. Note that I've pruned the unit test host info JSON to only include the fields that are now actually sent from the content nodes (i.e. only the `last` metric value and no metric descriptions etc).
@hmusum FYI

The stored entry count encompasses both visible documents and tombstones. Using this count rather than bucket count avoids any issues where a node only containing empty buckets (i.e. no actual data) is prohibited from being marked as permanently down.

Entry count is cross-checked with the visible document count; if the former is zero, the latter should always be zero as well.

Since entry/doc counts were only recently introduced as part of the `HostInfo` payload, we have to handle the case where these do not exist. If entry count is not present, the decision to allow or disallow the transition falls back to the bucket count check.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


vekterli:
A cluster controller system test needs to be updated and merged at the same time as this, as it expects a different disallow-reason message

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -107,6 +110,45 @@ private static boolean noChanges(NodeState oldWantedState, NodeState newWantedSt
...
+    private static Optional<Result> checkZeroEntriesStoredOnContentNode(HostInfo hostInfo, int nodeIndex) {
+        var entriesMetric = hostInfo.getMetrics().getValueAt(ENTRIES_METRIC_NAME, DEFAULT_SPACE_METRIC_DIMENSIONS);
+        if (entriesMetric.isEmpty() || entriesMetric.get().getLast() == null) {
+            // To allow for rolling upgrades in clusters with content node versions that do not report
+            // an entry count, defer to legacy bucket count check if the entry metric can't be found.
+            return Optional.empty();
+        }
+        long lastEntries = entriesMetric.get().getLast();
+        if (lastEntries > 0) {
+            return Optional.of(disallow("The storage node stores %d document entries".formatted(lastEntries)));
```
hakonhall, line 122: It may be interesting to report back the breakdown of docs vs tombstones in this message?
An even more verbose message would also print the legacy number of buckets so it is possible to see when there is an inconsistency there.

hakonhall (dismissed):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
vekterli: Will change the returned message to be of the format `The storage node stores X documents and Y tombstones across Z buckets`

vekterli:
@hakonhall PTAL. As part of extending the failure message I've refactored how metrics are accessed by the checks. I have also made the value checks more conservative by failing if we should ever have unexpected arithmetic over-/underflows that take us into negative ranges.

hakonhall (dismissed):

hakonhall (approved):

PR #30080, opened 2024-01-26
Revert "Revert "Jonmv/reapply leaner token buffer""

jonmv:
Reverts vespa-engine/vespa#30078

@vekterli


vekterli (approved):
:+1: revert-revert branching race condition...!

jonmv:
Hehe :)

PR #30084, opened 2024-01-28
Minor cleanup

bratseth:
Note: Changes the name of some public constructs introduced after last release.


baldersheim (approved):

baldersheim:
That should not be any problem, I doubt anyone has started using those.

PR #30106, opened 2024-01-30
use greatCircleDistance to compute distance(foo).km

arnej27959:
@geirst please review
@andreer FYI


geirst (approved):
👍

andreer:
Improvement for new use cases, but a change in behaviour - something we should bring up for arch review on Thursday?

bratseth:
Isn't this just that it used to give wrong numbers at long ranges and with this it doesn't?

andreer:
Yes, we could consider this a bugfix. In theory someone could have a ranking calculation or something that depends on it but seems pretty unlikely.

PR #30107, opened 2024-01-30
GC completely unused parameters from the days of VDS

baldersheim:
@vekterli PR

PR #30108, opened 2024-01-30
Balder/gc some unused configs

baldersheim:
@vekterli PR
Rebase of #30107


storage/src/tests/common/testhelper.cpp
```diff
@@ -59,15 +59,13 @@ vdstestlib::DirConfig getStandardConfig(bool storagenode, const std::string & ro
     config = &dc.addConfig("stor-communicationmanager");
     config->set("rpcport", "0");
     config->set("mbusport", "0");
-    config = &dc.addConfig("stor-bucketdb");
     config->set("chunklevel", "0");
```
vekterli, line 62: This line can probably be removed as well, since it referred to `stor-bucketdb`

storageserver/src/tests/testhelper.cpp
```diff
@@ -41,10 +41,8 @@ vdstestlib::DirConfig getStandardConfig(bool storagenode) {
     config = &dc.addConfig("stor-communicationmanager");
     config->set("rpcport", "0");
     config->set("mbusport", "0");
-    config = &dc.addConfig("stor-bucketdb");
     config->set("chunklevel", "0");
```
vekterli, line 44: Stale `stor-bucketdb` config can be removed here as well

vekterli (dismissed):
:+1:

storage/src/tests/common/testhelper.cpp
baldersheim: Thanks,fixed

storageserver/src/tests/testhelper.cpp
baldersheim: Thanks, fixed

PR #30109, opened 2024-01-31
Revert "Balder/gc some unused configs"

aressem:
Reverts vespa-engine/vespa#30108

Broke the [GetsDuringStateTransitionsTest::test_gets_during_state_transitions] performance test.


hmusum:
I think we should just fix the test , change is good. I'll look at it

baldersheim:
Fixed in https://github.com/vespa-engine/system-test/pull/3271

aressem:
👍

PR #30116, opened 2024-01-31
- Make separate path for reduction of indexed tensors.

baldersheim:
- This brough down reduce time from 60s to 3.7s or a factor of 16x for the splade embedder if using reduce.
- A reduce based splade embedder now uses 8.0s vs 7.0s for a custom rolled out version, translated to 14% overhead. Please enter the commit message for your changes. Lines starting

@bratseth PR
@jobergum FYI

### DIFF
--- a/vespajlib/abi-spec.json
+++ b/vespajlib/abi-spec.json
@@ -728,8 +728,10 @@
       "final"
     ],
     "methods" : [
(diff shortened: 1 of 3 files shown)


bratseth (dismissed):
Nice!

PR #30118, opened 2024-01-31
- Add alternative sparsify implementation using generic tensor.reduce…

baldersheim:
…/map.

- Add options for specifying which one to use in tests and performance benchmark. Based on original implementation prior to custom reduce with the following improvements.
 - Apply Math.log after reduction which is the samp optimization as done in the custom implementation.
 - Join the 2 separate single dimension reduce statements into single 2 dimensional reduce.

@jobergum PR


jobergum (approved):
Great, more readable now!

PR #30129, opened 2024-02-01
GC void config from stor-visitor.def

baldersheim:
@vekterli PR
@geirst FYI

### DIFF
--- a/storage/src/vespa/storage/visiting/visitor.cpp
+++ b/storage/src/vespa/storage/visiting/visitor.cpp
@@ -193,7 +193,7 @@ Visitor::Visitor(StorageComponent& component)
       _trace(DEFAULT_TRACE_MEMORY_LIMIT),
       _messageHandler(nullptr),
       _messageSession(),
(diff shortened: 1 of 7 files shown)


storage/src/vespa/storage/visiting/visitor.cpp
```diff
@@ -192,6 +192,8 @@ Visitor::Visitor(StorageComponent& component)
       _hitCounter(),
       _trace(DEFAULT_TRACE_MEMORY_LIMIT),
       _messageHandler(nullptr),
+      _messageSession(),
+      _documentPriority(documentapi::Priority::PRI_NORMAL_1),
```
vekterli, line 196: Consider setting `PRI_NORMAL_3` instead as that corresponds to the default client feed priority (yes I know, it's not exactly obvious and one more reason to throw onto the "remove client operation priorities" pile 🥸).

vekterli (dismissed):
:+1:

vekterli (approved):

PR #30142, opened 2024-02-02
- Always inhibit_default_merges_when_global_merges_pending

baldersheim:
- Only show config to the code that needs it.
- Avoid using config autogenerated internals around in the code.

@vekterli PR


vekterli (approved):
:+1:

PR #30144, opened 2024-02-02
Change parent class of search::streaming::PhraseQueryNode from

toregge:
search::streaming::AndQueryNode to search::streaming::MultiTerm.

@geirst : please review


geirst (approved):
👍

PR #30157, opened 2024-02-02
Balder/restart with fast update path and metadata only fetch phase

baldersheim:
@vekterli This is the combination of #30150 and #30149.
It has fewer test failures and might be a better starting point


baldersheim:
Better to continue from #30164

PR #30164, opened 2024-02-03
- Hardcode enable_metadata_only_fetch_phase_for_inconsistent_updates …

baldersheim:
…and restart_with_fast_update_path_if_all_get_timestamps_are_consistent to true.

- The tests expecting depending on these flags specify these values explicit.

@vekterli PR

### DIFF
--- a/storage/src/tests/distributor/distributor_stripe_test.cpp
+++ b/storage/src/tests/distributor/distributor_stripe_test.cpp
@@ -144,10 +144,6 @@ struct DistributorStripeTest : Test, DistributorStripeTestUtil {
         });
     }
 
-    void configure_update_fast_path_restart_enabled() {
-        configure_stripe_with([&](auto& builder) { (void) builder; });
-    }
-
     void configure_merge_operations_disabled(bool disabled) {
         configure_stripe_with([&](auto& builder) {
             builder.mergeOperationsDisabled = disabled;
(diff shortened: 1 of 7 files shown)


storage/src/tests/distributor/distributor_stripe_test.cpp
```diff
@@ -144,10 +144,8 @@ struct DistributorStripeTest : Test, DistributorStripeTestUtil {
         });
     }
 
-    void configure_update_fast_path_restart_enabled(bool enabled) {
-        configure_stripe_with([&](auto& builder) {
-            builder.restartWithFastUpdatePathIfAllGetTimestampsAreConsistent = enabled;
-        });
+    void configure_update_fast_path_restart_enabled() {
+        configure_stripe_with([&](auto& builder) { (void) builder; });
```
vekterli, line 148: Consider adding a small comment on why this is currently a no-op, lest we forget many moons from now

storage/src/tests/distributor/distributor_stripe_test.cpp
```diff
@@ -162,10 +160,8 @@ struct DistributorStripeTest : Test, DistributorStripeTestUtil {
         });
     }
 
-    void configure_metadata_update_phase_enabled(bool enabled) {
-        configure_stripe_with([&](auto& builder) {
-            builder.enableMetadataOnlyFetchPhaseForInconsistentUpdates = enabled;
-        });
+    void configure_metadata_update_phase_enabled() {
+        configure_stripe_with([&](auto& builder) { (void) builder; });
```
vekterli, line 164: Consider a comment re: no-op here as well

vekterli (dismissed):
:+1:

storage/src/tests/distributor/distributor_stripe_test.cpp
baldersheim: Good catch, actually unused.

storage/src/tests/distributor/distributor_stripe_test.cpp
baldersheim: Unused, GCed

PR #30177, opened 2024-02-05
adjust strict OR flow

havardpe:
A simple heap-based strict OR is not able to benefit from short circuiting of the per-document evaluation. An 'optimal' strict OR would need more book-keeping and would probably be slower than the simple implementation in most normal cases.

@geirst please review
@baldersheim FYI


geirst (approved):
👍

PR #30184, opened 2024-02-05
Align flow stats calculation for leaf blueprints.

geirst:
A special case is added for attributes without a known hit estimate, instead of trying to deduce this based on the size of the (legacy) absolute hit estimate.

@havardpe please review

### DIFF
--- a/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
+++ b/searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
@@ -129,6 +129,7 @@ private:
     // as only a few ISearchContext implementations exposes the query term.
     vespalib::string _query_term;
     ISearchContext::UP _search_context;
+    attribute::HitEstimate _hit_estimate;
     enum Type {INT, FLOAT, OTHER};
     Type _type;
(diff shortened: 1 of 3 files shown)


searchlib/src/vespa/searchlib/attribute/attribute_blueprint_factory.cpp
```diff
@@ -139,6 +140,18 @@ class AttributeFieldBlueprint : public SimpleLeafBlueprint
                             QueryTermSimple::UP term, const SearchContextParams &params);
     ~AttributeFieldBlueprint() override;
 
+    search::queryeval::FlowStats calculate_flow_stats(uint32_t docid_limit) const override {
+        double rel_est = abs_to_rel_est(getState().estimate().estHits, docid_limit);
```
havardpe, line 144: no need to calculate this if we do not know, should be moved to else case. Also, no need to ask the state for the value back since we already have the value stored directly as a member.

geirst: Fixed. PTAL

havardpe (approved):

PR #30199, opened 2024-02-06
make default flow stats more explicit

havardpe:
for both simple and complex leafs
account for number of inner children in complex leafs account for seek nesting for complex leafs with children

@geirst please review


geirst (approved):
👍

PR #30211, opened 2024-02-07
benchmark non-strict iterators

havardpe:
measure ns per seek
also calculate experimental flow stats and calculate ms per cost

@geirst please review


geirst (approved):
👍

PR #30224, opened 2024-02-08
- Complete dumping of 1 index field before progressing to the next.

baldersheim:
- This prevents allocating memory buffers, and file descriptors for all fields concurrently.
- It will reduce memory footprint during flush if there are many fields.

@toregge or @geirst PR

### DIFF
--- a/searchlib/src/vespa/searchlib/index/indexbuilder.h
+++ b/searchlib/src/vespa/searchlib/index/indexbuilder.h
@@ -2,12 +2,21 @@
 #pragma once
 
 #include <vespa/vespalib/stllike/string.h>
+#include <memory>
 
 namespace search::index {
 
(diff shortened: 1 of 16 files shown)


toregge (approved):

searchlib/src/vespa/searchlib/index/indexbuilder.h
```diff
@@ -2,13 +2,22 @@
...
 #include <vespa/vespalib/stllike/string.h>
+#include <memory>
 
 namespace search::index {
 
 class DocIdAndFeatures;
 class Schema;
 class WordDocElementWordPosFeatures;
 
+class FieldIndexBuilder {
```
geirst, line 13: Please add class comment.

geirst (commented):
👍

PR #30225, opened 2024-02-09
All embedders are the same

bratseth:
This is to avoid a validation override from changed indexing expression when embedder details are changed.


jobergum (commented):
100

PR #30226, opened 2024-02-09
- Let there only be one way to wire query to the grouping hits.

baldersheim:
@bratseth PR


baldersheim:
Note that the only semantic change here is line 69 in HitConverter.java => hit.setQuery(hitContext.getQuery());

The rest is just wiring of the DocumentDatabase required for the next PR, and unification of how the query is propagated.

baldersheim:
Or @bjorncs PR

bjorncs (commented):
👍

PR #30228, opened 2024-02-09
- Ensure that the sddocname field of the hit is set.

baldersheim:
- Also test that query is as expected from the HitConverter.

@bratseth or @bjorncs PR
Continuation of #30226


bjorncs (commented):
👍

PR #30233, opened 2024-02-09
Only mark the summaryclass that you actually decode as filled.

baldersheim:
@bratseth or @geirst PR


bratseth (commented):
wat ...

baldersheim:
w ? a? t ?

PR #30234, opened 2024-02-10
- DocId is an overloaded type already. Just use string as directly fo…

baldersheim:
…r readability.

- Add noexcept and and using.

@geirst PR


geirst (approved):
👍

PR #30239, opened 2024-02-10
- Drop the optimization from long to int and just use long throughout.

baldersheim:
- That enables that numeric hashes are also handled efficiently, without resorting to strings.

@bratseth PR


bratseth (approved):
👍

PR #30240, opened 2024-02-11
- Use explicit given wanted hit count.

baldersheim:
- Do not rely on using _hits.size(). There is no contract they are the same. They are so for now so only by implementation.

@geirst PR


geirst (dismissed):
👍

PR #30243, opened 2024-02-12
account for heap cost in strict OR

havardpe:
- make it easier to do flow analysis directly on FlowStats
- use FlowStats in tests (less custom test code)
- added flow_tuning.h for special sauce
- strict flow must now have in-flow of 1.0
- split low-level flow tests into full/partial

@geirst please review


geirst (approved):
👍

PR #30244, opened 2024-02-12
- Add all hits to the hit collector.

baldersheim:
- Maintain a heap on the side, and keep heap property when producing results and features.
- Drop teh pointer to the document once it drops off the heap.

@geirst PR


geirst (approved):
👍

PR #30254, opened 2024-02-13
Increase the size of blocks malloc can pool from 8M to 16M

baldersheim:
@jobergum PR


config-model/src/main/java/com/yahoo/vespa/model/container/Container.java
```diff
@@ -107,7 +107,7 @@ protected Container(TreeConfigProducer<?> parent, String name, boolean retired,
         addChild(new SimpleComponent("com.yahoo.container.jdisc.ConfiguredApplication$ApplicationContext"));
 
         appendJvmOptions(jvmOmitStackTraceInFastThrowOption(deployState.featureFlags()));
-        addEnvironmentVariable("VESPA_MALLOC_MMAP_THRESHOLD","0x800000");
+        addEnvironmentVariable("VESPA_MALLOC_MMAP_THRESHOLD","0x1000000"); // 16G
```
jobergum, line 110: G or M?

baldersheim: M :)

PR #30273, opened 2024-02-14
Use Schemainfo instead of Indexfacts.

baldersheim:
@bratseth PR, you will get a break after this one :)


bratseth (approved):
Great!

I'll take PR's removing dependency on IndexingFacts the whole day and night ... :-)

PR #30274, opened 2024-02-14
Install config definition for lucene linguistics

hmusum:


dainiusjocas:
What does this change do? What is the effect?

hmusum:
> What does this change do? What is the effect?

It install all config definition files found in `src/main/resources/configdefinitions` (in this case `lucene-analysis.def`) so that it is available for reference (as described in https://docs.vespa.ai/en/reference/config-files.html#config-definition-files) in an installation of Vespa. All builtin config definitions in Vespa should be installed like this.

The config server / config model will parse this file and use it when validating corresponding `<config ...>` in services.xml (the config definition _can_ be supplied in the application package as well and will be used in the same way, but there are some limitations/drawbacks regarding e.g. applications compiling against old versions, error handling etc.).

baldersheim (approved):

PR #30275, opened 2024-02-14
No need to have more than 1s shutdownIdleTimeout as default. Running …

baldersheim:
…tests takes a lot longer with high timeout.

@bjorncs PR


bjorncs (approved):

bjorncs:
FYI @jonmv

PR #30277, opened 2024-02-14
Add search::predicate::SimpleIndexSaver.

toregge:
Remove extra logging when saving predicate attribute.

@geirst : please review
@baldersheim : FYI


searchlib/src/vespa/searchlib/attribute/predicate_attribute.cpp
```diff
@@ -141,11 +141,8 @@ PredicateAttribute::before_inc_generation(generation_t current_gen)
 void
 PredicateAttribute::onSave(IAttributeSaveTarget &saveTarget) {
     LOG(info, "Saving predicate attribute version %d", getVersion());
-    vespalib::string name(getBaseFileName());
```
geirst, line 144: Consider adding the attribute name to the log message above.

geirst (approved):
👍

searchlib/src/vespa/searchlib/attribute/predicate_attribute.cpp
toregge: Log message will be extended in next PR.

PR #30286, opened 2024-02-15
Specify DNS record types in Azure

freva:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Dns.java
```diff
@@ -25,12 +25,7 @@ public enum RecordType { FORWARD, PUBLIC_FORWARD, REVERSE }
 
     /** Returns the set of DNS record types for a host and its children and the given version (ipv6), host type, etc. */
     public static Set<RecordType> recordTypesFor(IP.Version ipVersion, NodeType hostType, CloudName cloudName, boolean enclave) {
-        if (cloudName == CloudName.AWS)
-            return enclave ?
-                   EnumSet.of(RecordType.FORWARD, RecordType.PUBLIC_FORWARD) :
-                   EnumSet.of(RecordType.FORWARD, RecordType.PUBLIC_FORWARD, RecordType.REVERSE);
-
-        if (cloudName == CloudName.GCP) {
+        if (cloudName == CloudName.AWS || cloudName == CloudName.GCP) {
```
hakonhall, line 28: This seems better for AWS.  There is still one discrepancy:  AmazonRoute53ClientImpl allows `#IPv4 addresses = 1` for a tenant host, and in this case the empty set should be returned here.  Not sure if this is possible.  With Andreas' new inclave setup?

In any case, until AR53CI is refactored to use this class it's not that important to resolve. It could be ignored or left as a TODO.

hakonhall (approved):

PR #30291, opened 2024-02-16
Test that predicate index saver protected by a generation guard

toregge:
observes a snapshot of the predicate index.

@geirst : please review


geirst (approved):
👍

PR #30295, opened 2024-02-16
use flow for fetchPostings

havardpe:
note that only AND/OR are wired to their respective flows here, which means only strict OR should change (all children getting full in-flow)

@geirst please review


geirst (approved):
👍

PR #30300, opened 2024-02-16
Add new Protobuf-based MessageBus DocumentAPI protocol

vekterli:
@geirst please review schema/C++ code (primary focus on `routable_factories_8.cpp`), @jonmv please review schema/Java code (primary focus on `RoutableFactories80.java` and its codec design). Note that this commit by itself will **fail** its PR-build due to missing cross-language test files. This is intentional, as these are tied to the protocol version and will be added as a separate commit.

This adds an entirely new implementation of the internal MessageBus DocumentAPI protocol, which shall be functionally 1-to-1 compatible with the existing legacy protocol.

> [!CAUTION]
> **MUST NOT** be merged before a concrete Vespa release version (in the future) has been assigned to the protocol

New protobuf schemas have been added to the top-level documentapi module, which are separated into different domains of responsibility:

 * CRUD messages
 * Visiting messages
 * Data inspection messages

As well as a schema for shared, common message types.

Both C++ and Java protocol implementations separate serialization and deserialization into a codec abstraction per message type, which hides the boilerplate required for Protobuf buffer management. The Java version is a tad more verbose due to generics type-erasure.

This protocol does _not_ currently support lazy (de-)serialization in Java, as the existing mechanisms for doing so are inherently tied to the legacy protocol version. Performance tests will decide if we need to introduce such functionality to the new protocol version.

To avoid having the new protocol go live in production, this commit changes the semantics of how MessageBus version reporting works (at least for the near future); instead of reporting the current Vespa _release_ version, it reports the highest supported _protocol_ version. This lets us conditionally enable the new protocol by reporting a MessageBus version greater than or equal to the protocol version _iff_ the protocol should be active. This is currently done with a dedicated (presumably short-lived) environment variable.

The new protocol is disabled by default.

Other changes:
 * Protocol tests have been moved up one package directory level to be aligned with the actual package of the classes they test. This allows for using package-protected constructors in the serialization tests.
 * `DocumentDeserializer` now exposes the underlying document type repo/manager. This is done to detangle `Document`/`DocumentUpdate` deserialization from the underlying wire buffer management.
 * `RemoveLocationMessage` at long last contains a bucket space, which was forgotten when we initially added this concept to the other messages, and where the pain of adding it in later was too big (not so anymore!).

Unit tests for both C++ and Java have been hoisted from the legacy test suite, cleaned up and extended with additional cases. The C++ tests use the old unit test kit and should receive a good follow-up washing and GTest-rewrite.

**Important**: due to how MessageBus protocol versioning works, the final protocol version is _not_ yet decided, as setting it requires syncing against our build systems. A follow-up commit will assign the final version as well as include all required binary test files.


documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
```diff
@@ -0,0 +1,931 @@
...
+        static <DocApiT extends Routable, ProtoT extends AbstractMessage> ProtobufCodecBuilder<DocApiT, ProtoT>
+        of(Class<DocApiT> apiClass, Class<ProtoT> protoClass) {
+            return new ProtobufCodecBuilder<>(apiClass, protoClass);
+        }
+
+        ProtobufCodecBuilder<DocApiT, ProtoT> encoder(Function<DocApiT, ProtoT> fn) {
+            if (encoderFn != null) {
+                throw new IllegalArgumentException("Encoder already set");
+            }
+            encoderFn = fn;
```
jonmv, line 97: Could consider `requireNonNull(fn)` as well, to be in line with the above check. Likewise for the `decoder(...)` below.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
```diff
@@ -0,0 +1,931 @@
...
+            return this;
+        }
+
+        ProtobufCodecBuilder<DocApiT, ProtoT> decoder(ProtoT messagePrototype, Function<ProtoT, DocApiT> fn) {
```
jonmv, line 101: Did you consider just passing the `parser()` here, instead of the `getDefaultInstance()`? It's possible this is the idiomatic way with protobuf, but that's not obvious to me. That also lets you get rid of the cast on L108. 
Similarly for `decoderWithRepo(...)`.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
```diff
@@ -0,0 +1,931 @@
...
+            if (decoderFn != null) {
+                throw new IllegalArgumentException("Decoder already set");
+            }
+            decoderFn = (buf) -> {
+                try {
+                    var protoObj = messagePrototype.getParserForType().parseFrom(buf.getBuf().getByteBuffer());
+                    return fn.apply(protoClass.cast(protoObj));
+                } catch (IOException e) {
```
jonmv, line 109: Are we guaranteed the `IOException` is never thrown?

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
jonmv: I see the nullity is guarded against in `build()`, so this is less important.

documentapi/src/test/java/com/yahoo/documentapi/messagebus/protocol/Messages80TestCase.java
```diff
@@ -0,0 +1,729 @@
...
+                var decoded = (PutDocumentMessage)deserialize("PutDocumentMessage-create", DocumentProtocol.MESSAGE_PUTDOCUMENT, lang);
+                assertTrue(decoded.getCreateIfNonExistent());
+                assertEquals(decoded.getDocumentPut(), decoded.getDocumentPut());
+            });
+        }
+
+        @Override
+        public void run() {
+            var msg = new PutDocumentMessage(new DocumentPut(new Document(protocol.getDocumentTypeManager().getDocumentType("testdoc"), "id:ns:testdoc::")));
+            msg.setTimestamp(666);
```
jonmv, line 147: 👹

documentapi/src/test/java/com/yahoo/documentapi/messagebus/protocol/Messages80TestCase.java
```diff
@@ -0,0 +1,729 @@
...
+
+    class UpdateDocumentMessageTest implements RunnableTest {
+        @Override
+        public void run() {
+            var docType = protocol.getDocumentTypeManager().getDocumentType("testdoc");
+            var update = new DocumentUpdate(docType, new DocumentId("id:ns:testdoc::"));
+            update.addFieldPathUpdate(new RemoveFieldPathUpdate(docType, "intfield", "testdoc.intfield > 0"));
+
+            var msg = new UpdateDocumentMessage(update);
+            msg.setNewTimestamp(777);
```
jonmv, line 186: 👼

messagebus/src/main/java/com/yahoo/messagebus/network/rpc/RPCNetwork.java
```diff
@@ -301,15 +300,34 @@ public boolean destroy() {
         return false;
     }
 
+    private static Version deriveSupportedProtocolVersion() {
+        // This is a very leaky abstraction, but since MessageBus only exchanges versions
+        // (and not a set of supported protocols), we have to do this workaround.
+        // Disallow-version MUST be lower than that used as a protocol lower bound in
+        // DocumentProtocol.java and the exact same as that used in C++ for the same purposes.
+        // ... Or else!
+        // TODO remove this glorious hack once protobuf protocol is enabled by default
```
jonmv, line 309: 🙈

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
jonmv: Yes, but with no trace of what went wrong?

jonmv (commented):
This is great! I just have minor comments.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
vekterli: Good point. I see there is a higher-level `decode` method which does at least print out the exception message as a warning. Will runtime-wrap and re-throw instead of just returning null

messagebus/src/main/java/com/yahoo/messagebus/network/rpc/RPCNetwork.java
vekterli: 👀 💦

jonmv: Will this usage of `protobuf` be sufficient to never need the "version negotiation" again? I'm not super-familiar with this glorious new technology. I mean, where does _it_ check protocol versions, or whatever it does.

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
jonmv: I think that's a good idea.

messagebus/src/main/java/com/yahoo/messagebus/network/rpc/RPCNetwork.java
vekterli: Also, enabling the Protobuf protocol by default would just change the reported version to always be one that is >= the version that triggers both sides to use the correct codecs. The same dance could then be done again later if we wanted to add a new protocol with a version > the Protobuf protocol version.

documentapi/src/vespa/documentapi/messagebus/routable_factories_8.h
```diff
@@ -0,0 +1,72 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include "iroutablefactory.h"
+
+namespace document { class DocumentTypeRepo; }
+
+namespace documentapi::messagebus {
+
+class RoutableFactories80 {
```
geirst, line 10: Consider adding class comment.

documentapi/src/vespa/documentapi/messagebus/routable_factories_8.cpp
```diff
@@ -0,0 +1,883 @@
...
+
+void set_global_id(protobuf::GlobalId& dest, const document::GlobalId& src) {
+    char tmp[document::GlobalId::LENGTH];
+    memcpy(tmp, src.get(), document::GlobalId::LENGTH);
+    dest.set_raw_gid(tmp, document::GlobalId::LENGTH);
+}
+
+document::GlobalId get_global_id(const protobuf::GlobalId& src) {
+    if (src.raw_gid().size() != document::GlobalId::LENGTH) {
+        throw document::DeserializeException("Unexpected serialized protobuf GlobalId size");
```
geirst, line 74: Consider adding actual and expected lengths to exception message.

geirst (dismissed):
C++ code changes 👍

documentapi/src/vespa/documentapi/messagebus/routable_factories_8.cpp
vekterli: Done

documentapi/src/vespa/documentapi/messagebus/routable_factories_8.h
vekterli: Done

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
vekterli: Have changed the code to use the underlying parser directly

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
vekterli: Now has explicit catching and logging for both encoding and decoding

documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
vekterli: Since this particular API is explicitly internal use only I'll keep this as-is for now

geirst (dismissed):
👍

jonmv:
🍾

PR #30301, opened 2024-02-16
Extend predicate index unit test to cover case where the copies of

toregge:
frozen roots in simple index saver are needed.

@geirst : please review


baldersheim (approved):

geirst (commented):
👍

PR #30318, opened 2024-02-18
Remove constexpr specifier causing inline warning when compiling with…

toregge:
… gcc 14.

@baldersheim : please review

### DIFF
--- a/eval/src/vespa/eval/eval/aggr.h
+++ b/eval/src/vespa/eval/eval/aggr.h
@@ -170,7 +170,7 @@ public:
     constexpr Median() noexcept : _seen() {}
     constexpr Median(T value) : _seen({value}) {}
     constexpr void sample(T value) { _seen.push_back(value); }
-    constexpr void merge(const Median &rhs) {
+    void merge(const Median &rhs) {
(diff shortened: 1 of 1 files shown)


eval/src/vespa/eval/eval/aggr.h
```diff
@@ -170,7 +170,7 @@ template <typename T> class Median {
     constexpr Median() noexcept : _seen() {}
     constexpr Median(T value) : _seen({value}) {}
     constexpr void sample(T value) { _seen.push_back(value); }
-    constexpr void merge(const Median &rhs) {
+    void merge(const Median &rhs) {
```
baldersheim, line 173: This feels like the wrong solution. I assume this is due to not having noinline on m_insert_aux.

baldersheim: I suggest we try to come up with another solution.

toregge: m_insert_aux is constexpr. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97681 indicates that noinline attribute might be ignored.

PR #30329, opened 2024-02-19
make flow calc mandatory (pure virtual)

havardpe:
add appropriate flow calc for various blueprints

@geirst please review


geirst (approved):
👍

PR #30358, opened 2024-02-21
Support simulation of a filter in non-strict context.

geirst:
@havardpe please review


havardpe (approved):
we might want to be careful and call FlowStats for flow_stats to separate it from the actual flow, which is more directly related to the number of seeks.

PR #30359, opened 2024-02-21
allow inputs { query(foo) string }

arnej27959:
@bratseth please review
not really sure how useful this is, the effect is the same as if the input was not declared at all.


container-search/src/main/java/com/yahoo/search/query/properties/RankProfileInputProperties.java
```diff
@@ -39,9 +41,16 @@ public RankProfileInputProperties(SchemaInfo schemaInfo, Query query, Map<String
...
-                TensorType expectedType = typeOf(name);
-                if (expectedType != null) {
-                    value = tensorConverter.convertTo(expectedType,
+                var expectedType = typeOf(name);
+                System.err.println("setting rank feature '" + name + "' -> " + value + " :: " + value.getClass());
+                if (expectedType != null && expectedType.declaredString()) {
+                    System.err.println("expected type: declared string");
+                    var e = new IllegalArgumentException("foo");
+                    e.fillInStackTrace();
+                    e.printStackTrace();
```
bratseth, line 50: Remove printlns, and throw ...

bratseth (dismissed):
... otherwise LGTM!

container-search/src/main/java/com/yahoo/search/query/properties/RankProfileInputProperties.java
arnej27959: ouch, will fix.

arnej27959:
@bratseth please take another look now
@hmusum FYI - handles SD with "long" now

bratseth (approved):
👍

bratseth:
Was there something with demanding newline breaking something that needs to be handled before merging?

hmusum:
I think we can merge this, the integration test failed with:

`Failed parsing rank-profile from 'related_articles_factual_grade.profile': Encountered " <IDENTIFIER> "long"" at line 4, column 41.

Was expecting one of:

"\n" ...
    <TENSOR_TYPE> ...
    "}" ...
    ":" ...
    "double" ...
    
At position:
       query(storyRecentSpikeStartTime) long: 0`

and handling long seems to be covered by this PR. 
If there are still issues the integration test should cover that as well.

PR #30361, opened 2024-02-21
Add embed + NN test

bratseth:


arnej27959 (approved):
👍

freva:
Breaks on factory:
```
vespa/container-search/src/test/java/com/yahoo/search/searchers/ValidateNearestNeighborTestCase.java:[85,93] incompatible types: com.yahoo.tensor.TensorType cannot be converted to com.yahoo.search.schema.RankProfile.InputType
```

PR #30373, opened 2024-02-21
Revert "demand newline between declarations of inputs"

hmusum:
Reverts vespa-engine/vespa#30355

Some prod apps fail in integration test after this change, should be fixed before merging again


geirst (approved):
👍

arnej27959:
👍 not totally unexpected, but we want to fix those and re-apply later.

PR #30375, opened 2024-02-22
Create directories in a more robust way

hmusum:
Observed an error in a system test where it looked like there was a race between checking that the directory exists and creating it


arnej27959 (approved):
👍

PR #30388, opened 2024-02-23
know the in-flow when sorting blueprints

havardpe:
also added the option and tagging mechanic needed to force selected blueprints to be strict even in a non-strict context as well as calculating how much time this saves you.


geirst (approved):
👍

PR #30398, opened 2024-02-23
Don't do a hex dump of messages into the log upon protocol deserialization failures

vekterli:
@geirst please review

I like my logs filled with pages and pages of assorted fluff as much as the next guy, but this is a bit on the excessive side.


geirst (approved):
👍

PR #30401, opened 2024-02-26
Revert "Use class names instead of strings"

hmusum:
Reverts vespa-engine/vespa#30357

Broke ~at least~ 1 system test, need to look into it


gjoranv:
👍

PR #30402, opened 2024-02-26
Explicitly shut down Protobuf library prior to distributor process exit

vekterli:
@geirst please review

This avoids false positives from Valgrind et al about memory leaks.

The distributor is most susceptible to this false positive since error reporting in the DocumentAPI protocol may cause Protobuf descriptors to be lazily loaded when an incoming message fails to be decoded (usually due to a missing document type).


geirst (approved):
👍

PR #30408, opened 2024-02-26
More graceful reporting of Protobuf protocol decode errors

vekterli:
@geirst please review. This should address the protocol-related system test failure (which fails due to errors in the Vespa log).

If a message fails decoding due to e.g. a missing document type during document decode, that will be reported as a protocol-level decode error, even if that message itself is perfectly compliant. Report such exceptions as warnings instead (by propagating an exception up to the `DocumentProcotol` caller), but add more context to the exception message to make it clear that it happened in a decode-context.

Example of old (and rather confusing) message in logs:
```
WARNING distributor  vds.documentprotocol  Document type foo not found
```
Now becomes:
```
WARNING distributor  vds.documentprotocol Failed decoding message of type PutDocumentRequest: Document type foo not found
```


geirst (approved):
👍

PR #30414, opened 2024-02-27
Handle `InvalidJsonException` by default. Fix mapper ordering

bjorncs:

### DIFF
--- a/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
+++ b/container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
@@ -211,7 +211,11 @@ class RestApiImpl implements RestApi {
             exceptionMappers.addAll(RestApiMappers.DEFAULT_EXCEPTION_MAPPERS);
         }
         // Topologically sort children before superclasses, so most the specific match is found by iterating through mappers in order.
-        exceptionMappers.sort((a, b) -> (a.type.isAssignableFrom(b.type) ? 1 : 0) + (b.type.isAssignableFrom(a.type) ? -1 : 0));
+        exceptionMappers.sort((l, r) -> {
+            if (l.type.equals(r.type)) return 0;
+            if (l.type.isAssignableFrom(r.type)) return 1;
+            return -1;
+        });
         return exceptionMappers;
     }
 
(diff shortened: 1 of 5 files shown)


freva (dismissed):

bjorncs:
PTAL

container-core/src/main/java/com/yahoo/restapi/RestApiImpl.java
```diff
@@ -211,7 +211,13 @@ private static List<ExceptionMapperHolder<?>> combineWithDefaultExceptionMappers
             exceptionMappers.addAll(RestApiMappers.DEFAULT_EXCEPTION_MAPPERS);
         }
         // Topologically sort children before superclasses, so most the specific match is found by iterating through mappers in order.
-        exceptionMappers.sort((a, b) -> (a.type.isAssignableFrom(b.type) ? 1 : 0) + (b.type.isAssignableFrom(a.type) ? -1 : 0));
+        exceptionMappers.sort((l, r) -> {
+            if (l.type.equals(r.type)) return 0;
+            if (l.type.isAssignableFrom(r.type)) return 1;
+            if (r.type.isAssignableFrom(l.type)) return -1;
+            // Ensure global ordering when no inheritance relationship exists
```
jonmv, line 218: Why this over stable sort order, which lets users define desired order?

jonmv: Hmm, actually, why should it matter at all, without multiple inheritance?

bjorncs: I got incorrect ordering with the previous implementation, most likely to the sort algorithm assuming transitive equality from comparison of disconnected types.

bjorncs: So the intention is to always select the most specific exception mapper for a given exception instance, e.g selecting the mapper for `RestApiException` over `RuntimeException` for an exception of the former type.

jonmv: If types `A` and `B` are siblings, then only one of them can be assignable from instance `e`, not both, and so the ordering of `A` and `B` doesn't matter?
Do you have an example?

bjorncs: I have deleted the sample code sadly. You can recreate through a simple unit test by making a `List<Exception>` with various exception types and sorting with a custom comparator.

jonmv: Aha, never mind, I see what you're getting at. This doesn't define a total order, and so the sorting fails. Right, but I'm surprised if sorting disconnected children by some _other_ unrelated comparator helps at all.

jonmv:
Great that you caught this bug!

jonmv (approved):

PR #30430, opened 2024-02-28
Add test showing that load balancer might be removed erroneously for …

hmusum:
…combined cluster

Looks like load balancer refer to container cluster,  whereas LoadBalancerExpirer should check for the actual activated nodes used, which are the nodes in the content cluster for combined cluster.


mpolden (approved):
👍

PR #30435, opened 2024-02-29
Ensure that combined load balancer is not expired too early

mpolden:
Combined - the gift that keeps on giving.

(Parameterization gave me some initialization errors I didn't want to
spend time on)

@hmusum


hmusum (approved):
Great!

PR #30454, opened 2024-03-01
Set fixed vespa-build-dependencies version

aressem:

Because we have fixed runtime deps like vespa-onnxruntime we must use exact version for vespa-build-dependencies. If not we risk getting a newer version in the develop/build environment compared to the runtime environment.

FYI, @toregge


baldersheim (approved):

baldersheim:
I think that we are a bit too strict on the onnxruntime and a few others. I do not think wee need exact version.
That would also make upgrades a lot simpler.

PR #30468, opened 2024-03-04
- GC unused ancient code. documents selection validation is

baldersheim:
done correctly in ContentCluster. GC old leftower from 5.0.

@vekterli or @geirst PR


vekterli (approved):
🦖👍

PR #30472, opened 2024-03-04
Simplify MultipleDocumentDatabasesConfigProducer according to doc.

baldersheim:
@geirst PR

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/search/IndexedSearchCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/search/IndexedSearchCluster.java
@@ -45,7 +45,6 @@ public class IndexedSearchCluster extends SearchCluster
     private SearchCoverage searchCoverage;
 
     private final List<DocumentDatabase> documentDbs = new LinkedList<>();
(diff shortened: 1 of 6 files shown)


config-model/src/main/java/com/yahoo/vespa/model/search/IndexedSearchCluster.java
```diff
@@ -216,59 +214,42 @@ public String toString() {
...
+                    docDb.getConfig(builder);
+                }
             }
-        }
 
-        @Override
-        public void getConfig(SchemaInfoConfig.Builder builder) {
-            for (DocumentDatabase docDb : docDbs) {
-                docDb.getConfig(builder);
+            public void getConfig(SchemaInfoConfig.Builder builder) {
```
geirst, line 224: Something wrong with the indentation? Applies for the functions below as well.

geirst (dismissed):
👍

config-model/src/main/java/com/yahoo/vespa/model/search/IndexedSearchCluster.java
baldersheim: Good point, I wonder if this happened the last time intellij converted to a record too.

PR #30474, opened 2024-03-04
avoid re-using attributeField for sorting, would overwrite

arnej27959:
@baldersheim please review

### DIFF
--- a/streamingvisitors/src/vespa/searchvisitor/searchvisitor.cpp
+++ b/streamingvisitors/src/vespa/searchvisitor/searchvisitor.cpp
@@ -966,6 +966,15 @@ void SearchVisitor::setupAttributeVector(const FieldPath &fieldPath) {
     }
 }
 
+namespace {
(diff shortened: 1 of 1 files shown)


baldersheim (dismissed):

streamingvisitors/src/vespa/searchvisitor/searchvisitor.cpp
```diff
@@ -966,6 +966,15 @@ void SearchVisitor::setupAttributeVector(const FieldPath &fieldPath) {
     }
 }
 
+namespace {
+bool notContained(const std::vector<size_t> sortList, size_t idx) {
```
geirst, line 970: Did you mean to pass by value (and not by reference)?

arnej27959: good catch, thanks

geirst (approved):
👍

baldersheim (approved):

PR #30485, opened 2024-03-05
Add lucene-linguistics to platform bundles

hmusum:
Allow customers to specify LuceneLinguistics as a component without specifying bundle (so that Vespa's provided lucene-linguistics bundle will be used)


hmusum:
@gjoranv please take a look

gjoranv (approved):
Sorry for late review

PR #30487, opened 2024-03-05
Revert "Balder/let document database provide vsm config too"

baldersheim:
Reverts vespa-engine/vespa#30477
@geirst PR Broke system tests


geirst (commented):
👍

PR #30494, opened 2024-03-05
Enforce document timestamp requirements for updates in backend

vekterli:
@geirst please review

The document API has long since had a special field for update operations where an optional expected _existing_ backend timestamp can be specified, and where the update should only go through iff there is a timestamp match.

This has been supported on the distributor all along, but only when write-repair is taking place (i.e. rarely), but the actual backend support has been lacking. No one has complained yet since this is very much not an advertised feature, but if we want to e.g. use this feature for improvements to batch updates we should ensure that it works as expected.

With this commit, a non-zero "old timestamp" field is cross-checked against the existing document, and the update is only applied if the actual and expected timestamps match.

Note: this is a replica-level decision, but  if replicas are out of sync this shall be transparently (and equivalently) handled on the distributor as part of write-repair.

### DIFF
--- a/storage/src/tests/persistence/testandsettest.cpp
+++ b/storage/src/tests/persistence/testandsettest.cpp
@@ -65,7 +65,7 @@ struct TestAndSetTest : PersistenceTestUtils {
         PersistenceTestUtils::TearDown();
     }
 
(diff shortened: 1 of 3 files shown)


geirst (approved):
👍

PR #30499, opened 2024-03-06
- Set the complete list of the config where it contains the complete …

baldersheim:
…world.

- Append where it is intended. This is a workaround for not having a map type in config. Sometimes it works to have multiple producers produce the same config, sometimes not. This tries to make it more deterministic. Currently other tricks have been employed to avoid appending more to the list than intended.

@geirst PR

### DIFF
--- a/container-core/src/main/resources/configdefinitions/container.qr-searchers.def
+++ b/container-core/src/main/resources/configdefinitions/container.qr-searchers.def
@@ -64,7 +64,7 @@ searchcluster[].name string
 searchcluster[].searchdef[] string
 
 ## configid that may be used to get rank-profiles config for the cluster.
-searchcluster[].rankprofiles.configid reference default=""
+searchcluster[].rankprofiles_configid string default=""
 
 ## Indexing mode of search cluster.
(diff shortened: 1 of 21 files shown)


container-core/src/main/resources/configdefinitions/container.qr-searchers.def
```diff
@@ -64,7 +64,7 @@ searchcluster[].name string
 searchcluster[].searchdef[] string
 
 ## configid that may be used to get rank-profiles config for the cluster.
-searchcluster[].rankprofiles.configid reference default=""
+searchcluster[].rankprofiles_configid string default=""
```
geirst, line 67: Any side effects of this change, or is `reference` and `string` more or like the same?

geirst (approved):
👍

container-core/src/main/resources/configdefinitions/container.qr-searchers.def
baldersheim: Should be the same. Reference was pre-historic. Both @hmusum and @gjoranv recommended replacement.

PR #30511, opened 2024-03-07
Get the schema from the query

baldersheim:
@geirst or @hmusum PR


container-search/src/main/java/com/yahoo/vespa/streamingvisitors/StreamingSearcher.java
```diff
@@ -139,12 +135,15 @@ public Result doSearch2(Query query, Execution execution) {
             return new Result(query, ErrorMessage.createIllegalQuery("Streaming search requires either " +
                                                                      "streaming.groupname or streaming.selection"));
         }
+        // Cluster searcher guartees that there will one, and only one schema here
```
hmusum, line 138:
```suggestion
        // Cluster searcher guarantees that there will be one, and only one schema here
```

hmusum (dismissed):

container-search/src/main/java/com/yahoo/vespa/streamingvisitors/StreamingSearcher.java
baldersheim: Thanks

PR #30526, opened 2024-03-08
Add server-sent events (SSE) renderer

lesters:
@bratseth Please review
@hmusum FYI

`EventStream` renderer here is for rendering "events" that are added asynchronously to the hitgroup, and rendered as they occur by flushing to the output stream. Currently, this is primarily for rendering tokens as they arrive asynchronously from a language model, but could in principle be extended to be any form of event.

This renderer is enabled by `format=sse` in the query string, and requires a `EventStream` HitGroup (or DataList). The normal JSON renderer is able to render the `EventStream`, but would have to wait until all events are produced before returning a response.

Rendering traces and the underlying result set from a query is not yet implemented here (commented in the code) as it was not quite trivial to call out to the `JsonRenderer` for parts of the output.

### DIFF
--- a/container-core/src/main/java/com/yahoo/processing/response/DataList.java
+++ b/container-core/src/main/java/com/yahoo/processing/response/DataList.java
@@ -80,7 +80,7 @@ public interface DataList<DATATYPE extends Data> extends Data {
     void addDataListener(Runnable runnable);
 
     /**
-     * Notify this list that is will never be accessed again, neither for read nor write.
+     * Notify this list that it will never be accessed again, neither for read nor write.
      * Implementations can override this as an optimization to release any data held in the list
      * for garbage collection.
      *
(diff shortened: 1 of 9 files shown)


bratseth:
- Du we need a new renderer - couldn't this capability just be added to AsynchronousSectionedRenderer?
- Should EventStream extend AbstractDataList instead of HitGroup since it isn't a list of hits?

bratseth:
Regarding the renderer:
 I had forgotten about this format - it's fine!
I think it would be nice to have default handling of this in the AsynchronousSectionedRenderer, so that we could have hybrid Results, but then it would need to be rendered as some kind of JSON of course.

Regarding Hits or not: 
You could have a Result with a single GeneratedHit which is a DataList?

lesters:
@bratseth Updated the `EventStream` to be a `DataList` - with some changes to `JsonRenderer` to allow for rendering of this.

bratseth (approved):
Great!

PR #30553, opened 2024-03-09
Stem prefix items

bratseth:
If we are searching a stemmed index, it's probably better to stem terms also when we are searching for prefixes.


bratseth:
To quickly see the point of this scroll down to the bottom of the changes.

arnej27959 (approved):
I'm still not sure if this is helpful.  Searching in a (stemmed or not) index with a prefix item is probably a mistake, since indexes does not currently support prefix search.  Maybe it should be flagged with an error message by backend in that case?

bratseth:
Yes, you're right. I want to add support for prefix search in indexes though, and maybe also stemmed attributes, but we can let this wait for that.

esolitos:
@bratseth @arnej27959 What should be the conclusion of this, a year later? 😃

PR #30564, opened 2024-03-11
Balder/prepare for backend per schema

baldersheim:
@geirst PR


container-search/src/test/java/com/yahoo/prelude/test/IntegrationTestCase.java
```diff
@@ -1,174 +0,0 @@
...
-import org.junit.jupiter.api.Test;
-
-/**
- * Runs a query thru the configured search chain from a real http channel
- * to a mock fdispatch channel. The setup is rather complicated, as the point is
- * to span as much of the processing from query to result as possible.
- *
- * @author bratseth
- */
-public class IntegrationTestCase {
```
geirst, line 17: Were the changes in this files intentional?

geirst (approved):
👍  But look at the changes to IntegrationTestCase first.

container-search/src/test/java/com/yahoo/prelude/test/IntegrationTestCase.java
baldersheim: Intention was to remove test as everything was already commented out, and the comments and the code indicated it had been dead before it was born.

PR #30613, opened 2024-03-13
Balder/cleanup backend construction

baldersheim:
@hmusum PR
Hopefully makes it easier to reason about when things are valid and not by doing all setup in constructors.


hmusum (approved):
Great!

PR #30636, opened 2024-03-14
Allow prometheus format from /state/v1/metrics

olaaun:
FYI @yngveaasheim

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
@@ -31,6 +31,7 @@ import java.io.PrintStream;
 import java.net.URI;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
@@ -38,6 +39,7 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import static com.yahoo.container.jdisc.state.JsonUtil.sanitizeDouble;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * A handler which returns state (health) information from this container instance: Status, metrics and vespa version.
@@ -121,7 +123,7 @@ public class StateHandler extends AbstractRequestHandler implements CapabilityRe
     }
 
     private String resolveContentType(URI requestUri) {
-        if (resolvePath(requestUri).equals(HISTOGRAMS_PATH)) {
+        if (resolvePath(requestUri).equals(HISTOGRAMS_PATH) || isPrometheusRequest(requestUri.getQuery())) {
             return "text/plain; charset=utf-8";
         } else {
             return "application/json";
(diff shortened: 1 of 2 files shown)


container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
```diff
@@ -212,6 +216,56 @@ private ObjectNode buildJsonForConsumer(String consumer) {
...
+                .append("{").append(dimension).append("}")
+                .append(" ").append(sanitizeIfDouble(value)).append(" ")
+                .append(timeStamp).append("\n");
+    }
+
+    private String toPrometheusDimensions(MetricDimensions dimensions) {
+        if (dimensions == null) return "";
+        StringBuilder builder = new StringBuilder();
+        dimensions.forEach(entry -> {
+            var sanitized = prometheusSanitizedName(entry.getKey()) + "=\"" + entry.getValue() + "\",";
```
vekterli, line 263: Unless `entry.getValue()` is pre-sanitized we might want to add some escaping here to ensure our label values are always edible:
> `label_value` can be any sequence of UTF-8 characters, but the backslash (`\`), double-quote (`"`), and line feed (`\n`) characters have to be escaped as `\\`, `\"`, and `\n`, respectively

container-core/src/main/java/com/yahoo/container/jdisc/state/StateHandler.java
```diff
@@ -212,6 +216,56 @@ private ObjectNode buildJsonForConsumer(String consumer) {
...
 
+    private byte[] buildPrometheusForConsumer(String consumer) {
+        var snapshot = getSnapshot();
+        if (snapshot == null)
+            return new byte[0];
+
+        var timestamp = snapshot.getToTime(TimeUnit.MILLISECONDS);
+        var builder = new StringBuilder();
+        builder.append("# NOTE: THIS API IS NOT INTENDED FOR PUBLIC USE\n");
+        for (var tuple : collapseMetrics(snapshot, consumer)) {
```
vekterli, line 227: If my understanding of `collapseMetrics` _and_ the [Prometheus exposition docs](https://prometheus.io/docs/instrumenting/exposition_formats/#grouping-and-sorting) is correct, I think we need to add an extra aggregation step that groups all samples for the same metric _name_ under the same `HELP`/`TYPE` umbrella:

> All lines for a given metric must be provided as one single group, with the optional `HELP` and `TYPE` lines first (in no particular order). Beyond that, reproducible sorting in repeated expositions is preferred but not required, i.e. do not sort if the computational cost is prohibitive.
>
> Each line must have a unique combination of a metric name and labels. Otherwise, the ingestion behavior is undefined.

I.e. if we have two distinct samples (different dimensions) for a gauge `foo`, the different time series this metric results in need to be grouped together in the output:
```
...
# TYPE foo_sum untyped
foo_sum{bar="baz"} 123
foo_sum{bar="zoid"} 456
# TYPE foo_count untyped
foo_count{bar="baz"} 2
foo_count{bar="zoid"} 4
...
```

vekterli (approved):
:+1: some questions/comments for consideration, in particular regarding if we need to explicitly group emitted time series prior to output

PR #30647, opened 2024-03-15
Use filter settings from rank profiles and query terms in streaming s…

toregge:
…earch.

@baldersheim : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/query/streaming/dot_product_term.cpp
+++ b/searchlib/src/vespa/searchlib/query/streaming/dot_product_term.cpp
@@ -51,8 +51,9 @@ DotProductTerm::unpack_scores(Scores& scores, std::optional<double> score_thresh
 }
 
 void
-DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data)
+DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
 {
+    (void) index_env;
     Scores scores;
     build_scores(scores);
     unpack_scores(scores, std::nullopt, docid, td, match_data);
--- a/searchlib/src/vespa/searchlib/query/streaming/dot_product_term.cpp
+++ b/searchlib/src/vespa/searchlib/query/streaming/dot_product_term.cpp
@@ -51,9 +51,8 @@ DotProductTerm::unpack_scores(Scores& scores, std::optional<double> score_thresh
 }
 
 void
-DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
+DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment&)
 {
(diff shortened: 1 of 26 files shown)


searchlib/src/vespa/searchlib/query/streaming/weighted_set_term.cpp
```diff
@@ -18,8 +18,9 @@ WeightedSetTerm::WeightedSetTerm(std::unique_ptr<QueryNodeResultBase> result_bas
 WeightedSetTerm::~WeightedSetTerm() = default;
 
 void
-WeightedSetTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data)
+WeightedSetTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
 {
+    (void) index_env;
```
baldersheim, line 23: When the type is specific you do not need a name for the parameter, and then you do not need the explicit (void).

searchlib/src/vespa/searchlib/query/streaming/wand_term.cpp
```diff
@@ -34,8 +34,9 @@ WandTerm::evaluate() const
 }
 
 void
-WandTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data)
+WandTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
 {
+    (void) index_env;
```
baldersheim, line 39: When the type is specific you do not need a name for the parameter, and then you do not need the explicit (void).

searchlib/src/vespa/searchlib/query/streaming/in_term.cpp
```diff
@@ -21,8 +21,9 @@ InTerm::InTerm(std::unique_ptr<QueryNodeResultBase> result_base, const string &
 InTerm::~InTerm() = default;
 
 void
-InTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data)
+InTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
 {
+    (void) index_env;
```
baldersheim, line 26: When the type is specific you do not need a name for the parameter, and then you do not need the explicit (void).

searchlib/src/vespa/searchlib/query/streaming/dot_product_term.cpp
```diff
@@ -51,8 +51,9 @@ DotProductTerm::unpack_scores(Scores& scores, std::optional<double> score_thresh
 }
 
 void
-DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data)
+DotProductTerm::unpack_match_data(uint32_t docid, const ITermData& td, MatchData& match_data, const fef::IIndexEnvironment& index_env)
 {
+    (void) index_env;
```
baldersheim, line 56: When the type is specific you do not need a name for the parameter, and then you do not need the explicit (void).

baldersheim (dismissed):

baldersheim (approved):

PR #30648, opened 2024-03-15
CLI: Add query header option

mpolden:
I wrote `--header=K=V` in the review, but decided on `--header="K: V"` for
symmetry with `curl`.

@lesters


lesters (approved):
👍

PR #30649, opened 2024-03-15
Remove scaling suggestion replaced by list

olaaun:

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/ApplicationSerializer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/ApplicationSerializer.java
@@ -56,6 +56,7 @@ public class ApplicationSerializer {
     private static final String maxResourcesKey = "max";
     private static final String groupSizeKey = "groupSize";
     private static final String requiredKey = "required";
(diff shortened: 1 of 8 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/persistence/ApplicationSerializer.java
```diff
@@ -141,8 +140,6 @@ private static void toSlime(Cluster cluster, Cursor clusterObject) {
         toSlime(cluster.maxResources(), clusterObject.setObject(maxResourcesKey));
         toSlime(cluster.groupSize(), clusterObject.setObject(groupSizeKey));
         clusterObject.setBool(requiredKey, cluster.required());
-        // TODO(olaa): Remove 'suggested' once API clients migrate to suggestion list
-        toSlime(cluster.suggested(), clusterObject.setObject(suggestedKey));
```
mpolden, line 145: Configservers upgrade one-by-one, so you have to keep writing the field until there are no readers left. Stop reading the field first. Once that change has rolled out you can stop writing it and remove it entirely.

mpolden (changes requested):

mpolden (approved):

PR #30657, opened 2024-03-15
Attempt of supporting mapping array to mapped 2d tensor for sparse encoders

jobergum:
Still trying to figure out this is the prettiest way.. 

The intention is to support the following for the upcoming `splade-embedder` that maps a single-valued string to a mapped sparse tensor. 

```
field sparse_embeddings type tensor<float>(chunk{), token{}) {
  indexing: input array_string_chunks | embed splade chunk | attribute | summary
}
```


bratseth (approved):
Looks good to me!

PR #30679, opened 2024-03-19
Support internal metric rendering in Prometheus text format in C++

vekterli:
@baldersheim please review
@olaaun @yngveaasheim FYI

Maps all internal metrics to one or more labelled time series. Due to poor compatibility between the data model (and sampling strategy) of the legacy metrics framework and that of Prometheus, all time series are emitted as `untyped` metrics.

This is a stop-gap solution on the way to "properly" supporting Prometheus exposition (which would require more fundamental changes to our internal metric libraries), and the output of this renderer should therefore only be used for internal purposes.

This PR adds wiring to the storage-internal status page rendering of metrics—state V1 is not yet wired.


metrics/src/vespa/metrics/prometheus_writer.cpp
```diff
@@ -0,0 +1,287 @@
...
+    if (set.getOwner()) {
+        assert(!_path.empty());
+        _path.pop_back();
+    }
+}
+
+bool PrometheusWriter::visitCountMetric(const AbstractCountMetric& m, bool) {
+    auto full_path = metric_to_path_ref(m.getName());
+    auto labels    = as_prometheus_labels(m);
+    _samples.emplace_back(TimeSeriesSample{full_path, "count", labels, {m.getLongValue("count")}});
```
baldersheim, line 209: Is the TimeSeriesSample needed.
Are there ambiguities preventing just emplace_back(full_path, "count", labels, {m.getLongValue("count")}) ?

baldersheim (dismissed):

metrics/src/vespa/metrics/prometheus_writer.cpp
vekterli: This was done due to avoid type deduction problems when `emplace_back`-forwarding a nested `{}` initializer of the `std::variant` argument, but in retrospect this isn't needed at all since just passing the value itself will implicitly construct the variant with the correct value. Will simplify.

PR #30697, opened 2024-03-21
new publiccd endpoint

tokle:


flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
```diff
@@ -431,6 +431,13 @@ public class PermanentFlags {
             CLOUD_ACCOUNT
     );
 
+    public static final UnboundStringFlag REFRESH_IDENTITY_AFTER = defineStringFlag(
```
mpolden, line 434: Probably doesn't belong here?

tokle: correct ...

esolitos:
We'll re-handle this when the time comes.

PR #30716, opened 2024-03-22
Revert "fold AND and SAND items into top-level WEAKAND"

arnej27959:
Reverts vespa-engine/vespa#30707

this broke various system tests; needs it some more exploration and probably a feature flag.

@bratseth @baldersheim @jobergum FYI


geirst (approved):
👍

PR #30717, opened 2024-03-22
Balder/cluster.schema map to.cluster

baldersheim:
@bratseth PR


baldersheim:
@geirst FYI

baldersheim:
I will let factory have a pass first

bratseth (commented):
LGTM, but I think we need a test of the usage scenario this addresses.

baldersheim:
I will add unit tests after Easter. There are system tests.

PR #30755, opened 2024-03-29
Jobergum/add support for binarization and matryoshka

jobergum:
This adds support for binarization to the hf embedder (packing 1024 floats to 128 int8s) after getting the first k floats from an embedding model.  The binarization/quantization is triggered by asking for int8 precision. Similar to the colbert-embedder.

For example for a model that outputs 384 floats, we can binarize the full output into 48 int8:

```
field embedding type tensor<int8>(x[48]) {
   indexing: input text | embed my-embedder | index
   attribute {
     distance-metric: hamming
   }
}
```

Or we can just use the 24*8 first values from the embedding model

```
field embedding type tensor<int8>(x[24]) {
   indexing: input text | embed my-embedder | index
   attribute {
     distance-metric: hamming
   }
}
```

At query time, we can get both representations by invoking the model twice and then have re-scoring with the float,unpacked version in ranking. 


jobergum:
System test in https://github.com/vespa-engine/system-test/pull/3658

jobergum:
Thoughts on this @bratseth ?

bratseth (approved):
Lovely ... 💯

PR #30773, opened 2024-04-02
Lesters/update platform bundles for rag 2

lesters:
@bratseth Please review
@hmusum FYI

To fix reason for: https://github.com/vespa-engine/vespa/pull/30753

The package `ai.vespa.llm` existed in both `vespajlib` and `container-search`, so the interface `LanguageModel` was not properly exported since it was the one in `vespajlib`. I've renamed the package in `vespajlib` to `ai.vespa.languagemodels`. It is marked as Beta so I assume this is ok. Since it is the packages in `container-search` that we expose to the non-Java users, I thought it should probably have the shortest name.

Would welcome any other solutions.


bratseth:
Isn't the logical structure

ai.vespa.llm in vespajlib, and
ai.vespa.search.llm in container.search?

lesters:
Yes, except the LLM clients could be used in doc procs as well, so `ai.vespa.search.llm` isn't quite logical either. But actually, when looking closer at this, there are no classes in `ai.vespa.llm` in `container-search`, they are in `ai.vespa.llm.search` or `ai.vespa.llm.clients`.

bratseth:
Docproc shouldn't really depend on classes in search (Except to do search).
So, I think we should put the configurable OpenAIClient ("OpenAI") in some shared module - linguistics? 
I also suggest renaming it to ConfigurableOpenAIClient. This package could be e.g ai.vespa.llm.component?
The stuff in container-search should be the searcher and any RAG-specific functionality, and I think nested under ai.vespa.search rather than the other way around.

bratseth:
Yes, I think that makes sense? We can do it in another PR though.

lesters:
So, moved classes in `ai.vespa.llm.search` to `ai.vespa.search.llm`. 

I have not touched the clients, I think they make sense in `ai.vespa.llm.clients`. But they are still in container-search. As they can be used for instance in a doc proc, I agree they should probably find another home. But I don't think linguistics make sense though. Maybe container-core ? But we can move that in another PR.

lesters:
Ok to merge?

hmusum:
Looks good to me, I suggest merging it now

bratseth:
> But I don't think linguistics make sense though. Maybe container-core ?

container-core would work too but this isn't really a core thing though?
Why don't you think linguistics makes sense?

bratseth:
... or model-integration would also make sense.

PR #30780, opened 2024-04-02
Re-enable async LLM test

lesters:
@hmusum Please review.

In https://github.com/vespa-engine/vespa/pull/30779 I added a stealth change: upping fixed thread pool count to 2. When testing this at 1, locally on my computer I had maybe 10 tests failing out of 10000. With 2, 0 out of 10000. This should fix the instability.


hmusum (approved):
Great!

PR #30783, opened 2024-04-02
Disable proxy protocol on jdisc containers in Azure

hakonhall:


bjorncs (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/Zone.java
```diff
@@ -54,7 +54,7 @@ public Zone(Cloud cloud, SystemName systemName, Environment environment, RegionN
         this.region = region;
     }
 
-    // TODO(mpolden): For compatibility with older config models. Remove when versions < 8.76 are gone
+    // TODO(mpolden): For compatibility with older config models. Remove when versions < 8.327 are gone
```
bjorncs, line 57: Consider adding a `@Deprecated(forRemoval = true)`

hakonhall: Will do in a follow-up PR.

PR #30800, opened 2024-04-03
Add SimpleTokenScript to SimpleTokenizer

MariusArhaug:
FYI @bjorncs


linguistics/src/main/java/com/yahoo/language/simple/SimpleTokenizer.java
```diff
@@ -95,6 +103,11 @@ private TokenType determineType(TokenType tokenType, TokenType characterType) {
         return tokenType;
     }
```
arnej27959, line 105: what is the intent here? something like
// if any character is LATIN, use that as token script; otherwise use script of first character seen.

but we may consider using first "real" script seen also, ignoring COMMON and INHERITED if possible

arnej27959 (dismissed):
one minor comment, otherwise looks good to me

arnej27959 (approved):
👍

Issue #27967, opened 2023-08-04
Add embedding cache for query context inference to hf-embedder


jobergum:
For a model's lifetime (active container), the output tensor will not change for the same input and could be cached. 

This is worthwhile, given the significant amount of computing for text-embedding models. 

```
 <component id="e5-small-q" type="hugging-face-embedder">
            <transformer-model path="model/e5-small-v2-int8.onnx"/>
            <tokenizer-model path="model/tokenizer.json"/>
            <cache>128</cache>
</component>
```
No need for TTL/expiry since the model will not change for the lifecycle of the embedder instance.

bjorncs:
This is another use case for adopting Triton Inference Server as ONNX runtime in the container. Triton provides a built-in [response cache](https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/user_guide/response_cache.html).

PR #30809, opened 2024-04-04
Add onnx output caching to embedder  (allow different post-processing of model outputs)

jobergum:
After https://github.com/vespa-engine/vespa/pull/30769 & https://github.com/vespa-engine/vespa/pull/30766 we can cache the output of a model inference and perform different post-processing depending on the tensor cell type and dimensionality. 

The primary use cases are:

- Dimension flexibility 🪆 using only the first k vector dimensions 
- Dimension precision flexibility - using binary quantization packing k floats into k/8 int8  
- Future native embedder implementations supporting models that emits both sparse, dense and mixed representations, e.g [M3](https://pyvespa.readthedocs.io/en/latest/examples/mother-of-all-embedding-models-cloud.html) 

```
field embedding type tensor<int8>(x[16]) {
  input | text | embed my-embedder-id | attribute
}

field embedding type tensor<float>(x[128]) {
  input | text | embed my-embedder-id | attribute
}
```

And per query request

```
rank-profile coarse-to-fine {
  inputs {
    query(q) tensor<float>(x[128])
    query(q_binary) tensor<int8>(x[16])
  }
}
```
Then the following should will be one inference call, but with different post processing
of the model output. 

```
input.query(q)=embed(my-embedder-id,"the text to embed")
input.query(q_binary)=embed(my-embedder-id,"the text to embed")
```

Without invoking the expensive model inference part twice, only the post-processing of the model changes. 

## Important assumptions 

The PR makes an important assumption that there is a single Context instance per embedder instance per request since it uses the input text string as they cache key. If this assumption is wrong then then it would break when inputting the same text contents to multiple embedder instances as the model outputs would not be compatible, leading to all kinds of issues. 

Also this is only a per request cache (not a cache across requests as discussed in https://github.com/vespa-engine/vespa/issues/27967) .

```
field embedding type tensor<float>(x[128]) {
  input | text | embed my-embedder-id
}

field embedding type tensor<float>(x[128]) {
  input | text | embed my-other-embedder-id
}
```

### DIFF
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/ExecutionContext.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/ExecutionContext.java
@@ -22,7 +22,7 @@ public class ExecutionContext implements FieldTypeAdapter, FieldValueAdapter {
     private final FieldValueAdapter adapter;
     private FieldValue value;
     private Language language;
-    private final Map<String, Object> cache = LazyMap.newHashMap();
+    private final Map<Object, Object> cache = LazyMap.newHashMap();
 
     public ExecutionContext() {
         this(null);
@@ -125,12 +125,12 @@ public class ExecutionContext implements FieldTypeAdapter, FieldValueAdapter {
     }
 
     /** Returns a cached value, or null if not present. */
-    public Object getCachedValue(String key) {
+    public Object getCachedValue(Object key) {
         return cache.get(key);
     }
 
     /** Returns a mutable reference to the cache of this. */
-    public Map<String, Object> getCache() {
+    public Map<Object, Object> getCache() {
         return cache;
     }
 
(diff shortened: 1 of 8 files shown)


bjorncs (dismissed):

jobergum:
It looks like my assumption is wrong. The cache instance is owned by ExecutionContext so it's global and not unique per embedder id.

bratseth:
Yes, it's shared for the duration of the processing of the document/update.
I suggest creating a CacheKey that contains both the value and the embedder id:

record EmbedderCacheKey(ComponentId embedderId, Object embeddedValue) { }

bratseth (commented):
Otherwise, LGTM!

bratseth:
I also think we shouldn't build the tensor input if it turns out we won't need it because we have a cached value. I'm working on a commit to add to this PR.

bratseth:
@jobergum please take a look

bjorncs (dismissed):

jobergum:
Thanks @bratseth. I' will adapt the HF embedder as well to cache more of the compute and change the hash key from string to a record.

bjorncs:
With any object allowed as key it's important that all types of the composite key do have a sensible `equals()`/`hashcode()` implementation.

bjorncs (approved):

jobergum:
Updated performance test to use multiple embed calls, so need a couple of runs before merging https://github.com/vespa-engine/system-test/pull/3740

PR #30810, opened 2024-04-04
Limit amount of work in DocumentV1 Q to 4096, or 3s.

baldersheim:
Having only a fixed length here does no work well when throughput is low.

@jonmv or @bjorncs PR

### DIFF
--- a/vespaclient-container-plugin/src/main/resources/configdefinitions/document-operation-executor.def
+++ b/vespaclient-container-plugin/src/main/resources/configdefinitions/document-operation-executor.def
@@ -7,3 +7,5 @@ resendDelayMillis     int default=10
 # Bound on number of document operations to keep in retry queue — further operations are rejected
 maxThrottled          int default=4096
 
+# Max age in seconds of message in throttled Q.
+maxThrottledAge       double default=3.0
--- a/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
+++ b/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
@@ -113,6 +113,7 @@ public class DocumentV1ApiTest {
             .build();
     final DocumentOperationExecutorConfig executorConfig = new DocumentOperationExecutorConfig.Builder()
             .maxThrottled(2)
+            .maxThrottledAge(1.0)
             .resendDelayMillis(1 << 30)
             .build();
     final DocumentmanagerConfig docConfig = Deriver.getDocumentManagerConfig("src/test/cfg/music.sd")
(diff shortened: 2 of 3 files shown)


vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
```diff
@@ -188,6 +189,59 @@ public void testResolveCluster() {
...
+        driver.close();
+    }
+
+    @Test
+    public void testOverLoadByAge() {
+        RequestHandlerTestDriver driver = new RequestHandlerTestDriver(handler);
+        // OVERLOAD is a 429
+        access.session.expect((id, parameters) -> new Result(Result.ResultType.TRANSIENT_ERROR, Result.toError(Result.ResultType.TRANSIENT_ERROR)));
+        var response1 = driver.sendRequest("http://localhost/document/v1/space/music/number/1/two", POST, "{\"fields\": {}}");
+        try { Thread.sleep(3_000); } catch (InterruptedException e) {}
```
bjorncs, line 227: There is a `HttpRequest` constructor taking timestamp as nano seconds. Consider adding a new `RequestHandlerTestDriver.sendRequest()` overload with timestamp. That way we can lower the test cost by a couple of magnitudes ;)

vespaclient-container-plugin/src/main/resources/configdefinitions/document-operation-executor.def
```diff
@@ -7,3 +7,5 @@ resendDelayMillis     int default=10
 # Bound on number of document operations to keep in retry queue — further operations are rejected
 maxThrottled          int default=4096
 
+# Max age in seconds of message in throttled Q.
+maxThrottledAge       double default=3.0
```
bjorncs, line 11: 3 seconds sounds fine for me. @jonmv any objection?

bjorncs (approved):

vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
baldersheim: But that is not the same timestamp as 
`private final long jvmRelativeCreatedAt = System.nanoTime();`

bjorncs: Okay, in that case we need another overload.

jonmv (commented):
This looks great!

baldersheim:
Unfortunately it was not enough. You will fill the Q with 4k elements when starting up anyway before there is any pushback from mbus. After discussion with @bjorncs we decided to try without the Q and get direct feedback from mbus, since that already has a dynamic Q.

baldersheim:
Revert until all our main clients have been verified.

PR #30815, opened 2024-04-04
- Avoid a Q in the document v1 handler. Rely only on mbus Q.

baldersheim:
- This avoid filling a potentially large Q, and allows for more direct and better feedback to client.

@bjorncs and @jonmv PR


bjorncs (commented):
👍

PR #30816, opened 2024-04-04
Add significance model registry to linguistics

MariusArhaug:
FYI @geirst

PR #30819, opened 2024-04-04
Add significance config model registry

MariusArhaug:


config-model/src/test/java/com/yahoo/vespa/model/significance/test/SignificanceModelTestCase.java
```diff
@@ -0,0 +1,28 @@
...
+public class SignificanceModelTestCase {
+    private VespaModel createModel(String filename) {
+        return new VespaModelCreatorWithFilePkg(filename).create();
+    }
+
+    @Test
+    void testIndexGreaterThanNumNodes() {
+        VespaModel vespaModel = createModel("src/test/cfg/significance");
+        ApplicationContainerCluster containerCluster = vespaModel.getContainerClusters().get("container");
+        assertEquals(1, containerCluster.getContainers().size());
```
bjorncs, line 25: Test that the component is configured. The test now only checks that there is a container cluster.

config-model/src/main/java/com/yahoo/vespa/model/container/component/SignificanceModelRegistry.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * A registry for significance models.
+ *
+ * @author MariusArhaug
+ *
+ */
+public class SignificanceModelRegistry extends SimpleComponent implements SignificanceConfig.Producer {
+
+    private static final String CLASS = "com.yahoo.search.significance.DefaultSignificanceModelRegistry";
```
bjorncs, line 26: Update to reflect requested change in https://github.com/vespa-engine/vespa/pull/30816#discussion_r1557038341

config-model/src/test/java/com/yahoo/vespa/model/significance/test/SignificanceModelTestCase.java
MariusArhaug: Added another test for this

config-model/src/test/java/com/yahoo/vespa/model/significance/test/SignificanceModelTestCase.java
```diff
@@ -1,13 +1,19 @@
...
+import com.yahoo.search.significance.config.SignificanceConfig;
 import com.yahoo.vespa.model.VespaModel;
 import com.yahoo.vespa.model.container.ApplicationContainerCluster;
+import com.yahoo.vespa.model.container.component.SignificanceModelRegistry;
 import com.yahoo.vespa.model.test.utils.VespaModelCreatorWithFilePkg;
 import org.junit.jupiter.api.Test;
 
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.*;
```
bjorncs, line 16: Don't use wildcard imports. It's common practice to disable that automation in IntelliJ.

config-model/src/test/cfg/significance/services.xml
```diff
@@ -0,0 +1,29 @@
...
+    <redundancy>1</redundancy>
+
+    <documents>
+      <document type="music" mode="index"/>
+    </documents>
+
+    <group name="mycluster">
+      <node hostalias="node0" distribution-key="0" capacity="0.79"/>
+    </group>
+  </content>
```
bjorncs, line 27: Check whether this can be removed. Same for the `music.sd` file. The test should be as narrow as possible to reduce the cost of maintaining the test in the future.

bjorncs (approved):

PR #30820, opened 2024-04-04
When trottled we need to back down further.

baldersheim:
Need to back down below the minimum size of mbus Q size.
This needs closer thought, but at least this should bive a better picture for the tests.
@bjorncs @jonmv


jonmv:
Yes ... the minimum number of inflight operations was probably too high here, still.

bjorncs (commented):
Last time I tried the same change it resulted in regressions for multiple feeding performance tests 🤞

PR #30826, opened 2024-04-05
Balder/nap insteadof beautysleep

baldersheim:
@jonmv or @bjorncs PR

### DIFF
--- a/vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/JettyCluster.java
+++ b/vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/JettyCluster.java
@@ -72,7 +72,7 @@ class JettyCluster implements Cluster {
 
     JettyCluster(FeedClientBuilderImpl b) throws IOException {
         this.client = createHttpClient(b);
-        this.endpoints = b.endpoints.stream().map(Endpoint::new).toList();
+        this.endpoints = b.endpoints.stream().map(Endpoint::new).collect(Collectors.toList());
         this.compression = b.compression;
     }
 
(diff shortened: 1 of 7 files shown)


vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/JettyCluster.java
```diff
@@ -239,11 +239,7 @@ private static String endpointUri(URI uri) {
         return String.format("%s://%s:%s", uri.getScheme(), uri.getHost(), portOf(uri));
     }
 
-    private static class JettyResponse implements HttpResponse {
-        final Response response;
-        final byte[] content;
-
-        JettyResponse(Response response, byte[] content) { this.response = response; this.content = content; }
+    private record JettyResponse(Response response, byte[] content) implements HttpResponse {
```
bjorncs, line 242: Records are not available on JDK11 - the version used by our internal Hadoop integration.

bjorncs (changes requested):

vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/JettyCluster.java
baldersheim: Reverted, but shouldn't this be limited in the pom file so intellij would not suggest it ?

baldersheim: Did not fail maven build either....

bjorncs (approved):

vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/JettyCluster.java
bjorncs: That is unexpected. The build file targets 11 for all clients.

jonmv:
👍

PR #30848, opened 2024-04-08
Avoid spotbugs annotations by including them in provided scope to tri…

baldersheim:
…ck compiler.

@bjorncs PR


bjorncs (dismissed):
FYI Spotbug-annotation is an OSGi bundle. Bundles embedding zkfacade will fail to load at runtime.

bjorncs:
@gjoranv Is it possible to deny-list certain packages from being imported in bundle-plugin configuration?

baldersheim:
clustercontroller-core does the same thing and that seems to work fine.

bjorncs:
@baldersheim Ok, let's try this in CI then.

gjoranv:
> @gjoranv Is it possible to deny-list certain packages from being imported in bundle-plugin configuration?

Not out of the box. Probably should be.

baldersheim:
Closing in favor of #30854

PR #30849, opened 2024-04-08
Add flag for using new path disk encryption key metadata

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -422,6 +422,12 @@ public class Flags {
             "Role definitions for the system",
             "Takes effect immediately");
 
+    public static UnboundBooleanFlag NEW_PATH_FOR_DISK_ENCRYPTION_KEY_METADATA  = defineFeatureFlag(
+            "new-path-for-disk-encryption-key-metadata", false,
+            List.of("hmusum"), "2024-04-08", "2024-06-01",
+            "Whether to read and write disk encryption key to new path",
+            "Takes effect on next host admin tick.");
```
hakonhall, line 429: Will be read only on boot?

hakonhall (dismissed):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hmusum: Yes, thanks, description updated

hakonhall (approved):

PR #30860, opened 2024-04-09
Support pipelining (batching) of mutating ops to same bucket

vekterli:
@baldersheim please review

Bucket operations require either exclusive (single writer) or shared (multiple readers) access. Prior to this commit, this means that many enqueued feed operations to the same bucket introduce pipeline stalls due to each operation having to wait for all prior operations to the bucket to complete entirely (including fsync of WAL append). This is a likely scenario when
feeding a document set that was previously acquired through visiting, as such documents will inherently be output in bucket-order.

With this commit, a configurable number of feed operations (put, remove and update) bound for the exact same bucket may be sent asynchronously to the persistence provider in the context of the _same_ write lock. This mirrors how merge operations work for puts and removes.

Batching is fairly conservative, and will _not_ batch across further messages when any of the following holds:

 * A non-feed operation is encountered
 * More than one mutating operation is encountered for the same document ID
 * No more persistence throttler tokens can be acquired
 * Max batch size has been reached

Updating the bucket DB, assigning bucket info and sending replies is deferred until _all_ batched operations complete.

Max batch size is (re-)configurable live and defaults to a batch size of 1, which shall have the exact same semantics as the legacy behavior. A later follow-up will add a feature flag to control this.

Additionally, clock sampling for persistence threads have been abstracted away to allow for mocking in tests (no need for sleep!).

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -1027,11 +1027,11 @@ constexpr bool is_batchable_feed_op(api::MessageType::Id id) noexcept {
 document::GlobalId gid_from_feed_op(const api::StorageMessage& msg) {
     switch (msg.getType().getId()) {
     case api::MessageType::PUT_ID:
-        return dynamic_cast<const api::PutCommand&>(msg).getDocumentId().getGlobalId();
+        return static_cast<const api::PutCommand&>(msg).getDocumentId().getGlobalId();
     case api::MessageType::REMOVE_ID:
-        return dynamic_cast<const api::RemoveCommand&>(msg).getDocumentId().getGlobalId();
+        return static_cast<const api::RemoveCommand&>(msg).getDocumentId().getGlobalId();
     case api::MessageType::UPDATE_ID:
-        return dynamic_cast<const api::UpdateCommand&>(msg).getDocumentId().getGlobalId();
+        return static_cast<const api::UpdateCommand&>(msg).getDocumentId().getGlobalId();
     default: abort();
     }
 }
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -12,6 +12,7 @@
 #include <vespa/storage/persistence/messages.h>
 #include <vespa/storageapi/message/stat.h>
 #include <vespa/vespalib/stllike/hash_map.hpp>
+#include <vespa/vespalib/stllike/hash_set.hpp>
 #include <vespa/vespalib/util/exceptions.h>
 #include <vespa/vespalib/util/string_escape.h>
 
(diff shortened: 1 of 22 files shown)


vekterli:
Seems to be some nondeterminism in a test, will try to reproduce and remove.

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -992,6 +1008,92 @@ FileStorHandlerImpl::Stripe::getNextMessage(vespalib::steady_time deadline)
...
+
+// Precondition: msg must be a feed operation request (put, remove, update)
+document::GlobalId gid_from_feed_op(const api::StorageMessage& msg) {
+    switch (msg.getType().getId()) {
+    case api::MessageType::PUT_ID:
+        return dynamic_cast<const api::PutCommand&>(msg).getDocumentId().getGlobalId();
+    case api::MessageType::REMOVE_ID:
+        return dynamic_cast<const api::RemoveCommand&>(msg).getDocumentId().getGlobalId();
+    case api::MessageType::UPDATE_ID:
+        return dynamic_cast<const api::UpdateCommand&>(msg).getDocumentId().getGlobalId();
```
baldersheim, line 1034: dynamic_cast is a rather expensive operation. Consider replacing with static_cast or a getAsXXMsg cast safe method.

baldersheim (approved):

baldersheim (dismissed):

baldersheim:
PR build has spotted something I did not  .....

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
vekterli: Replaced with `static_cast`. Downcast-safe type invariant shall be maintained by the message's own type ID tracking. If it's not, we'd have other, bigger problems.

toregge:
Running all unit tests in parallel causes the following failure on x86_64 :

```
gdb) where
#0  raise (sig=11) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  <signal handler called>
#2  0x00007f0895e556aa in storage::MessageTracker::sendReply (this=0x7f0874001e80)
    at /home/tegge/git/vespa/storage/src/vespa/storage/persistence/persistenceutil.cpp:155
#3  0x00007f0895e3ad31 in operator() (response=..., __closure=0x7f0874000da8)
    at /opt/rh/gcc-toolset-13/root/usr/include/c++/13/bits/unique_ptr.h:199
#4  storage::(anonymous namespace)::LambdaResultTask<storage::AsyncHandler::handlePut(storage::api::PutCommand&, storage::MessageTracker::UP) const::<lambda(storage::spi::Result::UP)> >::run(void) (this=0x7f0874000d90)
    at /home/tegge/git/vespa/storage/src/vespa/storage/persistence/asynchandler.cpp:66
#5  0x00007f088c84a4ba in vespalib::ThreadStackExecutorBase::run (this=0x20ee000)
    at /home/tegge/git/vespa/vespalib/src/vespa/vespalib/util/threadstackexecutorbase.cpp:125
#6  0x00007f0896711dd9 in storage::(anonymous namespace)::test_executor (worker=...)
    at /home/tegge/git/vespa/storage/src/tests/common/teststorageapp.cpp:96
#7  0x00007f0888220b23 in std::execute_native_thread_routine (__p=0x20ee5c0) at ../../../../../libstdc++-v3/src/c++11/thread.cc:80
#8  0x00007f08884fb1ca in start_thread (arg=<optimized out>) at pthread_create.c:479
#9  0x00007f0887838e73 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) 
```

baldersheim (approved):

vekterli:
A few sins have been absolved... Some new, some old.

PR #30870, opened 2024-04-10
consolidate solutions into a single heuristic algorithm

havardpe:
use error relative to minimal cost rather than error potential

@geirst please review


geirst (approved):
👍

PR #30872, opened 2024-04-10
Support populating attribute where docid sets matching terms/children…

geirst:
… are disjunct.

This also adds testing of IN operator with many tokens (up to 10k), which uses disjunct terms/children.

In addition calculation of "alternative cost" is removed, as it doesn't provide any value at the moment.

@havardpe please review

### DIFF
--- a/searchlib/src/tests/queryeval/iterator_benchmark/benchmark_blueprint_factory.cpp
+++ b/searchlib/src/tests/queryeval/iterator_benchmark/benchmark_blueprint_factory.cpp
@@ -42,11 +42,11 @@ calc_hits_per_term(uint32_t num_docs, double op_hit_ratio, uint32_t children, Qu
 }
 
 std::unique_ptr<BenchmarkSearchable>
-make_searchable(const FieldConfig& cfg, uint32_t num_docs, const HitSpecs& hit_specs)
+make_searchable(const FieldConfig& cfg, uint32_t num_docs, const HitSpecs& hit_specs, bool disjunct_terms)
 {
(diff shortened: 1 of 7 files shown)


searchlib/src/tests/queryeval/iterator_benchmark/benchmark_blueprint_factory.cpp
```diff
@@ -159,24 +159,30 @@ class MyFactory : public BenchmarkBlueprintFactory {
...
     HitSpecs hit_specs(55555);
-    hit_specs.add(default_values_per_document, num_docs);
+    if (!disjunct_children) {
+        hit_specs.add(default_values_per_document, num_docs);
+    }
     _terms = hit_specs.add(children, hits_per_term);
-    _searchable = make_searchable(field_cfg, num_docs, hit_specs);
+    if (disjunct_children) {
+        uint32_t op_num_hits = num_docs * op_hit_ratio;
+        hit_specs.add(default_values_per_document, num_docs - op_num_hits);
```
havardpe, line 183: why less here?

havardpe (dismissed):

searchlib/src/tests/queryeval/iterator_benchmark/benchmark_blueprint_factory.cpp
geirst: This ensures that the remaining docids are populate with a "default value". Only a single default value is supported when using disjunct children, so I have altered the code to reflect that.

PR #30875, opened 2024-04-10
Lesters/add local llms

lesters:
@bratseth Please review
@hmusum FYI
@arnej27959 FYI
@andreer FYI

This is **initial** support for running LLMs locally using llama.cpp through java-llama.cpp and the container-llama module. This supports most things such as concurrency and GPU support if shared libraries have been compiled with it. The `LocalLLMTest.java` can be run on a laptop and the included libraries support Mac GPUs. You just need to download a GGUF file for instance from HuggingFace.

The `llm-local-client.def` contains the most significant model parameters, but we need more experience with this to be able to tune this properly and run in the context of Vespa without blowing stuff up. Things like total memory use, GPU memory use, context window size, number of parallel completions and so on. I'll be adding some system and performance tests shortly to start this.


container-search/src/main/java/ai/vespa/llm/clients/LocalLLM.java
```diff
@@ -0,0 +1,137 @@
...
+                }
+                completionFuture.complete(Completion.FinishReason.stop);
+            });
+        } catch (RejectedExecutionException e) {
+            // If we have too many requests (active + any waiting in queue), we reject the completion
+            int activeCount = executor.getActiveCount();
+            int queueSize = executor.getQueue().size();
+            String error = String.format("Rejected completion due to too many requests, " +
+                    "%d active, %d in queue", activeCount, queueSize);
+            logger.info(error);
```
bratseth, line 130: Should we throw here?
Probably want to return a 429 and not fill the log with these ...

container-search/src/main/resources/configdefinitions/llm-local-client.def
```diff
@@ -0,0 +1,32 @@
+# Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package=ai.vespa.llm.clients
+
+# Url to the model to use
```
bratseth, line 4: Use a "model" parameter type instead of url to have the path, url and cloud options handled as one.

container-search/src/test/java/ai/vespa/llm/clients/LocalLLMTest.java
```diff
@@ -0,0 +1,232 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for LocalLLM. Tests are disabled due to size of models.
```
bratseth, line 24: Maybe we can find a super-small model that can be used for unit testing?

bratseth (commented):
Great! See comments, also repeating my request to move the generic model stuff out of search.

lesters:
@bratseth Found a relatively small LLM (1.1Mb) which is suitable for unit testing

lesters:
Added DO NOT MERGE for some run-time testing of the dependencies.

@gjoranv Please take a look.

container-llama/src/main/java/de/kherud/llama/package-info.java
```diff
@@ -0,0 +1,8 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+/**
+ * @author lesters
+ */
+@ExportPackage
+package de.kherud.llama;
```
gjoranv, line 6: Do we really need to export this package?

gjoranv (dismissed):
Please verify that vespa-yahoo builds before merging, to ensure that dependency enforcers don't fail.

lesters:
@gjoranv Thanks for reminding me, have built vespa-yahoo and added dependencies there.

bratseth (approved):
Great!

PR #30878, opened 2024-04-11
Validate url for models

hmusum:


config-model/src/test/java/com/yahoo/vespa/model/container/component/ModelTest.java
```diff
@@ -0,0 +1,36 @@
...
+    @Test
+    void invalid_url(){
+        var xml = """
+                <component id="bert-embedder" type="bert-embedder">
+                  <transformer-model url="models/e5-base-v2.onnx" />
+                  <tokenizer-vocab path="models/vocab.txt"/>
+                </component>
+                """;
+
+        try {
```
bjorncs, line 27: Consider simplifying by using Junit5's `assertThrows` instead of the try-catch-fail pattern.

bjorncs (dismissed):

config-model/src/test/java/com/yahoo/vespa/model/container/component/ModelTest.java
hmusum: Good idea, done

config-model/src/test/java/com/yahoo/vespa/model/container/component/ModelTest.java
```diff
@@ -0,0 +1,33 @@
...
+                  <transformer-model url="models/e5-base-v2.onnx" />
+                  <tokenizer-vocab path="models/vocab.txt"/>
+                </component>
+                """;
+
+        var state = new DeployState.Builder().build();
+        var element = XML.getDocument(xml).getDocumentElement();
+        assertThrows(IllegalArgumentException.class,
+                     () -> Model.fromXml(state, element, "transformer-model", Set.of()),
+                     "Invalid url 'models/e5-base-v2.onnx': url has no 'scheme' component");
```
bjorncs, line 30:
```suggestion
var exception = assertThrows(IllegalArgumentException.class,
                     () -> Model.fromXml(state, element, "transformer-model", Set.of()));
org.junit.jupiter.api.Assertions.assertEquals("Invalid url 'models/e5-base-v2.onnx': url has no 'scheme' component", exception.getMessage());                   
```

bjorncs (approved):

PR #30887, opened 2024-04-11
Improve detection of updated config for service layer process.

toregge:
@vekterli : please review


vekterli (approved):
:+1: an oversight when unifying and rewiring a whole bunch of config subscriptions earlier

PR #30903, opened 2024-04-12
add code to AndFlow that can perform additional incremental reordering

havardpe:
of children based on the future possibility of forcing those children to be strict.

update unit test to use code from AndFlow and add tests for non-strict AND ordering as well.

@geirst please review


geirst (approved):
👍

PR #30905, opened 2024-04-12
improve strict and sorting when allowed

havardpe:
@geirst please review

This will not affect production code, but will affect factory via the sort_by_cost feature flag (hopefully not in a negative way)


geirst (approved):
👍

PR #30920, opened 2024-04-15
pass in_flow to intermediate sort functions

havardpe:
improve non-strict AND sorting by letting allow_force_strict affect ordering.

@geirst please review


geirst (approved):
👍

PR #30928, opened 2024-04-15
- Remove duplicates of the timestamp.

baldersheim:
- Avoid 17M of duplicates of 0.0.
- Avoid duplicating the optionally sanitized metric name.

@bratseth or @bjorncs PR

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
@@ -28,6 +28,7 @@ import static ai.vespa.metricsproxy.metric.model.DimensionId.toDimensionId;
  * @author Jo Kristian Bergum
  */
 public class MetricsParser {
+    private static final Double ZERO_DOUBLE = 0d;
     public interface Collector {
         void accept(Metric metric);
     }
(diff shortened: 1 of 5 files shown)


bratseth (dismissed):
👍

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
```diff
@@ -28,6 +28,7 @@
  * @author Jo Kristian Bergum
  */
 public class MetricsParser {
+    private static final Double ZERO_DOUBLE = 0d;
```
bjorncs, line 31: This shouldn't be necessary as long as `Double.valueOf()` is used

bjorncs (dismissed):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
baldersheim: Double.valueof does not have any of that as far as I can tell.
Long.valueof has for [-128, 127]. However when I use Long.valueof intellij says it can be dropped due to autoboxing.
Does that mean that auoboxing uses X.valueof, or is Intellij not correct. ?

bjorncs (approved):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/MetricsParser.java
bjorncs: That is surprising. The Javadoc clearly states that `Double.valueOf()` should be used for caching/performance. I'm not sure what auto-boxing does. Th use of the `Double` constructor should be avoided though.

PR #30929, opened 2024-04-16
Revert "- Avoid a Q in the document v1 handler. Rely only on mbus Q."

baldersheim:
Reverts vespa-engine/vespa#30815
@bjorncs PR


jonmv (approved):

bjorncs (commented):
👍

PR #30931, opened 2024-04-16
Lesters/add local llms 3

lesters:
@gjoranv Please review.
@hmusum FYI.
@bratseth FYI.

Reapplication of https://github.com/vespa-engine/vespa/pull/30916, but removes addition of `model-integration` as compile scope. The two exported classes that should be able to be specified in `services.xml` without a `bundle` attribute are added to the  `PlatformBundles` hack, similarly to the rest of the classes in container-search and -docproc.

Also disabled the unit test for local LLMs, as the run time varied wildly depending on platform. On the open source build, these tests took around 1 second, but on vespa-yahoo build they took 814s. This is likely due to the native libraries packaged with the java llama dependency, which work for the architecture used in the open source build, but not for vespa-yahoo. The system-test in https://github.com/vespa-engine/system-test/pull/3779 will test this functionality with correct native dependencies.


hmusum (approved):
Let's try it

gjoranv:
👍

PR #30933, opened 2024-04-16
- Optimize distance calculation for tensors with single dense subspace.

baldersheim:
- Let EmptySubspace be invalid.
- Add noexcept to get_tensor(s).

@geirst and @havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/distance_calculator.h
+++ b/searchlib/src/vespa/searchlib/tensor/distance_calculator.h
@@ -39,11 +39,14 @@ public:
     const BoundDistanceFunction& function() const noexcept { return *_dist_fun; }
     bool has_single_subspace() const noexcept { return _attr_tensor.getTensorType().is_dense(); }
 
-    template<bool has_single_subspace=false>
+    template<bool has_single_subspace>
     double calc_raw_score(uint32_t docid) const {
         if (has_single_subspace) {
-            double distance = _dist_fun->calc(_attr_tensor.get_vector(docid, 0));
-            return _dist_fun->to_rawscore(distance);
+            auto cells = _attr_tensor.get_vector(docid, 0);
+            if (cells.size == 0) [[unlikely]] {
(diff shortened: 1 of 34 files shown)


havardpe:
I suggest using a TypedCells with size 0 rather than adding an invalid bool to signal that the cells are not present. Note that most tensor code will expect that the appropriate number of cells are present. We might need to have an empty subspace with a pointer to the appropriate number of zero values combined with a size of zero, making it work either way...

baldersheim:
That is the approach used now. It has a proper backing of zero valued cells, but says its size is zero.

searchlib/src/vespa/searchlib/tensor/distance_calculator.h
```diff
@@ -36,27 +36,44 @@ class DistanceCalculator {
...
+    template<bool has_single_subspace=false>
     double calc_raw_score(uint32_t docid) const {
-        auto vectors = _attr_tensor.get_vectors(docid);
-        double result = _dist_fun->min_rawscore();
-        for (uint32_t i = 0; i < vectors.subspaces(); ++i) {
-            double distance = _dist_fun->calc(vectors.cells(i));
-            double score = _dist_fun->to_rawscore(distance);
-            result = std::max(result, score);
+        if (has_single_subspace) {
+            double distance = _dist_fun->calc(_attr_tensor.get_vector(docid, 0));
```
geirst, line 45: Shouldn't we check for `cells.size == 0` here as well? If so, that case should be covered by a unit test.

geirst: We should also use `max(_dist_fun->min_rawscore(), this_score)` to match existing behaviour.

geirst (approved):
👍  But I would like to see a unit test (if not already existing)

baldersheim:
It was already tested, it failed when I started specifying the template based on tensor type.

PR #30934, opened 2024-04-16
Use memory mapped disk index dictionary .ssdat file when large

toregge:
and with sufficient padding.  This reduces memory usage spikes.

@geirst : please review


geirst (approved):
👍

PR #30936, opened 2024-04-16
To avoid keeping all Prometheus Metrics in memory while aggregating, …

baldersheim:
…iterate multiple time and build

a single metric family while serializing.
This will significantly reduce amount of memory required.

@bjorncs @bratseth PR

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/prometheus/PrometheusModel.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/prometheus/PrometheusModel.java
@@ -1,10 +1,13 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package ai.vespa.metricsproxy.metric.model.prometheus;
 
+import ai.vespa.metricsproxy.metric.model.DimensionId;
 import ai.vespa.metricsproxy.metric.model.MetricId;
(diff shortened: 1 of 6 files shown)


metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/prometheus/PrometheusModel.java
```diff
@@ -45,4 +53,31 @@ public String serialize() {
...
+                    labels.add(labelName);
+                    labelValues.add(entry.getValue());
+                }
+                labels.add("vespa_service");
+                labelValues.add(serviceName);
+
+                Number metric = packet.metrics().get(metricId);
+                if (metric != null) {
+                    sampleList.add(new Collector.MetricFamilySamples.Sample(metricId.getIdForPrometheus(), labels, labelValues, metric.doubleValue(), timeStamp));
+                }
```
bjorncs, line 77: Can the code skip aggregating values into `labels` and `labelValues` if `metric` is null?

bjorncs (dismissed):

bratseth (dismissed):

metrics-proxy/src/main/java/ai/vespa/metricsproxy/metric/model/prometheus/PrometheusModel.java
baldersheim: Thanks, yes it can, fixed

PR #30947, opened 2024-04-17
Set docid limit early in blueprint subtree.

toregge:
Cap matching doc count to total doc count.

@geirst : please review
@havardpe : FYI


geirst (approved):
👍

PR #30950, opened 2024-04-17
Balder/typedcells as value param

baldersheim:
@geirst or @toregge PR

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hamming_distance.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/hamming_distance.cpp
@@ -6,6 +6,7 @@
 
 using vespalib::typify_invoke;
 using vespalib::eval::TypifyCellType;
+using vespalib::eval::TypedCells;
 
 namespace search::tensor {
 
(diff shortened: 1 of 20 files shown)


searchlib/src/vespa/searchlib/tensor/hamming_distance.cpp
```diff
@@ -6,80 +6,58 @@
...
-    explicit BoundHammingDistance(const vespalib::eval::TypedCells& lhs)
+    explicit BoundHammingDistance(TypedCells lhs)
         : _tmpSpace(lhs.size),
           _lhs_vector(_tmpSpace.storeLhs(lhs))
     {}
-    double calc(const vespalib::eval::TypedCells& rhs) const override {
+    double calc(TypedCells rhs) const noexcept override {
         size_t sz = _lhs_vector.size();
         vespalib::ConstArrayRef<FloatType> rhs_vector = _tmpSpace.convertRhs(rhs);
-        assert(sz == rhs_vector.size());
```
toregge, line 48: This assert should probably be kept, to have same set of sanity checks regardless of FloatType.

toregge (approved):

searchlib/src/vespa/searchlib/tensor/hamming_distance.cpp
baldersheim: This is in the innerloop, and for hamming distance the inner loop is very short.
Since it has never triggered in any of the calculators I think it is safe to remove it here.
We should also see if we can remove them in other places too.

geirst (approved):
👍

PR #30974, opened 2024-04-19
Stop adding extra documentid field to document type.

toregge:
@geirst : please review

Since autumn 2022 the document summary framework in the backend has supported
handling summary fields as pure virtual fields. A summary field is just an
output name, a source field (from the document type) and an optional transform
specifying how to produce the output. All usages of summary fields are virtual
as of Vespa 8.253.17 (November 2023), and the generation and population of
some summary fields as extra document fields can now be removed.


geirst (approved):
👍

Issue #30720, opened 2024-03-22
[Feature request] Prefix match support fuzziness


KazaBen:
**Is your feature request related to a problem? Please describe.**

When implementing autocomplete application, most of the queries are incomplete words.
Sometimes misspells happen in incomplete queries too. Let's look at an example:

`Document = Fjallraven` (notice double L)
`Query = Fjalr` (incomplete query, with misspell: only a single L)

It is a must do for autocomplete application to be still able to match `Fjallraven` with user query `Fjalr`.

**Describe the solution you'd like**

When using [Prefix match](https://docs.vespa.ai/en/text-matching.html#prefix-match) add support for [Fuzziness](https://docs.vespa.ai/en/reference/query-language-reference.html#fuzzy).

**Describe alternatives you've considered**

Combining prefix matching with word fuzzy matching: `term (prefix) contains OR term contains fuzzy` 
However, it would not cover the previously mentioned common case:
`Document = Fjallraven` (notice double L)
`Query = Fjalr` (incomplete query, with misspell: only a single L)

`term (prefix) contains` couldn't match as it query has a misspell and there is no fuzziness support
`term contains fuzzy` couldn't match as it is incomplete word

**Additional context**

This exact functionality exists in [Elasticsearch’s Completion Suggester Fuzzy Queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html#fuzzy)

vekterli:
@KazaBen I've started working on this feature between a few other things. Implementation will be a two-step process:

1. Add support for prefix matching to our core Levenshtein algorithm implementations (Deterministic Finite Automata used for max edits {1, 2} and a generalized Levenshtein matrix fallback for max edits > 2). **Work in progress**.
2. Wire this new prefix matching mode into the query evaluation pipeline. **Not yet started**.

I should be able to give an estimate once I've started work on part 2 (hopefully within a few days, assuming nothing else pops up) and have gotten a gist of the complexity involved.

vekterli:
Update: part 2 has been completed and its PR is pending code review. Once it has been merged and a new Vespa version has been released containing the changes, fuzzy prefix matching can be used by adding a `prefix:true` annotation to your query term. Example YQL:
```
{maxEditDistance:1,prefix:true}fuzzy("Fjalr")
```
Assuming no other blockers, I'd expect it to be available as part of an Open-source release some time next week. I'll update this issue with a concrete version number once that happens.

A few notes:
[...]

PR #30976, opened 2024-04-19
Wire fuzzy prefix matching support through the query stack

vekterli:
@geirst please review.
@toregge FYI

Adds `prefix:[true|false]` annotation support to the `fuzzy` query operator in the YQL and JSON query languages. Fuzzy prefix matching semantics are wired through to the matcher implementations for both indexed and streaming search.

Example usage:
```
  {maxEditDistance:1,prefix:true}fuzzy("foo")
```
Will match `foo`, `foobar`, `foxtrot`, `zookeeper` and so on.

It can be combined with the existing prefix locking feature:
```
  {maxEditDistance:1,prefixLength:2,prefix:true}fuzzy("foo")
```
Which will match `foo`, `foobar`, `foxtrot` etc, but _not_ `zookeeper` since the locked prefix (`fo`) does not match.

Due to the complexities involved with extending the legacy binary query stack representation, signalling prefix matching for the fuzzy term is done by pragmatically adding a new, generic "prefix matching" term-level flag. This is currently ignored for everything except fuzzy query items.

Modernizing the query stack format to make it more extensible (i.e. move encoding to Protobuf) is on the backlog...!

This relates to #30720.

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -555,11 +555,15 @@
       "public"
     ],
     "methods" : [
+      "public void <init>(java.lang.String, boolean, java.lang.String, int, int, boolean)",
       "public void <init>(java.lang.String, boolean, java.lang.String, int, int)",
       "public void setMaxEditDistance(int)",
       "public void setPrefixLength(int)",
       "public int getPrefixLength()",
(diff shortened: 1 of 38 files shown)


geirst (approved):
👍

PR #30985, opened 2024-04-21
Bratseth/group size reduction

bratseth:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeSpec.java
```diff
@@ -80,9 +80,11 @@ default boolean canResize(NodeResources currentNodeResources, NodeResources curr
         return false;
     }
 
+    NodeSpec withExtra(int nodeCount);
```
freva, line 83: This seems complicated and kind of bypasses the rest of the algorithm. How about extending `NodeAllocation::shouldRetire` with something like
```
var group = candidate.allocation().get().membership().cluster().group();
if (group.isPresent() && requested.count().isPresent()) {
    long nodesInGroup = nodes.values().stream().filter(n -> groupOf(n).equals(group)).count();
    if (nodesInGroup >= requested.groupSize())
        return Retirement.someNewReason;
}
```

freva (dismissed):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/NodeSpec.java
bratseth: Yes, thanks. I agree it is nicer, I didn't do it because the GroupAssigner prioritizes which nodes to retire after seeing the entire set, but considering it closer I think we should end up with the same subset retired in most cases either way.

freva (approved):

PR #30990, opened 2024-04-22
Upgrade jllama to 3.0.1

lesters:
@arnej27959 Please review


baldersheim:
This is the same as #30987 by renovate, which will autoclose.

PR #31001, opened 2024-04-23
Add user-agent with contact

jobergum:

### DIFF
--- a/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
+++ b/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
@@ -22,7 +22,7 @@ class UrlDownloader implements Downloader {
 
     private static final Logger log = Logger.getLogger(UrlDownloader.class.getName());
     private static final String CONTENTS_FILE_NAME = "contents";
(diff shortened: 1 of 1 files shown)


hmusum (dismissed):
👍

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
```diff
@@ -22,11 +22,13 @@ class UrlDownloader implements Downloader {
 
     private static final Logger log = Logger.getLogger(UrlDownloader.class.getName());
     private static final String CONTENTS_FILE_NAME = "contents";
+    private static final String USER_AGENT_MODEL_DOWNLOADER = "Vespa.ai config-proxy url model downloader (https://github.com/vespa-engine/vespa)";
```
oyving, line 25: Let's not block a fix behind this comment, but we should have a RFC compliant User-Agent string. E.g. `User-Agent: Vespa/<version> (model download - https://github.com/vespa-engine/vespa)`

jobergum: Thanks, check now

bjorncs (approved):
Consider using `Vespa.VERSION` instead of hardcoded '8.x'

PR #31018, opened 2024-04-24
Add significance rank profile

MariusArhaug:
FYI @geirst


container-search/src/main/java/com/yahoo/search/query/Ranking.java
```diff
@@ -217,6 +219,12 @@ public RankProperties getProperties() {
     /** Returns whether rank features should be dumped with the result of this query, default false */
     public boolean getListFeatures() { return listFeatures; }
 
+    /** Set whether to use significance in ranking */
+    public void setUseSignificance(boolean useSignificance) { this.useSignificance = useSignificance; }
```
bjorncs, line 223: Annotate with `@com.yahoo.api.annotations.Beta`

bjorncs (approved):

PR #31019, opened 2024-04-24
Update defaults for local LLM config

lesters:
@hmusum Please review

Setting `useGpu` to true should not harm anything as it will not be enabled if a GPU is not detected. Also setting `parallelRequests` to a default of 1 should give a better initial user experience before tuning to handle larger traffic loads.


hmusum (commented):
👍

PR #31046, opened 2024-04-25
Use non_existing_attribute_value to signal that value is the default … MERGEOK

baldersheim:
…value for an attribute that has not been set.

@geirst or @havardpe @toregge PR


geirst (dismissed):
👍 @havardpe please also review

havardpe (dismissed):
consider adding a test for some cells that have the flag set to true

geirst (approved):
👍

PR #31047, opened 2024-04-25
add some resource attributes and more per-service labels

arnej27959:
@yngveaasheim or @olaaun please review


config-model/src/main/java/com/yahoo/vespa/model/admin/otel/OpenTelemetryConfigGenerator.java
```diff
@@ -203,4 +251,68 @@ void addStatePorts(List<StatePortInfo> portList) {
...
+        return applicationId.application().value();
+    }
+    private String appInstanceAttr() {
+        if (applicationId == null) return "noAppInstance";
+        return applicationId.instance().value();
+    }
+    private String cloudAttr() {
+        if (zone == null) return "noCloud";
+        return zone.cloud().name().value();
+    }
```
olaaun, line 282: Maybe we should add the dimensions iff the value is not null?

arnej27959: done, please take another look

olaaun (approved):

PR #31052, opened 2024-04-25
estimate actual cost

havardpe:
@geirst please review

This covers baseline cost estimation with extra strictness and non-optimal ordering
still to come: better cost estimates for complex leafs and btree/disk_index blueprints


geirst (approved):
👍

PR #31057, opened 2024-04-26
Allow scorer for wand to carry state

baldersheim:
@geirst and @havardpe PR


baldersheim:
Oral approval from @havardpe

PR #31062, opened 2024-04-26
Propagate "create if missing"-flag outside binary Update payload in protocols

vekterli:
@geirst please review. Most added lines are tests.
@baldersheim FYI

Avoids potentially having to deserialize the entire update just to get to a single bit of information that is technically metadata existing orthogonally to the _document_ update itself.

To ensure backwards/forwards compatibility, the flag is propagated as a Protobuf `enum` where the default value is a special "unspecified" sentinel, implying an old sender.

Since the Java protocol implementation always eagerly deserializes messages, it unconditionally assigns the `create_if_missing` field when sending and completely ignores it when receiving.

The C++ protocol implementation observes and propagates the field iff set. Otherwise the flag is deferred to the update object as before. This applies to both the DocumentAPI and StorageAPI protocols.


geirst (approved):
👍

PR #31097, opened 2024-05-02
Restart on deploy for local LLMs

lesters:
@hmusum Please review.

For discussion. Is this a valid approach?


hmusum:
Yes, looks good to me

PR #31113, opened 2024-05-04
Require a higher bar to take groups in rotation

bratseth:
If a majority of groups have been recently added, avoid taking any that are indexing a slightly faster than the rest in rotation before it has 97% of the content.

Notice that to achieve this we have to keep groups out of rotation until we get positive info from the groups. I'm not quite sure if that may have any negative consequences.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchGroupsImpl.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchGroupsImpl.java
@@ -7,14 +7,17 @@ import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
 
+/**
+ * @author baldersheim
+ */
 public class SearchGroupsImpl implements SearchGroups {
 
     private final Map<Integer, Group> groups;
-    private final double minActivedocsPercentage;
+    private final double minActiveDocsPercentage;
 
-    public SearchGroupsImpl(Map<Integer, Group> groups, double minActivedocsPercentage) {
+    public SearchGroupsImpl(Map<Integer, Group> groups, double minActiveDocsPercentage) {
         this.groups = Map.copyOf(groups);
-        this.minActivedocsPercentage = minActivedocsPercentage;
+        this.minActiveDocsPercentage = minActiveDocsPercentage;
     }
 
     @Override public Group get(int id) { return groups.get(id); }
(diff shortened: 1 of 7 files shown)


container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchGroupsImpl.java
```diff
@@ -23,23 +26,38 @@ public SearchGroupsImpl(Map<Integer, Group> groups, double minActivedocsPercenta
...
+                                             long groupDocumentCount, long medianDocumentCount, long maxDocumentCount) {
+        if (medianDocumentCount <= 0) return true;
+        if (currentIsGroupCoverageSufficient) {
+            // To take a group *out of* rotation, require that it has less active documents than the median.
+            // This avoids scenarios where incorrect accounting in a single group takes all other groups offline.
+            double documentCoverage = 100.0 * (double) groupDocumentCount / medianDocumentCount;
+            return documentCoverage >= minActiveDocsPercentage;
+        }
+        else {
+            // to put a group *in* rotation, require that it has as many documents as the largest group,
```
baldersheim, line 46: I wonder if there are situations where this will prevent groups that should have been included to join.
 Need to let this one mature a bit.

What about median(of all groups above average) ?

bratseth: That would be better in the case where you have one group with anomalously much data, at least three existing groups, and some new groups added. 

But if you had one group and are adding N new, then the larger N gets, the closer the average gets towards what those new groups have indexed so far on average, and if more than 1 of the new ones happens to go enough above the others to go above average, then 1 or more of them will be added prematurely.

baldersheim: I guess either of these solutions are fine, but I suggest we run them by @vekterli first.

baldersheim (approved):

container-search/src/main/java/com/yahoo/search/dispatch/searchcluster/SearchGroupsImpl.java
vekterli: One concern is that using "sticky" state (i.e. "does the group already have sufficient coverage?") as an input to the group coverage algorithm introduces an asymmetry between the decisions made by a container that has been up for some time versus one that has just started up. This because internal state is transient and groups are considered _down_ initially upon container startup.

Hypothetical scenario: the cluster is in some edge case where all but 1 group has coverage that is < 97% of max active but >= median active. If the container was alive when these groups had >= 97% coverage, they will _remain_ available for serving. If the container then restarts, this information is lost and the groups will _not_ be available for serving until they have reached >= 97% coverage once more.

I don't see this as something that should happen during normal operation, and these changes should make the behavior when adding fresh groups much better than it is today. So we'll probably have to observe it in practice for a bit before making any adjustments (related; should this behavior be guarded by a feature flag?).

vekterli (approved):
:+1:

PR #31137, opened 2024-05-07
Include value of 'idref' attribute when deciding which elements to ke…

hmusum:
…ep when preprocessing

We already include value of 'id' to key which is used to decide which elements are equal, but also need to include value of 'idref' so that we know which elements to keep e.g. in a list of searchers referred to by 'idref' in a search chain.


gjoranv (approved):
Perhaps we should deprecate 'idref'?

PR #31143, opened 2024-05-07
More help text for vespa log

kkraune:
- @mpolden Can you also add some code to encourage the user to run `vespa help log` when this happens, so it is known what to do next after reading the below.


mpolden:
`vespa log` is supported in the next version of CLI/Vespa.

Issue #30972, opened 2024-04-19
NPE while deploying 


ddashenkov:
I'm experiencing a `NullPointerException` (see the stacktrace below) resulting in a 500 status when deploying Vespa.

It seems to have appeared after I've changed a field from `indexing: attribute` to `indexing: attribute  | index`. Here is a snippet of the schema:
```diff
schema product {

    document product {

        ...

        field nesteds type array<Nested> {
            struct-field foo {
-               indexing: attribute
+               indexing: attribute | index
                match: exact
            }
        }
        
        struct Nested {
            field foo type string {}
        }
    }
}
```

When I do `vespa deploy`, the process fails with a 500 HTTP status. The docker container's logs contain the following message:
```
Unexpected error building default.default\nexception=\njava.lang.NullPointerException
at java.base/java.util.Objects.requireNonNull(Objects.java:209)
at java.base/java.util.TreeMap.getEntry(TreeMap.java:345)
at java.base/java.util.TreeMap.get(TreeMap.java:279)
at com.yahoo.schema.Schema.getAttribute(Schema.java:659)
at com.yahoo.schema.processing.TypedTransformProvider.requiresTransform(TypedTransformProvider.java:33)
at com.yahoo.vespa.indexinglanguage.ValueTransformProvider.shouldConvert(ValueTransformProvider.java:41)
at com.yahoo.vespa.indexinglanguage.ExpressionConverter.convert(ExpressionConverter.java:13)
at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
at java.base/java.util.LinkedList$LLSpliterator.forEachRemaining(LinkedList.java:1242)
at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:622)
at java.base/java.util.stream.ReferencePipeline.toList(ReferencePipeline.java:627)
[...]

PR #31144, opened 2024-05-07
Unit test that struct field with index and exact match gives expected

toregge:
log warning.

@geirst : please review

This is related to Issue #30972


geirst (approved):
👍

PR #31154, opened 2024-05-08
Make sure to close state v1 updater

hmusum:
Found this when looking at a shutdown of config server that was unsuccessful, suspecting that this prevented clean shutdown


hakonhall (approved):
Great

PR #31156, opened 2024-05-08
Explicitly track number of entries received through visiting

vekterli:
@jonmv please review. I think the changed counts in the tests should be accurate (but please take a look 👀).

Using the underlying session's `VisitorStatistics` may not be 1-1 with the actual number of entries the data handler has been invoked with. This causes issues if anyone tries to cross-check the document count emitted as part of the results vs. the number of entries actually present in the result array.

The session updates its statistics based on the what is returned from the backend as part of _successful_ `CreateVisitorReply` messages. If a CreateVisitor returns with a transient error, the statistics will not be updated, but it's unspecified how many (if any) document entries that particular visitor may already have pushed to the client directly from the content nodes.


jonmv (approved):

jonmv:
I guess the "documentCount" field is sort of useless.

jonmv:
It's useful for the other visit operations, though. And for humans who want a quick overview, i suppose ...

PR #31158, opened 2024-05-09
Reduce min inflight for go feeder, and start lower in both go and java

jonmv:
@baldersheim


jonmv:
When mbus starts at 20, there's no point in starting higher, just to immediately be throttled.

jonmv:
(The default is 8 connections, which now translates to starting at 16 inflight)

baldersheim (dismissed):

baldersheim (commented):
There are some tests that needs an update

PR #31165, opened 2024-05-10
use least common term to drive strict phrase search

havardpe:
stop using beginId since it only makes sense for the first (or full) docid space

remove misleading assert. We could still end up with an infinite loop without triggering the assert if the driving term matched the next document. Also, this assert is not used with other iterators that require strict children.

@geirst please review
@baldersheim FYI


geirst (approved):
👍

PR #31167, opened 2024-05-10
Add embedder metrics to vespa9 metricset

yngveaasheim:

Also move adding of embedder metrics to a better method.

@jobergum or @bjorncs Please review and merge.


jobergum (approved):

jobergum:
Are these then default in Prometheus etc?

yngveaasheim:
Yes, correct. For public, we use a smaller set for publiccd.

PR #31180, opened 2024-05-13
Explicitly track number of locally entries received through visiting

vekterli:
@jonmv please review. Same as #31156 with an added case for local vs. remote data handler.

Using the underlying session's `VisitorStatistics` may not be 1-1 with the actual number of entries the data handler has been invoked with. This causes issues if anyone tries to cross-check the document count emitted as part of the results vs. the number of entries actually present in the result array.

The session updates its statistics based on the what is returned from the backend as part of _successful_ `CreateVisitorReply` messages. If a CreateVisitor returns with a transient error, the statistics will not be updated, but it's unspecified how many (if any) document entries that particular visitor may already have pushed to the client directly from the content nodes.

Note that the locally tracked number is only reported if the session itself is receiving the document data; if a remote data handler is in use we have to report what the `VisitorStatistics` give us.


jonmv (approved):
Very nice!

PR #31193, opened 2024-05-13
Bratseth/quota capacity policies

bratseth:
@hakonhall please review at least the commits up to the last

@oyving please review the last one

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/QuotaValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/QuotaValidator.java
@@ -1,9 +1,12 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.model.application.validation;
 
+import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.Capacity;
+import com.yahoo.config.provision.CapacityPolicies;
 import com.yahoo.config.provision.ClusterResources;
 import com.yahoo.config.provision.ClusterSpec;
+import com.yahoo.config.provision.Exclusivity;
 import com.yahoo.config.provision.NodeResources;
 import com.yahoo.config.provision.QuotaExceededException;
 import com.yahoo.config.provision.SystemName;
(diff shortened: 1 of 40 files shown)


config-model/src/main/java/com/yahoo/vespa/model/application/validation/QuotaValidator.java
```diff
@@ -31,25 +34,34 @@ public class QuotaValidator implements Validator {
 
     @Override
     public void validate(Context context) {
+        var zone = context.deployState().zone();
+        var flagSource = context.deployState().flagSource();
+        var capacityPolicies = new CapacityPolicies(zone, new Exclusivity(zone, flagSource), flagSource);
```
hakonhall, line 39: Ideally the flags module would not become a config model dependency.

Both the SHARED_HOST flag and ADMIN_CLUSTER_NODE_ARCHITECTURE can be resolved in ModelContextImpl and then accessed here via `context.deployState().featureFlags()`.

SHARED_HOST resolves to a SharedHost class defined in the flags module, but the only thing needed for this PR is `Predicate<ClusterSpec.Type> exclusiveClusterType`.

bratseth: Well ADMIN_CLUSTER_NODE_ARCHITECTURE is used to do adminClusterNodeArchitecture.with(INSTANCE_ID, instance.serializedForm())

and SHARED_HOSTS to call both hasClusterType and supportsClusterType on the flag.

So, while doable it will be a bit messy.

These flags are not consumed in the config-model and whats more they are permanent, so the solution you suggest isn't actually needed. Maybe we could find a cleaner alternative - create an interface FlagSource implements that is part of e.g config-provisioning? Could also expose only the permanent flags perhaps?

bratseth: ... so part of the mismatch here is that the ModelContextImpl flags are meant to be resolved for a single model, while the instances we need to pass this on exist across applications and so need to do their own resolution. I've tried now, but this just doesn't fit that way of doing it, I actually think it's much cleaner then to just allow FlagSource to pass through the model as in this PR.

hakonhall: You mean CapacityPolicies? As it is now it could also be viewed as depending on the application and have to be created per application. I made https://github.com/vespa-engine/vespa/compare/master...hakonhall/quota-capacity-policies to verify this works.

That branch also makes flags depend on config-provisioning, instead of the other way around, which will allow type-safe setting of flag dimensions for most dimensions.  This is demonstrated with the shared host flag.

This PR changes the NodeRepository constructor, which requires a corresponding change in the internal repo (https://git.ouryahoo.com/vespa/vespa-yahoo/compare/master...hakonhall/quota-capacity-policies).

hakonhall: One benefit of resolving flags in ModelContextImpl is that each flag is resolved only once.

During NodeRepositoryProvisioner.prepare(), for example, some flags are resolved multiple times.  Since flags can change at any time, this means e.g. host-memory can have one value during one half of prepare(), and another value for the remainder.  It sounds wrong - a flag value should have one value throughout the preparation of a cluster.  I'm guessing most of these cases will converge in the next prepare, but I fear this may cause an incident at some point or at least will make certain behavior almost impossible to understand.

One benefit of adding flags as a dependency of the config model, is that it makes it possible to use the cluster dimension inside the config model.  Since flags are currently resolved at the application level in ModelContextImpl, if you need different values per cluster, the flag will have to resolve to a map from cluster type (or id) to some value.

bratseth: Nice!

bratseth:
Continued in https://github.com/vespa-engine/vespa/pull/31241

PR #31199, opened 2024-05-14
Adjust invalid feature name message when setting up first phase and

toregge:
second phase rank feature.

@geirst : please review


geirst (approved):
👍

PR #31204, opened 2024-05-14
Speed up bfloat16 to float conversion

baldersheim:
@vekterli and @havardpe PR

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
@@ -1,11 +1,13 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
 #include "temporary_vector_store.h"
+#include <vespa/vespalib/hwaccelrated/iaccelrated.h>
 
 using vespalib::ConstArrayRef;
 using vespalib::ArrayRef;
 using vespalib::eval::CellType;
 using vespalib::eval::TypedCells;
+using vespalib::hwaccelrated::IAccelrated;
 
 namespace search::tensor {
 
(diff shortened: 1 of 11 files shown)


searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
```diff
@@ -1,30 +1,43 @@
...
 convert_cells(ArrayRef<ToType> space, TypedCells cells) noexcept
 {
-    assert(cells.size == space.size());
-    auto old_cells = cells.typify<FromType>();
+    auto old_cells = cells.unsafe_typify<FromType>();
     ToType *p = space.data();
     for (FromType value : old_cells) {
-        ToType conv(value);
-        *p++ = conv;
+        *p++ = value;
```
vekterli, line 27: Are all other cell types well-defined with just a cast? Also, should we have a `static_cast` here to make the conversion more obvious?

searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
```diff
@@ -1,30 +1,43 @@
...
     }
     return space;
 }
 
+template<>
+ConstArrayRef<float>
+convert_cells<vespalib::BFloat16, float>(ArrayRef<float> space, TypedCells cells) noexcept
+{
+    static const IAccelrated & accelrator = IAccelrated::getAccelerator();
+    accelrator.convert_bfloat16_to_float(reinterpret_cast<const uint16_t *>(cells.data), space.data(), space.size());
```
vekterli, line 37: Nit: `s/accelrator/accelerator` 👀

vekterli (dismissed):
:+1: cross-checked with Godbolt, and the bf16 conversion routine gets auto-vectorized as expected

havardpe (dismissed):
I think this violates some aliasing rules, but hopefully that does not matter

searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
baldersheim: The type are double, float, BFloat16 and Int8Float. As far as I can tell they should be fine.

searchlib/src/vespa/searchlib/tensor/temporary_vector_store.cpp
baldersheim: Fixed

PR #31205, opened 2024-05-15
Improve javadoc

bratseth:


linguistics/src/main/java/com/yahoo/language/process/Segmenter.java
```diff
@@ -6,22 +6,23 @@
 import java.util.List;
 
 /**
- * Interface providing segmentation, i.e. splitting of CJK character blocks into separate tokens. This is primarily a
- * convenience feature for users who don't need full tokenization (or who use a separate tokenizer and only need CJK
- * processing).
+ * A segmnenter splits a string into separate segments (such as words) without applying any further
```
arnej27959, line 9: segmnenter -> segmenter

arnej27959 (dismissed):
👍

linguistics/src/main/java/com/yahoo/language/process/Segmenter.java
bratseth, line 9:
```suggestion
 * A segmenter splits a string into separate segments (such as words) without applying any further
```

PR #31207, opened 2024-05-15
Add method to extract instance name from SANs

freva:


vespa-athenz/src/main/java/com/yahoo/vespa/athenz/tls/AthenzX509CertificateUtils.java
```diff
@@ -68,12 +68,16 @@ public static Optional<String> getInstanceId(X509Certificate cert) {
...
-        return getAthenzUniqueInstanceIdFromSanUri(sans)
+        return getLastSegmentFromSanUri(sans, "athenz://instanceid/")
                 .or(() -> getAthenzUniqueInstanceIdFromSanDns(sans));
     }
 
-    private static Optional<String> getAthenzUniqueInstanceIdFromSanUri(List<SubjectAlternativeName> sans) {
-        String uriPrefix = "athenz://instanceid/";
+    /** @return Athenz unique instance name from the Subject Alternative Name extension */
+    public static Optional<String> getInstanceName(List<SubjectAlternativeName> sans) {
+        return getLastSegmentFromSanUri(sans, "athenz://instancename/");
```
hakonhall, line 77: The returned value has the format project/instancename - so return a record with those fields?

hakonhall (approved):

vespa-athenz/src/main/java/com/yahoo/vespa/athenz/tls/AthenzX509CertificateUtils.java
freva: The returned value is only the instance name, since it returns the last path segment

PR #31240, opened 2024-05-18
- Carry datatype explicit instead of implicit via field inheritance.

baldersheim:
@toregge PR


toregge (dismissed):

baldersheim:
@toregge There are still a few usages of getDataType sprinkled around. Are they easy to get rid of ?

toregge (dismissed):

toregge:
It's a mixed situation, some of the calls to getDataType() should be easy to remove, while others might require checking the summary field source instead (e.g. `Bolding` processor).  Vsm commands such as `FLATTENJUNIPER` and `FLATTENSPACE` (where the summary field typically have multiple sources) adds more complexity. Multiple sources can mean two different things (flattening or set of nested fields that should be in summary) in streaming mode.

baldersheim:
Then we need to handle them obe by one then.

PR #31241, opened 2024-05-19
Hakonhall/quota capacity policies

hakonhall:


bratseth (dismissed):

hakonhall:
```
        Failed to set up first component graph
        exception=
        java.lang.RuntimeException: When resolving dependencies of 'NodeRepository' of type 'com.yahoo.vespa.hosted.provision.NodeRepository'
        Caused by: java.lang.IllegalStateException: No global component of class com.yahoo.config.provision.Exclusivity to inject into component 'NodeRepository' of type 'com.yahoo.vespa.hosted.provision.NodeRepository'.
```

hakonhall:
working on a fix

PR #31242, opened 2024-05-19
Avoid injecting Exclusivity

hakonhall:
Fixes

```
        Failed to set up first component graph
        exception=
        java.lang.RuntimeException: When resolving dependencies of 'NodeRepository' of type 'com.yahoo.vespa.hosted.provision.NodeRepository'
        Caused by: java.lang.IllegalStateException: No global component of class com.yahoo.config.provision.Exclusivity to inject into component 'NodeRepository' of type 'com.yahoo.vespa.hosted.provision.NodeRepository'.
```

Introduced with https://github.com/vespa-engine/vespa/pull/31241.


bratseth (approved):
Thanks. I always forget we don't support transitive construction because I want it so much 😥

PR #31248, opened 2024-05-19
Use pointer to vespalib::eval::Value.

toregge:
@baldersheim : please review


searchcore/src/vespa/searchcore/proton/common/attribute_updater.cpp
```diff
@@ -503,7 +503,7 @@ void
 AttributeUpdater::updateValue(TensorAttribute &vec, uint32_t lid, const FieldValue &val)
 {
     validate_field_value_type(FieldValue::Type::TENSOR, val, "TensorAttribute", "TensorFieldValue");
-    const auto &tensor = static_cast<const TensorFieldValue &>(val).getAsTensorPtr();
+    const auto tensor = static_cast<const TensorFieldValue &>(val).getAsTensorPtr();
```
baldersheim, line 506: I think best practice is to actually use **const auto *** so it is easier to see that it is a pointer.
I know that atleast @vekterli favors that.

baldersheim (approved):

PR #31255, opened 2024-05-21
Improve error message when visiting with wrong document type

jonmv:

Instead of complaining about missing a bucket space mapping, we instead tell the user that the document type was not found in the cluster, which is the actual cause.


vekterli (approved):
:+1:

PR #31279, opened 2024-05-23
Dont pollute the ccache for sanitizer builds.

aressem:


.buildkite/bootstrap-cmake.sh
```diff
@@ -7,10 +7,12 @@ source /etc/profile.d/enable-gcc-toolset.sh
 VESPA_CMAKE_SANITIZERS_OPTION=""
 if [[ $VESPA_USE_SANITIZER != null ]]; then
     VESPA_CMAKE_SANITIZERS_OPTION="-DVESPA_USE_SANITIZER=$VESPA_USE_SANITIZER"
+    VESPA_CMAKE_CCACHE_OPTION="-DVESPA_USE_CCACHE=true"
```
hmusum, line 11: I would have thought this should be false, but maybe I am misunderstanding how the flag is used?

aressem: This should be false. Thanks 👍 Fixed in new commit.

hmusum (approved):

PR #31280, opened 2024-05-23
Separate out admin-only hosts in parent-host-not-ready message

jonmv:
@hakonhall


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/provisioning/Activator.java
```diff
@@ -169,13 +171,24 @@ private static void validateParentHosts(ApplicationId application, NodeList allN
                                              .matching(node -> parentHostnames.contains(node.hostname()))
                                              .hostnames();
 
+        Set<String> applicationParentHostnames = new HashSet<>(parentHostnames);
+        applicationParentHostnames.removeIf(host -> allNodes.childrenOf(host).type(Type.combined, Type.container, Type.content).isEmpty());
```
jonmv, line 175: Hmm, should probably check for children among `potentialChildren`, only.

hakonhall (approved):

PR #31282, opened 2024-05-23
- Wire in a common shared heap for WeakAnd.

baldersheim:
@havardpe PR
This wires teh MatchParams to the WeadAndSearch and prepares the tests, but does not use it yet.
It also deinlines some test code.
Should have no semantic meaning.
All to make the essential PR small.


baldersheim:
@geirst PR

havardpe (approved):

PR #31285, opened 2024-05-23
Jonmv/support new tester nodes syntax

jonmv:
@freva


freva (dismissed):

freva:
```
[ERROR] Class com.yahoo.text.XML: Extra method public static java.lang.String toString(org.w3c.dom.Node)
```

freva (approved):

PR #31298, opened 2024-05-24
Include cost of all distributed ONNX models, not just those with cust…

jonmv:
…om config

@hmusum please review.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
@@ -815,8 +815,9 @@ public class ContainerModelBuilder extends ConfigModelBuilder<ContainerModel> {
                                                                                         !container.getHostResource().realResources().gpuResources().isZero());
                 onnxModel.setGpuDevice(gpuDevice, hasGpu);
             }
-            cluster.onnxModelCostCalculator().registerModel(context.getApplicationPackage().getFile(onnxModel.getFilePath()), onnxModel.onnxModelOptions());
         }
+        for (OnnxModel onnxModel : models.asMap().values())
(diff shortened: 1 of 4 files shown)


baldersheim (dismissed):

baldersheim:
Is this easily testable in unit test ?

jonmv:
I'm hoping it is. Trying to find out now.

baldersheim:
@bjorncs PR

hmusum:
Not sure of I understand the changes that will follow from this. Consider this scenario for an app:

1. Cluster 1 configures model A and B, model A with custom config
2. Cluster 2 configures model C and D, model C with custom config 

Previously we would have a cost for cluster 1 based on model A and another cost for cluster 2 based on model C.
This might miss cost for model B in cluster 1 and cost for model D in cluster 2.

With the new code I assume we will use the cost for A+B+C+D for both cluster 1 and cluster 2? Or just A+B for cluster 1 and C+D for cluster 2?

I agree that some testing would be nice understand the implications.

jonmv:
I'm not sure what you mean with clusters configuring only models A, B, vs C, D. I couldn't find any indication that this is cluster-dependent—as far as I can see, all clusters will always load all four models. Or is this done lazily?

bjorncs:
I'm not sure whether it's lazy loaded or at component graph construction time.

jonmv:
It looks to me like it's construction time. But I guess that doesn't matter much, anyway. The point is these models are all available on all clusters that declare `<model-evaluation>`, AFAICS, and we should therefore factor in their cost for all clusters too ... right?

baldersheim:
If all are loaded all cost should be included.

jonmv:
They currently are.

jonmv:
I tested by deploying an app with some small models, and `<model-evaluation>`, then deploying again, the only change being that I added lots on unused .onnx models to `/models`, and the memory util went up correspondingly.

PR #31305, opened 2024-05-24
Handle fulfilment precision loss

mpolden:
@bratseth @bjormel


bjormel (approved):

bjormel:
@mpolden should we log some more debug data on this to try to identify where the calculation goes wrong?
https://github.com/vespa-engine/vespa/pull/31308

bratseth (approved):

PR #31324, opened 2024-05-28
- Separate priority Q functionality into a separate base class WeakAn…

baldersheim:
…dPriorityQueue.

- Keep thread safety in the SharedWeakAndpriorityQueue.
- Only use thread safe variant if more than 1 thread in the search bundle.

@havardpe @geirst PR

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/blueprintbuilder.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/blueprintbuilder.cpp
@@ -67,7 +67,8 @@ private:
     void buildIntermediate(IntermediateBlueprint *b, NodeType &n) __attribute__((noinline));
 
     void buildWeakAnd(ProtonWeakAnd &n) {
-        auto *wand = new WeakAndBlueprint(n.getTargetNumHits(), _requestContext.get_attribute_blueprint_params().weakand_range);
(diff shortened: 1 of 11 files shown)


searchcore/src/vespa/searchcore/proton/matching/blueprintbuilder.cpp
```diff
@@ -67,7 +67,8 @@ class BlueprintBuilderVisitor :
     void buildIntermediate(IntermediateBlueprint *b, NodeType &n) __attribute__((noinline));
 
     void buildWeakAnd(ProtonWeakAnd &n) {
-        auto *wand = new WeakAndBlueprint(n.getTargetNumHits(), _requestContext.get_attribute_blueprint_params().weakand_range);
+        auto *wand = new WeakAndBlueprint(n.getTargetNumHits(), _requestContext.get_attribute_blueprint_params().weakand_range,
+                                          _requestContext.thread_bundle().size() > 1);
```
havardpe, line 71: should use use is_search_multi_threaded here as well

havardpe (dismissed):
looks good

searchcore/src/vespa/searchcore/proton/matching/blueprintbuilder.cpp
baldersheim: CreateBlueprintVisitorHelper was not used here, but made similar method to unify code.

PR #31333, opened 2024-05-29
Add test showing that inheriting a non-existent document summary give…

hmusum:
…s NullPointerException

Today we give a warning and plan to fix this in Vespa 9. This test shows that we might as well throw as we get a NullPointerException and config server returns and internal server error when this happens, which is worse.

@baldersheim the fix is easy, but it is marked for change in Vespa 9, any objections to changing it now?

### DIFF
--- a/config-model/src/test/java/com/yahoo/schema/processing/SummaryConsistencyTestCase.java
+++ b/config-model/src/test/java/com/yahoo/schema/processing/SummaryConsistencyTestCase.java
@@ -5,6 +5,7 @@ import com.yahoo.schema.Schema;
 import com.yahoo.schema.ApplicationBuilder;
 import com.yahoo.schema.parser.ParseException;
 import com.yahoo.vespa.documentmodel.SummaryTransform;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import static com.yahoo.config.model.test.TestUtil.joinLines;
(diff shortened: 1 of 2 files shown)


hmusum:
I now see that this statement will fail if we do this:
`    document-summary content-summary inherits default {
`

baldersheim (approved):

config-model/src/test/java/com/yahoo/schema/processing/SummaryConsistencyTestCase.java
```diff
@@ -42,4 +43,26 @@ void attribute_combiner_transform_is_set_when_source_is_array_of_struct_with_onl
         Schema schema = ApplicationBuilder.createFromString(sd).getSchema();
         assertEquals(SummaryTransform.ATTRIBUTECOMBINER, schema.getSummaryField("elem_array_unfiltered").getTransform());
     }
+
+    @Test
+    @Disabled
+    void testDocumentTypesWithInheritanceOfNonExistingField() throws ParseException {
```
geirst, line 49: Should this be `testDocumentSummaryWithInheritanceOfNonExistingSummary` ?

hmusum: Yes, thanks, will fix in next PR

PR #31339, opened 2024-05-29
Throw if inheriting non-existent document summary

hmusum:
Special handling of inheriting default, which I assume will always be ok

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/documentmodel/DocumentSummary.java
+++ b/config-model/src/main/java/com/yahoo/vespa/documentmodel/DocumentSummary.java
@@ -1,7 +1,6 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.documentmodel;
 
-import com.yahoo.config.application.api.DeployLogger;
 import com.yahoo.schema.Schema;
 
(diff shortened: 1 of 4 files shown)


config-model/src/main/java/com/yahoo/vespa/documentmodel/DocumentSummary.java
```diff
@@ -118,14 +116,11 @@ public String toString() {
...
-    public void validate(DeployLogger logger) {
+    public void validate() {
         for (var inheritedName : inherited) {
             var inheritedSummary = owner.getSummary(inheritedName);
-            if (inheritedSummary == null) {
-                // TODO Vespa 9: Throw IllegalArgumentException instead
-                logger.logApplicationPackage(Level.WARNING,
-                                             this + " inherits '" + inheritedName + "' but this is not present in " + owner);
-            }
+            if (inheritedSummary == null && ! inheritedName.equals("default"))
```
baldersheim, line 122: Inheriting 'default' does not make sense. I suggest logging a warning for now. Then we we might be able to remove the if later on.

hmusum: Sounds like a good idea, done

baldersheim (approved):

PR #31340, opened 2024-05-29
Rewrite match loop communicator unit test to gtest.

toregge:
@geirst : please review


geirst:
I am not sure it makes sense to rewrite this as multi-threaded tests with `TEST_MT_F()` is simpler with the old test framework. @havardpe please take a look.

havardpe (approved):

PR #31342, opened 2024-05-29
Add hidden RerankRescorer class and use it to get second phase scores

toregge:
into the result set earlier.

@geirst : please review


geirst (approved):
👍

PR #31355, opened 2024-05-30
Avoid buffering entire prometheus model serialisation

jonmv:
@baldersheim or @gjoranv please review.


baldersheim (dismissed):
Check if the same should be done for the other ones too.

gjoranv (dismissed):

baldersheim (approved):

PR #31358, opened 2024-05-30
Add an opportunistic weakand searcher that will first replace the wea…

baldersheim:
…kAnd with AND, and fall back to original plan if not enough hits.

@bjorncs PR

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/searchers/OpportunisticWeakAndSearcher.java
+++ b/container-search/src/main/java/com/yahoo/search/searchers/OpportunisticWeakAndSearcher.java
@@ -1,0 +1,83 @@
+package com.yahoo.search.searchers;
+
+import com.yahoo.api.annotations.Beta;
+import com.yahoo.component.chain.dependencies.After;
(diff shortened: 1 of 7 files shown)


bjorncs (approved):

container-search/src/main/java/com/yahoo/search/searchers/OpportunisticWeakAndSearcher.java
```diff
@@ -0,0 +1,83 @@
+package com.yahoo.search.searchers;
```
geirst, line 1: Please add copyright.

container-search/src/test/java/com/yahoo/search/searchers/OpportunisticWeakAndSearcherTestCase.java
```diff
@@ -0,0 +1,40 @@
+package com.yahoo.search.searchers;
```
geirst, line 1: Please add copyright.

bratseth:
👍

PR #31360, opened 2024-05-30
Log if getting exception when downloading app-package

hmusum:
Setting log level to info for now, not sure what kind of issues this might catch, I'll consider changing to warning if/when I see one of these

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
@@ -94,7 +94,10 @@ public class ApplicationPackageMaintainer extends ConfigServerMaintainer {
                                     return;
                                 }
                             }
-                            catch (Exception ignored) { }
+                            catch (Exception e) {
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
```diff
@@ -94,7 +94,10 @@ protected double maintain() {
                                     return;
                                 }
                             }
-                            catch (Exception ignored) { }
+                            catch (Exception e) {
+                                log.info("Exception when downloading application package (" + appFileReference + ")" +
```
baldersheim, line 98: No success is already logged at info level at line 102.
I would just make this a warning right away as that stands out more and is easier to detect.

baldersheim (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
hmusum: Makes sense, done, need a new review

baldersheim (approved):

PR #31375, opened 2024-05-31
Implement second phase rank drop limit for hit collector.

toregge:
@geirst : please review


toregge:
The default value should always be dropped when second phase rank drop limit is set.

PR #31379, opened 2024-05-31
Match XML tags directive only with environment match (or implicit prod)

jonmv:
@bratseth


jonmv:
I had a look at the usual suspects, and this won't break any of them.

bratseth (approved):
Nice

PR #31399, opened 2024-06-03
Do a proper sync with thread that shuts down old cluster

jonmv:
@hmusum please review and merge


hmusum (approved):

aressem:
@jonmv This might have caused instability for system tests with different tests timing out. Could be worth to try a revert if you think the same.

PR #31403, opened 2024-06-03
Extend index properties api to cover second phase rank score drop limit.

toregge:
@geirst : please review


geirst (approved):
👍

searchlib/src/vespa/searchlib/fef/ranksetup.h
```diff
@@ -60,6 +61,7 @@ class RankSetup
     double                   _degradationSamplePercentage;
     double                   _degradationPostFilterMultiplier;
     feature_t                _rankScoreDropLimit;
+    std::optional<double>    _second_phase_rank_score_drop_limit;
```
baldersheim, line 64: I think it would be better to to just solve this the same way as with the rankScoreDropLimit in first-phase, without std::optional. To different ways of solving the same problem is not ideal.

PR #31404, opened 2024-06-03
Add feature flag for enforcing strictly increasing cluster state versions

vekterli:
@baldersheim please review.

Applies to `stor-server` config for both distributors and storage nodes.

Defaults to `false` for now.


baldersheim (approved):

vekterli:
PR build froze and timed out, should be entirely unrelated.

PR #31406, opened 2024-06-03
Use targetHits from WeakAnd directly. Also add tracing of fallback to…

baldersheim:
… weakAnd

@jonmv or @bjorncs PR


bratseth (approved):

bjorncs (commented):
👍

PR #31415, opened 2024-06-04
Revert "Do a proper sync with thread that shuts down old cluster"

baldersheim:
Reverts vespa-engine/vespa#31399
Suspected test instability during shutdown of feed-client
@aressem or @jonmv


aressem (approved):

jonmv:
Yes, probably this. I'm looking into it.

PR #31421, opened 2024-06-04
Rename existing rank drop limit and rank score drop limit to

toregge:
first phase rank score drop limit.
Stop using std::isnan to mark no value.

@geirst : please review


baldersheim (approved):

geirst (commented):
👍

PR #31446, opened 2024-06-05
Increase container ideal memory load to 0.9

bratseth:
Vespa will now attempt to allocate memory to this level. If we don't allow it, all allocations will be unfulfiled, since adding more resources won't decrease used memory per container.

@bjormel is 0.8 -> 0.9 enough?


bjormel (commented):
👍

PR #31448, opened 2024-06-05
Preprocess application package in temp dir, then move atomically to d…

hmusum:
…estination.

### DIFF
--- a/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
+++ b/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
@@ -52,6 +52,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.Reader;
 import java.io.StringReader;
+import java.nio.file.AccessDeniedException;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
```diff
@@ -613,24 +614,38 @@ private void preprocessXML(File destination, File inputXml, Zone zone) throws IO
...
+            // when shutting down.
+            // Temp directory needs to be on the same file system as appDir for 'move' to work,
+            // if it fails (with DirectoryNotEmptyException (!)) we need to use 'copy' instead
+            // (this will always be the case for the application package for a standalone container).
+            Files.move(tempDir, preprocessedDir.toPath());
+        } catch (AccessDeniedException e) {
+            preprocess(appDir, preprocessedDir, zone);
+        }
+        FilesApplicationPackage preprocessedApp = fromFile(preprocessedDir, includeSourceFiles);
+        preprocessedApp.copyUserDefsIntoApplication();
```
jonmv, line 631: Should this also be inside the atomic move?

jonmv (approved):

PR #31452, opened 2024-06-05
Use a short default wait for cloud deployment

mpolden:
@jonmv

### DIFF
--- a/client/go/internal/cli/cmd/deploy.go
+++ b/client/go/internal/cli/cmd/deploy.go
@@ -59,7 +59,6 @@ $ vespa deploy -t cloud -z perf.aws-us-east-1c`,
 			if err != nil {
 				return err
 			}
-			timeout := time.Duration(waitSecs) * time.Second
 			opts := vespa.DeploymentOptions{ApplicationPackage: pkg, Target: target}
 			if versionArg != "" {
 				version, err := version.Parse(versionArg)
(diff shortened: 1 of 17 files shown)


client/go/internal/cli/cmd/deploy.go
```diff
@@ -73,6 +72,12 @@ $ vespa deploy -t cloud -z perf.aws-us-east-1c`,
 					return err
 				}
 			}
+			if opts.Target.IsCloud() && waitSecs == 0 && !cmd.PersistentFlags().Changed("wait") {
+				// If --wait is not explicitly given, we always wait a few seconds in Cloud to catch fast failures, e.g.
+				// invalid application package
+				waitSecs = 2
```
jonmv, line 78: 2sec is fine for small packages, but probably not for larger stuff. But I guess we want people to use a larger --wait?

jonmv (dismissed):

client/go/internal/cli/cmd/deploy.go
jonmv: Is 10s too long? Probably ...

mpolden: Thanks! The suggestion was to add a short wait that catches fast errors in most cases. However, it shouldn't fail if this short wait elapses. I'll fix that.

mpolden:
PTAL

jonmv:
👍

PR #31453, opened 2024-06-05
Revert "Preprocess application package in temp dir, then move atomically to d…"

hmusum:
Reverts vespa-engine/vespa#31448

Please review only, merge in case this fails


jonmv (approved):
🤦

PR #31456, opened 2024-06-05
Bjorncs/content in access log

bjorncs:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
@@ -42,7 +42,7 @@ public class HostedSslConnectorFactory extends ConnectorFactory {
         this.remotePortHeaders = List.copyOf(builder.remotePortHeaders);
         this.knownServerNames = Collections.unmodifiableSet(new TreeSet<>(builder.knownServerNames));
         builder.requestPrefixForLoggingContent.forEach(prefix -> {
-            var regex = "^.*:[0|1](\\.\\d+)?:\\d+[a-zA-Z]+$";
+            var regex = "^.*:[01](\\.\\d+)?:\\d+[a-zA-Z]+$";
             if (!prefix.matches(regex))
                 throw new IllegalArgumentException("Invalid prefix '%s, must match regex '%s'".formatted(prefix, regex));
         });
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
@@ -42,7 +42,7 @@ public class HostedSslConnectorFactory extends ConnectorFactory {
         this.remotePortHeaders = List.copyOf(builder.remotePortHeaders);
         this.knownServerNames = Collections.unmodifiableSet(new TreeSet<>(builder.knownServerNames));
         builder.requestPrefixForLoggingContent.forEach(prefix -> {
-            var regex = "^.*:[0|1](\\.\\d+)?$";
+            var regex = "^.*:[0|1](\\.\\d+)?:\\d+[a-zA-Z]+$";
             if (!prefix.matches(regex))
                 throw new IllegalArgumentException("Invalid prefix '%s, must match regex '%s'".formatted(prefix, regex));
         });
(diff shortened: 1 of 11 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
```diff
@@ -42,7 +42,7 @@ private HostedSslConnectorFactory(Builder builder) {
         this.remotePortHeaders = List.copyOf(builder.remotePortHeaders);
         this.knownServerNames = Collections.unmodifiableSet(new TreeSet<>(builder.knownServerNames));
         builder.requestPrefixForLoggingContent.forEach(prefix -> {
-            var regex = "^.*:[0|1](\\.\\d+)?$";
+            var regex = "^.*:[0|1](\\.\\d+)?:\\d+[a-zA-Z]+$";
```
jonmv, line 45:
```suggestion
            var regex = "^.*:[01](\\.\\d+)?:\\d+[a-zA-Z]+$";
```
This should be changed the other place too, from the previous PR!

jonmv (dismissed):
Please fix regex, otherwise LGTM!

config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
jonmv: To ensure there are exactly _three_ parts, as expected in later code, perhaps also replace the leading `.*` with `[^:]*`.

vespajlib/src/main/java/ai/vespa/utils/BytesQuantity.java
```diff
@@ -0,0 +1,89 @@
...
+        if (!matcher.matches())
+            throw new IllegalArgumentException(
+                    "Bytes quantity '%s' does not match pattern '%s'".formatted(value, PATTERN.pattern()));
+        var digits = Long.parseLong(matcher.group("digits"));
+        var unit = Unit.fromString(matcher.group("unit"));
+        return BytesQuantity.of(digits, unit);
+    }
+
+    public String asPrettyString() {
+        long remaining = bytes;
```
jonmv, line 73:
```suggestion
        if (bytes == 1) return "1 byte";
        long remaining = bytes;
```

jonmv: But to parse this requires more thought >_<

config-model/src/main/java/com/yahoo/vespa/model/container/http/ssl/HostedSslConnectorFactory.java
jonmv: So ... this may fail during preparation, but is set in a feature flag? 
I think the flag should also enforce the pattern, then. Does it?

container-core/src/test/java/com/yahoo/jdisc/http/server/jetty/HttpServerTest.java
```diff
@@ -745,10 +745,10 @@ void requireThatRequestIsTrackedInAccessLog() throws IOException, InterruptedExc
                 new ServerConfig.Builder(),
                 new ConnectorConfig.Builder().accessLog(
                         new ConnectorConfig.AccessLog.Builder()
-                                .contentPathPrefixes("/status:1")
-                                .contentPathPrefixes("/state/v1:0.001")),
+                                .contentPathPrefixes("/search:1:1k")
```
hakonhall, line 748: In these examples the suffix is always a one-character, so the regex part could be replaced with a more strict and self-documenting [kMG]?

hakonhall (dismissed):

bjorncs:
PTAL @hakonhall @jonmv

container-core/src/test/java/com/yahoo/jdisc/http/server/jetty/HttpServerTest.java
bjorncs: See [6467827](https://github.com/vespa-engine/vespa/pull/31456/commits/64678270e5ac9ab3a9f2b6523b4c2fc80999c4d9)

vespajlib/src/main/java/ai/vespa/utils/BytesQuantity.java
bjorncs: See [d681820](https://github.com/vespa-engine/vespa/pull/31456/commits/d6818209d90b407dc1a9bde5cae4269c071eea2d). It's fine that parsing accepts invalid English grammar ;)

hakonhall (approved):

PR #31457, opened 2024-06-05
Catch correct exception

hmusum:


jonmv:
Ouch, didn't see this.

jonmv (approved):

PR #31462, opened 2024-06-05
Reapply "Preprocess application package in temp dir, then move atomically to destination"

hmusum:
Reverts vespa-engine/vespa#31460

### DIFF
--- a/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
+++ b/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
@@ -72,6 +72,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import static com.yahoo.text.Lowercase.toLowerCase;
+import static java.util.logging.Level.INFO;
 
 
 /**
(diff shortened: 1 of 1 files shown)


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
```diff
@@ -625,7 +626,7 @@ public ApplicationPackage preprocess(Zone zone, DeployLogger logger) throws IOEx
             // if it fails (with DirectoryNotEmptyException (!)) we need to use 'copy' instead
             // (this will always be the case for the application package for a standalone container).
             Files.move(tempDir, preprocessedDir.toPath());
-        } catch (DirectoryNotEmptyException e) {
+        } catch (AccessDeniedException | DirectoryNotEmptyException e) {
```
jonmv, line 629: Is it indeterminate what is thrown?

jonmv (approved):

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
hmusum: There are 2 possible failures:
* Creating a temp dir fails => AccessDeniedException
* Temp dir is on another partition/file system than destination, fails when doing `move` => DirectoryNotEmptyException (yes, strange)

The first one happens when preprocessing a standalone container's application package, where we sometimes are allowed to create a temp dir in parent dir and sometimes aren't. If we are allowed, but are on another partition, the second failure will happen. But it might be some other way to make this more understandable/cleaner...

PR #31463, opened 2024-06-06
Remove duplicate tracing of dependencies

bratseth:


container-search/src/main/java/com/yahoo/prelude/query/NumericInItem.java
```diff
@@ -2,21 +2,23 @@
 package com.yahoo.prelude.query;
 
 import com.yahoo.compress.IntegerCompressor;
+import com.yahoo.prelude.query.textualrepresentation.Discloser;
```
baldersheim, line 5: Spurious import ?

baldersheim (dismissed):

baldersheim (approved):

PR #31464, opened 2024-06-06
Bratseth/grouping trace

bratseth:


container-search/src/main/java/com/yahoo/search/searchchain/Execution.java
```diff
@@ -499,16 +499,6 @@ public Result search(Query query) {
         return (Result)super.process(query);
     }
 
-    @Override
-    protected void onInvoking(Request request, Processor processor) {
-        super.onInvoking(request,processor);
```
baldersheim, line 504: Traced elsewhere ..?

baldersheim (changes requested):
abi-spec mismatch

baldersheim:
Rebase after #31463 is merged.

PR #31471, opened 2024-06-06
use flow analysis to calculate estimated seeks and cost

havardpe:
@geirst please review


havardpe:
@geirst please have a look at the new version

geirst (approved):
👍

PR #31476, opened 2024-06-06
Implement clone and disclose in InItems

bratseth:


container-search/src/main/java/com/yahoo/prelude/query/StringInItem.java
```diff
@@ -89,4 +102,11 @@ public int hashCode() {
...
 
-}
+    @Override
+    public StringInItem clone() {
+        StringInItem clone = (StringInItem) super.clone();
+        clone.tokens = new HashSet<>(tokens);
+        return clone;
+    }
+
+}
```
toregge, line 112: Consider adding newline at end of file.

toregge (dismissed):

toregge:
the abi spec needs to be regenerated.

PR #31477, opened 2024-06-07
Log when some ZK nodes are not present in a session, but should be

jonmv:
@hmusum


configserver/src/test/java/com/yahoo/vespa/config/server/session/SessionPreparerTest.java
```diff
@@ -206,6 +206,7 @@ public void require_no_warning_for_overlapping_host_for_same_appid() throws IOEx
         HostRegistry hostValidator = new HostRegistry();
         ApplicationId applicationId = applicationId();
         hostValidator.update(applicationId, List.of("mytesthost"));
+        //createSessionZooKeeperClient().createNewSession(Instant.now());
```
jonmv, line 209:
```suggestion
```

baldersheim (approved):

hmusum (approved):

hmusum (approved):

PR #31479, opened 2024-06-07
Fall back to parent ctime for session creation

jonmv:
@hmusum


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionZooKeeperClient.java
```diff
@@ -383,7 +388,7 @@ public void createNewSession(Instant createTime) {
         CuratorTransaction transaction = new CuratorTransaction(curator);
         transaction.add(CuratorOperations.create(sessionPath.getAbsolute()));
         transaction.add(CuratorOperations.create(sessionPath.append(UPLOAD_BARRIER).getAbsolute()));
-        transaction.add(createWriteStatusTransaction(Session.Status.NEW).operations());
+        transaction.add(CuratorOperations.create(sessionStatusPath.getAbsolute(), Utf8.toBytes(Status.NEW.name())));
```
jonmv, line 391: Just to avoid looking up a path which never exists at this time.

hmusum (approved):
Nice!

PR #31531, opened 2024-06-11
Add support for .vespaignore

mpolden:
This adds support for having a `.vespaignore` file in the application package,
which contains a list of path patterns to exclude from the final
`application.zip`.

We'll do a architect review of this feature, so it's not documented yet. Will do
that after the review.n

@jonmv


jonmv (approved):
Nice.

PR #31538, opened 2024-06-11
Speed up insertion into HNSW index by referencing vector memory in mo…

geirst:
…re cases.

When inserting vectors into a HnswIndex we use DistanceFunctionFactory::for_insertion_vector() to create a distance function each time distance calculations are needed for a candidate vector. All these vectors are of the same type, as given by the tensor type of the TensorAttribute. The lifetime of vectors is also handled by them either being stored in the TensorAttribute, or existing in the document being inserted.

This PR speeds up insertion for vectors of type int8, float and double. This is done by referencing the vector memory directly in the distance function, instead of copying (and transforming) into a tempory memory buffer. Vectors of type bfloat16 are still transformed to float before distance calculations.

@arnej27959 please review
@baldersheim @toregge FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
@@ -13,11 +13,12 @@ using vespalib::eval::Int8Float;
 
 namespace search::tensor {
 
-template<typename FloatType>
+template <typename VectorStoreType>
 class BoundAngularDistance final : public BoundDistanceFunction {
 private:
+    using FloatType = VectorStoreType::FloatType;
     const vespalib::hwaccelrated::IAccelrated & _computer;
-    mutable TemporaryVectorStore<FloatType> _tmpSpace;
+    mutable VectorStoreType _tmpSpace;
     const vespalib::ConstArrayRef<FloatType> _lhs;
     double _lhs_norm_sq;
 public:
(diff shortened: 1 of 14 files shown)


searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
```diff
@@ -66,25 +66,37 @@ class BoundAngularDistance final : public BoundDistanceFunction {
...
 }
 
-template <typename FloatType>
+template <typename FloatType, bool reference_insertion_vector>
 BoundDistanceFunction::UP
-AngularDistanceFunctionFactory<FloatType>::for_insertion_vector(TypedCells lhs) const {
-    using DFT = BoundAngularDistance<FloatType>;
-    return std::make_unique<DFT>(lhs);
+AngularDistanceFunctionFactory<FloatType, reference_insertion_vector>::for_insertion_vector(TypedCells lhs) const {
+    if constexpr (reference_insertion_vector) {
```
baldersheim, line 86: I recommend that you let reference_insertion_vector be a normal argument to the constructor. Then you do not need an extra template argument. Less template arguments and a more local change.

searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
```diff
@@ -13,11 +13,11 @@ using vespalib::eval::Int8Float;
 
 namespace search::tensor {
 
-template<typename FloatType>
+template <typename FloatType, typename VectorStoreType>
```
baldersheim, line 16: Given that cell type in distance function is always the same as the cell type in the VectorStore, if will be sufficient with just the VectorStoreType as a template argument.
Just add a public `using CellType = FloatType;` to each of Temporary and Reference store classes.

baldersheim (dismissed):
I suggest reducing template arguments.

geirst:
@baldersheim PTAL. I have reduced the number of template arguments as discussed.

searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
geirst: Fixed

searchlib/src/vespa/searchlib/tensor/angular_distance.cpp
geirst: Fixed

baldersheim (approved):

baldersheim (approved):

PR #31541, opened 2024-06-12
Revert "Hmusum/remove qrserver remnants"

aressem:
Reverts vespa-engine/vespa#31540

Broke https://factory.vespa.oath.cloud/testrun/87210014/test/Integrationtests::test_subscription


hmusum:
I'll update the test instead

hmusum:
Fixed by updating test in https://github.com/vespa-engine/system-test/pull/4057

PR #31544, opened 2024-06-12
Support setting 'select' in a query profile

bratseth:


vespajlib/src/test/java/com/yahoo/tensor/MixedTensorTestCase.java
```diff
@@ -159,4 +164,25 @@ public void testTwoMappedTwoIndexedBuilding() {
                 tensor.toString());
     }
 
+    @Test
+    public void testSplitIntoDense() {
```
baldersheim, line 168: Unrelated I assume.
Looks like documentation for the code you gave to @andreer  the other day.

baldersheim (approved):
Functionality lost between 7 or 8, or later on ?

vespajlib/src/test/java/com/yahoo/tensor/MixedTensorTestCase.java
bratseth: Yes, decided it was worth keeping + lazy ...

bratseth:
I think later than 7 -> 8 but not sure.

PR #31546, opened 2024-06-12
Add distribution config bundle to `ClusterStateBundle` in C++

vekterli:
@geirst please review. 

This is part of prerequisite work needed before we can start pushing distribution config from the cluster controller and rewiring how distribution config is propagated and used in the backends.

Move distribution config transforms to `vdslib` so that the distribution config bundle can contain derived configs for all bucket spaces in one central place.

Add a content-node internal wrapper around state+config that is exposed by the bucket space as one atomic unit rather than fetching the two separately and thus non-atomically.

Also, rename `Distribution::serialize()` to `Distribution::serialized()` since it returns a const ref to a cached serialized form and does not do on-demand serialization.


vdslib/src/vespa/vdslib/distribution/distribution_config_bundle.h
```diff
@@ -0,0 +1,48 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include "bucket_space_distribution_configs.h"
+#include "distribution.h"
+#include <iosfwd>
+
+namespace storage::lib {
+
+class DistributionConfigBundle {
```
geirst, line 10: Consider adding class comment.

geirst (approved):
👍

vdslib/src/vespa/vdslib/distribution/distribution_config_bundle.h
vekterli: Will be done as part of a follow-up; this code is currently in motion 🏃‍➡️

PR #31562, opened 2024-06-13
Use readonly distance heap in exact nearest neighbor iterators during

toregge:
later matching phases.

@geirst : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/queryeval/exact_nearest_neighbor_iterator.cpp
+++ b/searchlib/src/vespa/searchlib/queryeval/exact_nearest_neighbor_iterator.cpp
@@ -19,13 +19,14 @@ namespace search::queryeval {
  * Keeps a heap of the K best hit distances.
  * Currently always does brute-force scanning, which is very expensive.
  **/
-template <bool strict, bool has_filter, bool readonly_distance_heap, bool has_single_subspace>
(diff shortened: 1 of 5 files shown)


geirst (dismissed):
👍

searchlib/src/vespa/searchlib/queryeval/exact_nearest_neighbor_iterator.cpp
```diff
@@ -19,7 +19,7 @@ namespace search::queryeval {
  * Keeps a heap of the K best hit distances.
  * Currently always does brute-force scanning, which is very expensive.
  **/
-template <bool strict, bool has_filter, bool has_single_subspace>
+template <bool strict, bool has_filter, bool readonly_distance_heap, bool has_single_subspace>
```
baldersheim, line 22: No need to readonly_distance_heap as a template argument. It is only used in unpack, which a very cold method.

geirst (approved):
👍

baldersheim (approved):

searchlib/src/vespa/searchlib/queryeval/exact_nearest_neighbor_iterator.cpp
baldersheim: Perfect

baldersheim (approved):

PR #31569, opened 2024-06-14
Accept distribution config from cluster controller on distributor

vekterli:
@geirst please review

If a received cluster state bundle contains an embedded distribution config, the distributor will act _as if_ it had atomically received and processed a distribution config change followed by a new cluster state, but where no bucket info requests are sent for the config change.

If a distributor observes a state bundle containing distribution config it will note this internally and explicitly ignore any further received distribution configs _not_ arriving from the cluster controller. To handle downgrades and rollbacks, it undoes this toggle if it later observes a state bundle _without_ distribution config, reverting to using internal node config instead.

Note: the `DistributionConfigBundle::clone_with_new_distribution()` functionality is added but not used here—it is expected to be used later for the content node work (or I'll remove it).

### DIFF
--- a/config/src/vespa/config/print/asciiconfigreader.h
+++ b/config/src/vespa/config/print/asciiconfigreader.h
@@ -3,6 +3,8 @@
 
 #include "configreader.h"
 
+namespace vespalib { class asciistream; }
+
 namespace config {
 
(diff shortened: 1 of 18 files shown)


geirst (dismissed):
👍

PR #31579, opened 2024-06-16
Adjust include statement to handle out of source builds.

toregge:
@baldersheim : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/engine/search_protocol_proto.h
+++ b/searchlib/src/vespa/searchlib/engine/search_protocol_proto.h
@@ -2,4 +2,4 @@
 
 #pragma once
 
-#include "search_protocol.pb.h"
+#include <vespa/searchlib/engine/search_protocol.pb.h>


baldersheim (approved):
Is teher something special with the generated protobuf files that makes this necessary ?

toregge:
The generated protobuf files are written to the build directory which is different from the source directory for out of source builds. 

Two alternatives for out of source build:

1.  Adjust include statement, as in [logd](https://github.com/vespa-engine/vespa/blob/430d03c2459569b45e2a167b86796d02fbc07c09/logd/src/logd/log_protocol_proto.h#L5) and [storage](https://github.com/vespa-engine/vespa/blob/430d03c2459569b45e2a167b86796d02fbc07c09/storage/src/vespa/storage/storageserver/rpc/rpc_envelope_proto.h#L5) and 8 other locations.
2. Adjust cmake setup to add more include directories as in [document](https://github.com/vespa-engine/vespa/blob/430d03c2459569b45e2a167b86796d02fbc07c09/document/src/vespa/document/select/CMakeLists.txt#L14) and more generally in [cmake setup for generated config files](https://github.com/vespa-engine/vespa/blob/430d03c2459569b45e2a167b86796d02fbc07c09/functions.cmake#L160) (which only works for unit tests and applications due to the property being private).

PR #31588, opened 2024-06-17
Remove ancient, non-functional distributor init node state from tests

vekterli:
@geirst please review.

This has not been anything more than a no-op at best, as there is no (and to the best of my knowledge—never has been) an init state for distributors.

🦖🧹💨


geirst (approved):
👍

PR #31596, opened 2024-06-17
allow profiling weak and children

havardpe:
- new decoration strategy: transform_children
- track child reordering (to make profile results match the structure of the blueprint tree)

@geirst please review


geirst (approved):
👍

PR #31602, opened 2024-06-18
Support non-cloud target in vespa log

mpolden:
@jonmv

FYI @kkraune: Next version of CLI will support `vespa log` when using
self-hosted Vespa >= 8.359.0.


jonmv (approved):
Wohoo!

PR #31605, opened 2024-06-18
only use transform_children for profiling

havardpe:
also use it for multi-bit-vector source blender optimization

@geirst please review
@baldersheim FYI


geirst (approved):
👍

PR #31626, opened 2024-06-19
Make sure to always delete temp dir

hmusum:


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
```diff
@@ -626,8 +627,13 @@ public ApplicationPackage preprocess(Zone zone, DeployLogger logger) throws IOEx
             // if it fails (with DirectoryNotEmptyException (!)) we need to use 'copy' instead
             // (this will always be the case for the application package for a standalone container).
             Files.move(tempDir, preprocessedDir.toPath());
+            tempDir = null;
         } catch (AccessDeniedException | DirectoryNotEmptyException e) {
             preprocess(appDir, preprocessedDir, zone);
+            IOUtils.recursiveDeleteDir(tempDir.toFile());
```
baldersheim, line 633: Should you remove this one and always do the cleanup in the finally clause, if not I guess you should set tempDir to null here too.

hakonhall, line 633: Actually the deletion will naturally be done in the finally, so this statement can be removed.

hakonhall (dismissed):

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
hmusum: Yes, I didn't intend this one to be there, mixed up when editing, will fix

hmusum: See comment below

hakonhall (approved):

baldersheim (approved):

PR #31658, opened 2024-06-20
Handle cluster state bundle with distribution config on content node

vekterli:
@geirst please review 🎣

If a cluster state bundle contains distribution config, this is internally propagated via the `StateManager` component to all registered state listeners. One such state listener is `FileStorManager`, which updates the content node-internal bucket space repository.

All `SetSystemStateCommand` and internal config-aware components (`StateManager` and `ChangedBucketOwnershipHandler`) now explicitly track whether the cluster controller provides distribution config, or if the internally provided config should be used (including fallback to internal config if necessary). This logic duplication is not optimal, but will go away once the only way to get distribution config is via the cluster controller.


baldersheim:
@vekterli merge conflict ....

geirst (dismissed):
👍

PR #31659, opened 2024-06-20
Consider port differences between versions when getting log

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/ApplicationRepository.java
```diff
@@ -1169,33 +1177,40 @@ private void logConfigChangeActions(ConfigChangeActions actions, DeployLogger lo
...
-        int port = servicePort(logService);
-        return HttpURL.create(Scheme.http, DomainName.of(logServerHostInfo.getHostname()), port, HttpURL.Path.parse("logs"));
+                .map(hostInfo -> hostInfo.getServices().stream()
+                        .filter(service -> LOGSERVER_CONTAINER.serviceName.equals(service.getServiceType()))
+                        .findFirst()
+                        .or(() -> hostInfo.getServices().stream()
+                                .filter(service -> CONTAINER.serviceName.equals(service.getServiceType()))
+                                .findFirst())
+                        .orElseThrow(() -> new IllegalArgumentException("No container running on logserver host")))
+                .collect(Collectors.toMap(ServiceInfo::getHostName, this::servicePort, (a, b) -> a));
```
jonmv, line 1213: I think you actually need `groupingBy` here?

jonmv: But then it's simpler to just return a list of pairs.

hmusum: Right, there _could_ be different hostnames, I guess. Will fix

hmusum:

jonmv (approved):
Nice!

Issue #31433, opened 2024-06-04
Bug with vespa when using a linear combination score of two embeddings and varying hits=k?


drei34:
Hello,

I am seeing some behavior I can't explain. We have a query which tries to pull relevant documents from a DB given two embeddings per document (and at runtime, we get two embeddings for this query). To do this, we use a rank profile which uses a weighted average (50/50 say) and there is an OrItem and disjunction in the Java code. However, when I vary hits=10 or hits=1000 and use this rank profile I see complete different document ids being retrieved. We can't explain this. Have people seen this before and is this something with an easy explanation? If not I can post more, but I'd probably need to create an example from MS Marco etc since this is propriety data. Does hits affect the set of initial documents retrieved (then the results makes sense)? However, the documentation makes me think hits just limits the output length but it seems to be doing more.

jobergum:
Hey, would appreciate some more details indeed. I assume that you are doing

```
where ({targetHits:100}nearestNeighbor(x, q)) or ({targetHits:100}nearestNeighbor(y,q1))
```

and that you combine the two scores in a ranking profile? In any case, the [hits](https://docs.vespa.ai/en/reference/query-api-reference.html#hits) parameter has no functional impact on the retrieval or ranking in the above example, where we expect about 200 (or more) documents to be exposed to ranking per content node . That said, hits can have impact on serving performance and which can increase the chance of triggering graceful coverage degradation where vespa terminates early.  So please check that coverage is 100%, see [result coverage degradation](https://docs.vespa.ai/en/graceful-degradation.html).

drei34:
Hey! So we are using HNSW and I am not changing the targetHits parameter; the default is 100 so I assume this stays the same. The issue is when we change hits=10 or hits=100 or hits=1000 we see a different set of documents in the top 10. The query looks basically like below if this helps. If this cannot be explained, I can add more:

SELECT * FROM XYZ WHERE (data contains 'ABC' AND userQuery())&query="Some documents about trees"&type=any&hits=10&retriever=hybrid&restrictSchema=XYZ;

We embed "Some documents about trees" using two different models. We use the rank profile "hybrid" which is a 50/50 weighted sum cosine similarity from two transformers. However, the thing we can't explain is this behavior? Should the results for hits =10, hits=100 and hits=1000 not be contained within each other? Is hits=1000 > targetHits=100 causing the problem? HNSW? Have you see this before?

jobergum:
[...]

PR #31664, opened 2024-06-21
Balder/targethits equal max targethits and hits

baldersheim:
@geirst and @bratseth PR
This relates to #31433


bratseth (dismissed):
👍 👍

bratseth:
Need ABI spec update due to `isNExplicit`.
Btw, I would suggest calling it `nIsExplicit`

baldersheim:
Fixed

geirst (commented):
👍

PR #31673, opened 2024-06-21
Lower log level

hmusum:
Avoid noise from this when nodes have not yet started, this is not critical, so use INFO level.

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
@@ -76,7 +76,7 @@ public class NodeMetricsDbMaintainer extends NodeRepositoryMaintainer {
                                 ApplicationId application) {
         if (exception != null) {
             if (failures.get() < maxWarningsPerInvocation)
-                log.log(Level.WARNING, "Could not update metrics for " + application + ": " +
+                log.log(Level.INFO, "Could not update metrics for " + application + ": " +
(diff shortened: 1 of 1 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
```diff
@@ -76,7 +76,7 @@ private void handleResponse(MetricsResponse response,
                                 ApplicationId application) {
         if (exception != null) {
             if (failures.get() < maxWarningsPerInvocation)
-                log.log(Level.WARNING, "Could not update metrics for " + application + ": " +
+                log.log(Level.INFO, "Could not update metrics for " + application + ": " +
```
baldersheim, line 79: What is maxWarningPerInvocation ... should it be reduced.

baldersheim (dismissed):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/NodeMetricsDbMaintainer.java
hmusum: Hm, yes, good point, I think we could just remove it when lowering log level

hmusum: It is set to 2 now, so I am not sure how useful this logging is. But we should probably only do this for prod zones, to avoid noise in zones where we often provision new nodes on deployment

baldersheim (approved):

PR #31674, opened 2024-06-21
Require Go 1.22

mpolden:
Since Go 1.21, Go has the ability to bootstrap the required Go version
automatically. See https://go.dev/doc/toolchain

This means that we no longer need to limit ourselves to the exact Go version
provided by the OS package.

@aressem


aressem:
The build environment was not happy with this:
``` 
12:28:36 mkdir -p /sd/workspace/src/github.com/vespa-engine/vespa/build/go && GOTMPDIR=/sd/workspace/src/github.com/vespa-engine/vespa/build/go go test ./...
12:28:36 go: go.mod requires go >= 1.22 (running go 1.21.9; GOTOOLCHAIN=local)
12:28:36 make: *** [Makefile:140: test] Error 1
```

aressem (dismissed):

mpolden:
Seems to work now. I think this will also allow Renovate to update the Go version itself (as far as I can tell it's enabled automatically)

mpolden:
Made it independent of `CI` to make it work for developers using the dev container image.

PR #31675, opened 2024-06-21
Support decoding distribution config as part of cluster state bundles in C++

vekterli:
@geirst please review.

This mirrors 1-1 how the cluster controller will encode distribution config in state bundles (Coming Soon™️).

Actually _encoding_ config in the same format as that used for decoding config payloads is not directly supported, so we do our own roundabout conversion as part of testing.


geirst (approved):
👍

PR #31709, opened 2024-06-24
Jonmv/provide filter on data plane filters

jonmv:
@bjorncs please review.

### DIFF
--- a/container-core/src/main/java/com/yahoo/jdisc/http/filter/RequestFilter.java
+++ b/container-core/src/main/java/com/yahoo/jdisc/http/filter/RequestFilter.java
@@ -1,14 +1,13 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.jdisc.http.filter;
 
-import com.yahoo.component.chain.model.Chainable;
 import com.yahoo.jdisc.handler.ResponseHandler;
 import com.yahoo.jdisc.http.HttpRequest;
 
 /**
  * @author Einar M R Rosenvinge
  */
-public interface RequestFilter extends com.yahoo.jdisc.SharedResource, RequestFilterBase, Chainable {
+public interface RequestFilter extends com.yahoo.jdisc.SharedResource, RequestFilterBase {
 
     void filter(HttpRequest request, ResponseHandler handler);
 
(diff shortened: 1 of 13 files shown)


jdisc-security-filters/src/main/java/com/yahoo/jdisc/http/filter/security/base/JsonSecurityRequestFilterBase.java
```diff
@@ -25,7 +26,7 @@
  *
  * @author bjorncs
  */
-public abstract class JsonSecurityRequestFilterBase extends AbstractComponent implements SecurityRequestFilter {
+public abstract class JsonSecurityRequestFilterBase extends ChainedComponent implements SecurityRequestFilter {
```
jonmv, line 29: It's sad that the next person has to figure out _their_ filter isn't a child of this class, too, but I don't see a good solution here. Do you?

bjorncs: I'm not familiar with the framework. If it do require class inheritance I don't see any quick fix.

bjorncs (dismissed):
Nice! 👍 🎖️

container-core/src/main/java/com/yahoo/jdisc/http/filter/RequestFilter.java
```diff
@@ -1,13 +1,14 @@
...
 
+import com.yahoo.component.chain.model.Chainable;
 import com.yahoo.jdisc.handler.ResponseHandler;
 import com.yahoo.jdisc.http.HttpRequest;
 
 /**
  * @author Einar M R Rosenvinge
  */
-public interface RequestFilter extends com.yahoo.jdisc.SharedResource, RequestFilterBase {
+public interface RequestFilter extends com.yahoo.jdisc.SharedResource, RequestFilterBase, Chainable {
```
bjorncs, line 11: Any reason for not having `RequestFilterBase` / `ResponseFilterBase` inherit from `Chainable` instead?

bjorncs (dismissed):

container-core/src/main/java/com/yahoo/jdisc/http/filter/RequestFilter.java
jonmv: Probably not.

PR #31713, opened 2024-06-25
Don't attempt to remove orphans

bratseth:


baldersheim (approved):
I guess this might warrant a unit unit test ...

arnej27959 (approved):
looks good to me, but how can this trigger? looks like some invariant was broken already...

bratseth:
Yes, the only possibility I could find is duplicate items + removal with the iterator, which I fixed here.

PR #31730, opened 2024-06-26
Include distribution config as part of cluster state bundle

vekterli:
@geirst please review.

Opaque config Slime representation is encoded as part of the existing bundle object. Mirrors decoding in C++ impl.

New cluster state version and broadcast is triggered if the distribution config changes.

Including distribution config as part of the state bundle is controlled by a dedicated config parameter, which currently defaults to `false` (i.e. disabled).


geirst (approved):
👍

PR #31732, opened 2024-06-26
Add feature flag for zookeeper jute maxbuffer

hmusum:


hakonhall (approved):
Large znodes may cause the quorum to become non-convergent and unstable according to the doc.  I'd vote against raising it w/o a way to test this or stand-by for recovery.

hmusum:
Noted, we will try to change how we handle a some large items in the application package to avoid the need for doing this

PR #31736, opened 2024-06-27
Re-enable LLMSearcherTest

lesters:
@hmusum Please review.

This was previously disabled due to sometimes failing. This change removes the race condition in the `drain` here, as the second drain could conceivably result in 0 events before the code in the first drain processes the events.


hmusum (approved):
👍

PR #31746, opened 2024-06-28
Jonmv/provide filter on data plane filters

jonmv:
@bjorncs or @freva please review and merge. This is just a reapply with a fix to the ABI spec.


freva (approved):

bjorncs (commented):
👍

PR #31748, opened 2024-06-28
Also pass hostname to log handler

freva:
Just noticed this was removed in https://github.com/vespa-engine/vespa/commit/dff253253457e1cd570de3c55364e235cf8ce2de#diff-5c31444588efca8dabc84e0614374e23bcb8f1b83ede2f768054cc5f5e896c45R220, not sure if there was any reason to this? `LogHandler` supports filtering by hostname, so this should be propagated.


jonmv (approved):
Was a misdirected attempt at hygiene, thanks!

PR #31750, opened 2024-06-28
increase timeout to reduce re-starts (VESPANG-79)

havardpe:
@hmusum please review


hmusum (approved):

baldersheim (commented):
You should check the go code too. I think that is used in most cases.

hmusum:
For the go code, see https://github.com/vespa-engine/vespa/pull/31758

PR #31752, opened 2024-06-28
Emit cluster event with config diff on distribution config change

vekterli:
@geirst please review.

Adds a utility for computing the difference between two distribution configs with arbitrary parameter and topology changes, and for rendering this difference in a human-readable way. This diff is included in the cluster-level event upon distribution config changes.


geirst (approved):
👍

PR #31754, opened 2024-06-28
Prefer document frequency from query properties when calculating idf …

toregge:
…for bm25 feature.

@geirst : please review


geirst (approved):
👍

PR #31755, opened 2024-06-28
add flag for SegmentItem to trigger folding it into parent WeakAnd

arnej27959:
@bratseth please review

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/query/WeakAndItem.java
+++ b/container-search/src/main/java/com/yahoo/prelude/query/WeakAndItem.java
@@ -87,6 +87,32 @@ public final class WeakAndItem extends NonReducibleCompositeItem {
         putString(index, buffer);
     }
 
+    private WeakAndItem foldSegments() {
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/prelude/query/WeakAndItem.java
```diff
@@ -87,6 +87,32 @@ protected void encodeThis(ByteBuffer buffer) {
...
+                }
+            } else {
+                result.addItem(child);
+            }
+        }
+        return result;
+    }
+
+    @Override
+    public int encode(ByteBuffer buffer) {
```
bratseth, line 105: super.encode will encode this, then encode all the items in this (without folding). Won't the total result of this be a bit strange?

arnej27959: super.encode() should be called when no folding is needed; maybe the logic was not clear. will refactor some...

arnej27959:

bratseth (approved):
👍

PR #31781, opened 2024-07-01
Do not ignore convergence timeouts in status command

mpolden:
@jonmv

### DIFF
--- a/client/go/internal/cli/cmd/status_test.go
+++ b/client/go/internal/cli/cmd/status_test.go
@@ -153,7 +153,7 @@ func TestStatusCloudDeployment(t *testing.T) {
 		Body:   []byte(`{"active": true, "status": "running"}`),
 	})
 	assert.NotNil(t, cli.Run("status", "deployment"))
-	assert.Equal(t, `Timed out waiting for deployment to converge. See https://console.vespa-cloud.com/tenant/t1/application/a1/dev/instance/i1/job/dev-us-north-1/run/1337 for more details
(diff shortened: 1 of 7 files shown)


client/go/internal/cli/cmd/status_test.go
```diff
@@ -141,6 +141,23 @@ func TestStatusCloudDeployment(t *testing.T) {
...
+		Status: 200,
+		Body:   []byte(`{"runs": [{"id": 1337}]}`),
+	})
+	client.NextResponse(mock.HTTPResponse{
+		URI:    "/application/v4/tenant/t1/application/a1/instance/i1/job/dev-us-north-1/run/1337?after=-1",
+		Status: 200,
+		Body:   []byte(`{"active": true, "status": "running"}`),
+	})
+	assert.NotNil(t, cli.Run("status", "deployment"))
+	assert.Equal(t, `Timed out waiting for deployment to converge. See https://console.vespa-cloud.com/tenant/t1/application/a1/dev/instance/i1/job/dev-us-north-1/run/1337 for more details
```
jonmv, line 156: Couldn't we return a message more like "deployment started" or "deployment running", and then the "use wait" hint? Obviously only on the invocation where we actually start a new job.

jonmv (dismissed):

client/go/internal/cli/cmd/status_test.go
jonmv: ... or is this just a different API? I guess maybe it is. But the messages are the same?

mpolden: Changed the message.

PR #31782, opened 2024-07-01
Explain that developer key is tenant-wide

mpolden:
@kkraune


kkraune (approved):
Great! One small step for better usability :-)

PR #31797, opened 2024-07-01
Revert "Jonmv/remove unused stuff" MERGEOK

baldersheim:
Reverts vespa-engine/vespa#31792
@bratseth @jonmv PR
When both screwdriver and buildkite says no, it is wise to listen...


bratseth:
I thought it was caused by a differently-named internal PR but maybe not ...

jonmv:
The branch names should matter for that, unless that changed with new build setup, @aressem ?

aressem:
Vespa is not together with internal PRs anymore.

PR #31804, opened 2024-07-02
Handle distributor maintenance operation cancellation edge case

vekterli:
@geirst please review.

If a maintenance operation reply comes from a node that went down after its original request was sent _and_ cancelling is enabled, there is a discrepancy where the distributor stripe message tracker does not know about the operation, but the maintenance operation owner _does_ know about it.

This must be handled explicitly, as the message tracker returns a default constructed bucket when the message is not known. This default bucket must not be propagated out of the function, or we'll transitively trigger an invariant check failure.


geirst (approved):
👍

PR #31806, opened 2024-07-02
Balder/avoid trying to take substring outside of string view

baldersheim:
@toregge PR
@vekterli Take a look at the last commit

### DIFF
--- a/document/src/vespa/document/select/simpleparser.cpp
+++ b/document/src/vespa/document/select/simpleparser.cpp
@@ -42,7 +42,7 @@ bool IdSpecParser::parse(vespalib::stringref s)
                             ((len == 9) && (strncasecmp(&s[startPos], "namespace", 9) == 0)))
                         {
                             retval = true;
-                            setValue(ValueNode::UP(new IdValueNode(_bucketIdFactory, "id", s.substr(startPos, len), widthBits, divisionBits)));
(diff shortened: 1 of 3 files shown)


document/src/vespa/document/select/simpleparser.cpp
```diff
@@ -6,16 +6,25 @@
...
 
-bool icmp(char c, char l)
-{
-    return tolower(c) == l;
+void
+Parser::setRemaining(vespalib::stringref s, size_t fromPos) {
+    if ((fromPos + 1) < s.size()) {
+        _remaining = s.substr(fromPos);
+    } else {
+        _remaining = "";
```
vekterli, line 26: According to https://en.cppreference.com/w/cpp/string/basic_string_view/substr throws

> [`std::out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range) if `pos > size()`.

I.e. it's well-defined when `fromPos <= s.size()`. `s.substr(s.size())` thus returns a `string_view` with a length of zero, which is analogous to `_remaining = ""` in the else-case.

Maybe this could be simplified to `_remaining = s.substr(std::min(fromPos, s.size()))`?

vekterli (dismissed):
:+1:

document/src/vespa/document/select/simpleparser.cpp
baldersheim: Indeed, thanks.

PR #31807, opened 2024-07-02
Interns/magnus/completion

Mangern:
Add completion feature, currently applies to:
- Keyword snippets when inside a body
- Types when declaring a field

Also defined some general utility methods.


theodorklauritzen (approved):
LGTM

PR #31814, opened 2024-07-02
Integrate parser classes from config-model to language-server

theodorklauritzen:
Integrated the language server with the classes from config-model. Since they are in different packages a lot of methods and classes in com.yahoo.schema.parser are now public.

I have also added a way to ignore parser exceptions from `verifyThatIngoreable` if a boolean is set. `verifyThat` still works as earlier.

There will probably be further changes after expansion of the grammar in the language server. They will most likely set more methods to public and change some specific `verifyThat` to `verifyThatIgnoreable`.


arnej27959 (approved):
👍

PR #31816, opened 2024-07-02
Interns/magnus/completion

Mangern:
More completion
- rank-profile
- struct
- fieldset with field lookup
- compound types (basic)


theodorklauritzen (approved):
LGTM

PR #31817, opened 2024-07-02
Move <dev> to upper level in deployment spec, with endpoints

jonmv:
@bratseth please review, but don't merge yet.


bratseth (dismissed):
Nice 👍

PR #31824, opened 2024-07-03
Report correct autoscaling status when disabled through feature flag

mpolden:
@bratseth

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
@@ -10,6 +10,7 @@ import com.yahoo.vespa.hosted.provision.autoscale.Autoscaling.Status;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.logging.Logger;
 
 /**
  * The autoscaler gives advice about what resources should be allocated to a cluster based on observed behavior.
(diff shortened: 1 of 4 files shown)


mpolden:
or @bjormel

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/autoscale/Autoscaler.java
```diff
@@ -53,28 +56,29 @@ public List<Autoscaling> suggest(Application application, Cluster cluster, NodeL
...
      */
-    public Autoscaling autoscale(Application application, Cluster cluster, NodeList clusterNodes, boolean enableDetailedLogging) {
+    public Autoscaling autoscale(Application application, Cluster cluster, NodeList clusterNodes, boolean enabled, boolean logDetails) {
         var limits = Limits.of(cluster);
         var model = model(application, cluster, clusterNodes);
         if (model.isEmpty()) return Autoscaling.empty();
 
-        if (! limits.isEmpty() && cluster.minResources().equals(cluster.maxResources()))
+        if (!enabled || (! limits.isEmpty() && cluster.minResources().equals(cluster.maxResources())))
             return Autoscaling.dontScale(Autoscaling.Status.unavailable, "Autoscaling is not enabled", model);
```
bratseth, line 70: Maybe we should output two different messages depending on the reason so that people get why it's disabled even if they themselves have enabled it?

bratseth (dismissed):

PR #31828, opened 2024-07-03
Arnej/configurable legacy query parsing

arnej27959:
@bratseth please review
@baldersheim please take a look also
@geirst @havardpe FYI


bratseth (dismissed):
Look solid to me. 

One though: Once this is wired as throughly as this I think it will likely find other uses when we want to tweak parsing behavior/defaults and so be here to stay until Vespa is carried to the one million galaxies we'll eventually rule. So maybe it would be better to call it "ParserSettings" or something like that?

baldersheim (dismissed):

PR #31834, opened 2024-07-03
Emit full cluster state bundle string on state manager status page

vekterli:
@geirst please review

Since cluster state bundles now contain distribution config, this will also show the history of distribution config changes received on a particular node.


geirst (approved):
👍

PR #31840, opened 2024-07-04
Create builder from existing json

bjorncs:

### DIFF
--- a/vespajlib/src/main/java/ai/vespa/json/Json.java
+++ b/vespajlib/src/main/java/ai/vespa/json/Json.java
@@ -211,6 +211,18 @@ public class Json implements Iterable<Json> {
             if (cursor.type() != Type.ARRAY) throw new InvalidJsonException("Input is not an array");
             return new Builder.Array(cursor);
         }
+        public static Builder.Object fromObject(Json json) {
+            json.requireType(Type.OBJECT);
+            var builder = newObject();
+            SlimeUtils.copyObject(json.inspector, builder.cursor);
+            return builder;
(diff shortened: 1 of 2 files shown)


vespajlib/src/main/java/ai/vespa/json/Json.java
```diff
@@ -211,6 +211,18 @@ public static Builder.Array existingSlimeArrayCursor(Cursor cursor) {
             if (cursor.type() != Type.ARRAY) throw new InvalidJsonException("Input is not an array");
             return new Builder.Array(cursor);
         }
+        public static Builder.Object fromObject(Json json) {
```
hakonhall, line 214: There should be a from(Json) too, that delegates to one of Object or Array depending on the type?

hakonhall (approved):

vespajlib/src/main/java/ai/vespa/json/Json.java
bjorncs: That method would have the return type `Builder` which does not provide any methods for mutating the structure.

hakonhall: makes sense

bjorncs: @jonmv `json.requireType(Type.OBJECT)` will throw `InvalidJsonException`

jonmv: Yes ... >_<

PR #31860, opened 2024-07-05
Interns/theodorkl/expand grammar

theodorklauritzen:
Add comments as semantic tokens

Refactor SchemaNode with multiple classes that extends SchemaNode


Mangern (approved):
lgtm

PR #31846, opened 2024-07-04
- make vespalib::strivg construction from std::string_view explicit t…

baldersheim:
…o bring it closer to std::string.

- Clean up mix and matcg with string/string_view.

@toregge PR

PR #31865, opened 2024-07-05
Revert "- make vespalib::strivg construction from std::string_view explicit t…"

aressem:
Reverts vespa-engine/vespa#31846

Caused this test to start failing:
https://factory.vespa.oath.cloud/testrun/88877974/test/MapTypeTest::test_map__STREAMING


baldersheim:
Fixed in #31870

PR #31871, opened 2024-07-06
Enable heterogenous lookup in map from field name to field id.

toregge:
@baldersheim : please review


toregge:
Note that when map key is changed to `std::string` then each compare during the `find` might involve a temporary `std::string_view`, resulting in a tradeoff between multiple cheap temporaries versus one slightly more expensive one.

baldersheim (approved):

baldersheim:
I assume that the compiler will optimize that away.

PR #31873, opened 2024-07-06
Revert "Parse sameElement"

baldersheim:
Reverts vespa-engine/vespa#31869

This cause a system test to fail. Not sure if the test is relevant, or it points out an ambiguity with same element.
https://factory.vespa.oath.cloud/testrun/88919366/test/IndexAddressing::test_index_addressing__INDEXED


bratseth:
I'll look into it.

PR #31880, opened 2024-07-06
Add temporary abort to root out potential issues before switching to …

baldersheim:
…std::string.

@toregge PR


vespalib/src/vespa/vespalib/stllike/string.h
```diff
@@ -431,6 +431,8 @@ class small_string
             _bufferSize = StackSize;
             if (s) {
                 memcpy(_stack, s, _sz);
+	    } else {
+                abort();
```
toregge, line 435: Consider only aborting if _sz is nonzero. And add an abort in constructor at line 36 when `s` is null.

toregge (changes requested):

vespalib/src/vespa/vespalib/stllike/string.h
baldersheim: It was intentionally rather strict, but I agree that it should be enough to mimic std::string.

toregge (approved):

PR #31881, opened 2024-07-07
Balder/factor out small string

baldersheim:
@toregge PR
This also contains the commit in #31880


vespalib/src/tests/stllike/string_test.cpp
```diff
@@ -449,4 +449,10 @@ TEST("starts_with has expected semantics for std::string_view") {
     EXPECT_FALSE(ar.starts_with("oobar"));
 }
 
+TEST("test allowed nullptr construction") {
+    EXPECT_TRUE(vespalib::string(nullptr, 0).empty());
```
toregge, line 453: Consider adding a comment about this being legal but not legitimate cases.

toregge (dismissed):

vespalib/src/tests/stllike/string_test.cpp
baldersheim: Done

toregge (dismissed):

PR #31867, opened 2024-07-05
Revert "Make query rate growth measurement less sensitive to sudden bursts"

hakonhall:
Reverts vespa-engine/vespa#31866

Broke ControllerMaintainer.

PR #31896, opened 2024-07-08
Reapply "Make query rate growth measurement less sensitive to sudden bursts"

mpolden:
Reverts vespa-engine/vespa#31867


mpolden:
@bratseth or @bjormel

bjormel (approved):

PR #31901, opened 2024-07-08
Propagate document/tombstone timestamps over the wire during visiting

vekterli:
@geirst please review.

This revives the ability to get persistence timestamps in visiting, which for technical legacy protocol reasons™️ has been gone for a long time. Now that the DocumentAPI protocol is Protobuf-based, such woes are no more.

Note that the timestamps are not _yet_ wired to the default visitor message handler functions, but they are available as part of the raw messages themselves.

Also includes misc cleanup of backend visitor unit tests.


geirst (approved):
👍

PR #31906, opened 2024-07-08
Interns/magnus/workspace

Mangern:
Implement document inheritance.
- Parse all .sd files in workspace on init
- Manage inheritance relationships in a DAG
- Lookup struct types in inherited documents
- Update descendants when a document changes


theodorklauritzen (approved):
Very good

PR #31913, opened 2024-07-08
Restore lost element in attribute memory usage calculation.

toregge:
@vekterli or @baldersheim : please review


baldersheim (approved):

vekterli (approved):
:+1:

PR #31920, opened 2024-07-09
Revert "Enable std::string as default string implementation." MERGEOK

vekterli:
@toregge please review
@baldersheim FYI

Some refeeding performance tests show slightly degraded numbers after the flip. Temporarily reverting to avoid any surprises while we investigate.

Reverts vespa-engine/vespa#31894


toregge (approved):

toregge:
When running the feeding attributes performance test manually in a dev environment then the refeeding with 1 thread throughput is improved by 10-20% by switching from vespalib::string to std::string. But using the feature flags from `docker/include/feature-flags.json` in the system-test repo causes the throughput to be 25-30% worse for the same switch.

PR #31925, opened 2024-07-10
Return overflowing tokens to avoid memory leak in HF tokenizer

lesters:
@bjorncs Please review
@havardpe FYI


baldersheim (dismissed):

bjorncs (dismissed):

bjorncs:
Consider adding a comment in the code for reason why this option is set

PR #31926, opened 2024-07-10
Resolve dimension bindings in ranking expressions

bratseth:
People can parametrize the names of features in ranking expression functions, and then naturally expects to be able to do the same with tensor dimension names. This adds support for that in type resolution and expression generation, which is enough for supporting this in rank profiles.

We should also support it in direct evaluation of Java expressions
for completeness. I'll do that separately, this PR is already large enough.


arnej27959 (approved):
looks mostly good to me, please check if you need to modify
GenerateToStringContext also.

bratseth:
I think it's correct like this since GenerateToStringContext lets the dimensions in the local type spec override the normal (modified) resolution.

PR #31927, opened 2024-07-10
Always skip the {

bratseth:


bratseth:
I couldn't guess the query which causes SO but this should fix it.

bjormel:
Thanks! I will see if I am able to find a request to verify

bjormel (approved):

baldersheim (approved):

PR #31934, opened 2024-07-10
Test that new state versions are published when startup timestamps change

vekterli:
@geirst please review. This functionality is already present, but I want to extend the test coverage.

We want to ensure that all changes to the cluster's active topology trigger new cluster state versions. Otherwise we risk ambiguity of what the state of nodes actually were for a given version.


geirst (approved):
👍

PR #31943, opened 2024-07-11
Don't render stack traces on illegal argument

bratseth:
Since IllegalArgumentException indicates something wrong with the request, we shouldn't expect the stack trace to contain anything useful and rendering just adds noise.

@baldersheim or @arnej27959

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
+++ b/container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
@@ -305,7 +305,7 @@ public class JsonRenderer extends AsynchronousSectionedRenderer<Result> {
             if (message != null) {
                 generator.writeStringField(ERROR_MESSAGE, message);
             }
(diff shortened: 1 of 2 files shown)


container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
```diff
@@ -317,6 +317,10 @@ protected void renderErrors(Set<ErrorMessage> errors) throws IOException {
         generator.writeEndArray();
     }
 
+    protected boolean shouldRenderStacktraceOf(Throwable cause) {
```
arnej27959, line 320: either update abi-spec or make this private.  Also consider moving the
`cause.getStackTrace().length > 0` condition into `shouldRenderStacktraceOf()`

bratseth: Thanks, added. It makes sense to be able to override this in a subclass. For the same reason the existence check should be where it is.

PR #31947, opened 2024-07-12
Guard against buffer overread when comparing sort blobs.

toregge:
@vekterli : please review
@geirst, @baldersheim : FYI


vekterli (approved):
:+1:

PR #31952, opened 2024-07-14
Avoid negative values passed to isspace in vespalib.

toregge:
@vekterli : please review


vekterli (approved):
:+1:

vespalib/src/vespa/vespalib/stllike/asciistream.cpp
```diff
@@ -215,7 +216,7 @@ T
 strToInt(T & v, const char *begin, const char *end)
 {
     const char * curr = begin;
-    for (;(curr < end) && std::isspace(*curr); curr++);
```
baldersheim, line 218: common eat_white ...

baldersheim (commented):
👍

PR #31953, opened 2024-07-14
Avoid negative values passed to isspace in document module.

toregge:
@vekterli : please review


baldersheim (approved):

vekterli (approved):
:+1: would probably do good to have some utility `cctype` functions in vespalib with explicit conversions to avoid this common pitfall

Issue #31945, opened 2024-07-12
summary: matched-elements-only doesn't work with weightedSet/in query syntax.


dadib:
**Describe the bug**
With `summary: matched-elements-only` on a map field, the matched elements are not returned when using `weightedSet` or `in` query syntax.

**To Reproduce**
The schema has the following map field
```
    field tags type map<string, string> {
      indexing: summary
      summary: matched-elements-only
      struct-field key {
        indexing: attribute
        match: exact
      }
      struct-field value {
        indexing: attribute
        match: exact
      }
    }
```
The query `SELECT * FROM root_asset WHERE tags.value contains 'test_value'` returns the matched field in summary as expected:
```
  {
        "id": "index:text/0/618014578ff5f43cc510eb75",
        "relevance": 1.0,
        "source": "text",
        "fields": {
          "matchfeatures": {
            "attribute(titleTokens).count": 1.0,
            "attributeMatch(tags.value)": 0.96,
            "attributeMatch(tags.value).completeness": 0.96,
            "attributeMatch(tags.value).fieldCompleteness": 0.2,
            "attributeMatch(tags.value).matches": 1.0,
            "attributeMatch(tags.value).queryCompleteness": 1.0,
            "attributeMatch(title).matches": 0.0,
            "attributeMatch(titleTokens).matches": 0.0
          },
          "sddocname": "root_asset",
          "title": "TEST",
          "tags": {
            "test_key": "test_value"
          }
        }
      }
```
but when using `weightedSet` or `in` it gets the hit but fails to return the matched field in the summary
```
SELECT * FROM root_asset WHERE (weightedSet(tags.value, {'test_value':1})) 
SELECT * FROM root_asset WHERE tags.value in ('test_value') 
```
```
      {
        "id": "index:text/0/618014578ff5f43cc510eb75",
        "relevance": 1.0,
        "source": "text",
        "fields": {
          "matchfeatures": {
            "attribute(titleTokens).count": 1.0,
            "attributeMatch(tags.value)": 0.96,
            "attributeMatch(tags.value).completeness": 0.96,
            "attributeMatch(tags.value).fieldCompleteness": 0.2,
            "attributeMatch(tags.value).matches": 1.0,
            "attributeMatch(tags.value).queryCompleteness": 1.0,
            "attributeMatch(title).matches": 0.0,
            "attributeMatch(titleTokens).matches": 0.0
          },
          "sddocname": "root_asset",
          "title": "TEST"
        }
      }
```

[...]

PR #31974, opened 2024-07-16
Calculate matched elements for searches in struct attributes

toregge:
with in and weightedset operator (indexed mode only).

This is related to issue #31945

@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #31975, opened 2024-07-16
Revert "Enable std::span instead of vespalib::ArrayRef."

toregge:
Reverts vespa-engine/vespa#31973

@vekterli : please review

In case revert is needed.


vekterli (approved):
:+1:

baldersheim:
Not needed

PR #31979, opened 2024-07-17
Interns/magnus/parent child

Mangern:
Support for import field statement, checking the document reference and referencing the imported field.


theodorklauritzen (approved):
LGTM

PR #31981, opened 2024-07-17
Detect conflict between tensor field and non-tensor field in fieldset.

toregge:
@vekterli : please review
@geirst : FYI

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/processing/FieldSetSettings.java
+++ b/config-model/src/main/java/com/yahoo/schema/processing/FieldSetSettings.java
@@ -25,6 +25,8 @@ import java.util.LinkedList;
 // (this requires adding normalizing and stemming settings to FieldSet).
 public class FieldSetSettings extends Processor {
 
+    private static String fieldSetDocUrl = "https://docs.vespa.ai/en/reference/schema-reference.html#fieldset";
+
     public FieldSetSettings(Schema schema,
                             DeployLogger deployLogger,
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/schema/processing/FieldSetSettings.java
```diff
@@ -104,4 +108,21 @@ private void checkStemming(Schema schema, FieldSet fieldSet) {
...
+            if (field.getDataType() instanceof TensorDataType) {
+                tensorFields.add(field.getName());
+            } else {
+                nonTensorFields.add(field.getName());
+            }
+        }
+        if (!tensorFields.isEmpty() && !nonTensorFields.isEmpty()) {
+            throw new IllegalArgumentException("For schema '" + schema.getName() + "', fieldset '" + fieldSet.getName() + "': " +
+                    "Illegal mixing of tensor fields ['" + String.join("','", tensorFields) + "'] " +
+                    "and non-tensor fields ['" + String.join("','", nonTensorFields) + "']");
```
vekterli, line 125: Consider instead something like `"For schema 'X', fieldset 'Y': tensor fields ['foo', 'bar'] cannot be mixed with non-tensor fields ['baz', 'zoid'] in the same fieldset. See https://docs.vespa.ai/en/reference/schema-reference.html#fieldset"`

vekterli (dismissed):
:+1:

vekterli (approved):

PR #31986, opened 2024-07-18
Retry connecting to OpenAI

lesters:
@bjorncs Please review.

Sometimes, particularly in cloud, if the connection to OpenAI hadn't been used in a while, the connect attempt would fail with a "Connection reset" error. This PR introduces a retry mechanism.

Please particularly take a look at the cases when we should retry, which is made clear in the `shouldRetry` method.

### DIFF
--- a/vespajlib/src/main/java/ai/vespa/llm/client/openai/OpenAiClient.java
+++ b/vespajlib/src/main/java/ai/vespa/llm/client/openai/OpenAiClient.java
@@ -161,7 +161,7 @@ public class OpenAiClient implements LanguageModel {
                 waitBeforeRetry();
                 completeAsyncAttempt(context, attempt + 1);
             } else {
-                context.completionFuture().completeExceptionally(new RuntimeException("OpenAI: max retries reached"));
+                context.completionFuture().completeExceptionally(new RuntimeException("OpenAI: max retries reached", exception));
             }
         } else {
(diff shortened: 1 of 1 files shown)


vespajlib/src/main/java/ai/vespa/llm/client/openai/OpenAiClient.java
```diff
@@ -67,44 +76,101 @@ public List<Completion> complete(Prompt prompt, InferenceParameters options) {
...
+        return false;
+    }
+
+    private void handleHttpException(Throwable exception, CompletionContext context, int attempt) {
+        if (shouldRetry(exception)) {
+            if (attempt < MAX_RETRIES) {
+                waitBeforeRetry();
+                completeAsyncAttempt(context, attempt + 1);
+            } else {
+                context.completionFuture().completeExceptionally(new RuntimeException("OpenAI: max retries reached"));
```
bjorncs, line 164: Consider including the original exception as cause to the new `RuntimeException` instance.

bjorncs (dismissed):

vespajlib/src/main/java/ai/vespa/llm/client/openai/OpenAiClient.java
lesters: Thanks!

PR #31988, opened 2024-07-19
Support custom paths for cli cert checks 

olaughter:

This aims to resolve: https://github.com/vespa-engine/vespa/issues/31978

I'm hoping I took this in the right direction! I was keen to give this a go. I'm very happy to take feedback and/or see maintainers make necessary changes to this, etc.

The tldr behaviour that i've sought to add is that cert checks will now use the cert paths defined in `services.xml` if any are found, otherwise it looks for the default cert path.

Look forward to hearing your review!

### DIFF
--- a/client/go/internal/cli/cmd/cert.go
+++ b/client/go/internal/cli/cmd/cert.go
@@ -157,7 +157,8 @@ func doCertAdd(cli *CLI, overwriteCertificate bool, args []string) error {
 	if err != nil {
 		return err
 	}
-	if pkg.HasCertificate() && !overwriteCertificate {
+	defaultCertPath := []string{"security/clients.pem"}
+	if pkg.HasCertificate(defaultCertPath) && !overwriteCertificate {
 		return errHint(fmt.Errorf("application package '%s' already contains a certificate", pkg.Path), "Use -f flag to force overwriting")
 	}
 	return requireCertificate(true, false, cli, target, pkg)
@@ -182,7 +183,11 @@ func requireCertificate(force, ignoreZip bool, cli *CLI, target vespa.Target, pk
 	if force {
 		return copyCertificate(tlsOptions, cli, pkg)
 	}
-	if pkg.HasCertificate() {
+	servicesXML, err := readServicesXML(pkg)
+	if err != nil {
+		return fmt.Errorf("Error reading services.xml: %w", err)
+	}
+	if pkg.HasCertificate(servicesXML.CertPaths()) {
 		if cli.isCI() {
 			return nil // A matching certificate is not required in CI environments
 		}
--- a/client/go/internal/vespa/application.go
+++ b/client/go/internal/vespa/application.go
@@ -20,7 +20,22 @@ type ApplicationPackage struct {
 	TestPath string
 }
 
-func (ap *ApplicationPackage) HasCertificate() bool { return ap.hasFile("security", "clients.pem") }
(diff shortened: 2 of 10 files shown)


client/go/internal/cli/cmd/cert.go
```diff
@@ -182,7 +183,11 @@ func requireCertificate(force, ignoreZip bool, cli *CLI, target vespa.Target, pk
 	if force {
 		return copyCertificate(tlsOptions, cli, pkg)
 	}
-	if pkg.HasCertificate() {
+	servicesXML, err := readServicesXML(pkg)
+	if err != nil {
+		return fmt.Errorf("Error reading services.xml: %w", err)
```
mpolden, line 188: Nit: Error messages should be lower-case as this may be concatenated with other errors further up.

client/go/internal/vespa/application.go
```diff
@@ -20,7 +20,22 @@ type ApplicationPackage struct {
...
+			return true
+		} else {
+			return false
+		}
+	} else {
+		for _, certPath := range certPaths {
+			if ap.hasFile(certPath) {
+				return true
+			}
+		}
```
mpolden, line 35: This should check that all files are present, not just one.

client/go/internal/vespa/application.go
```diff
@@ -20,7 +20,22 @@ type ApplicationPackage struct {
...
+		return false
+	}
+}
 
 func (ap *ApplicationPackage) HasMatchingCertificate(certificatePEM []byte) (bool, error) {
```
mpolden, line 40: This needs to consider all certificates now, not just `clients.pem`. I think `ApplicationPackage` could have a method that reads the certificate paths from services XML, then both `HasCertificate` and `HasMatchingCertificate` can use this method and you won't have to pass around `certPaths` everywhere.

mpolden (changes requested):
Good job! I have some suggestions though.

client/go/internal/vespa/application.go
olaughter: Thanks @mpolden, I'll look at refactoring for this. The issue I had originally was that the cert.go file imports ApplicationPackage, leading to a circular import if application.go wants to make use of the file parsing from cert.go. I'll revisit and see what I can do to make it work

client/go/internal/cli/cmd/cert.go
olaughter: Ah good spot!

client/go/internal/vespa/application.go
olaughter: 👍

client/go/internal/vespa/application.go
olaughter: Hey, I've been playing around with this and think I could do with some steer on solving it.

I've been trying to get around the circular imports by changing the `vespa` package. `xml/config.go` imports `System`, from `vespa`, but moving that to its own package ended up with a lot of files changed without solving the issue. I also tried extracting ApplicationPackage, but again couldn't get that to work on its own. There might be a way of solving it this way but I don't feel confident enough to do it as it seems it would take a more dramatic refactoring then I'm comfortable with.

The only other option I could think of was to recreate some of the xml parsing functionality within application package. It wouldn't be very dry, but would enable this. What do you think is the best way forward?

(btw I'm on the vespa slack so let me know if its easier to chat this through there)

PR #31993, opened 2024-07-21
Cache-friendly prime table

blacktea:

This PR optimizes the getting modulo method. Based on the benchmarks https://quick-bench.com/q/-bPdO1GtFx1DQ6Ve5dZhaah_eso, the performance increased by up to 4 times.

Method:
The original `__stl_prime_list` table is split into arrays of 8 elements. It allows the cache line to accommodate these arrays decreasing cache misses.
The `__upper_bound_list_size` array is used to find a nested array where the given `newSize` is located.

Testing:
The method is checked here: https://godbolt.org/z/bd7rM71f6

### DIFF
--- a/vespalib/src/vespa/vespalib/stllike/hashtable.cpp
+++ b/vespalib/src/vespa/vespalib/stllike/hashtable.cpp
@@ -4,33 +4,32 @@
 
 namespace {
 
-static const unsigned long __stl_prime_list[] =
+static const size_t __chunk_size{8};
+static const unsigned long __stl_prime_list[][__chunk_size] =
 {
-            7ul,        17ul,         53ul,         97ul,        193ul,
-          389ul,       769ul,       1543ul,       3079ul,       6151ul,
-        12289ul,     24593ul,      49157ul,      98317ul,     196613ul,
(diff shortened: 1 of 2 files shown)


blacktea:
Note: I'd avoid using double underscores as well since they are supposed to be reserved by compilers. 
Let me know if you want me to rename variable names.

baldersheim (changes requested):

jonmv:
If this is a critical piece of code, it's much faster to find the relevant part of the list straight from the order of the msb: https://quick-bench.com/q/9W-F5Q5sCBJr96PXjsPq2a9PFGU, and then comparing the only two relevant elements. I have no idea how to write C++, though, so don't look too hard at the details :p Also, this is obviously only true while the list has the property that the primes are between consecutive powers of two (except for the first one, which is handled with the `+ 8` 🙈).

baldersheim:
I have never observed this as consuming time. Perhaps @blacktea can add some info on how this was observed ?
I agree the optimal solution would be to use the msb as lookup index.

baldersheim:
@blacktea Do you intend to complete this one anytime soon ?

baldersheim:
Implemented msb index lookup in #32168, closing

PR #32015, opened 2024-07-27
Sort components

hmusum:
Components are used when generating config, make sure ordering is consistent

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/component/ConfigProducerGroup.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/component/ConfigProducerGroup.java
@@ -9,8 +9,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
-
 
 /**
  * A group of config producers that have a component id.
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/vespa/model/container/component/ConfigProducerGroup.java
```diff
@@ -51,14 +51,15 @@ public Collection<CHILD> getComponents() {
...
+        List<T> result = new ArrayList<>();
 
         for (CHILD child: getChildren().values()) {
             if (componentClass.isInstance(child)) {
                 result.add(componentClass.cast(child));
             }
         }
-        return Collections.unmodifiableCollection(result);
+        // We need consistent ordering
+        return result.stream().sorted().toList();
```
baldersheim, line 62: Since you are introducing streams here I would go all the way with
```
return getChildren().values().stream()
                .filter(componentClass::isInstance)
                .map(componentClass::cast)
                .sorted().toList();
```

baldersheim (dismissed):

config-model/src/main/java/com/yahoo/vespa/model/container/component/ConfigProducerGroup.java
hmusum: Yeah, considered that, but did the least possible change. Probably better to go all the way, as you say.

baldersheim (approved):
I removed some unused imports.

PR #32062, opened 2024-07-31
Add sdm-enabled-provisioning feature flag

olaaun:

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -539,6 +539,12 @@ public class Flags {
             "Whether to sync companies to HubSpot",
             "Takes effect immediately");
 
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -539,6 +539,12 @@ public class Flags {
             "Whether to sync companies to HubSpot",
             "Takes effect immediately");
 
+    public static final UnboundBooleanFlag SDM_ENABLED_PROVISIONING = defineFeatureFlag(
+            "sdm-enabled-provisioning", true,
+            List.of("olaa"), "2024-07-31", "2024-10-01",
+            "Whether to provision with SDM enabled",
+            "Takes effect immediately");
```
freva, line 546: Currently the check is where we create the client, which happens once when we start the configservers. If we want this to take effect immediately, the check should be moved to `CloudHostProvisioner::strongDMClient`

freva (approved):

PR #32069, opened 2024-08-01
Minor cleanup, no functional changes

hmusum:

### DIFF
--- a/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
+++ b/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
@@ -41,49 +41,38 @@ public class FilesApplicationFile extends ApplicationFile {
     }
 
     @Override
-    public boolean isDirectory() {
-        return file.isDirectory();
-    }
+    public boolean isDirectory() { return file.isDirectory(); }
 
     @Override
-    public boolean exists() {
-        return file.exists();
-    }
+    public boolean exists() { return file.exists(); }
 
     @Override
     public ApplicationFile delete() {
-        log.log(Level.FINE, () -> "Delete " + file);
-        if (file.isDirectory() && !listFiles().isEmpty()) {
-            throw new RuntimeException("files. Can't delete, directory not empty: " + this  + "(" + listFiles() + ")." + listFiles().size());
-        }
-        if (file.isDirectory() && file.listFiles() != null && file.listFiles().length > 0) {
-            for (File f : file.listFiles()) {
-                deleteFile(f);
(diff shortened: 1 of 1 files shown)


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
```diff
@@ -41,49 +41,38 @@ public FilesApplicationFile(Path path, File file) {
...
-            if (path.isDirectory()) {
-                File[] files = path.listFiles();
-                for(int i=0; i<files.length; i++) {
-                    if(files[i].isDirectory()) {
-                        deleteFile(files[i]);
-                    } else {
-                        files[i].delete();
-                    }
-                }
+        if (path.exists() && path.isDirectory()) {
```
hakonhall, line 69: isDirectory => exists, so remove `path.exists()` ?

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
```diff
@@ -41,49 +41,38 @@ public FilesApplicationFile(Path path, File file) {
...
     public ApplicationFile delete() {
-        log.log(Level.FINE, () -> "Delete " + file);
-        if (file.isDirectory() && !listFiles().isEmpty()) {
-            throw new RuntimeException("files. Can't delete, directory not empty: " + this  + "(" + listFiles() + ")." + listFiles().size());
-        }
-        if (file.isDirectory() && file.listFiles() != null && file.listFiles().length > 0) {
-            for (File f : file.listFiles()) {
-                deleteFile(f);
+        if (file.isDirectory()) {
+            if (!listFiles().isEmpty())
```
hakonhall, line 52: I believe this `delete()` can only be invoked on a directory if the directory only has entries that start with `.`, and in that case any such file and directory-trees are deleted.

This seems strange to me - does it make sense to you?

It is possibly related to the fact that this class creates a .meta directory after deletion.

hakonhall (dismissed):

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
hmusum: Fixed

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
hmusum: Discussed offline, added some TODOs

PR #32088, opened 2024-08-06
Hmusum/support deleting local and remote session in one go

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -686,28 +686,27 @@ public class SessionRepository {
         // All known sessions, both local (file) and remote (zookeeper)
         Set<Long> sessions = getLocalSessionsIdsFromFileSystem();
         sessions.addAll(getRemoteSessionsFromZooKeeper());
-
         log.log(Level.FINE, () -> "Sessions for tenant " + tenantName + ": " + sessions);
 
-        // TODO: Sessions created last 30 seconds are considered new, exclude these to avoid deleting new sessions
+        // Skip sessions newly added (we might have a session in the file system, but not in ZooKeeper,
+        // we will exclude these)
         Set<Long> newSessions = findNewSessionsInFileSystem();
+        sessions.removeAll(newSessions);
+
         // Avoid deleting too many in one run
         int deleteMax = (int) Math.min(1000, Math.max(50, sessions.size() * 0.05));
         int deleted = 0;
-        try {
-            for (Long sessionId : sessions) {
-                // Skip sessions newly added (we might have a session in the file system, but not in ZooKeeper,
-                // we will exclude these)
-                if (newSessions.contains(sessionId)) continue;
-
+        for (Long sessionId : sessions) {
+            try {
                 Session session = remoteSessionCache.get(sessionId);
                 if (session == null)
                     session = new RemoteSession(tenantName, sessionId, createSessionZooKeeperClient(sessionId));
 
-                Instant createTime = session.getCreateTime();
                 Session.Status status = session.getStatus();
-                if (session.getStatus() == ACTIVATE && sessionIsActiveForApplication.test(session)) continue;
+                boolean activeForApplication = sessionIsActiveForApplication.test(session);
+                if (status == ACTIVATE && activeForApplication) continue;
 
+                Instant createTime = session.getCreateTime();
                 boolean hasExpired = hasExpired(createTime);
                 if (hasExpired) {
                     log.log(Level.FINE, () -> "Remote session " + sessionId + " for " + tenantName + " has expired, deleting it");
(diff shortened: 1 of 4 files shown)


configserver/src/test/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainerTest.java
```diff
@@ -189,6 +186,64 @@ public void testDeletingInactiveSessions3() throws IOException {
...
+        assertNumberOfLocalSessions(3);
+        assertNumberOfRemoteSessions(3);
+
+        // advance clock more than session lifetime
+        clock.advance(Duration.ofSeconds(applicationRepository.configserverConfig().sessionLifetime()));
+
+        // All sessions except session id 3 should be removed after maintainer has run
+        maintainer.maintain(Duration.ofSeconds(5));
+        assertNumberOfLocalSessions(2);
+        assertNumberOfRemoteSessions(2);
```
jonmv: Why 2, then?

configserver/src/test/java/com/yahoo/vespa/config/server/maintenance/SessionsMaintainerTest.java
```diff
@@ -210,16 +210,19 @@ public void testDeletionWithDelay() {
...
+        assertLocalSessions(Set.of(2L, 3L, 4L));
+
+        // advance clock session lifetime minus time elapsed since session 2 was created
+        // minus 2 seconds
+        clock.advance(Duration.ofSeconds(applicationRepository.configserverConfig().sessionLifetime())
+                              .minus(Duration.ofSeconds(20).plus(Duration.ofSeconds(2))));
+
+        // Sessions 3 and 4 should be left after maintainer has run, 2 should be deleted
+        // Test this with advancing clock 5 seconds between deletion of local and remote sessions
+        maintainer.maintain(() -> clock.advance(Duration.ofSeconds(5)));
```
jonmv: 👍

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,77 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+                    if ( ! sessionIsActiveForApplication.test(localSession)) {
+                        log.log(Level.FINE, () -> "Will delete inactive session " + sessionId + " created " +
+                                createTime + " for '" + applicationId + "'");
+                        deleteLocalSession(sessionId);
+                    }
+                }
+                if (deleted >= deleteMax)
+                    return;
+            }
+        } catch (Throwable e) { // Make sure to catch here, to avoid executor just dying in case of issues ...
```
jonmv, line 745: Should the try-catch be per session, to avoid one failing session blocking deletion of others? Alternatively, use a random or time-dependent subset of sessions—perhaps the first 0.05, but in reverse creation time order?

jonmv: Ah, they're in somewhat random order already, from being a hash set of longs ... 
and using reverse creation order potentially leaves older sessions hanging, of course. 
So some true randomness, then, or moving the try-catch down a step.

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,77 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+        // TODO: Sessions created last 30 seconds are considered new, exclude these to avoid deleting new sessions
+        Set<Long> newSessions = findNewSessionsInFileSystem();
+        // Avoid deleting too many in one run
+        int deleteMax = (int) Math.min(1000, Math.max(50, sessions.size() * 0.05));
+        int deleted = 0;
+        try {
+            for (Long sessionId : sessions) {
+                // Skip sessions newly added (we might have a session in the file system, but not in ZooKeeper,
+                // we will exclude these)
+                if (newSessions.contains(sessionId)) continue;
```
jonmv, line 701: Perhaps simpler to `sessions.removeAll(newSessions)` outside the loop?

jonmv: Well, this fits with the below `continue`, so perhaps not ...

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,77 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+                    log.log(Level.FINE, () -> "Remote session " + sessionId + " for " + tenantName + " has expired, deleting it");
+                    deleteRemoteSessionFromZooKeeper(session);
+                    deleted++;
+                }
+
+                log.log(Level.FINE, () -> "Candidate local session for deletion: " + sessionId +
+                        ", created: " + createTime + ", status " + status + ", can be deleted: " + canBeDeleted(sessionId, status) +
+                        ", hasExpired: " + hasExpired);
+                if (hasExpired && canBeDeleted(sessionId, status)) {
+                    log.log(Level.FINE, () -> "expired: " + hasExpired + ", can be deleted: " + canBeDeleted(sessionId, status));
```
jonmv, line 722: I'd think `canBeDeleted` is always true here, unless there's a race, which we don't want?

jonmv: Looks like a redundant log line, with the logging a few lines up?

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,77 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+                    deleteLocalSession(sessionId);
+                } else if (createTime.plus(Duration.ofDays(1)).isBefore(clock.instant())) {
```
jonmv, line 724: Could expiration time ever exceed 1 day? If it does, this becomes a bit weird—it's not "expired", but we attempt forceful deletion anyway.

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,77 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+                    LocalSession localSession;
+                    log.log(Level.FINE, () -> "more than 1 day old: " + sessionId);
+                    try {
+                        localSession = getSessionFromFile(sessionId);
+                    } catch (Exception e) {
+                        log.log(Level.FINE, () -> "could not get session from file: " + sessionId + ": " + e.getMessage());
+                        continue;
+                    }
+                    Optional<ApplicationId> applicationId = localSession.getOptionalApplicationId();
+                    if (applicationId.isEmpty()) continue;
```
jonmv, line 734: Are these expected states?

jonmv (commented):
I like this a lot better. I believe there could still be cases where we delete a remote session, but not the local one, but that it would only happen if something else was wrong (`UNKNOWN` session status, or inconsistent state (`ACTION` session which isn't "the active session" of an application).

hmusum:
I think I have addressed most of you comments, there are some corner cases and perhaps excessive checking before deleting in the code, but rather safe than sorry...

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,97 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+                boolean activeForApplication = sessionIsActiveForApplication.test(session);
+                if (status == ACTIVATE && activeForApplication) continue;
+
+                Instant createTime = session.getCreateTime();
+                boolean hasExpired = hasExpired(createTime);
+                if ( ! hasExpired) continue;
+
+                log.log(Level.FINE, () -> "Remote session " + sessionId + " for " + tenantName + " has expired, deleting it");
+                deleteRemoteSessionFromZooKeeper(session);
+                deleted++;
```
jonmv, line 715: So I guess this is racy: status and active-session are written in one transaction, but here we read them one after the other, so we could potentially see `PREPARED` and `true`, and then go on to delete the remote session, if it had spent a long time in prepare (not impossible when a session fails to activate because of host provisioning). 
I guess we could handle that by simply re-reading status if `activeForApplication` but `status != ACTIVATE`, or re-reading active-session in the other case? If they still disagree, we should probably warn about it?

jonmv: I wonder if we should also bail out if they disagree, unless the session `isOldAndCanBeDeleted(...)`, like for local sessions?

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -682,6 +682,97 @@ private boolean sessionLifeTimeElapsed(Instant created) {
...
+
+                log.log(Level.FINE, () -> "Expired local session is candidate for deletion: " + sessionId +
+                        ", created: " + createTime + ", status " + status + ", can be deleted: " + canBeDeleted(sessionId, status));
+                // Need to check if it can be deleted, as we might have a (local) session in UNKNOWN
+                // state that we should not delete
```
jonmv, line 720: Do you remember how we could end up in this state?

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
jonmv: Hmm, we could also read the session status as `PREPARED`, and active as `false`, but then the session is activated while we delete it. I don't remember without looking at the code, but I believe those writes are protected by a lock. Alternatively, we could do the deletion with a zknode version condition ... which I'm not sure we expose through our own APIs >_<

hmusum: Will do this while holding session lock instead

hmusum:

jonmv (approved):
Wohoo!

PR #32097, opened 2024-08-09
Interns/languageserver

Mangern:
This PR includes the language server we worked on for the last 8 weeks. Most changes are in the integrations directory, and consists of a language server for the Schema language, as well as clients for VSCode and IntelliJ Ultimate.
We also made minor changes to config-model: setting some fields to public instead of package-private, to be used in our parser implementation. It only affects classes in com.yahoo.schema.parser. 

Refer to README.md in integration/schema-language-server.


Mangern:
@arnej27959 Some files in config-model are changed, with fields set to public. Please confirm that this is OK.

Mangern:
@arnej27959 Can you confirm that our changes in config-model are OK?

bratseth (approved):

PR #32139, opened 2024-08-13
Remove direct dependency on github.com/alessio/shellescape

mpolden:
It was easier to simply remove our use of this library. There's still an
indirect dependency on this, but that's not our responsibility:

```
$ go mod why github.com/alessio/shellescape
# github.com/alessio/shellescape
github.com/vespa-engine/vespa/client/go/internal/cli/auth
github.com/zalando/go-keyring
github.com/alessio/shellescape
```

Closes #32044.

@baldersheim


baldersheim (dismissed):
Even better with less dependencies, but there is a merge conflict.

PR #32146, opened 2024-08-14
Migrate workflow to delete old Cloudsmith artifacts to GitHub Actions

glebashnik:

## What

Adds a GitHub action workflow to delete old Cloudsmith artifacts.
The workflow is triggered either manually or on a daily schedule.

## Why

Part of the process of retiring Screwdriver as CI/CD runner.

## Additional Info

The corresponding Screwdriver pipeline is not removed since we don't have a dashboard for Github Actions yet.

See successful test run:
https://github.com/vespa-engine/vespa/actions/runs/10387039260/job/28759628746

### DIFF
--- a/.github/scripts/delete-old-cloudsmith-artifacts.sh
+++ b/.github/scripts/delete-old-cloudsmith-artifacts.sh
@@ -30,9 +30,7 @@ done
 echo "Deleting the following RPMs:"
 cat $RPMS_TO_DELETE
 
-echo $CLOUDSMITH_API_CREDS
-
-if [[ -n $SCREWDRIVER ]] && [[ -z $SD_PULL_REQUEST ]]; then
+if [[ -n $GITHUB ]] && [[ -z $SD_PULL_REQUEST ]]; then
     for RPMID in $(cat $RPMS_TO_DELETE); do
       curl -sLf -u "$CLOUDSMITH_API_CREDS" -X DELETE \
         --header 'accept: application/json' \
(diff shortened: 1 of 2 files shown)


glebashnik:
@bjormel Could you look at this test run and verify that it deleted what it should and no more:
https://github.com/vespa-engine/vespa/actions/runs/10387039260/job/28759628746

I haven't changed the core logic so should be ok but just in case.

.github/scripts/delete-old-cloudsmith-artifacts.sh
```diff
@@ -0,0 +1,42 @@
...
+
+for VERSION in $VERSIONS_TO_DELETE; do
+  curl -sLf --header 'accept: application/json' \
+    "https://api.cloudsmith.io/v1/packages/vespa/open-source-rpms/?query=version:${VERSION}" | jq -re '.[] | .slug' >> $RPMS_TO_DELETE
+done
+
+echo "Deleting the following RPMs:"
+cat $RPMS_TO_DELETE
+
+if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
```
esolitos, line 33: This should actually be the other way around! It should **not** delete during a PR. 😅 

```suggestion
if [ "$GITHUB_EVENT_NAME" != "pull_request" ]; then
```

esolitos (changes requested):
Found an actual bug there!

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: Alternatively:
```suggestion
if [ "$GITHUB_EVENT_NAME" == "schedule" ]; then
```

glebashnik: Nice, thanks!

esolitos (approved):

PR #32151, opened 2024-08-14
Extracting memory info from /proc/pid/smaps is very costly so reduce …

baldersheim:
…frequency to once every minute.

@geirst PR


geirst (approved):
👍

PR #32152, opened 2024-08-14
- Only require similarity of 1 percent, or a minimum of 1_Mi. a fixed…

baldersheim:
… limit of 1_Mi is to strict when you are operating with plenty of Gb.

@geirst PR


geirst (approved):
👍

Issue #32153, opened 2024-08-14
Slow convergence of nodes due to download of file references sometimes taking a long time


hmusum:
Reported on Slack:

* The convergence of all types of nodes are fast (sub seconds) except container. And the convergence time for container is multiple of 20s. It looks like there's a 20s timeouts somewhere. When connected to the correct config server, it returns immediately. Most of times, it returns in about 20s, sometimes 40s. (I have configured 3 config servers). (edited) 
* It seems the distribution of files with type path to the container node takes multiple of 20s to happen during app convergence.

hmusum:
Note: The improvement needed has been known, but impact was not expected to be this large. Note that nodes eventually converge, but it takes longer than necessary.

papa99do:
I've printed out the debug log for filedistribution package on 3 config servers, and 1 container server, and sorted them in chronological order. I also added some comments about my understanding of what happened. Hope this analysis can help pinpoint the cause of this issue. 

```
# I have 3 config servers node0 (172.19.0.7), node1 (172.19.0.8), node2 (172.19.0.2) and 1 container node4 (172.19.0.3) in this setup
# Session 49 was created on node0, after the session was active, node4 need to download two json files `50b8474089c54d6/marqo_index_settings.json` and `569b5a7c041bcb4a/marqo_index_settings_history.json` to reload the components
# In this scenario, node4 managed to download the first file immediately from node1 relayed to node0, but waited 20s to download the second file.
# In the worst scenario, it waited 20s for each document, taking 40s in total to download both of them 

# session 49 being prepared on node0
1723647539.683213	node0.vespanet	16/113	configserver	Container.com.yahoo.vespa.config.server.filedistribution.FileDirectory	debug	Adding file /opt/vespa/var/db/vespa/config_server/serverdb/tenants/default/sessions/49 (created 2024-08-14T14:58:59.078960009Z, modified 2024-08-14T14:58:59.310960009Z, size 4096)
1723647539.683317	node0.vespanet	16/113	configserver	Container.com.yahoo.vespa.config.server.filedistribution.FileDirectory	debug	Copying /opt/vespa/var/db/vespa/config_server/serverdb/tenants/default/sessions/49 to /opt/vespa/var/db/vespa/filedistribution/writing5797371890501607843/49
[...]

PR #32154, opened 2024-08-15
Download file reference directly from source

hmusum:
When getting a 'filedistribution.setFileReferencesToDownload' RPC request, use the fact that the client is the one that has the file to download from that same source instead of potentially trying some other config server that might not have it.

See https://github.com/vespa-engine/vespa/issues/32153

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/rpc/RpcServer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/rpc/RpcServer.java
@@ -574,14 +574,19 @@ public class RpcServer implements Runnable, ConfigActivationListener, TenantList
     private void setFileReferencesToDownload(Request req) {
         req.detach();
         rpcAuthorizer.authorizeFileRequest(req)
-                .thenRun(() -> { // okay to do in authorizer thread as downloadIfNeeded is async
+                .thenRun(() -> { // okay to do in authorizer thread as downloadFromSource is async
                     String[] fileReferenceStrings = req.parameters().get(0).asStringArray();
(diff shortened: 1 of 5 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/rpc/RpcServer.java
```diff
@@ -574,14 +574,19 @@ private void serveFile(Request request) {
...
+                    var client = req.target();
+                    var peerSpec = client.peerSpec();
                     Stream.of(fileReferenceStrings)
                             .map(FileReference::new)
-                            .forEach(fileReference -> downloader.downloadIfNeeded(
-                                    new FileReferenceDownload(fileReference,
-                                                              req.target().toString(),
-                                                              false /* downloadFromOtherSourceIfNotFound */)));
+                            .forEach(fileReference -> {
+                                log.log(FINE, "Downloading file reference " + fileReference.value() + " from " + peerSpec);
```
baldersheim, line 587: I would log after downloadFromSource and include the return value in the log message.

baldersheim (dismissed):
Looks good, just one comment to add some info in log message.

configserver/src/main/java/com/yahoo/vespa/config/server/rpc/RpcServer.java
hmusum: Yes, good idea, done

hmusum:
@baldersheim please take a look again

baldersheim (approved):

PR #32167, opened 2024-08-19
Hmusum/require compression types in request

hmusum:
We have supplied compression type in request for a long time, we are allowed to changes this after 2 months (see https://vespa.ai/releases#versions)


baldersheim (approved):
Which compression types are supported, and what is default.

hmusum:
These ones:
`private static final List<CompressionType> compressionTypesToServe = compressionTypesAsList(List.of("zstd", "lz4", "gzip")); // In preferred order
`

Client supplies the set of compression types it supports (which has been the same list as above for a long time now), so in effect `zstd` will be used.

PR #32168, opened 2024-08-19
Use msbIdx for direct lookup, and then additional check if prev is la…

baldersheim:
…rge enough.

@vekterli or @jonmv PR
This supeceeds #31993

### DIFF
--- a/vespalib/src/tests/stllike/hashtable_test.cpp
+++ b/vespalib/src/tests/stllike/hashtable_test.cpp
@@ -72,6 +72,8 @@ TEST("require that getModuloStl always return a larger number in 32 bit integer 
         size_t num = 1ul << i;
         size_t prime = hashtable_base::getModuloStl(num);
         EXPECT_GREATER_EQUAL(prime, num);
+        EXPECT_EQUAL(prime, hashtable_base::getModuloStl(prime));
+        EXPECT_GREATER(hashtable_base::getModuloStl(prime+1), prime + 1);
         printf("%lu <= %lu\n", num, prime);
     }
     for (size_t i=0; i < 32; i++) {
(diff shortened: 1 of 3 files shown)


jonmv (dismissed):

vespalib/src/tests/stllike/hashtable_test.cpp
```diff
@@ -69,9 +67,24 @@ TEST("require that hashtable<int> can be copied") {
...
+        EXPECT_GREATER_EQUAL(prime, num);
+        printf("%lu <= %lu\n", num, prime);
+    }
+    for (size_t i=0; i < 32; i++) {
+        size_t num = (1ul << i) - 1;
+        size_t prime = hashtable_base::getModuloStl(num);
+        EXPECT_GREATER_EQUAL(prime, num);
+        printf("%lu <= %lu\n", num, prime);
+    }
+}
```
jonmv, line 83: Perhaps also test all the primes, and their ++1?

baldersheim: Done

vespalib/src/tests/stllike/hashtable_test.cpp
```diff
@@ -69,9 +67,26 @@ TEST("require that hashtable<int> can be copied") {
     EXPECT_EQUAL(42, *table2.find(42));
 }
 
+TEST("require that getModuloStl always return a larger number in 32 bit integer range") {
+    for (size_t i=0; i < 32; i++) {
+        size_t num = 1ul << i;
+        size_t prime = hashtable_base::getModuloStl(num);
+        EXPECT_GREATER_EQUAL(prime, num);
+        EXPECT_EQUAL(prime, hashtable_base::getModuloStl(prime));
+        EXPECT_GREATER(hashtable_base::getModuloStl(prime+1), prime + 1);
```
jonmv, line 76: This can only work because `0 < 7` (for `i = 31`), but ok 😁

vekterli (commented):
:+1: just to verify, I ran a brute force check between the old and new code for all values in `[0, UINT32_MAX]` and the outputs were identical for the same input size

PR #32174, opened 2024-08-19
Detect conflict between tensor field and non-tensor field in fieldset.

toregge:
@geirst : please review
@hmusum : FYI


geirst (approved):
👍

PR #32182, opened 2024-08-20
Handle sessions with no corresponding application id

hmusum:


jonmv (approved):
Ah, yes, I guess it has to be like that.

PR #32183, opened 2024-08-20
Use 16x unrolling for generic floating point squared euclidean distance computations

vekterli:
@baldersheim please review. Noticed this when looking at the generated ARM NEON machine code.

This lets the auto-vectorized code use all 4 32-bit lanes of a 128-bit SIMD register instead of just 2, which doubles performance compared to `double` (which is in line with what could be expected).

### DIFF
--- a/vespalib/src/vespa/vespalib/hwaccelerated/generic.cpp
+++ b/vespalib/src/vespa/vespalib/hwaccelerated/generic.cpp
@@ -169,12 +169,12 @@ GenericAccelrator::squaredEuclideanDistance(const int8_t * a, const int8_t * b, 
 
 double
 GenericAccelrator::squaredEuclideanDistance(const float * a, const float * b, size_t sz) const noexcept {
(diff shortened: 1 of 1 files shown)


baldersheim (dismissed):

vekterli:
After some experiments, a 16x unrolling factor seems to be optimal for both `float` and `double`. Further increases either has no effect or _reduces_ performance.

Before:
```
double : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 589
float  : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 570
int8_t : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 82
```

After:
```
double : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 121
float  : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 81
int8_t : sum=2610046000000.000000 of N=1000000 and vector length=1000 took 82
```

PR #32186, opened 2024-08-20
Avoid this escape.

baldersheim:
@bjorncs PR

### DIFF
--- a/vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
+++ b/vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
@@ -164,9 +164,7 @@ public abstract class Maintainer implements Runnable {
 
         Instant now = clock.instant();
         long nodeOffset = clusterHostnames.indexOf(hostname) * interval.toMillis() / clusterHostnames.size();
-        String c = getClass().getName();
-        System.out.println("class=" + c);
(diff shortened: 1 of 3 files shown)


vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
```diff
@@ -164,7 +164,9 @@ Duration staggeredDelay(Duration interval, String hostname, List<String> cluster
 
         Instant now = clock.instant();
         long nodeOffset = clusterHostnames.indexOf(hostname) * interval.toMillis() / clusterHostnames.size();
-        long maintainerOffset = getClass().getName().hashCode() % interval.toMillis();
+        String c = getClass().getName();
+        System.out.println("class=" + c);
```
bjorncs, line 168: `System.out.println` intentional?

bjorncs (dismissed):

vespajlib/src/main/java/com/yahoo/concurrent/maintenance/Maintainer.java
baldersheim: Not at all ... 👎

baldersheim: Fixed, thanks

PR #32195, opened 2024-08-20
Use std::string directly.

toregge:
@baldersheim : please review
@geirst : FYI


baldersheim (approved):
This is not  reviewable by a human, so factory is the judge.

geirst:
💯

PR #32204, opened 2024-08-21
Add OpenNLP config to serverdb

bratseth:
@aressem is this sufficient to get the config file in this module into serverdb?


aressem:
This should be enough so that it ends up in `/opt/vespa/share/vespa/configdefinitions/`.

aressem (approved):

PR #32215, opened 2024-08-21
Handle file reference not found as a temporary error

hmusum:
So that request for file reference is retried if not found in 1 minute

### DIFF
--- a/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/FileDistributionRpcServer.java
+++ b/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/FileDistributionRpcServer.java
@@ -77,7 +77,7 @@ class FileDistributionRpcServer {
     private static final int baseErrorCode = 0x10000;
     private static final int baseFileProviderErrorCode = baseErrorCode + 0x1000;
 
-    private static final int fileReferenceDoesNotExists = baseFileProviderErrorCode;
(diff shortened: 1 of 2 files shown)


config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/FileDistributionRpcServer.java
```diff
@@ -77,7 +77,7 @@ private void declareMethods() {
     private static final int baseErrorCode = 0x10000;
     private static final int baseFileProviderErrorCode = baseErrorCode + 0x1000;
 
-    private static final int fileReferenceDoesNotExists = baseFileProviderErrorCode;
+    private static final int fileReferenceNotFound = baseFileProviderErrorCode;
```
baldersheim, line 80: should this be capital letters, like ABORT, CONNECTION etc ?
I am not sure what is java practice.

baldersheim (approved):

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/FileDistributionRpcServer.java
hmusum: Both are used, but should probably be capital letters, yes. I want to do some more cleanup related to this, so will keep this in mind in a future PR

PR #32217, opened 2024-08-22
feat: add container image for query-builder client application

esolitos:
## What

This adds a simple Dockerfile and pipeline to build a container `ghcr.io/vespa-engine/vespa/query-builder` container image.

## Why

This provides a ready-made tool for the developers to use when testing/exploring Vespa, without the need to clone this repo and having node installed locally.

## Additional Info

As this is a developer-oriented tool all the container image tags are prefixed with `dev-`, to clarify that this is not to be used in a production environment.

### DIFF
--- a/client/js/app/Dockerfile
+++ b/client/js/app/Dockerfile
@@ -1,0 +1,10 @@
+FROM  node:lts-alpine
+
+COPY . /app
+
+WORKDIR /app
+
+RUN yarn install && \
+  yarn build
+
+CMD [ "yarn", "preview", "--host" ]
--- a/.github/workflows/query-builder-container.yml
+++ b/.github/workflows/query-builder-container.yml
@@ -1,0 +1,70 @@
+name: Build "Query Builder" Container
+
+on:
+  push:
+    tags:
+      - v*
+  pull_request:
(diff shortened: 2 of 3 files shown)


aressem (dismissed):

client/js/app/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM  node:lts-alpine
+
+COPY . /app
+
+WORKDIR /app
+
+RUN yarn install && \
+  yarn build
+
+CMD [ "yarn", "dev", "--host" ]
```
freva, line 10: This should be `yarn preview --port 3000 --host`, dev will start server with hot module reload, it wont use the built code from `yarn build`.

Consider adding `--port 3000` here: https://github.com/vespa-engine/vespa/blob/6cb5d731084c13cebdfd7e4e413143d87c9511f0/client/js/app/package.json#L8 for symmetry with `yarn dev`, then the command here only needs to replace `dev` with `preview`

esolitos: Sounds good, I updated accordingly. 😊

freva (approved):

PR #32221, opened 2024-08-22
remove tenant secrets secret store integration

tokle:


configserver/src/main/java/com/yahoo/vespa/config/server/http/SecretStoreValidator.java
```diff
@@ -25,22 +24,19 @@
...
 
     // http for easier testing. VespaHttpClient rewrites to https
     private static final String PROTOCOL = "http://";
     private static final String AWS_PARAMETER_VALIDATION_HANDLER_POSTFIX = ":4080/validate-secret-store";
-    private final SecretStore secretStore;
     private final CloseableHttpClient httpClient = VespaHttpClientBuilder.custom().buildClient();;
 
-    public SecretStoreValidator(SecretStore secretStore) {
-        this.secretStore = secretStore;
+    public SecretStoreValidator() {
```
gjoranv, line 36: Empty public ctor can be removed.

gjoranv (approved):

PR #32223, opened 2024-08-22
Fix significance model resolver, better tests, handling of upper case in query words

glebashnik:

Implements the following logic for resolving models:
  - When language is explicitly tagged on query
      - Use language if available from the model registry, fail otherwise.
      - If “un” try both “un” and “en”.
  - When language is implicitly detected
      - Use language if available from the model registry. Fallback to “un” then “en”, fail if none are available.

Handling of query words containing upper case chars.

### DIFF
--- a/container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
+++ b/container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
@@ -301,8 +301,9 @@ public class SignificanceSearcherTest {
                      errorMessage.getDetailedMessage());
     }
 
-    // Naming convention for tests that follow:
-    // Explicit language - language set for a query
+    // Tests that follow verify model resolving logic in different scenarios.
+    // Test naming convention is as follows:
+    // Explicit language - language set in a query
     // Implicit language - automatically detected language
     // Missing language - language without a model
     // Unknown language - Language.UNKNOWN
--- a/container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
+++ b/container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
@@ -2,13 +2,11 @@
 package com.yahoo.search.significance.test;
 
 import com.yahoo.component.chain.Chain;
-import com.yahoo.config.subscription.ConfigGetter;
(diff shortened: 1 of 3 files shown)


container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
```diff
@@ -305,50 +303,112 @@ public void failsOnConflictingSignificanceConfiguration() {
...
+        var result = searchWordWithLanguage(existingWord, Optional.of(explicitLanguage), Optional.empty());
 
-        Query q2 = new Query();
-        q2.getModel().setLanguage(Language.FRENCH);
-        q2.getRanking().setProfile("significance-ranking");
-        AndItem root2 = new AndItem();
-        WordItem tmp2;
-        tmp2 = new WordItem("hello", true);
-        root2.addItem(tmp2);
+        var resultRoot = (AndItem) result.getQuery().getModel().getQueryTree().getRoot();
```
geirst, line 341: Consider adding a helper function that extracts the DocumentFrequency from the result, to avoid these 3 lines being duplicated in all test cases.

geirst (dismissed):
👍 Just one comment to improve readability of the tests.

container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
glebashnik: Makes sense. Extracted two methods and removed duplication.

container-search/src/test/java/com/yahoo/search/significance/test/SignificanceSearcherTest.java
```diff
@@ -332,27 +332,34 @@ private Optional<DocumentFrequency> getDocumentFrequencyWithEnglish(String word)
         return makeDocumentFrequency(model.documentFrequency(word));
     }
 
+    private static WordItem getFistWord(Result result) {
```
geirst, line 335: Should this be `getFirstWord`?

glebashnik: Yep, autocomplete :). Fixed it now.

geirst (approved):

PR #32225, opened 2024-08-22
Avoid this escape.

baldersheim:
@vekterli PR


documentapi/src/main/java/com/yahoo/documentapi/ProgressToken.java
```diff
@@ -75,59 +75,49 @@ public void setState(BucketState state) {
...
-                // MSBs differ
-                return ((key >>> 63) > (other.key >>> 63)) ? 1 : -1;
-            }
-            // Mask off MSBs since we've already checked them, and with MSB != 1
-            // we know the ordering will be consistent
-            if ((key & 0x7FFFFFFFFFFFFFFFL) < (other.key & 0x7FFFFFFFFFFFFFFFL)) {
-                return -1;
-            } else if ((key & 0x7FFFFFFFFFFFFFFFL) > (other.key & 0x7FFFFFFFFFFFFFFFL)) {
-                return 1;
+                if ((key & 0x8000000000000000L) != (other.key & 0x8000000000000000L)) {
```
vekterli, line 86: Unintentional added indent?

vekterli (commented):
My main concern is that all of these classes are contained in `@PublicAPI` packages, so removing or `final`-izing classes (top-level classes in particular) might be a breaking change. Other than that LGTM.

vekterli (approved):
:+1:

documentapi/src/main/java/com/yahoo/documentapi/ProgressToken.java
baldersheim: fixed

PR #32279, opened 2024-08-27
Make the gram root settable

bratseth:
Less work to support it than to keep explaining how to do it in your app.


arnej27959 (approved):
👍

PR #32287, opened 2024-08-28
fix: fetch all git references in publish-cli workflow - MERGEOK

esolitos:
## What

In the "Publish CLI" action adds [the `fetch-depth: 0` option](https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches) to the checkout github actions.

## Why

The deploy process does a check to see if there has been any changes in the codebase, but with a shallow clone it cannot do that, by setting `fetch-depth: 0` the GH Action will fetch also all references.

--- 

### DIFF
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -89,6 +89,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - uses: actions/download-artifact@v4
         with:


bjormel (approved):

freva:
Please observe merge stops even for changes that don't affect the build because each merge to this repo triggers a new Vespa version build followed by all the system/performance/integration tests that take a few hours to run, this means that when a fix (for the issue that caused the merge stop) is merged, it may take longer to get it out, which makes the merge stop even longer :)

esolitos:
Good to know, sorry about that!

PR #32293, opened 2024-08-29
Use filename in URL when storing downloaded file on disk

hmusum:
Rationale: To be able to decompress or transform a downloaded file based on file suffix

@toregge please review


hmusum:
@toregge please take a look again

toregge (approved):
:+1:

PR #32297, opened 2024-08-29
Remove cluster state version downgrade fallback

vekterli:
@geirst please review

To avoid inherent race conditions with overlapping cluster controller leader intervals (caused by the old leader not yet knowing it has been deposed) where both an old state version and a newer state version is concurrently published, we want to only accept strictly increasing version numbers (for the lifetime of a process; these are currently not durably stored on content nodes). On the cluster controllers themselves, this version number is backed by a ZooKeeper quorum, ensuring that it _is_ durably stored where it matters the most.

A content node only observing strictly increasing version numbers is an invariant that holds _unless_ an explicit fallback is triggered, where we can still accept an older version.

This fallback was intended to be a "failsafe" if ZooKeeper state on the cluster controllers was lost, but its implementation depended on information that is not actually present in all CC RPCs, meaning that it could kick in even when not intended, thus rendering the race condition protection void.

The CC RPC in question is not easily extensible, so instead remove the fallback entirely. This has the bonus of content nodes actually being able to rely on the version invariant internally. Downside is that content node and distributor processes must be restarted to accept a lower state version upon ZK state loss, but in that case you probably have bigger problems.

### DIFF
--- a/storage/src/tests/storageserver/statemanagertest.cpp
+++ b/storage/src/tests/storageserver/statemanagertest.cpp
@@ -33,10 +33,8 @@ struct StateManagerTest : Test, NodeStateReporter {
     void SetUp() override;
     void TearDown() override;
 
-    static std::shared_ptr<api::SetSystemStateCommand> make_set_state_cmd(std::string_view state_str, uint16_t cc_index) {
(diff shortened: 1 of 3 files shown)


geirst (approved):
👍

PR #32317, opened 2024-09-02
fix bug in StructuredParser

yjtakamabe:
use explicitIndex for parsing correctly.
bug introduced in https://github.com/vespa-engine/vespa/pull/23686


bratseth (approved):
Thanks a lot and sorry for any trouble this caused you.

yjtakamabe:
Thank you for your approval.

PR #32324, opened 2024-09-03
Fix: Language server publish github action

theodorklauritzen:
Include the compiled typescript code to the extension. Then the github action will not throw an error while publishing.


bratseth (commented):
Isn't "out" a directory for generated artefacts, which should therefore be ignored?

theodorklauritzen:
It should be ignored by git, but not by `.vscodeignore`. If the out folder is ignored in this file, the generated artefacts will not be included in the extension. And then the extension will not have any source code.

bratseth:
Got it.

bratseth (approved):

PR #32354, opened 2024-09-05
LSP: Verify that attribute functions are referencing fields that are stored as an attribute

theodorklauritzen:
Add a quick fix to automatically fix fields not stored as an attribute, but references to as an attribute.


Mangern (dismissed):
good

Mangern (approved):
Grammar is correct.

PR #32359, opened 2024-09-05
Refactor LSP release action

theodorklauritzen:
Make the LSP release action fetch the latest release from GitHub releases rather than storing the latest version in the source code.


Mangern (approved):
lgtm

PR #32364, opened 2024-09-05
Interns/magnus/documentation java

Mangern:
Move documentation fetching (for hover) from Python to Java. It now runs on the client in the background when the language server starts.
Documentation fetching is still tested during build.


theodorklauritzen (dismissed):
Glimrende

theodorklauritzen (approved):

PR #32369, opened 2024-09-06
Use session lifetime when calculating if a session is new

hmusum:
Makes it possible to test with smaller session lifetime


hmusum:
@aressem FYI, should fix failing system test

aressem (approved):

PR #32370, opened 2024-09-06
Move directory, then delete it

hmusum:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
@@ -94,6 +94,7 @@ import static com.yahoo.vespa.config.server.session.Session.Status.PREPARE;
 import static com.yahoo.vespa.config.server.session.Session.Status.UNKNOWN;
 import static com.yahoo.vespa.curator.Curator.CompletionWaiter;
 import static com.yahoo.vespa.flags.Dimension.INSTANCE_ID;
+import static com.yahoo.yolean.Exceptions.uncheck;
 import static java.nio.file.Files.readAttributes;
 
 /**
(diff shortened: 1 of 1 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
```diff
@@ -1172,18 +1173,15 @@ private interface FileOperation extends Transaction.Operation {
...
-        }
+    private record DeleteOperation(java.nio.file.Path pathToDelete) implements FileOperation {
 
         @Override
         public void commit() {
-            // TODO: Check delete access in prepare()
-            IOUtils.recursiveDeleteDir(new File(pathToDelete));
+            var tempDir = uncheck(() -> Files.createTempDirectory("delete"));
+            uncheck(() -> {
+                Files.move(pathToDelete, tempDir, StandardCopyOption.ATOMIC_MOVE);
```
hakonhall, line 1182: The temporary directory and pathToDelete are not guaranteed to be on the same file system, and in fact is not in hosted(?), and hence not atomic?

hakonhall (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionRepository.java
hmusum: We set java.io.tmpdir in startup scripts to make sure they are on the same file system (but I agree that this is not obvious from the code, so maybe we should have some common code that does this and documents behavior)

hakonhall: I think sessions ends up under /opt/vespa/var/db, while java.io.tmpdir is /opt/vespa/var/tmp?  Those are different mount points in hosted.  So I think this move will end up throwing AtomicMoveNotSupportedException.

You want to avoid seeing a half-deleted directory `pathToDelete`?  You could perhaps rename pathToDelete to another name in the same directory or to /opt/vespa/var/db, and then delete?

hmusum: Yes, thanks, you are right, will change to use a temp dir one level above sessions

hmusum:
@hakonhall please take a look again

hakonhall (approved):

PR #32371, opened 2024-09-06
Include concurrent operation rejection reason in error message

vekterli:
@geirst please review.

Rejection can happen either from concurrent mutations or due to reindexing being in progress for the bucket containing the document. Distinguish between these two cases in the returned error message to make debugging the underlying rejection reason easier.


geirst (approved):
👍

PR #32385, opened 2024-09-09
Bratseth/symmetric normalizing

bratseth:


bratseth:
Please review but don't merge.

arnej27959 (approved):
is there a good reason why the new stem() API has a different argument ordering (swapped StemMode and Language)?  otherwise, looks good to me.

bratseth:
It's to harmonize with the order used in tokenize.

PR #32390, opened 2024-09-11
Use a Python venv for language server build

oyving:
This creates a virtual environment that Python can install requests into, and use that to run the build script.

As a next step we should probably only use this with a specific profile so it's not used on every developer laptop.


theodorklauritzen:
@Mangern started to solve this issue last week and are almost finished, see PR #32364. This will remove python from the build stage. This will probably we ready on Friday. If this is too late, we can merge this PR

oyving:
Great! Then we can close this PR 👍

PR #32398, opened 2024-09-13
Interns/magnus/documentation java

Mangern:
Move documentation fetching (for hover) from Python to Java. It now runs on the client in the background when the language server starts.
Documentation fetching is still tested during build


Mangern:
We had to add some dependencies to the vespa-dependencies-enforcer, not sure if it is the correct way.

theodorklauritzen (dismissed):
This should cover the whole python script 👍

tokle (dismissed):
👍

theodorklauritzen (approved):
LGTM

PR #32413, opened 2024-09-17
Ensure deferred tasks are aborted if blocking messages are explicitly cleared

vekterli:
@geirst please review.

When a node becomes unavailable in the cluster state (or disappears from the distribution config), pending operations towards it are currently explicitly cleared in the pending message tracker. The code responsible for clearing messages was not functionally symmetric with the code that handles replies in how deferred tasks were handled.

Normally a deferred task expects to be run once there exist no further pending operations towards a particular bucket, or to be aborted at any point in time. This was handled as expected during reply processing and process shutdown, but not when clearing messages for individual nodes.

In particular, this meant that it was possible for deferred tasks scheduled by reindexing visitors to stall feeding to certain buckets when they raced with nodes becoming unavailable, as such tasks hold bucket-level feed locks.

With these changes, clearing operations for a node implicitly aborts deferred tasks towards all buckets touched by pending operations to the cleared node.

### DIFF
--- a/storage/src/tests/distributor/pendingmessagetrackertest.cpp
+++ b/storage/src/tests/distributor/pendingmessagetrackertest.cpp
@@ -35,22 +35,29 @@ namespace {
 
 class RequestBuilder {
     uint16_t _toNode;
+    uint32_t _bucket_idx;
     std::chrono::milliseconds _atTime;
 public:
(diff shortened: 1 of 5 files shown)


geirst (approved):
👍

PR #32418, opened 2024-09-17
Add unit test for zc compression.

toregge:
@geirst : please review


toregge:
Rebased and updated in #32425

PR #32425, opened 2024-09-18
Add unit test for zc compression.

toregge:
Avoid buffer overflow when encoding values >= 256Mi.

@geirst : please review


geirst (approved):
👍

PR #32426, opened 2024-09-18
Fix warning message for paged attributes and remote storage

hmusum:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/application/validation/PagedAttributesRemoteStorageValidator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/application/validation/PagedAttributesRemoteStorageValidator.java
@@ -60,6 +60,7 @@ public class PagedAttributesRemoteStorageValidator implements Validator {
     private static String join(List<Attribute> fields) {
         var ret = fields.stream()
                 .map(Attribute::getName)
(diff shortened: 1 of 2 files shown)


arnej27959 (dismissed):
we should probably cap the list if it's too long, to avoid unreadable message

config-model/src/main/java/com/yahoo/vespa/model/application/validation/PagedAttributesRemoteStorageValidator.java
```diff
@@ -56,11 +57,12 @@ private static void validatePagedAttributes(DeployLogger logger, String clusterN
                 .formatted(clusterName, join(fields)));
     }
 
-    private static String join(Set<Attribute> fields) {
-        return fields.stream()
+    private static String join(List<Attribute> fields) {
+        var ret = fields.stream()
```
arnej27959, line 61: add `.limit(10)` to actually limit also?

hmusum: Thanks, f**ked up commit, so had to redo it, apparently lost the most important line when doing that.

arnej27959 (approved):
👍

PR #32433, opened 2024-09-19
fix(cli): only publish if changes detected since last release

### LINKED ISSUE #32365: Vespa CLI 8.405.15 did not get marked as latest release
👋 looks like 8.405.15 did not get marked as latest release, raise this issue for some awareness. Thanks!

relates to https://github.com/Homebrew/homebrew-core/pull/183574

esolitos:
## What

Changes on the "Publish Vespa CLI" Github workflow:

1. The publish job now checks if there are any changes between the last release tag and the new version being published. This is done by using the `git diff` command to compare the changes in the `client/go` directory.
2. If no changes are detected, the workflow will skip the publish step.
3. The workflow also handles the case where a dev version (starting with "0.0.0-dev") is being used, in which case the changes check is skipped.

## Why

This ensures that the CLI is only published when there are actual changes, preventing unnecessary releases.

A further reason is that we had 2 separate systems checking for changes, one for the Github Releases and another one for Homebrew, this leads to inconsistencies, so this PR merges the validation into an initial check.

## Additional info

- Fixes #32365

### DIFF
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -53,11 +53,21 @@ jobs:
       - name: Check for Changes
         id: check-diff
         env:
-          VERSION_PREFIX: 'v8*' # We need the 8 or else "vespa-7." will match and sort before "v8.x.y"
+          VERSION_PATTERN: '^v8\.' # We need the 8 prefix or else "vespa-7." will match and sort before "v8.x.y"
+          GH_TOKEN: ${{ github.token }}
         run: |
+          set -x
+
           # Checks for changes since the last release.
           new_tag="v${{ steps.meta.outputs.version }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-cli.yml
```diff
@@ -80,8 +112,8 @@ jobs:
   publish:
     runs-on: macos-latest
 
-    # Publish the CLI when a tag is pushed or a workflow is triggered manually.
-    if: contains(fromJSON('["push", "workflow_dispatch"]'), github.event_name)
+    # Publish the CLI only if changes were detected and the event was a push or workflow_dispatch.
+    if: needs.prepare.outputs.changes-detected == 'true' && contains(fromJSON('["push", "workflow_dispatch"]'), github.event_name)
```
freva, line 116: Maybe not require a diff if this is a workflow_dispatch?

freva (dismissed):

freva (dismissed):

freva (approved):

PR #32438, opened 2024-09-20
group independent rename

andreer:


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/HostRenamer.java
```diff
@@ -33,51 +35,57 @@ public HostRenamer(NodeRepository nodeRepository, Duration interval, Metric metr
         this.hostnameSchemeFlag = Flags.HOSTNAME_SCHEME.bindTo(nodeRepository.flagSource());
     }
 
+    record ApplicationGroup(ApplicationId applicationId, Integer groupIndex){};
```
mpolden, line 38: Ideally this should be `ClusterGroup(ClusterId id, int groupIndex)` so that for example container cluster group 0 and content cluster group 0 within the same app can be retired simultaneously (there's already a `ClusterId` that can be used for this)

mpolden (approved):
Great!

PR #32446, opened 2024-09-21
Revert "add new dependencies to intellij plugin and set infinity until version"

hmusum:
Reverts vespa-engine/vespa#32440

Build fails with:
`invalid major version: lsp-v2`

so I assume this is caused by this change


aressem (approved):

hmusum:
@Mangern FYI

PR #32453, opened 2024-09-23
Flag to override s1 token

tokle:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -559,6 +560,13 @@ public class Flags {
             "Initial backoff time in milliseconds when failing to download a file reference",
             "Takes effect on restart of Docker container");
 
+    public static final UnboundBooleanFlag OVERRIDE_S1_TOKEN = defineFeatureFlag(
+            "override-s1-token", false,
+            List.of("mortent"), "2024-09-23", "2024-11-01",
+            "Override the S1 token",
+            "Takes effect on next host-admin run",
+            ZONE_ID, HOSTNAME);
```
hakonhall, line 568: Avoid adding ZONE_ID unless you actually want to override zone dimension when evaluating the flag [in the controller].

hakonhall (dismissed):

hakonhall (approved):

PR #32476, opened 2024-09-26
Language server: target newest versions of IntelliJ

Mangern:
This change was reverted earlier because it was suspected it broke the build. The broken build was fixed later.


theodorklauritzen (approved):
LGTM

PR #32477, opened 2024-09-26
fix ringbuffer expand() code

arnej27959:
@jonmv please review
the issue here was: only one of two cases was handled
case A) `newOffSet == offset`; enough room for second arraycopy after newOffset + toWrite
case B) `newOffset == offset + size`; `newOffset + toWrite == newSize`; triggered exception in second arrraycopy.
removed `if (toWrite < size)` because it seems unlikely and arraycopy handles copying 0 bytes.


jonmv (approved):
Yikes, thanks!

PR #32480, opened 2024-09-26
vespa-logfmt now shows all levels by default;

arnej27959:
also extend unit tests and fix a corner case

@hmusum please review


hmusum (approved):
LGTM

PR #32490, opened 2024-09-30
Inject AsmSecretStore

lesters:
@gjoranv For discussion.

Note: moved `asm-secret.def` from `jdisc-cloud-aws` to `container-disc`.


gjoranv (dismissed):

gjoranv:
You have to move the config to a separate Java package and add an export-package for it.

lesters:
Added a export package for that config (and moved it to another package)

gjoranv (approved):

PR #32491, opened 2024-09-30
Require positive proof of an index field to segment

bratseth:


container-search/src/main/java/com/yahoo/search/yql/YqlParser.java
```diff
@@ -1558,7 +1560,7 @@ private Item instantiateWordItem(String field,
     }
 
     private boolean shouldSegment(String field, boolean fromQuery) {
-        return fromQuery && ! indexFactsSession.getIndex(indexNameExpander.expand(field)).isAttribute();
```
bratseth, line 1561: This is the actual change.

arnej27959 (approved):
looks good to me

PR #32492, opened 2024-09-30
Fix class name in script

hmusum:
Not used in any system tests, will extend a test to use it if this is a tool we still want to support


vekterli (approved):
:+1: we should check what information this (legacy) tool actually emits—it's possible that most, if not all, of it is available via backend metrics instead

hmusum:
Decided after some discussion to remove this tool instead: https://github.com/vespa-engine/vespa/pull/32498

PR #32503, opened 2024-10-02
fix: improve handling of failed dnf list command

esolitos:
## What

Updates the script to delete old artifacts in Cloudsmith:

- Add a `|| true` after the `dnf list` command to ensure that the script continues to execute even if the `dnf list` command returns non-zero status code.
- Adds the standard set of bash options at the beginning of the script.

## Why

- fixes an issue where if there are no artifacts to remove the Workflow would be marked as failed

## Additional info

Failed workflows: https://github.com/vespa-engine/vespa/actions/workflows/delete-old-versions-in-archive.yml?query=is%3Afailure+branch%3Amaster

---

### DIFF
--- a/.github/scripts/delete-old-cloudsmith-artifacts.sh
+++ b/.github/scripts/delete-old-cloudsmith-artifacts.sh
@@ -54,10 +54,7 @@ done
 
 echo
 if [[ "${GITHUB_EVENT_NAME}" == "schedule" || "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
-    echo '\033[1;33mDeleting the following RPMs:\033[0m'
-    echo
-    cat "${RPMS_TO_DELETE}"
-    echo
+    printf '\033[1;33mDeleting the following RPMs:\033[0m\n\n' "${RPMS_TO_DELETE}"
 
     while IFS= read -r RPMID
     do
@@ -68,8 +65,7 @@ if [[ "${GITHUB_EVENT_NAME}" == "schedule" || "${GITHUB_EVENT_NAME}" == "workflo
         "https://api.cloudsmith.io/v1/packages/vespa/open-source-rpms/${RPMID}/"
     done < "${RPMS_TO_DELETE}"
 else
-    echo '\033[1;32m#### Dry-run mode ####\033[0m'
-    echo '\033[0;32mWould have deleted the following RPMs:\033[0m'
-    echo
+    printf '\033[1;32m#### Dry-run mode ####\033[0m\n'
+    printf '\033[0;32mWould have deleted the following RPMs:\033[0m\n\n'
     cat "${RPMS_TO_DELETE}"
 fi
--- a/.github/scripts/delete-old-cloudsmith-artifacts.sh
+++ b/.github/scripts/delete-old-cloudsmith-artifacts.sh
@@ -1,8 +1,16 @@
 #!/bin/bash
 # Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+
+if [ "${RUNNER_DEBUG:-}" == "1" ]; then
+    set -o xtrace
+fi
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
 dnf install -y 'dnf-command(config-manager)' jq
 
-set -euo pipefail
 MAX_NUMBER_OF_RELEASES=32
 
 # Cloudsmith repo
@@ -12,7 +20,7 @@ dnf config-manager --add-repo '/tmp/vespa-open-source-rpms.repo'
 rm -f /tmp/vespa-open-source-rpms.repo
 
 # Allow the last Vespa 7 release to remain in the repo
-VERSIONS_TO_DELETE=$(dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa | awk '/[0-9].*\.[0-9].*\.[0-9].*/{print $2}' | sort -V | grep -v "7.594.36" | head -n -$MAX_NUMBER_OF_RELEASES)
+VERSIONS_TO_DELETE=$(dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa || true | awk '/[0-9].*\.[0-9].*\.[0-9].*/{print $2}' | sort -V | grep -v "7.594.36" | head -n -$MAX_NUMBER_OF_RELEASES)
 
 if [[ -z "$VERSIONS_TO_DELETE" ]]; then
   echo "No old RPM versions to delete found. Exiting."
@@ -38,5 +46,3 @@ if [[ "$GITHUB_EVENT_NAME" == "schedule" || "$GITHUB_EVENT_NAME" == "workflow_di
         "https://api.cloudsmith.io/v1/packages/vespa/open-source-rpms/$RPMID/"
     done
 fi
-
-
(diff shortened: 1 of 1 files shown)


bjormel (dismissed):

.github/scripts/delete-old-cloudsmith-artifacts.sh
```diff
@@ -1,42 +1,67 @@
...
-# Allow the last Vespa 7 release to remain in the repo
-VERSIONS_TO_DELETE=$(dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa | awk '/[0-9].*\.[0-9].*\.[0-9].*/{print $2}' | sort -V | grep -v "7.594.36" | head -n -$MAX_NUMBER_OF_RELEASES)
+VESPA_VERSIONS=$(dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa || true)
+if [[ -z "$VESPA_VERSIONS" ]]; then
+  echo "No Vespa versions found, nothing to do. Exiting."
+  exit 0
+fi
 
+# Note: Allow the last Vespa 7 release to remain in the repo!
+VERSIONS_TO_DELETE=$(echo "${VESPA_VERSIONS}" | awk '/[0-9].*\.[0-9].*\.[0-9].*/{print $2}' | sort -V | grep -v "${LAST_VESPA_7_RELEASE}" | head -n -$MAX_NUMBER_OF_RELEASES)
```
hakonhall, line 32: Over-generalized regex - you probably mean `/[0-9]+\.[0-9]+\.[0-9]+/` ?

.github/scripts/delete-old-cloudsmith-artifacts.sh
```diff
@@ -1,42 +1,67 @@
...
-echo "Deleting the following RPMs:"
-cat $RPMS_TO_DELETE
-
+echo
 if [[ "$GITHUB_EVENT_NAME" == "schedule" || "$GITHUB_EVENT_NAME" == "workflow_dispatch" ]]; then
-    for RPMID in $(cat $RPMS_TO_DELETE); do
-      curl -sSLf -X DELETE \
+    echo '\033[1;33mDeleting the following RPMs:\033[0m'
+    echo
+    cat "${RPMS_TO_DELETE}"
```
hakonhall, line 51: I see you sometimes use curly braces when referencing parameters and sometimes not (when either is semantically equivalent).  If you want my 2 cents I prefer to avoid curly braces as it is easier on the eyes and fingers.

In a somewhat similar vein, I prefer to always reference parameters inside double-quotes (unless that would be semantically wrong), possibly with the exception of integer variables :thinking: .

hakonhall (dismissed):

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: I just copied the existing regex and indeed I like your suggestion better. 

I'll tell you what.. Given we want to handle the 8+ only, I'll give you one even better:

```suggestion
VERSIONS_TO_DELETE=$(echo "${VESPA_VERSIONS}" | awk '/[8-9]\.[0-9]+\.[0-9]+/{print $2}' | sort -V | head -n -$MAX_NUMBER_OF_RELEASES)
```

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: > I see you sometimes use curly braces when referencing parameters and sometimes not (when either is semantically equivalent).

The "sometimes not" is only where I might have forgotten or copied some code from somewhere and not noticed. 
My approach is to use it _always_, for:
- consistency and not need a context switch to quickly identify variables
- using them everywhere possible avoids the need to remember edge cases _(like $9 is ok and $10 isn't)_
- it is consistent with other language templates (python for example)


>  I prefer to always reference parameters inside double-quotes (unless that would be semantically wrong), possibly with the exception of integer variables

That's also my take, if you see around I tend do always quote any usage of variables. Similarly to the `${}`, where they are lacking I probably missed them (unless it was required for the specific usage of the var) or was some code taken from an existing implementation. 😅

.github/scripts/delete-old-cloudsmith-artifacts.sh
hakonhall: Sweet

.github/scripts/delete-old-cloudsmith-artifacts.sh
hakonhall: I think you're in the minority on always wanting to use curly braces.

.github/scripts/delete-old-cloudsmith-artifacts.sh
```diff
@@ -1,42 +1,74 @@
...
-curl -1sLf 'https://dl.cloudsmith.io/public/vespa/open-source-rpms/config.rpm.txt?distro=el&codename=8' > /tmp/vespa-open-source-rpms.repo
+curl  --silent --show-error --location --fail \
+  --tlsv1 --output /tmp/vespa-open-source-rpms.repo \
+  'https://dl.cloudsmith.io/public/vespa/open-source-rpms/config.rpm.txt?distro=el&codename=8'
 dnf config-manager --add-repo '/tmp/vespa-open-source-rpms.repo'
 rm -f /tmp/vespa-open-source-rpms.repo
 
-# Allow the last Vespa 7 release to remain in the repo
-VERSIONS_TO_DELETE=$(dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa | awk '/[0-9].*\.[0-9].*\.[0-9].*/{print $2}' | sort -V | grep -v "7.594.36" | head -n -$MAX_NUMBER_OF_RELEASES)
+dnf list -y --quiet --showduplicates --disablerepo='*' --enablerepo=vespa-open-source-rpms vespa >/tmp/dnf-stdout 2>/tmp/dnf-stderr || true
```
hakonhall, line 24: By testing for failure I was thinking of testing the exit status, so something like

```
if DNF_STDOUT=$(dnf list -y --quiet --showduplicates); then
  ...
else
  ... failure, be sure to exit with some message, whether something was written to stderr or not
fi
```

Usually, a failing command writes something to stderr, so it is equivalent.

hakonhall (dismissed):

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: My understanding was that it is best practice. 🙃 

I'll leave this here and tackle this discussion when we look into a linter/code-style as part of [VESPANG-511](https://vespaai.atlassian.net/browse/VESPANG-511).

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: I believe your solution is better because `dnf` does not seem to care about the `--quiet`  option and _only the first run_ it writes _"Importing GPG key 0x0D35DA7B"_ to stderr... 😒

.github/scripts/delete-old-cloudsmith-artifacts.sh
hakonhall: :+1:

.github/scripts/delete-old-cloudsmith-artifacts.sh
esolitos: I solved it by adding `dnf makecache` before, so that it's pre-cached. 🙃

hakonhall (approved):

PR #32507, opened 2024-10-02
prevent NPE after close()

arnej27959:
@bjorncs please review


container-search/src/main/java/com/yahoo/search/result/HitGroup.java
```diff
@@ -970,7 +970,7 @@ public void addDataListener(Runnable runnable) {
     @Override
     public void close() {
         super.close();
-        hits = null;
+        hits = new ListenableArrayList<>(0);
```
bjorncs, line 973: As a performance optimization consider adding a static "null" instance as a constant to `ListenableArrayList` (e.g `ListenableArrayList.EMPTY`).

bjorncs (approved):

PR #32509, opened 2024-10-02
keep original index name if uncertain

arnej27959:
* if there are several index names differing only in casing, keep the original from the request.

@hmusum please review
@geirst @bratseth FYI


bratseth (approved):
👍

hmusum:
👍

PR #32520, opened 2024-10-04
feat: setting to set java path in IntelliJ plugin

Mangern:
This makes it possible to set a custom path to the java installation used for running the language server in intellij.


Mangern:
@bratseth Reference to the mail we got.

theodorklauritzen (approved):
LGTM

PR #32521, opened 2024-10-04
Fixed hardcoded doc ids in significance model generator

glebashnik:

- Fixed mismatch between doc type registered for parsing and the one in doc id.
- Removed doc type as parameter, a dummy doc type is used instead since we are only interested in content.

### DIFF
--- a/vespaclient-java/src/main/java/com/yahoo/vespasignificance/ClientParameters.java
+++ b/vespaclient-java/src/main/java/com/yahoo/vespasignificance/ClientParameters.java
@@ -23,9 +23,6 @@ public class ClientParameters {
     // Language for the program
     public final String language;
 
-    // Document type identifier
-    public final String docType;
-
     // Zstandard compression
     public final boolean zstCompression;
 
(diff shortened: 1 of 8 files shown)


document/src/main/java/com/yahoo/document/json/JsonReader.java
```diff
@@ -83,6 +84,35 @@ ParsedDocumentOperation readSingleDocument(DocumentOperationType operationType,
         return operation;
     }
 
+    /**
+     * Reads a single operation. The operation is not expected to be part of an array.
+     *
+     * @return the parsed document operation
+     */
+    public ParsedDocumentOperation readSingleDocument() {
```
geirst, line 92: Please add unit test for the new function in document/src/test/java/com/yahoo/document/json/JsonReaderTestCase.java.

geirst (changes requested):
Overall looks good, please add unit test.

glebashnik:
@geirst New implementation, avoids changes in JsonReader.
Uses dummy doc type and doc id instead.
They don't matter as long as doc type is in doc id.
We are only interested in content.
Made unit tests for model generator more robust.

geirst (approved):

PR #32522, opened 2024-10-04
Bratseth/binarize

bratseth:


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/BinarizeExpression.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @author bratseth
+ */
+public class BinarizeExpression extends Expression  {
+
+    private final double threshold;
+
+    //private DataType targetType;
+
+    /** The type this bth consumes and produces. */
```
arnej27959, line 25: bth -> both

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/BinarizeExpression.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param threshold the value which the tensor cell value must be larger than to be set to 1 and not 0.
+     */
+    public BinarizeExpression(double threshold) {
+        super(TensorDataType.any());
+        this.threshold = threshold;
+    }
+
+    @Override
+    public void setStatementOutput(DocumentType documentType, Field field) {
+//        if (! (field.getDataType() instanceof TensorDataType))
```
arnej27959, line 40: we want to pick up both the input and output tensor types and ensure they are compatible; we need to make sure we can extend this so binarize can take a `tensor<float>(x[24])` as input and produce `tensor<int8>(y[3])` as output in the future.  It looks like it might be possible to do something here, why did you comment out the check?

arnej27959 (commented):
I'm a bit worried about how future-proof this is, looks like a WIP

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/BinarizeExpression.java
bratseth: My plan is to create a separate pack_bits for that.

Longer term I want to support arbitrary tensor expressions in indexing, and then we can't derive the behavior of an expression (i.e whether it should pack) from the output type.

As it is, this expression works with any tensor and does not modify the type, so there is nothing to check (I think?).

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/BinarizeExpression.java
bratseth: Thanks fixed

bratseth:
> I'm a bit worried about how future-proof this is, looks like a WIP

I wanted to keep this simple and dumb specifically for future-proofing, but maybe you have something else in mind?

Could argue that as this is just a special case of applying a tensor expression we should rather focus on that ...

arnej27959 (approved):
👍

PR #32524, opened 2024-10-04
Interns/magnus/tensorexpressions

Mangern:
### Language server
This PR specifically fixes the case:

expression: tensor<float>(d0[1],d1[10])(d1)
where d1 is a reference to the tensor dimension in the tensor expression body.

Earlier the d1 was correctly marked as a reference in the body, but the dimensions were not marked as definitions in the type definition.


theodorklauritzen (approved):
LGTM

PR #32529, opened 2024-10-04
more hex support in mixed tensor parsing

arnej27959:
* now also handles hex input with out-of-order dimensions
* also, add support for short-form mixed tensors with several mapped dimensions, using basically the same syntax as C++ implementation
* refactor and consolidate the various places handling different cell types
* detect garbage after ending brace
* fix various unit tests that had tensors with extra ending brace

@bratseth please review
@lesters @havardpe FYI


bratseth (approved):
Great!

PR #32537, opened 2024-10-07
Bratseth/pack bits

bratseth:
@geirst please review. 

Sorry for the many commits, this was a bit exploratory.

Now that we have contextual type resolving, I want to replace the earlier half-baked attempts at type resolving by it, but this diff is already large enough.

### DIFF
--- a/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/AnyDataType.java
+++ b/indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/AnyDataType.java
@@ -1,3 +1,4 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.vespa.indexinglanguage.expressions;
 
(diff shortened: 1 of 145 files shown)


bratseth:
No rush, but have you seen this @geirst ?

bratseth:
Take your time @geirst ... :-)

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/AnyDataType.java
```diff
@@ -0,0 +1,29 @@
+package com.yahoo.vespa.indexinglanguage.expressions;
```
geirst, line 1: Please add copyright.

geirst (dismissed):
👍 I am not sure if the new input / output type resolving in StatementExpression.doVerify() is tested by existing unit tests or if new tests should be added (especially error situations).

bratseth:
Thanks! Yes, more work is needed on this type resolving for sure. This is just the bare minimum to support both embed and pack_bits in the same statement.

PR #32539, opened 2024-10-08
Add UNKNOWN_CONFIG_DEFINITION and FORWARD_ALL_LOG_LEVELS to Permanent…

hmusum:
…Flags

Moving these from Flags to PermanentFlags. Removing first flag will require a lot of work that will take time before being able to remove, second one we probably need to keep permanently until we have had running for a long time, to get some experience with how this works out in practice.


arnej27959 (approved):
👍

PR #32549, opened 2024-10-08
Disable creation of '._' files in tarball on mac os

gjoranv:
See:
https://superuser.com/questions/61185/why-do-i-get-files-like-foo-in-my-tarball-on-os-x

I got the problem with dot-underscore files in the unpacked application packages again after wiping the mac. This time, the 'com.apple.quarantine' xattr was not set, but instead 'com.apple.provenance'. The latter seems impossible to remove. Hence, this solution.


mpolden (dismissed):

gjoranv:
Unfortunately, this seems to fail on Linux:
```

[ERROR]   CompressedApplicationInputStreamTest.require_that_gnu_tared_file_can_be_unpacked:95->assertTestApp:67 expected:<3> but was:<0>
[ERROR]   CompressedApplicationInputStreamTest.require_that_nested_app_can_be_unpacked:107 expected:<5> but was:<0>

```

hmusum (approved):

gjoranv:
I tried checking out the 'vespa' git repo from scratch, but the same thing happened. I wonder if it could be because I use https (with git-credentials-manager) instead of SSH.

PR #32553, opened 2024-10-09
Move to use permanent flags

hmusum:


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -385,13 +385,6 @@ public class Flags {
             "Takes effect at redeployment",
             INSTANCE_ID);
 
-    public static final UnboundStringFlag UNKNOWN_CONFIG_DEFINITION = defineStringFlag(
```
hakonhall, line 388: There should never be 2 definitions of a flag.  AFAIK the only problem is that a random one will be chosen for any given JVM. I'll make a PR to catch this in unit testing.

hakonhall (approved):

PR #32555, opened 2024-10-09
Use timestamp TaS conditions from Document V1 handler when possible

vekterli:
@jonmv please review 👁️👄👁️

Timestamp-predicated test-and-set operations are cheaper to evaluate in the backend, are safe in the face of concurrent feeding and work as expected when the batch selection uses imported fields (since the selection itself is only evaluated during the visiting phase, _not_ during the mutating operation processing phase).

Since we now receive backend persistence timestamps as part of visiting, patch these into the generated `TestAndSetCondition` instances sent for each resulting update/remove operation (iff present).

To allow for backwards compatibility, the generated TaS condition has _both_ a required timestamp as well as the original batch operation selection. The selection is only used as a fallback if backends are on a version too old to support timestamp requirements. On nodes with timestamp TaS support, the selection is _ignored_ entirely when a required timestamp is present.

Also note that if a content node is too old to support sending persistence timestamps, it will default to 0 the receiver which also preserves legacy behavior.

### DIFF
--- a/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
+++ b/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
@@ -1018,6 +1018,71 @@ public class DocumentV1ApiTest {
         driver.close();
     }
 
+    @Test
+    public void batch_update_rewrites_tas_condition_with_timestamp_predicate_if_provided_by_backend() {
+        var driver = new RequestHandlerTestDriver(handler); // try-with-resources hangs the test on assertion failure, which isn't optimal
+        List<AckToken> tokens = List.of(new AckToken(null), new AckToken(null), new AckToken(null), new AckToken(null));
+        long backendTimestamp = 1234567890;
+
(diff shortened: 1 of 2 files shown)


vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/DocumentV1ApiTest.java
```diff
@@ -1018,6 +1018,71 @@ public void testResponses() {
...
+            var expectedCondition = TestAndSetCondition.ofRequiredTimestampWithSelectionFallback(backendTimestamp, "optimist");
+            assertEquals(expectedCondition, ((DocumentUpdate) update).getCondition());
+            parameters.responseHandler().get().handleResponse(new UpdateResponse(0, false));
+            return new Result();
+        });
+        var response = driver.sendRequest("http://localhost/document/v1/space/music/docid?selection=optimist&cluster=content&timeChunk=10", PUT,
+                """
+                        {
+                          "fields": {
+                            "artist": { "assign": "Jahn Teigen" }
```
jonmv, line 1045: <3

jonmv (approved):
# 💯

PR #32579, opened 2024-10-11
fix: correctly unset untilBuild in intellij

Mangern:
Due to reasons explained [here](https://intellij-support.jetbrains.com/hc/en-us/community/posts/18554337303186-How-can-we-declare-a-plugin-compatible-with-future-versions) we need to explicitly set untilBuild to null to make the intellij plugin compatible with all future versions of IntelliJ.


theodorklauritzen (approved):
LGTM

PR #32581, opened 2024-10-11
Rename attribute writer OnWriteDoneType to OnWriteDoneConstRefType.

toregge:
@geirst : please review


toregge:
After discussion, it was decided on a slightly larger step for the change.

PR #32606, opened 2024-10-17
less throw of UnsupportedOperation

arnej27959:
when the query asks for something that can't be done, throw an exception that indicates that the problem is the input, and won't cause a stack trace to be shown.

@bratseth please review


bratseth (approved):
👍

PR #32616, opened 2024-10-19
Revert "Get rid of the screwdriver stuff."

hmusum:
Reverts vespa-engine/vespa#32615


hmusum:
Attempted fix in https://github.com/vespaai/yahoo-vespa-yahoo/pull/112

PR #32617, opened 2024-10-19
Revert "unsupported operation should give a nice error"

arnej27959:
Reverts vespa-engine/vespa#32572

after discussion with @bratseth we will instead change to throwing IllegalInputException in those places where the query does something we don't accept, and use UnsupportedOperation only for cases where it indicates a bug in Java code (our own or plugins).  If we see stack traces in search results we should ideally investigate and make some sort of fix.

@bjorncs @hmusum FYI
@bratseth review and merge


hmusum (approved):

bjorncs (commented):
👍

PR #32626, opened 2024-10-21
extend RankingExpression grammar

arnej27959:
* allow alternate mixed-tensor literal form

this should help for tensor playground.

@lesters please review
@havardpe FYI


lesters (approved):
LGTM!

PR #32628, opened 2024-10-22
Revert "Bratseth/pack bits"

aressem:
Reverts vespa-engine/vespa#32537

Suspecting that this broke a few system tests.


bratseth:
Thanks. Not unlikely ...

bratseth (approved):

PR #32637, opened 2024-10-22
Record time of each snapshot event

mpolden:
@hakonhall

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/backup/Snapshot.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/backup/Snapshot.java
@@ -27,10 +27,18 @@ public record Snapshot(SnapshotId id, HostName hostname, State state, Instant cr
     }
 
     public Snapshot with(State state) {
-        if (state.compareTo(this.state) < 0) {
+        if (!canChangeTo(state)) {
             throw new IllegalArgumentException("Cannot change state of " + this + " to " + state);
         }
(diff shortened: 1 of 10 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/backup/Snapshot.java
```diff
@@ -1,36 +1,86 @@
...
+        if (!canChangeTo(state)) {
             throw new IllegalArgumentException("Cannot change state of " + this + " to " + state);
         }
-        return new Snapshot(id, hostname, state, createdAt, cluster, clusterIndex);
+        return new Snapshot(id, hostname, state, history.with(state, at), cluster, clusterIndex);
+    }
+
+    private boolean canChangeTo(State state) {
+        // Allow repeated restores
+        if (state == State.restoring && this.state == State.restored) return true;
```
hakonhall, line 40: This doesn't sound like something we want to always allow?
Perhaps an operator should force-change the state away from restored first?

hakonhall (approved):

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/backup/Snapshot.java
mpolden: You mean restoring->restored->created->restoring? I'll think about it. Right now I want to allow this for testing purposes.

PR #32648, opened 2024-10-24
Secret store subclass

gjoranv:


jdisc-cloud-aws/src/main/java/ai/vespa/secret/aws/AsmTenantSecretReader.java
```diff
@@ -0,0 +1,58 @@
+package ai.vespa.secret.aws;
```
lesters, line 1: copyright header?

lesters (approved):
LGTM!

jdisc-cloud-aws/src/main/java/ai/vespa/secret/aws/AsmTenantSecretReader.java
gjoranv: Thanks, will fix in a new PR

PR #32664, opened 2024-10-25
Interns/magnus/servicesxml

Mangern:
This adds initial basic support when editing the services.xml file by extending the LemMinX language server.
This means it only works in vscode and standalone installations, in particular not IntelliJ yet.
Current functionality:

- Docs on hover
- Validation of the services.xml specification
- Go to definition on document types.


theodorklauritzen (approved):
LGTM

PR #32667, opened 2024-10-25
Interns/magnus/servicesxml

Mangern:
Checks if documents exist in the application when they are added to services.xml.
Also adds config-model as a dependency.


theodorklauritzen (approved):
LGTM

PR #32671, opened 2024-10-25
Interns/magnus/lemminx vespa

Mangern:
Fixes a bug where XML support crashes due to wrong .rng files being extracted.


theodorklauritzen (approved):
LGTM

PR #32682, opened 2024-10-26
Do not reduce number of distribution bits in a running system

hmusum:
Reducing distribution bits may not always be safe, as previously disjoint bucket subtrees are joined back into the same subtrees which will be problematic, as we assume unique timestamps per unique document in a given bucket.


vekterli (approved):
:+1: nice!

PR #32691, opened 2024-10-28
Clear conntrack table flag

hakonhall:


andreer (approved):
We should only clear the conntrack table for ipv6 + udp though.

hakonhall:
Yes, I wondered if I should mentioned that in the flag, but decided against it for being too verbose.  But it is taken care of in the next PR.

PR #32729, opened 2024-10-31
use separate limits for separate strategies

havardpe:
note that not scoring stop words inside weak and is not yet supported

@arnej27959 please review


arnej27959 (approved):
looks good to me

PR #32741, opened 2024-11-01
Interns/magnus/resolve symbols

Mangern:
Language server: tries to resolve more symbol references than what was done earlier after each parse.
Should fix errors related to not detecting rank profiles.


theodorklauritzen (approved):
LGTM

PR #32748, opened 2024-11-02
erase max score from vectorized terms where appropriate

havardpe:
also change how initial threshold adjusting works; choose the word closest to the adjust limit.

@arnej27959 please review


arnej27959 (approved):
👍

PR #32791, opened 2024-11-06
create new feature flag allow-status-page

lperry25:
create new feature flag allow-status-page to allow tenants to easily see status pages

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
@@ -367,6 +367,12 @@ public class PermanentFlags {
             "Takes effect on next redeployment",
             INSTANCE_ID);
 
+    public static final UnboundBooleanFlag ALLOW_STATUS_PAGE = defineFeatureFlag(
+            "allow-status-page", false,
+            "Shows link to status page for nodes of a specific tenant",
+            "takes effect on browser reload of api/user/v1/user",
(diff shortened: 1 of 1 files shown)


flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
```diff
@@ -367,6 +367,12 @@ public class PermanentFlags {
             "Takes effect on next redeployment",
             INSTANCE_ID);
 
+    public static final UnboundBooleanFlag ALLOW_STATUS_PAGE = defineFeatureFlag(
+            "allow-status-page", false,
+            "Shows link to status page for nodes of a specific tenant",
```
freva, line 372: "status page" is a little vague, this should at least mention that this is the "cluster controller status page"

lperry25: I updated the description and renamed the flag now

freva (dismissed):

kkraune:
This is not necessarily just for the cluster controller status pages, it is also for searchnode in another view

kkraune (changes requested):
rename to more generic

freva:
Reverted back to generic name, added TENANT_ID dimension. PTAL @kkraune

kkraune:
looks good to me!

freva (approved):

aressem (approved):

kkraune (approved):

PR #32798, opened 2024-11-06
enable avoiding cores on stdio errors

havardpe:
@hmusum please review


havardpe:
the onnx model analyzer should already catch std::exception which should avoid cores

hmusum (approved):
Nice

PR #32810, opened 2024-11-07
MERGEOK Allow 'any' as output from for_each

bratseth:


hmusum (approved):
LGTM. @geirst might want to have a look

geirst (commented):
👍

PR #32812, opened 2024-11-07
take care to never resurrect removed children

arnej27959:
@havardpe or @geirst please review


havardpe (approved):

arnej27959:
We decided to fix this in another way.

PR #32813, opened 2024-11-08
No point in depending on epel-release because other requrements depen…

aressem:
…d on epel-release being installed before the vespa packages.


toregge (approved):
:+1:

PR #32814, opened 2024-11-08
Feat: Add experimental support for yql files

theodorklauritzen:
This adds support for:
- Parsing of YQL queries in `.yql`files
- Simple semantic tokens for YQL
- Simple semantic token for grouping in YQL
- Code action to run YQL queries diretcly from editors
- New abstract Node class that YQLNode and SchemaNode inherits from


Mangern (approved):
Probably good we merged this now

PR #32815, opened 2024-11-08
Interns/magnus/lemminx vespa

Mangern:
XML server: code action to create document, completion on document types.


theodorklauritzen (approved):
LGTM

PR #32818, opened 2024-11-08
retry missing docsums to all other nodes

arnej27959:
This could solve the issue where documents move between nodes; I have tested it manually with redundancy=2 and searchable-copies=2 to see that it could work.
We should discuss this more on Monday, so do not merge for now.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/dispatch/rpc/RpcProtobufFillInvoker.java
+++ b/container-search/src/main/java/com/yahoo/search/dispatch/rpc/RpcProtobufFillInvoker.java
@@ -222,8 +222,6 @@ public class RpcProtobufFillInvoker extends FillInvoker {
         }
     }
 
-    private static int monkey = 0;
-
     private List<FastHit> fill(Result result, List<FastHit> hits, String summaryClass, byte[] payload, boolean ignoreErrors) {
         try {
             var protobuf = SearchProtocol.DocsumReply.parseFrom(payload);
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/search/dispatch/rpc/RpcProtobufFillInvoker.java
```diff
@@ -135,34 +141,59 @@ private void sendDocsumsRequest(int nodeId, List<FastHit> hits, byte[] payload,
...
+                        }
+                    }
+                }
+                if (retryMap.size() > 0) {
+                    sendFillRequestByNode(result, summaryClass, retryMap);
+                }
+                while (outstandingResponses > 0) {
+                    var responseAndHits = getNextResponse(result);
+                    if (responseAndHits == null) {
+                        log.log(Level.WARNING, "Timed out waiting for summary data. " + outstandingResponses + " responses outstanding.");
```
bjorncs, line 188: Do we want to log as 'warning' or could 'fine' suffice? Could a container overload scenario be worsen by excessive logging?

container-search/src/main/java/com/yahoo/search/dispatch/rpc/RpcProtobufFillInvoker.java
```diff
@@ -135,34 +143,73 @@ private void sendDocsumsRequest(int nodeId, List<FastHit> hits, byte[] payload,
...
-                skippedHits += processResponse(result, response, hitsContext, summaryClass);
+                skippedHits.addAll(processOneResponse(result, responseAndHits, summaryClass, false));
                 outstandingResponses--;
             }
-            if (skippedHits != 0) {
+            if (skippedHits.isEmpty()) {
+                // all done OK
+                return;
+            }
+            int numSkipped = skippedHits.size();
```
geirst, line 179: I suggest documenting why retrying docsum fetch is needed in some cases to better explain the following code. Also consider putting all of the retrying in a separate function to make it easier to read processResponses().

havardpe:
@andreer FYI

bjorncs (approved):

geirst (approved):

PR #32822, opened 2024-11-09
Bratseth/indexing type inference 2

bratseth:


bratseth:
I'll let factory have a look

PR #32833, opened 2024-11-11
Support setting zookeeper preAllocSize with feature flag

hmusum:

This will make it possible to reduce this setting significantly, to avoid writing an unnecessarily huge amount of data in e.g. system tests

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -484,6 +484,13 @@ public class Flags {
             "Whether node snapshots should be created when host storage is discarded",
             "Takes effect immediately");
 
+    public static final UnboundLongFlag ZOOKEEPER_PRE_ALLOC_SIZE = defineLongFlag(
+            "zookeeper-pre-alloc-size", 65536,
+            List.of("hmusum"), "2024-11-11", "2025-01-11",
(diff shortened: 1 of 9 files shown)


flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -484,6 +484,13 @@ public class Flags {
             "Whether node snapshots should be created when host storage is discarded",
             "Takes effect immediately");
 
+    public static final UnboundLongFlag ZOOKEEPER_PRE_ALLOC_SIZE = defineLongFlag(
+            "zookeeper-pre-alloc-size", 65536,
+            List.of("hmusum"), "2024-11-11", "2025-01-11",
+            "Setting for zookeeper.preAllocSize flag, can be reduced from default value "
+            + "e.g. when running tests to avoid writing a large, sparse, mostly unused file",
+            "Takes effect on restart of Docker container");
```
vekterli, line 492: Would be nice if this explicitly mentioned the size unit, since it's not immediately obvious if this is in bytes or KiB (maybe the flag name should mention it as well?).

vekterli (dismissed):
:+1:

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
hmusum: Yes, agree, done.

vekterli (approved):

PR #32834, opened 2024-11-11
add "vespa auth show" command

arnej27959:
@mpolden please review


mpolden (dismissed):
LGTM, but a test would be nice. I think it can be simple as `api_key_test.go` + mocking of the user V1 response.

mpolden (dismissed):

mpolden (approved):

PR #32842, opened 2024-11-12
Glebashnik/feed field generator

glebashnik:


model-integration/src/main/java/ai/vespa/generative/LanguageModelGenerator.java
```diff
@@ -0,0 +1,29 @@
+package ai.vespa.generative;
```
bratseth, line 1: Same, I think we could put this in llm.generation

And maybe, since what separates it from Generator isn't that it's generating from a language model, but that it's taking an additional config specifying a prompt, we could call it something like PromptedGenerator, or ConfiguredGenerator?

linguistics/src/main/java/com/yahoo/language/process/Generator.java
```diff
@@ -0,0 +1,128 @@
...
+    default Map<String, Generator> asMap() {
+        return asMap(defaultGeneratorId);
+    }
+
+    // Returns this generator instance as a map with the given name
+    default Map<String, Generator> asMap(String name) {
+        return Map.of(name, this);
+    }
+
+    String generate(String prompt, Context context, DataType dataType);
```
bratseth, line 29: Please remove the DataType argument, I don't think you need it and sine it is a document side concept it makes it weird to use this interface in other contexts.

bratseth, line 29: Why not use the Prompt class here so we have the option of passing prompts through that aren't just strings.

model-integration/src/main/java/ai/vespa/llm/clients/LocalLLM.java
```diff
@@ -31,7 +34,7 @@
  *
  * @author lesters
  */
-public class LocalLLM extends AbstractComponent implements LanguageModel {
+public class LocalLLM extends AbstractComponent implements LanguageModel, Generator {
```
bratseth, line 37: The role of Generator is to apply a LanguageModel to a generation task specified by a context. This is a more general API so it's a bit messy for it to know about that context (and also to call itself through a utility defined elsewhere). I see the purpose though ... maybe there's a better solution, but I need to get some breakfast now.

indexinglanguage/src/main/javacc/IndexingParser.jj
```diff
@@ -33,6 +33,7 @@ import com.yahoo.text.StringUtilities;
 import com.yahoo.vespa.indexinglanguage.expressions.*;
```
bratseth, line 33: Unfortunately, atm we have two parsers. You need to also make these same additions in IndexingParser.ccc in schema-language-server.

model-integration/src/main/java/ai/vespa/llm/generation/OnnxEncoderDecoderTextGenerator.java
```diff
@@ -21,16 +22,14 @@
...
- * See llm.generator.def for configurable parameters.
- *
+ * Component for generating text using SentencePiece tokenizer and encoder and decoder language models in ONNX format.
+ * See onnx-encoder-decoder-text-generator.def for configuration options.
+ * TODO: Consider if this should be a LanguageModel implementation instead of a TextGenerator.
+ * 
  * @author lesters
  */
 @Beta
-public class Generator extends AbstractComponent {
```
glebashnik, line 33: Renaming it to something more concrete.
It's beta and this class is not used anywhere so I assume renaming is ok.

model-integration/src/main/java/ai/vespa/llm/generation/LanguageModelTextGenerator.java
```diff
@@ -0,0 +1,81 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.logging.Logger;
+
+/**
+ * Text generator that uses a language model, configured with a config definition.
+ *
+ * @author glebashnik
+ */
+public class LanguageModelTextGenerator extends AbstractComponent implements TextGenerator {
```
glebashnik, line 22: A bit verbose naming but quite precise.
I also considered LMTextGenerator.

linguistics/src/main/java/com/yahoo/language/process/TextGenerator.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * Generates text given a prompt.
+ * 
+ * @author glebashnik
+ */
+public interface TextGenerator {
```
glebashnik, line 16: Renamed from Generator to TextGenerator since Generator is too general.

glebashnik:
@bratseth Let me know if you think the design and naming are good enough.
Left some explanations in comments here.

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/GenerateExpression.java
```diff
@@ -82,7 +82,7 @@ public DataType setOutputType(DataType outputType, VerificationContext context)
             throw new VerificationException(this, "Generate expression requires either a string or array<string> output type, but got "
                     + outputType.getName());
         
-        super.setOutputType(null, outputType, null, context); // todo: Why not set actualOutput to outputType?
+        super.setOutputType(null, outputType, null, context); // TODO: Why not set actualOutput to outputType?
```
bratseth, line 85: It won't make any difference when required and actual is the same.
(The reason for separating them is just that the required type may be a supertype of the actually produced type.)

bratseth (dismissed):
Looks great to me!

bjorncs (approved):

PR #32843, opened 2024-11-12
add the --file option to "vespa query"

arnej27959:
@mpolden please review
@kkraune FYI


mpolden (commented):
I think this needs a test, ideally one where all parameters come from a file and one with a combination of file+arguments.

mpolden (approved):
Great!

PR #32851, opened 2024-11-14
Tenant vault deployment data

gjoranv:

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
@@ -5,6 +5,7 @@ import com.yahoo.component.Version;
 import com.yahoo.config.FileReference;
 import com.yahoo.config.model.api.Quota;
 import com.yahoo.config.model.api.TenantSecretStore;
+import com.yahoo.config.model.api.TenantVault;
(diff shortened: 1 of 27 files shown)


tokle (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
```diff
@@ -54,6 +56,7 @@ void write(SessionZooKeeperClient zooKeeperClient, ApplicationId applicationId,
                                                              dockerImageRepository,
                                                              athenzDomain,
                                                              quota,
+                                                             tenantVaults,
```
hmusum, line 59: You need to call `zookeeperClient.writeTenantVaults()` before `if (writeSessionData.value())`
as well, this code is only called if a feature flag is set. This code should be cleaned up and explained more I see, I'll look into that

hmusum (changes requested):
Generally looks great, need one change

configserver/src/main/java/com/yahoo/vespa/config/server/session/SessionSerializer.java
gjoranv: Thanks! Missed that one.

gjoranv:
@hmusum PTAL

hmusum (approved):
👍

PR #32872, opened 2024-11-18
Log warning (shown in console) when using admin version 2.0 on cloud

hmusum:


config-model/src/main/java/com/yahoo/config/model/admin/AdminModel.java
```diff
@@ -78,7 +79,10 @@ public BuilderV2() {
 
         @Override
         public void doBuild(AdminModel model, Element adminElement, ConfigModelContext modelContext) {
-            if (modelContext.getDeployState().isHosted()) { // admin v4 is used on hosted: Build a default V4 instead
+            // admin v4 is used on hosted: Build a default V4 instead
+            if (modelContext.getDeployState().isHosted()) {
+                modelContext.getDeployLogger().logApplicationPackage(WARNING, "<admin> version 2.0 is not supported " +
```
kkraune, line 84: I think it is _supported_ as it works today:
```suggestion
                modelContext.getDeployLogger().logApplicationPackage(WARNING, "<admin> version 2.0 is deprecated" +
```

hmusum: Yes, that's right, thanks

oyving (approved):

PR #32914, opened 2024-11-21
add basic trace about search request and result

arnej27959:
@geirst please review
@vekterli was this what you needed?


vekterli:
Great! Infinitely better than the information void that was there previously 🥳

geirst (approved):

PR #32931, opened 2024-11-22
Interns/magnus/xmlintellij

Mangern:
Add support for services.xml file in IntelliJ plugin.
Also changed the documentation fetcher to make documentation hovers smaller - specifically the HTML tables from the docs did not display nicely.


theodorklauritzen (approved):
LGTM

PR #32933, opened 2024-11-22
for completeness, allow user to force WeakAnd using "~" prefix

arnej27959:
@bratseth please review
not sure how useful this really is, but it felt weird not having it.


bratseth (approved):
👍

PR #32934, opened 2024-11-22
Initial commit of Logstash plugins

radu-gheorghe:
Moved from the logstash-plugins repository.


bratseth (approved):
👍  we should point to the license file in the root by addint the standard one-liner at the top of the files, but we can do it after merging.

PR #32935, opened 2024-11-22
Remove special handling of position

bratseth:
Position is handled a bit specially so I added some tests, and I don't think we need the special handling in the config model.


arnej27959 (approved):
agreed - this looks like remnant of ancient backwards compatibility.   If system tests pass then it should be OK.

PR #32947, opened 2024-11-26
Ean/specific sso enforcement error

ean:

Is it possible to attach hints to an error? To avoid the final "Unauthorized" error.

### DIFF
--- a/client/go/internal/cli/cmd/login.go
+++ b/client/go/internal/cli/cmd/login.go
@@ -75,8 +75,6 @@ This command runs a browser-based authentication flow for the Vespa Cloud contro
 				return fmt.Errorf("login error: %w", err)
 			}
 
-			cli.printSuccess("Logged in")
-
 			// store the refresh token
 			secretsStore := &auth.Keyring{}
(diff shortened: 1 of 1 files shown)


client/go/internal/cli/cmd/login.go
```diff
@@ -72,7 +72,15 @@ This command runs a browser-based authentication flow for the Vespa Cloud contro
 			})
 
 			if err != nil {
-				return fmt.Errorf("login error: %w", err)
+				switch err.Error() {
+				case "600":
+					cli.printErr(fmt.Errorf("Unable to login through social channels"),
+						"Your organization does not allow logging in through our social channels",
+						"Please login by entering your address into the email field")
+					return fmt.Errorf("Unauthorized")
```
mpolden, line 80: You can simply return `errHint(fmt.Errorf("Unauthorized"), "Your organization does not allow logging in through our social channels", "Please login by entering your address into the email field")` here, and it will be printed automatically.

mpolden (approved):

PR #32961, opened 2024-11-28
Add support for explicit trimming of LRU cache

vekterli:
@havardpe please review. Last one... <sup>for now...</sup> 😼

`lrucache_map`: add a `trim()` function which evicts elements in old-to-new order until the LRU policy stays stop. Note that this does not use soft limits; if the cache has only a single element, and it is over-sized, trimming will remove the entry and leave the cache empty.

`cache`: always use the trimming functionality (for both segments) when reconfiguring capacities instead of special-casing eviction. Trim protected before probationary to ensure spillover to probationary segment.

### DIFF
--- a/vespalib/src/vespa/vespalib/stllike/cache.hpp
+++ b/vespalib/src/vespa/vespalib/stllike/cache.hpp
@@ -92,12 +92,6 @@ cache<P>::SizeConstrainedLru::try_get_and_ref(const KeyT& key, ValueT& val_out) 
 }
 
 template <typename P>
-void
-cache<P>::SizeConstrainedLru::trim() {
-    Lru::trim();
-}
-
-template <typename P>
 std::vector<typename P::Key>
 cache<P>::SizeConstrainedLru::dump_segment_keys_in_lru_order() {
     std::vector<KeyT> lru_keys;
--- a/vespalib/src/vespa/vespalib/stllike/cache.hpp
+++ b/vespalib/src/vespa/vespalib/stllike/cache.hpp
@@ -46,7 +46,6 @@ cache<P>::SizeConstrainedLru::insert_and_update_size(const KeyT& key, ValueT val
     add_size_bytes(_owner.calcSize(key, value));
     auto insert_res = Lru::insert(key, std::move(value));
     assert(insert_res.second);
-
 }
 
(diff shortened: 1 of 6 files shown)


vespalib/src/vespa/vespalib/stllike/cache.hpp
```diff
@@ -94,16 +93,8 @@ cache<P>::SizeConstrainedLru::try_get_and_ref(const KeyT& key, ValueT& val_out)
...
-    auto iter = Lru::begin();
-    while (iter != Lru::end()) {
-        // Entries will not be transitioned out of the cache via the probationary segment,
-        // so invoke the removal callback directly.
-        _owner.onRemove(iter.key());
-        iter = Lru::erase(iter); // This does _not_ invoke `removeOldest()`
-    }
-    set_size_bytes(0);
+cache<P>::SizeConstrainedLru::trim() {
+    Lru::trim();
```
havardpe, line 97: isn't this a subclass and would do this already?

vespalib/src/vespa/vespalib/stllike/lrucache_map.hpp
```diff
@@ -146,12 +146,16 @@ lrucache_map<P>::erase(const iterator & it)
...
-bool
+void
 lrucache_map<P>::verifyInternals()
 {
-    bool retval(true);
-    assert(_head != LinkedValueBase::npos);
-    assert(_tail != LinkedValueBase::npos);
+    if (_head == LinkedValueBase::npos || _tail == LinkedValueBase::npos) {
+        // If _either_ head or tail is invalid, they must both be.
+        assert(_head == LinkedValueBase::npos);
```
havardpe, line 154: consider using REQUIRE_EQ as a modern replacement for assert that also works with GTest and reports actual values

havardpe (dismissed):

vespalib/src/vespa/vespalib/stllike/cache.hpp
vekterli: Yes—the parent class is `protected`, but I'll bring `Lru::trim` explicitly into scope instead.

havardpe (approved):

PR #32993, opened 2024-12-04
Support tensor field presence checks as part of document selection evaluation

vekterli:
@geirst @toregge please review. Can be reviewed per commit, if desired.

For the C++ implementation, adds a sentinel `TensorValue` selection-internal type which is instantiated iff a field is a tensor, and which only allows for comparing against (explicit or implicit) `null` values. All other comparisons poison the expression with `Invalid` results. In other words, tensor _contents_ cannot be inspected as part of selection evaluation.

This sentinel is also instantiated by the search core tensor attribute wiring.

The Java implementation has very different evaluation semantics, so the changes are mostly about enforcing `Invalid` results when using tensors in ways that are not supported.

Also doing some casual code cleanup.


geirst (approved):
👍

toregge (approved):

PR #32995, opened 2024-12-05
Use a lower log level when there is no master cluster controller

hmusum:
When a request fails with master cluster controller the client will get that response, I don't think we need to log it as a warning, response and metrics shuold be enough.


vekterli (approved):
:+1:

PR #33004, opened 2024-12-05
feat: Add flag for legacy trust store option

esolitos:
## What

Defines a new Feature Flag which allows switching CA trust store.

## Why

Enables gradual roll out of the new CA.

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/Flags.java
@@ -496,6 +496,13 @@ public class Flags {
             "Takes after controller restart",
             SYSTEM);
 
+    public static final UnboundBooleanFlag USE_LEGACY_STORE = defineFeatureFlag(
+            "use-legacy-trust-store", true,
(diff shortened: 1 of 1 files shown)


freva (dismissed):

flags/src/main/java/com/yahoo/vespa/flags/Flags.java
```diff
@@ -496,6 +496,13 @@ public class Flags {
             "Takes after controller restart",
             SYSTEM);
 
+    public static final UnboundBooleanFlag USE_LEGACY_STORE = defineFeatureFlag(
+            "use-legacy-trust-store", true,
+            List.of("marlon"), "2024-12-05", "2025-02-01",
+            "Use legacy trust store for CA, or new one",
+            "Takes effect on restart of OCI containers",
+            SYSTEM);
```
hakonhall, line 504: No need for SYSTEM.
The list of dimension here should be the list of dimensions you are explicitly setting before resolving the flag value.

freva (approved):

PR #33008, opened 2024-12-06
Remove some ZooKeeper patches when using ZK 3.9.3

hmusum:
3.9.3 fixes https://issues.apache.org/jira/browse/ZOOKEEPER-4712 and https://issues.apache.org/jira/browse/ZOOKEEPER-4541 in particular, which means we should not need these patches anymore


hakonhall (approved):
Great!

PR #33010, opened 2024-12-09
Make vespa-zkflw work with fewer arguments

hmusum:
Use defaults for hostname, port and command
Require 0, 1 or 3 arguments to script

### DIFF
--- a/zookeeper-command-line-client/src/main/sh/vespa-zkflw
+++ b/zookeeper-command-line-client/src/main/sh/vespa-zkflw
@@ -77,6 +77,24 @@ export ROOT
 
 # END environment bootstrap section
 
+# Defaults
+server="localhost"
+port="2181"
+command="stat"
+
+if [ "$#" -eq 1 ]; then
+    command=$1
+elif [ "$#" -eq 2 ] || [ "$#" -gt 3 ]; then
+    echo "Need 0, 1 (command) or 3 arguments (host, port, command)"
+    exit 1
+elif [ "$#" -eq 3 ]; then
+    server=$1
+    port=$2
+    command=$3
+fi
+
+args="$server $port $command"
+
 java -cp ${VESPA_HOME}/lib/jars/zookeeper-command-line-client-jar-with-dependencies.jar \
      -Dlogback.configurationFile="logback-vespa.xml" -Xms32m -Xmx512m \
-     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain "$@"
+     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain $args
--- a/zookeeper-command-line-client/src/main/sh/vespa-zkflw
+++ b/zookeeper-command-line-client/src/main/sh/vespa-zkflw
@@ -93,8 +93,8 @@ elif [ "$#" -eq 3 ]; then
     command=$3
 fi
 
-args="$server $port $command"
+args=("$server" "$port" "$command")
 
 java -cp ${VESPA_HOME}/lib/jars/zookeeper-command-line-client-jar-with-dependencies.jar \
      -Dlogback.configurationFile="logback-vespa.xml" -Xms32m -Xmx512m \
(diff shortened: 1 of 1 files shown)


zookeeper-command-line-client/src/main/sh/vespa-zkflw
```diff
@@ -77,6 +77,24 @@ export ROOT
...
+    port=$2
+    command=$3
+fi
+
+args="$server $port $command"
+
 java -cp ${VESPA_HOME}/lib/jars/zookeeper-command-line-client-jar-with-dependencies.jar \
      -Dlogback.configurationFile="logback-vespa.xml" -Xms32m -Xmx512m \
-     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain "$@"
+     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain $args
```
hakonhall, line 100: If command ever contains something that means something special in shell, e.g. ;, \, quotes, etc, then this will not work.  The correct way of preserving arguments would be to make args an array, and then do e.g. args=("$1" "$2" "$3").

hakonhall (dismissed):

zookeeper-command-line-client/src/main/sh/vespa-zkflw
hmusum: Thanks, that's better, fixd

zookeeper-command-line-client/src/main/sh/vespa-zkflw
```diff
@@ -77,6 +77,24 @@ export ROOT
 
 # END environment bootstrap section
 
+# Defaults
+server="localhost"
+port="2181"
+command="stat"
+
+if [ "$#" -eq 1 ]; then
+    command=$1
```
hakonhall, line 86: args=("$1")

zookeeper-command-line-client/src/main/sh/vespa-zkflw
```diff
@@ -77,6 +77,24 @@ export ROOT
...
+elif [ "$#" -eq 2 ] || [ "$#" -gt 3 ]; then
+    echo "Need 0, 1 (command) or 3 arguments (host, port, command)"
+    exit 1
+elif [ "$#" -eq 3 ]; then
+    server=$1
```
hakonhall, line 91: args=("$1" "$2" "$3")

zookeeper-command-line-client/src/main/sh/vespa-zkflw
```diff
@@ -77,6 +77,24 @@ export ROOT
...
+    port=$2
+    command=$3
+fi
+
+args=("$server" "$port" "$command")
```
hakonhall, line 96: Remove this line, as `args` is set above.

zookeeper-command-line-client/src/main/sh/vespa-zkflw
```diff
@@ -77,6 +77,24 @@ export ROOT
...
+
 java -cp ${VESPA_HOME}/lib/jars/zookeeper-command-line-client-jar-with-dependencies.jar \
      -Dlogback.configurationFile="logback-vespa.xml" -Xms32m -Xmx512m \
-     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain "$@"
+     com.yahoo.vespa.zookeeper.cli.FourLetterWordMain ${args[@]}
```
hakonhall, line 100: `${args[@]}` needs to be double-quoted.

zookeeper-command-line-client/src/main/sh/vespa-zkflw
hmusum: We always need to use all 3 arguments, so cannot do it this way when there is just 0 or 1 arguments

hakonhall: :+1: Got it

hakonhall (approved):

PR #33012, opened 2024-12-09
Wire SLRU segment size ratio into Proton config

vekterli:
@geirst please review

By default SLRU semantics are disabled, but a segment ratio greater than 0 will enable it implicitly.

Also add a guard against negative values in config, since we don't support auto-adjusting based on available memory yet.

### DIFF
--- a/searchlib/src/vespa/searchlib/diskindex/posting_list_cache.h
+++ b/searchlib/src/vespa/searchlib/diskindex/posting_list_cache.h
@@ -20,6 +20,44 @@ private:
     std::unique_ptr<Cache> _cache;
     std::unique_ptr<BitVectorCache> _bitvector_cache;
 public:
+    class CacheSizingParams {
+        size_t _posting_max_bytes   = 0;
+        size_t _bitvector_max_bytes = 0;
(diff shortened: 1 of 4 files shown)


searchlib/src/vespa/searchlib/diskindex/posting_list_cache.h
```diff
@@ -20,6 +20,44 @@ class PostingListCache : public IPostingListCache {
...
+        }
+
+        [[nodiscard]] size_t bitvector_slru_protected_bytes() const noexcept {
+            if (_bitvector_slru_protected_ratio <= 0) {
+                return 0;
+            }
+            return static_cast<size_t>(static_cast<double>(_bitvector_max_bytes) * _bitvector_slru_protected_ratio);
+        }
+        [[nodiscard]] size_t bitvector_slru_probationary_bytes() const noexcept {
+            return _bitvector_max_bytes - posting_slru_protected_bytes();
```
geirst, line 56: Should this use `bitvector_slru_protected_bytes()` instead?

vekterli: Definitely. Fixed.

geirst (approved):

PR #33018, opened 2024-12-10
Update Netty to 4.1.115.Final

hmusum:


vespa-dependencies-enforcer/allowed-maven-dependencies.txt
```diff
@@ -78,8 +78,8 @@ io.jsonwebtoken:jjwt-api:${java-jjwt.vespa.version}
 io.jsonwebtoken:jjwt-impl:${java-jjwt.vespa.version}
 io.jsonwebtoken:jjwt-jackson:${java-jjwt.vespa.version}
 io.netty:netty-buffer:${netty.vespa.version}
-io.netty:netty-codec-http2:${netty.vespa.version}
```
hakonhall, line 81: So we will depend on a version 115 that is not allowed ? :thinking:

hakonhall (approved):

vespa-dependencies-enforcer/allowed-maven-dependencies.txt
hmusum: We don't depend on io.netty:netty-codec-http{,2} directly, maybe we should to be able to use 4.1.115.Final, but I don't know if that will break something else.

PR #33022, opened 2024-12-10
Integrate optional LFU frequency sketch into (S)LRU cache

vekterli:
@havardpe please review today's advent of cache 🎅 oh ho ho ho! Most added lines are tests or comments.

Allows a cache to be configured to use a probabilistic LFU frequency sketch for gating insertions to its segments. An element that is a candidate for insertion will only be allowed into a segment if it is estimated to be more frequently accessed than the element it would displace (note that if the cache is below max capacity the insertion would not displace any element(s), so it is always allowed).

LFU gating works in both LRU and SLRU modes. In both modes, initial insertion into the probationary segment is gated (for cache read-through and write-through). In SLRU mode, promotion from probationary to protected is _also_ gated. In the case that promotion is denied, the candidate element is placed at the LRU head of the probationary segment instead, giving it another chance.

The configured sketch element count should be at least as large as the maximum _expected_ number of elements that the cache can hold at once.

The default size is 0, i.e. LFU functionality is disabled.

Setting the size to >0 will always create a new sketch. The sketch will be initialized with the cache keys that are currently present in the cache segments, giving each existing entry an estimated frequency of 1. All preexisting frequency information about entries _not_ currently in the cache will be lost.

### DIFF
--- a/vespalib/src/vespa/vespalib/stllike/cache.h
+++ b/vespalib/src/vespa/vespalib/stllike/cache.h
@@ -276,7 +276,9 @@ public:
      * Setting the size to >0 will always create a new sketch. The sketch will be
      * initialized with the cache keys that are currently present in the cache segments,
      * giving each existing entry an estimated frequency of 1. All preexisting frequency
-     * information about entries _not_ currently in the cache will be lost.
+     * information about entries _not_ currently in the cache will be lost. To avoid
+     * pathological frequency estimates for existing entries, the sketch has a lower
+     * bound size of max(existing cache element count, cache_max_elem_count).
      */
     void set_frequency_sketch_size(size_t cache_max_elem_count);
 
--- a/vespalib/src/vespa/vespalib/stllike/cache.h
+++ b/vespalib/src/vespa/vespalib/stllike/cache.h
@@ -3,6 +3,7 @@
 
 #include "lrucache_map.h"
 #include <vespa/vespalib/util/memoryusage.h>
+#include <vespa/vespalib/util/relative_frequency_sketch.h>
 #include <atomic>
 #include <mutex>
 #include <vector>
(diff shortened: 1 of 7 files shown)


vespalib/src/vespa/vespalib/stllike/cache.hpp
```diff
@@ -230,6 +253,64 @@ cache<P>::cache(BackingStore& backing_store, size_t max_bytes)
...
+                                uint8_t candidate_freq) const noexcept
+{
+    if (!_sketch) {
+        return true; // Trivially accepts insertion, since there's no LFU policy
+    }
+    // TODO > capacity_bytes() instead of >=, this uses >= to be symmetric with removeOldest()
+    const bool would_displace = ((segment.size() >= segment.capacity()) ||
+                                 (segment.size_bytes() + calcSize(key, value)) >= segment.capacity_bytes());
+    const K* victim;
+    if (would_displace && (victim = segment.last_key_or_nullptr()) != nullptr) {
```
havardpe, line 299: consider re-writing to not use assignment inside check

vespalib/src/vespa/vespalib/stllike/cache.h
```diff
@@ -127,6 +139,10 @@ class cache {
         //   size_bytes() <= capacity() && size() <= maxElements()
         using Lru::trim;
 
+        // Invokes functor `fn` for each segment key in LRU order (new to old)
```
havardpe, line 142: technically the wrong order. Maybe we should consider enforcing a big enough sketch to not decay due to filling in all cache elements whenever a size > 0 is selected?

havardpe (dismissed):

vespalib/src/vespa/vespalib/stllike/cache.hpp
vekterli: Refactored to instead early-out on `!would_displace`

vespalib/src/vespa/vespalib/stllike/cache.h
vekterli: Added a lower bound on the existing cache `size()` when setting a non-zero sketch size. The iteration order is an artefact of the only available iteration order of the underlying `lrucache_map`, so not doing anything with it for now at least.

havardpe (approved):

PR #33038, opened 2024-12-13
Don't fail early when using reference attribute in grouping expression.

geirst:
This is not yet supported in the backend, so it still results in an error in the query result. This change is done to unblock some failing tests.

@hmusum please review
@toregge FYI


hmusum (approved):

geirst:
@ean FYI

PR #33041, opened 2024-12-16
Add doc link to warning when there are fieldset inconsistencies

hmusum:
@radu-gheorghe please review.

Adding just link to doc for now. I see some unexpected behavior when trying to show warning for all fields in a schema, so will follow up that in another PR.


radu-gheorghe (approved):
Looks good to me. Thanks for doing this!

PR #33044, opened 2024-12-17
Special handling for "Latin Capital Letter I with Dot Above":

arnej27959:
Java follow Unicode standard which says lowercasing should produce "i" plus "Combining Dot Above".  Our C++ code doesn't agree that this makes sense, and note that this is the only character which produces two characters as output from lowercasing.

Ideally we should do "Case Folding" instead of lowercasing.

@bratseth please review
@hmusum FYI


bratseth (dismissed):

arnej27959:
@geirst or. @toregge take a look at the C++ unit test changes.  It turns out it also tested Java-only lowercasing comparing to reference values, and I found that it only compared the first character for fastlib/vespalib variants.

geirst (approved):

PR #33048, opened 2024-12-17
Try to warn about all fields having inconsistent settings when they a…

hmusum:
…re used in a field set

Note that this is not done for match settings, as that will change behavior, but we should probably do that change, as it might be wrong now (e.g. you can get 'text' matching when the 2 fields in a fieldset have matching 'exact' and 'word')

@bratseth or @arnej27959 might have opinions about this


arnej27959 (commented):
This looks problematic to me.  The new text claims that "matching.getType()" will be used, but the "return" that you want to remove will prevent that from happening.  So we must either do the change or change how we produce the new text.  Also, it doesn't warn about all fields like the commit message would indicate.

hmusum:
Thanks, looks good, merging

arnej27959 (approved):
👍

PR #33059, opened 2024-12-27
[logstash-plugins] output: added options to remove metadata fields

radu-gheorghe:
…from documents after using them for put/update.

The default assumption is that if we want the ID from the document, we assume we store the whole document (including the ID). But if we don't want the ID in the document, we can't remove it like we do with e.g. @timestamp (using the `mutate` filter, as in the README example), because we wouldn't have it at index time. The option I see now is to delete the ID from the document as a new option of the output plugin.

Similarly, we have new options for namespace, document type and operation, in case we want them to be dynamic (i.e. from document fields).

With this PR there's also a critical bugfix: explicit `id_field` didn't work. It would work with the default `id` from the document and it did generate UUIDs, but we have explicit `id_field` in the README even :)


kkraune (approved):
👍

PR #33061, opened 2024-12-30
[logstash-plugins] trim gem size for output

radu-gheorghe:
It got to 46MB and now it's back to 16MB. Packed a couple of minor improvements in this PR as well.

I feel like the plugins are quite feature-complete at this point, though I have some nice big improvements that I'm planning to add.


kkraune (approved):
👍

PR #33078, opened 2025-01-03
Revert "give a better estimate in WhiteListBlueprint"

toregge:
Reverts vespa-engine/vespa#33076

3 system tests fails with the new estimate.


toregge:
Revert not needed.

PR #33089, opened 2025-01-07
Use normal abseil-cpp for CentOS Stream 10 and Fedora.

toregge:
@arnej27959 : please review

This is related to vespa-engine/vespa-3rdparty-deps#294


arnej27959 (approved):
👍

PR #33099, opened 2025-01-10
[logstash-plugins] adds dead letter queue support to Vespa output

radu-gheorghe:
Now if we have a permanent failure (e.g. field not in the schema, exceeded number of retries) we can put them in the dead letter queue and process them later. As opposed to dropping them :)

More details on how DLQ works can be found [here](https://www.elastic.co/guide/en/logstash/current/dead-letter-queues.html).

The only downside is that I couldn't find any other Java plugin that supports DLQ, and I think there's some missing functionality in Logstash itself, because I couldn't convince my `context` to return anything other than a `NullDeadLetterQueueWriter`. The bottom line is that I couldn't reuse DLQ settings from `logstash.yml`, so **we have specific settings in the plugin to configure DLQ**. Which I guess would have been needed anyway (as an override) in case you have multiple outputs defined and you only want DLQ for some.


kkraune (approved):
👍

PR #33107, opened 2025-01-14
Set memory resource limit lower for content nodes with less than 8 Gib memory

hmusum:
Use feature flag (default value same as for nodes with more memory) to set memory limit for content nodes in vespa cloud

### DIFF
--- a/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
+++ b/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
@@ -1708,7 +1708,7 @@ public class ContentClusterTest extends ContentBaseTest {
         assertEquals(0.864, protonConfigBuilder.build().writefilter().disklimit(), 0.001);
         assertEquals(0.84, protonConfigBuilder.build().writefilter().memorylimit(), 0.001);
 
-        // Resource limits should for content nodes should be changed based on new values above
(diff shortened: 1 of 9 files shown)


config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
```diff
@@ -1666,6 +1667,57 @@ void strictly_increasing_cluster_state_versions_config_controlled_by_feature_fla
...
 
+    @Test
+    void testResourceLimitsForSmallNodes() throws Exception {
+        var services = """
+                <services>
+                  <content version="1.0" id="foo">
+                    <redundancy>1</redundancy>
+                    <documents/>
+                    <nodes count='1'>
+                      <resources vcpu='1' memory='3Gb' />
```
geirst, line 1678: Consider testing with a 8GB flavor, as that is considered small.

geirst (dismissed):

config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
hmusum: 👍 Done

hmusum:
@geirst please take a look again

geirst (approved):

PR #33113, opened 2025-01-14
Stop using vespa_string in cache test.

toregge:
@vekterli : please review


vekterli (approved):
:+1: we should consider refactoring how this test deals with cache element sizes and limits at some point to avoid the current magic numbers

PR #33114, opened 2025-01-14
Log a warning if setting content node resource limits directly

hmusum:
Discouraged, should set for cluster controller instead. Note that it's not allowed to set these limits in vespa cloud


hmusum:
@geirst please review

geirst (approved):

Issue #32645, opened 2024-10-24
Full visit without any selection criteria is dropping handful of documents with discrepancy in documentCount


nehajatav:
**Describe the bug**
Every time we run a full visit of the cluster we are losing out on few documents exactly same in number as the #documents misreported in documentCount of response. The missed documents were not consistently the same, but they were not created nor reindexed during the visit (created in 2021)

**To Reproduce**
Run full visit using slices
Very few visit calls results in #documents in json lesser than documentCount json key provided in visit response

**Expected behavior**
We expected the missed document to be there in a full visit run esp since the document was an old one and not recently created

**Environment (please complete the following information):**
 - OS: RHEL8
 - Infrastructure: Podman

**Vespa version**
8.332.5

vekterli:
From my understanding this is observed when using `stream=true`, correct? Does this also happen if `stream=false`, i.e. buffered response mode?

vekterli:
After locally running hundreds upon hundreds of full cluster (streaming) Document V1 visits across many loops (with a 50K document set), I have observed that on very rare occasions a small number of documents (usually 1-2) will be missing from the output.

This anomaly appears to be highly timing sensitive and transient. I have thus far only observed it when visiting using `stream=true` (with the usual caveats that absence of proof does not imply proof of absence etc.). Assuming this observation holds, it may hint towards a potential latent race condition in the asynchronous response rendering pipeline, though nothing is certain at this point.

I will prioritize looking into this. First step is to find a way to reduce the number of testable dimensions down to something that can (at least semi-) reliably reproduce the issue.

nehajatav:
Thanks much @vekterli ! We are pushing this version to all our envs now. Will reopen if we find something like this again

PR #33126, opened 2025-01-16
Ensure document is visible in render queue prior to visitor ACK

vekterli:
@bjorncs please review.

Document V1 API visitors asynchronously invoke a callback every time a document is received from a content node. This callback previously ACKed the operation, rendered the document JSON to a byte buffer and enqueued the buffer to an internal render queue (in a lock-free manner).

This ordering has a subtle flaw. If an ACK is sent for the final document of a visitor, it is possible (but unlikely) that the content node can observe the completion of the whole visitor and unwind this entirely back to the client (via the distributor) _before_ the document JSON rendering and buffer enqueueing has had time to complete. This will invoke another completion callback which flushes all remaining document buffers to the wire and marks the response as done. Any buffers not on the queue at that point will never be rendered, i.e. will appear to the client as if lost. This is what happens in the above ACK race.

This commit simply moves response ACKing to explicitly happen-after JSON rendering and buffer enqueueing to avoid premature session completion.

This is expected to resolve what was observed in #32645.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
@@ -818,17 +818,23 @@ public final class DocumentV1ApiHandler extends AbstractRequestHandler {
 
         /** Writes documents to an internal queue, which is flushed regularly. */
         void writeDocument(DocumentWriter documentWriter, CompletionHandler completionHandler) throws IOException {
-            if (completionHandler != null) {
-                acks.add(completionHandler);
-                ackDocuments();
-            }
-
             // Serialise document and add to queue, not necessarily in the order dictated by "written" above,
             // i.e., the first 128 documents in the queue are not necessarily the ones ack'ed early.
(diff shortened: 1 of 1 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -818,18 +818,24 @@ void writeDocumentRemoval(DocumentId id, CompletionHandler completionHandler) th
...
             myOut.write(','); // Prepend rather than append, to avoid double memory copying.
             documentWriter.write(myOut);
             docs.add(myOut);
 
+            // It is crucial that ACKing of in-flight operations happens _after_ the document payload is
+            // visible in the `docs` queue. Otherwise, there is a risk that the ACK sets in motion a
+            // full unwind of the entire visitor session from the content node back to the client session
+            // (that's us), triggering the `onDone` callback and transitively the final flush of enqueued
+            // documents. If `myOut` is then not part of `docs`, it won't be rendered at all.
+            // This is a Dark Souls-tier distributed race condition.
```
bjorncs, line 833: 😆

bjorncs (approved):

bratseth:
edb ...

PR #33131, opened 2025-01-17
Support json query with "where: <boolean>"

hmusum:
Also reorder switch statements.
No functional changes.


bratseth (approved):
Surely the LAST place we needed to add boolean 😌

radu-gheorghe:
That was fast! Thanks a lot @hmusum 🙏

PR #33137, opened 2025-01-20
Add supported authentication methods to ZoneEndpoint configuration

evgiz:
Add configuration to indicate which authentication methods are supported for zone endpoints - to be used in load balancer provisioning to set up appropriate resources.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
@@ -33,6 +33,7 @@ import com.yahoo.config.provision.InstanceName;
 import com.yahoo.config.provision.NodeType;
 import com.yahoo.config.provision.Zone;
 import com.yahoo.config.provision.ZoneEndpoint;
+import com.yahoo.config.provision.zone.AuthMethod;
 import com.yahoo.config.provision.zone.ZoneId;
 import com.yahoo.container.bundle.BundleInstantiationSpecification;
 import com.yahoo.container.jdisc.DataplaneProxyService;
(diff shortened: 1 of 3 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/ZoneEndpoint.java
```diff
@@ -27,16 +27,25 @@ public class ZoneEndpoint {
...
      */
     public static final int generation = 0;
-    public static final ZoneEndpoint defaultEndpoint = new ZoneEndpoint(true, false, List.of());
-    public static final ZoneEndpoint privateEndpoint = new ZoneEndpoint(false, false, List.of());
+    public static final ZoneEndpoint defaultEndpoint = new ZoneEndpoint(true, false, false, List.of());
+    public static final ZoneEndpoint privateEndpoint = new ZoneEndpoint(false, false, false, List.of());
 
     private final boolean isPublicEndpoint;
     private final boolean isPrivateEndpoint;
+    private final boolean supportsTokenAuthentication;
```
mpolden, line 35: I think this should be `List<AuthMethod> authMethods` instead, to make it more extensible. The `defaultEndpoint` and `privateEndpoint` should have `List.of(AuthMethod.mtls)` as the default.

The getter/builders methods should then be named `authMethods/withAuthMethods`.

mpolden (approved):

config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -1034,7 +1035,20 @@ private ZoneEndpoint zoneEndpoint(ConfigModelContext context, ClusterSpec.Id clu
         InstanceName instance = context.properties().applicationId().instance();
         ZoneId zone = ZoneId.from(context.properties().zone().environment(),
                                   context.properties().zone().region());
-        return context.getApplicationPackage().getDeploymentSpec().zoneEndpoint(instance, zone, cluster);
+
+        var supportsTokenAuthentication = context.properties()
+                .endpoints()
+                .stream()
+                .anyMatch(endpoint -> endpoint.authMethod() == ApplicationClusterEndpoint.AuthMethod.token);
```
mpolden, line 1042: You also need to match the cluster here, since this is the zone endpoint for a particular cluster and auth method is a per-cluster setting.

PR #33141, opened 2025-01-21
Always use 16 distribution bits in dev in vespa cloud

hmusum:


config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
```diff
@@ -591,21 +591,12 @@ public int distributionBits() {
...
-            return distributionBits;
-        }
+        return distributionBits;
     }
 
     private boolean zoneEnvImplies16DistributionBits() {
-        // We want perf to behave like prod as much as possible.
-        return (zone.environment() == Environment.prod) || (zone.environment() == Environment.perf);
+        // We want dev and perf to behave like prod as much as possible.
+        return Set.of(Environment.dev, Environment.perf, Environment.prod).contains(zone.environment());
```
vekterli, line 599: Looks like there's also an `isAnyOf(...)` method in `Environment`, but can't see it in use anywhere else.

vekterli (dismissed):
:+1:

config-model/src/main/java/com/yahoo/vespa/model/content/cluster/ContentCluster.java
hmusum: Nice! We better use it somewhere then. Fixed

vekterli (approved):

PR #33150, opened 2025-01-21
Revert "Add serialization for zone endpoint auth methods in load balancer (zookeeper)"

hmusum:
Reverts vespa-engine/vespa#33144

Seeing some strange things for first publiccd controller that upgraded:
`[2025-01-21 15:11:56.217] WARNING configserver     Container.com.yahoo.vespa.hosted.provision.provisioning.Preparer     Failed to prepare load balancers for hosted-vespa.controller container cluster 'controller' 8.468.12: Could not provision load balancer hosted-vespa:controller:default:controller. The operation will be retried on next deployment.: Could not provision target group for application hosted-vespa.controller.default (Ignoring because bootstrap deployment)`

and an exception:

`[2025-01-21 15:11:56.220] INFO    configserver     Container.com.yahoo.vespa.hosted.provision.provisioning.InfraDeployerImpl
        Failed to activate hosted-vespa.controller
        exception=
        com.yahoo.config.provision.exception.LoadBalancerServiceException: Could not provision load balancer hosted-vespa:controller:default:controller. The operation will be retried on next deployment.
        at com.yahoo.vespa.hosted.provision.provisioning.LoadBalancerProvisioner.provisionInstance(LoadBalancerProvisioner.java:265)
        at com.yahoo.vespa.hosted.provision.provisioning.LoadBalancerProvisioner.prepare(LoadBalancerProvisioner.java:217)
        at com.yahoo.vespa.hosted.provision.provisioning.LoadBalancerProvisioner.prepare(LoadBalancerProvisioner.java:106)
        at com.yahoo.vespa.hosted.provision.provisioning.Preparer.lambda$prepare$2(Preparer.java:74)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at com.yahoo.vespa.hosted.provision.provisioning.Preparer.prepare(Preparer.java:74)
        at com.yahoo.vespa.hosted.provision.provisioning.NodeRepositoryProvisioner.prepare(NodeRepositoryProvisioner.java:110)
        at com.yahoo.vespa.hosted.provision.provisioning.InfraDeployerImpl$InfraDeployment.prepare(InfraDeployerImpl.java:93)
        at com.yahoo.vespa.hosted.provision.provisioning.InfraDeployerImpl$InfraDeployment.activate(InfraDeployerImpl.java:103)
[...]

PR #33153, opened 2025-01-22
Reapply "Add serialization for zone endpoint auth methods in load balancer (zookeeper)"

evgiz:
Reapply revert from vespa-engine/vespa#33150

Provisioning issue was caused by an outdated zone tag. Auth method change indicated a config change which triggered a validation check. When tags are updated this change should proceed as intended. @freva 

As for the empty `authMethods` list in ZK, this should be updated in `LoadBalancerProvisioner::223` and be correctly provided to activate in `LoadBalancerProvisioner::134`, so even if ZK was updated during deploy this shouldn't be an issue. Could you verify @mpolden?


freva (approved):
I've updated the tags.

mpolden (approved):

hmusum (approved):

PR #33164, opened 2025-01-23
Revert "Set timeout for RPC call proton.prepareRestart to infinity"

hmusum:
Reverts vespa-engine/vespa#32376

This does not work in practice as sentinel blocks until this has finished before stopping the process, need a lower timeout

@arnej27959 please review
@toregge FYI


arnej27959 (approved):
👍

PR #33172, opened 2025-01-24
Reduce max relative merge memory usage from 3% to 1.5%

vekterli:
@geirst please review. Reducing the upper bound will be done separately once we have observed that this change does not incur an adverse merge performance impact in practice.
@toregge FYI

Merge memory limits are by default set as a percentage of the process' memory limit, capped by lower and upper bounds.

On nodes with the vast majority of memory taken up by attributes, using 3% of memory on merging might be a much larger relative percentage jump in _heap_ memory, which can cause feed block limits to be crossed if the node's already operating very close to the limit.


vekterli:
Will fix config model test that cross-checks this limit

geirst (approved):

PR #33174, opened 2025-01-24
Add option for ignoring non-existent documents

mpolden:
The user might want to read a set of documents where one or more of them may not
exist, without CLI returning errors. This was requested by @wix-mikej.

@bratseth or @kkraune

### DIFF
--- a/client/go/internal/cli/cmd/document.go
+++ b/client/go/internal/cli/cmd/document.go
@@ -101,10 +101,10 @@ func sendOperation(op document.Operation, args []string, timeoutSecs int, waiter
 		service.CurlWriter.InputFile = filename
 	}
 	result := client.Send(doc)
-	return printResult(cli, operationResult(false, doc, service, result), false)
+	return printResult(cli, operationResult(false, doc, service, result), false, false)
 }
 
-func readDocuments(ids []string, timeoutSecs int, waiter *Waiter, printCurl bool, cli *CLI, fieldSet string, headers []string) error {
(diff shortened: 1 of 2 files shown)


wix-mikej:
There is an important distinction between ignoring missing document errors, and ignoring all errors which this diff appears to do. We wanted the "document missing" error still reported via stderr (so it didn't break `vespa feed`), but didn't want it to exit early (or with an error). If it ran into any other errors it exits w/ failure and we logged the error for review or retry. 

In the end we used the following change against v8.460.16:
```diff
diff --git a/client/go/internal/cli/cmd/document.go b/client/go/internal/cli/cmd/document.go
index 267f0e91c29..4e084c20c77 100644
--- a/client/go/internal/cli/cmd/document.go
+++ b/client/go/internal/cli/cmd/document.go
@@ -122,7 +122,9 @@ func readDocuments(ids []string, timeoutSecs int, waiter *Waiter, printCurl bool
        for _, docId := range parsedIds {
                result := client.Get(docId, fieldSet)
                if err := printResult(cli, operationResult(true, document.Document{Id: docId}, service, result), true); err != nil {
-                       return err
+                       if result.HTTPStatus != 404 {
+                               return err
+                       }
                }
        }
```

Also, I apologize for not providing a PR for our change as I had indicated I would. We've been really busy launching a couple things, and I hadn't had a chance to put something together. What you have is basically what I was considering, save the detail above.

mpolden:
Thanks! I considered ignoring 404 only, but thought that might be too strict for your use case. I'll make the change.

arnej27959 (dismissed):
looks good to me

mpolden:
Needs another approval after I fixed a conflict.

arnej27959 (approved):
👍

PR #33179, opened 2025-01-25
Bratseth/type inference cleanup

bratseth:
@arnej I did a little more while the build was broken.

Could you take a look at the last (real) commit, "@bratseth
[Don't return current type from verifiction](https://github.com/vespa-engine/vespa/pull/33179/commits/1d4fabcb0c63501b93e6f89e0c4e12a6f204c981)"

Should be harmless as this was only used by tests but necessitated lots of test changes.


arnej27959 (dismissed):
I'm a bit worried about the unit tests that you have removed, it looks like some of the things they test will stop working now, and it's hard to be certain that it won't be a problem in the larger setting.

bratseth:
I need to move the testing to a level where entire statements are tested since the type resolution is at the statement level. I did that with for_each here but maybe I missed something. What tests did you have in mind?

bratseth:
Ok, thanks! Yes the simplification isn't complete with this.

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/Expression.java
```diff
@@ -158,24 +158,24 @@ public final void verify(DocumentType type) {
...
 
-    public final DocumentUpdate verify(DocumentUpdate upd) {
-        return verify(new SimpleAdapterFactory(), upd);
+    public final void verify(DocumentUpdate upd) {
+        verify(new SimpleAdapterFactory(), upd);
     }
 
-    public final DocumentUpdate verify(AdapterFactory factory, DocumentUpdate upd) {
+    public final void verify(AdapterFactory factory, DocumentUpdate upd) {
         DocumentUpdate ret = null;
```
arnej27959, line 179: it looks like "ret" is no longer used and the logic computing it can be removed

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/Expression.java
```diff
@@ -181,21 +187,19 @@ public final DocumentUpdate verify(AdapterFactory factory, DocumentUpdate upd) {
                 ret = output;
             }
         }
-        return ret;
     }
 
     public final DocumentUpdate verify(UpdateAdapter adapter) {
```
arnej27959, line 192: this should probably be void as well

arnej27959 (dismissed):
looks good to me, only minor suggestions

bratseth:
Both done - thanks!

arnej27959 (approved):
👍

hmusum:
One of our integration tests started failing with:

`Invalid application: For schema 'foo', field 'bar': Invalid expression 'input baz': This produces type Array<string> but Array<any> is required expected:<0> but was:<2>`

which seems related to the changes here

bratseth:
Thanks! I'll just revert for now then.

PR #33191, opened 2025-01-27
Propagate WeakAnd-related query property overrides

vekterli:
@geirst please review
@toregge FYI

Adds a new `weakand` property hierarchy under `ranking.matching` and exposes `stopwordLimit` and `adjustTarget`, which shall map 1-1 to the existing rank profile settings bearing the same names.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
+++ b/container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
@@ -94,12 +94,12 @@ public class QueryProperties extends Properties {
         addDualCasedRM(map, Matching.POST_FILTER_THRESHOLD, GetterSetter.of(query -> query.getRanking().getMatching().getPostFilterThreshold(), (query, value) -> query.getRanking().getMatching().setPostFilterThreshold(asDouble(value, 1.0))));
         addDualCasedRM(map, Matching.APPROXIMATE_THRESHOLD, GetterSetter.of(query -> query.getRanking().getMatching().getApproximateThreshold(), (query, value) -> query.getRanking().getMatching().setApproximateThreshold(asDouble(value, 0.05))));
         addDualCasedRM(map, Matching.TARGET_HITS_MAX_ADJUSTMENT_FACTOR, GetterSetter.of(query -> query.getRanking().getMatching().getTargetHitsMaxAdjustmentFactor(), (query, value) -> query.getRanking().getMatching().setTargetHitsMaxAdjustmentFactor(asDouble(value, 20.0))));
-        map.put(CompoundName.fromComponents(Ranking.RANKING, Ranking.MATCHING, Matching.WEAKAND, WeakAnd.STOP_WORD_LIMIT),
-                GetterSetter.of(query -> query.getRanking().getMatching().getWeakAnd().getStopWordLimit(),
-                               (query, value) -> query.getRanking().getMatching().getWeakAnd().setStopWordLimit(asDouble(value, 1.0))));
+        map.put(CompoundName.fromComponents(Ranking.RANKING, Ranking.MATCHING, Matching.WEAKAND, WeakAnd.STOPWORD_LIMIT),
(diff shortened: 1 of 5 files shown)


container-search/src/main/java/com/yahoo/search/query/ranking/WeakAnd.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Query properties exposing WeakAnd-specific settings.
+ *
+ * @author vekterli
+ */
+public class WeakAnd implements Cloneable {
+
+    private static final QueryProfileType argumentType;
+
+    public static final String STOP_WORD_LIMIT = "stopwordLimit";
```
geirst, line 18: I think this should be _STOPWORD_LIMIT_ to match the name in the rank profile.

container-search/src/main/java/com/yahoo/search/query/ranking/WeakAnd.java
```diff
@@ -0,0 +1,83 @@
...
+        argumentType = new QueryProfileType(Matching.WEAKAND);
+        argumentType.setStrict(true);
+        argumentType.setBuiltin(true);
+        argumentType.addField(new FieldDescription(STOP_WORD_LIMIT, "double"));
+        argumentType.addField(new FieldDescription(ADJUST_TARGET, "double"));
+    }
+
+    public static QueryProfileType getArgumentType() { return argumentType; }
+
+    private Double stopWordLimit = null;
```
geirst, line 31: I think this should be _stopwordLimit_, _getStopwordLimit()_, and _setStopwordLimit()_ to match the name in the rank profile.

container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
```diff
@@ -93,6 +94,12 @@ private static Map<CompoundName, GetterSetter> createPropertySetterMap() {
...
         addDualCasedRM(map, Matching.TARGET_HITS_MAX_ADJUSTMENT_FACTOR, GetterSetter.of(query -> query.getRanking().getMatching().getTargetHitsMaxAdjustmentFactor(), (query, value) -> query.getRanking().getMatching().setTargetHitsMaxAdjustmentFactor(asDouble(value, 20.0))));
+        map.put(CompoundName.fromComponents(Ranking.RANKING, Ranking.MATCHING, Matching.WEAKAND, WeakAnd.STOP_WORD_LIMIT),
+                GetterSetter.of(query -> query.getRanking().getMatching().getWeakAnd().getStopWordLimit(),
+                               (query, value) -> query.getRanking().getMatching().getWeakAnd().setStopWordLimit(asDouble(value, 1.0))));
+        map.put(CompoundName.fromComponents(Ranking.RANKING, Ranking.MATCHING, Matching.WEAKAND, WeakAnd.ADJUST_TARGET),
+                GetterSetter.of(query -> query.getRanking().getMatching().getWeakAnd().getAdjustTarget(),
+                               (query, value) -> query.getRanking().getMatching().getWeakAnd().setAdjustTarget(asDouble(value, 0.0))));
```
geirst, line 102: The default should be 1.0, meaning the feature is turned off.

container-search/src/main/java/com/yahoo/search/query/ranking/WeakAnd.java
vekterli: Updated. `getStopwordLimit` et al just somehow _feel_ a bit more awkward than `getStopWordLimit` to me 🙈

container-search/src/main/java/com/yahoo/search/query/properties/QueryProperties.java
vekterli: Updated. Not sure what the semantics are of the default value here, since we already explicitly test that _not_ providing the query property returns a `null` `Double` value, i.e. implicitly turned off.

geirst (dismissed):

geirst (approved):

PR #33201, opened 2025-01-28
Glebashnik/fix openai lm component

glebashnik:

This should fix the bug with empty OpenAPI key when retrieved from secret store.
The problem was my misunderstanding of setApiKey, which actually modifies the parameters.
Did some refactoring to make it more obvious and improved tests to catch this bug.


bjorncs (approved):

bjorncs:
PR build failure is unrelated

PR #33204, opened 2025-01-28
Fail schema parsing when values are outside allowed range

vekterli:
@geirst please review
@toregge FYI

Applies to rank profile fields `filter-threshold`, `adjust-target` and `stopword-limit`.


geirst (approved):

vekterli:
Buildkite error looks entirely unrelated to this PR.

PR #33216, opened 2025-01-30
Remove instructions about stopping and starting services from validat…

hmusum:
…ion error message

Handled by automation on Vespa Cloud, remove instructions about starting and stopping in that case.


bjormel (approved):
Thanks!

PR #33217, opened 2025-01-30
add quoting when sorting on "[rank]" or "[relevance]"

arnej27959:
@geirst please review


bratseth (approved):

geirst (approved):
Please add a unit test as well.

PR #33220, opened 2025-01-30
Catch NoSuchMetoderror when building models

hmusum:


configserver/src/main/java/com/yahoo/vespa/config/server/modelfactory/ModelsBuilder.java
```diff
@@ -131,7 +132,9 @@ public List<MODELRESULT> buildModels(ApplicationId applicationId,
                     }
                     else {
                         log.log(Level.WARNING, "Unexpected error building " + applicationId, e);
-                        throw new InternalServerException("Unexpected error building " + applicationId, e);
+                        // Wrap it in a RuntimeException if it is an Error
+                        Exception ex = (e instanceof RuntimeException) ? (RuntimeException) e : new RuntimeException(e);
```
hakonhall, line 136:
```suggestion
                        Exception ex = (e instanceof RuntimeException re) ? re : new RuntimeException(e);
```

hakonhall (dismissed):

hakonhall (approved):

PR #33227, opened 2025-01-31
Block raw attributes from attributeMatch, dotProduct, match and

toregge:
nativeAttributeMatch rank features.
Block raw and string attributes from age, freshness and debugAttributeWait rank features.

@vekterli : please review
@geirst : FYI


geirst (approved):

vekterli (approved):
:+1:

PR #33232, opened 2025-01-31
optional hex encoding for dense parts of tensors

arnej27959:
@bjorncs and/or @bratseth please review

this is just the low-level json generation; wiring in container-search will follow in next PR.


bratseth (approved):
👍

PR #33239, opened 2025-02-03
Route all TermItems in SameElement serialization

bratseth:


bratseth:
Tbc not all term item subtypes are currently allowed in SameElement, but the decision should only be made there.

arnej27959 (approved):
LGTM

PR #33240, opened 2025-02-03
chore: push forward the expire of the Legacy Trust Store feature flag.

esolitos:
## What

Pushed forward the expiration date of the Legacy Trust Store feature flag

## Why

We're still rolling this out.


freva (approved):
Optimistic, are we? :laughing:

hmusum (approved):

PR #33246, opened 2025-02-03
[logstash-plugins] Add token authentication support

radu-gheorghe:
This is done for both the input and the output plugins.

Reminder to self: publish gems after merging 🙂


kkraune (approved):
👍

PR #33258, opened 2025-02-04
Create dummy document v1 api handler when not configured

hmusum:
If there is no document api configured (no <document-api> in services.xml) creatae a dummy handler that will return a message with pointers for how to configure it.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DummyDocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DummyDocumentV1ApiHandler.java
@@ -40,6 +40,7 @@ public final class DummyDocumentV1ApiHandler extends ThreadedHttpRequestHandler 
         public HttpErrorResponse(int code, String msg) {
             super(code);
             Cursor root = slime.setObject();
+            root.setString("error-code", "NOT_FOUND");
             root.setString("message", msg);
         }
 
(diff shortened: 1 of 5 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DummyDocumentV1ApiHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private static class HttpErrorResponse extends HttpResponse {
+
+        private final Slime slime = new Slime();
+
+        public HttpErrorResponse(int code, String msg) {
+            super(code);
+            Cursor root = slime.setObject();
+            root.setString("message", msg);
```
bjorncs, line 43: Consider adding a field `error-code` or `code` with value `404`. If the returned JSON is copy-pasted it will still indicate that this was a 404 Not Found response.

bjorncs (dismissed):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DummyDocumentV1ApiHandler.java
hmusum: Added, with value NOT_FOUND, which is what is used in other APIs

bjorncs (approved):

gjoranv (commented):
👍

PR #33264, opened 2025-02-05
Adjust query environment API to return search::index::FieldLengthInfo

toregge:
instead of average field length.

@vekterli : please review


vekterli (approved):
:+1:

PR #33276, opened 2025-02-07
Use Logstash to generate schema from data

radu-gheorghe:
The objective here is for new users to get an application package that works with their data as easily as possible. For development use, to get started.

The flow would be as follows:

1. User installs Logstash + Vespa Output Plugin and configures it to get data from the source of choice (files, DB, etc.)
2. User enables `detect_schema => true` in the Logstash config.
3. Logstash goes through the data, generates an application package and uploads it. Or upload the package manually (`deploy_package => false`).
4. User removes the `detect_schema ` flag and re-runs Logstash to upload the actual data.
5. Profit 🙂 

The initial commit of this PR is a PoC of the flow, where it should detect strings and integers. Any feedback is appreciated!

What still needs to be done (the list will likely grow during development):

- [x] **upload config**
- [x] **configurable definitions** (e.g. should we enable BM25?) again with sensible defaults. Right now we have something but it's just a PoC with string and long and it's not configurable.
- [x] **multiple field definitions per document field**. For example, we don't know if strings are intended to be used for full-text search (`index`) or grouping/ranking (`attribute`) and we may want both.
- [x] **support commonly used field types**. Need to test with some sample apps, e.g. with tensors.
- [x] **be thread-safe** (Logstash has `$num_processors` output threads by default). The plan is to use a lock file when the thread starts to read existing package files and release the lock when done writing.
- [x] **append to an existing package**. Currently we write a new package at every batch.
- [x] **deal with conflicts**. What if we thought something was an `int` and then realized it's a `float`? We should have some configurable rules there with sensible defaults.
- [x] **authentication support** (e.g. for Vespa Cloud)
- [x] add debug **logging for how each field was detected**
- [x] **better config validation**, now that we'll have a lot more config options.
- [x] **cleanup, add unit and integration tests**
- [x] **update readme and docs**

Postponed/cancelled:
- [ ] **multiple schemas support**. In Logstash we can deduce the document type name from events. I wouldn't implement it at this stage, because this feature is aimed mostly at first-time Vespa users, and they won't use multiple document types/schemas.
- [ ] **allow customized ordering of fields in the schema** (see the comment from @ldalves below). I think it would be more complicated to come up with a way for user to set rules on where each field should go. At the moment, we simply add them as we see them. The user could manually re-arrange fields before deploying. Maybe down the road we might want to have some order (e.g. string fields first), but it would still not solve the mentioned problem (e.g. array-of-struct).


ldalves:
Hey @radu-gheorghe! I haven't looked deep at the code, but here are some points that I think it would be "nice" to have:

**Schema Preview and Iteration**

- **Quick Feedback Loop**:
It would be really useful if the generated schema could be displayed immediately (for example, via a preview mode or as part of the “dry-run” output) so that users can quickly see what the schema will actually look like. I know we could get this inside the application package, but this helps in iterating over the schema design quickly.

- **Visual Diff or Versioning**:
Consider providing a “diff” view between the previous schema and the newly generated one. This can help users understand what’s changed, especially when the plugin is re-run with evolving data and the user is just trying out things. 

- **Detailed Logging and Feedback**:
Integrate detailed logs during the schema generation process - with a `--debug` flag, maybe? - so users can see why a particular field was classified a certain way. This can help in troubleshooting and understanding the automatic inference process.

**User Control Over Schema Order and Definitions**

- **Customizable Field Ordering**:
Allowing users to control the order in which fields appear in the generated schema could greatly improve the developer experience. For instance, if a user wants to have the array-of-struct definitions closer to their corresponding struct definitions (rather than having them automatically placed at the top or bottom), a configurable ordering mechanism would be very welcome. I got this experience with pyvespa where it doesn't allow this ordering and then my schema didn't ended-up the way I planed. :)

Overall, great job, the PoC looks promising. :)

radu-gheorghe:
Thanks a lot for commenting @ldalves! And for your nice words! Let me address each point:

- feedback loop: you should be able to do that today - the app package is generated and you can look at it, change it and deploy it manually. The auto-upload from Logstash is just a convenience shortcut.
- diff/versioning - I think that would be the scope of a different tool, because the point isn't so much to iterate over the app package using Logstash, but rather for Logstash to provide a good-enough application package to get started. You'd iterate on it later (using other tools, some of which are in the making :D)
- detailed logging - great idea, +1! Added it to the list in the description.
- field ordering - I can see why it's nice for visibility. Added to the list as well.

Thanks again!

integration/logstash-plugins/logstash-output-vespa/src/main/java/org/logstashplugins/SelfSignedCertGenerator.java
```diff
@@ -0,0 +1,144 @@
+package org.logstashplugins;
```
ldalves, line 1: Missing `// Copyright Vespa.ai. ....`

integration/logstash-plugins/logstash-output-vespa/README.md
```diff
@@ -2,48 +2,99 @@
...
+2. **Sending data to Vespa**: once you have an application package deployed, you can send data to Vespa.
 
-### Integration tests
-To run integration tests, you'll need to have a Vespa instance running with an app deployed that supports an "id" field. And Logstash installed.
+### Mode 1: generating an application package
 
-Check out the `integration-test` directory for more information.
+If you're just getting started with Vespa, you can use the `detect_schema` option to generate an application package that works with the data you're sending. That application package can be deployed to [Vespa Cloud](https://cloud.vespa.ai) or a local Vespa instance.
+
+To process the data, the `input` and `filter` sections are the same as for normal usage (there's an example below). The output section is different. Here's an example for Vespa Cloud:
```
kkraune, line 35: not sure what "normal usage" is - this is OSS usage, I presume?

kkraune: oh, I think normal usage is for feeding, nevermind this

kkraune (dismissed):
👍  I did not review the Java code in detail

kkraune:
@hmusum  I added you for the java part

hmusum (approved):
Java code not reviewed in detail, but generally looks good to me

radu-gheorghe:
Thanks, @hmusum and @kkraune and @ldalves. I'm merging this now.

gjoranv (commented):
I don't have a lot of domain knowledge, but this code looks good!

PR #33278, opened 2025-02-07
Increase default size of search handler thread pool

bjorncs:


andreer (approved):
Big change :) For those that already have an override of e.g. 4 they don't get 40 now right?

andreer (approved):

bjorncs:
No, it does not affect those we have overridden the default.

bjorncs:
If we observe any apps with issues we'll use the feature flag to rollback for those apps.

Issue #31619, opened 2024-06-19
Implement rank-score-drop-limit for global phase


geirst:
Currently, only `first-phase` and `second-phase` support `rank-score-drop-limit`. In some contexts, you can discover that the retrieved document was irrelevant after having spent more resources on re-ranking in `global-phase`.

This is a spinoff from https://github.com/vespa-engine/vespa/issues/29955, which added support for `rank-score-drop-limit` in `second-phase`.

dainiusjocas:
Let's close?

[Implementation](https://github.com/vespa-engine/vespa/pull/33298)
[Docs](https://github.com/vespa-engine/documentation/commit/e3f462201ca2b5a6725faedc26c31349c4c9aa0c)
[Pyvespa](https://github.com/vespa-engine/pyvespa/pull/1017)
[IDE](https://github.com/vespa-engine/vespa/pull/33381)

Issue #29955, opened 2024-01-17
Implement rank-score-drop-limit for second phase


jobergum:
Currently, only first-phase supports `rank-score-drop-limit.` In many contexts, you can discover that the retrieved document was irrelevant after the second-phase expression or even after the global phase after having spent more resources on the problem, and we should support drop limits for both the second-phase and global phase. For example, transformer-based cross-encoders have a pretty good interpretable score range.

xansrnitu:
Hi @jobergum , 
Good to see this enhancement ticket.
Currently, I am using binary quantisation (https://pyvespa.readthedocs.io/en/latest/examples/multilingual-multi-vector-reps-with-cohere-cloud.html) for my use case. However, as the cosine similarity is calculated at the second ranking phase and rank-score-drop-limit is available in the second phase, a huge list of hits are being returned from the vespa layer. Most of these don't meet my required similarity threshold and so, I drop these in my application layer. 
So, rank-score-drop-limit support in the second layer will help here. 
Until this is available, id there a way to prune my hits on the ranking layer itself, somehow?

Thanks.

jobergum:
Thank you @xansrnitu!

You can prune in a custom [searcher implementation ](https://docs.vespa.ai/en/searcher-development.html). This ticket concerns being able to prune without having to write a custom searcher implementation in both the second-phase and global-phase. There is no workarounds except a custom searcher implementation.

geirst:
_rank-score-drop-limit_ for second phase is available in Vespa 8.354.46.
Documentation: https://docs.vespa.ai/en/reference/schema-reference.html#secondphase-rank-score-drop-limit 
System test: https://github.com/vespa-engine/system-test/pull/4040, https://github.com/vespa-engine/system-test/pull/4051

geirst:
I have created https://github.com/vespa-engine/vespa/issues/31619 for supporting _rank-score-drop-limit_ in global phase. Closing this issue.

PR #33298, opened 2025-02-11
Initial commit for the global-phase rank-score-drop-limit

dainiusjocas:

Implements https://github.com/vespa-engine/vespa/issues/31619

Continuation of https://github.com/vespa-engine/vespa/issues/29955

The idea is to iterate through all hits in the global phase **after re-ranking** and drop those with a lower score than the set threshold.

Or should drop in the global phase work like in the second phase: drop within the reranked hits?

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/RankProfile.java
+++ b/config-model/src/main/java/com/yahoo/schema/RankProfile.java
@@ -111,6 +111,7 @@ public class RankProfile implements Cloneable {
     /** The drop limit used to drop hits with rank score less than or equal to this value */
     private double rankScoreDropLimit = -Double.MAX_VALUE;
     private double secondPhaseRankScoreDropLimit = -Double.MAX_VALUE;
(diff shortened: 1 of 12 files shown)


arnej27959:
This looks good, but I'm not sure about the total-hit-count adjustment; if this feature is active you won't be able to get any further hits so maybe it should be clamped to the hits you actually keep.
Please run
`mvn install -Dabicheck.writeSpec`
in the container-search module to update the ABI spec (as you added some new public methods).

arnej27959 (approved):
this looks good, we'll pull it in and I will make / extend a system test.

dainiusjocas:
I didn't fully understand your point about the total-hit-count adjustment. Is it related to paging?

arnej27959:
paging: yes - if you get a high total-hit-count it's expected that you can repeat the query asking for more hits (or with some offset) to get more results, but when this limit triggers all the hits below will be dropped as well.  I'll test it and see how it looks, it's probably OK as-is.

PR #33300, opened 2025-02-11
Handle Fuzzy items in SameElement correctly

bratseth:


vekterli (approved):
:+1: had to test this locally first to ensure the backend doesn't implode when receiving a `sameElement` + `fuzzy` combo. Looks like it works out of the box with indexed mode, but doesn't work with streaming mode.

PR #33308, opened 2025-02-12
Use deployment spec zone endpoint configuration directly

evgiz:
With the addition of `<endpoint auth-method='token'/>` attribute in `DeploymentSpecXmlReader`, using the `DeploymentSpec` directly in this builder should be valid (but may not consider whether `token` is present in `services.xml`). I don't believe `ContainerEndpoint` can be used, given that auth data from `ZoneEndpoint` config is lost during endpoint generation and we can't know whether the endpoint should actually be in private DNS. 

Could you verify that this is valid approach @mpolden ?


mpolden (approved):
Agree, I think this is unnecessary now.

PR #33322, opened 2025-02-14
Suuport feature flag RESOURCE_LIMIT_LOW_WATERMARK_DIFFERENCE

hmusum:
Support setting cluster controller config based on the feature flag. Config set is cluster_feed_block_noise_level, which by default is 0.0.

The flag value is an absolute number, so e.g. 0.01 implies that a block limit of 0.8 effectively becomes 0.79 for an already blocked node. This value will be in effect for both disk and memory resource limits.


vekterli (approved):
:+1:

PR #33330, opened 2025-02-14
Return 404 if document type does not exist

johansolbakken:


vekterli (dismissed):
:+1:

vekterli:
I see there are a couple of `storageserver` tests that need to have their source of truth updated:
```
[  FAILED  ] CommunicationManagerTest.unmapped_bucket_space_documentapi_request_returns_error_reply
[  FAILED  ] CommunicationManagerTest.unmapped_bucket_space_for_get_documentapi_request_returns_error_reply
```
After a bit of additional pondering on the issue I'm also wondering if we want to restrict _when_ we remap this error code to 404. It makes sense for Gets, but the bucket space resolving logic is also invoked for mutating operations like Puts. What would be the appropriate error code here? It's still conceptually trying to go via a resource (document type) that does not exist, but it's not a read-operation... The existing 500 response is very suboptimal either way.

vekterli (approved):
:+1:

PR #33331, opened 2025-02-14
Avoid parsing which will be overwritten

bratseth:
The string in this test fails the heuristic parser and that caused an error even when overriding the grammar since the query string was lazily parsed with the default grammar just to be immediately thrown away.


arnej27959 (approved):
👍

PR #33339, opened 2025-02-17
Allow for multiple certificates and comments in security/clients.pem

ean:
This will allow deployments using the CLI when client.pem contain more data than only the local client certificiate.


mpolden (commented):
LGTM, but could use a test.

ean:
> LGTM, but could use a test.

Agreed, fixed now.

mpolden (approved):

PR #33349, opened 2025-02-18
Global phase rank-score-drop-limit query override

dainiusjocas:

It seems that currently, it is not possible to override `ranking.globalPhase.rankScoreDropLimit`. Hope this should help.


arnej27959 (approved):
👍

PR #33357, opened 2025-02-19
remove always expensive phrases support

havardpe:
(the feature flag has been removed)

@hmusum please review


hmusum (approved):
Looks good to me 👍

PR #33358, opened 2025-02-19
Added cause exception message to error message in LLMSearcher.

glebashnik:

### DIFF
--- a/container-search/src/main/java/ai/vespa/search/llm/LLMSearcher.java
+++ b/container-search/src/main/java/ai/vespa/search/llm/LLMSearcher.java
@@ -177,8 +177,7 @@ public class LLMSearcher extends Searcher {
         } else if (exception.getCause() instanceof LanguageModelException) {
             errorCode = ((LanguageModelException) exception.getCause()).code();
         }
-        eventStream.error(languageModelId, new ErrorMessage(errorCode, exception.getMessage() +
-                (exception.getCause() != null ? ": " + exception.getCause().getMessage() : "")));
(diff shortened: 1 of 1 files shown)


container-search/src/main/java/ai/vespa/search/llm/LLMSearcher.java
```diff
@@ -177,7 +177,8 @@ private void handleException(EventStream eventStream, Throwable exception) {
         } else if (exception.getCause() instanceof LanguageModelException) {
             errorCode = ((LanguageModelException) exception.getCause()).code();
         }
-        eventStream.error(languageModelId, new ErrorMessage(errorCode, exception.getMessage()));
+        eventStream.error(languageModelId, new ErrorMessage(errorCode, exception.getMessage() +
+                (exception.getCause() != null ? ": " + exception.getCause().getMessage() : "")));
```
bratseth, line 181:
```suggestion
        eventStream.error(languageModelId, new ErrorMessage(errorCode, "Error in LLM text generation", com.yahoo.yolean.Exceptions.toMessageString(exception)));
```

bratseth: 1. Suggest using both a high level and detail message.
2. Generate the detail message using the utility for it.

bratseth (changes requested):

container-search/src/main/java/ai/vespa/search/llm/LLMSearcher.java
glebashnik: Neat!

bratseth (approved):

PR #33367, opened 2025-02-20
Add constructor taking `ContainerThreadPool`

bjorncs:
Ensure handler is using the thread pool intended for request handlers, not the common pool.

FYI @glebashnik @gjoranv


hmusum (approved):

gjoranv (commented):
👍

PR #33375, opened 2025-02-20
Prepare for removing feature flag ENFORCE_STRICTLY_INCREASING_CLUSTER…

hmusum:
…_STATE_VERSIONS

Set feature flag default value and config default value to true. Value for feature flag definition has been true for a long time, and there are no feature flag overrides.


vekterli (approved):
:+1:

PR #33383, opened 2025-02-21
Revert "check if timeout actually happens before giving it as a reason"

aressem:
Reverts vespa-engine/vespa#33374


aressem:
Suspecting that this one broke : https://factory.vespa.ai/testrun/106551701/test/ReportCoverage::test_reportcoverage__INDEXED

bjorncs (approved):

hmusum:
Yes, could be that test needs an update, reporting timeout true when this was not the real reason was one thing that was fixed in Arne's changes, to suggest to hold this a bit

PR #33384, opened 2025-02-21
Do not retire if host-flavor flag change

freva:
I believe this was added before we had `HostFlavorUpgrader`, we should use that instead as this will retire all nodes in cluster at the same time, which we almost never want, and if we do, we can manually retire instead.


hakonhall (approved):
great

freva:
FYI: @bjormel

PR #33388, opened 2025-02-21
timelines for query and worst thread

havardpe:
also flat profiling for first and second phase ranking for the worst thread

worst performers (search/thread) is compared to the (exclusive) average of the other samples


havardpe:
@andreer please review

andreer (approved):

PR #33403, opened 2025-02-24
Arnej/add gpu types

arnej27959:
@bratseth please review
@hmusum @oyving FYI

Note: the "plus", "minus", and "multiply" methods in GpuResources change semantics here, needs double-checking.

For later: The "cost()" method should probably take gpuType into account .

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
@@ -167,18 +167,21 @@ public class NodeResources {
             if (other.isZero()) return this;
             if (this.isZero()) return other;
             if (this.type != other.type) throw new IllegalArgumentException("Bad add: " + this + " + " + other);
-            double mem = Math.max(this.memoryGiB, other.memoryGiB);
-            return new NodeResources.GpuResources(this.type, this.count() + other.count(), mem);
+            var thisMem = this.count() * this.memoryGiB();
+            var otherMem = other.count() * other.memoryGiB();
+            return new NodeResources.GpuResources(this.type, 1, thisMem + otherMem);
         }
 
         public GpuResources minus(GpuResources other) {
-            if (other.isZero() || this.isZero()) return this;
+            if (other.isZero()) return this;
             if (this.type != other.type) throw new IllegalArgumentException("Bad minus: " + this + " + " + other);
-            return new NodeResources.GpuResources(this.type, this.count() - other.count(), this.memoryGiB);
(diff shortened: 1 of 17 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
```diff
@@ -145,33 +160,20 @@ private boolean lessThan(GpuResources other) {
 
         public GpuResources plus(GpuResources other) {
             if (other.isZero()) return this;
-            var thisMem = this.count() * this.memoryGiB();
-            var otherMem = other.count() * other.memoryGiB();
-            return new NodeResources.GpuResources(1, thisMem + otherMem);
+            if (this.isZero()) return other;
+            if (this.type != other.type) throw new IllegalArgumentException("Bad add: " + this + " + " + other);
+            double mem = Math.max(this.memoryGiB, other.memoryGiB);
```
bratseth, line 170: Are you changing the definition this way on purpose? The previous definition preserved the total memory of the two resources added, and since memory is what we charge for that does make sense. With this, 1 gpu with 8Gb memory + 10 with 4Gb become 11 with 8Gb memory. Wouldn't the minimal change you need here just be to only add the type verification?

config-provisioning/src/main/java/com/yahoo/config/provision/NodeResources.java
```diff
@@ -145,33 +160,20 @@ private boolean lessThan(GpuResources other) {
...
         }
 
         public GpuResources minus(GpuResources other) {
-            if (other.isZero()) return this;
-            var thisMem = this.count() * this.memoryGiB();
-            var otherMem = other.count() * other.memoryGiB();
-            return new NodeResources.GpuResources(1, thisMem - otherMem);
+            if (other.isZero() || this.isZero()) return this;
+            if (this.type != other.type) throw new IllegalArgumentException("Bad minus: " + this + " + " + other);
+            return new NodeResources.GpuResources(this.type, this.count() - other.count(), this.memoryGiB);
```
bratseth, line 177: And here memory differences are ignored?

bratseth (commented):
LGTM apart from the questions on plus/minus.

arnej27959:
I'll revert the plus/minus changes; I wanted to be proactive since we probably want to change pricing to be based on GPU type instead of just memory GB, but that probably needs some more design (and we must fix the databases etc used for resources and billing to have GPU type available there as well first).

arnej27959:
Please take another look now

arnej27959:
@bratseth ping

bratseth (approved):
👍

PR #33409, opened 2025-02-24
Use feature macros to check for constexpr std::string support.

toregge:
@vekterli : please review


toregge:
No longer relevant. gcc14-g++ is used on AlmaLinux 2023.

PR #33413, opened 2025-02-24
stop forcing timeout as reason

arnej27959:
this code is no longer needed and forced "timeout" as reason in all cases, making some system tests unhappy.


hmusum (approved):
Looks good to me

bratseth (commented):
👍

PR #33422, opened 2025-02-25
Log if min-node-ratio-per-group is low and there are many small groups.

hmusum:
It's better to advise users to set this to 1 in these cases, so the whole group goes down if one node goes down, so log and point to doc about this.

An alternative i to just set min-node-ratio-per-group to 1 in code and don't log.


vekterli (approved):
:+1:

PR #33423, opened 2025-02-25
Stop using feature flag ENFORCE_STRICTLY_INCREASING_CLUSTER_STATE_VER…

hmusum:
…SIONS

Default config value is true, so this is a no-op


vekterli (approved):
:+1:

PR #33426, opened 2025-02-26
batch chars when escaping json

andreer:


andreer:
JMH says it's an improvement (although less so for small strings)
```
Benchmark          Mode  Cnt     Score     Error  Units
StuffDoer.render  thrpt    5  1909.663 ± 231.894  ops/s

Benchmark          Mode  Cnt     Score     Error  Units
StuffDoer.render  thrpt    5  2973.112 ± 593.253  ops/s
```

bratseth (dismissed):
Maybe add a comment so it does get cleaned up later, otherwise looks .... like this kind of code looks to me!

bjorncs (approved):

PR #33428, opened 2025-02-26
Estimate bytes read from disk as part of flush and use this estimate

toregge:
in prepare for restart flush strategy. This allows prepare restart flush strategy to consider paged attributes as more expensive to flush.

@vekterli : please review
@geirst : FYI


geirst (approved):

vekterli (approved):
:+1:

PR #33433, opened 2025-02-26
Stop using enforceStrictlyIncreasingClusterStateVersions()

hmusum:
No need to set config value, default value is true.


vekterli (approved):
:+1:

PR #33436, opened 2025-02-26
Bratseth/type resolution cleanup

bratseth:


bratseth:
@arnej27959 just some more cleanup + fix an unwarranted assumption in CatExpression

arnej27959 (approved):
seems good to me

PR #33443, opened 2025-02-27
Support overriding filterThreshold per query.

geirst:
This is the equivalent of setting filter-threshold in a rank profile.

@toregge please review
@vekterli FYI

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/ranking/Matching.java
+++ b/container-search/src/main/java/com/yahoo/search/query/ranking/Matching.java
@@ -26,6 +26,7 @@ public class Matching implements Cloneable {
     public static final String POST_FILTER_THRESHOLD = "postFilterThreshold";
     public static final String APPROXIMATE_THRESHOLD = "approximateThreshold";
     public static final String TARGET_HITS_MAX_ADJUSTMENT_FACTOR = "targetHitsMaxAdjustmentFactor";
+    public static final String FILTER_THRESHOLD = "filterThreshold";
     public static final String WEAKAND = "weakand";
 
     static {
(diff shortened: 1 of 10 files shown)


container-search/src/main/java/com/yahoo/search/query/ranking/Matching.java
```diff
@@ -52,6 +54,7 @@ public class Matching implements Cloneable {
     private Double postFilterThreshold = null;
     private Double approximateThreshold = null;
     private Double targetHitsMaxAdjustmentFactor = null;
+    private Double filterThreshold;
```
toregge, line 57: Consider initializing `filterThreshold` to `null`, as in the lines above.

toregge (dismissed):

container-search/src/main/java/com/yahoo/search/query/ranking/Matching.java
geirst: Fixed

geirst:
@toregge PTAL

searchlib/src/vespa/searchlib/diskindex/disktermblueprint.cpp
```diff
@@ -38,12 +38,12 @@ getName(uint32_t indexId)
 
 }
 
-DiskTermBlueprint::DiskTermBlueprint(const FieldSpec & field,
+DiskTermBlueprint::DiskTermBlueprint(const FieldSpec field,
```
vekterli, line 41: Should this be un-`const`ed now that it's by value and is being `std::move`d away from?

geirst: Yes, fixed

toregge (approved):

PR #33447, opened 2025-02-27
chore: update paths for new Yahoo certificate bundle

esolitos:
## What

* Replace `yahoo_certificate_bundle.[jks|pem|pkcs12]` with the updated `yahoo_tw_certificate_bundle.[jks|pem|pkcs12]`
* Replace the `athenz_certificate_bundle.[jks|pem|pkcs12]` with the updated `athenz_tw_certificate_bundle.[jks|pem|pkcs12]`

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/otel/OpenTelemetryConfigGenerator.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/otel/OpenTelemetryConfigGenerator.java
@@ -20,9 +20,9 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


ean (dismissed):

config-model/src/main/java/com/yahoo/vespa/model/admin/otel/OpenTelemetryConfigGenerator.java
freva, line 1: Not needed

vespa-athenz/src/main/resources/configdefinitions/vespa.athenz.identity.sia-provider.def
freva, line 1: We should not change the default, instead override explicitly where this is needed.

freva (approved):

PR #33457, opened 2025-03-01
add Java program generating data for IsWordChar()

arnej27959:
@toregge (or anyone else) please review


toregge (dismissed):

arnej27959:
@bratseth please review the Java bits
@toregge please review the C++ bits

bratseth:
LGTM! 👍   Did you not include isSymbol() (emojis) as word characters on purpose? (I'm not sure if we should here)

arnej27959:
The emojis etc will have to be handled separately later, yes.

bratseth (approved):

PR #33458, opened 2025-03-01
remove some unreachable code

arnej27959:
@toregge please review and merge


vespalib/src/vespa/fastlib/text/unicodeutil.cpp
```diff
@@ -6,28 +6,6 @@
...
 
-namespace {
-
-class Initialize
-{
-public:
-    Initialize() noexcept { Fast_UnicodeUtil::InitTables(); }
-};
-
-Initialize _g_initializer;
```
toregge, line 17: This instance caused the constructor for `Initialize` to be called, thus `Fast_UnicodeUtil::InitTables()` is called once, before main is called at program startup time. But InitTables doesn't seem to do anything useful.

toregge (approved):

PR #33465, opened 2025-03-03
Add data plane token authentication support to vespa cli

evgiz:
Usage:

```bash
$ vespa status
Container default at https://foo.bar.z.vespa-app.cloud/ is ready (token)
Container default at https://bar.foo.z.vespa-app.cloud/ is ready (mtls)

$ export VESPA_CLI_DATA_PLANE_TOKEN="my-token"
$ vespa query <my-query>
<success response from foo.bar endpoint>
```

Supports `vespa query` and `vespa feed` and resolves token endpoint automatically when the environment variable is set. Requires a token endpoint in the cluster. May break tools depending on a single url output from `vespa status --format plain`.


evgiz:
> In this PR it will always default to the token endpoint if the environment variable is set?

Yes, if `VESPA_CLI_DATA_PLANE_TOKEN` environment variable is set, `query` and `feed` will only resolve `token` endpoints. If one does not exist, the command will fail with an error that token endpoint is missing.

client/go/internal/cli/cmd/feed.go
```diff
@@ -136,10 +136,18 @@ func createServices(n int, timeout time.Duration, cli *CLI, waiter *Waiter) ([]h
 	if err != nil {
 		return nil, "", err
 	}
+
+	token := cli.Environment["VESPA_CLI_DATA_PLANE_TOKEN"]
+	authMethod := "mtls"
+	if token != "" {
+		authMethod = "token"
+	}
```
esolitos, line 144: Should we add a debug message here?

client/go/internal/cli/cmd/waiter.go
```diff
@@ -33,6 +33,10 @@ func (w *Waiter) DeployService(target vespa.Target) (*vespa.Service, error) {
 
 // Service returns the service identified by cluster ID, available on target.
 func (w *Waiter) Service(target vespa.Target, cluster string) (*vespa.Service, error) {
+	return w.ServiceWithAuthMethod(target, cluster, "mtls")
+}
+
+func (w *Waiter) ServiceWithAuthMethod(target vespa.Target, cluster string, authMethod string) (*vespa.Service, error) {
```
esolitos, line 39: Or perhaps the debug message about which Authentication method is used could be set directly in there and when `VESPA_CLI_DATA_PLANE_TOKEN` we simply mention that it's found.

client/go/internal/vespa/target.go
```diff
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fatih/color"
```
esolitos, line 17: Needs to weigh adding a dependency vs. defining a constant for the ANSI color here. 🤔

client/go/internal/vespa/target.go
```diff
@@ -221,16 +223,26 @@ func (s *Service) Description() string {
...
 	}
-	names := make([]string, len(services))
-	for i, s := range services {
+	names := make([]string, 0, len(services))
+	for _, s := range services {
 		if name == s.Name {
 			return s, nil
 		}
-		names[i] = s.Name
+		var prettyName = color.CyanString("%s", s.Name)
```
esolitos, line 241: When adding ANSI color we should also add an option to print colourless, not all shells support ansi colors.

Furthermore: if we use colors, we should define a system and apply thought out the whole messages, not just "in one place". 😊

esolitos (commented):
Did a quick review with some suggestions. 😊 

Additional question: *Do we cache the endpoints in any form, or are they re-checked each time?*

client/go/internal/cli/cmd/waiter.go
evgiz: Good idea, I'll look into the cli tracing system

client/go/internal/vespa/target.go
evgiz: `github.com/fatih/color` is already a dependency!

client/go/internal/vespa/target.go
esolitos: 👍🏼  Right, I didn't realize that was already a dependency, apologies. 😊

arnej27959 (approved):
this seems reasonable; we'll need to add it to the "document" command (and refactor some common code) as well, but that's for another PR.

PR #33480, opened 2025-03-03
Rewrite searchlib features unit tests to gtest.

toregge:
@vekterli : please review

### DIFF
--- a/searchlib/src/tests/features/constant/constant_test.cpp
+++ b/searchlib/src/tests/features/constant/constant_test.cpp
@@ -11,8 +11,7 @@
 #include <vespa/eval/eval/value.h>
 #include <vespa/eval/eval/test/value_compare.h>
 #include <vespa/vespalib/util/stringfmt.h>
-#include <vespa/vespalib/testkit/test_kit.h>
-#include <vespa/vespalib/testkit/test_master.hpp>
+#include <vespa/vespalib/gtest/gtest.h>
 
 using search::feature_t;
 using namespace search::fef;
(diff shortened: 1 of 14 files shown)


searchlib/src/tests/features/constant/constant_test.cpp
```diff
@@ -36,28 +35,25 @@ Value::UP make_tensor(const TensorSpec &spec) {
...
-    {
-        setup_search_features(factory);
-    }
+    ConstantTest();
+    ~ConstantTest() override;
     bool setup() { return test.setup(); }
     const Value &extractTensor(uint32_t docid) {
         Value::CREF value = test.resolveObjectFeature(docid);
-        ASSERT_TRUE(value.get().type().has_dimensions());
+        EXPECT_TRUE(value.get().type().has_dimensions());
```
vekterli, line 48: Is this required to hold to prevent the caller from crashing instead of failing the test gracefully? If so, maybe it should be a regular `assert` instead

searchlib/src/tests/features/constant/constant_test.cpp
```diff
@@ -36,28 +35,25 @@ Value::UP make_tensor(const TensorSpec &spec) {
...
         return value.get();
     }
     const Value &executeTensor(uint32_t docId = 1) {
         return extractTensor(docId);
     }
     double extractDouble(uint32_t docid) {
         Value::CREF value = test.resolveObjectFeature(docid);
-        ASSERT_TRUE(value.get().type().is_double());
+        EXPECT_TRUE(value.get().type().is_double());
```
vekterli, line 56: Same here (unless `as_double` can throw or return a sentinel value)

vekterli (approved):
:+1:

searchlib/src/tests/features/constant/constant_test.cpp
toregge: No. operator== makes transient tensor specs and comares those.

PR #33482, opened 2025-03-04
Explore hnsw index data stores.

toregge:
@vekterli : please review
@geirst : FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/tensor/hnsw_index_explorer.cpp
+++ b/searchlib/src/vespa/searchlib/tensor/hnsw_index_explorer.cpp
@@ -10,6 +10,14 @@ using vespalib::slime::ObjectInserter;
 
 namespace search::tensor {
 
+namespace {
+
+const std::string LEVELS_STORE_NAME("levels_store");
+const std::string LINKS_STORE_NAME("links_store");
+const std::string NODEID_STORE_NAME("nodeid_store");
+
+}
+
 template <HnswIndexType type>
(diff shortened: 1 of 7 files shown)


searchlib/src/vespa/searchlib/tensor/hnsw_index_explorer.cpp
```diff
@@ -63,6 +71,30 @@ HnswIndexExplorer<type>::get_state(const vespalib::slime::Inserter& inserter, bo
                    cfg.neighbors_to_explore_at_construction());
 }
 
+template <HnswIndexType type>
+std::vector<std::string>
+HnswIndexExplorer<type>::get_children_names() const
+{
+    return { LEVELS_STORE_NAME, LINKS_STORE_NAME, NODEID_STORE_NAME };
```
vekterli, line 78: To be symmetric with `get_child`, should `NODEID_STORE_NAME` only be returned if `type == HnswIndexType::MULTI`?

vekterli (approved):
:+1:

searchlib/src/vespa/searchlib/tensor/hnsw_index_explorer.cpp
toregge: The name is effectively hidden if get_child returns an empty unique_ptr for the name. The same can bee seen in `Proton::get_children_names` vs `Proton::get_child`.

PR #33490, opened 2025-03-04
Extend expiry date for feature flag

hmusum:
Maybe this should be in PermanentFlags?


arnej27959 (approved):
Will try to remove this flag soonish. Extend for now.

PR #33491, opened 2025-03-04
add more token support (document and visit)

arnej27959:
move debug-mode flag to common code, but make it hidden
use printDebug instead of trace, enabled by --debug-mode
separate out some common code for token handling

Note: also clears key/cert file options, to avoid confusion, especially when printing the curl command

@evgiz please review
@bratseth FYI


evgiz (dismissed):
Nice! `--debug-mode` flag is a useful addition

esolitos (dismissed):

esolitos (approved):

PR #33493, opened 2025-03-04
Show number of true bits for single bool attribute in attribute vector

toregge:
state explorer.

@vekterli : please review


geirst (approved):

vekterli (approved):
:+1:

PR #33494, opened 2025-03-04
Take number of document dbs into account when calculating initial number

toregge:
of entries in document meta store vector for streaming mode.

@vekterli : please review
@geirst : FYI


geirst (approved):

vekterli (approved):
:+1:

PR #33496, opened 2025-03-05
Don't log at info level until some time has passed

hmusum:
Log at debug level when not able to download application package, but start logging at info level when some time has passed.

### DIFF
--- a/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
+++ b/configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
@@ -132,7 +132,7 @@ public class ApplicationPackageMaintainer extends ConfigServerMaintainer {
     }
 
     private void logFailure(Session session, FileReference fileReference, ApplicationId applicationId) {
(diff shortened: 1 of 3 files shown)


configserver/src/main/java/com/yahoo/vespa/config/server/maintenance/ApplicationPackageMaintainer.java
```diff
@@ -129,4 +131,13 @@ private void createLocalSessionIfMissing(ApplicationId applicationId, long sessi
             sessionRepository.createLocalSessionFromDistributedApplicationPackage(sessionId);
     }
 
+    private void logFailure(Session session, FileReference fileReference, ApplicationId applicationId) {
+        var deployed = Instant.ofEpochSecond(session.getMetaData().getDeployTimestamp());
```
hakonhall, line 135: You document deployTimestamp as being in milliseconds since Epoch but use ofEpochSecond?

hmusum: Thanks, fixed

hakonhall (approved):

PR #33512, opened 2025-03-06
special case json rendering for array of string

andreer:


bjorncs (dismissed):
Consider adding a unit test for optimized path if not already present

container-search/src/main/java/com/yahoo/search/rendering/JsonRenderer.java
```diff
@@ -751,6 +751,27 @@ private Inspector maybeConvertData(Inspector data) {
...
+                    }
+                }
+
+                generator.writeStartArray();
+                for (int i = 0; i < entries; i++) {
+                    byte[] utf8 = data.entry(i).asUtf8();
+                    try {
+                        generator.writeUTF8String(utf8, 0, utf8.length);
+                    } catch (IOException ex) {
+                        throw new RuntimeException(ex);
```
bjorncs, line 769: You may want to leave a comment here if exception should never happen. If not ensure that `RuntimeException` is handled correctly. Other code paths wraps this in `UncheckedIOException`. You can likely just not catch is as `renderInspector` already throws `IOException`

andreer: Thanks, completely missed this from my PoC implementation

bjorncs (dismissed):

bjorncs (approved):

PR #33516, opened 2025-03-06
Let System know its target type.

arnej27959:
This helps getting us the correct target type when detecting the system from an URL target.

@bratseth please review
@lperry25 @evgiz FYI


bratseth (approved):
👍

PR #33518, opened 2025-03-06
deprecate and do API cleanup

arnej27959:
@vekterli please review
@bjorncs FYI

note: this changes some APIs that it's not expected that any vespa users will use, but it should be backwards (source) compatible anyways.  The goal here is to mark the "6" version as deprecated and only use it indirectly, so it can be removed at some later time.


vekterli (approved):
:+1:

PR #33528, opened 2025-03-07
Respect searchable-copies 1 also with min-redundancy

bratseth:


config-model/src/main/java/com/yahoo/vespa/model/content/cluster/RedundancyBuilder.java
```diff
@@ -60,6 +62,7 @@ public Redundancy build(boolean isHosted, Boolean isStreaming, int subGroups, in
         if (isHosted) {
             if (globalMinRedundancy != null && ( finalRedundancy == null || finalRedundancy * leafGroups < globalMinRedundancy ))
                 initialRedundancy = finalRedundancy = (int)Math.ceil((double)globalMinRedundancy / leafGroups);
+            System.out.println("ready copies here is " + readyCopies);
```
bjormel: Remove the `System.out.println`

bjormel: I did

bjormel (approved):

PR #33535, opened 2025-03-10
Increase number of buffers for TensorBufferStore from 8Ki to 16Ki.

toregge:
This comes with a cost of 192Ki extra bytes used for the data store (8Ki * sizeof(vespalib::datastore::BufferAndMeta)).

@vekterli : please review
@geirst : FYI


vekterli (approved):
:+1:

PR #33543, opened 2025-03-11
Change default value of use_async_message_handling_on_schedule to true

hmusum:


hmusum:
Got a segfault in PR build here, would like some help here @vekterli

vekterli:
Likely some tests that implicitly expect the config to be `false` and have message ordering or barrier semantics that depend on this. Will take a look.

PR #33547, opened 2025-03-11
Revert "Respect searchable-copies 1 also with min-redundancy"

hmusum:
Reverts vespa-engine/vespa#33528

Fails for a customer that has this config:

```
        <min-redundancy>2</min-redundancy>

 <engine>
            <proton>
                <searchable-copies>2</searchable-copies>

        <nodes deploy:environment="dev" count="2" groups="2">
```

with:

```

Caused by: com.yahoo.vespa.config.server.http.InvalidApplicationException: Invalid application
...
Caused by: java.lang.IllegalArgumentException: In content cluster 'foo'
...	
Caused by: java.lang.IllegalArgumentException: Number of searchable copies (2) can not be higher than final redundancy (1)
```


bratseth:
Thanks. They need it fixed though ...

PR #33548, opened 2025-03-11
Non-functional changes only

bratseth:


gjoranv (approved):

gjoranv (changes requested):
The BlendingSearchTestCase fails, where this wrapper ctor is used:
```
        public BlendingSearcherWrapper() {
            blendingField = null;
        }
```
From what I can see, it would work to replace null with empty string.

bratseth:
Fixed - thanks!

gjoranv (approved):

PR #33557, opened 2025-03-11
Add internal malloc information to state explorer

vekterli:

Adds a new `mallocinfo` sub-path to the top-level custom component structure, which emits malloc-related information.

In particular:
 1. The detected malloc implementation, if known.
 2. Implementation-independent info via `mallinfo()` or `mallinfo2()`, if supported by the platform. Keep a manpage at the ready when deciphering the fields...
 3. Malloc implementation-specific information for the implementations we know about. Currently only covers `vespamalloc`.

Internal `vespamalloc` information is retrieved by going via a weak symbol reference that is valid (i.e. non-`nullptr`) iff the expected symbol is exposed from the `vespamalloc` shared library. This is also how we will detect other malloc implementations at runtime in the future(tm).

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/server/CMakeLists.txt
+++ b/searchcore/src/vespa/searchcore/proton/server/CMakeLists.txt
@@ -70,6 +70,7 @@ vespa_add_library(searchcore_server STATIC
     maintenancecontroller.cpp
     maintenancedocumentsubdb.cpp
     maintenancejobrunner.cpp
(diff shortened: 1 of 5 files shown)


geirst (approved):

toregge (approved):

vekterli:
Will add this to the `CustomStateApi` system test since it checks the top-level paths. Holding off merge until that is ready.

PR #33559, opened 2025-03-12
Use openai java client

thomasht86:

Adding Official OpenAI Java Client library. 
Will enable us to leverage that for e.g. Structured Output, Embeddings, Moderations ++ in the future. 

- Moved the `OpenAiClient` from `vespajlib`to `model-integration`.
- Added [Official OpenAI Java client](https://github.com/openai/openai-java) as dependency.
- Rewrite `OpenAiClient` to use the official Client. 
- Add more `Completion.Finishreasons`. 
- Expand with tests to test for compatibility with other openai-compatible inference providers (together.ai)

Please run tests and review critically. (I can share api-keys in 1password)


thomasht86:
Will merge `OpenAIClient` with `OpenAI` and make it implement `ConfigurableLanguageModel` as discussed.

thomasht86:
please review @glebashnik @bjorncs

bjorncs (commented):
See comments

model-integration/src/main/java/ai/vespa/llm/clients/OpenAI.java
```diff
@@ -2,71 +2,215 @@
...
+            return defaultSyncClient;
+        }
+        return OpenAIOkHttpClient.builder()
+                .apiKey(apiKey)
+                .baseUrl(endpoint)
+                .responseValidation(false)
+                .build();
+    }
+    
+    private OpenAIClientAsync getAsyncClient(String apiKey, String endpoint) {
```
glebashnik, line 97: @thomasht86 It looks to me that this way caching will never work since cachedApiKey and cachedEndpoint are nulls to start with and the only way to set their values is for them not be null.

In addition, defaultAsyncClient needs to be null to be set to a new value, which means that it will only be set the first time and never reset. Is it intentional?

glebashnik: Same for getSyncClient

model-integration/src/main/java/ai/vespa/llm/clients/OpenAI.java
```diff
@@ -2,71 +2,215 @@
...
     @Override
     public CompletableFuture<Completion.FinishReason> completeAsync(
             Prompt prompt, InferenceParameters parameters, Consumer<Completion> consumer) {
         var preparedParameters = prepareParameters(parameters);
-        return client.completeAsync(prompt, preparedParameters, consumer);
+        String apiKey = preparedParameters.getApiKey().orElse(null);
+        String endpoint = preparedParameters.getEndpoint().orElse(DEFAULT_ENDPOINT);
+        
+        if (apiKey == null) {
+            throw new IllegalArgumentException("API key must be provided either in configuration or in InferenceParameters");
```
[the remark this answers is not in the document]
glebashnik: Right, it should be completed exceptionally instead of throwing exception directly.
Otherwise LLM searcher that uses completeAsync will not handle it correctly.

glebashnik (changes requested):

model-integration/src/main/java/ai/vespa/llm/clients/OpenAI.java
thomasht86: they are set here:

                cachedApiKey = apiKey;
                cachedEndpoint = endpoint;

won't that work?

vespajlib/src/test/java/ai/vespa/llm/client/openai/OpenAiClientCompletionTest.java
```diff
@@ -1,46 +0,0 @@
...
-import ai.vespa.llm.completion.StringPrompt;
-import org.junit.jupiter.api.Disabled;
-import org.junit.jupiter.api.Test;
-
-import java.util.Map;
-
-/**
- * @author bratseth
- */
-public class OpenAiClientCompletionTest {
```
glebashnik, line 14: Would be nice to have tests for error handling.

glebashnik (changes requested):

glebashnik (approved):
"consider" refactorings are optional.

PR #33573, opened 2025-03-13
Magnus/lsp xml preprocessing

Mangern:
Fix problem of showing errors on valid services.xml. It makes the validation less strict, so some invalid xml documents may now not show errors.

Also it now doesn't run on other xml files than services.xml.


theodorklauritzen (approved):
Great work!

PR #33580, opened 2025-03-13
Add metric for HTTP compliance violations

bjorncs:
@olaaun Please review the metric wiring.
@glebashnik Please review everything else.


glebashnik (approved):

olaaun:
Is this metric intended only for Jetty 12 rollout? If so, consider adding it to the Vespa9 set only, so that we can remove it when not needed anymore

bjorncs:
@olaaun We need it in the legacy metric system as well. Which metric set does that use?

olaaun:
Ack, that's `Vespa`

olaaun (approved):

PR #33578, opened 2025-03-13
Fail if 'boost' is lower than 'threads'

bjorncs:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #33581, opened 2025-03-13
Revert "Fail if 'boost' is lower than 'threads'"

hmusum:
Reverts vespa-engine/vespa#33578

One integration test started failing, customer needs to update their app first.

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerThreadpool.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/container/ContainerThreadpool.java
@@ -50,8 +50,6 @@ public abstract class ContainerThreadpool extends SimpleComponent implements Con
             if (threadsElem != null) {
                 min = parseMultiplier(threadsElem.getTextContent());
                 max = threadsElem.hasAttribute("boost") ? parseMultiplier(threadsElem.getAttribute("boost")) : min;
-                if (Math.abs(min) > Math.abs(max))
-                    throw new IllegalArgumentException("For <threadpool>: 'boost' must be greater than <threads>");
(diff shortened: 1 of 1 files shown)


arnej27959 (approved):

bjormel:
Maybe we rather should change the implementation to match the semantics instead?
Some sample definitions of the term 'boost' 
> If one thing boosts another, it causes it to increase, improve, or be more 
> Britannica Dictionary definition of BOOST
> [count]
> an increase in amount
>> a boost in wheat production
>> a boost in sales
>> Exercise can sometimes provide a boost of energy.
>> After layoffs at the company, employees needed a boost in morale.

bratseth:
I agree but maybe too late to change now ...

bjormel:
The documentation is in line with the semantics, alt least the first part:
https://docs.vespa.ai/en/reference/services-search.html#threads.boost
> The number of **additional** threads relative to CPU cores. 
But we also say
> Default value is 10. Boost will be disabled if boost==threads.
Which contradicts the first statement

bjormel:
I was trying to update the documentation to match the code, but when I read what I had written... it felt wrong:
> The maximum number of threads relative to CPU cores. <boost> - <threads> will be created on demand when the thread pool is fully utilized. Default value is <threads>, which implies that no extra threads will be created on demand.

hmusum:
@bjormel reading the code and the doc I think we could introduce e.g `max` instead and deprecate `boost`. That will make the doc you write better. Your doc is still correct, though, and better than we have today IMHO.

bjormel:
I will update the doc then, waiting for the perfect solution 👍

PR #33587, opened 2025-03-14
LSP: Open the yql query results file in a split screen

theodorklauritzen:

This will open the results from a YQL query in vscode in a split screen, and not a new tab.


Mangern (commented):
We can do it by exploiting 'takeFocus' flag instead.

PR #33588, opened 2025-03-14
Use search_int instead of assertInt in streamingvisitors searcher

toregge:
unit test.

@vekterli : please review


vekterli (approved):
:+1:

PR #33589, opened 2025-03-14
Do response metric aggregation through Jetty's request log interface

bjorncs:
Handlers are not invoked if the request is failed during HTTP request parsing, e.g. invalid headers. Use the request log interface instead to ensure we those responses are covered in our metrics.


freva (approved):

bjorncs:
FYI @glebashnik @esolitos

PR #33590, opened 2025-03-14
First step in changing MMAP madvise default to SEQUENTIAL for search.

geirst:
Benchmarking of lexical search using a Wikipedia dataset has shown that madvise SEQUENTIAL for disk index posting list files results in lower 99 percentile query latencies and better utilization of the I/O subsystem.

Rollout is complete for node flavors with slow disks.

@vekterli please review
@bjorncs @toregge FYI


vekterli (approved):
:+1:

PR #33593, opened 2025-03-17
Revert back to previous semantic of headers with empty value

bjorncs:
Ignore all headers with empty value to preserve the semantic of the previous Jetty 11 integration.

FYI @glebashnik @bjormel


glebashnik (dismissed):
Looks good

evgiz (approved):

PR #33597, opened 2025-03-17
handle overlapping indexed terms better.

arnej27959:
* pick the longest term at a given position;
* any terms within that span are considered as alternative variants (ala stems)

@toregge please review
@bratseth @bjormel FYI


bratseth (dismissed):

geirst (commented):
Please add new unit test(s) in searchsummary/src/tests/docsummary/annotation_converter/annotation_converter_test.cpp.

toregge (commented):
Unit test for annotation converter should be extended.

toregge (approved):
👍

Issue #33316, opened 2025-02-13
Schema field names 'id' and 'version' should be reserved


yuvalshimron:
**Describe the bug**
When I define schemas with the fields "id" and "version" this seems to be accepted by the platform.
However, it generates sources which do not compile because the getters/setters of these fields collide with parent classes.

E.g. in clear() method we get:
`'setId(com.yahoo.document.DocumentId)' in 'com.yahoo.document.Document' cannot be applied to '(java.lang.Long)'`

And in getStructType() we get:
`'getVersion()' in 'com.taboola.vespa.document.Campaign_base.Embedding_meta_data' clashes with 'getVersion()' in 'com.yahoo.document.datatypes.Struct'; attempting to use incompatible return type`

**To Reproduce**
Create a schema with these fields:
```
field id type long {
  indexing: attribute | summary
  rank:filter
}

...

struct embedding_meta_data {
  field end_point_name type string {}
  field version type string {} # <------------ Here
  field update_time type long {}
}
```
And then generate sources.
The sources should show the compilation errors.

**Vespa version**
8.457.32

yuvalshimron:
I also noticed another issue - one of my generated documents has a duplicate field inside the `getDocumentType()` method.
This causes this error:
```
Caused by: java.lang.IllegalArgumentException: Struct already has field field 'publishers_ecvr' of type tensor<float>(publisher_id{})
```
Works fine when I comment out the field but why does this happens? :)

arnej27959:
please provide a schema (preferably as small as possibly) reproducing the publishers_ecvr issue, it's hard to say why that would happen.

arnej27959:
both issues fixed by
https://github.com/vespa-engine/vespa/pull/33601

yuvalshimron:
Thanks so much!

PR #33601, opened 2025-03-17
try to warn/workaround field names clashing with built-in metadata

arnej27959:
WIP - there's still another issue, it seems.

@hmusum please take a look, is this a reasonable (pragmatic) solution?
@bratseth FYI

can any of you think of more field names that should be "reserved"?


bratseth (dismissed):
I can't think of any others ...

hmusum (dismissed):
Yes, I think this is OK

arnej27959:
please take another look - now with attempted fix for the other issue reported in https://github.com/vespa-engine/vespa/issues/33316

hmusum (dismissed):

hmusum (approved):

PR #33611, opened 2025-03-19
Prepare for protobuf descriptor name being a string view.

toregge:
@vekterli : please review


vekterli (approved):
:+1: mental note to self to replace with `std::format` at some opportune time.

PR #33613, opened 2025-03-19
Use `= default` for inline defined virtual destructors, part 2

vekterli:
@toregge please review. Continuation of #33612.


toregge (approved):
:+1:

PR #33614, opened 2025-03-19
Avoid non-public endpoint in Azure test environments

hakonhall:
Currently getting "Azure only supports public endpoints" error when controller tries to deploy to test and staging environments.

### DIFF
--- a/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentSpec.java
+++ b/config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentSpec.java
@@ -241,7 +241,9 @@ public final class DeploymentSpec {
         if (   zone.environment().isTest()
             && instances().stream()
                           .anyMatch(spec -> spec.zoneEndpoints().getOrDefault(cluster, Map.of()).values().stream()
-                                                .anyMatch(endpoint -> ! endpoint.isPublicEndpoint())))
+                                                .anyMatch(endpoint -> ! endpoint.isPublicEndpoint()))
+               // Remove once Azure Private Link has been implemented
+            && !zone.region().value().startsWith("azure-"))
             return ZoneEndpoint.privateEndpoint;
 
         if (zone.environment().isManuallyDeployed())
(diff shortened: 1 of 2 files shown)


config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentSpec.java
```diff
@@ -241,7 +241,9 @@ public ZoneEndpoint zoneEndpoint(InstanceName instance, ZoneId zone, ClusterSpec
         if (   zone.environment().isTest()
             && instances().stream()
                           .anyMatch(spec -> spec.zoneEndpoints().getOrDefault(cluster, Map.of()).values().stream()
-                                                .anyMatch(endpoint -> ! endpoint.isPublicEndpoint())))
+                                                .anyMatch(endpoint -> ! endpoint.isPublicEndpoint()))
+               // Remove once Azure Private Link has been implemented
+            && !zone.region().value().startsWith("azure-"))
```
hakonhall, line 246: This is done to avoid requiring additional method parameters.  Changing the method signature requires preserving the old for backwards compatibility with old models, and adding the new.

config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentSpec.java
```diff
@@ -241,7 +241,9 @@ public ZoneEndpoint zoneEndpoint(InstanceName instance, ZoneId zone, ClusterSpec
         if (   zone.environment().isTest()
             && instances().stream()
```
hakonhall, line 242: Removing this requires changes across model-config-api etc etc. That's going to be a bit risky even with flag. I can try that separately and in a later PR.

freva: What kind of changes are required if this is removed? Updating unit tests?

freva (approved):

config-model-api/src/main/java/com/yahoo/config/application/api/DeploymentSpec.java
hakonhall: I want to guard removing this with a flag.  The flag must either be available in this class but the fan-out seems daunting.  Or, it needs to be passed as a parameter of the zoneEndpoint() method, which crosses the config-model boundary, which means a new method must be added.  Evaluating the flag upstream will then depends on the application being on a new Vespa version. The flag needs to be evaluated in ModelContextImpl. And disabling the flag triggers only on the next redeploy (which should be OK AFAIK).

PR #33631, opened 2025-03-20
Cleanup old `NULL`/literal 0 -> `nullptr`, part 1 of N MERGEOK

vekterli:
@toregge please review. Many more remain, but it's a small start.


toregge (approved):
👍

PR #33634, opened 2025-03-20
Improve naming

bratseth:
Only renaming and removal of an unnecessary interface - no functional changes.


arnej27959 (approved):
seems ok to me

PR #33638, opened 2025-03-21
Bratseth/cycle detection

bratseth:


searchlib/src/main/java/com/yahoo/searchlib/rankingexpression/Reference.java
```diff
@@ -46,6 +49,8 @@ private Reference(String name, Arguments arguments, String output, boolean isIde
         Objects.requireNonNull(name, "name cannot be null");
         Objects.requireNonNull(arguments, "arguments cannot be null");
         this.arguments = arguments;
+        if (featureNames.contains(name))
```
arnej27959, line 52: I think this should use FeatureNames.isSimpleFeature instead.

bratseth: Requires moving FeatureNames out of config-model, but okay, done.

bratseth:
@arnej27959 PTAL

arnej27959 (approved):
👍

PR #33641, opened 2025-03-21
deduce tensor type of elementwise rank feature

arnej27959:
@havardpe please review
@andreer FYI


bratseth (dismissed):
👍 

Longer term we should extract knowledge of this out of the context.

havardpe (approved):

PR #33644, opened 2025-03-21
Prepare for more options in the sort spec for attribute vectors.

geirst:
This change allows selecting the appropriate implementation for writing sort blobs for attribute vectors at setup time instead of per document. Currently, only a default writer is used by calling the IAttributeVector::serializeForXXXSort() API. This API will be removed in the near future.

@toregge please review

### DIFF
--- a/searchlib/src/vespa/searchcommon/attribute/iattributevector.h
+++ b/searchlib/src/vespa/searchcommon/attribute/iattributevector.h
@@ -2,8 +2,9 @@
 
 #pragma once
 
-#include "collectiontype.h"
 #include "basictype.h"
+#include "collectiontype.h"
+#include "default_sort_blob_writer.h"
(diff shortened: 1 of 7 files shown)


searchlib/src/vespa/searchcommon/attribute/iattributevector.h
```diff
@@ -24,6 +25,7 @@ namespace search::attribute {
 
 class IMultiValueAttribute;
 class ISearchContext;
+class ISortBlobWriter;
```
toregge, line 28: Not needed, class is already defined due to including `default_sort_blob_writer.h`.

toregge (approved):

searchlib/src/vespa/searchcommon/attribute/iattributevector.h
geirst: Resolved offline.

PR #33647, opened 2025-03-21
start comparing with median peer

havardpe:
+ protect against bad timings from qrs
+ auto-aligned tables with box drawing characters

@andreer please review


andreer (approved):
👍

PR #33650, opened 2025-03-23
Support cased matching in indexed fields

bratseth:


arnej27959:
while the code changes here look like they should work, the entire feature is likely to cause problems.  For tokenized text it's very possible that the linguistics code will lowercase anyway, maybe always, maybe only when a stem is found.  Getting accent removal matches when you probably wanted "exact" matching is also very surprising.  There should probably be a warning if you use `match: cased` together with `match: text`, and also if you haven't specified `normalizing: none` - to be conservative, I would prefer only allowing this together with `match: exact` actually.

bratseth:
I think there are two potential problems:
- The linguistics library will lowercase anyway (you'll still match, but not cased). OpenNlp does not do this but LuceneLinguistics with default settings do. I think if you plug in your own linguistics implementation you're expected to know what you are doing though.
- Non-cased tokens are not stemmed. This is what OpenNlp does in the cases I've observed. But, arguably that is correct more than 50% of the time.

arnej27959:
Experimented with dutch stemming and OpenNlp: text "School En Kindergarten" gives indexed tokens 
"Schol", "En", "Kindergart", and you'll get match for "School", "Schol", and "Schöl".  So far all the requests I've seen for this feature was when people already had "match: exact" so I would still prefer that to be recommended in doc, at least.

arnej27959 (approved):
👍

PR #33653, opened 2025-03-23
Add chunker

bratseth:
- Add chunker
- Extract invocation context to top level and reuse


bratseth:
Ping @glebashnik

bratseth:
Ping @glebashnik

glebashnik:
On it!

linguistics/src/main/java/com/yahoo/language/process/Chunker.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Splits a text into multiple chunks. The chunks should perferably contain all the content
+     * of the original text, and can be overlapping.
+     *
+     * @param text the text to split into chunks
+     * @param context the context which may influence a chunker's behavior
+     * @return the resulting chunks
+     * @throws IllegalArgumentException if the language is not supported by this
+     */
+    List<String> chunk(String text, Context context);
```
glebashnik, line 41: Consider wrapping strings into chunk object to store chunk locations and perhaps some other metadata about chunk source.
I don't have a specific use case for this in mind right now but it seems it can be useful to know the origin of a chunk, maybe even be able to attach some arbitrary metadata.
Feel free to ignore it doesn't alling with the chunk purpose.

linguistics/src/main/java/com/yahoo/language/process/Chunker.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A chunker converts splits a text string into multiple smaller strings.
+ *
+ * @author bratseth
+ */
+public interface Chunker {
```
glebashnik, line 14: A bit confusing to have both Chunker and Segmenter.
Can we merge them?
If not, make it clear in javadoc what is the conceptual difference between them.

glebashnik (changes requested):

linguistics/src/main/java/com/yahoo/language/process/Chunker.java
bratseth: Thanks. Yes we need both because of your other suggestion, and the need to pass a context.
I suspect we'll want the Language in that context later, and if so we can let Chunker extend and delegate to Segmenter.

glebashnik (approved):
Looks good!

PR #33663, opened 2025-03-25
Do not join in `completeAsync` (OpenAI-client)

thomasht86:

Will continue adding test as discussed.


glebashnik (approved):

bratseth:
Could we find a way to unit test this?

thomasht86:
Yes, we are working on it!

PR #33674, opened 2025-03-26
Add (disabled) RAGSearcherWithOpenAITest.java

thomasht86:

Adding a test of RAGSearcher, using OpenAI-Client that tests that `execution.search(query);` returns before stream is completed. In practice, tests that `completeAsync`-method of OpenAI-client is actually async and non-blocking all the Need to be disabled until we find a way to either mock an OpenAI-compatible server or find a way to actually run one as part of unit tests.

### DIFF
--- a/container-search/src/test/java/ai/vespa/search/llm/RAGSearcherWithOpenAITest.java
+++ b/container-search/src/test/java/ai/vespa/search/llm/RAGSearcherWithOpenAITest.java
@@ -17,6 +17,8 @@ import com.yahoo.search.result.Hit;
 import com.yahoo.search.searchchain.Execution;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
 
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


container-search/src/test/java/ai/vespa/search/llm/RAGSearcherWithOpenAITest.java
```diff
@@ -0,0 +1,115 @@
...
+                "llm.prompt", "{context}\nGiven these documents, answer this query as concisely as possible: @query"
+        );
+        var query = new Query("?" + LLMSearcherTest.toUrlParams(queryParams));
+
+        var result = execution.search(query);
+        var stream = (EventStream) result.hits().get(0);
+        // assert that stream is not complete - (if completeAsync had been blocking, it would be complete)
+        // For this assertion to fail, OpenAI-completion must have been completed in less time than the execution time
+        // of returning the search. Consider this more robust than asserting on timings.
+        assert !stream.incoming().isComplete();
```
glebashnik, line 86: Better to use assertFalse, more consistent with other tests.
```suggestion
       assertFalse(stream.incoming().isComplete()))
```

glebashnik (changes requested):
Looks good, minor change.

glebashnik (approved):

PR #33680, opened 2025-03-26
VESPANG-1389: Add option to stream the HTTP responses

bjormel:


bratseth:
Perhaps we should merge this with default=false for now then, and have somebody test it. I can ask DeviantArt once this is available.

bjormel:
> Perhaps we should merge this with default=false for now then, and have somebody test it. I can ask DeviantArt once this is available.

The default is now set to false

sarzhann (approved):
No objections from my side, but @vekterli or @bratseth have better understandings of safeness of this change 🤷‍♂️

PR #33684, opened 2025-03-27
Add info about effective feed block limit if feed blocked

hmusum:
Suggestions to how we can give some info about effective limits, @vekterli let's discuss offline


vekterli (approved):
:+1:

PR #33685, opened 2025-03-27
Disable verbose logging for malicious requests

bjorncs:

### DIFF
--- a/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyHttpServer.java
+++ b/container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyHttpServer.java
@@ -19,6 +19,7 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.SslConnectionFactory;
 import org.eclipse.jetty.server.handler.ContextHandlerCollection;
+import org.eclipse.jetty.server.handler.ErrorHandler;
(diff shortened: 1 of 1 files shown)


esolitos (approved):

container-core/src/main/java/com/yahoo/jdisc/http/server/jetty/JettyHttpServer.java
```diff
@@ -45,6 +46,14 @@ public class JettyHttpServer extends AbstractResource implements ServerProvider
 
     private final static Logger log = Logger.getLogger(JettyHttpServer.class.getName());
 
+    @SuppressWarnings("LoggerInitializedWithForeignClass")
+    private static final Logger jettyErrorHandlerLog = Logger.getLogger(ErrorHandler.class.getName());
+    static {
+        // Disable warning log entry whenever the rendered message is too large to fit the output buffer.
+        // The response content will be empty in those cases. This is fine as it's typically a result of a malicious request.
+        jettyErrorHandlerLog.setLevel(Level.SEVERE);
```
gjoranv, line 54: Doesn't this silence all non-severe logging from ErrorHandler?  It's not what the comment says. I would expect something more fine-grained, but perhaps that's not possible?

bjorncs: Both are correct. It silences everything, and the handler does not log anything else at info/warn, so effectively it's only silencing that particular message.

PR #33693, opened 2025-03-28
Stop using feature flag SYMMETRIC_PUT_AND_ACTIVATE_REPLICA_SELECTION

hmusum:
Has been set to true for a long time

### DIFF
--- a/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
+++ b/config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
@@ -1550,19 +1550,9 @@ public class ContentClusterTest extends ContentBaseTest {
         assertTrue(resolveDistributorOperationCancellationConfig(2));
     }
 
-    private boolean resolveDistributorSymmetricReplicaSelectionConfig(Boolean flagValue) throws Exception {
-        return resolveDistributorConfig((props) -> {
-            if (flagValue != null) {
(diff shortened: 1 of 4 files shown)


config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
```diff
@@ -1550,19 +1550,9 @@ void distributor_operation_cancelling_config_controlled_by_properties() throws E
...
-            }
-        }).symmetric_put_and_activate_replica_selection();
-    }
-
     @Test
     void distributor_symmetric_replica_selection_config_controlled_by_properties() throws Exception {
-        assertFalse(resolveDistributorSymmetricReplicaSelectionConfig(null)); // defaults to false
-        assertFalse(resolveDistributorSymmetricReplicaSelectionConfig(false));
-        assertTrue(resolveDistributorSymmetricReplicaSelectionConfig(true));
+        assertTrue(resolveDistributorConfig((props) -> {}).symmetric_put_and_activate_replica_selection());
```
vekterli, line 1555: This test should probably either be removed or renamed to imply the config default is always `true`

vekterli (dismissed):
:+1:

config-model/src/test/java/com/yahoo/vespa/model/content/ContentClusterTest.java
hmusum: Yes, thanks, removed

vekterli (approved):

PR #33699, opened 2025-03-28
Magnus/lsp yql comments

Mangern:

YQL comments supported in the parser and highlighted correctly in .yql files. They start with // as this was already defined in the parser.
Also added test for semantic token partitioning and refactored the testing boilerplate to make writing tests simpler.


theodorklauritzen:
This works almost great. However the comments inside the grouping language is not supported when querying. When trying to run queries from with the lsp with comments inside the grouping language, vespa responds with an error.

Mangern:
That makes sense, oops. It is of course not supported in the normal grouping parser. We could strip comments before sending the query. It would not conform with the normal parser, however I think being able to comment inside grouping language makes sense for an application like this.

bratseth:
Maybe we should just add it to the normal parser.

Issue #32406, opened 2024-09-16
Assertion '_features_pos == no_skip.get_features_pos()' fails


lukasmoellerch:
**Describe the bug**
The following assertion fails:

`void search::diskindex::Zc4PostingReaderBase::L1Skip::check(const search::diskindex::Zc4PostingReaderBase::NoSkipBase&, bool, bool): Assertion '_features_pos == no_skip.get_features_pos()' failed.`

**To Reproduce**
I sadly don't have a way to reproduce this issue in an isolated way currently and can also not fully disclose the schema definition that the data was indexed with. We saw this repeatedly happen on this setup with a specific dataset. 

The general timeline was the following: We indexed multiple million documents to the cluster, eventually during a disk index fusion the assertion failure occurred. We then upgraded the cluster to v8.398.4, removed the schema and fed the same data into the cluster, seeing the same assertion failure happen once again after a couple millions documents.

**Expected behavior**
It shouldn't fail

**Screenshots**
-

**Environment (please complete the following information):**
 - OS: Ubuntu-22.04
 - Infrastructure: Azure Virtual Machine with Kubernetes
 - Versions 22.04

(AKSUbuntu-2204gen2containerd-202406.07.0)

**Vespa version**
v8.398.4

**Additional context**

toregge:
If the difference between the two values are a multiple of 4Gi then it might be a 32-bits overflow for the feature position when writing a chunk of the posting list for that word. If you have a very common word that occurs many times in most documents (e.g. having an indexed field that is an array of string or weighted set of string with many long values) such that the average features length for that word is 2 KiB or more (for each document) then the total feature length for that document across 256Ki documents can become larger than 512 MiB (4 Gib).

The [`max-occurrences`](https://docs.vespa.ai/en/reference/schema-reference.html#max-occurrences) setting for the field can be used to limit the number of indexed tokens for the word from each string, but it has limited effect when many strings are indexed for that document field and it will also affect handling of phrases and near operators.

If the difference between the two values is not a multiple of 4Gi then the problem is likely elsewhere, e.g.. something went wrong when reading features from the file, causing the decoding to be out of sync with the encoding, probably due to some unhandled corner case.

lukasmoellerch:
Thanks @toregge for the quick response & the insights.

We got the problem reproduced on the same data again after switching from ext4 to xfs and we have some details that might help with figuring this out: Only one field seems to be effected, a metadata field of type `array<string>` with bm25 and bolding enabled:

[...]

PR #33702, opened 2025-03-30
Use 64-bits bit length when writing feature data for a posting list chunk MERGEOK

toregge:
…hunk.

@vekterli : please review
@geirst : FYI

This is related to issue #32406


vekterli (approved):
:+1:

PR #33706, opened 2025-03-31
Gather memory usage from another source

hmusum:
Avoid expensive lookup of memory usage from `/proc/<pid>/smaps`, using `/proc/<pid>/statm` is cheaper

### DIFF
--- a/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/SystemPoller.java
+++ b/metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/SystemPoller.java
@@ -38,7 +38,6 @@ public class SystemPoller {
     private static final Logger log = Logger.getLogger(SystemPoller.class.getName());
     private static final int memoryTypeVirtual = 0;
     private static final int memoryTypeResident = 1;
-    private static final long pageSize = getPageSize();
     private static final MetricId CPU = MetricId.toMetricId("cpu");
     private static final MetricId CPU_UTIL = MetricId.toMetricId("cpu_util");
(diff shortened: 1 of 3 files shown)


metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/SystemPoller.java
```diff
@@ -305,4 +300,15 @@ public void run() {
             poller.poll();
         }
     }
+
+    private static int getPageSize() {
+        try {
+            return Integer.parseInt(new ProcessExecuter().exec("getconfig PAGESIZE").getSecond().trim());
```
boeker, line 306: Isn't the utility called `getconf` (instead of `getconfig`)?

boeker (changes requested):
See comment

boeker (dismissed):
`getconf` fixed 👍

hmusum:
@boeker please take a look again

metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/SystemPoller.java
```diff
@@ -107,41 +111,35 @@ void stop() {
      * @return array[0] = memoryResident, array[1] = memoryVirtual (kB units)
```
boeker, line 111: The javadoc says `kB units`.

boeker (commented):
The javadoc of `getMemoryUsage` says that the returned values are in kB. However, the returned values are in bytes, and if I am not mistaken, the returned values were in bytes already before these changes, so I guess the javadoc already was wrong before?!

boeker (approved):
👍

PR #33711, opened 2025-03-31
chore: Ensure safe variable usage in upload script

esolitos:
## What

* Added `-S` on curl
* Iprove handling of vars

## Why

Improve debug on release errors

### DIFF
--- a/.github/scripts/upload-rpm-to-cloudsmith.sh
+++ b/.github/scripts/upload-rpm-to-cloudsmith.sh
@@ -1,14 +1,17 @@
 #!/bin/bash
 # Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
-set -euo pipefail
+set -o xtrace
+set -o pipefail
+set -o nounset
+set -o errexit
 
 if (( $# < 1 )); then
   echo "Usage: $0 <RPM file>"
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


glebashnik (dismissed):
Good improvements.

.github/scripts/upload-rpm-to-cloudsmith.sh
```diff
@@ -8,7 +8,7 @@ if (( $# < 1 )); then
   exit 1
 fi
 
-if [[ -z $CLOUDSMITH_API_TOKEN ]]; then
+if [[ -z "${CLOUDSMITH_API_TOKEN:-}" ]]; then
```
arnej27959, line 11: I don't understand the rationale for this change.  The `:-` part does literally nothing; the braces are nice for readability, but not needed; I would prefer to not use double brackets (they are bash only) but if you use them the quotes aren't needed either.  Consider `if [ -z "${CLOUDSMITH_API_TOKEN}" ];` or even
`if [ "${CLOUDSMITH_API_TOKEN} = "" ];`

esolitos: Wouldn't  `if [ "${CLOUDSMITH_API_TOKEN} = "" ];` fail if the variable is not defined because of the `set -u` ?

arnej27959 (approved):
I hadn't used the "set -u" aka "set -o nounset" flag before, so I had to check exactly how that worked...

PR #33714, opened 2025-04-01
Simplify test case and allow failure on idle timeout

bjorncs:
The client may under certain conditions block until an idle timeout is reached.


vekterli (approved):
:+1:

PR #33722, opened 2025-04-02
Add support for IN operator in json select queries

boeker:
Previously, the in operator could be used in yql queries but not in select queries.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/SelectParser.java
+++ b/container-search/src/main/java/com/yahoo/search/query/SelectParser.java
@@ -3,6 +3,8 @@ package com.yahoo.search.query;
 
 import com.google.common.base.Preconditions;
 import com.yahoo.prelude.query.FalseItem;
+import com.yahoo.prelude.query.NumericInItem;
+import com.yahoo.prelude.query.StringInItem;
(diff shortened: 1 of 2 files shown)


container-search/src/main/java/com/yahoo/search/query/SelectParser.java
```diff
@@ -625,6 +630,56 @@ private Item buildEquals(String key, Inspector value) {
...
+            throw new IllegalArgumentException("The in operator is only supported for integer and string fields. The field '" +
+                    field + "' is not of these types");
+        }
+        if (index.isInteger() && index.isString()) {
+            throw new IllegalArgumentException("The in operator is not supported for fieldsets with a mix of integer and string fields. The fieldset '" +
+                    field + "' has both");
+        }
+
+        Collection<Inspector> values = children.values();
+        Item item = null;
```
hmusum, line 652: Unused, remove?

hmusum (dismissed):
Great!

container-search/src/main/java/com/yahoo/search/query/SelectParser.java
boeker: You're right!

hmusum (approved):

PR #33738, opened 2025-04-05
Avoid problematic definition of internal close member function.

toregge:
@vekterli : please review


vekterli (approved):
:+1: interesting how the compiler didn't scream about this from day one...

PR #33763, opened 2025-04-10
Remove unused feature flag

hmusum:
I assume we have built all machine images after 2025-02-13?


hakonhall (approved):

hakonhall:
They are built, but not promoted everywhere yet which is the requirement.  Once it has been promoted I can remove some code from InstallYumPackage.  InstallYumPackage has the same TODO, so we can remove this flag.

Issue #33755, opened 2025-04-09
Query dispatch fails with mixed group states (maintenance + diskLimitReached)


juhkin:
**Describe the bug**

Vespa dispatcher rejects queries in grouped distribution setup if one group is in maintenance and other group has intermittent feedblocks.

When running long operations that involve reprocessing a large number of documents, the expected approach is to place one content group into maintenance mode while the other continues to serve queries even if that active group becomes feed blocked. However, Vespa appears to incorrectly consider the maintenance-mode group as still needing to be accounted for when dispatching queries. This leads to query rejection instead of falling back to the remaining (feed blocked but otherwise available) group.

**To Reproduce**
1. Create grouped distribution with 2 content groups
2. Add one group to maintenance mode (we had most issue during controlled shutdown and reprocessing) 
```
[2025-04-09 08:51:48.528] WARNING container        Container.com.yahoo.search.dispatch.searchcluster.SearchCluster	Cluster dispatcher.iprally: group 1 has reduced coverage: Active documents: 0/116124981, Target active documents: 0, working nodes: 0/24
```
3. One group to feed blocked state

![Image](https://github.com/user-attachments/assets/4150b4da-b92b-4843-89c6-903bc33dac2c)

4. Dispatcher rejects queries due to `java.lang.IllegalStateException: No suitable groups to dispatch query. Rejected: [0, 1]\n\tat com.yahoo.search.dispatch.Dispatcher.getInternalInvoker(Dispatcher.java:348)`

**Expected behavior**
The Vespa dispatcher should fallback to dispatching queries to the feed blocked group, similar to how it behaves when there's only one content group. The presence of feed blocks should not prevent query dispatch as long as a group is technically available to serve queries.

**Screenshots**

![Image](https://github.com/user-attachments/assets/8061cb15-1213-4dac-bdd8-7a5977673283)

![Image](https://github.com/user-attachments/assets/11aefa50-5ffe-47a9-931b-cd927bbafa4d)

**Environment:**
 - Infrastructure: Google Kubernetes Engine 
 - Versions 1.30.8-gke.1261000

**Vespa version**
8.460.16

vekterli:
It is not the intention that query dispatching should be affected by feed block. Do you have any log messages from the dispatcher regarding the status of group 0 (i.e. the feed blocked group) prior to observing the exception?

juhkin:
[...]

PR #33787, opened 2025-04-14
Don't reject feed blocked group if it has enough active documents

hmusum:
Prefer a feed blocked group before one without coverage (i.e. one that is in maintenance).

https://github.com/vespa-engine/vespa/issues/33755


bratseth:
There's always another corner case inside the corner case with this :-)

Would it be easier to understand this if we instead made the condition for rejecting a group 
(is blocking feed && has fewer docs than the max observed)?

hmusum:
Yes, makes sense, I'll look into it

hmusum:
@bratseth please take a look

bratseth (approved):

PR #33793, opened 2025-04-14
Havardpe/new profiling tags

havardpe:
@toregge please review


toregge (approved):

toregge (changes requested):
searchlib_queryeval_profiled_iterator_test_app unit tests fails on x86_64 platforms. The order of calling `OR()` and `blend()` is not defined in `create_iterator_tree()` , thus the enumeration order is undefined.

havardpe:
@toregge PTAL

toregge (approved):

PR #33807, opened 2025-04-16
Add missing includes. Avoid triggering bogus gcc 15 warning.

toregge:
@boeker : please review


boeker (approved):
That warning is really strange. 👍

PR #33808, opened 2025-04-16
The imported target GTest::GTest has been deprecated since CMake 3.20.

toregge:
Use the imported target GTest::gtest instead (added in CMake 3.20). Require CMake 3.20 or newer.

@boeker : please review


boeker (approved):
👍

PR #33826, opened 2025-04-22
Classify clear as a complete field replacement

bratseth:
This lets 'clear' updates be handled when the
indexing statement for that field depends on multiple input fields.


arnej27959 (approved):
seems ok to me

PR #33827, opened 2025-04-22
Stop considering feed blocked status per group in dispatcher

hmusum:
Since feed blocked is for the whole cluster it does not make sense to consider individual groups and their feed blocked status, simplify accordingly.


bratseth (approved):
nice

PR #33828, opened 2025-04-23
Preserve published method

bratseth:


container-search/src/main/java/com/yahoo/search/grouping/request/GroupingOperation.java
```diff
@@ -50,6 +50,28 @@ protected GroupingOperation(String image, String label) {
         super(image, label);
     }
 
+    /** @deprecated pass a FilterExpression (which can be null) */
+    @Deprecated // TODO: Remove on Vespa 9
+    protected GroupingOperation(GroupingOperation parentOfCopy,
```
bjorncs, line 55: Symbols with protected access modifier is only visible for sub-classes and same package, and therefore not public API in this scope. Is inheriting from `GroupingOperation` something that application code will actually do?

bratseth: At least one have.

bjorncs (approved):

PR #33829, opened 2025-04-23
newer jllama versions bundle llama.cpp into libjllama.so

arnej27959:
@glebashnik please review


glebashnik:
On it, will run disabled local llm tests to see that everything works.

glebashnik (approved):

PR #33831, opened 2025-04-23
Expose stable cluster document and byte count metrics from cluster controller

vekterli:
@hakonhall please review cluster controller changes in particular (756d7e57b09d4896580503deea2ddcf4d81a07ea)
@toregge please review distributor changes in particular (32604189e26012087400847638d9c1f4106f88f8)

Report back aggregated cross-node and cross-bucket space document counts and byte sizes from the distributors to the cluster controller. By having aggregation that is cluster state version/transition-aware, the controller can report consistent values for total document counts and byte sizes.

This is unlike distributed gathering of distributor metrics, which risks sampling from different nodes on different sides of a state transition, causing seemingly inconsistent values to be observed (becoming consistent again once all sampling takes place on the same side of the transition edge).

The new host info values reported are the same as those emitted by the distributor via the existing docs/bytes stored-metrics. They are communicated to the controller using the existing host info `distributor` object, adding a new `global-stats` object underneath.

Two new metrics are added on the cluster controller:
  - `cluster-controller.stored-document-count`
  - `cluster-controller.stored-document-bytes`

Also unify `MetricUpdater` default metric context to always include certain dimensions.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
@@ -573,10 +573,13 @@ public class FleetController implements NodeListener, SlobrokListener, SystemSta
     }
 
     private void updateMasterClusterSyncMetrics() {
-        var stats = stateVersionTracker.getAggregatedClusterStats().getAggregatedStats();
-        if (stats.hasUpdatesFromAllDistributors()) {
-            GlobalBucketSyncStatsCalculator.clusterBucketsOutOfSyncRatio(stats.getGlobalStats())
+        var stats = stateVersionTracker.getAggregatedClusterStats();
+        var aggrStats = stats.getAggregatedStats();
+        if (aggrStats.hasUpdatesFromAllDistributors()) {
+            GlobalBucketSyncStatsCalculator.clusterBucketsOutOfSyncRatio(aggrStats.getGlobalStats())
                     .ifPresent(metricUpdater::updateClusterBucketsOutOfSyncRatio);
+            metricUpdater.updateClusterDocumentMetrics(
+                    stats.getAggregatedDocumentCountTotal(), stats.getAggregatedBytesTotal());
         }
     }
 
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/DistributorGlobalStats.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/DistributorGlobalStats.java
@@ -1,0 +1,18 @@
+package com.yahoo.vespa.clustercontroller.core.hostinfo;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Cross-node/bucket-space/replica statistics from a particular distributor.
+ *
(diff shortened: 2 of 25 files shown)


toregge (approved):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
```diff
@@ -573,10 +573,13 @@ public void tick() throws Exception {
     }
 
     private void updateMasterClusterSyncMetrics() {
-        var stats = stateVersionTracker.getAggregatedClusterStats().getAggregatedStats();
-        if (stats.hasUpdatesFromAllDistributors()) {
-            GlobalBucketSyncStatsCalculator.clusterBucketsOutOfSyncRatio(stats.getGlobalStats())
+        var stats = stateVersionTracker.getAggregatedClusterStats();
+        var aggrStats = stats.getAggregatedStats();
+        if (aggrStats.hasUpdatesFromAllDistributors()) {
```
hakonhall, line 578: If we go to the `else`-part for an extended period of time, the stored-document-count metrics may appear to be stable even though they are stale and not trustworthy.  What's the best way to capture this?  Have another metric on the ratio of distributors that we have updates from, and... alert on it if it stays <1.0 for too long?  Or always show such a graph next to stored-document-count metrics?  Set the metric to 0 if the `if` has not been entered within the last X minutes?

hakonhall (approved):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/FleetController.java
vekterli: We have a dedicated, time-sensitive `cluster-controller.nodes-not-converged` metric that is intended to capture this scenario, and that can generate actionable alerts if > 0. I therefore think it should be OK to not have any explicit failure detection as part of these new metrics.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/DistributorGlobalStats.java
```diff
@@ -0,0 +1,18 @@
...
+ * Cross-node/bucket-space/replica statistics from a particular distributor.
+ *
+ * Fields are nullable to be able to distinguish between a node being on a version that
+ * does not send the fields vs. a node that sends the fields with zero values.
+ *
+ * @param storedDocumentCount Number of documents stored across all bucket spaces on all
+ *                            content node replicas controlled by the distributor.
+ * @param storedDocumentBytes Combined byte size of the documents reported by storedDocumentCount.
+ */
+public record DistributorGlobalStats(@JsonProperty("stored-document-count") Long storedDocumentCount,
```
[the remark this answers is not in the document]
vekterli: The host info Jackson `ObjectMapper` is explicitly configured with
```java
mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
I just double-checked this since I realized none of the other existing host info classes had `@JsonIgnoreProperties` set, but still happily accepted unknown fields in the JSON... 👀

It makes sense to have type-level annotations instead, but I think I will do this as a separate change.

PR #33837, opened 2025-04-23
Add test source file missing from `CMakeLists.txt`

vekterli:
@toregge please review.

Seemingly slipped between the cracks when the test was moved here from another module at some point. Luckily the unit tests have not been silently broken...


toregge (approved):
👍

PR #33842, opened 2025-04-24
Implement /state/v1/version in vespalib

boeker:
Previously, only some services supported /state/v1/version. This commit adds support for it to the remaining services.


vekterli (approved):
:+1:

PR #33855, opened 2025-04-25
Fix inconsistencies w.r.t. toString() in Version in vespalib

boeker:
Previously, the class Version only had a toString() method. This PR renames this method to toFullString() and implements a toString() method to mimic the behavior of the corresponding Java class. The behavior of toFullString() is slightly changed to avoid inconsistencies and follow the behavior of the corresponding Java class (and the doc string).

### DIFF
--- a/vespalib/src/tests/component/component.cpp
+++ b/vespalib/src/tests/component/component.cpp
@@ -157,7 +157,8 @@ TEST("testEmpty")
     Version ev;
     VersionSpecification evs;
 
-    EXPECT_EQUAL("", ev.toFullString());
+    EXPECT_EQUAL("", ev.toString());
+    EXPECT_EQUAL("0.0.0", ev.toFullString());
     EXPECT_EQUAL("*.*.*", evs.toFullString());
 
     EXPECT_TRUE(ev == Version(0,0,0,""));
--- a/vespalib/src/tests/component/component.cpp
+++ b/vespalib/src/tests/component/component.cpp
@@ -198,6 +198,79 @@ TEST("testSimple")
     EXPECT_TRUE(vs.matches(v));
 }
 
+TEST("testZeros")
+{
+    // test Version:
+
+    Version v000(0, 0, 0, "");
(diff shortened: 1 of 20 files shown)


vespalib/src/tests/component/component.cpp
```diff
@@ -197,6 +198,79 @@ TEST("testSimple")
...
+    Version v110s("1.1");
+    EXPECT_EQUAL("1.1", v110s.toAbbreviatedString());
+    EXPECT_EQUAL("1.1.0", v110.toString());
+
+    Version v111(1, 1, 1, "");
+    EXPECT_EQUAL("1.1.1", v111.toAbbreviatedString());
+    EXPECT_EQUAL("1.1.1", v111.toString());
+    Version v111s("1.1.1");
+    EXPECT_EQUAL("1.1.1", v111s.toAbbreviatedString());
+    EXPECT_EQUAL("1.1.1", v111s.toString());
```
vekterli, line 231: It may border on paranoia to do so, but it could be useful to specify distinct numbers for the major/minor/micro parts of the version string (e.g. `1.2.3`) to catch any unintended errors caused by e.g. reusing one parsed component across multiple fields.

vespalib/src/tests/component/component.cpp
```diff
@@ -197,6 +198,79 @@ TEST("testSimple")
     EXPECT_TRUE(vs.matches(v));
 }
 
+TEST("testZeros")
```
vekterli, line 201: Consider increasing the verbosity of the test name so that it states _what_ property of zeroes it expects to hold. E.g. something like `"Version toAbbreviatedString() truncates trailing zeroed components while toString() does not"`.

Also consider factoring out `VersionSpecification` specific parts to its own test case. This also means the `// test Foo:` comments can be removed without any information loss. Other than that, great test coverage!

vekterli (dismissed):
:+1: only a couple minor comments

vekterli (approved):
:+1:

PR #33858, opened 2025-04-25
Bratseth/clear is complete take 2

bratseth:


bratseth:
@arnej27959 please review

This is now just 
- improving the error message when sending an update for a field which has an indexing statement depending on multiple inputs.
- adding tests that demonstrates the current behavior.
- cleanup.

arnej27959 (approved):
seems ok to me

PR #33864, opened 2025-04-25
experiment with json to nested table conversion

havardpe:
@andreer please review


andreer:
Koan: What is the width of a non-printable character?

andreer (approved):

PR #33869, opened 2025-04-26
Bratseth/embed expression

bratseth:


linguistics-components/src/test/java/ai/vespa/language/chunker/SentenceChunkerTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author bratseth
+ */
+public class SentenceChunkerTest {
+
+    @Test
+    public void testChunker() {
```
glebashnik, line 18: Should include test sentences with abbreviations with period, e.g. a.m. Mrs. etc.

linguistics-components/src/main/java/ai/vespa/language/chunker/SentenceChunker.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author bratseth
+ */
+public class SentenceChunker implements Chunker {
+
+    @Override
+    public boolean isReversible() { return true; }
+
+    @Override
+    public List<Chunk> chunk(String inputText, Context context) {
```
glebashnik, line 22: This will return false positives for abbreviations that end in lowercase letters (a.m. or Mrs. or etc.).
More sophisticated methods are usually used to split sentence, e.g. see OpenNLP implementation of sentence detection.
This is often language-specific.

glebashnik (changes requested):
Looks good. Nice refactorings.
We should probably improve `SentenceChunker`, see the comment below.
Also consider how to deal with multiple-languages, e.g. language-specific chunker vs multilingual chunker.
Should `Chunker` distinguish between languages, e.g. include language in `Context`?

glebashnik:
Branch and PR name "embed-expression" should be "chunk-expression".

linguistics-components/src/main/java/ai/vespa/language/chunker/SentenceChunker.java
bratseth: Yes, a lot of possible improvements like that, but I think open-ended. Hopefully we can use some library for that. Here I just wanted something simple without dependencies.

integration/schema-language-server/language-server/src/main/ccc/indexinglanguage/IndexingParser.ccc
```diff
@@ -461,8 +488,7 @@ Expression embedExp() :
       ( SCAN((identifierStr)+) => (embedderArgument = identifierStr()) { embedderArguments.add(embedderArgument); } )*
     )
     {
-        // return new EmbedExpression(embedders, embedderId, embedderArguments);
-        return null;
```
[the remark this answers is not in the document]
bratseth: It was missing a parameter so I suppose somebody did it to make it compile ...

bratseth:
Language is already in the Context parent class.

PR #33890, opened 2025-04-28
refactor Value children

arnej27959:
add some safety bits when using size from the union
add helpers to enable modern C++ iteration

please review carefully


toregge (dismissed):
@vekterli : FYI

toregge (approved):

PR #33902, opened 2025-04-29
ordering expressions seemed to be double-executed

arnej27959:


havardpe (commented):
but why is the other execute better than this one...

PR #33903, opened 2025-04-29
Fix typo in vectorized acceleration type names

vekterli:
@toregge please review


toregge (changes requested):
`/home/tegge/git/vespa/vespalib/src/vespa/vespalib/hwaccelerated/iaccelerated.cpp:27:48: error: template argument 1 is invalid
   27 |         return std::make_unique<Avx2Accelrator>();
`

Incomplete rename.

vekterli:
I see there are some other usages that escaped my casual grepping as well... Will fix.

toregge (approved):

PR #33905, opened 2025-04-30
Change owner of routing-generator module and some minor cleanup

hmusum:


arnej27959 (dismissed):
it looks like there will be conflicts between these PRs changing the CODEOWNERS file, but we'll see...

hmusum:
Yes, thanks, should be fixed with the latest change

arnej27959 (approved):
👍

PR #33906, opened 2025-04-30
fix certificates config regression

radu-gheorghe:
With the detect-schema changes, I broke the mTLS certificate options (unless defaults were used). Now it should be OK (including unit tests).


kkraune (dismissed):
👍 Thanks

not sure about the CODEOWNERS commit

kkraune (approved):

PR #33912, opened 2025-04-30
Remove searchlib true unit test.

toregge:
@vekterli : please review


vekterli (approved):
:+1: at the very least, this was a very stable unit test

PR #33913, opened 2025-04-30
Fetch element gap from index environment for native proximity and

toregge:
term distance rank features.

@vekterli : please review


vekterli (approved):
:+1:

PR #33914, opened 2025-04-30
Only disable if in Azure and only mTLS endpoint is enabled

bjorncs:


config-model/src/main/java/com/yahoo/vespa/model/container/xml/ContainerModelBuilder.java
```diff
@@ -25,6 +25,7 @@
 import com.yahoo.config.provision.AthenzDomain;
 import com.yahoo.config.provision.AthenzService;
 import com.yahoo.config.provision.Capacity;
+import com.yahoo.config.provision.CloudName;
```
hakonhall, line 28:
```suggestion
```

hakonhall (dismissed):

ldalves (approved):

PR #33932, opened 2025-05-02
Extend iterator constness when iterating over const children.

toregge:
@arnej27959 : please review
@vekterli : FYI

Note: AppleClang doesn't yet support cbegin()/cend() member functions on std::span, but using begin()/end() should be ok when template argument to std::span is a const type.


arnej27959 (approved):
👍

PR #33939, opened 2025-05-03
Add fixed-length chunker

bratseth:


docprocs/src/main/java/com/yahoo/docprocs/indexing/IndexingProcessor.java
```diff
@@ -71,7 +70,7 @@ public IndexingProcessor(DocumentTypeManager documentTypeManager,
              new ScriptManager(documentTypeManager,
                                ilscriptsConfig,
                                linguistics,
-                               toMap(chunkers, DefaultChunkerProvider.class),
+                               toMap(chunkers, null),
```
glebashnik, line 73: Why chunkers are handled differently than embedders, generators?

glebashnik (dismissed):

docprocs/src/main/java/com/yahoo/docprocs/indexing/IndexingProcessor.java
bratseth: Embedding and generation requires configuring an instance in services.xml.

I think some useful chunking strategies can be done purely in Java code without configuration. If so it makes sense to provide a selection of chunkers by default rather than requiring component setup in services.xml.

bratseth:
@glebashnik please take a look again

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A chunker which splits a text into chunks at the first word break after a given target chunk length
+ * (or precisely at that length, for CJK languages).
+ *
+ * @author bratseth
+ */
+public class FixedLengthChunker implements Chunker {
```
glebashnik, line 16: Consider including length units (characters) in the class name and comment, e.g FixedCharacterLengthChunker.
It is common to split text by fixed number of tokens so we might want to implement FixedTokenLengthChunker using a tokenizer.

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private List<Chunk> computeChunks(String inputText, int chunkLength, boolean isCjk) {
+        var text = new UnicodeString(inputText);
+        List<Chunk> chunks = new ArrayList<>();
+        var currentChunk = new StringBuilder();
+        int currentLength = 0;
+        for (int i = 0; i < text.length();) {
+            int currentChar = text.codePointAt(i);
+            currentChunk.appendCodePoint(currentChar);
+            if (++currentLength >= chunkLength && (isCjk || !characters.isLetterOrDigit(currentChar))) {
```
glebashnik, line 38: Consider explaining limitations of this approach in comments, e.g. it will split floats 12.4.

glebashnik (changes requested):

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
bratseth: I think we won't since it is harder to fit that into the indexing pipeline, and (imho) it doesn't matter unless the length is much shorter than what we've heard people use so far (since the avg. difference between character and chunk count will be near constant).

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
bratseth: Improved doc.

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
bratseth: Improved doc.

PR #33958, opened 2025-05-06
Add system name parameter to factory-command create-release

gjoranv:
- To allow calling it from the promote-to-pubilc gh workflow (uses 'public', while buildkite uses 'opensource')


esolitos (changes requested):
You might want to change also the "update-released-time" entry 😉

esolitos (approved):
PTAL after my changes @gjoranv

PR #33965, opened 2025-05-07
Extend flag expiry dates for 2 flags

hmusum:


esolitos (approved):

esolitos:
BTW: Those two flags feel they should be permanent, or am I wrong?

ean:
> BTW: Those two flags feel they should be permanent, or am I wrong?

they should not be flags, but instead part of the soon to be implemented(tm) tenant policy. I mostly kept them as temporary to nag me every 3 months that we need to do this properly.

PR #33968, opened 2025-05-08
Give more info when warning about searchable copies > 1 and hnsw index

hmusum:
Log schema name and fields and change to INFO level.


arnej27959 (approved):
👍

PR #33969, opened 2025-05-08
Update customer RPM service model

evgiz:


hakonhall (dismissed):
It would be nice if this class could follow the style used by the other custom flags in this directory.

hakonhall:
- Add either @JsonIgnore or @JsonProperty on public methods
- Avoid @JsonProperty on private fields (I didn't think this worked!?)
- Avoid relying on Jackson error handling as much as possible, and instead verify explicitly in our code.  In particular do not require any fields (Use Double instead of double, no `require = true`). So the only way for deserialization to fail is on bad JSON.  But you may end up with only nulls.  For some fields, an absent field can be taken as equivalent to a default value, and that's fine to fix in the constructor.

evgiz:
`@JsonIgnore` is redundant when we have the class decorator `@JsonIgnoreProperties(ignoreUnknown = true)`. Pretty sure jackson uses some reflection magic internally, so access modifiers like `private` have no effect. I've removed the`required=true` annotation 👍

hakonhall:
`@JsonIgnoreProperties(ignoreUnknown = true)` is to ignore any unknown fields when deserializing, whereas `@JsonIgnore` on public getter-methods (w/o arguments) will make sure Jackson don't interpret the method as a property and render a JSON field for it.  There should be no '@JsonIgnore on setter-methods` as the constructor is used for deserialization.

hakonhall:
:thinking: , not entirely sure how annotations would work for a record.  I think it would just work to use @JsonProperty on the constructor arguments.

evgiz:
Both serialization & deserialization works as expected including with missing/extra fields, so in that case the default behavior is sufficient. Both `Role` and `RoleList` follow the same style as this class. If jackson supports records easily we could of course do that, but all other custom flags are class-based.

hakonhall:
Migrating to record is just to make the class more compact and simpler to read.  It's fine not to do it - often a class needs a method or similar that prevents it from being a record. Most of the other classes in this package was made before record was available in Java, but they should probably also be migrated at some point.

Jackson is very flexible, so there are many ways to (de)serialize classes that makes it non-obvious what's going on.  I believe I started annotating with @JsonIgnore after I found out that Jackson interprets public getter methods as-if they are annotated with @JsonProperty.  By using @JsonIgnore we disable Jackson smartness which can be surprising and confusing to developers not familiar with it.  By default, most things work as you expect, but I find it better to be explicit to both aid the developer, and to shut down Jackson smartness.

But don't feel you *have* to use @JsonProperty - it's something I use especially around flag serialization, but it's less common elsewhere.

hakonhall (approved):

PR #33970, opened 2025-05-08
enable specifying element filtering separate from command

havardpe:
@toregge @arnej27959 please review


toregge (changes requested):
Need to update various summary.cfg files too.

toregge (approved):

PR #33971, opened 2025-05-08
Sync fill_matching_elements() callback signature with

toregge:
fillSummaryFeatures() callback signature.

@havardpe : please review


toregge:
Will keep old signature on callback.

PR #33973, opened 2025-05-08
rewrite partial fill logic with better semantics

arnej27959:
Before this, doing fill(result, "foo") could result in only getting parts of the "foo" document-summary if you also had "select a,b,c ..." in your YQL statement.

Now, only fill(result, "default") or equivalent will trigger the special only-some-field fetching.  Also, the hits will be marked with hit.setFilled("[f:a,b,c]") so that it doesn't lie about what we actually have.

We probably want to redefine some semantics so it's possible to see the difference between at least some of these:
* execution.fill(result)
* execution.fill(result, null)
* execution.fill(result, result.getQuery().getPresentation().getSummary())
* execution.fill(result, "mysummary")

NOTE: no unit tests yet, some semantics are still open for change

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/DocsumDefinition.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/DocsumDefinition.java
@@ -5,6 +5,7 @@ import com.yahoo.data.access.Inspector;
 import com.yahoo.search.schema.DocumentSummary;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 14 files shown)


bjorncs (approved):

dainiusjocas:
Would this rewrite work such that "select matchfeatures from...." would not fetch the summary at all?

dainiusjocas:
> No, I think it would still do a summary fetch to get matchfeatures, but that's a good point - it should be possible to avoid that, and this change puts most of the necessary pieces in place.

Should I create a GH issue for such a feature? Or it is going to be implemented anyway?

arnej27959:
github issue would be nice, thank you

dainiusjocas:
> github issue would be nice, thank you

And here it is https://github.com/vespa-engine/vespa/issues/33979 🙏

bratseth (approved):
Great!

PR #33976, opened 2025-05-09
Theodorkl/Lsp filter and chunk support

theodorklauritzen:
Add support for chunk in indexing language, and filter in grouping language.

### DIFF
--- a/integration/schema-language-server/language-server/src/main/ccc/indexinglanguage/IndexingParser.ccc
+++ b/integration/schema-language-server/language-server/src/main/ccc/indexinglanguage/IndexingParser.ccc
@@ -456,11 +456,12 @@ Expression chunkExp() :
 }
 
     (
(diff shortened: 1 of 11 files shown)


Mangern (dismissed):
Clean change

Mangern (approved):
sweet

bratseth:
Why are some return values set to null but not others?

theodorklauritzen:
Some of the return values are set to null because the environment around the parser is not properly configured. This was initially done since some of the constructors and functions validated the data and could throw errors. This would stop the parsing of the rest of the document. Other places it referenced variables that were null, and that would throw a nullpointer-exception.

We can review the parser to try to resolve these issues, although we haven't prioritized it.

bratseth:
Thanks, got it.

bratseth:
This should be solved by https://github.com/vespa-engine/vespa/pull/34294

It would be nice with a unit test that tested this on the lsp side though.

PR #33983, opened 2025-05-09
Update elements setting for matched-elements-only in summary config.

toregge:
@havardpe : please review
@arnej27959 : FYI


config-model/src/main/java/com/yahoo/schema/derived/SummaryClass.java
```diff
@@ -165,6 +168,17 @@ static String getSource(SummaryField summaryField, Schema schema) {
         }
     }
 
+    static SummaryConfig.Classes.Fields.Elements convertElementsSelector(SummaryElementsSelector elementsSelector) {
```
havardpe, line 171: consider returning the builder here

havardpe (dismissed):

config-model/src/main/java/com/yahoo/schema/derived/SummaryClass.java
toregge: Done.

havardpe (approved):

PR #33988, opened 2025-05-09
Remove unused, deprecated method to get zone endpoint

hmusum:
Only tests was using this and no config models in use are using it anymore. Part 1, more to come.


hakonhall (approved):
great

PR #33990, opened 2025-05-09
remove confusing script reordering

arnej27959:
There's some old reordering magic for indexing scripts that I do not understand, but its effect is basically to bunch all expressions doing set_language before the expressions that perform indexing.  This makes it really hard to use multiple set_language calls to switch languages.
Do we have any idea if this is (still) necessary?

This PR will keep order for fields in the document, as far as I can figure out.  Maybe we need to just try it and see.


bratseth:
Wat ... never heard of that and it explicitly breaks what we say in the documentation.

bratseth (approved):

PR #33993, opened 2025-05-10
Bratseth/summary fields cleanup

bratseth:
Cleanup + better error messages. I suggest looking at the commits in order.


toregge (approved):
👍

PR #33994, opened 2025-05-10
Try harder to keep original order of indexing expressions.

arnej27959:
Expressions using an input that is modified by some other expression will be moved so they get the original input. Fix allExtraFields() method to avoid extra reordering.

Another try on this keeping some reordering...
@hmusum FYI


bratseth (approved):
Looks good to me, given the premise. Seems like we should be able to stop doing this at all since document fields content cannot be modified, we only add annotations.

arnej27959:
yes, annotations seems to be the only problem.  I'm guessing that if we allow "tokenize" and "ngram" to clear annotations from their input, we could remove this reordering also.

PR #33995, opened 2025-05-11
Cache generations

bratseth:


glebashnik:
@bratseth
Could you clarify how long does this cache lives?
My understanding is the entry point where the context is created in the expression itself, thus lives as long as expression is executed.
In this case, what is the use case for this cache?

bratseth:
It's cached for the execution of the entire "indexing script" for one document operation.
The indexing script is the collection of statements created from a schema.
This is a bit hard to see because the indexing script is itself an indexing expression (ScriptExpression), and the logic connecting execution to a document update is in DocumentScript in the docprocs module (the latter I would like to improve).

So, caching lets us reuse intermediate results of expensive indexing statements in a single document, such as using the same generation both for indexing and embedding.

glebashnik:
I see. Thanks for the explanation.

glebashnik (approved):

PR #33998, opened 2025-05-12
Fix calculation of rss

hmusum:


metrics-proxy/src/main/java/ai/vespa/metricsproxy/service/SystemPoller.java
```diff
@@ -135,11 +135,14 @@ static long[] getMemoryUsage(VespaService service, int pageSize) {
     static long[] getMemoryUsage(String s, int pageSize) throws IOException{
         long[] size = new long[2];
         // statm line: "size rss shared text lib data dt"
+        // Note: From man proc_pid_statm man page: rss is the same as VmRSS in /proc/pid/status
+        // Note 2: From man proc_pid_status:  VmRSS  Resident set size.  Note that the value here is the sum of RssAnon, RssFile, and  RssShmem.
+        // So we need to subtract the shared memory from the resident set size to get what we want
```
vekterli, line 140: Could be useful to have a "Note 3" which states that `shared` is `RssFile+RssShmem` so that it's more obvious that we're then getting `RssAnon` from the subtraction.

vekterli (dismissed):
:+1:

vekterli (approved):

PR #34006, opened 2025-05-13
configurable language detectConfidenceThreshold

arnej27959:
fixing an old ticket - @kkraune does it seem reasonable to ask users to configure a number between 1 and 10?

@bratseth please review code and test changes.

There's another change mixed in here, since I noticed that the OpenNLP detector will return 3-letter language codes that we didn't pick up, so I added some mappings for IANA language subtag registry to ensure that we pick up more (at least "cmn" for Mandarin Chinese).


bratseth (approved):
👍

PR #34011, opened 2025-05-14
Add app, instance and arch dims for customer rpm flag

evgiz:
This setup assumes that missing/empty dimension value for no filtering is possible (or alternatively that we can use an empty blacklist).

E.g we can assume that a setup like this is possible:
- Flag config `(tenant_a, <no value>, <no value>, x86_64)`
- Applies to `(tenant_a, some_app, some_instance, x86_64)`

Could you confirm @hakonhall ?

### DIFF
--- a/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
+++ b/flags/src/main/java/com/yahoo/vespa/flags/PermanentFlags.java
@@ -178,9 +178,9 @@ public class PermanentFlags {
 
     public static final UnboundJacksonFlag<CustomerRpmServiceList> CUSTOMER_RPM_SERVICES = defineJacksonFlag(
             "customer-rpm-services", CustomerRpmServiceList.empty(), CustomerRpmServiceList.class,
-            "Specifies customer rpm services to run on enclave hosts.",
+            "Specifies customer rpm services to run on enclave tenant hosts.",
             "Takes effect on next host admin tick.",
-            TENANT_ID);
+            TENANT_ID, APPLICATION, INSTANCE_ID, ARCHITECTURE);
 
     private static final String VERSION_QUALIFIER_REGEX = "[a-zA-Z0-9_-]+";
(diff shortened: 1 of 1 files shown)


hakonhall:
What is "this setup"?   Just this PR seen in isolation?  If so:  These dimensions are only for informational purposes so has no effect either way.  We may introduce a meaning to them in a later stage, e.g. verifying that exactly all of them are set and none others.

hakonhall (approved):

evgiz:
The dimensions are not purely informational - looks like they're used for validation in `FlagDefinition`, resolve in the controller and in various serialization contexts.  By "this setup" I mean this PR in combination with the flag configuration described:

> E.g we can assume that a setup like this is possible:
> 
> - Flag config `(tenant_a, <no value>, <no value>, x86_64)`
> - Applies to `(tenant_a, some_app, some_instance, x86_64)`

Since I'm not familiar with the way flag-distribution and binding is implemented, I'm curious if apis or validation checks could fail in this scenario and if the correct value would be bound given that the host queries the flag source with the full set of values.

evgiz:
The `FlagDefinition` constructor called by `Flags::define` does validation checking on those dimensions and exposes them to other methods with various business logic, so I wouldn't say that's purely informational. But I digress.

The intended goal of this PR is supporting both:
- A tenant-global feature flag value
- A tenant *app/instance specific* flag value

Consider the case when:
- Flag config is defined with `value=a` and whitelisted conditions `tenant=mytenant`, `architecture=x86_64`
- Host admin binds to the flag with dimensions `tenant=mytenant`, `app=myapp`, `instance=myinstance`, `architecture=x86_64`

When querying against a flag in this way, when only a subset of the queried dimensions exist in the flag condition config, *will the config be matched and the value `a` be returned?*.

If not, we'll need some other way to manage this.

hakonhall:
Yes.  If a flag has an override, the value is resolved by finding the first rule that has all of its conditions met.  In your case you mention only one rule, and the 2 conditions are true, so its value is returned.  If no rule is found, or if not override is specified, the default value is used.

PR #34015, opened 2025-05-14
Bratseth/smarter chunking

bratseth:
Somewhat smarter chunking. I recommend following each commit.

### DIFF
--- a/linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
+++ b/linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
@@ -63,7 +63,7 @@ public class FixedLengthChunker implements Chunker {
             while (index < text.length()) {
                 int currentChar = text.codePointAt(index);
                 currentChunk.appendCodePoint(currentChar);
-                if (++currentLength >= targetLength && (isCjk || (!isLetter(index) && !isLetter(nextIndex())))) {
+                if (endOfChunk(++currentLength)) {
                     chunks.add(new Chunk(currentChunk.toString()));
                     currentChunk.setLength(0);
                     currentLength = 0;
@@ -73,6 +73,12 @@ public class FixedLengthChunker implements Chunker {
             if (currentLength > 0)
                 chunks.add(new Chunk(currentChunk.toString()));
             return chunks;
+        }
+
+        private boolean endOfChunk(int currentLength) {
+             if (currentLength < targetLength) return false;
+             if (isCjk) return true;
+             return !isLetter(index) && !isLetter(nextIndex());
         }
 
         int charAt(int index) {
--- a/linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
+++ b/linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
@@ -8,8 +8,15 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
- * A chunker which splits a text into chunks at the first non-word/letter character after a given target chunk length
- * measured in codepoints (or precisely at that length, for CJK languages).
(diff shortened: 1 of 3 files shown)


linguistics/src/test/java/ai/vespa/language/chunker/FixedLengthChunkerTest.java
```diff
@@ -22,23 +22,54 @@ public void testChunker() {
...
+
+        // -> First chunk is 20 long
+        assertChunks(20, "first sentence, end. Next sentence.    ",
+                     "first sentence, end.", " Next sentence.    ");
+        // -> First chunk is 21 long as we get to a soft boundary
+        assertChunks(20, "first sentence, stop.Next sentence.    ",
+                     "first sentence, stop.", "Next sentence.    ");
+        // -> First chunk is 22 long, as there is no boundary before reaching the hard limit
+        assertChunks(20, "first sentence, endNext sentence.      ",
+                     "first sentence, endNex", "t sentence.      ");
```
glebashnik, line 45: Splitting words like this can lead to meaningless tokens and degrade retrieval quality.

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
```diff
@@ -8,8 +8,18 @@
...
+ * target chunk length measured in characters (or precisely at that length, for CJK languages).
+ *
+ * If there are no double non-letter/digit characters within 5% above the target length,
+ * the chunk split will be at the first single non-letter/digit character.
+ *
+ * If there are no double non-letter/digit characters within 10% above the target length,
+ * the chunk split will be at that position, so the absolute max chunk length will be 10% above the target
+ * length.
+ *
+ * The given target chunk length is adjusted down to prefer a more even chunk length distribution to account for the
```
glebashnik, line 21: This would make more sense with word/token-level chunks, since we embed and index on word/token level.
I suspect that for character-level chucks, these 5% adjustments are negligible relative to variations in word length.

glebashnik (changes requested):

linguistics/src/test/java/ai/vespa/language/chunker/FixedLengthChunkerTest.java
bratseth: In these tests we use unrealistically short chunks lengths to test boundary conditions.
With a chunk size of 1000 characters this condition will kick in at length 1100, so you need to have a contiguous sequence of 100 letter/digit characters before it is split like this.
The purpose of this is robustness given such degenerate input.

linguistics/src/main/java/ai/vespa/language/chunker/FixedLengthChunker.java
bratseth: I don't quite understand what you mean. Is this to the 5% margin before we split on words not sentences, or ensuring similar distances?

bratseth: ... sorry, I got it now. Yes, often this won't matter much, but if your avg. size is slightly above your fixed-length it matters a lot.

glebashnik: Ok, got it now.
"double non-letter/digit" is a proxy of (part of) sentence separator.
"single non-letter/digit" character is a proxy for word separator.
It does make sense to have a bit longer chunk to finish sentence/word.

linguistics/src/test/java/ai/vespa/language/chunker/FixedLengthChunkerTest.java
glebashnik: Right, this is only a problem with unrealistically small chunks.

glebashnik (approved):

PR #34021, opened 2025-05-14
avoid extra fill() round-trips

arnej27959:
This is a bit of a hack until more semantics are changed / more cleanup done.  Should make actual behavior OK as before, at least in the simple cases.

@bjormel FYI


frodelu (approved):

bratseth (approved):
Ok, looks "good" to me :-)

PR #34023, opened 2025-05-14
Make summary elements selector consistent the same way as [MERGEOK]

toregge:
summary transform is made consistent.

@havardpe : please review
@bratseth : FYI


bratseth (approved):
Nice catch. Should be MERGEOK I think?

toregge:
MERGEOK tag added. The changes in vespa-engine/vespa#34001 depends on summary element selector being consistent for a summary across all document-summary instances, thus vespa 8.520.11 and newer have issues with matched-elements-only (until this PR is merged).

PR #34025, opened 2025-05-15
handle indexing statements with optional inputs

arnej27959:
* an indexing statement that has just one required input but also some optional input can also be considered as a single-input statement.

this is to make it possible (but not easy) to allow partial update clearing (aka assign null) for fields with extra set_language wrappers.


bratseth (approved):

bratseth:
It really is time for us to add the "text" type :-)

PR #34029, opened 2025-05-15
avoid entire fill phase when selecting only matchfeatures

arnej27959:
note: this also skips fill when fetching a subset of fields already filled.

@bratseth please review
@havardpe FYI

this also skips extra fills if you fill first a,b,c,d and then later fill a,c fields.

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
@@ -12,6 +12,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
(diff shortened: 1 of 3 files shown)


container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
```diff
@@ -196,6 +204,8 @@ private Set<String> getFieldsForClass(String wantClass) {
             else if (isFieldListRequest(wantClass)) {
                 var fieldSet = parseFieldList(wantClass);
                 knownSummaryClasses.put(wantClass, fieldSet);
+            } else {
+                log.log(Level.WARNING, "unknown summary class: " + wantClass);
```
bratseth, line 208: We should avoid adding log lines on invalid requests, throw instead?

bratseth (approved):

PR #34036, opened 2025-05-16
Magnus/lsp refactor rank expressions

Mangern:

Simplify the way ranking expressions are stored, parsed and embedded into the syntax tree.
The previous implementation did a lot of nontrivial manipulation of the syntax tree which made it hard to modify or implement new features. This should hopefully make the process simpler.

The main difference is moving away from regex to capture ranking expressions, and rather expand the grammar definition of the schema language.


theodorklauritzen (approved):
LGTM

PR #34039, opened 2025-05-17
Handle multiple stems with Lucene linguistics

bratseth:


arnej27959 (approved):
👍 
We may want to allow startOffset inside range from current startOffset to same plus current origin length, but this should be a safe improvement as far as I can see.

PR #34040, opened 2025-05-18
Bratseth/auto unpack bits

bratseth:
I want to do some validation on this, but it'll need some more work to connect the dots - this just keeps track of which fields it is.


arnej27959 (approved):
👍

PR #34041, opened 2025-05-19
Add container memory check into vespa-wrapper

onurkaracali:

### DIFF
--- a/client/go/internal/admin/vespa-wrapper/services/container.go
+++ b/client/go/internal/admin/vespa-wrapper/services/container.go
@@ -10,7 +10,7 @@ import (
 	"github.com/vespa-engine/vespa/client/go/internal/osutil"
 )
 
-const minRequiredMemoryInBytes = 4 * 1024 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


client/go/internal/admin/vespa-wrapper/services/container.go
```diff
@@ -0,0 +1,101 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/vespa-engine/vespa/client/go/internal/osutil"
+)
+
+const minRequiredMemoryInBytes = 4 * 1024 * 1024 * 1024
```
andreer, line 13: The actual available memory when i start a 4G podman machine is lower:
MemAvailable:    3516160 kB

Suggest we lower the number from 4 to 3.2

onurkaracali: Updated it to `3 * 1024 * 1024 * 1024`

andreer (approved):

PR #34044, opened 2025-05-19
refactoring StreamingBackend

arnej27959:
* use Iterable / modern for loops
* differentiate between invalid query and invalid programmatic fill
* move helper classes and methods only used for streaming from VespaBackend to StreamingBackend
* pack some data used together into a Context, which can also carry around a PartialSummaryHandler (as yet unused)
* no functional changes

@bjorncs or @bratseth please review

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
@@ -227,9 +227,12 @@ public abstract class VespaBackend {
             } else {
                 var db = getDocumentDatabase(query);
                 if (db != null && ! db.getDocsumDefinitionSet().hasDocsum(summaryClass)) {
+                    // intentional reference compare, true when execution.fill(result) was called:
                     if (summaryClass == query.getPresentation().getSummary()) {
+                        // problem comes from the query:
                         throw new IllegalInputException("invalid presentation.summary=" + summaryClass);
                     } else {
+                        // problem comes from a Searcher doing fill with explicit summaryClass:
                         throw new IllegalArgumentException("invalid fill() with summaryClass=" + summaryClass);
                     }
                 }
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
@@ -20,7 +20,6 @@ import com.yahoo.search.result.Hit;
 import com.yahoo.searchlib.aggregation.Grouping;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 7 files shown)


bratseth (dismissed):
👍

container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
```diff
@@ -228,7 +227,11 @@ protected String ensureLegalSummaryClass(Query query, String summaryClass) {
             } else {
                 var db = getDocumentDatabase(query);
                 if (db != null && ! db.getDocsumDefinitionSet().hasDocsum(summaryClass)) {
-                    throw new IllegalInputException("invalid presentation.summary=" + summaryClass);
+                    if (summaryClass == query.getPresentation().getSummary()) {
```
bjorncs, line 230: Is reference equality intended here? I assume we want `.equals()` instead?

container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
```diff
@@ -228,7 +227,11 @@ protected String ensureLegalSummaryClass(Query query, String summaryClass) {
...
+                        throw new IllegalInputException("invalid presentation.summary=" + summaryClass);
+                    } else {
+                        throw new IllegalArgumentException("invalid fill() with summaryClass=" + summaryClass);
```
bjorncs, line 233: What's the difference of using `IllegalArgumentException` vs `IllegalInputException` in this context?

container-search/src/main/java/com/yahoo/vespa/streamingvisitors/StreamingBackend.java
```diff
@@ -368,10 +377,95 @@ public VisitorSession createVisitorSession(VisitorParameters params) throws Pars
...
+        String lastError = null;
+        int packetIndex = 0;
+        for (Hit hit : iterableHits(result)) {
+            // System.err.println("fill? hit " + hit);
+            if (hit instanceof FastHit fastHit && ! hit.isFilled(summaryClass)) {
+                // System.err.println("yes for " + summaryClass);
+
+                DocsumPacket docsum = packets[packetIndex];
+
+                packetIndex++;
```
bjorncs, line 439: The mapping from hit to packet index is confusing. Why can't `summaryMap` be used directly?

bjorncs (commented):
See comments. Let's talk offline.

container-search/src/main/java/com/yahoo/prelude/fastsearch/VespaBackend.java
bjorncs: This was intentional - please add a comment describing why.

container-search/src/main/java/com/yahoo/vespa/streamingvisitors/StreamingBackend.java
bjorncs: Alternatively remove the `! hit.isFilled(summaryClass)` check if it's guaranteed to always evaluate to `true`

bjorncs (approved):

PR #34051, opened 2025-05-19
add logic to handle [presentation]

arnej27959:
@bratseth please review
@bjorncs FYI

note 1: some of the tricky logic here is to avoid changing existing behavior unnecessarily.
note 2: this should be in sync with the decision from architecture meeting.


bratseth (approved):
Nice

PR #34058, opened 2025-05-20
Remove MatchedElementsFilterDFW, use CopyDFW instead.

toregge:
Rename corresponding unit test.

@havardpe : please review
@vekterli : FYI


searchsummary/src/tests/docsummary/copy_dfw/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+vespa_add_executable(searchsummary_copy_dfw_test_app TEST
```
havardpe, line 2: I think there is a new CMake function to add a test now... (not having to write the name 3 times)

havardpe (approved):

PR #34063, opened 2025-05-20
more usage of PartialSummaryHandler:

arnej27959:
* move summary class validation
* avoid fill with explicit query.getPresentation().getSummary()
* prepare to change fill(result) to use [presentation]

@bjorncs / @bratseth please review


bratseth (approved):
👍

container-search/src/main/java/com/yahoo/search/handler/SearchHandler.java
```diff
@@ -349,7 +349,7 @@ public Result searchAndFill(Query query, Chain<? extends Searcher> searchChain)
 
         ensureQuerySet(result, query);
         // TODO: consider fill(result, "[presentation]") instead
```
bjorncs, line 351: TODO is still relevant?

PR #34067, opened 2025-05-21
Block lid space compaction job while draining outstanding operations.

toregge:
@arnej27959 : please review
@vekterli : FYI


arnej27959 (approved):
this looks good to me

PR #34071, opened 2025-05-21
Drop MATCHED_ELEMENTS_FILTER and MATCHED_ATTRIBUTE_ELEMENTS_FILTER su…

toregge:
…mmary

transforms in config model.

@havardpe : please review
@vekterli : FYI


config-model/src/main/java/com/yahoo/schema/processing/AddDataTypeAndTransformToSummaryOfImportedFields.java
```diff
@@ -64,12 +64,7 @@ private static void setAttributeCombinerTransform(ImmutableSDField field, Summar
         if (summaryField.hasUnresolvedType()) {
             summaryField.setResolvedDataType(field.getDataType());
         }
-        if (summaryField.getTransform() == SummaryTransform.MATCHED_ATTRIBUTE_ELEMENTS_FILTER) {
-            // This field already has the correct transform.
-            return;
-        } else if (summaryField.getTransform() == SummaryTransform.MATCHED_ELEMENTS_FILTER) {
-            summaryField.setTransform(SummaryTransform.MATCHED_ATTRIBUTE_ELEMENTS_FILTER);
-        } else {
+        if (summaryField.getTransform() == SummaryTransform.NONE) {
```
havardpe, line 67: what if 'none of the above'?

config-model/src/main/java/com/yahoo/schema/processing/MatchedElementsOnlyResolver.java
```diff
@@ -45,13 +46,7 @@ private void processSummaryField(DocumentSummary summary, SummaryField field, bo
         var sourceField = schema.getField(field.getSingleSource());
         if (sourceField != null) {
             if (isSupportedComplexField(sourceField)) {
```
havardpe, line 48: could be simplified further?

havardpe (approved):

config-model/src/main/java/com/yahoo/schema/processing/AddDataTypeAndTransformToSummaryOfImportedFields.java
toregge: The code is structured to match the setAttributeTransform() member function: only set the transform if it has not yet been set (implied by it still being NONE).

PR #34077, opened 2025-05-22
Require validation override when changing hnsw index settings

hmusum:


config-model/src/test/java/com/yahoo/vespa/model/application/validation/change/search/AttributeChangeValidatorTest.java
```diff
@@ -190,27 +201,83 @@ void changing_distance_metric_with_hnsw_index_enabled_requires_restart() throws
...
+                  indexing: attribute | index
+                  index {
+                    hnsw {
+                      neighbors-to-explore-at-insert: 100
+                    }
+                  }
+                }
+                """;
+
+        var expectedOutput = "Field 'f1' changed: change hnsw index property 'neighbors-to-explore-at-insert' from '200' to '100'";
```
andreer, line 263: > Adding or removing the [hnsw index](https://docs.vespa.ai/en/reference/schema-reference.html?mode=cloud#index-hnsw) on a tensor attribute field, or changing the **distance-metric** or max-links-per-node requires a restart to rebuild the index. **Changing neighbors-to-explore-at-insert requires a restart, but does not rebuild the index.**

Suggest we only require validation for those things that rebuild the index (make sure to include change of distance-metric)

hmusum: Yes, that makes more sense, thanks

hmusum:
@andreer please take a look again

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/search/AttributeChangeValidator.java
```diff
@@ -136,27 +136,37 @@ private static void validateAttributePredicate(ClusterSpec.Id id,
...
-    private static <T> void validateAttributeProperty(ClusterSpec.Id id,
+    private <T> void validateAttributeProperty(ClusterSpec.Id id,
                                                       Attribute current, Attribute next,
                                                       Function<Attribute, T> settingValueProvider, String setting,
                                                       List<VespaConfigChangeAction> result) {
         T currentValue = settingValueProvider.apply(current);
         T nextValue = settingValueProvider.apply(next);
         if ( ! Objects.equals(currentValue, nextValue)) {
             String message = String.format("change property '%s' from '%s' to '%s'", setting, currentValue, nextValue);
+            if (hasHnswIndex(current) && hasHnswIndex(next))
```
andreer, line 147: Is this intended to catch _adding_ a hnsw index? If so, should the first hasHnswIndex be negated?

andreer (approved):

config-model/src/main/java/com/yahoo/vespa/model/application/validation/change/search/AttributeChangeValidator.java
hmusum: It catches only changes to `distance-metric` when already having a hnsw index. I wasn't intending to require a validation override when adding a hnsw index in this PR, sounds a bit restrictive, but I am not sure...

andreer: Agreed, was just me not understanding the code here then :)

PR #34080, opened 2025-05-22
Check fnet frt values encoding and decoding with no values.

toregge:
@vekterli : please review
@havardpe : FYI


vekterli (approved):
:+1:

vekterli:
Once [N3322](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf) has made its way through all our toolchains, we can remove these special case branches... :eyes:

PR #33677, opened 2025-03-26
Clean-up golang part

sarzhann:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

## What
This pull request introduces several changes to the Go client codebase, focusing on improving code quality, adding linting support, and addressing minor issues such as formatting, error handling, and code cleanup. The most important changes include enabling linting with `golangci-lint`, adding a configuration file for linters, and making various code improvements across multiple files.

### Linting and Configuration:
* Added `golangci-lint` as a step in the GitHub Actions workflow for the Go client to ensure code quality and detect issues in new changes.
* Introduced a `.golangci.yaml` configuration file to enable specific linters and define exclusions and formatting rules.

### Code Quality Improvements:
* Replaced numeric file permission literals (e.g., `0755`) with octal literals (e.g., `0o755`) for better readability and consistency.
* Simplified loop constructs by removing unnecessary variables

### Error Handling and Formatting:
* Improved error messages by standardizing capitalization and formatting
* Enhanced error handling by using `errors.As` for type assertion and wrapping errors with additional context. 

### Cleanup and Refactoring:
* Removed redundant variable declarations and unused code blocks to simplify function implementations.
* Refactored function signatures and return statements for clarity and consistency.

PR #34085, opened 2025-05-23
Reapply "chore: clean-up client golang coding standards"

esolitos:
Reapply vespa-engine/vespa#33677


sarzhann:
I would prefer to reapply with multiple commits 🙈

PR #34086, opened 2025-05-23
allow unescaped newline/tab inside JSON strings

arnej27959:
@bjorncs please review
@radu-gheorghe FYI


radu-gheorghe:
Awesome, thanks! This will work for both vespa-cli and HTTP requests, right?

bjorncs (dismissed):

bjorncs:
Consider adding a comment on why it's allowed in `JsonSingleLevelMap`

arnej27959:
@radu-gheorghe yes - one change is just for accepting newlines inside JSON strings received in POST, the other is for "vespa query --file" so it can parse the same.

bjorncs (approved):

PR #34094, opened 2025-05-23
Test and fix LightGBM categorical integer handling

thomasht86:

Another follow-up fix to #34084 to make sure the categorical integers are not enclosed in quotes in the ranking expression. 
@bjorncs please review.


model-integration/src/main/java/ai/vespa/rankingexpression/importer/lightgbm/LightGBMParser.java
```diff
@@ -27,6 +27,7 @@ class LightGBMParser {
     private final List<LightGBMNode> nodes;
     private final List<String> featureNames;
     private final Map<Integer, List<String>> categoryValues;  // pr feature index
+    private Map<Integer, Boolean> categoricalIntegerFeatures = new HashMap<>();  // true if feature values are integers
```
bratseth, line 30: You could use a Set containing only the true entires.

bratseth (approved):

PR #34097, opened 2025-05-23
Don't fail if cgroupsv2 memory limit is not available use available sys memory in container startup memory check

onurkaracali:

### DIFF
--- a/client/go/internal/admin/vespa-wrapper/services/container.go
+++ b/client/go/internal/admin/vespa-wrapper/services/container.go
@@ -18,14 +18,10 @@ func VerifyAvailableMemory() {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


andreer (dismissed):

onurkaracali:
> Hmm ... if the cgroup memory limit exists and is > 3GB but the amount of actual memory is < 3GB it should probably still fail the check, not sure that happens now

Yes, you are right. This is the case when a podman machine has a lower memory limit than the container's limit. I tested this with 2G podman machine and 4G container limit. We should use the min of them

andreer (approved):

hmusum (approved):
Nice! Merging

PR #34099, opened 2025-05-23
Stop using feature flag SEARCH_MMAP_ADVISE

hmusum:


hmusum:
@vekterli do you think we should we remove this flag as done her, keep it, or move it to PermanentFlags?

vekterli:
I'm not aware of any use cases where configuring this away from the current default makes a lot of sense. If there are, we should probably make it into an explicit tuning setting in `services.xml` instead. It can always be set via a config override in the worst case.

hmusum:
Good, I think it's best to remove it then, please review

vekterli (approved):
:+1:

PR #34112, opened 2025-05-26
Don't set max_feed_op_batch_size config (same as default value)

hmusum:
Default value set to 64 in https://github.com/vespa-engine/vespa/pull/34083


vekterli (approved):
:+1:

PR #34117, opened 2025-05-27
AwsCredentials with custom duration

olaaun:
`ztsClient.getAwsTemporaryCredentials` returns credentials with default expiry if duration is `null`


vespa-athenz/src/main/java/com/yahoo/vespa/athenz/aws/AwsCredentials.java
```diff
@@ -27,43 +27,49 @@ public class AwsCredentials implements AutoCloseable {
     private final ZtsClient ztsClient;
     private final String externalId;
     private final boolean close;
+    private final Duration duration;
```
hakonhall, line 30: It's not obvious from the name of this field, nor its type, that this can legitimately be null.  Consider fixing this.

hakonhall (dismissed):

hakonhall (approved):

PR #34124, opened 2025-05-27
fix(release-java-artifacts): Add dummy POM for Maven publishing

esolitos:
- Added a dummy POM file to enable Maven publishing for Java artifacts


bjorncs (approved):
Assuming `repositoryDirectory` is a valid parameter

PR #34147, opened 2025-05-30
Add support for comments in GroupingParser

Mangern:
This adds support for comments in the grouping parser. This was already supported in the YQL parser, so it would be natural to support it in grouping operations as well.
Allowing comments enables writing more readable expressions across several lines when editing queries in an editor-like environment.


arnej27959:
personally I would prefer # for comment-to-end-of-line, @bratseth do you think it's possible to support that as well, or instead of //

bratseth:
Yes '#' should work fine.

arnej27959:
@bratseth please approve or suggest changes...

bratseth (approved):

PR #34151, opened 2025-05-30
Implement application [list|show]

ean:
List all application for a tenant, and show basic information about them. For now only their instances and deployment zones are shown.

### DIFF
--- a/client/go/internal/cli/cmd/application.go
+++ b/client/go/internal/cli/cmd/application.go
@@ -10,7 +10,7 @@ import (
 )
 
 func newApplicationCmd() *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Hidden: true,
 		Use:    "application",
 		Short:  "",
@@ -24,13 +24,15 @@ $ vespa application show -a <tenant>.<application>`,
 			return fmt.Errorf("invalid command: %s", args[0])
 		},
 	}
+	return cmd
 }
 
 func newApplicationListCmd(cli *CLI) *cobra.Command {
+	var listAllApplications bool
 	cmd := &cobra.Command{
 		Use:   "list",
 		Short: "List all application for a given tenant",
-		Long: `The tenant is provided by specifying -a <tenant>.IGNORED.
+		Long: `The tenant is provided by specifying -a <tenant>.ignored
 The applications are listed without any extra information. In the format <tenant>.<application>.`,
 		DisableAutoGenTag: true,
 		SilenceUsage:      true,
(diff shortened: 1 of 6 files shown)


client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+func newApplicationShowCmd(cli *CLI) *cobra.Command {
+	var format string
+	cmd := &cobra.Command{
+		Use:               "show",
+		Short:             "Show information about a given application",
+		Long:              `Show the known instances of an application, and the Vespa Zones they are deployed in.`,
+		DisableAutoGenTag: true,
+		SilenceUsage:      true,
+		Example: `$ vespa application show -a <tenant>.<application>
+$ vespa application show -a <tenant.application> --format plain`,
```
esolitos: I tested this (`vespa application show`) and I have some notes:

1. I initially run without `-a` option, because I did not RTFM, so: `bin/vespa application show my-tenant.my-app`. I was surprided to get an answer like the following:
  ```
    my-other-tenant.default:
      dev.aws-us-east-1c
  ```
2. The options `--quiet`, `--zone`, `--instance`, `--cluster` don't see to have any effect at all? They seem to be global, but imho *if possible* they should be omitted from this command, as the are only confusing.

client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/vespa-engine/vespa/client/go/internal/vespa"
+)
+
+func newApplicationCmd() *cobra.Command {
+	return &cobra.Command{
+		Hidden: true,
+		Use:    "application",
```
esolitos, line 15: Could we alias this to `app` as well?

client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+}
+
+func newApplicationListCmd(cli *CLI) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List all application for a given tenant",
+		Long: `The tenant is provided by specifying -a <tenant>.IGNORED.
+The applications are listed without any extra information. In the format <tenant>.<application>.`,
```
esolitos: Feedback after testing:

1. The `-a` flag is not properly documented, it can be used but doesn't show in the Flags.
2. The `-a` option is a but odd, expects an "application" name with the tenante prefix. 🤔 Perhaps we should either supporte `-t <tenant-name>` or support _also_ `-a <tenant>`, so without the application name.

client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+		DisableAutoGenTag: true,
+		SilenceUsage:      true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			target, err := cli.target(targetOptions{noCertificate: true, supportedType: cloudTargetOnly})
+			if err != nil {
+				return err
+			}
+			if target.Type() != vespa.TargetCloud && target.Type() != vespa.TargetHosted {
+				return fmt.Errorf("application list does not support %s target", target.Type())
+			}
```
esolitos: We could provide a better hint, like:
```
❯ bin/vespa application list
Error: command does not support local target

Hint: to switch mode simply run the following
 $ vespa config set target cloud
```

client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+				return err
+			}
+
+			seen := map[string]bool{}
+			for _, app := range tenant.Applications {
+				app := fmt.Sprintf("%s.%s", app.Tenant, app.Application)
+				// skip application with multiple instances
+				if seen[app] {
+					continue
+				}
```
esolitos: Add an option (default?) to filter out applications without active deployments?

client/go/internal/cli/cmd/application.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/vespa-engine/vespa/client/go/internal/vespa"
+)
+
+func newApplicationCmd() *cobra.Command {
+	return &cobra.Command{
+		Hidden: true,
```
esolitos, line 14: Why hidden?

esolitos (commented):
Looks good, I have some suggestions and comments. 😊

client/go/internal/cli/cmd/application.go
ean: Because of the issues you mention (especially `-a <tenant>.ignored`, and it being very limited in output right now)

client/go/internal/cli/cmd/application.go
ean: Yes, I agree. But -a is a required flag when setting up the cloud target. And I don't want to rewrite the entire CLI right now.
Hence the hiding of the subcommand for now, while still being able to provide this to users that need to list their applications programmatically.

client/go/internal/cli/cmd/application.go
esolitos: Given what we discussed IRL, we could simply output a warning when those options are set in this command.

client/go/internal/cli/cmd/application.go
ean: Let's wait until we look at making this `public` (not hidden)

esolitos (dismissed):

esolitos (approved):

PR #34156, opened 2025-06-02
Bratseth/grammar decompose

bratseth:
We really need 56 different query types to express the multidimensional variants we support, so it was time to split it into separate parameters.


arnej27959 (approved):
looks like a great idea to me, let us see if factory agrees 👍

PR #34157, opened 2025-06-02
Track, aggregate and report content node response (network) errors

vekterli:
@toregge please review the distributor-side changes in particular (cb06a19bd64b30ac249a59e6d55b22fff7c86baa)
@hakonhall please review the cluster controller-side changes in particular (83e6dff37a668867619e0307f536a9686f968470)

**Distributor-side**
This PR adds distributor-internal per-node monotonic counters that track the number of messages sent and received, with error categorization of responses. Currently categorizes response errors into those that are network-related, clock sync (NTP) related and "other" for whatever's left over. We currently only have semantics for caring about network-related errors, so the focus is on that.

Message stats tracking is added to the per-stripe `PendingMessageTracker` components, as well as the top level distributor bucket DB updater. The latter is required to observe state convergence problems, as the top level bucket DB updater handles all communication with content nodes during cluster state changes on behalf of the stripes.

The state manager's host info reporting has been extended to include all nodes for which network errors have been observed during the previous statistics non-overlapping sampling period. This period is currently hard-coded to 60 seconds.

**Cluster controller-side**
The PR adds parsing and handling of the `response-stats` field in the `HostInfo` payload from distributor nodes. Currently only network errors are considered. These are further aggregated per distributor per content node, similar to how we track bucket space-specific statistics already. Aggregation is sparse, so in a healthy cluster the expected additional overhead is minimal.

Unlike the existing statistics, error statistics are updated even if the distributor has _not_ converged to the most recent cluster state version. This is to avoid catch-22 situations where a distributor fails to converge to the newest version due to connectivity errors, but where error statistics are not tracked and reported because the node has not yet managed to converge.

In this initial implementation, errors are only surfaced on the cluster controller status page as part of the content node information view. If at least one distributor reports network errors, a warning will be printed for the node. If the reporting distributor is _also_ not converging to the latest cluster state version, an explicit convergence warning will also be printed.

Aggregation is currently gated by a new live config parameter. If disabled, existing aggregated information will be cleared. In this initial PR, the config flag is set to _disabled_ by default. Will enable in a separate PR.


vekterli:
Sorry for the repeated force-pushing—the Buildkite compilation for some reason triggers `hash_map`-related linkage errors that do _not_ show up when I build locally :eyes:

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ContentClusterErrorStats.java
```diff
@@ -0,0 +1,63 @@
...
+    public Map<Integer, ContentNodeErrorStats> getAllNodeErrorStats() {
+        return contentNodeErrorStats;
+    }
+
+    public void clearAllStats() {
+        contentNodeErrorStats.clear();
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
hakonhall: Consider making this class a record.  For instance, it will automatically provide you with a getter for contentNodeErrorStats, an equals, hashCode, and toString.

If ContentClusterStats is made a record too, the methods you fix there would have been automatically fixed.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResponseStats.java
```diff
@@ -0,0 +1,91 @@
...
+    public boolean hasClockSkewErrors() { return clockSkewErrorCount > 0; }
+    public boolean hasUncategorizedErrors() { return uncategorizedErrorCount > 0; }
+
+    public boolean hasErrors() {
+        return hasNetworkErrors() || hasClockSkewErrors() || hasUncategorizedErrors();
+    }
+
+    public double sampleWindowSec() { return sampleWindowSec; }
+
+    public long totalResponseCount() { return totalResponseCount; }
```
hakonhall: This class is ideal as a record.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ClusterStatsAggregator.java
```diff
@@ -105,6 +118,37 @@ void updateForDistributor(int distributorIndex, ContentClusterStats clusterStats
         addStatsFromDistributor(distributorIndex, clusterStats);
     }
 
+    void updateErrorStatsFromDistributor(int distributorIndex, ContentClusterErrorStats updatedErrorStats) {
```
hakonhall, line 121: It's not entirely clear to me what happens during topology changes, e.g. if node 10 is removed from the content cluster.  Will all data related to node 10 be removed from the data structures, or will there be a memory leak?

toregge (dismissed):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ClusterStatsAggregator.java
vekterli: The `ClusterStatsAggregator` instance is created and owned by a singular `ClusterStateView` instance, which in turn is created anew when any new cluster state version is _published_ (see `StateVersionTracker.updateStatesForNewVersion`).

I.e. for better or worse, a new state version wipes all statistics, both for bucket spaces and error reporting. The assumption is that this is fine in practice, since the error reporting is primarily intended to help explain why a cluster is consistently failing to converge to a new state version, so there will quickly be sufficient evidence gathered to point out a likely culprit.

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ContentClusterErrorStats.java
vekterli: Done (for both mentioned classes)

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/hostinfo/ResponseStats.java
vekterli: Done

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ContentClusterErrorStats.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public ContentClusterErrorStats(Set<Integer> storageNodes) {
+        this(toEmptyErrorStats(storageNodes));
+    }
+
+    public ContentNodeErrorStats getNodeErrorStats(Integer index) {
+        return contentNodeErrorStats.get(index);
+    }
+
+    public Map<Integer, ContentNodeErrorStats> getAllNodeErrorStats() {
```
hakonhall: Consider just using the automatically provided getter for a record member.

hakonhall (dismissed):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/ContentClusterErrorStats.java
vekterli: Done

hakonhall (approved):

Issue #28769, opened 2023-10-03
model.filter/filter is ignored with yql while recall works as expected


jobergum:
## Using `recall` in combination with YQL

Using [recall](https://docs.vespa.ai/en/reference/query-api-reference.html#recall)

>This is identical to [filter](https://docs.vespa.ai/en/reference/query-api-reference.html#model.filter), except that recall terms are not exposed to the ranking framework and thus not ranked. As such, one can not use unprefixed terms; they must either be positive or negative

```
{'yql': 'select id,title,text from sources * where {targetHits:10}nearestNeighbor(embedding, q)', 'query': 'how does the coronavirus respond to changes in the weather', 'ranking.profile': 'dense', 'presentation.format.tensors': 'short-value', 'hits': 3, 'language': 'en', 'timeout': '15s', 'presentation.timing': 'true', 'input.query(q)': 'embed(bge, "Represent this sentence for searching relevant passages: how does the coronavirus respond to changes in the weather")', 'recall': '+text:ARIMA +text:"novel coronavirus illness"', 'tracelevel': 3}
```
Gives the following correct query tree

```
{'message': 'sc0.num0 search to dispatch: query=[AND NEAREST_NEIGHBOR {field=embedding,queryTensorName=q,hnsw.exploreAdditionalHits=0,distanceThreshold=Infinity,approximate=true,targetHits=10} |text:arima |text:"novel coronaviru illness"] timeout=14977ms offset=0 hits=3 rankprofile[dense]
```
Which is the expected query tree. filter (not highlight), and ranking is disabled. 

## Using `model.filter` in combination with YQL 

```
{'yql': 'select id,title,text from sources * where {targetHits:10}nearestNeighbor(embedding, q)', 'query': 'how does the coronavirus respond to changes in the weather', 'ranking.profile': 'dense', 'presentation.format.tensors': 'short-value', 'hits': 3, 'language': 'en', 'timeout': '15s', 'presentation.timing': 'true', 'input.query(q)': 'embed(bge, "Represent this sentence for searching relevant passages: how does the coronavirus respond to changes in the weather")', 'filter': '+text:ARIMA +text:"novel coronavirus illness"', 'tracelevel': 3}
```

It gives the following incorrect query tree (filter is silently dropped) 
```
[...]

PR #34162, opened 2025-06-02
Require query string is specified whenever filter is specified

bjorncs:
See https://github.com/vespa-engine/vespa/issues/28769


bratseth (commented):
Will this go from silently ignoring to returning an error? If so I think it might break someone badly, and we could avoid that by adding the error without returning?

bjorncs:
@bratseth You're right. The last commit will add an error without failing the query.

bratseth (approved):

PR #34163, opened 2025-06-02
Revert "Bratseth/grammar decompose"

bratseth:
Reverts vespa-engine/vespa#34156

@arnej27959 the 'deploy all apps' integration test did _not_ agree.


arnej27959:
local search has query profile with:
```
<query-profile id="default">
    <field name="timeout">1</field>
    <field name="model.type">all</field>  <!-- required for VESPA-8+, since default value is 'weakAnd' -->
</query-profile>
```

bratseth:
Thanks!

PR #34167, opened 2025-06-03
Revert "Get maintenance job token in lid space compation job."

toregge:
Reverts vespa-engine/vespa#34161

This caused deadlocks.

@vekterli : please review
@arnej27959 : FYI

PR #34172, opened 2025-06-03
Reapply get maintenance job token in lid space compation job.

toregge:
@arnej27959 : please review
@vekterli : FYI

This reapplies #34161 (reverted in #34167), the changes in #34168 should prevent the deadlock.


arnej27959 (approved):
this should be fine to merge now

PR #34186, opened 2025-06-05
Magnus/lsp improved grouping support

Mangern:
Improved completion in YQL and Grouping expressions, still work to do on this, but now we get more completions.


theodorklauritzen (approved):
LGTM

PR #34192, opened 2025-06-05
Revert "Remove unused methods" [MERGEOK]

hmusum:
Reverts vespa-engine/vespa#34183

Looks like we had an app on an older version that I missed, need to revert


vekterli (approved):
:+1:

PR #34196, opened 2025-06-05
chore(messagebus): increase maxPendingCount to 4096

bjorncs:
Based on the config definition comment it’s not in use. Let’s see if it has any impact in the CI environment.


vekterli (approved):
:+1:

PR #34204, opened 2025-06-06
allow cors from vespa console (and localhost)

andreer:


container-disc/src/main/java/com/yahoo/container/jdisc/DataplaneProxyService.java
```diff
@@ -235,6 +240,30 @@ static String nginxConfig(
...
+                    """ : "";
+
+            String corsHandler = isDevEnvironment ? """
+                    if ($request_method = OPTIONS) {
+                                  add_header Access-Control-Allow-Origin $allow_origin;
+                                  add_header Vary Origin;
+                                  add_header Access-Control-Allow-Headers "Authorization";
+                                  add_header Content-Length 0;
+                                  add_header Content-Type text/plain;
+                                  return 200 "";
```
bjorncs, line 258: It's most common to return a 204 (No Content) response where `Content-Length` and `Content-Type` is not required.

container-disc/src/main/java/com/yahoo/container/jdisc/DataplaneProxyService.java
```diff
@@ -235,6 +240,30 @@ static String nginxConfig(
...
+                                }
+                                add_header Access-Control-Allow-Origin $allow_origin;
+                                add_header Vary Origin;
```
bjorncs, line 261: Afaik the `Access-Control-Request-Method` response header is required for pre-flight requests

container-disc/src/test/java/com/yahoo/container/jdisc/DataplaneProxyServiceTest.java
```diff
@@ -205,7 +205,7 @@ public void defaultProxyConfigUsesProxyProtocol() throws IOException {
         Path dummy = fs.getPath("/dummy.pem");
         Files.createFile(dummy);
 
-        String out = DataplaneProxyService.nginxConfig(templatePath, dummy, dummy, dummy, dummy, 1, 2, List.of("e"), fs.getPath("/pfx"), false);
+        String out = DataplaneProxyService.nginxConfig(templatePath, dummy, dummy, dummy, dummy, 1, 2, List.of("e"), fs.getPath("/pfx"), false, false);
```
bjorncs, line 208: I guess we cannot unit test against the real template as it's closed-source 😞

bjorncs (approved):

container-disc/src/test/java/com/yahoo/container/jdisc/DataplaneProxyServiceTest.java
andreer: Yes, unfortunate ...

PR #34218, opened 2025-06-10
Reapply Azure DNS entry types

olaaun:
This reverts commit ec1befeacd27ab4ea0ca9615674ecd7facbf4248, reversing changes made to 0f309e19c2ad28587df1595c9902d767f381190d.


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/node/Dns.java
```diff
@@ -31,7 +31,7 @@ public static Set<RecordType> recordTypesFor(IP.Version ipVersion, NodeType host
     }
 
     public static Set<RecordType> recordTypesFor(IP.Version ipVersion, NodeType hostType, CloudName cloudName, boolean enclave, boolean allowReverse) {
-        if (cloudName == CloudName.AWS || cloudName == CloudName.GCP) {
```
hakonhall, line 34: If the flag is false, this change would lose PUBLIC_FORWARD for v4.

hakonhall (approved):

PR #34219, opened 2025-06-10
Updated sidecars flag

glebashnik:


bjorncs (approved):
If you need further Jackson annotated classes, consider representing them as a Java record instead.

PR #34222, opened 2025-06-10
Only log lidspace compaction (re-)start events when debug logging is enabled

vekterli:
@toregge please review
@arnej27959 FYI

lidspace compaction throttling might cause a lot of stop-and-go for compaction, causing an excessive amount of (re-)start edge events. Should find a more elegant solution for this, but for now simply avoid logging any events for these cases unless `debug` level is active for the event logger component. Completion events are unconditionally logged as before.

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/common/eventlogger.cpp
+++ b/searchcore/src/vespa/searchcore/proton/common/eventlogger.cpp
@@ -206,17 +206,22 @@ EventLogger::lidSpaceCompactionStart(const string &subDbName,
                                      double lidBloatFactor, double allowedLidBloatFactor,
                                      uint32_t lidLimit, uint32_t lowestFreeLid)
 {
-    JSONStringer jstr;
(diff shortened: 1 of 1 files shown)


toregge (approved):
:+1: 

Note that we still have some info logs that is likely to produce even more output:

https://github.com/vespa-engine/vespa/blob/66753ec6320c77c5f95ae057436d6584e2bd9fa7/searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job.cpp#L213
https://github.com/vespa-engine/vespa/blob/66753ec6320c77c5f95ae057436d6584e2bd9fa7/searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job.cpp#L222
https://github.com/vespa-engine/vespa/blob/66753ec6320c77c5f95ae057436d6584e2bd9fa7/searchcore/src/vespa/searchcore/proton/server/lid_space_compaction_job.cpp#L229

PR #34223, opened 2025-06-10
Change default value of cluster_feed_block_noise_level from 0.0 to 0.01

hmusum:
When feeding is blocked resource usage needs to decrease below another, lower limit before getting unblocked. The change here is to make this lower limit 1 percent point below the configured limit.


hmusum:
@vekterli please take a look

vekterli (approved):
:+1:

PR #34225, opened 2025-06-10
fix: ensure request logging is applied to the default HTTP setup

bjorncs:


gjoranv (approved):
It would be nice to see an updated or new unit test here, but assuming it will be too costly.

PR #34229, opened 2025-06-10
Don't set job run status to sucess when command status is non-zero.

gjoranv:
- I wrongfully assumed that post-command would not run if the command failed. In this case, it's no big deal, though, as the pre-exit hook will shortly after set job status to failure.

### DIFF
--- a/.buildkite/plugins/factory-reporter/hooks/post-command
+++ b/.buildkite/plugins/factory-reporter/hooks/post-command
@@ -12,27 +12,34 @@ if [[ $SKIP_BUILDKITE_PLUGINS == "true" ]]; then
     echo "SKIP_BUILDKITE_PLUGINS is set. Skipping factory reporter"
     exit 0
 fi
+if [[ $BUILDKITE_PULL_REQUEST != "false" ]]; then
+    exit 0
+fi
 if [[ -z $PIPELINE ]]; then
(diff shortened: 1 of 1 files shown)


esolitos (approved):

.buildkite/plugins/factory-reporter/hooks/post-command
```diff
@@ -12,27 +12,36 @@ if [[ $SKIP_BUILDKITE_PLUGINS == "true" ]]; then
...
+START_SECONDS=$(buildkite-agent meta-data get start-seconds)
+if [[ -z $START_SECONDS ]]; then
+  echo "Error: start-seconds not set as meta-data" >&2
+  exit 1
+fi
 
+FACTORY_COMMAND=$(buildkite-agent meta-data get factory-command)
+if [[ -z $FACTORY_COMMAND ]]; then
+  echo "Error: factory-command not set as meta-data" >&2
+  exit 1
```
esolitos, line 32: I'm conflicted if we should exit with 1 in there two cases, afterall yes it's a "bad condition", but... 🤷🏼

gjoranv: This means something that should happen did not happen in a previous step, so I think it's a failure

PR #34240, opened 2025-06-11
Reduce time to wait before resetting restart penalty

hmusum:
Today a service that has been restarted will have a restart penalty until the service has been running 10 * MAX_RESTART_PENALTY, reduce this to 5 * MAX_RESTART_PENALTY, should be enough.


arnej27959 (approved):
👍

PR #34247, opened 2025-06-12
 Add initial-inflight-factor option to vespa-feed-client to mitigate the slow start

yjtakamabe:
## Is your feature request related to a problem? Please describe.
The feed client experiences a slow start because the default initial target inflight value is too low in our environment, causing a 2-3 minute delay in the time it takes to reach optimal throughput.

## Describe the solution you'd like
I would like an option called "initial-inflight-factor" that allows us to configure a higher initial target inflight.
This would help the feed client ramp up more quickly and avoid performance bottlenecks at startup.

```
    --initial-inflight-factor <arg>       Multiplier for minInflight to
                                          determine the initial
                                          targetInflight
```

```
$ vespa-feed-client \
  --initial-inflight-factor 64 \
  --file /path/to/json/file \
  --endpoint https://container-endpoint:443/
```

## Describe alternatives you've considered
One alternative we considered was setting the number of "connections" to 512 in order to mitigate the slow start.
However, this approach consumes significant machine resources.

```
$ vespa-feed-client \
  --connections 512 \
  --file /path/to/json/file \
  --endpoint https://container-endpoint:443/
```

## Additional context
In our environment, we set the initial-inflight-factor to 64.

### DIFF
--- a/vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/DynamicThrottler.java
+++ b/vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/DynamicThrottler.java
@@ -28,7 +28,7 @@ public class DynamicThrottler extends StaticThrottler {
 
     public DynamicThrottler(FeedClientBuilderImpl builder) {
         super(builder);
-        targetInflight = new AtomicLong(minInflight);
+        targetInflight = new AtomicLong(minInflight * Math.min(builder.initialInflightFactor, 256)); // maxInflight = 256 * minInflight;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


vespa-feed-client/src/main/java/ai/vespa/feed/client/impl/DynamicThrottler.java
```diff
@@ -28,7 +28,7 @@ public class DynamicThrottler extends StaticThrottler {
 
     public DynamicThrottler(FeedClientBuilderImpl builder) {
         super(builder);
-        targetInflight = new AtomicLong(minInflight);
+        targetInflight = new AtomicLong(minInflight * Math.min(builder.initialInflightFactor, 256)); // maxInflight = 256 * minInflight;
```
bjorncs: Please modify the expression to removed the hardcoded constant `256`, and instead cap the overall `targetInflight` to not exceed `maxInflight`.

yjtakamabe: Thank you for your feedback.
 I will remove the hardcoded constant 256 and update the expression.

bjorncs (approved):

bjorncs:
Thanks for your contribution @yjtakamabe - much appreciated! 👍 This will most likely be shipped with the next release (8.537).

yjtakamabe:
Thank you so much for reviewing and approving my contribution.

PR #34249, opened 2025-06-12
Use apache http client in config proxy

hmusum:
To be able to use DefaultHttpRequestRetryStrategy, which will retry responses with status codes 429 and 503 and retries requests when getting these

### DIFF
--- a/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
+++ b/config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
@@ -3,13 +3,16 @@ package com.yahoo.vespa.config.proxy.filedistribution;
 
 import org.apache.hc.client5.http.classic.HttpClient;
 import org.apache.hc.client5.http.classic.methods.HttpGet;
+import org.apache.hc.client5.http.config.RequestConfig;
 import org.apache.hc.client5.http.impl.DefaultHttpRequestRetryStrategy;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.core5.http.ClassicHttpResponse;
 import org.apache.hc.core5.http.message.BasicHeader;
+import org.apache.hc.core5.util.Timeout;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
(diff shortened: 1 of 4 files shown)


config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
```diff
@@ -72,4 +78,12 @@ boolean alreadyDownloaded(File downloadDir) {
         File contents = new File(downloadDir,fileName());
         return contents.exists() && contents.length() > 0;
     }
+
+    private static HttpClient createClient() {
+        return HttpClientBuilder.create()
+                                .setRetryStrategy(new DefaultHttpRequestRetryStrategy())
+                                .setUserAgent(USER_AGENT_MODEL_DOWNLOADER)
+                                .build();
```
bjorncs: Consider specifying connect timeout and friends. Some timeouts might be disabled by default. Afaik the default connect timeout is 3 minutes and response timeout is disabled by default.

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
```diff
@@ -23,47 +28,48 @@ class UrlDownloader {
...
-                    log.log(Level.FINE, () -> "URL '" + this.url + "' available at " + contentsPath);
-                    log.log(Level.INFO, String.format("Download of URL '%s' done in %.3f seconds",
-                                                      this.url, (System.currentTimeMillis() - start) / 1000.0));
-                    return Optional.of(contentsPath);
-                } else {
-                    log.log(Level.SEVERE, "Downloaded URL '" + this.url + "' not found, returning error");
-                    return Optional.empty();
-                }
+    private Optional<File> writeContent(File downloadDir, ClassicHttpResponse resp, File contentsPath, long start) throws IOException {
+        try (var in = resp.getEntity().getContent()) {
```
bjorncs: FYI `getEntity()` will return `null` for responses without body

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
hmusum: Done, thanks

config-proxy/src/main/java/com/yahoo/vespa/config/proxy/filedistribution/UrlDownloader.java
hmusum: Handled now

bjorncs (approved):

PR #34254, opened 2025-06-12
more decimals in tensor output

andreer:
This should ensure we have enough to round-trip a float (and will make sure all our existing examples in the tensor playground look nice in the refreshed UI)


havardpe:
I do not think this will change what you think it will change...

andreer:
You're right, after testing it doesn't work. I'll fix the examples instead.

PR #34255, opened 2025-06-12
remove casts no longer needed

arnej27959:
@vekterli please review - probably leftovers from when we used vespalib::string


vekterli (approved):
:+1:

PR #34260, opened 2025-06-12
bump jllama version to 4.1.0

arnej27959:
@glebashnik please review
@toregge FYI


glebashnik (approved):
Looks good!
I updated and ran disabled LLM tests on larger model locally.

PR #34262, opened 2025-06-12
use gtest for vbench unit tests

havardpe:
@toregge please review

### DIFF
--- a/vbench/src/tests/timer/timer_test.cpp
+++ b/vbench/src/tests/timer/timer_test.cpp
@@ -1,17 +1,17 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
-#include <vespa/vespalib/testkit/test_kit.h>
+#include <vespa/vespalib/gtest/gtest.h>
 #include <vbench/test/all.h>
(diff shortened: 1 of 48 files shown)


vbench/src/tests/timer/timer_test.cpp
```diff
@@ -1,17 +1,17 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
-#include <vespa/vespalib/testkit/test_kit.h>
+#include <vespa/vespalib/gtest/gtest.h>
 #include <vbench/test/all.h>
 #include <vespa/vespalib/util/time.h>
 
 using namespace vbench;
 
-IGNORE_TEST("timer") {
+TEST(TimerTest, ignored_timer_test) {
```
toregge, line 8: Consider `DISABLED_` prefix instead of `ignored_` prefix if test should be disabled.

toregge (approved):

vbench/src/tests/timer/timer_test.cpp
havardpe: I do not want to disable it

PR #34275, opened 2025-06-13
Implement downloading private huggingface models

onurkaracali:


model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
```diff
@@ -0,0 +1,78 @@
...
+            return Paths.get(file.getAbsolutePath());
+        };
+
+
+    public String getModelPathResolvingIfNecessary(ModelReference modelReference) {
+        if (isModelDownloadRequired(modelReference)) {
+            return resolveModelAndReturnPath(modelReference);
+        }
+
+        return modelReference.toString();
```
evgiz: To string will return the serialized version for the binary blob (config server -> config proxy communication). What we want is `ModelReference::value`.

config/src/main/java/com/yahoo/vespa/config/UrlDownloader.java
```diff
@@ -66,24 +67,40 @@ public boolean isValid() {
...
+        }
 
-       if (request.checkReturnTypes("s")) {
-           return new File(request.returnValues().get(0).asString());
-       } else if (! request.isError()) {
-           throw new RuntimeException("Invalid response: " + request.returnValues());
-       } else {
-           throw new RuntimeException("Wait for " + urlReference + " failed: " + request.errorMessage() + " (" + request.errorCode() + ")");
-       }
+        public static DownloadOptions withAuthToken(String authToken ) {
```
evgiz: I've not seen static metods with the `with` prefix elsewhere, mostly used for instance builder patterns. Perhaps other naming?

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
evgiz: Actually, seems like it *also* returns the path string if resolved.... Still prefer `::value`, double behaviors like this can be confusing.

onurkaracali: Did a small refactor here, better to just return Path using value. HF component will decide on to use as a Path or String.

config/src/main/java/com/yahoo/vespa/config/UrlDownloader.java
onurkaracali: Updated naming.

evgiz (dismissed):
LGTM, let's wait for @bjorncs to also take a look

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelper.java
```diff
@@ -0,0 +1,10 @@
+package ai.vespa.embedding.support;
```
bjorncs: Consider moving all classes to `ai.vespa.embedding`. I prefer not having too many small Java packages. Another benefit is that helper classes only used in `ai.vespa.embedding` can reduce its visibility to package-private, which in my opinion improves code quality (easier to understand the intended scope of a type). In this case we will still have to make these two types `public` as the hugging face embedder resides in a separate package, unless we move the Huggingface embedder to the parent package.

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
```diff
@@ -0,0 +1,78 @@
+package ai.vespa.embedding.support;
```
bjorncs: Remember to add following to all new Java files / types:

1. Copyright
2. A short class description (unless the type is super trivial)
3. Javadoc author tag

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
```diff
@@ -0,0 +1,78 @@
...
+                    urlReference,
+                    downloadOptions,
+                    MODEL_DOWNLOAD_TIMEOUT
+            );
+
+            return Paths.get(file.getAbsolutePath());
+        };
+
+
+    public Path getModelPathResolvingIfNecessary(ModelReference modelReference) {
```
bjorncs: Use `@Override` annotation whenever implementing or overriding a method from parent type.

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+    // For test purposes
+    ModelPathHelperImpl(Secrets secrets, ModelResolverFunction modelResolverFunction) {
+        this.secrets = secrets;
+        this.modelResolverFunction = modelResolverFunction;
+    }
+
+    private ModelResolverFunction defaultModelResolverFunction =
+         (urlReference, downloadOptions) -> {
+            UrlDownloader urlDownloader = new UrlDownloader();
```
bjorncs: Either reuse the `UrlDownloader` across invocations or wrap in try-finally to ensure that any resources like client socket is released immediately after the `waitFor` invocation is completed.

bjorncs (changes requested):

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
onurkaracali: Addressed.

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
onurkaracali: Addressed.

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelperImpl.java
onurkaracali: Addressed.

model-integration/src/main/java/ai/vespa/embedding/support/ModelPathHelper.java
onurkaracali: Addressed.

bjorncs (approved):
LGTM 👍

PR #34284, opened 2025-06-16
Revert "Extend model reference configuration to support delayed authenticated url resolve"

evgiz:
Reverts vespa-engine/vespa#34233


onurkaracali (dismissed):

hmusum:
I think you need add default implementations of `secretRef()` and `unresolved()`, since they are used by models already out there, which depends on these methods existing

hmusum (approved):

PR #34289, opened 2025-06-17
Move tracking of max attribute address space usage from

toregge:
ResourceUsageTracker to new AttributeUsageNotifier class.

@vekterli : please review


vekterli (approved):
:+1:

PR #34294, opened 2025-06-17
Bratseth/ignore components

bratseth:
@glebashnik context in https://github.com/vespa-engine/vespa/pull/33976


indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/expressions/Components.java
```diff
@@ -0,0 +1,189 @@
...
+        public boolean contains(String componentId) { return true; }
+
+        @Override
+        public TYPE get(String componentId) {
+            return failingComponent("Components can not be created in this environment");
+        }
+
+    }
+
+    /** A collection of components of a single type, of which one is selected. */
```
glebashnik: This class doesn't represent a collection of components, rather:
```suggestion
    /** Selected component of a specific type. */
```

glebashnik (dismissed):
Looks good.
One nitpick in the comment.

PR #34301, opened 2025-06-18
chore: extract deployment specific flags into a trait

esolitos:
## What

* Remove the global flags (list below) and make a new generic structure to provide such flags in each command:
    * `-a / --application`
    * `-c / --cluster`
    * `-i / --instance`
    * `-t / --target`
    * `-z / --zone`
* Add some tests where regressions where found after manual testing

## Why

* Only use flags that are actually needed by each command.


esolitos:
@evgiz / @ean Please install this locally and use it for a while, if you find regressions I'll add more tests. 😊

esolitos:
(Just rebased)

sarzhann (dismissed):
Let's ship it and see if it will cause some system tests to fail (as I faced before with Golang Cleanup)

esolitos:
Resolved merge conflicts.

sarzhann (approved):

PR #34311, opened 2025-06-19
make rendezvous destructible

havardpe:
... to be used by Nexus

@toregge please review
@arnej27959 FYI


toregge (approved):

havardpe:
the destroy is a bit too strong right now (back in time)

PR #34322, opened 2025-06-20
chore(doc-v1-api): increase window size update rate in InstrumentedThrottlePolicy

bjorncs:
Set the resize rate to 2 to increase the window size update frequency. This change aims to improve the responsiveness of the
throttle policy by adjusting the rate at which the window size is updated.


vekterli (approved):
:+1: fascinating...

PR #34331, opened 2025-06-23
Add relative max attribute address space usage with limit to Resource…

toregge:
…UsageState.

@vekterli : please review


vekterli (approved):
:+1:

PR #34334, opened 2025-06-23
add bounds checking for onGetX(size_t index)

arnej27959:
For onGetInteger, onGetFloat and onGetString:
When accessing out of bounds (typically index 0 for an empty array) you would get a core dump.  Also, for array inside array you would access outer[index].inner[index] which does not make sense, use API without index for the inner result node.

@toregge please review
@bjorncs @glebashnik FYI

### DIFF
--- a/searchlib/src/vespa/searchlib/expression/resultvector.h
+++ b/searchlib/src/vespa/searchlib/expression/resultvector.h
@@ -101,9 +101,15 @@ private:
     void visitMembers(vespalib::ObjectVisitor &visitor) const override { ::visit(visitor, "Vector", _result); }
     size_t onSize() const override { return _result.size(); }
     const vespalib::Identifiable::RuntimeClass & getBaseClass() const override { return B::_RTClass; }
-    int64_t onGetInteger(size_t index) const override { return _result[index].getInteger(index); }
-    double onGetFloat(size_t index)    const override { return _result[index].getFloat(index); }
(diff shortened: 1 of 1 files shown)


toregge (approved):

searchlib/src/vespa/searchlib/expression/resultvector.h
```diff
@@ -101,9 +101,15 @@ class ResultNodeVectorT : public ResultNodeVector
     void visitMembers(vespalib::ObjectVisitor &visitor) const override { ::visit(visitor, "Vector", _result); }
     size_t onSize() const override { return _result.size(); }
     const vespalib::Identifiable::RuntimeClass & getBaseClass() const override { return B::_RTClass; }
-    int64_t onGetInteger(size_t index) const override { return _result[index].getInteger(index); }
-    double onGetFloat(size_t index)    const override { return _result[index].getFloat(index); }
-    ConstBufferRef onGetString(size_t index, BufferRef buf) const override { return  _result[index].getString(index, buf); }
+    int64_t onGetInteger(size_t index) const override {
+        return index < _result.size() ? _result[index].getInteger() : 0L;
```
bjorncs, line 105: We should pick default values that are consistent with other default values in context of grouping. See examples in https://github.com/vespa-engine/system-test/blob/master/tests/search/grouping_adv/grouping_base.rb#L282-L325

arnej27959: good catch Bjørn, will make another PR for that

PR #34344, opened 2025-06-24
minimal change to allow "equiv()" inside near()

arnej27959:
@havardpe or @bratseth please review


arnej27959:
turns out this "just works" (TM) when I tried it.

bratseth (approved):
Nice!

PR #34350, opened 2025-06-25
Arnej/run integration tests as such

arnej27959:
@gjoranv please review
@bjorncs @bratseth FYI

note: follow-up PR is in progress, splitting tests in container-core

### DIFF
--- a/config-model/pom.xml
+++ b/config-model/pom.xml
@@ -341,6 +341,7 @@
               <classesDirectory>target/test-classes</classesDirectory>
               <outputDirectory>tmp/</outputDirectory>
               <finalName>simpleplugin</finalName>
+              <classifier>simpleplugin</classifier>
               <includes>
                 <include>**/test/SimplePlugin.class</include>
                 <include>**/test/SimpleService.class</include>
--- a/container-core/pom.xml
+++ b/container-core/pom.xml
@@ -461,6 +461,20 @@
         </executions>
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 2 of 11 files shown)


config-model/pom.xml
```diff
@@ -341,6 +341,7 @@
               <classesDirectory>target/test-classes</classesDirectory>
               <outputDirectory>tmp/</outputDirectory>
               <finalName>simpleplugin</finalName>
+              <classifier>simpleplugin</classifier>
```
gjoranv, line 344: Why is classifier added here?

container-core/pom.xml
```diff
@@ -461,6 +461,20 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-test</id>
+            <goals>
+              <goal>test</goal>
+            </goals>
+            <phase>integration-test</phase>
```
gjoranv, line 473: Why is running surefire moved to a later phase here?

gjoranv (approved):
Great!

config-model/pom.xml
arnej27959: in some circumstances, other modules would pick simpleplugin.jar instead
of config-model.jar - leading to strange and confusing errors.

container-core/pom.xml
arnej27959: Many of these tests are really integration tests that require bundles to
be built.  We will rename those from FooTest to FooIT and use failsafe
plugin to run them, this is just a stopgap temporary solution.

PR #34362, opened 2025-06-26
Allow mirroring copr rpms manually, not only on schedule. MERGEOK

gjoranv:
- Add a 'dry-run' input param to skip the mirroring itself.

FYI: @glebashnik


arnej27959 (approved):
👍

PR #34368, opened 2025-06-27
Bjorncs/gguf embedder

bjorncs:

### DIFF
--- a/model-integration/src/main/java/ai/vespa/embedding/GgufEmbedder.java
+++ b/model-integration/src/main/java/ai/vespa/embedding/GgufEmbedder.java
@@ -27,6 +27,9 @@ public class GgufEmbedder extends AbstractComponent implements Embedder {
     private static final Logger log = Logger.getLogger(GgufEmbedder.class.getName());
     private final LlamaModel model;
     private final int maxPromptTokens;
+    private final String prependQuery;
+    private final String prependDocument;
+    private final boolean normalize;
 
     public static class Exception extends RuntimeException {
         public Exception(Throwable cause) { super(cause); }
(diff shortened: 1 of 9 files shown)


model-integration/src/main/java/ai/vespa/embedding/GgufEmbedder.java
```diff
@@ -100,10 +107,15 @@ public void deconstruct() {
     }
 
     /**
-     * Performs a naive truncation of the prompt to the maximum number of tokens.
+     * Add prefix to prompt if configured, then do naive truncation of the prompt to the maximum number of tokens.
      * Tokens are assumed to be independent and that the token sequence can be safely truncated at any position.
      */
-    private String truncatePrompt(String text) {
+    private String prependAndTruncatePrompt(String text, Context context) {
+        if (!prependQuery.isBlank() && context.getDestination().startsWith("query")) {
```
glebashnik, line 114: Is the "query" prefix for destination hardcoded somewhere?
Consider defining it as a constant used in both places.
This will help to find references to this prefix.

thomasht86 (approved):
Nice!

model-integration/src/main/java/ai/vespa/embedding/Normalize.java
```diff
@@ -0,0 +1,33 @@
...
+import com.yahoo.tensor.Tensor;
+import com.yahoo.tensor.TensorAddress;
+import com.yahoo.tensor.TensorType;
+
+/**
+ * A utility class for normalizing embeddings.
+ *
+ * @author bjorncs
+ */
+class Normalize {
```
glebashnik, line 13: Is it some special type of normalization useful for embeddings only or general tensor normalization. If we don't have it already in tensor framework consider adding it there.

Also consider better naming, e.g. Normalizer, TensorNormalizer

glebashnik (changes requested):
Some minor comments.

model-integration/src/main/java/ai/vespa/embedding/GgufEmbedder.java
bjorncs: I don't know actually - I've just followed the pattern from the existing `HuggingFaceEmbedder.java`-embedder.

model-integration/src/main/java/ai/vespa/embedding/Normalize.java
bjorncs: It's specific for embedders, which is why it's in the embedding Java package and is package-private. I'll rename it in a follow-up PR.

bjorncs: FYI renamed in [0ecb54e](https://github.com/vespa-engine/vespa/pull/34371/commits/0ecb54efbbbbdef98da41f64b337098e8936f2ae)

PR #34371, opened 2025-06-27
Bjorncs/support onnx model external data files

bjorncs:
FYI @evgiz @onurkaracali


thomasht86:
LGTM!
Great that you use uv.
Consider whether inline script metadata for the python script would be sufficient (would allow you to remove pyproject.toml and uv.lock-file) so may be better suited for single file. 
See https://docs.astral.sh/uv/guides/scripts/#improving-reproducibility.

thomasht86 (dismissed):

thomasht86 (approved):

bjorncs:
Addresses #28761.

glebashnik (approved):
Looks good!

PR #34372, opened 2025-06-28
Remove unused method searchMmapAdvise

hmusum:
(unit test removed, since this is the default config setting now)


vekterli (approved):
:+1:

PR #34376, opened 2025-06-29
get MD5 from bytes directly

arnej27959:
the old code would convert from binary data to string, back to bytes, and then perform MD5.  This gave different results depending on LC_CTYPE settings.  Use the raw bytes directly to get the same MD5 as "md5sum ./src/test/apps/content/foo/bar/test.jar"

NOTE: this will (obviously) change the MD5 checksums generated - it seems likely to me that it could have bad effects, so we need to discuss it and go over scenarios before merging.  But the current situation is bad in any case.

### DIFF
--- a/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
+++ b/config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
@@ -207,7 +207,7 @@ public class FilesApplicationFile extends ApplicationFile {
             if (file.isDirectory()) {
                 return new MetaData(ContentStatusNew, "");
             } else {
-                return new MetaData(ContentStatusNew, ConfigUtils.getMd5(IOUtils.readAll(createReader())));
+                return new MetaData(ContentStatusNew, ConfigUtils.getMd5(IOUtils.readFileBytes(file)));
             }
         } catch (IOException | IllegalArgumentException e) {
             return null;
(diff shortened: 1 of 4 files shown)


hmusum:
Seems to me that this is only used in HTTP responses for application and session content, which I think is OK to change/fix. I'll take a closer look tomorrow.

bratseth (approved):
👍

hmusum (approved):

hmusum:
This looks good to go, there should be no issue to change this

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
```diff
@@ -207,7 +207,7 @@ public MetaData getMetaData() {
             if (file.isDirectory()) {
                 return new MetaData(ContentStatusNew, "");
             } else {
-                return new MetaData(ContentStatusNew, ConfigUtils.getMd5(IOUtils.readAll(createReader())));
+                return new MetaData(ContentStatusNew, ConfigUtils.getMd5(IOUtils.readFileBytes(file)));
```
bjorncs, line 210: FYI use can use `Files.readAllBytes()` - no need for our `IOUtils` here.

bjorncs (approved):

gjoranv (approved):

config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationFile.java
arnej27959: Not quite - we have a `File` but `Files.readAllBytes()` want a `Path`

bjorncs: https://docs.oracle.com/javase/8/docs/api/java/io/File.html#toPath--

PR #34378, opened 2025-06-30
sync rpmbuild setup

arnej27959:
* use maven 3.9.9 via wrapper
* run mvn test as part of check

@toregge please review

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -330,9 +330,9 @@ export PATH="%{_prefix}-deps/bin:$PATH"
 
 %if 0%{?_use_mvn_wrapper}
 mvn -B wrapper:wrapper -Dmaven=3.9.9 -N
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -329,10 +329,13 @@ export FACTORY_VESPA_VERSION=%{version}
 export PATH="%{_prefix}-deps/bin:$PATH"
 
 %if 0%{?_use_mvn_wrapper}
-mvn -B wrapper:wrapper -Dmaven=3.8.8 -N
+mvn -B wrapper:wrapper -Dmaven=3.9.9 -N
+%define _mvn_cmd $(pwd)/mvnw
```
toregge: Consider using `%global` instead of `%define`, to avoid scope issue

toregge (approved):

PR #34380, opened 2025-06-30
Implement ACORN-1 in HNSW Search

boeker:
This is a proposed implementation of a heuristics based on [ACORN-1](https://arxiv.org/abs/2403.04871) in the filtered HNSW search.

The rank-profile parameter `acorn-one-threshold` and the query equivalent `ranking.matching.acornOneThreshold` determine the index property `vespa.matching.nns.acorn_one_upper_limit`, which controls when ACORN-1 is being used. Currently, this defaults to `0.0`, which means that ACORN-1 is disabled by default. To enable it, experiments suggest that it should be set to `0.4`, i.e., it is being used when at most 40% of documents remain after filtering. Additionally, one should consider adapting `vespa.matching.global_filter.lower_limit`/`approximateThreshold`  from the default `0.05` to really make use of these heuristics. The 5% mark seems to be where one benefits most from these heuristics.

Another parameter pair`acorn-one-exploration`/`ranking.matching.acornOneExploration` determines the index property `vespa.matching.nns.acorn_one_exploration`, which controls how aggressively three-hop neighborhoods are being explored in addition to two-hop neighborhoods. One can set this to `0.0` to not explore three-hop neighborhoods and to `1.0` to explore all three-hop neighborhoods (if deemed necessary). Currently, it defaults to `0.01`, which means that three-hop neighborhoods are explored only very reluctantly.


arnej27959 (dismissed):
all looks good to me!

arnej27959 (approved):
went through this again now and it looks good to me 👍

PR #34381, opened 2025-06-30
Implement Adaptive Beam Search in HNSW Search

boeker:
This is a proposed implementation of [adaptive beam search](https://arxiv.org/abs/2505.15636) in the HNSW search, which is an early termination criterion that usually performs better (in terms of recall per visited nodes/response time) than just searching for additional hits.

Unlike the `hnsw.exploreAdditionalHits` annotation, adaptive beam search is controlled by the rank-profile parameter `adaptive-beam-search-slack` and the query equivalent `ranking.matching.adaptiveBeamSearchSlack`. These determine the index property `vespa.matching.nns.adaptive_beam_search_slack`. Currently, this defaults to `0.0`, which means that adaptive beam search is not used. To enable it, one has to set it to a value of at most `1.0`.


searchlib/src/vespa/searchlib/tensor/hnsw_index.cpp
```diff
@@ -388,7 +388,7 @@ HnswIndex<type>::search_layer_helper(const BoundDistanceFunction &df, uint32_t n
 
     while (!candidates.empty()) {
         auto cand = candidates.top();
-        if (cand.distance > (1.0 + adaptive_beam_search_slack) * limit_dist) {
+        if (cand.distance > (1.0 + exploration_slack) * limit_dist) {
```
arnej27959, line 391: I think we have to rewrite slightly to handle the corner case that `limit_dist`can be negative, but I’ll make a separate PR for that.

arnej27959 (approved):
Looks great 👍

PR #34385, opened 2025-07-01
Compute tensor rendering options from request once

vekterli:
@bjorncs please review

Previously did so twice for every rendered document. The request does not change between documents, so neither does the tensor rendering options.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
@@ -709,6 +709,7 @@ public final class DocumentV1ApiHandler extends AbstractRequestHandler {
         private final AtomicLong documentsWritten = new AtomicLong();
         private final AtomicLong documentsFlushed = new AtomicLong();
         private final AtomicLong documentsAcked = new AtomicLong();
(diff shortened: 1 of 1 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -742,6 +744,25 @@ static JsonResponse create(HttpRequest request, String message, ResponseHandler
             return response;
         }
 
+        private static JsonFormat.EncodeOptions createTensorOptionsFromRequest(HttpRequest request) {
+            // TODO: Flip default on Vespa 9 to "short-value"
+            String format = "short";
+            if (request != null && request.parameters().containsKey("format.tensors")) {
```
bjorncs, line 750: Is `request` ever `null` here?

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
vekterli: Seemingly so, transitively via `DocumentV1ApiHandler.handleFeedOperation()` which explicitly passes `null` as the request. This all happens in a context where a `HttpRequest` is available, but it is not useful for response rendering there.

PR #34393, opened 2025-07-02
fix(grouping-parser): enable unicode support

bjorncs:
Allow unicode string literals, e.g. in regex for filter expression.


vekterli (approved):
:+1:

PR #34398, opened 2025-07-02
feat(indexing-language): don't annotate if text is likely binary data

bjorncs:
Use number of replacement characters as proxy


arnej27959 (dismissed):
👍

bjorncs:
Ignore the branch name...

bratseth:
Shouldn't we throw an exception here?

bjorncs:
> Shouldn't we throw an exception here?

Perhaps in the future - for now we believe it's too risky (e.g. break reindexing of already indexed documents).

vekterli (approved):
:+1:

PR #34403, opened 2025-07-03
Better jvm option for metrics proxy

glebashnik:

### DIFF
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
@@ -39,7 +39,6 @@ import com.yahoo.vespa.model.container.component.SystemBindingPattern;
 import java.nio.file.Path;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.logging.Logger;
--- a/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
+++ b/config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
@@ -27,6 +27,7 @@ import com.yahoo.config.provision.ApplicationId;
 import com.yahoo.config.provision.Zone;
 import com.yahoo.container.jdisc.ThreadedHttpRequestHandler;
 import com.yahoo.osgi.provider.model.ComponentModel;
+import com.yahoo.search.config.QrStartConfig;
 import com.yahoo.vespa.model.admin.Admin;
 import com.yahoo.vespa.model.admin.monitoring.MetricsConsumer;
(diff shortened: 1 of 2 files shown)


config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
```diff
@@ -184,7 +185,16 @@ public void getConfig(ApplicationDimensionsConfig.Builder builder) {
...
 
-
+    public void getConfig(QrStartConfig.Builder builder) {
+        builder.jvm
+                .verbosegc(false)
+                .availableProcessors(1)
+                .minHeapsize(256)
+                .heapsize(256)
+                .heapSizeAsPercentageOfPhysicalMemory(0)
+                .gcopts(G1GC);
```
bjorncs: Wire in `gcopts` from feature flag

config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
```diff
@@ -184,7 +185,16 @@ public void getConfig(ApplicationDimensionsConfig.Builder builder) {
         }
     }
 
-
+    public void getConfig(QrStartConfig.Builder builder) {
```
bjorncs, line 189: Add `@Override`

bjorncs (changes requested):

config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
glebashnik: Added

config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
glebashnik: Calling super method that does it.

config-model/src/main/java/com/yahoo/vespa/model/admin/metricsproxy/MetricsProxyContainerCluster.java
```diff
@@ -185,14 +185,13 @@ public void getConfig(ApplicationDimensionsConfig.Builder builder) {
...
     public void getConfig(QrStartConfig.Builder builder) {
+        super.getConfig(builder);
+        
         builder.jvm
-                .verbosegc(false)
-                .availableProcessors(1)
                 .minHeapsize(256)
-                .heapsize(256)
-                .heapSizeAsPercentageOfPhysicalMemory(0)
-                .gcopts(G1GC);
```
bjorncs: Consider adding a unit test as the configuration now is inherited from parent class.

glebashnik: Makes sense.
Added unit test, check it out!

bjorncs (approved):

PR #34406, opened 2025-07-03
Log unknown vespa version with info level

hmusum:
After feedback in https://github.com/vespa-engine/vespa/issues/34397


arnej27959 (approved):
looks good to me, worst case we may need to limit logging here somewhat.

PR #34414, opened 2025-07-07
Add experimental streaming JSONL output to the Document V1 API handler

vekterli:
@bjorncs please review. The majority of added lines is comments or tests. No testing of actual wiring from the Document V1 handler yet, since the actual semantics for how to enable JSONL remain TBD. Changes to the existing `DocumentV1ApiHandler` code are intentionally kept to a minimum—more stuff can be factored out separately later.

This commit adds a streaming JSONL response writer where puts (i.e. document instances) and removes are written in the Vespa JSONL feed format. Since the output also contains other non-put/remove entries, this is a strict _superset_ of the feed format, and clients must take that into account.

One major difference from the "restful" JSON responses (streaming or not) is that continuation tokens are emitted inline as they are updated by the visitor session, rather than being emitted once as part of the response epilogue. This means a client can continuously observe updated token values and simply remember the last one they have observed. Visiting can then restart from that point with minimal loss of progress.

The actual structure/contents of the emitted JSONL lines are not yet set in stone, and this is therefore currently to be considered an entirely experimental feature that should not yet be used in production. Correspondingly, enabling this output format requires the use of an undocumented, magic HTTP parameter, which will **stop working** once the JSONL output format is set in stone.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/BufferedContentChannelResponseWriter.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/BufferedContentChannelResponseWriter.java
@@ -1,0 +1,89 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.document.restapi.resource;
+
+import com.yahoo.jdisc.Response;
+import com.yahoo.jdisc.handler.BufferedContentChannel;
+import com.yahoo.jdisc.handler.CompletionHandler;
+import com.yahoo.jdisc.handler.ContentChannel;
+import com.yahoo.jdisc.handler.ResponseHandler;
+import com.yahoo.vespa.http.server.Headers;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.logging.Logger;
+
+import static java.util.logging.Level.FINE;
+import static java.util.logging.Level.WARNING;
+
+/**
+ * Abstraction around buffered/streaming writing to a {@link ResponseHandler}.
+ */
+class BufferedContentChannelResponseWriter implements ResponseWriter {
+
+    private static final Logger log = Logger.getLogger(BufferedContentChannelResponseWriter.class.getName());
+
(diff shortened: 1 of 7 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/TraceJsonRenderer.java
```diff
@@ -0,0 +1,29 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.yahoo.messagebus.TraceNode;
+
+import java.io.IOException;
+
+/**
+ * Simple wrapper for rendering trace trees as recursive JSON structures.
+ */
+class TraceJsonRenderer {
+
```
bjorncs, line 13: Consider adding a `private` constructor if class should not be instantiated

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/BufferedContentChannelResponseWriter.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        buffer.write(buf, handler);
+    }
+
+    @Override
+    public void close() throws IOException {
+        synchronized (lock) {
+            try {
+                if (channel == null) {
+                    log.log(WARNING, "Close called before response was committed, in " + getClass().getName());
```
bjorncs, line 78: Why include the class name in message?
Is this check intended as an assert?

bjorncs (dismissed):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/BufferedContentChannelResponseWriter.java
vekterli: This method is basically a transplant of `DocumentV1ApiHandler.JsonResponse.close()`, so this has also been carried over. Agreed that it seems more like an invariant violation than anything else.

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/TraceJsonRenderer.java
vekterli: Done

bjorncs (approved):

PR #34420, opened 2025-07-08
Use preserialized JSON field name strings for Document V1 output

vekterli:
@bjorncs please review

Avoids having to do the same UTF-8 encoding/character escaping work over and over for every static field name string written.


bjorncs (approved):
Clever 👍

PR #34421, opened 2025-07-08
Clean up archaic `fn(void)` empty parameter list syntax

vekterli:
@toregge please review

No changes in semantics.


toregge (approved):
:+1:

PR #34424, opened 2025-07-08
Factor out `JsonResponse` class from `DocumentV1ApiHandler`

vekterli:
@bjorncs please review. Chipping away at the overall heft of the handler class. Not intended to change any semantics.

* Factor out other dependencies needed to make it work at a package-level.
* Disambiguate `JsonResponse` factory methods.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/Group.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/Group.java
@@ -1,0 +1,68 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.document.restapi.resource;
+
+import com.yahoo.text.Text;
+
+import java.util.Objects;
+
+/**
+ * Represents visiting constrained over a document subset that shares the same
+ * location affinity ('number' or 'group' value).
+ *
+ * @author Jon Marius Venstad
+ */
+class Group {
+
+    private final String docIdPart;
+    private final String selection;
+
+    private Group(String docIdPart, String selection) {
+        this.docIdPart = docIdPart;
+        this.selection = selection;
+    }
+
+    public static Group of(long value) {
+        String stringValue = Long.toUnsignedString(value);
+        return new Group("n=" + stringValue, "id.user==" + stringValue);
+    }
+
+    public static Group of(String value) {
+        Text.validateTextString(value)
+                .ifPresent(codePoint -> {
+                    throw new IllegalArgumentException(String.format("Illegal code point U%04X in group", codePoint));
+                });
+
+        return new Group("g=" + value, "id.group=='" + value.replaceAll("'", "\\\\'") + "'");
+    }
+
+    public String docIdPart() {
+        return docIdPart;
+    }
+
+    public String selection() {
+        return selection;
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/Group.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Objects;
+
+/**
+ * Represents visiting constrained over a document subset that shares the same
+ * location affinity ('number' or 'group' value).
+ *
+ * @author Jon Marius Venstad
+ */
+class Group {
```
bjorncs: Consider rewriting as a Java `records`. You'll get rid of the getters, `hashCode()`, `equals()` and `toString()`. The only downside afaik is a non-private constructor, but that does not really matter here.

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/JsonResponse.java
```diff
@@ -0,0 +1,334 @@
...
+        json.writeFieldName(JsonNames.ID);
+        json.writeString(id.toString());
+    }
+
+    @Override
+    public synchronized void writeTrace(Trace trace) throws IOException {
+        if (trace != null && !trace.getRoot().isEmpty()) {
+            TraceJsonRenderer.writeTrace(json, trace.getRoot());
+        }
+    }
```
bjorncs, line 186: Consider returning `this` so the write methods can be chained

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/JsonResponse.java
```diff
@@ -0,0 +1,334 @@
...
+        new JsonWriter(json, tensorOptions()).writeFields(document);
+    }
+
+    @Override
+    public synchronized void writeDocumentsArrayStart() throws IOException {
+        json.writeFieldName(JsonNames.DOCUMENTS);
+        json.writeStartArray();
+    }
+
+    private interface DocumentWriter {
```
bjorncs, line 211: Consider annotating the interface with `@FunctionalInterface` to make its intention clear.

bjorncs (dismissed):
Only 1.5k lines left in `DocumentV1ApiHandler`... 😅 👍

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/JsonResponse.java
vekterli: Done

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/Group.java
vekterli: Done

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/JsonResponse.java
vekterli: Will consider for a follow-up. These methods are from the `StreamableJsonResponse` interface, so changing the signature would also necessitate changing the JSONL response renderer at the same time.

PR #34439, opened 2025-07-10
Fix vespamalloc edge case in `realloc`

vekterli:
@toregge please review
@arnej27959 FYI

Code did not properly constrain the amount of memory copied across old and new allocations if the following held:

 1. the old allocation was at least as large as the limit set by `mallopt(M_MMAP_THRESHOLD)`, which means it got its own dedicated "off-heap" anomymous memory mapped buffer.
 2. the new allocation size is _smaller_ than the old allocation size.

This could lead to heap overwrites or—with a likelihood increasing with the size of the old buffer—triggering a `SIGSEGV` due to touching memory outside mapped arena blocks.

This commit ensures that memory copied is capped to the _minimum_ of the old and new allocation sizes.

This bug has never been observed in action for Vespa's own C++ binaries since `realloc` is used extremely rarely internally, but it can be triggered fairly reliably when `vespamalloc` is used via `LD_PRELOAD` for other processes that reallocate large buffers.

### DIFF
--- a/vespamalloc/src/tests/test1/new_test.cpp
+++ b/vespamalloc/src/tests/test1/new_test.cpp
@@ -266,6 +266,29 @@ TEST(NewTest, test_realloc_large_buffers) {
     EXPECT_EQ(1, mallopt(M_MMAP_THRESHOLD, 1_Gi));
 }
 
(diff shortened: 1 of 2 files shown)


toregge (approved):
:+1:

PR #34440, opened 2025-07-10
feat(indexing-language): don't annotate if text is likely binary data

bjorncs:
Skip tokenization if it's a reindexing operation with some heavily throttled logging. If not, fail hard with an exception.

Extended the previous PR with detection of reindexing operations (https://github.com/vespa-engine/vespa/pull/34398). The threshold are the same as before.

FYI @bratseth @hmusum

### DIFF
--- a/docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
+++ b/docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
@@ -237,7 +237,7 @@ public class DocumentScriptTestCase {
         docType.addField("myField", fieldValue.getDataType());
         Document doc = new Document(docType, "id:ns:myDocumentType::");
         doc.setFieldValue("myField", fieldValue.clone());
-        doc = newScript(docType).execute(fieldValuesFactory, doc, isReindexingOperation(input));
+        doc = newScript(docType).execute(fieldValuesFactory, doc, false);
         return doc.getFieldValue("myField");
     }
 
@@ -347,7 +347,7 @@ public class DocumentScriptTestCase {
     }
 
     private static Document execute(Document document) {
-        return newScript(document.getDataType()).execute(new FieldValuesFactory(), document, isReindexingOperation(input));
+        return newScript(document.getDataType()).execute(new FieldValuesFactory(), document, false);
     }
 
     private static DocumentUpdate execute(DocumentUpdate update) {
--- a/docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
+++ b/docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
@@ -237,7 +237,7 @@ public class DocumentScriptTestCase {
         docType.addField("myField", fieldValue.getDataType());
         Document doc = new Document(docType, "id:ns:myDocumentType::");
         doc.setFieldValue("myField", fieldValue.clone());
-        doc = newScript(docType).execute(fieldValuesFactory, doc);
+        doc = newScript(docType).execute(fieldValuesFactory, doc, isReindexingOperation(input));
         return doc.getFieldValue("myField");
     }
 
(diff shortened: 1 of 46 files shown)


bratseth:
Great!

The schema/il-script parsers (there are two right now) must also be extended to parse the new tokenization parameters (max-replacement-characters-ratio, and max-replacement-characters). You can find them by searching for "max-token-length".

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
```diff
@@ -146,4 +172,30 @@ private static void addAnnotationSpan(String input, SpanList parent, Token token
         }
     }
 
+    /**
+     * Use the ratio of Unicode replacement characters as heuristic if the text is likely representing binary data
+     *
+     * @see <a href="https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character">...</a>
+     * @throws IllegalArgumentException if text is likely binary data and not a reindexing operation
+     */
+    private boolean isLikelyBinaryData(String text, DocumentId docId, boolean isReindexingOperation) {
```
vekterli: Although documented, it might be a little bit surprising that a method prefixed with "is" can throw. Maybe `checkLikelyBinaryData` or something similar, which sounds less predicate-ey? Or pull the actual exception vs return value decision up to the caller, which would also remove the need for passing down the doc ID or reindexing state.

vekterli (dismissed):
:+1: with the caveat emptor that the docproc changes should be verified by someone rather more well traveled in those lands than me

docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
```diff
@@ -237,7 +237,7 @@ private static FieldValue processDocument(FieldValue fieldValue) {
         docType.addField("myField", fieldValue.getDataType());
         Document doc = new Document(docType, "id:ns:myDocumentType::");
         doc.setFieldValue("myField", fieldValue.clone());
-        doc = newScript(docType).execute(fieldValuesFactory, doc);
+        doc = newScript(docType).execute(fieldValuesFactory, doc, isReindexingOperation(input));
```
arnej27959:
```suggestion
        doc = newScript(docType).execute(fieldValuesFactory, doc, false);
```

docprocs/src/test/java/com/yahoo/docprocs/indexing/DocumentScriptTestCase.java
```diff
@@ -347,7 +347,7 @@ private static void assertFail(String expectedException, DocumentUpdate update)
     }
 
     private static Document execute(Document document) {
-        return newScript(document.getDataType()).execute(new FieldValuesFactory(), document);
+        return newScript(document.getDataType()).execute(new FieldValuesFactory(), document, isReindexingOperation(input));
```
arnej27959:
```suggestion
        return newScript(document.getDataType()).execute(new FieldValuesFactory(), document, false);
```

arnej27959 (dismissed):
a couple of test methods needs fixing, rest looks good to me

indexinglanguage/src/main/java/com/yahoo/vespa/indexinglanguage/linguistics/LinguisticsAnnotator.java
bjorncs: Producing an actionable error message requires the intermediate calculations. I'll rename the method and improve the comment section.

bjorncs:
I've resolved the requested changes from @arnej27959 / @vekterli and increased the default thresholds to more conservative values. 

@bratseth Imho these settings should rather be global instead of field specific. We can make it easier to override later - if so we should generalize the threshold names (not tie them to replacement characters only). For now the generic config override mechanism should suffice.

vekterli (approved):
:+1:

PR #34445, opened 2025-07-10
Bratseth/near segments

bratseth:
@arnej27959 the change that matters is in the last commit


bratseth:
@arnej27959 ping

arnej27959 (approved):
I guess this is ok.

bratseth:
Better than throwing them away.

PR #34447, opened 2025-07-11
More `NULL` -> `nullptr` cleanups

vekterli:
@toregge please review.

With this, the final obviously greppable old-school `NULL`s shall have been purged from the C++ code base (aside from some comments and use in string literals). A few `NULL`s still remain in a couple of C files, but not touching those yet since `nullptr` is a C23 feature.


toregge (approved):
👍

PR #34450, opened 2025-07-11
Test replace pom versions

gjoranv:
All BATS tests will be run whenever a shell script or a bats test is modified in this repo. Adding a new test does not require any other changes.

Each test added will help the AI write good tests with less effort from the developer.

Caveat: Changes to shell scripts with no file suffix will not be detected.

### DIFF
--- a/.buildkite/replace-vespa-version-in-poms.sh
+++ b/.buildkite/replace-vespa-version-in-poms.sh
@@ -4,14 +4,25 @@
 set -euo pipefail
 
 if [[ $# -ne 2 ]]; then
-    echo "Usage: $0 <Vespa version> <path>"
+    echo "Usage: $(basename "$0") <Vespa version> <path>"
     exit 1
 fi
 
 readonly VESPA_VERSION=$1
 readonly DIR=$2
 
-find -L "$DIR" -name "pom.xml" -exec sed -i \
+# Fail if DIR does not exist or is not a directory
+if [[ ! -d "$DIR" ]]; then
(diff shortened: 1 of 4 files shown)


hmusum (approved):

hmusum (changes requested):
Looks like this needs a minor update:


In /workspace/build/buildkite/vespaai/vespa-engine-vespa/.buildkite/replace-vespa-version-in-poms.sh line 31:
--
  | find -L "$DIR" -name "pom.xml" -exec $SED_INPLACE \
  | ^----------^ SC2086: Double quote to prevent globbing and word splitting.

gjoranv:
No, that's intentionally unquoted.

> The variable $SED_INPLACE is intentionally left unquoted in the find ... -exec command. This is necessary for the shell to perform word splitting.
> 
> 
> If SED_INPLACE is "sed -i ''", the shell splits it into three separate arguments: sed, -i, and '', which is the correct syntax for the sed command on macOS.
> If SED_INPLACE is "sed -i", it's split into sed and -i, which is correct for Linux.
> If you were to enclose $SED_INPLACE in double quotes ("$SED_INPLACE"), the shell would treat the entire string as a single command name (e.g., "sed -i") which would fail because no such command exists.

hmusum:
Sure, but the build fails, so I assume you need to do something to skip checking this then?

hmusum (approved):

esolitos:
Great addition!

PR #34455, opened 2025-07-12
Allow query profiles to set YQL grammar defaults

bratseth:
This is useful when there is a client that produces YQL statements containing userinput strings, as opposed to using one or a few canned YQL statements where only values varies. The disadvantage is that the YQL statements cannot be fully understood in isolation.


arnej27959 (dismissed):
Looks good to me

arnej27959 (approved):

PR #34466, opened 2025-07-15
Add debug option `--replica-override` to `vespa-get` for selecting specific node to retrieve documents from

johansolbakken:
This change enables effective debugging of content consistency issues in Vespa content clusters by allowing the retrieval of documents from specific replicas.

Usage example – retrieve from node with ID 2:
```shell
$ vespa-get --replica-override 2 "id:test:music::user-specific-id"
```

If the document does not exist or there is no node with the given ID, `vespa-get` will return "document not found".

### DIFF
--- a/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
+++ b/documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
@@ -195,19 +195,13 @@ abstract class RoutableFactories80 {
         return fieldSet.getSpec();
     }
 
-    private static DocapiCommon.DebugGetFromReplica toProtoDebugReplica(Integer nodeId) {
-        if (nodeId == null) {
-            return null;
-        }
+    private static DocapiCommon.DebugGetFromReplica toProtoDebugReplica(int nodeId) {
         return DocapiCommon.DebugGetFromReplica.newBuilder()
-                .setNodeId(nodeId)
-                .build();
+            .setNodeId(nodeId)
+            .build();
     }
 
-    private static Integer fromProtoDebugReplica(DocapiCommon.DebugGetFromReplica proto) {
-        if (proto == null) {
-            return null;
-        }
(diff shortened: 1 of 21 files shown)


documentapi/src/main/java/com/yahoo/documentapi/messagebus/protocol/RoutableFactories80.java
```diff
@@ -306,15 +306,27 @@ private static String fromProtoClusterState(DocapiCommon.ClusterState state) {
...
-                            .setFieldSet(toProtoFieldSet(apiMsg.getFieldSet()))
-                            .build())
-                .decoder(DocapiFeed.GetDocumentRequest.parser(), (protoMsg) ->
-                        new GetDocumentMessage(
-                                fromProtoDocId(protoMsg.getDocumentId()),
-                                fromProtoFieldSet(protoMsg.getFieldSet())))
+                            .setFieldSet(toProtoFieldSet(apiMsg.getFieldSet()));
+
+                        if (apiMsg.hasDebugReplicaNodeId()) {
+                            builder = builder.setDebugReplicaNodeId(apiMsg.getDebugReplicaNodeId());
```
vekterli: `builder = ...` can be omitted here since the `Builder` is just `this`-chaining for convenience

https://protobuf.dev/reference/java/java-generated/:
> Methods that modify the contents of a builder—including field setters—always return a reference to the builder (i.e. they `“return this;”`). This allows multiple method calls to be chained together in one line.

storage/src/vespa/storageapi/message/persistence.h
```diff
@@ -230,6 +231,10 @@ class GetCommand : public BucketInfoCommand {
         return api::LockingRequirements::Shared;
     }
 
+    void set_debug_replica_node_id(std::optional<uint16_t> node_id) noexcept { _debug_replica_node_id=node_id; }
```
vekterli: Nit: missing spaces around assignment op

vekterli (dismissed):
:+1:

vekterli (approved):
:+1:

PR #34467, opened 2025-07-15
Unit test prepare restart2 rpc handler.

toregge:
@vekterli : please review

### DIFF
--- a/searchcore/src/tests/proton/server/prepare_restart2_rpc_handler_test.cpp
+++ b/searchcore/src/tests/proton/server/prepare_restart2_rpc_handler_test.cpp
@@ -45,38 +45,38 @@ std::string as_string(const FRT_StringValue& v)
 /*
  * Base class that supports waiting for destruction.
  */
-class DestroyedFutureFactoryBase {
-    std::promise<void>       _destroyed_promise;
-    std::shared_future<void> _destroyed_future;
+class DestructionFutureFactoryBase {
+    std::promise<void>       _destruction_promise;
+    std::shared_future<void> _destruction_future;
 public:
-    DestroyedFutureFactoryBase();
-    virtual ~DestroyedFutureFactoryBase();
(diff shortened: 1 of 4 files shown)


searchcore/src/tests/proton/server/prepare_restart2_rpc_handler_test.cpp
```diff
@@ -0,0 +1,454 @@
...
+    }
+}
+
+void
+MyReturnHandler::HandleReturn()
+{
+    _returned = true;
+    auto& v = *_req->GetReturn();
+    ASSERT_EQ("bs", std::string(v.GetTypeString(), v.GetNumValues()));
+    _success = v[0]._intval8 == 1;
```
vekterli: Consider wrapping assignment rhs in parentheses

searchcore/src/tests/proton/server/prepare_restart2_rpc_handler_test.cpp
```diff
@@ -0,0 +1,454 @@
...
+/*
+ * Base class that supports waiting for destruction.
+ */
+class DestroyedFutureFactoryBase {
+    std::promise<void>       _destroyed_promise;
+    std::shared_future<void> _destroyed_future;
+public:
+    DestroyedFutureFactoryBase();
+    virtual ~DestroyedFutureFactoryBase();
+    [[nodiscard]] std::shared_future<void> get_destroyed_future() const noexcept { return _destroyed_future; }
```
vekterli: Strictly a personal preference, but I would find it more intuitive if these were called _destruction_ futures/promises instead of _destroyed_ futures, as to me the former implies something pending in the future(tm) whereas the latter implies the state of something in the present ("`get_destroyed_future()` returns a future whose state is somehow destroyed").

Also helps avoid potential confusion around the existing, separate notion of a _broken_ promise object (which _is_ a state in the present).

vekterli (dismissed):
:+1:

vekterli (approved):

PR #34475, opened 2025-07-16
VESPANG-967: add more info when refresh token cannot be persisted

sarzhann:

### DIFF
--- a/client/go/internal/cli/cmd/login.go
+++ b/client/go/internal/cli/cmd/login.go
@@ -18,88 +18,102 @@ import (
 // Use `expired` to run the login from other commands setup:
 // this will only affect the messages.
 func newLoginCmd(cli *CLI) *cobra.Command {
-	return &cobra.Command{
+	var useFileStorage bool
+	cmd := &cobra.Command{
 		Use:   "login",
 		Args:  cobra.NoArgs,
 		Short: "Authenticate Vespa CLI with Vespa Cloud control plane. This is preferred over api-key for interactive use",
 		Long: `Authenticate Vespa CLI with Vespa Cloud control plane. This is preferred over api-key for interactive use.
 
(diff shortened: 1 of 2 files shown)


client/go/internal/cli/cmd/login.go
```diff
@@ -87,6 +88,10 @@ This command runs a browser-based authentication flow for the Vespa Cloud contro
 			if err != nil {
 				// log the error but move on
 				cli.printWarning("Could not store the refresh token locally. You may need to login again once your access token expires")
```
frodelu: Add information about when access token refresh will be needed (30 minutes).

client/go/internal/cli/cmd/login.go
```diff
@@ -87,6 +88,10 @@ This command runs a browser-based authentication flow for the Vespa Cloud contro
...
+				cli.printWarning("To persist the refresh token using file storage (unencrypted), set VESPA_CLI_DUMMY_KEYRING=1 environment variable")
```
frodelu: Could we add explicit switch for this rather than environment variable? Scary if someone puts this in their `.bashrc` and forgets about it.

I would also perhaps give this a separate sentence on the security aspect instead of just saying "(unencrypted)" saying something like "Note: Storing the refresh token unencrypted on your file system means someone with access to this file can get unauthorized access to your application for the life of the refresh token (24 hours)".

...or something a bit shorter.

frodelu: Could we also link to some Vespa CLI documentation giving more information about the secret store implementation and debug advice?

frodelu (changes requested):
Would be great if we could do even more on the clarity front since the problem here is confusion from a usability standpoint.

frodelu (approved):

PR #34485, opened 2025-07-17
Magnus/lsp rank properties

Mangern:
rank-properties syntax was only parsed and displayed as strings before, this PR sends rank features from rank-properties to the rank expression parser, giving syntax highlighting and error checking.


johansolbakken (approved):
LGTM :+1:

PR #34507, opened 2025-07-21
Only allow the first failing thread to enter `mimalloc` error handler

vekterli:
@johansolbakken please review
@toregge FYI

When `malloc` starts failing in one thread due to process OOM conditions, it's highly likely that other threads will also fail around the same time (assuming their own thread-local free-lists aren't sufficient).

To avoid getting jumbled, interleaved output on `stderr` from a bunch of threads racing to the finish line of printing errors and triggering a process exit, only allow a single thread into the error handler. The rest shall obediently wait forever until the process is terminated by that one thread (where "forever" is expected to be on the order of microseconds in reality).

This also adds death tests for the existing error handler functionality. Testing of only allowing the first thread to enter is not included here since it's inherently a race condition (which results in either killing the process or hanging the thread, neither outcomes being the most test-friendly) and therefore hard to test deterministically.

### DIFF
--- a/vespalib/CMakeLists.txt
+++ b/vespalib/CMakeLists.txt
@@ -171,6 +171,7 @@ vespa_define_module(
     src/tests/util/generationhandler_stress
     src/tests/util/hamming
     src/tests/util/md5
+    src/tests/util/mimalloc
     src/tests/valgrind
     src/tests/visit_ranges
     src/tests/wakeup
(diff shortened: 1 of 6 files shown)


johansolbakken (dismissed):
lgtm 👍🏻

vekterli:
Looks like my incantations for getting the expected stack trace weren't sufficient on x64 (have only tested on Aarch64)... Might have to add some more indirections.

johansolbakken (approved):

PR #34510, opened 2025-07-22
Add `not()`, `and()` and `or()` filter predicates to grouping language

johansolbakken:
This PR adds `not()`, `and()` and `or()` filter predicates to grouping. This enables more complex grouping queries. 

Usage:

```text
all(group(stores) filter(and(or(regex("super.*", name), regex(".*store", name)), not(regex("oslo", location)))) each(summary())) 
```

The `MultiArgPredicateNode` class was added to facilitate predicates that take multiple arguments.


johansolbakken:
Related PR with system tests: https://github.com/vespa-engine/system-test/pull/4518

container-search/src/main/java/com/yahoo/search/grouping/request/AndPredicate.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Represents a logical conjunction (AND) of filter expressions used to match grouping elements.
+ *
+ * @author johsol
+ */
+@Beta
+public class AndPredicate extends FilterExpression {
+    private final List<FilterExpression> args;
+
+    public AndPredicate(List<FilterExpression> args) {
```
vekterli, line 18: The parser currently requires conjunctions to be at least binary. Consider adding a check that this holds also for `args` since these classes are `@PublicApi` and can be instantiated directly.

container-search/src/main/java/com/yahoo/search/grouping/request/OrPredicate.java
```diff
@@ -0,0 +1,37 @@
...
+ * Represents a logical disjunction (OR) of filter expressions used to match grouping elements.
+ *
+ * @author johsol
+ */
+@Beta
+public class OrPredicate extends FilterExpression {
+    private final List<FilterExpression> args;
+
+    public OrPredicate(List<FilterExpression> args) {
+        this.args = args;
```
vekterli, line 22: See arity comment for `AndPredicate`; also applies here.

integration/schema-language-server/language-server/src/main/ccc/grouping/GroupingParser.ccc
```diff
@@ -192,6 +193,7 @@ UNPARSED :
 // Production rules.
 //
 // --------------------------------------------------------------------------------
+/**/
```
vekterli: Nit: leftover comment?

searchlib/src/vespa/searchlib/expression/and_predicate_node.h
```diff
@@ -0,0 +1,29 @@
...
+ * Implements logical AND filter used in grouping expressions.
+ **/
+class AndPredicateNode : public MultiArgPredicateNode {
+public:
+    AndPredicateNode() noexcept;
+    ~AndPredicateNode() override;
+    AndPredicateNode* clone() const override { return new AndPredicateNode(*this); }
+
+    // for unit testing::
+    AndPredicateNode(const std::vector<FilterPredicateNode>& input);
```
vekterli: This does not seem to be used; remove?

Also, since `FilterPredicateNode` is an abstract class, this would have to be a vector of `unique_ptr<FilterPredicateNode>` (or some other polymorphic wrapper) and not _by value_. Otherwise you would in the general case get object slicing, though the caller will not compile when trying to slice to an abstract class.

I'm actually sort of surprised that the `MultiArgPredicateNode` ctor definition compiles. I'm not sure what the value is in allowing a loop over objects with inherently incomplete vtables... 🤔

searchlib/src/vespa/searchlib/expression/multi_arg_predicate_node.h
```diff
@@ -0,0 +1,31 @@
...
+ **/
+class MultiArgPredicateNode : public FilterPredicateNode {
+    using FilterPredicateNodeVector = std::vector<FilterPredicateNode::CP>;
+    FilterPredicateNodeVector _args;
+
+public:
+    MultiArgPredicateNode() noexcept;
+    ~MultiArgPredicateNode() override;
+
+    MultiArgPredicateNode(const std::vector<FilterPredicateNode>& input);
```
vekterli: See comment in `AndPredicateNode` on object slicing; consider removing.

searchlib/src/vespa/searchlib/expression/not_predicate_node.h
```diff
@@ -0,0 +1,33 @@
...
+class NotPredicateNode : public FilterPredicateNode {
+    FilterPredicateNode::CP _expression;
+
+public:
+    NotPredicateNode() noexcept;
+    ~NotPredicateNode();
+    NotPredicateNode* clone() const override { return new NotPredicateNode(*this); }
+
+    // for unit testing::
+    NotPredicateNode(const FilterPredicateNode& input);
```
vekterli: Prefer marking single-param constructors as `explicit` to avoid implicit conversions

searchlib/src/vespa/searchlib/expression/multi_arg_predicate_node.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "resultvector.h"
+
+namespace search::expression {
+
+using vespalib::Deserializer;
+using vespalib::Serializer;
+
+IMPLEMENT_IDENTIFIABLE_ABSTRACT_NS2(search, expression, MultiArgPredicateNode, FilterPredicateNode);
+
+MultiArgPredicateNode::MultiArgPredicateNode() noexcept : _args({}) {}
```
vekterli: Don't need an empty initializer list here, can just use the `_args()` default constructor

searchlib/src/vespa/searchlib/expression/or_predicate_node.h
```diff
@@ -0,0 +1,29 @@
...
+ * Implements logical OR filter used in grouping expressions.
+ **/
+class OrPredicateNode : public MultiArgPredicateNode {
+public:
+    OrPredicateNode() noexcept;
+    ~OrPredicateNode() override;
+    OrPredicateNode* clone() const override { return new OrPredicateNode(*this); }
+
+    // for unit testing::
+    OrPredicateNode(const std::vector<FilterPredicateNode>& input);
```
vekterli: See comment in `AndPredicateNode` on object slicing; consider removing.

searchlib/src/tests/expression/filter_predicate_node/filter_predicate_node_test.cpp
```diff
@@ -1,68 +1,143 @@
...
+template<typename... Nodes>
+std::unique_ptr<FilterPredicateNode> FilterPredicateNodesTest::make_and(Nodes... nodes) {
+    auto and_predicate = std::make_unique<AndPredicateNode>();
+    (and_predicate->args().push_back(std::move(nodes)), ...);
+    return and_predicate;
+}
+
+std::unique_ptr<ExpressionNode> FilterPredicateNodesTest::make_result(const std::string& value) {
+    return std::make_unique<ConstantNode>(
+        ResultNode::UP(std::make_unique<StringResultNode>(value)->clone()));
```
vekterli: Can this be reduced to `return std::make_unique<ConstantNode>(std::make_unique<StringResultNode>(value))`?

searchlib/src/vespa/searchlib/expression/not_predicate_node.cpp
```diff
@@ -0,0 +1,48 @@
...
+using vespalib::Serializer;
+
+IMPLEMENT_IDENTIFIABLE_NS2(search, expression, NotPredicateNode, FilterPredicateNode);
+
+NotPredicateNode::NotPredicateNode() noexcept = default;
+
+NotPredicateNode::~NotPredicateNode() = default;
+
+NotPredicateNode::NotPredicateNode(const FilterPredicateNode& input)
+  : _expression(input.clone())
```
vekterli: It looks like `CloneablePtr<T>` can be constructed with a `unique_ptr<T>`, so consider just passing in a `unique_ptr` directly to avoid a redundant `clone()` indirection. As a bonus, this simplifies `FilterPredicateNodesTest::make_not` by not having to take in a `unique_ptr` by const ref (which looks a bit peculiar).

searchlib/src/vespa/searchlib/expression/multi_arg_predicate_node.cpp
```diff
@@ -0,0 +1,37 @@
...
+void MultiArgPredicateNode::visitMembers(vespalib::ObjectVisitor& visitor) const {
+    visit(visitor, "args", _args);
+}
+
+void MultiArgPredicateNode::selectMembers(const vespalib::ObjectPredicate& predicate,
+                                       vespalib::ObjectOperation& operation) {
+    FilterPredicateNode::selectMembers(predicate, operation);
+    for(size_t i(0), m(_args.size()); i < m; i++) {
+        _args[i]->select(predicate, operation);
+    }
```
vekterli, line 53: Consider simplifying with `for (auto& arg : _args) {` as used elsewhere

searchlib/src/vespa/searchlib/expression/multi_arg_predicate_node.h
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Abstract base class for predicate nodes with multiple children.
+ **/
+class MultiArgPredicateNode : public FilterPredicateNode {
+    using FilterPredicateNodeVector = std::vector<FilterPredicateNode::CP>;
+    FilterPredicateNodeVector _args;
+
+public:
+    MultiArgPredicateNode() noexcept;
+    ~MultiArgPredicateNode() override;
```
vekterli, line 18: Since the `Identifiable` magic macro scaffolding of delights uses copy ctors and assignment operators, consider explicitly declaring the copy constructor and assignment opererator of this class and `= default`-ing them in the `.cpp` file to prevent the compiler having to do this inline every time (which puts more load on the linker).

This comment also applies to the other added node types.

searchlib/src/vespa/searchlib/expression/and_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+bool AndPredicateNode::allow(const DocId docId, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (!arg->allow(docId, rank)) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool AndPredicateNode::allow(const document::Document & doc, const HitRank rank) {
```
vekterli: Nit: `&` spacing (also applies to declaration)

searchlib/src/vespa/searchlib/expression/not_predicate_node.cpp
```diff
@@ -0,0 +1,48 @@
...
+Deserializer& NotPredicateNode::onDeserialize(Deserializer& is) {
+    is >> _expression;
+    return is;
+}
+
+bool NotPredicateNode::allow(DocId docId, HitRank rank) {
+    return !_expression->allow(docId, rank);
+}
+
+bool NotPredicateNode::allow(const document::Document & doc, HitRank rank) {
```
vekterli: Nit: `&` spacing (also applies to declaration)

searchlib/src/vespa/searchlib/expression/not_predicate_node.cpp
```diff
@@ -0,0 +1,48 @@
...
+    return !_expression->allow(doc, rank);
+}
+
+void NotPredicateNode::visitMembers(vespalib::ObjectVisitor& visitor) const {
+    visit(visitor, "expression", _expression);
+}
+
+void NotPredicateNode::selectMembers(const vespalib::ObjectPredicate& predicate,
+                                       vespalib::ObjectOperation& operation) {
```
vekterli: Nit: indentation alignment

searchlib/src/vespa/searchlib/expression/or_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+bool OrPredicateNode::allow(const DocId docId, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (arg->allow(docId, rank)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool OrPredicateNode::allow(const document::Document & doc, const HitRank rank) {
```
vekterli: Nit: `&` spacing (also applies to declaration)

searchlib/src/vespa/searchlib/expression/and_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return is;
+}
+
+bool AndPredicateNode::allow(const DocId docId, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (!arg->allow(docId, rank)) {
+            return false;
+        }
+    }
+    return true;
```
vekterli: Consider if this should be simplified with [`std::ranges::all_of`](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of.html)

searchlib/src/vespa/searchlib/expression/and_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+bool AndPredicateNode::allow(const document::Document & doc, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (!arg->allow(doc, rank)) {
+            return false;
+        }
+    }
+    return true;
```
vekterli: Consider if this should be simplified with [`std::ranges::all_of`](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of.html)

searchlib/src/vespa/searchlib/expression/or_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+    return is;
+}
+
+bool OrPredicateNode::allow(const DocId docId, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (arg->allow(docId, rank)) {
+            return true;
+        }
+    }
+    return false;
```
vekterli: Consider if this should be simplified with [`std::ranges::any_of`](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of.html)

searchlib/src/vespa/searchlib/expression/or_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+bool OrPredicateNode::allow(const document::Document & doc, const HitRank rank) {
+    for (auto& arg : args()) {
+        if (arg->allow(doc, rank)) {
+            return true;
+        }
+    }
+    return false;
```
vekterli: Consider if this should be simplified with [`std::ranges::any_of`](https://en.cppreference.com/w/cpp/algorithm/ranges/all_any_none_of.html)

searchlib/src/tests/expression/filter_predicate_node/filter_predicate_node_test.cpp
```diff
@@ -1,68 +1,143 @@
...
-class RegexFilterTest : public Fixture, public ::testing::Test
-{
-protected:
-    RegexFilterTest();
-    ~RegexFilterTest() override;
-};
+template<typename... Nodes>
+std::unique_ptr<FilterPredicateNode> FilterPredicateNodesTest::make_or(Nodes... nodes) {
+    auto or_predicate = std::make_unique<OrPredicateNode>();
+    (or_predicate->args().push_back(std::move(nodes)), ...);
```
vekterli: Consider `emplace_back` instead of `push_back`. Also, although I enjoy arg pack expansion as much as the next person (some hushed voices would say even more so), I would generally have preferred taking in a span or initializer list with `unique_ptr`s instead to avoid creating a bunch of new function instantiations. But this is fun, too 🙂

searchlib/src/tests/expression/filter_predicate_node/filter_predicate_node_test.cpp
```diff
@@ -1,68 +1,143 @@
...
+    return or_predicate;
+}
 
-RegexFilterTest::RegexFilterTest() = default;
-RegexFilterTest::~RegexFilterTest() = default;
+template<typename... Nodes>
+std::unique_ptr<FilterPredicateNode> FilterPredicateNodesTest::make_and(Nodes... nodes) {
+    auto and_predicate = std::make_unique<AndPredicateNode>();
+    (and_predicate->args().push_back(std::move(nodes)), ...);
```
vekterli: Consider `emplace_back` instead of `push_back`

searchlib/src/tests/expression/filter_predicate_node/filter_predicate_node_test.cpp
```diff
@@ -1,68 +1,143 @@
...
-{
-    EXPECT_TRUE(setup_doc("foobar").setup_node("foo.*").evaluate());
-    EXPECT_FALSE(setup_doc("foobar").setup_node("foo").evaluate());
-    EXPECT_FALSE(setup_doc("foobar").setup_node("bar").evaluate());
+TEST_F(FilterPredicateNodesTest, test_and_all_match) {
+    EXPECT_TRUE(
+        set_node(make_and(
+            make_regex("foo", make_result("foo")),
+            make_regex("bar", make_result("bar")))).
+        evaluate());
```
vekterli, line 153: Consider having a test for `and` and `or` with more than 2 children since they are _n_-ary operators

searchlib/src/vespa/searchlib/expression/multi_arg_predicate_node.cpp
```diff
@@ -0,0 +1,37 @@
...
+        _args.push_back(node.clone());
+    }
+}
+
+void MultiArgPredicateNode::visitMembers(vespalib::ObjectVisitor& visitor) const {
+    visit(visitor, "args", _args);
+}
+
+void MultiArgPredicateNode::selectMembers(const vespalib::ObjectPredicate& predicate,
+                                       vespalib::ObjectOperation& operation) {
```
vekterli: Nit: indent is off

container-search/src/test/java/com/yahoo/search/grouping/vespa/RequestBuilderTestCase.java
```diff
@@ -1088,6 +1104,18 @@ private static String toSimpleName(FilterExpressionNode filterExp) {
...
                 return "Regex [%s]".formatted(simpleName);
+            } else if (filterExp instanceof NotPredicateNode npn) {
+                var simpleName = npn.getExpression().map(LayoutWriter::toSimpleName).orElse("");
+                return "Not [%s]".formatted(simpleName);
+            } else if (filterExp instanceof OrPredicateNode opn) {
+                var simpleName = opn.getArgs().get().stream()
+                        .map(LayoutWriter::toSimpleName).collect(Collectors.joining(", "));
+                return "Or [%s]".formatted(simpleName);
+            }
+            else if (filterExp instanceof AndPredicateNode apn) {
```
vekterli: Nit: redundant line break after closing brace

searchlib/src/main/java/com/yahoo/searchlib/expression/MultiArgPredicateNode.java
```diff
@@ -0,0 +1,93 @@
...
+    public FilterExpressionNode getArg(int i) {
+        return args.get(i);
+    }
+
+    public int getNumArgs() {
+        return args.size();
+    }
+
+    public Optional<List<FilterExpressionNode>> getArgs() {
+        return Optional.ofNullable(args);
```
vekterli: Can `args` ever be `null` in a valid case? The rest of the class seems to always assume non-null, and if this also holds here then there's also probably no need for an `Optional`.

searchlib/src/main/java/com/yahoo/searchlib/expression/MultiArgPredicateNode.java
```diff
@@ -0,0 +1,93 @@
...
+    private List<FilterExpressionNode> args = new ArrayList<>();
+
+    public MultiArgPredicateNode() {}
+
+    public MultiArgPredicateNode(List<FilterExpressionNode> args) {
+        this.args = args;
+    }
+
+    public final MultiArgPredicateNode addArg(FilterExpressionNode arg) {
+        Objects.requireNonNull(arg, "Can not add null arg"); // throws NullPointerException
```
vekterli: This comment is not really needed, the implication is given by `requireNonNull`

searchlib/src/main/java/com/yahoo/searchlib/expression/MultiArgPredicateNode.java
```diff
@@ -0,0 +1,93 @@
...
+        return args.equals(rhs.args);
+    }
+
+    @Override
+    public int hashCode() {
+        int ret = 0;
+        for (FilterExpressionNode node : args) {
+            ret += node.hashCode();
+        }
+        return ret;
```
vekterli: Should this instead just defer to `Objects.hashCode(args)` (which in turns defers to `args.hashCode()` when non-null)? A simple linear combination of sub-hashes is not likely to generate a good resulting hash; `List.hashCode` uses element-wise prime multiplication for that reason (see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#hashCode--)

searchlib/src/main/java/com/yahoo/searchlib/expression/NotPredicateNode.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public FilterExpressionNode clone() {
+        return new NotPredicateNode(expression != null ? expression.clone() : null);
+    }
+
+    @Override
+    protected void onSerialize(Serializer buf) {
+        // Future-proofing: allow optional expression in protocol in case it's added later to the grouping language
+        serializeOptional(buf, expression);
```
vekterli, line 44: Does this comment apply here? I'm not sure what the semantics of `not` node with an empty expression should be...

searchlib/src/vespa/searchlib/expression/and_predicate_node.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+Serializer& AndPredicateNode::onSerialize(Serializer& os) const {
+    return os << args();
+}
+
+Deserializer& AndPredicateNode::onDeserialize(Deserializer& is) {
+    is >> args();
+    return is;
+}
```
vekterli: `onSerialize` and `onDeserialize` is currently duplicated for `AndPredicate` and `OrPredicate`. Since they are both handled by `MultiArgPredicateNode` in Java, it makes sense for this to be the case here as well.

Also, since `IdentifiablePtr` is serialized as an _optional_ object, it seems to me that we either have to detect `nullptr`-ness of deserialized objects here and fail the deserialization with an appropriate exception, or we have to skip `nullptr` objects during traversal. I prefer the former since I don't see any valid semantics for missing elements for predicates.

Otherwise we'll send the search core to Narnia when dereferencing missing objects during calls to `allow`.

searchlib/src/vespa/searchlib/expression/filter_predicate_node.h
```diff
@@ -11,6 +11,8 @@ namespace search::expression {
 class FilterPredicateNode : public vespalib::Identifiable
 {
 public:
+    using CP = vespalib::IdentifiablePtr<FilterPredicateNode>;
```
vekterli: Although `IdentifiablePtr` is-a `CloneablePtr`, this should preferably be renamed to `IP` to avoid confusion since `IdentifiablePtr` has explicit serialization/deserialization semantics

vekterli (commented):
A few comments for consideration~ 🥸

searchlib/src/main/java/com/yahoo/searchlib/expression/MultiArgPredicateNode.java
johansolbakken: Agree. Removing optional and assuming not null where this class is used.

vekterli (approved):
:+1:

vekterli:
@arnej27959 @bjorncs FYI

searchlib/src/vespa/searchlib/expression/and_predicate_node.cpp
```diff
@@ -0,0 +1,28 @@
...
+AndPredicateNode::AndPredicateNode() noexcept = default;
+
+AndPredicateNode::~AndPredicateNode() = default;
+
+AndPredicateNode::AndPredicateNode(const AndPredicateNode&) = default;
+
+AndPredicateNode& AndPredicateNode::operator=(const AndPredicateNode&) = default;
+
+bool AndPredicateNode::allow(const DocId docId, const HitRank rank) {
+    return std::ranges::all_of(args(), [docId, rank](auto arg){ return arg->allow(docId, rank); });
```
arnej27959, line 22:
```suggestion
    return std::ranges::all_of(args(), [docId, rank](auto &arg){ return arg->allow(docId, rank); });
```
This (and the other 3 similar places) needs to operate on the wired argument, not on a copy of it.

johansolbakken: Thanks, this solved the issues I had!

PR #34515, opened 2025-07-24
Adding dimension `ClusterSpec.Type` to `nodes.active` to compare against alive in alerting

bjormel:
…nst `alive` in alerting

### DIFF
--- a/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
+++ b/node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
@@ -201,7 +201,7 @@ public final class Node implements Nodelike {
     public Optional<Allocation> allocation() { return allocation; }
 
     /** Returns the current allocation when it must exist, or throw exception there is not allocation. */
-    private Allocation requireAllocation(String message) {
+    public Allocation requireAllocation(String message) {
         final Optional<Allocation> allocation = this.allocation;
         if ( ! allocation.isPresent())
             throw new IllegalStateException(message + " for  " + hostname() + ": The node is unallocated");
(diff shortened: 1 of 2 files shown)


node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
```diff
@@ -100,13 +101,17 @@ private void updateAllocationMetrics(NodeList nodes) {
                     nonActiveNodes++;
                 }
             }
+            AtomicReference<ClusterSpec.Type> clusterType = new AtomicReference<>();
+            allocatedNodes.stream().findFirst().flatMap(Node::allocation).ifPresent(allocation -> {
```
olaaun: No need for AtomicReference.
Consider replacing ifPresent with get. The map is already filtered on `node.allocation().isPresent()`

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/Node.java
```diff
@@ -201,7 +201,7 @@ public boolean hasParent(String hostname) {
     public Optional<Allocation> allocation() { return allocation; }
 
     /** Returns the current allocation when it must exist, or throw exception there is not allocation. */
-    private Allocation requireAllocation(String message) {
+    public Allocation requireAllocation(String message) {
```
olaaun: Why public?

bjormel: I am too old, 👴 ⚒️

node-repository/src/main/java/com/yahoo/vespa/hosted/provision/maintenance/MetricsReporter.java
bjormel: Thanks

olaaun (approved):

PR #34517, opened 2025-07-25
Magnus/lsp grouping and not or

Mangern:
Fix highlighting and completion of and, or, not in filter expressions in grouping language.


johansolbakken (approved):
👍🏻

PR #34521, opened 2025-07-27
Refer to canonical names

bratseth:


arnej27959 (dismissed):
Documentation mentions 
`query.type` still, should that be picked up as well? Also not sure where/how the `type` alias triggers.

bratseth:
Aliases are not allowed in query profiles, but added support for setting this as request parameters with alias as well.

arnej27959 (approved):
👍

PR #34524, opened 2025-07-29
Fix parameter in filter-first heuristic

boeker:
In the filter-first heuristic, the size of the two-hop neighborhood was compared to the maximum size of a three-hop neighborhood. This was not intended. This fixes this and adapts the default value of `acorn-one-exploration`/`ranking.matching.acornOneExploration` accordingly. It now defaults to `0.3`. (To retain the behavior from before this change, one has to multiply the parameter by `2 * max-links-per-node`. By default, this is `32`.)

In addition, an unnecessary filter check is removed.


earti (approved):
Looks good to Brage and me!

Issue #34525, opened 2025-07-29
Add a ranking feature to the global-phase which represents the current relevance score


dainiusjocas:
**Is your feature request related to a problem? Please describe.**

When ranking profiles are composed, the ranking profile which contains the global-phase has no good way to take the current relevance score:
- `firstPhase` represents the score after the first-phase ranking
- `secondPhase` may not be specified at all (e.g. when only the global reranking makes sense)
- Calculating the second-phase ranking function might be expensive.
- specifying `secondPhase` in the `match-features` triggers the calculation of the feature.

Consider ranking profiles:

```
rank-profile base {
  first-phase {
    expression: random()
  }
}

rank-profile global {
  first-phase {
    expression: if (firstPhase > 50, firstPhase, -1)
    rerank-count: 10000
  }
}

rank-profile final inherits base, global {}
```

Then consider if a rank profile with the second phase is added:

```

rank-profile with_second_phase {
  second-phase {
    expression: random()
    rerank-count: 100
  }
}

rank-profile final_2 inherits base, global, with_second_phase {}
```

Now `final_2` just discards the second phase score. Which 

**Describe the solution you'd like**

In the global phase, I'd like to use the relevance score as it was received from content nodes, regardless of whether the second phase was calculated or not.

**Describe alternatives you've considered**

Duplicate rank profiles and use `secondPhase` when the `second-phase` is defined. This approach is not great because (2) it recalculates the second-phase score for all hits being reranked in the global phase. This hurts in particular if we have many content nodes (e.g. 20) and the global phase operates on more than `rerankCount` of the second phase. Also, (2) when the global phase expression has some logic which you don't want to duplicate.

To overcome the duplication problem, we can wrap the score in a ranking function, and override the value in the leaf ranking profile. E.g.:

```
rank-profile global2 {
  function current_relevance_score() {
     expression: firstPhase
  }
  first-phase {
    expression: if (current_relevance_score > 50, current_relevance_score, -1)
    rerank-count: 10000
  }
}

rank-profile final_3 inherits base, global2, with_second_phase {
  function current_relevance_score() {
    expression: secondPhase
  }
}
```

Other approach which almost works is to leverage the `firstPhaseRank` match feature:

```
rank-profile global2 {
  function current_relevance_score() {
     expression: firstPhase
  }
  first-phase {
    expression: if (current_relevance_score > 50, current_relevance_score, -1)
    rerank-count: 10000
  }
}

[...]

PR #34526, opened 2025-07-29
feat: bind relevanceScore to the global-phase scorer

dainiusjocas:
Follow up on https://github.com/vespa-engine/vespa/issues/34525

In the global phase, let's add the `relevanceScore` ranking expression that represents the relevance score of the hit.

This value typically would be either the first or the second phase score.

The `secondPhase` doesn't represent the scaled scores for hits below the `second-phase.rerank-count`.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/ranking/HitRescorer.java
+++ b/container-search/src/main/java/com/yahoo/search/ranking/HitRescorer.java
@@ -49,6 +49,8 @@ class HitRescorer {
     }
 
     private static double evalScorer(WrappedHit wrapped, Evaluator scorer, List<MatchFeatureInput> fromMF) {
+        // Add before match-features so that the value can be overriden.
+        scorer.bind("currentRelevanceScore", Tensor.from(wrapped.getScore()));
         for (var argSpec : fromMF) {
             var asTensor = wrapped.getTensor(argSpec.matchFeatureName());
             if (asTensor != null) {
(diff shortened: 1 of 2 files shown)


laurynasjs (approved):

kjocevicius (approved):

arnej27959 (dismissed):
This looks good to me, so if the name is OK let’s go ahead 👍

bratseth:
Maybe "score" is unnecessary and currentRelevance is enough?

dainiusjocas:
> Maybe "score" is unnecessary and currentRelevance is enough?

`currentRelevance` would be better.

Or `relevanceScore` because in java object `hit.getRelevance().getScore();`

bratseth:
Yes agreed, relevanceScore is also good.

container-search/src/main/java/com/yahoo/search/ranking/HitRescorer.java
```diff
@@ -49,6 +49,8 @@ void runNormalizers() {
     }
 
     private static double evalScorer(WrappedHit wrapped, Evaluator scorer, List<MatchFeatureInput> fromMF) {
+        // Add before match-features are bind, so that the value can be overridden.
```
arnej27959:
```suggestion
        // Add before match-features are bound, so that the value can be overridden.
```

arnej27959 (changes requested):
small language fix in comment; rest looks very good

bratseth (approved):

arnej27959:
OK it seems I didn't remember how strict we made this; will fix shortly...

PR #34527, opened 2025-07-30
Magnus/lsp illegal argument handling

Mangern:
Fixed bug where parser would crash if an invalid regex was given in grouping filter(regex(...)).
Generalized handling of illegal arguments in the parsers.
Minor change to uri element in onnx-model (which is not supported so there is always an illegal argument).


johansolbakken (approved):
👍🏻

PR #34532, opened 2025-07-31
add statistics to CrossLinker

arnej27959:
* also, do copy to temporary file and move into place.

@hmusum please review


hmusum (approved):
👍

PR #34533, opened 2025-07-31
Magnus/lsp commands

Mangern:
This adds some lsp/executeCommand capabilities to the language server:
- get schemas: list the schemas in the current workspace
- get schema fields: given a schema, retrieve the fields.
- write yql query: open a yql file and write a query to it

These are useful to control the language server externally, needed for the code in the browser project


johansolbakken (approved):
👍🏻

PR #34534, opened 2025-07-31
Arnej/global phase fix

arnej27959:
@bratseth or @hmusum please review

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/RankProfile.java
+++ b/config-model/src/main/java/com/yahoo/schema/RankProfile.java
@@ -1214,7 +1214,7 @@ public class RankProfile implements Cloneable {
             }
             List<FeatureList> addIfMissing = new ArrayList<>();
             for (String input : needInputs) {
-                if (input.startsWith("constant(") || input.startsWith("query(")) {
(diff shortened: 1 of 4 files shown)


arnej27959:
dainiusjocas FYI

hmusum (approved):
👍

dainiusjocas:
What is the key idea here?

arnej27959:
key ideas:
- The actual expression is done by [FunctionEvaluator](https://github.com/vespa-engine/vespa/blob/master/model-evaluation/src/main/java/ai/vespa/models/evaluation/FunctionEvaluator.java) which requires `bind()` for exactly the input arguments needed by the global-phase ranking expression, `bind()` for unknown name is not allowed
- resolving the sources for inputs is done during setup, not per query
- constant values are extracted and wired during setup as well
- values from the query are extracted once per query
- values from match-features (and now also `relevanceScore` if needed) are extracted from each `Hit`
- adding extra match-features (and validating that all inputs will be present) is done at deploy time, in [config-model](https://github.com/vespa-engine/vespa/pull/34534/files#diff-06bf1ce36d331b080f2f023d5ba863e0c79acd02c5d23e1a0775478b2c80b5f2)

I hope that makes sense (it's more complex than I remembered at first).

PR #34539, opened 2025-08-01
Use `constexpr` `std::bit_cast` for BFloat16 float<->u16 conversions

vekterli:
@arnej27959 please review

`memcpy` is not `constexpr`, so using it in a  `constexpr` function in an actual compile-time context will fail. Using `std::bit_cast` avoids this issue.


arnej27959 (approved):
👍

PR #34540, opened 2025-08-01
Tokenize early in linguistics mode

bratseth:
Since 'linguistics' mode will do all token processing in one pass, tokenization should be invoked instead of segmentation, as segmentation returns the original form of each token, anticipating further processing.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/yql/YqlParser.java
+++ b/container-search/src/main/java/com/yahoo/search/yql/YqlParser.java
@@ -23,8 +23,12 @@ import com.yahoo.geo.DistanceParser;
 import com.yahoo.geo.ParsedDegree;
 import com.yahoo.language.Language;
 import com.yahoo.language.detect.Detector;
+import com.yahoo.language.process.LinguisticsParameters;
 import com.yahoo.language.process.Normalizer;
 import com.yahoo.language.process.Segmenter;
+import com.yahoo.language.process.StemMode;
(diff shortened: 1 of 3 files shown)


container-search/src/main/java/com/yahoo/search/yql/YqlParser.java
```diff
@@ -1685,26 +1691,46 @@ private TaggableItem segment(String field, OperatorNode<ExpressionOperator> ast,
...
         Substring substring = getSubstring(ast);
-        Language usedLanguage = language == null ? currentlyParsing.getLanguage() : language;
-        if (substring != null) {
+        if (substring != null)
             toSegment = substring.getValue();
+        List<String> segments = segment(ast, toSegment, language);
+        return itemFromSegments(segments, field, ast, wordData, toSegment, fromQuery, parent);
+    }
+
+    private List<String> segment(OperatorNode<ExpressionOperator> ast, String toSegment, Language language) {
```
arnej27959, line 1700: Maybe rename this method to `segmentOrTokenize`?

arnej27959 (approved):
Minor suggestion only

container-search/src/main/java/com/yahoo/search/yql/YqlParser.java
bratseth: Too long to be worth it imho

PR #34541, opened 2025-08-01
Ignore output to stdout

hmusum:
Avoid filling log with stdout output from dependencies that unfortunately write to stdout in some situations. We will not get output to stdout from our own code, but that should use proper logging anyway, so I don't think this should be an issue. Trying to change this with standalone container first.


bratseth:
I think some application devs write to stdout to debug?

hmusum:
Yes, and I guess some of our devs could do that in some cases as well, so not a good ide. Closing

PR #34543, opened 2025-08-01
Carry oveer the weight of the source term to the composite

bratseth:
This is necessary because a phrase enforces that children added have the same weight as itself.


arnej27959 (approved):
Looks good to me 👍

PR #34546, opened 2025-08-02
Give a warning if select-elements-by references undefined summary feature

hmusum:


bratseth (approved):
Isn't it better to just throw right away?

hmusum:
Burnt child dreads the fire... And I'm on vacation, so don't want to place burden on others. Will throw in next PR

PR #34553, opened 2025-08-04
Magnus/lsp small bugs

Mangern:
Add some missing rank feature properties.
Ignore diagnostics on onnx-model outside rank-profile (which is deprecated).


sebastiannberg (approved):
lgtm

PR #34560, opened 2025-08-05
chore(lsp): Support significance {} with highlighting & completion

Mangern:
Main fix: Add the significance {} block to semantic tokens and completion.
But also sorted the long lists alphabetically.


sebastiannberg (approved):
🚀

PR #34565, opened 2025-08-06
add casts for "char" variants

arnej27959:
I've experimented with this for replacing some of the `reinterpret_cast` code converting between `unsigned char *`  and `char *`

Not sure if the name `char_p_cast` is the best; we could also consider `sign_cast` (possibly extending to other similar casts).

I have also included a string literal operator after some research, this can replace various instances of `char_from_u8` and can actually be used in constexpr contexts (doesn't need `reinterpret_cast`).

Please review/discuss.


arnej27959:
@toregge please review and merge

toregge (approved):
👍

PR #34566, opened 2025-08-06
add "none" pooling strategy; use when ONNX models does it internally

arnej27959:
@glebashnik please review
@bjorncs @thomasht86 FYI


glebashnik:
@arnej27959 Should we have a test for this?
Maybe also include a comment that mentions when to use "none":
> when ONNX models does it internally

Pooling is part of the model?

glebashnik (approved):

PR #34572, opened 2025-08-08
Revert "chore: extract deployment specific flags into a trait" [MERGEOK]

bjormel:
Reverts vespa-engine/vespa#34301


esolitos (approved):
💢  I knew it..

PR #34573, opened 2025-08-11
Throw if select-elements-by references undefined summary feature

hmusum:
Followup to https://github.com/vespa-engine/vespa/pull/34546

Note that this fails one unit test, it seems like `getSummaryFeatures()` in `RankProfile` will not give the full summary feature, e.g. when using `elementwise(bm25(foo),x,double)` as in the failing test.

@arnej27959 do you know of any other way of doing this that will work?

### DIFF
--- a/config-model/src/main/java/com/yahoo/schema/processing/SummaryElementsSelectorValidator.java
+++ b/config-model/src/main/java/com/yahoo/schema/processing/SummaryElementsSelectorValidator.java
@@ -16,7 +16,6 @@ import com.yahoo.vespa.documentmodel.SummaryField;
 import com.yahoo.vespa.model.container.search.QueryProfiles;
 
 import java.util.Set;
-import java.util.logging.Level;
 
 import static com.yahoo.schema.document.ComplexAttributeFieldUtils.isSupportedComplexField;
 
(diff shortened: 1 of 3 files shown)


config-model/src/main/java/com/yahoo/schema/processing/SummaryElementsSelectorValidator.java
```diff
@@ -74,9 +73,7 @@ private void verifySelectElementsBy(DocumentSummary summary, SummaryField field,
         if (!found) {
```
arnej27959, line 73: Problem is on the previous line; the "s" has just "elementwise" as its name returned by .getName(); use .toString() instead to also get the arguments.

arnej27959 (commented):
a fix is needed for validation to work correctly

config-model/src/main/java/com/yahoo/schema/processing/SummaryElementsSelectorValidator.java
hmusum: Of course. Thanks!

arnej27959 (approved):
👍

PR #34574, opened 2025-08-11
Set distance to empty tensor fields to infinity in DistanceCalculator

boeker:
Currently, the distance to an empty tensor field is `std::numeric_limits<double>::max()` instead of `std::numeric_limits<double>::infinity()`.

This leads to unexpected behavior when using exact vector search via the `nearestNeighbor` operator. It can actually return documents that do not contain a vector for the tensor field, while both approximate search via HNSW and streaming search do not return such documents.

With this change to `std::numeric_limits<double>::infinity()`, all three search modes do not return such documents.

This fixes https://github.com/vespa-engine/vespa/issues/34514.


arnej27959 (approved):
👍

PR #34575, opened 2025-08-11
Fix check for nested string type

bratseth:
Unnecessarily checking for equality of the parent type meant that weighted sets with non-default settings for create-if-nonexistent etc. would not be treated as containing the given type.


bratseth:
Or @hmusum

arnej27959 (approved):
👍

PR #34578, opened 2025-08-12
ensure that comparisons handle NaN consistently

arnej27959:
@bjorncs please review
@vekterli @toregge FYI

### DIFF
--- a/searchlib/src/main/java/com/yahoo/searchlib/expression/FloatBucketResultNode.java
+++ b/searchlib/src/main/java/com/yahoo/searchlib/expression/FloatBucketResultNode.java
@@ -88,32 +88,34 @@ public class FloatBucketResultNode extends BucketResultNode {
         FloatBucketResultNode b = (FloatBucketResultNode)rhs;
         double f1 = from;
         double f2 = b.from;
-        if (Double.isNaN(f1)) {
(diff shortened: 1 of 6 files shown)


searchlib/src/main/java/com/yahoo/searchlib/expression/FloatBucketResultNode.java
```diff
@@ -88,13 +88,25 @@ protected int onCmp(ResultNode rhs) {
...
         if (f1 < f2) {
             return -1;
         } else if (f1 > f2) {
             return 1;
         } else {
             double t1 = to;
             double t2 = b.to;
+            if (Double.isNaN(t2)) {
+                return 1;
+            }
```
bjorncs: `t1` and `t2` are not equal even when `.to` both is `NaN`?

bjorncs (dismissed):

searchlib/src/main/java/com/yahoo/searchlib/expression/FloatBucketResultNode.java
arnej27959: right, needs fix

bjorncs (approved):

PR #34579, opened 2025-08-13
Output effective limit and, if applicable, why it is used

hmusum:
Effective limit is equal to configured limit if feed is not blocked, so no change in that scenario.


johansolbakken (approved):
👍🏻

PR #34584, opened 2025-08-14
add geoBoundingBox() syntax

arnej27959:
@bratseth please review
may want some more sanity checks in parsing?
I'm also considering adding something to Location API as the coordinate reordering is very non-intuitive.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/query/SelectParser.java
+++ b/container-search/src/main/java/com/yahoo/search/query/SelectParser.java
@@ -409,14 +409,14 @@ public class SelectParser implements Parser {
     }
 
     private Item buildGeoBoundingBox(String key, Inspector value) {
-        System.err.println("buildGeoBoundingBox: value=" + value);
         String field = value.entry(0).asString();
         var coord_1 = value.entry(1).asDouble();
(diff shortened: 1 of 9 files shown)


container-search/src/main/java/com/yahoo/search/query/SelectParser.java
```diff
@@ -406,6 +408,19 @@ private CompositeItem buildOr(String key, Inspector value) {
         return orItem;
     }
 
+    private Item buildGeoBoundingBox(String key, Inspector value) {
+        System.err.println("buildGeoBoundingBox: value=" + value);
```
bratseth:
```suggestion
```

bratseth (commented):
LGTM, notice the System.err leftover. 

Regarding the reordering the setBoundingBox: This method initializes the members and can only be called once, so it would be better done during initialization. How about fixing both these things at once, by deprecating the setter and adding a static Location.fromBoundingBox() that uses the new argument order? Should do the same for setGeoCircle then.

arnej27959:

bratseth (approved):
nice

PR #34588, opened 2025-08-15
Range filter in grouping

johansolbakken:
@arnej27959 please review
@bjorncs fyi

The range operator in grouping:
```
filter(range(1990, 2010, field))
filter(range(1990, 2010, field, true, true))
```

This PR:
- Adds `RangePredicate` class in Java public API.
- Extends GroupingParser.jj with the required grammar. 
- Adds range predicate node class in Java and C++, and allocates 178 in identifiable.h for this node.

### DIFF
--- a/searchlib/src/vespa/searchlib/expression/range_predicate_node.h
+++ b/searchlib/src/vespa/searchlib/expression/range_predicate_node.h
@@ -9,10 +9,10 @@ namespace search::expression {
 /**
  **/
 class RangePredicateNode : public FilterPredicateNode {
-    double _lower;
-    double _upper;
-    bool _lower_inclusive;
-    bool _upper_inclusive;
+    double _lower{};
+    double _upper{};
+    bool _lower_inclusive{};
+    bool _upper_inclusive{};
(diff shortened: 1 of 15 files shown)


searchlib/src/vespa/searchlib/expression/range_predicate_node.h
```diff
@@ -0,0 +1,41 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+#pragma once
+
+#include "expressiontree.h"
+#include "filter_predicate_node.h"
+
+namespace search::expression {
+
+/**
+ **/
```
arnej27959, line 14: Please add a simple description of the class here.
It's also missing from FilterPredicateNode and RegexPredicateNode (my bad), so try to make up something for those as well while you're at it.

arnej27959 (dismissed):
This looks good to me!

johansolbakken:
Fixing the language server errors

arnej27959 (approved):
looks great 👍

PR #34593, opened 2025-08-16
Deploy after modifying reindexing parameters

hmusum:
Makes it possible to change indexing speed to 0 (basically stopping reindexing) and have that change deployed immediately.


bratseth (approved):
👍

PR #34595, opened 2025-08-18
Make logical predicate filters in grouping infix instead of functional

johansolbakken:
Convert grouping filter predicates from functional to **infix** operators for readability. Introduces `filterPrimary` (parses unary expressions and atoms) and restructures the parser as `filterExp → orPredicate → andPredicate → notPredicate → filterPrimary`, yielding conventional precedence and associativity. `not` has the highest precedence (unary), `and` binds tighter than `or`, and both `and`/`or` are left-associative. Parentheses are supported via `filterPrimary` for explicit grouping. No semantic changes beyond syntax and explicit precedence.

```txt
# Before (functional)
and(or(a, b), not(c))

# Now (infix)
a or b and not c          # parsed as: a or (b and (not c))
(a or b) and not c        # explicit grouping
```

Also expands tests for precedence, associativity, unary `not`, and parentheses.

### DIFF
--- a/container-search/src/main/java/com/yahoo/search/grouping/request/AndPredicate.java
+++ b/container-search/src/main/java/com/yahoo/search/grouping/request/AndPredicate.java
@@ -28,9 +28,9 @@ public class AndPredicate extends FilterExpression {
 
     @Override
     public String toString() {
-        return "and(" + args.stream()
+        return args.stream()
                 .map(Object::toString)
-                .collect(Collectors.joining(", ")) + ")";
+                .collect(Collectors.joining(" and "));
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


container-search/src/main/java/com/yahoo/search/grouping/request/AndPredicate.java
```diff
@@ -28,9 +28,9 @@ public List<FilterExpression> getArgs() {
 
     @Override
     public String toString() {
-        return "and(" + args.stream()
+        return args.stream()
```
arnej27959: the toString() operations here don't add any parentheses anywhere, so you will lose the intended semantics if parens are necessary.  Simplest is probably to always add probably-unnecessary parens.

johansolbakken: Thanks, will fix it

arnej27959 (approved):
looks good to me

PR #34596, opened 2025-08-18
Refactored sidecars flag, added id and  minCpu, maxCpu, validation

glebashnik:


flags/src/main/java/com/yahoo/vespa/flags/custom/Sidecar.java
```diff
@@ -1,104 +1,46 @@
...
-                && Objects.equals(volumeMounts, that.volumeMounts)
-                && Objects.equals(envs, that.envs)
-                && Objects.equals(command, that.command);
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(image, quota);
+@JsonInclude(value = JsonInclude.Include.NON_NULL)
+public record Sidecar(
```
hakonhall, line 19: To be explicit and consistent with other and future flag custom classes, I would prefer if you add @JsonCreator and @JsonProperty annotations like you had before.

Same with the other custom flag classes in this PR.

hakonhall (approved):

PR #34600, opened 2025-08-19
Allow any items inside SameElement

bratseth:
@toregge PTAL

I want to add some schema-dependent validation separately.


toregge (approved):
@havardpe , @vekterli : FYI

PR #34605, opened 2025-08-20
Revert "Don't reduce SameElement with one child"

toregge:
Reverts vespa-engine/vespa#34601

This broke 2 system tests on factory for streaming search due to streaming search backend not handling SameElement with one child. Fix for streaming search backend is in PR #34607, this revert is not needed when that fix is merged.


bratseth:
Not needed as https://github.com/vespa-engine/vespa/pull/34607 is merged.

PR #34606, opened 2025-08-20
chore(jdisc-core): destroy stale containers after a grace period

bjorncs:
FYI @gjoranv

### DIFF
--- a/container-disc/src/main/java/com/yahoo/container/jdisc/metric/MetricUpdater.java
+++ b/container-disc/src/main/java/com/yahoo/container/jdisc/metric/MetricUpdater.java
@@ -2,11 +2,10 @@
 package com.yahoo.container.jdisc.metric;
 
 import ai.vespa.metrics.ContainerMetrics;
-import ai.vespa.metrics.ContainerMetrics;
 import com.yahoo.component.AbstractComponent;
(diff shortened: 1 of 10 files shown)


jdisc_core/src/main/java/com/yahoo/jdisc/core/DeactivatedDeactivatedContainerWatchdog.java
```diff
@@ -31,13 +31,12 @@
  *
  * @author bjorncs
  */
-class ContainerWatchdog implements ContainerWatchdogMetrics, AutoCloseable {
+class DeactivatedDeactivatedContainerWatchdog implements DeactivatedContainerWatchdogMetrics, AutoCloseable {
```
gjoranv: Is the repetition intentional? If so, what does it mean?

gjoranv (dismissed):

jdisc_core/src/main/java/com/yahoo/jdisc/core/DeactivatedDeactivatedContainerWatchdog.java
bjorncs: No - PTAL...

gjoranv (approved):

PR #34607, opened 2025-08-20
Handle SameElement with one child in streaming search.

toregge:
@havardpe or @bratseth : please review

With this applied, the revert in #34605 is not needed.


bratseth (approved):
👍

PR #34616, opened 2025-08-21
Support one content node in a group being down at a time

hmusum:
Rough draft of how this can be done. No support for more than one node at a time or configuring this in services.xml. Probably want to use another config as well, so >1 node is easier to express.

### DIFF
--- a/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
+++ b/clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
@@ -248,6 +248,10 @@ public class NodeStateChangeChecker {
                 return Optional.of(result);
             if (anotherNodeInGroupAlreadyAllowed(nodeInfo, description))
                 return Optional.of(allow());
+        } else if (maxNumberOfGroupsAllowedToBeDown == 0) {
+            result = otherNodeHasWantedState(nodeInfo);
+            if (result.notAllowed())
+                return Optional.of(result);
         } else {
             return checkIfOtherNodesHaveWantedState(nodeInfo, description, clusterState);
         }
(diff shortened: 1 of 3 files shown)


clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -204,24 +205,15 @@ private Result canSetStateUp(NodeInfo nodeInfo, NodeState oldWantedState) {
...
 
-    private Result canSetStateMaintenanceTemporarily(StorageNodeInfo nodeInfo, ClusterState clusterState,
-                                                     String newDescription) {
-        var result = checkIfStateSetWithDifferentDescription(nodeInfo, newDescription);
+    private Result canSetStateMaintenanceTemporarily(StorageNodeInfo nodeInfo, ClusterState clusterState, String description) {
+        var result = checkIfStateSetWithDifferentDescription(nodeInfo, description);
         if (result.notAllowed())
             return result;
 
         if (isGroupedSetup()) {
```
hakonhall: This looks good to me.  An alternative to changing these lines is to make this one-liner:

```suggestion
        if (isGroupedSetup() && maxNumberOfGroupsAllowedToBeDown != 0) {
```

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -384,16 +394,17 @@ private Result otherNodeHasWantedState(StorageNodeInfo nodeInfo) {
...
             if (index == nodeInfo.getNodeIndex()) continue;
 
+            var message = maxNumberOfGroupsAllowedToBeDown == 0
+                    ? "At most one node can have a wanted state: Other %s %d has wanted state %s"
+                    : "At most one node can have a wanted state when #groups = 1: Other %s %d has wanted state %s";
             State storageNodeWantedState = clusterInfo.getStorageNodeInfo(index).getUserWantedState().getState();
             if (storageNodeWantedState != UP) {
-                return disallow("At most one node can have a wanted state when #groups = 1: Other storage node " +
-                                index + " has wanted state " + storageNodeWantedState);
+                return disallow(message.formatted("storage node", index, storageNodeWantedState));
```
hakonhall: Nice.  Also, `clusterInfo.getStorageNodeInfo(index).type()` returns "storage node", and `clusterInfo.getDistributorNodeInfo(index).type()` returns "distributor, so perhaps it can be simplified even more. :thinking:

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
```diff
@@ -384,16 +394,17 @@ private Result otherNodeHasWantedState(StorageNodeInfo nodeInfo) {
             int index = configuredNode.index();
             if (index == nodeInfo.getNodeIndex()) continue;
 
+            var message = maxNumberOfGroupsAllowedToBeDown == 0
+                    ? "At most one node can have a wanted state: Other %s %d has wanted state %s"
+                    : "At most one node can have a wanted state when #groups = 1: Other %s %d has wanted state %s";
```
hakonhall: I think it's fine to just remove the " #groups = 1" part of the message to make them equal.

hakonhall (dismissed):

clustercontroller-core/src/main/java/com/yahoo/vespa/clustercontroller/core/NodeStateChangeChecker.java
hmusum: Nice, good suggestion, will do that

hakonhall (approved):

PR #34620, opened 2025-08-21
Serialize WordAlternatives inside Near

bratseth:
Also, serialize word annotations on items inside near


bratseth:
This is better than it was (failing on serialization), but I see we're also missing parsing of WordAlternatives annotations.

bratseth:
.. no strike that. This is good as is. PTAL @arnej27959

arnej27959 (approved):
we should probably have some more system tests for various combinations with alternatives inside other operators...

bratseth:
Agreed ...

PR #34625, opened 2025-08-24
add "keep" as synonym for "filter"

arnej27959:
as decided in architecture meeting.

@bratseth or @bjorncs please review
@johansolbakken FYI


bratseth (approved):
👍

PR #34626, opened 2025-08-24
Bratseth/application inheritance

bratseth:
@hmusum please review, I believe this is functionally complete apart from multiple inheritance and inheriting same-named artefacts.

This mixes cleanup and functionality, which is unfortunate, but it's the best I was able to do. I suggest going commit by commit, there is some backtracking but not very much.


config-application-package/src/main/java/com/yahoo/config/model/application/provider/FilesApplicationPackage.java
```diff
@@ -867,4 +553,121 @@ private boolean isSchemasSubDir(java.nio.file.Path relativeDirectory) {
...
+                                                  Map<String, FilesApplicationPackage> inheritableApplications) {
+        return fromDir(appDir, deployData, false, inheritableApplications);
+    }
+
+    /** Creates package from a local directory, typically deploy app   */
+    public static FilesApplicationPackage fromDir(File appDir,
+                                                  DeployData deployData,
+                                                  boolean includeSourceFiles,
+                                                  Map<String, FilesApplicationPackage> inheritableApplications) {
+        return new Builder(appDir).includeSourceFiles(includeSourceFiles).deployData(deployData).build();
```
hmusum:
```suggestion
        return new Builder(appDir).includeSourceFiles(includeSourceFiles)
                                  .deployData(deployData)
                                  .inheritableApplications(inheritableApplications)
                                  .build();
```

config-application-package/src/main/java/com/yahoo/config/model/application/provider/ApplicationDefinition.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Optional;
+
+/**
+ * Represents the content of the optional application.xml file in application packages.
+ *
+ * @author bratseth
+ */
+public class ApplicationDefinition {
+
+    private static final String applicationTag = "application";
```
hmusum: Unused?

bratseth: Yes (I removed validation in code) - removed, thanks!

hmusum (approved):
Overall design and structure looks good, less complicated than feared.

~2 comments, one of them seems to be a bug / missing a call in a builder method, that should be fixed, otherwise loos good~ Fixed now, great!

PR #34633, opened 2025-08-26
stop using boost::iterator_facade

arnej27959:
@vekterli please review

### DIFF
--- a/storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
+++ b/storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
@@ -3,7 +3,6 @@
 
 #include "prioritizedbucket.h"
 #include <vespa/storage/bucketdb/bucketdatabase.h>
-#include <boost/iterator/iterator_facade.hpp>
 
 namespace storage::distributor {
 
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
```diff
@@ -21,26 +20,30 @@ class BucketPriorityDatabase
...
-        >
+    class ConstIterator final
     {
         ConstIteratorImplPtr _impl;
     public:
+        using difference_type = std::ptrdiff_t;
+        using value_type = PrioritizedBucket const;
+        using pointer = value_type *;
+        using reference = value_type;
+        using iterator_category = std::input_iterator_tag;
```
vekterli: Based on the previous facade code I think this should be `std::forward_iterator_tag` (assuming that's what `boost::forward_traversal_tag` maps to). But see the other comment on whether we should just remove this....

storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
```diff
@@ -21,26 +20,30 @@ class BucketPriorityDatabase
...
+
         explicit ConstIterator(ConstIteratorImplPtr impl) noexcept
             : _impl(std::move(impl))
         {}
+        const PrioritizedBucket operator* () const noexcept { return dereference(); }
+        void operator++() noexcept { increment(); }
+        void operator++(int) noexcept { increment(); }
```
vekterli: If we want to satisfy the forward iterator concept, pre-increment should technically return a ref to `this` and post-increment should return a by-value copy of `this` prior the increment. But this iterator is inherently _not_ copyable... Does it break anything if we _remove_ the `iterator_category` and the post-increment overload entirely and stop pretending that this is a proper iterator, since it has really never been?

vekterli (dismissed):
:+1:

storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
arnej27959: it can't (easily) be a forward_iterator because the underlying implementation returns a temporary instead of a reference to something stored.

storage/src/vespa/storage/distributor/maintenance/bucketprioritydatabase.h
arnej27959: you're right that since this isn't copy-constructible it doesn't even fulfill the LegacyIterator requirements; so we can simplify and stop pretending it does.

vekterli (approved):

PR #34634, opened 2025-08-26
feat(kubernetes-operator): add kubernetes system

bjorncs:


onurkaracali (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
```diff
@@ -59,21 +55,25 @@ public String value() {
             case main -> "main";
             case Public -> "public";
             case PublicCd -> "publiccd";
+            case kubernetes -> "kubernetes";
         };
     }
 
     /** Whether the system is similar to Public, e.g. PublicCd. */
-    public boolean isPublic() { return isPublic; }
+    public boolean isPublic() { return this == Public || this == PublicCd; }
```
hakonhall: You will likely also need a method isMain(), which is what `!isPublic()` used to mean.

hakonhall (dismissed):

config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
bjorncs: Will that not be confusing that `isMain()` also returns true for `cd`?

hakonhall: Either document it (as done with isPublic), or come up with a better name?  Best alternative name I know of is "Yahoo", although that's also a cloud.  `isMainOrCd` (but then also `isPublicOrPublicCd`?)  How about `isPublicLike` and `isMainLike`?

bjorncs: Good idea!

onurkaracali (approved):

PR #34637, opened 2025-08-26
Handle equiv and phrase below near in streaming search.

toregge:
Limitation: does not take phrase length into account.

@vekterli : please review
@arnej27959 : FYI


vekterli (approved):
:+1:

PR #34646, opened 2025-08-27
use vespalib::getStackTraceFrames instead of boost

arnej27959:
@vekterli please review
@toregge FYI

### DIFF
--- a/vespalib/src/vespa/vespalib/util/backtrace.cpp
+++ b/vespalib/src/vespa/vespalib/util/backtrace.cpp
@@ -3,6 +3,7 @@
 #include <vespa/vespalib/util/backtrace.h>
 #include <vespa/vespalib/util/classname.h>
 #include <vespa/vespalib/stllike/asciistream.h>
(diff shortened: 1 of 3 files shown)


vekterli (commented):
Looks like a test is not entirely happy with the change
```
[ RUN      ] SignalHandlerTest.can_dump_stack_of_another_thread
--
  | /workspace/build/buildkite/vespaai/vespa-engine-vespa/vespalib/src/tests/signalhandler/signalhandler_test.cpp:51: Failure
  | Value of: trace
  | Expected: has substring "my_cool_function"
  | Actual: "Backtrace:"
  |  
  | [  FAILED  ] SignalHandlerTest.can_dump_stack_of_another_thread (1 ms)
```
Otherwise LGTM

arnej27959:
as it turns out, absl::GetStackTrace only returns the 4 frames inside signal handling (on x86_64).  So it's useless here; we should probably stop using it inside mimalloc_intercept.cpp as well.

arnej27959:
superseded

PR #34658, opened 2025-08-28
Arnej/cleanup evil logbuffer

arnej27959:
@vekterli please review
@toregge @havardpe FYI

### DIFF
--- a/vespalog/src/vespa/log/bufferedlogger.cpp
+++ b/vespalog/src/vespa/log/bufferedlogger.cpp
@@ -32,14 +32,18 @@ struct EntryKey {
     std::strong_ordering operator<=>(const EntryKey& entry) const = default;
 };
 
+struct PayLoad {
+    Logger::LogLevel level;
+    std::string file;
+    int line;
+    std::string message;
+    system_time timestamp;
+};
+
 /** Struct keeping information about log message. */
 struct Entry : EntryKey {
-    Logger::LogLevel _level;
-    std::string _file;
-    int _line;
-    std::string _message;
     uint32_t _count;
-    system_time _timestamp;
+    PayLoad payload;
 
     Entry(const Entry &);
     Entry & operator=(const Entry &);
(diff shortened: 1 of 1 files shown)


vespalog/src/vespa/log/bufferedlogger.cpp
```diff
@@ -8,16 +8,162 @@
...
+// Let each hit count for 5 seconds
+duration global_countFactor = VESPA_LOG_COUNTAGEFACTOR * 1s;
+
+// Don't let tokens from different loggers match each other, but
+// if in the same logger, you should have full control. Overlapping
+// tokens if you want is a feature.
+struct EntryKey {
+    Logger* const _logger;
+    const std::string _token;
+    std::strong_ordering operator<=>(const EntryKey& entry) const = default;
```
vekterli, line 34: This could probably also safely be tagged `noexcept`

vespalog/src/vespa/log/bufferedlogger.cpp
```diff
@@ -93,7 +183,7 @@ class BackingBuffer {
     void flush();
 
     /** Gives all current content of log buffer. Useful for debugging. */
-    std::string toString() const;
+    std::string toString();
```
vekterli, line 186: Can this still be `const`? It looks to me as if the function impl doesn't do anything mutable

vespalog/src/vespa/log/bufferedlogger.cpp
```diff
@@ -246,143 +255,113 @@ BufferedLogger::doLog(Logger& l, Logger::LogLevel level,
...
+        system_time oldest;
+        const Entry *toRemove = nullptr;
+        for (const auto & entry : _cache) {
+            system_time adjusted = entry.getAgeFactor();
+            if (toRemove == nullptr || adjusted < oldest) {
+                oldest = adjusted;
+                toRemove = &entry;
+            }
+            if (--numToCheck == 0) break;
+        }
```
vekterli, line 318: Although not likely an issue in practice since we're generally dealing with small log caches, we may consider ensuring sub-linear complexity for finding the eviction candidate when the cache is at capacity.

vekterli (approved):
:+1:

PR #34672, opened 2025-08-29
abort messages in priority order

arnej27959:
* Stripe::abort() was the only place iterating messages in sequence, but it shouldn't matter if we do it in priority order instead.
* Also, avoid exposing the internal PriorityQueue which was only used one place.


vekterli (approved):
:+1:

PR #34674, opened 2025-08-29
Disable mimalloc on darwin.

toregge:
@vekterli : please review


vekterli (approved):
:+1: very strange how `__attribute__((weak))` seemingly works fine in the existing malloc info explorer code (for detecting the runtime malloc impl), but not for this...

PR #34689, opened 2025-09-01
Replace multi_index_container with std::set pair

arnej27959:
replacing this boost::multi_index_container seems to be pretty straight-forward, keeping the code structure unchanged.
since this queue only holds api::CreateVisitorCommand it should only be performance critical for streaming search, I would guess.

@vekterli please review
@toregge FYI


vekterli (approved):
:+1:

PR #34690, opened 2025-09-01
use std::map/set for multi_index_container

arnej27959:
This is the most complicated replacement so far.
Note that in-place writes to the entries were expected, so std::set as the main container was out.
On the other hand, there was no iteration in sequence-order (anymore), so I could use
unordered_map (for possible better performance).
Because of how the iterators were used, I found I should model the standard input iterator closely,
hiding the underlying iterator better.
As a bonus, this meant we got non-const iterators some places more, enabling std::move a couple
of places where comments indicated it would be preferrable.

@vekterli and @toregge please review

### DIFF
--- a/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
+++ b/storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
@@ -736,13 +736,13 @@ FileStorHandlerImpl::remapMessage(api::StorageMessage& msg, const document::Buck
 void
 FileStorHandlerImpl::remapQueueNoLock(const RemapInfo& source, std::vector<RemapInfo*>& targets, Operation op)
 {
-    BucketIdx idx = stripe(source.bucket).exposeBucketIdx();
+    BucketIdxView idx = stripe(source.bucket).exposeBucketIdxView();
     auto range(idx.equal_range(source.bucket));
 
     std::vector<MessageEntry> entriesFound;
 
     // Find all the messages for the given bucket.
-    for (BucketIdx::iterator i = range.first; i != range.second; ++i) {
+    for (BucketIdxView::iterator i = range.first; i != range.second; ++i) {
         assert(i->_bucket == source.bucket);
         MessageEntry& entry = *i;
         entriesFound.push_back(std::move(entry));
@@ -853,8 +853,8 @@ FileStorHandlerImpl::Stripe::failOperations(const document::Bucket &bucket, cons
 {
     std::lock_guard guard(*_lock);
 
-    BucketIdx idx = exposeBucketIdx();
-    std::pair<BucketIdx::iterator, BucketIdx::iterator> range(idx.equal_range(bucket));
+    BucketIdxView idx = exposeBucketIdxView();
+    std::pair<BucketIdxView::iterator, BucketIdxView::iterator> range(idx.equal_range(bucket));
 
     for (auto iter = range.first; iter != range.second;) {
         // We want to post delete bucket to list before calling this
@@ -982,8 +982,8 @@ FileStorHandlerImpl::Stripe::next_message_impl(monitor_guard& guard, vespalib::s
     // second attempt. This is key to allowing the run loop to register
     // ticks at regular intervals while not busy-waiting.
     for (int attempt = 0; (attempt < 2) && !_owner.isPaused(); ++attempt) {
-        PriorityIdx idx = exposePriorityIdx();
-        PriorityIdx::iterator iter(idx.begin()), end(idx.end());
+        PriorityIdxView idx = exposePriorityIdxView();
+        PriorityIdxView::iterator iter(idx.begin()), end(idx.end());
         bool was_throttled = false;
 
         while ((iter != end) && operationIsInhibited(guard, iter->_bucket, *iter->_command)) {
@@ -1077,7 +1077,7 @@ FileStorHandlerImpl::Stripe::fill_feed_op_batch(monitor_guard& guard, LockedMess
 {
     assert(batch.size() == 1);
     assert(guard.owns_lock());
-    BucketIdx idx = exposeBucketIdx();
+    BucketIdxView idx = exposeBucketIdxView();
     auto bucket_msgs = idx.equal_range(batch.lock->getBucket());
     // Process in FIFO order (_not_ priority order) until we hit the end, a non-batchable operation
     // (implicit pipeline stall since bucket set might change) or can't get another throttle token.
@@ -1129,8 +1129,8 @@ FileStorHandlerImpl::Stripe::get_next_async_message(monitor_guard& guard)
     if (_owner.isPaused()) {
         return {};
     }
-    PriorityIdx idx = exposePriorityIdx();
-    PriorityIdx::iterator iter(idx.begin()), end(idx.end());
+    PriorityIdxView idx = exposePriorityIdxView();
+    PriorityIdxView::iterator iter(idx.begin()), end(idx.end());
 
     while ((iter != end) && operationIsInhibited(guard, iter->_bucket, *iter->_command)) {
         ++iter;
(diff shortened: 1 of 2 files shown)


storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -154,15 +288,17 @@ class FileStorHandlerImpl final
...
         [[nodiscard]] FileStorHandler::LockedMessageBatch next_message_batch(vespalib::steady_time now, vespalib::steady_time deadline);
         void dumpQueue(std::ostream & os) const;
         void dumpActiveHtml(std::ostream & os) const;
         void dumpQueueHtml(std::ostream & os) const;
         [[nodiscard]] std::mutex & exposeLock() { return *_lock; }
         void queue_emplace(MessageEntry entry) { _queue->emplace_back(std::move(entry)); }
-        [[nodiscard]] BucketIdx & exposeBucketIdx() { return bmi::get<2>(*_queue); }
+        [[nodiscard]] BucketIdx exposeBucketIdx() { return BucketIdx(*_queue); }
+        [[nodiscard]] PriorityIdx exposePriorityIdx() { return PriorityIdx(*_queue); }
+     // [[nodiscard]] ConstBucketIdx exposeBucketIdx() const { return ConstBucketIdx(*_queue); }
```
vekterli: Nit: leftover comment

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -734,16 +736,16 @@ FileStorHandlerImpl::remapMessage(api::StorageMessage& msg, const document::Buck
 void
 FileStorHandlerImpl::remapQueueNoLock(const RemapInfo& source, std::vector<RemapInfo*>& targets, Operation op)
 {
-    BucketIdx& idx(stripe(source.bucket).exposeBucketIdx());
+    BucketIdx idx = stripe(source.bucket).exposeBucketIdx();
```
vekterli: Maybe we should call this `BucketIdxView` instead of just `BucketIdx` to make it clear to a casual observer that we're not (accidentally or otherwise) copying an entire index structure by value

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.cpp
```diff
@@ -980,7 +982,7 @@ FileStorHandlerImpl::Stripe::next_message_impl(monitor_guard& guard, vespalib::s
     // second attempt. This is key to allowing the run loop to register
     // ticks at regular intervals while not busy-waiting.
     for (int attempt = 0; (attempt < 2) && !_owner.isPaused(); ++attempt) {
-        PriorityIdx& idx(bmi::get<1>(*_queue));
+        PriorityIdx idx = exposePriorityIdx();
```
vekterli: See other comment re: `Idx` vs `IdxView`

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+        using ByPriCmp = OrderCmp<compareByPriority>;
+        using ByBucketCmp = OrderCmp<compareByBucket>;
+
+        using ByPriSet = std::set<EntryPtr, ByPriCmp>;
+        using ByBucketSet = std::set<EntryPtr, ByBucketCmp>;
+
+        uint64_t _next_sequence_id = 1;
+        EntryMap _main_map;
+        ByPriSet _sequence_ids_by_priority;
+        ByBucketSet _sequence_ids_by_bucket;
```
vekterli: Consider making all the internal guts `private` to keep eager hands from poking at things from the outside

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+
+        size_t size() const { return _main_map.size(); }
+        bool empty() const { return _main_map.empty(); }
```
vekterli: Consider adding `noexcept` (and possibly `[[nodiscard]]`) for relevant methods

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+            using reference = value_type &;
+            using iterator_category = std::input_iterator_tag;
+
+            reference operator* () { return deref()->second; };
+            pointer operator-> () { return &deref()->second; };
+            void operator++() { ++_place; }
+            bool operator==(const ordered_iterator& other) {
+                return _place == other._place;
+            }
+            bool operator!=(const ordered_iterator& other) {
```
vekterli: Comparison ops should probably be `const noexcept`

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+            }
+            ordered_iterator(I p) : _place(p) {}
+            ordered_iterator(const ordered_iterator&) = default;
+            ordered_iterator& operator= (const ordered_iterator& other) = default;
+            auto deref() const { return *_place; }
+        private:
+            I _place;
+        };
+        struct ConstPriorityIdx {
+            using iterator = ordered_iterator<ByPriSet::const_iterator, const MapEntry>;
```
vekterli: Should this be `const_iterator`?

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+            PriorityIdx(PriorityQueue& q) : _q(q) {}
+            iterator erase(iterator it) {
+                EntryPtr p = it.deref();
+                ++it;
+                _q.remove(p->first);
+                return it;
+            }
+        };
+        struct ConstBucketIdx {
+            using iterator = ordered_iterator<ByBucketSet::const_iterator, const MapEntry>;
```
vekterli: Should this be `const_iterator`?

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+            template<typename T>
+            bool operator() (EntryPtr a, const T& b) const { return b.cvt(a) < b.cvt(); }
+            template<typename T>
+            bool operator() (const T& a, EntryPtr b) const { return a.cvt() < a.cvt(b); }
+        };
+        static bool compareByPriority(const MessageEntry& a, const MessageEntry&b) {
+            return a._priority < b._priority;
+        }
+        static bool compareByBucket(const MessageEntry& a, const MessageEntry&b) {
+            return a._bucket < b._bucket;
```
vekterli, line 86: I think it would make sense to hide both `_priority` and `_bucket` behind read-only accessors in `MessageEntry` (and make the fields `private`) to lessen the risk of anyone mutating fields that are used transitively by secondary indexes.

Ideally they would be `const` fields, but the type is copyable and movable today... Maybe it no longer has to be, since this code harkens back to the pre-C++11 era when we did not have `emplace` and had to walk uphill both ways, in the snow.

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
```diff
@@ -67,13 +60,154 @@ class FileStorHandlerImpl final
...
+            }
+            void erase(iterator it, iterator to) {
+                while (it != to) {
+                    it = erase(it);
+                }
+            }
+            struct BucketCompare {
+                const document::Bucket& bucket;
+                const document::Bucket& cvt() const { return bucket; }
+                const document::Bucket& cvt(EntryPtr p) const { return p->second._bucket; }
```
vekterli: I think calling this `convert` instead would be more immediately understandable from the callsite

vekterli (dismissed):
:+1: having a nice generic secondary index container abstraction in `vespalib` that ensures all sneaky invariants are always upheld would be great, as these are variants around the same theme 👀

storage/src/vespa/storage/persistence/filestorage/filestorhandlerimpl.h
arnej27959: Since this was just as bad in the original code, I think I'll make a separate PR for this; we should be able to make those fields const and `MessageEntry` move-only, as far as I can see.

vekterli (approved):
:+1:

PR #34692, opened 2025-09-02
Support adjusting cluster controller memory based on number of conten…

hmusum:
…t nodes

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
@@ -183,25 +183,26 @@ public class CapacityPolicies {
         double adjustment;
         if (isBetween(count, 0, 50)) {
             adjustment = 0;
-        } else if (isBetween(count, 51, 100)) {
+        } else if (isBetween(count, 50, 100)) {
             adjustment = 0.15;
-        } else if (isBetween(count, 101, 200)) {
+        } else if (isBetween(count, 100, 200)) {
             adjustment = 0.3;
-        } else if (isBetween(count, 201, 300)) {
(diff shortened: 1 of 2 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
```diff
@@ -144,13 +155,53 @@ private NodeResources defaultResources(ClusterSpec clusterSpec) {
...
+        }
+
+        double newMemory = memory + adjustment;
+        if (count > 50) {
+            log.log(INFO, "Adjusted cluster controller memory (" + count + " content nodes): " + newMemory + " GiB");
+        }
+        return newMemory;
+    }
+
+    public static boolean isBetween(int x, int lower, int upper) {
```
vekterli, line 204: Maybe either the method name or the variable names should indicate that the value range is inclusive and not exclusive.

vekterli (dismissed):
:+1: Outside the scope of this PR, but should we also consider auto-adjusting the number of cluster controller vCPUs for large clusters? The amount of CPU work needed for any given cluster is—as with memory—superlinear in the number of nodes (in particular due to aggregating cross-node statistics), so things can get pretty sweaty as clusters grow big.

config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
hmusum: Yes, thanks, I think I will change to use a range where lower is inclusive and upper is exclusive

hmusum:
Yes, vCPU dependent on nodes sound right, good point, will look into it after this has been tested a bit

vekterli (approved):

PR #34717, opened 2025-09-04
fix: remove unused imports causing build problems

Mangern:
These imports caused the intellij client build to fail probably due to lsp4j version mismatch.


theodorklauritzen (approved):
LGTM

PR #34719, opened 2025-09-04
Fix setting of address space limit in config

hmusum:
Last part was missing: Actually setting this in the config. Added, and also added to unit test


vekterli (approved):
:+1:

PR #34739, opened 2025-09-08
feat: add mirroring of Almalinux 9

esolitos:
## What

- Added support for multiple AlmaLinux OS versions (8 and 9) in the RPM publishing workflow.
- Improved robustness of workflow scripts by updating shell options, adding variable quotes, replacing for-loop with while-read, and simplifying error checks.

## Why

- Support AL9 in mirrors

### DIFF
--- a/.github/scripts/publish-unpublished-rpms-to-archive.sh
+++ b/.github/scripts/publish-unpublished-rpms-to-archive.sh
@@ -21,12 +21,14 @@ OSVERSION=$2
 ALLOWED_ARCHS=("x86_64" "aarch64")
 ALLOWED_VERSIONS=("8" "9")
 
-if [[ " ${ALLOWED_ARCHS[*]} " != " $RPMARCH " ]]; then
+# shellcheck disable=SC2076
+if [[ ! " ${ALLOWED_ARCHS[*]} " =~ " ${RPMARCH} " ]]; then
   echo "Architecture $RPMARCH not in allowed archs: ${ALLOWED_ARCHS[*]}"
   exit 1
 fi
 
-if [[ " ${ALLOWED_VERSIONS[*]} " != " $OSVERSION " ]]; then
+# shellcheck disable=SC2076
+if [[ ! " ${ALLOWED_VERSIONS[*]} " =~ " ${OSVERSION} " ]]; then
   echo "OS version $OSVERSION not in allowed versions: ${ALLOWED_VERSIONS[*]}"
   exit 1
 fi
--- a/.github/scripts/publish-unpublished-rpms-to-archive.sh
+++ b/.github/scripts/publish-unpublished-rpms-to-archive.sh
@@ -11,37 +11,37 @@ if [[ -n "${RUNNER_DEBUG:-}" ]]; then
     set -o xtrace
 fi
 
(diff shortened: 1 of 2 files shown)


.github/scripts/publish-unpublished-rpms-to-archive.sh
```diff
@@ -1,68 +1,82 @@
...
 rm -f /tmp/vespa-open-source-rpms.repo
 
-readonly COPR_PACKAGES=$(mktemp)
-trap "rm -f $COPR_PACKAGES" EXIT
-readonly DLDIR=$(mktemp -d)
-trap "rm -rf $DLDIR" EXIT
+COPR_PACKAGES=$(mktemp)
+readonly COPR_PACKAGES
+# shellcheck disable=SC2064
+trap "rm -f \"${COPR_PACKAGES}\"" EXIT
```
arnej27959: do we even need to do this type of cleanup given that it's running inside a github action container which will be deleted when the script is done anyway?  something to consider for later, maybe...

arnej27959 (dismissed):
as far as i can see this should be OK

.github/scripts/publish-unpublished-rpms-to-archive.sh
esolitos: Indeed, I did not question the process when I fixed the shellcheck warning.

gjoranv (dismissed):

esolitos:
Do not merge was set because this would bring us over the limit of the OSS storage offered by Cliudsmith.
Need to have a quick chat with @frodelu

.github/scripts/publish-unpublished-rpms-to-archive.sh
```diff
@@ -11,37 +11,37 @@ if [[ -n "${RUNNER_DEBUG:-}" ]]; then
...
-ARCH_VALID=false
-for archname in "${ALLOWED_ARCHS[@]}"; do
-  if [[ "$archname" == "$RPMARCH" ]]; then
-    ARCH_VALID=true
-    break
-  fi
-done
-
-if [[ "$ARCH_VALID" != "true" ]]; then
+if [[ " ${ALLOWED_ARCHS[*]} " != " $RPMARCH " ]]; then
```
arnej27959:
```suggestion
if [[ ! " ${ALLOWED_ARCHS[*]} " =~ " $RPMARCH " ]]; then
```

.github/scripts/publish-unpublished-rpms-to-archive.sh
```diff
@@ -11,37 +11,37 @@ if [[ -n "${RUNNER_DEBUG:-}" ]]; then
...
   echo "Architecture $RPMARCH not in allowed archs: ${ALLOWED_ARCHS[*]}"
   exit 1
 fi
 
+if [[ " ${ALLOWED_VERSIONS[*]} " != " $OSVERSION " ]]; then
```
arnej27959:
```suggestion
if [[ ! " ${ALLOWED_VERSIONS[*]} " =~ " $OSVERSION " ]]; then
```

arnej27959 (commented):
noticed two strange if expressions

gjoranv (approved):

PR #34743, opened 2025-09-08
Mention that rosetta is sadly still needed

bratseth:


README.md
```diff
@@ -90,7 +90,13 @@ for building Vespa, running unit tests and running system tests:
...
 ```
 
-2. For the system Java wrappers to find this JDK, symlink it with
+2. On ARM Macs (M1, M2 etc.), install intel compatibility since [grpc isn't properly maintained](https://github.com/grpc/grpc-java/issues/7690): 
+
+```sh
+softwareupdate --install-rosetta
+```
+
+4. For the system Java wrappers to find this JDK, symlink it with
```
bjorncs: This should be `3.` and the subsequent items updated.

bjorncs (dismissed):
Strange - I've been able to build the Java code without Rosetta installed.

bratseth:
I feel I have too (setting up a new Mac). I think you may easily end up installing it without really noticing if you run some intel binary in some other way.

bjorncs (approved):

PR #34748, opened 2025-09-09
Add search and document api to inheritable app

hmusum:
Isn't something like this needed?


bratseth (approved):
Eh, yes of course - I forgot about that. :-). Thanks!

PR #34749, opened 2025-09-09
Don't expose more info than needed

hmusum:


esolitos (dismissed):

configserver/src/main/java/com/yahoo/vespa/config/server/host/HostRegistry.java
```diff
@@ -44,8 +44,10 @@ public synchronized void update(ApplicationId key, Collection<String> newHosts)
     public synchronized void verifyHosts(ApplicationId applicationId, Collection<String> newHosts) {
         for (String host : newHosts) {
             if (hostAlreadyTaken(host, applicationId)) {
-                throw new IllegalArgumentException("'" + applicationId + "' tried to allocate host '" + host +
-                                                   "', but the host is already taken by '" + host2ApplicationId.get(host) + "'");
+                var message = "'" + applicationId + "' tried to allocate host '" + host +
+                              "', but the host is already taken by";
+                log.log(Level.SEVERE, () -> message + "' " + host2ApplicationId.get(host) + "'");
```
sarzhann: `level.SEVERE` means that it won't be shown in UI to the user, but will be accessible internally?

hmusum: Yes, this will be written to config server log only (vespa.log)

configserver/src/main/java/com/yahoo/vespa/config/server/host/HostRegistry.java
```diff
@@ -44,8 +44,10 @@ public synchronized void update(ApplicationId key, Collection<String> newHosts)
...
+                throw new IllegalArgumentException(message + " another host");
```
sarzhann: `... but the host is already taken by another host`... sounds a bit off 🙈

hmusum: Yes, let me improve it a bit

sarzhann (approved):

PR #34752, opened 2025-09-09
avoid relying on obscure bash features; assume python3 is OK already.

arnej27959:
@esolitos please review

### DIFF
--- a/.buildkite/prepare.sh
+++ b/.buildkite/prepare.sh
@@ -6,10 +6,14 @@ set -x
 
 "$SOURCE_DIR/.buildkite/replace-vespa-version-in-poms.sh" "$VESPA_VERSION" "$SOURCE_DIR"
 
+cd $SOURCE_DIR
+find . -name "*.java" |
+    awk -F/ '{print $2}' | sort -u > /tmp/list.java-modules.txt
+find . -name "package-info.java" -print0 | xargs -0 grep -HnE "@(com.yahoo.api.annotations.)?PublicApi.*"  |
+    awk -F/ '{print $2}' | sort -u > /tmp/list.public-api-modules.txt
+
 # We disable javadoc for all modules not marked as public API
-for MODULE in $(comm -2 -3 \
-                <(find . -name "*.java" | awk -F/ '{print $2}' | sort -u)
-                <(find . -name "package-info.java" -exec grep -HnE "@(com.yahoo.api.annotations.)?PublicApi.*" {} \; | awk -F/ '{print $2}' | sort -u)); do
+for MODULE in $(comm -2 -3 /tmp/list.java-modules.txt /tmp/list.public-api-modules.txt); do
     mkdir -p "$MODULE/src/main/javadoc"
     echo "No javadoc available for module" > "$MODULE/src/main/javadoc/README"
 done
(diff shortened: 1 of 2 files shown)


.buildkite/prepare.sh
```diff
@@ -6,20 +6,20 @@ set -x
...
 
 mkdir -p "$WORKDIR/artifacts/$ARCH/rpms"
 mkdir -p "$WORKDIR/artifacts/$ARCH/maven-repo"
 
-# TODO(aressem): This should move into the vespaengine/docker-image-build-alma* image
-# Make sure we use the latest python3 version installed and ensure that pip is installed
-# shellcheck shell=bash disable=SC2010
-PYBIN="$(ls /usr/bin/python3* | grep -E "/usr/bin/python3.[0-9]+$" |sort -n -k2 -t.|tail -1)"
-alternatives --set python3 "$PYBIN"
-dnf install -y "$(basename "$PYBIN")"-pip
```
esolitos, line 25: Looks good, but we need to make sure we update the default python in AL9, which still uses 3.9:

```bash
❯ docker run --rm -ti vespaengine/vespa-build-almalinux-9 bash

# which python3
/usr/bin/python3

# ls -l /usr/bin/python3
lrwxrwxrwx 1 root root 9 Sep  2 07:19 /usr/bin/python3 -> python3.9

# ls -l /usr/bin/python3*
lrwxrwxrwx 1 root root     9 Sep  2 07:19 /usr/bin/python3 -> python3.9
-rwxr-xr-x 1 root root 68496 Sep  2 07:20 /usr/bin/python3.9
``` 

_While AL8 is correctly using 3.12_

.buildkite/prepare.sh
```diff
@@ -6,20 +6,20 @@ set -x
...
+cd $SOURCE_DIR
+find . -name "*.java" |
+    awk -F/ '{print $2}' | sort -u > /tmp/list.java-modules.txt
+find . -name "package-info.java" -print0 | xargs -0 grep -HnE "@(com.yahoo.api.annotations.)?PublicApi.*"  |
+    awk -F/ '{print $2}' | sort -u > /tmp/list.public-api-modules.txt
+
 # We disable javadoc for all modules not marked as public API
-for MODULE in $(comm -2 -3 \
-                <(find . -name "*.java" | awk -F/ '{print $2}' | sort -u)
-                <(find . -name "package-info.java" -exec grep -HnE "@(com.yahoo.api.annotations.)?PublicApi.*" {} \; | awk -F/ '{print $2}' | sort -u)); do
```
esolitos, line 12: This is not very obscure, but I 100% grant that your implementation is more readable. 👍🏼 

😺

esolitos (dismissed):

.buildkite/prepare.sh
```diff
@@ -6,20 +6,20 @@ set -x
 
 "$SOURCE_DIR/.buildkite/replace-vespa-version-in-poms.sh" "$VESPA_VERSION" "$SOURCE_DIR"
 
+cd $SOURCE_DIR
```
esolitos: shellcheck complains of this. 😊 

```suggestion
cd "${SOURCE_DIR}" || exit 1
```

esolitos (dismissed):

arnej27959:
@esolitos turns out the javadoc hack wasn't working before, and should not be done here anyway, so please take another look now.

esolitos (approved):

esolitos:
Related PR: https://github.com/vespa-engine/docker-image-dev/pull/314

PR #34763, opened 2025-09-10
Use accumulator type for intermediate subtraction result

vekterli:
@arnej27959 please review

Avoids potential precision issues with `BFloat16` element type. This function overload is not yet wired to anything except tests and benchmarks, so this has not been possible to observe as an issue in practice.


arnej27959 (approved):
👍

PR #34775, opened 2025-09-11
Handle GNU-style lambda attributes differently on Clang and GCC

vekterli:
@toregge please review

Clang and GCC refuse to parse `noexcept HWY_ATTR` and `HWY_ATTR noexcept`, respectively. Add a compiler-specific macro to hide which form is actually used when defining lambdas that need compilation target attributes.


toregge (approved):
:+1:

PR #34768, opened 2025-09-11
Wire in BFloat16 acceleration for tensor ops and distance functions

vekterli:
@havardpe and @arnej27959 please review

PR #34779, opened 2025-09-11
Revert "Wire in BFloat16 acceleration for tensor ops and distance functions"

vekterli:
@arnej27959 or @havardpe please review

Some latencies are strongly down, some are strongly up. The latter I think is likely due to no longer implicitly converting to float vectors, meaning that the autovectorized BF16 functions are used, which are very slow. Need to let this one cook a bit more to figure out how to test this without breaking things...


arnej27959 (approved):
👍

vekterli:
Post hoc note: this reverted #34768.

PR #34781, opened 2025-09-11
Make `vespalib::BFloat16` conversions easier to optimize for the compiler

vekterli:
@arnej27959 please review compiler horse-whispering.

By using plain bit-casts and shifts we can seduce the compiler into generating much more efficient auto-vectorized code. Not using struct "aliasing" makes the auto-vectorized dot product and squared Euclidean distance kernels ~3x faster in local benchmarks. This also means we don't have to care about endianness.

Note that the Highway kernels are still measured to be up to several times faster on top of this, since they can directly emit e.g. dedicated BF16 dot product instructions on ARM.


arnej27959 (approved):
std::bit_cast is great. 👍

PR #34786, opened 2025-09-12
Wire BFloat16 acceleration for distance functions and dot/l2 distance operations, take two

vekterli:
@havardpe or @arnej27959 please review. This is just #34768 squashed with no changes.

New attempt now that #34781 is in place.


arnej27959 (approved):
👍

PR #34788, opened 2025-09-12
Fix use `vespamalloc` as default and temporary search for `libmimalloc.so.2`

johansolbakken:
- Use `vespamalloc` as the default if we cannot find `mimalloc`.
- Log a warning to the user if mimalloc cannot be found.
- Search for `libmimalloc.so.2` during testing. This will be removed once the symlink is set up in RPM.

@arnej27959 please review
@hmusum fyi


arnej27959 (approved):
👍

PR #34790, opened 2025-09-12
Revert "Wire BFloat16 acceleration for distance functions and dot/l2 distance operations, take two"

vekterli:
@arnej27959 please review. Merging should be held off until we have at least one perf run with the wiring+Highway both present.

Reverts vespa-engine/vespa#34786

Needs another spin in the tumble dryer. Working on speeding up the legacy BF16 dot product and squared Euclidean distance kernels.


vekterli:
(or @hmusum please review)

hmusum (approved):

PR #34793, opened 2025-09-13
chore(bootstrap): improve script robustness and quoting

esolitos:
Shellcheck fixes:
- Switch the shebang to use env for better portability.  
- Enable strict  error handling with errexit, pipefail, and nounset to catch failures   early.
- Add optional debug tracing controlled by the DEBUG environment variable.
- Fix quoting and array usage for Maven extra  options to correctly handle multiple arguments. 
- Quote variable  expansions consistently to prevent word splitting and globbing issues.


hmusum (approved):

gjoranv:
👍

PR #34795, opened 2025-09-13
Allow WordAltermatives in PhraseSegments

bratseth:
Allow alternatives in phrase segments.

Also fixes some output that was locale dependent.


arnej27959 (commented):
I'm pretty sure this won't work at all in the backend; alternatives inside phrase isn't implemented AFAIK.

bratseth:
It is allowed in PhraseItem though?

arnej27959:
if I remember correctly, it's allowed there because middle management was engaging in wishful thinking, under-estimating how hard things would be to implement, over-estimating how useful it was going to be, and then dropping the ball before the entire thing was finished.

bratseth:
It does seem somewhat unfinished. Also undocumented. However, we create this structure ourselves given the right tokenizer output, so something must be done ...

arnej27959 (dismissed):
This should work in backend now, can you make a system test that uses linguistics to make sure it works all the way?

arnej27959 (approved):
👍

PR #34796, opened 2025-09-14
unroll BF16 loops to 16

arnej27959:
microbenchmark shows this gives a large improvement both on x86_64 and aarch64.

@vekterli please review


vekterli (approved):
:+1: beat me to it 🙂

PR #34801, opened 2025-09-14
Revert back to gofmt

hmusum:
'make' now returns exit code 0 if gofumpt is not installed or cannot be found.
Would be nice to detect this and fail with some hint on how to install gofumpt.


sarzhann:
We could at least fallback to gofmt but completely ignoring formatters doesn't sound like a good idea...

sarzhann:
> We could at least fallback to gofmt but completely ignoring formatters doesn't sound like a good idea...

commented based on the outdated PR message...

sarzhann (approved):

PR #34804, opened 2025-09-15
Wire BFloat16 acceleration for distance functions and dot/l2 distance operations, take 3

vekterli:
@arnej27959 please review. Same as #34786 and #34768.

Testing effects of increased unrolling of auto-vectorized kernels (#34796).


arnej27959 (approved):
👍

PR #34806, opened 2025-09-15
Bump `vespa-build-dependencies` to 1.6.3

vekterli:
@arnej27959 please review. Holding off merge until commit refs are changed elsewhere.

This adds a build-time only dependency, so no changes to the runtime deps.


arnej27959 (approved):
👍

arnej27959 (approved):
👍

PR #34807, opened 2025-09-15
Resolve malloc implementation in config model

johansolbakken:
Assuming `vespa-deps` is `/opt/vespa-deps`. For testing mimalloc.


hmusum (approved):
👍

PR #34808, opened 2025-09-15
add code to handle equiv (actually "alternatives") inside phrase

arnej27959:
@havardpe please review
@toregge FYI


havardpe (commented):
I think we will get issues when searching a view containing multiple fields. The searchable will split the phrase per field and combine them with OR, but the equiv nodes packed inside will potentially all have multiple fields...

havardpe (approved):

PR #34809, opened 2025-09-15
Prepare for non-term children of same element blueprint.

toregge:
@havardpe : please review


havardpe (commented):
SameElementBlueprint should be an intermediate blueprint rather than a complex leaf (to propagate all relevant blueprint operations downwards)

havardpe (approved):

PR #34813, opened 2025-09-16
Remove support for admin v 3.0

hmusum:
Has been deprecated and undocumented for a long time, not in use by any application.

### DIFF
--- a/config-model/src/main/resources/schema/admin.rnc
+++ b/config-model/src/main/resources/schema/admin.rnc
@@ -1,5 +1,5 @@
 # Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
-Admin = AdminV2 | AdminV3 | AdminV4
+Admin = AdminV2 | AdminV4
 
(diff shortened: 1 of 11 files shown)


config-model/src/main/resources/schema/admin.rnc
```diff
@@ -1,5 +1,5 @@
 # Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
-Admin = AdminV2 | AdminV3 | AdminV4
+Admin = AdminV2 | AdminV4
```
esolitos, line 2: We're leaving 2.0?

esolitos (approved):
<3 Less code and less legacy. 😊

config-model/src/main/resources/schema/admin.rnc
hmusum: Yes, 2.0 is for open-source, 4.0 (and 3.0 before this) is for Vespa Cloud.

PR #34815, opened 2025-09-16
Use library filename without version

hmusum:
New dep available that symlinks to versioned file. Simplify.


johansolbakken (approved):
LGTM 👍🏻

PR #34816, opened 2025-09-16
move weighted term vectors

arnej27959:
* Makes it possible to use WeightedStringTermVector other places. Also, allow replicate_subterms to work more generally.

@toregge please review
@havardpe FYI


toregge (approved):
:+1:

PR #34823, opened 2025-09-17
fix(container-core): don't recreate sampler for every request

bjorncs:
Effective sampling rate was 100% as a new sampler instance was created for each request.


hmusum (approved):
👍

arnej27959 (approved):
the new code looks good to me

PR #34825, opened 2025-09-17
Add async signal safe backtracing using `unwind.h` where available

vekterli:
@arnej27959 please review. Signal handler backtrace tests pass on both arm64 and x64. `boost::stacktrace::safe_dump_to` ends up using `_Backtrace_Unwind` when possible, so this should be no less signal safe than what we already have. This implementation is also inspired by the one found in Boost, but simplified.

By using `_Unwind_Backtrace` with a custom callback, we can get stack traces that can punch through a signal handler boundary. The low-level unwinding in Abseil does not seem to do this reliably (at least not on x64), so we can't simply use that.

If there is no `<unwind.h>` fall back to returning zero frames, since we don't have a known async signal safe implementation.

Use this new stack trace method for the cross-thread stack trace dumping functionality.

The changes in #34646 should no longer be needed after this.


arnej27959 (approved):
👍

PR #34828, opened 2025-09-18
Reference counting for triton onnx model loading and unloading

glebashnik:

Implements reference counting for triton onnx model loading and unloading.
This is to avoid race conditions from multiple evaluators during redeployment.

Refactoring of exception handling in TritonClient.

### DIFF
--- a/model-integration/src/main/java/ai/vespa/triton/TritonOnnxModelLoader.java
+++ b/model-integration/src/main/java/ai/vespa/triton/TritonOnnxModelLoader.java
@@ -1,0 +1,86 @@
+package ai.vespa.triton;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.logging.Logger;
+
+/**
+ * Implements reference counting for loading and unloading ONNX models in NVIDIA Triton Inference Server.
+ * This is to avoid loading models that are already loaded and unloading models that are still in use.
+ * 
+ * @author glebashnik
+ */
+class TritonOnnxModelLoader {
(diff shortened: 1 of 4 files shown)


model-integration/src/main/java/ai/vespa/triton/TritonOnnxModelLoader.java
```diff
@@ -0,0 +1,86 @@
...
+    private final boolean isExplicitControlMode;
+
+    TritonOnnxModelLoader(TritonOnnxClient tritonClient, boolean isExplicitControlMode) {
+        Objects.requireNonNull(tritonClient);
+        this.client = tritonClient;
+        this.isExplicitControlMode = isExplicitControlMode;
+    }
+
+    TritonOnnxClient.ModelMetadata loadModel(String modelName) {
+        Objects.requireNonNull(modelName);
```
bjorncs, line 31: Consider add fine level logging when incrementing the reference and potentially loading the model

model-integration/src/main/java/ai/vespa/triton/TritonOnnxModelLoader.java
```diff
@@ -0,0 +1,86 @@
...
+                if (isExplicitControlMode) {
+                    client.unloadModel(modelName);
+                }
+            } else {
+                modelReferenceCounts.put(modelName, count - 1);
+            }
+        }
+    }
+
+    void unloadAllModels() {
```
bjorncs, line 70: Consider fail on any invocation to `loadModel()` after `unloadAllModels()` has been invoked, assuming that this is a cleanup/destroy method and the loader instance should not be reused afterwards.

bjorncs (approved):

PR #34834, opened 2025-09-18
fix(lsp): add rank score drop limit highlighting

Mangern:
Add highlighting for rank-score-drop-limit keyword in LSP.


theodorklauritzen (approved):
Incredible feature!

PR #34841, opened 2025-09-19
feat(docproc): introduce new result value for invalid input

bjorncs:


arnej27959 (approved):
This looks fine to me.  We can consider moving the InvalidInputException out of indexing-language if we find that the same type of failure makes sense for other document processors.

PR #34847, opened 2025-09-19
feat: add new system for Kubernetes CD

bjorncs:
Add new enum variant for `kubernetescd`.
Use same naming scheme for helper methods proping properties of a system.

### DIFF
--- a/config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
+++ b/config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
@@ -1,6 +1,7 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.config.provision;
 
+import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.Set;
 import java.util.function.Predicate;
@@ -15,10 +16,10 @@ import java.util.stream.Stream;
  */
 public enum SystemName {
 
-    /** Continuous deployment system */
+    /** Yahoo Continuous deployment system */
     cd,
 
-    /** Production system */
+    /** Yahoo Production system */
     main,
 
     /** System accessible to the public */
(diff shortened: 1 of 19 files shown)


config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
```diff
@@ -66,12 +71,23 @@ public String value() {
     // TODO Remove and use isPublicLike() instead
     public boolean isPublic() { return isPublicLike(); }
 
-    public boolean isMainLike() { return this == main || this == cd; }
+    // TODO Remove and use isYahoo() instead
+    public boolean isMainLike() { return isYahoo(); }
+
+    public boolean isYahoo() { return this == main || this == cd; }
```
hakonhall: I don't know - main is also the default system so it would be wrong to call it Yahoo for those cases.  And without also changing HostedSystem and most references to 'main' this is probably more confusing than not.

And why migrate 2 methods to *Like, but then introduce 2 methods not named *Like?

bjorncs: > I don't know - main is also the default system so it would be wrong to call it Yahoo for those cases. And without also changing HostedSystem and most references to 'main' this is probably more confusing than not.

I can take a stab at adding a `default` system and changing the default.

> And why migrate 2 methods to *Like, but then introduce 2 methods not named *Like?

The idea was to only use the `Like()` suffix where the prefix collides with a system name - `isCd()` is ambiguous but `isProd()` is not.

hakonhall: Another possibility is to call `main` "Yahoo's main system" and `CD` "Yahoo's CD system".  In types and identifiers one could write isYahooMain etc, which makes it easier to understand `main` wherever that is seen, but still get the point across that it is a Yahoo system.

bjorncs: It should in theory be safe to rename the enum variants, e.g. `main` => `yahooMain`, but I suspect we do not use `.value()` / `.from(String)` in all our (de)serialization code paths.

hakonhall (approved):

config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
```diff
@@ -11,42 +12,44 @@
...
-    /** Kubernetes */
-    kubernetes;
+    /** Kubernetes Production */
+    kubernetes,
+
+    /** Kubernetes CD */
+    kubernetesCd,
+
+    /** Default system (for unit tests and non-hosted) */
+    Default;
```
esolitos, line 42: Should this be on top perhaps? 😄

config-provisioning/src/main/java/com/yahoo/config/provision/SystemName.java
```diff
@@ -57,21 +60,37 @@ public String value() {
...
+            case kubernetesCd -> "kubernetescd";
+            case Default -> "default";
         };
     }
 
     /** Whether the system is similar to Public, e.g. PublicCd. */
-    public boolean isPublicLike() { return this == Public || this == PublicCd; }
+    public boolean isPublicCloudLike() { return this == Public || this == PublicCd; }
+
+    // TODO Remove and use isPublicCloudLike() instead
```
esolitos, line 71: Outdated TODO?

PR #34853, opened 2025-09-19
Rotate access log if larger than configured limit

bjormel:

### DIFF
--- a/container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
@@ -200,7 +200,7 @@ class LogFileHandler <LOGTYPE> {
         private PageCacheFriendlyFileOutputStream fileOutput = null;
         private long nextRotationTime = 0;
         private long fileSize = 0;
-        private final Duration fileSizeCheckInterval = Duration.ofMinutes(1);
+        final static Duration fileSizeCheckInterval = Duration.ofMinutes(1);
         private Instant lastFileSizeCheck = Instant.now();
         private final String filePattern;  // default to current directory, ms time stamp
         private volatile String fileName;
--- a/container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
+++ b/container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
@@ -19,6 +19,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
+import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


bjormel:
Documented in https://github.com/vespa-engine/documentation/pull/4122

container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -196,6 +206,10 @@ static class LogThread<LOGTYPE> extends Thread {
         private final String symlinkName;
         private final ExecutorService executor = createCompressionTaskExecutor();
         private final NativeIO nativeIO = new NativeIO();
+        private final long rotationSize; // Maximum file size in bytes (0 = disabled)
+        private final long sizeCheckInterval; // Interval between size checks
```
bjorncs: Either include the unit as part of the parameter name (e.g `Ms` suffix) or use `Duration` as type instead.

container-core/src/main/resources/configdefinitions/container.core.access-log.def
```diff
@@ -24,3 +24,9 @@ fileHandler.queueSize int default=10000
 
 # Buffer size for the output stream has a default of 256k
 fileHandler.bufferSize int default=262144
+
+# Maximum file size (in bytes) before rotation. 0 means disabled (only time-based rotation)
+fileHandler.rotationSize long default=0
+
+# Interval in milliseconds between file size checks
+fileHandler.sizeCheckInterval long default=60000
```
bjorncs: Most of our time related config parameters is specified in seconds and with double as unit. Consider doing the same here. If not make sure to indicate in parameter name the unit.

bjorncs: What is the rationale for making this configurable? It's a lot of wiring for a parameter that could potentially be a hardcoded constant somewhere in the Java code

container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -292,23 +310,86 @@ private void internalClose() {
...
             } catch (IOException e) {
                 logger.warning("Failed writing log record: " + Exceptions.toMessageString(e));
             }
         }
 
+        // Helper method to check if it's time to check size
+        private boolean shouldCheckSizeRotation(long now) {
+            // Only check if size rotation is enabled and enough time has passed
+            return rotationSize > 0 && 
+                   (now - lastSizeCheck) >= sizeCheckInterval;
```
bjorncs: Consider using `Duration` to perform time math and comparison.

container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -292,23 +310,86 @@ private void internalClose() {
...
             }
+            
+            if (shouldCheckSizeRotation(now)) {
+                syncFileSizeAndRotateIfNeeded();
+            }
+            
+            // Regular time-based rotation check
             if (now > nextRotationTime || fileOutput == null) {
                 internalRotateNow();
+                approximateFileSize = 0; // Reset size after rotation
```
bjorncs: Consider letting `internalRotateNow()` reset the `approximateFileSize`

container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -292,23 +310,86 @@ private void internalClose() {
...
+                // Write the log record
                 logWriter.write(r, fileOutput);
                 fileOutput.write('\n');
+                
+                // Calculate bytes written for this record
+                long bytesAfterWrite = fileOutput.getTotalBytesWritten();
+                long bytesWritten = bytesAfterWrite - bytesBeforeWrite;
+                
+                // Update approximate size
+                approximateFileSize += bytesWritten;
```
bjorncs: The `approximateFileSize` optimization seems a bit premature unless we have evidence that's sampling the actual file size with a more frequent check interval is too expensive.

container-core/src/main/java/com/yahoo/container/logging/LogFileHandler.java
```diff
@@ -292,23 +310,86 @@ private void internalClose() {
...
+                // Check if rotation is needed based on actual size
+                if (actualSize >= rotationSize) {
+                    logger.info("Rotating log file " + fileName + 
+                               " due to actual size: " + actualSize + " >= " + rotationSize);
+                    internalRotateNow();
+                    approximateFileSize = 0;
+                }
+            } catch (IOException e) {
+                // If we can't check the actual size, continue using approximate
+                logger.log(Level.FINE, "Could not sync file size for " + fileName, e);
```
bjorncs: The exception should probably be logged as a warning akin to the existing `IOException` handling in this class.

container-core/src/main/resources/configdefinitions/container.core.access-log.def
bjormel: @bjorncs  You mean both, or only `sizeCheckInterval`? I guess that a default at 1GB size could be ok?

bjorncs: Only size check interval.

container-core/src/test/java/com/yahoo/container/logging/LogFileHandlerTestCase.java
```diff
@@ -206,6 +213,253 @@ private void testcompression(Compression compression,
...
+
+    private void forceFileSizeCheck(LogFileHandler<String> handler) {
+        try {
+            Field lastCheckField = handler.logThread.getClass().getDeclaredField("lastFileSizeCheck");
+            lastCheckField.setAccessible(true);
+            lastCheckField.set(handler.logThread, Instant.now().minus(Duration.ofMinutes(2)));
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to force file size check", e);
+        }
+    }
```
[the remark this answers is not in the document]
bjormel: @bjorncs Should I introduce a `Clock` to avoid using reflection?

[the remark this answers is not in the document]
bjorncs: Agree - reflection is ugly. There is a `ManualClock` in test utils you can inject for the unit tests.

bjorncs (approved):

PR #34854, opened 2025-09-21
word alternative in phrase segments

arnej27959:
to make "stemming: all" work inside phrases, we need access to per-segment origin from StemList.
and then the PhraseSegment item needed some flexibility; this is technically an ABI change but I'm
guessing this is only used internally.

@bratseth please review


bratseth (dismissed):
👍  and yes, agreed

bratseth (dismissed):

bratseth (approved):
Nice

PR #34858, opened 2025-09-22
Increased Triton Onnx GRPC message size to 128 MIB from default 4 MIB

glebashnik:

### DIFF
--- a/model-integration/src/main/java/ai/vespa/triton/TritonOnnxClient.java
+++ b/model-integration/src/main/java/ai/vespa/triton/TritonOnnxClient.java
@@ -43,7 +43,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  */
 @Beta
 public class TritonOnnxClient implements AutoCloseable {
-    final int MAX_INBOUND_MESSAGE_SIZE_MIB = 32;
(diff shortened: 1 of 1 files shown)


model-integration/src/main/java/ai/vespa/triton/TritonOnnxClient.java
```diff
@@ -59,6 +59,7 @@ public static class TritonException extends RuntimeException {
     public TritonOnnxClient(TritonConfig config) {
         var ch = ManagedChannelBuilder.forTarget(config.target())
                 .usePlaintext()
+                .maxInboundMessageSize(MAX_INBOUND_MESSAGE_SIZE_MIB * 1024 * 1024)
```
bjorncs, line 62: Verify that this parameter to not used to allocated a fixed-size buffer upfront per inference

bjorncs (approved):

model-integration/src/main/java/ai/vespa/triton/TritonOnnxClient.java
glebashnik: As discussed here, client will not allocate a fixed-size buffer and this value is used purely for protection:
https://stackoverflow.com/questions/60248003/java-grpc-server-side-stream-keeps-increasing-max-inbound-message-size

PR #34859, opened 2025-09-22
Disable filter for descendants of near, onear and same element.

toregge:
@havardpe : please review
@vekterli , @arnej27959 : FYI

### DIFF
--- a/searchcore/src/vespa/searchcore/proton/matching/tag_needed_handles.cpp
+++ b/searchcore/src/vespa/searchcore/proton/matching/tag_needed_handles.cpp
@@ -15,15 +15,17 @@ namespace proton::matching {
 class TagNeededHandlesVisitor : public TemplateTermVisitor<TagNeededHandlesVisitor, ProtonNodeTypes>
 {
     uint32_t                 _inspecting_ancestor_nodes;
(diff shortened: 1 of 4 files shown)


searchcore/src/vespa/searchcore/proton/matching/tag_needed_handles.cpp
```diff
@@ -15,62 +15,72 @@ namespace proton::matching {
...
     }
 }
 
 void
 TagNeededHandlesVisitor::visit(ProtonNodeTypes::Equiv& n)
 {
-    maybe_visit_handles(n);
-    // Descendants unpack features to a separate match data instance.
+    maybe_visit_field_specs(n);
+    ++_changed_match_data;
```
havardpe, line 81: also bump inspecting ancestors here?

havardpe (approved):

searchcore/src/vespa/searchcore/proton/matching/tag_needed_handles.cpp
toregge: There should be no need to disable filter here unless there are inspecting ancestors above equiv.

PR #34864, opened 2025-09-24
Don't count RPC `TIMEOUT` code as a network error signal for statistics reporting

vekterli:
@havardpe please review

If an operation reaches its deadline in the content node persistence queues, it will be automatically and immediately failed out with a `TIMEOUT` response code. This means that observing this code at a higher level does not necessarily mean that a network error condition exists; there could just as well be an overload condition.


havardpe (approved):
maybe we need separate codes for different timeouts?

vekterli:
> maybe we need separate codes for different timeouts?

Yes, that would certainly be the best. This is a self-inflicted problem caused by the enum value "aliasing" between MessageBus and StorageAPI error codes.

PR #34873, opened 2025-09-25
fix(config-model): don't store a reference to the deploy state

bjorncs:
The `DeployState` instance contains a reference to the previous model. By keeping the reference to the deploy state in the config model instances, you'll end up with keeping all previous model generations in memory until OoM occur.

FYI @johansolbakken


hakonhall (approved):

hmusum:
Great, thanks!

PR #34875, opened 2025-09-25
just compare Thread directly

arnej27959:
Thread.getId() is deprecated in Java 21


vekterli (approved):
:+1: presumably the use of reference equality is always well-defined for `Thread.currentThread()`?

PR #34894, opened 2025-09-29
Arnej/java 21 profiles

arnej27959:
note: the various "client" packages needed an override of compilerArgs because they use their own `<release>` setting.

@bjorncs please review
@gjoranv @johansolbakken FYI

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -335,10 +335,45 @@
         </profile>
         <profile>
             <id>ffm-preview</id>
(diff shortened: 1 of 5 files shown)


parent/pom.xml
```diff
@@ -335,10 +335,45 @@
         </profile>
         <profile>
             <id>ffm-preview</id>
```
bjorncs, line 337: Consider renaming the profiles if they no longer are `ffm` specific.

bjorncs (approved):

PR #34901, opened 2025-09-30
Give cluster controllers for apps with many content nodes more memory

hmusum:


config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
```diff
@@ -176,39 +176,28 @@ private NodeResources clusterControllerResources(ClusterSpec clusterSpec, Archit
...
-            adjustment = 0.15;
-        } else if (isBetween(count, 100, 200)) {
-            adjustment = 0.3;
-        } else if (isBetween(count, 200, 300)) {
-            adjustment = 0.45;
-        } else {
-            adjustment = 0.6;
+
+        double adjustmentFactor = 0.20; // 20% increase per step for vespa version > 8.588
+        if (vespaMinorVersion <= 588) {
```
vekterli: To avoid future exciting surprises on major version 9 I think it would make sense to pass (and compare with) a `Version` instance and compare with `isBefore` or `isAfter` instead :eyes:

vekterli (dismissed):
:+1:

config-provisioning/src/main/java/com/yahoo/config/provision/CapacityPolicies.java
hmusum: Yeah, the plan was to remove this after a few releases, but this is a good point, thanks

vekterli (approved):

PR #34911, opened 2025-10-01
Refactor add global options for selecting command in `vespa-significance` tool

johansolbakken:
Before this change, providing no input to `vespa-significance` said use `--help`, but providing `--help` did not print any help message, only that `generate` was not provided.

This change makes `vespa-significance` extensible for additional commands and enhances the feedback provided by the tool. Motivated by functionality for generating significance model from indexes.

```shell
$ vespa-significance -h
usage: vespa-significance <command> [-h]
Commands:
  generate     Generate a significance model from a JSONL feed file.

Options:
  -h,--help   Show available commands.
```


hmusum (approved):
Nice!

PR #34920, opened 2025-10-02
Give more memory to cluster controllers in Vespa cloud

hmusum:
Nodes should have more than 1.3 GiB memory, so should be safe to use a sightly larger heap.


vekterli (approved):
:+1:

PR #34921, opened 2025-10-02
Add utility for parsing HTTP `Accept` header

vekterli:
@bjorncs please review. This is what happens when thinking _"surely this feature will be trivial to implement, let me just quickly check the RFC..."_ 🫣👹

This PR adds a matcher utility for deducing which media types a client should ideally receive, as communicated through the HTTP `Accept` header.

This sounds simple enough, but is complicated by the fact that a client may accept _many_ different media types, with varying levels of preference between the types. Media types may also contain arbitrary key/value parameters whose values may be quoted strings. This means that naive string matching or splitting may have subtle, yet exciting, failure modes. As a consequence, this matcher implements a tiny hand-knitted lexer+parser for the micro-DSL of the `Accept` header as specified by [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.accept).

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/AcceptHeaderMatcher.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/AcceptHeaderMatcher.java
@@ -1,0 +1,431 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+package com.yahoo.document.restapi.resource;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * <p>
+ * A matcher for deducing which media types a client should ideally receive, as
+ * communicated through the HTTP <code>Accept</code> header.
+ * </p>
+ * <p>
+ * This sounds simple enough, but is complicated by the fact that a client may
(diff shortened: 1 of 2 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/AcceptHeaderMatcher.java
```diff
@@ -0,0 +1,431 @@
...
+ * whose values may be quoted strings. This means that naive string matching or
+ * splitting may have subtle yet exciting failure modes. As a consequence, this
+ * matcher implements a tiny, bespoke parser for the micro-DSL of the Accept header.
+ * </p>
+ *
+ * @see <a href="https://httpwg.org/specs/rfc9110.html#field.accept">RFC 9110</a>
+ *
+ * @author vekterli
+ */
+public class AcceptHeaderMatcher {
```
bjorncs: Consider making package-private as it's only used from the same Java package.

bjorncs (dismissed):
Solid implementation, although a bit overkill... 😎

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/AcceptHeaderMatcher.java
vekterli: Good point; changed

bjorncs (approved):

vekterli:
> although a bit overkill... 😎

_If it's worth doing, it's worth overdoing_ 😼✨

PR #34926, opened 2025-10-02
improve test input building

havardpe:
Use a fluent API for building test input data instead of nested initializer lists in constructor calls. This makes the test code much more readable and easier to maintain.

Changes:
- Replace ElementPositions and MyTerm constructor with fluent API
- Add doc(), elem(), and pos() methods to MyTerm
- Use variadic template for pos() to support pos(1, 2, 3) syntax
- Update all test cases to use the new fluent API

@toregge please review

### DIFF
--- a/searchlib/src/tests/nearsearch/nearsearch_test.cpp
+++ b/searchlib/src/tests/nearsearch/nearsearch_test.cpp
@@ -27,58 +27,39 @@ using search::queryeval::test::MockElementGapInspector;
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-class ElementPositions {
-    uint32_t _id;
-    uint32_t _len;
-    std::vector<uint32_t> _pos;
-public:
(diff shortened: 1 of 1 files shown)


searchlib/src/tests/nearsearch/nearsearch_test.cpp
```diff
@@ -27,59 +27,40 @@ using search::queryeval::test::MockElementGapInspector;
...
 private:
-    std::vector<uint32_t> _docs;
-    std::vector<ElementPositions> _epos;
+    search::queryeval::FakeResult _result;
 
 public:
-    MyTerm(const std::vector<uint32_t>& doc, const std::vector<ElementPositions>& epos);
-    ~MyTerm();
+    MyTerm() = default;
+    ~MyTerm() = default;
```
toregge, line 36: Consider not inlining destructor, to avoid future warnings with gcc 15.

toregge (approved):

PR #34930, opened 2025-10-03
Use HTTP `Accept` header from client to select JSONL response format

vekterli:
@bjorncs please review

We only return JSONL if the client explicitly indicates it accepts the `application/jsonl` media type, and only if the `application/json` type is not specified with a greater preference.

For backwards compatibility reasons we assume a client will always accept `application/json` if nothing is specified.

### DIFF
--- a/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
+++ b/vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
@@ -120,9 +120,14 @@ import static java.util.stream.Collectors.toUnmodifiableMap;
 /**
  * Asynchronous HTTP handler for /document/v1
  *
- * @author jonmv
+ * @author Jon Marius Venstad
  */
 public final class DocumentV1ApiHandler extends AbstractRequestHandler {
+
+    private static class MediaType {
+        static final String JSON       = "application/json";
+        static final String JSON_LINES = "application/jsonl";
+    }
 
     private static final Duration defaultTimeout = Duration.ofSeconds(180); // Match document API default timeout.
     private static final Duration handlerTimeout = Duration.ofMillis(100); // Extra time to allow for handler, JDisc and jetty to complete.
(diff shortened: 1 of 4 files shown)


vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -1271,15 +1276,34 @@ private void visitWithRemote(HttpRequest request, VisitorParameters parameters,
...
+                // Note: ordering matters! If the client has equal preference for JSONL and JSON,
+                // _we_ prefer to use JSONL. Also, since "application/jsonl" is not IANA-standardized,
+                // we might want to extend this if there are (or will be) widely used aliases.
+                var bestTypes = acceptMatcher.preferredExactMediaTypes(MediaType.JSON_LINES, MediaType.JSON);
+                if (!bestTypes.isEmpty() && MediaType.JSON_LINES.equals(bestTypes.get(0))) {
+                    return true;
+                }
+            } catch (IllegalArgumentException e) {
+                log.fine(() -> "Failed to parse Accept header '%s': %s".formatted(combinedAcceptHeader, e.getMessage()));
+                // TODO should we reject with a Bad Request instead?
```
bjorncs, line 1296: Agree that we should do a 400 response here.

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
```diff
@@ -1271,15 +1276,34 @@ private void visitWithRemote(HttpRequest request, VisitorParameters parameters,
...
+            }
+        } // else: for backwards compatibility, always assume application/json is accepted
```
bjorncs, line 1298: In future we could return a `406 Not Acceptable` if non of the requested types are supported.

bjorncs (approved):

vespaclient-container-plugin/src/main/java/com/yahoo/document/restapi/resource/DocumentV1ApiHandler.java
vekterli: Will do as a later follow-up

PR #34931, opened 2025-10-03
Extract RootItem base class

arnej27959:
Create new RootItem class containing root item functionality from QueryTree, making QueryTree extend RootItem.
Also extract MockItem and FakeWSItem from test cases to correct package.

the reason for this is that I want to add Protobuf Serialization as package-private methods; also I think it reads better that QueryTree is a RootItem rather than CompositeItem directly.

@bratseth please review
@bjorncs FYI

### DIFF
--- a/container-search/abi-spec.json
+++ b/container-search/abi-spec.json
@@ -1452,6 +1452,33 @@
     ],
     "fields" : [ ]
   },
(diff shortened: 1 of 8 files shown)


bratseth (approved):
LGTM

PR #34932, opened 2025-10-03
prepare for java 25

arnej27959:
@gjoranv or @bratseth please review
@bjorncs FYI

### DIFF
--- a/bundle-plugin/src/main/java/com/yahoo/container/plugin/mojo/AssembleContainerPluginMojo.java
+++ b/bundle-plugin/src/main/java/com/yahoo/container/plugin/mojo/AssembleContainerPluginMojo.java
@@ -4,7 +4,6 @@ package com.yahoo.container.plugin.mojo;
 import com.yahoo.container.plugin.util.Artifacts;
 import org.apache.maven.model.Build;
 import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
@@ -17,6 +16,7 @@ import java.nio.file.Paths;
 import java.util.EnumMap;
 import java.util.Map;
 import java.util.jar.JarFile;
+import javax.inject.Inject;
 
 /**
  * @author Tony Vaagenes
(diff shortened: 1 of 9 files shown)


vespajlib/src/test/java/com/yahoo/io/reader/NamedReaderTestCase.java
```diff
@@ -36,6 +36,8 @@ public void testIt() {
     @Test
     public void testMethodMasking() {
         List<Method> unmaskedMethods = ClassValidator.unmaskedMethodsFromSuperclass(NamedReader.class);
+        // there is no problem inheriting the Reader implementation for these:
+        unmaskedMethods.removeIf(m -> m.getName().equals("readAllAsString") || m.getName().equals("readAllLines"));
```
bjorncs: WTF... This unit test seems exceptionally strange. Do we really need to test this?

bundle-plugin/src/main/java/com/yahoo/container/plugin/mojo/AssembleContainerPluginMojo.java
```diff
@@ -17,6 +16,7 @@
 import java.util.EnumMap;
 import java.util.Map;
 import java.util.jar.JarFile;
+import javax.inject.Inject;
```
bjorncs, line 19: FYI using this annotation might cause troubles downstream.

bratseth (dismissed):

gjoranv (dismissed):

vespajlib/src/test/java/com/yahoo/io/reader/NamedReaderTestCase.java
arnej27959: that was my reaction also, let's just remove it

bundle-plugin/src/main/java/com/yahoo/container/plugin/mojo/AssembleContainerPluginMojo.java
arnej27959: according to the [maven docs](https://maven.apache.org/maven-jsr330.html#how-to-use-jsr-330-in-plugins) this is the recommended alternative.

arnej27959:
@bjorncs please take another look now

gjoranv (approved):

bjorncs (approved):

PR #34934, opened 2025-10-03
bump vespa-openssl version MERGEOK

arnej27959:
@toregge please review
@vekterli FYI

### DIFF
--- a/dist/vespa.spec
+++ b/dist/vespa.spec
@@ -33,7 +33,7 @@
 %define _defattr_is_vespa_vespa 0
 %define _command_cmake cmake
 %global _vespa_abseil_cpp_version 20250127.1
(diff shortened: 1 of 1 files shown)


dist/vespa.spec
```diff
@@ -171,7 +171,7 @@ Requires: vespa-xxhash >= 0.8.1
 Requires: xxhash-libs >= 0.8.1
 %endif
 %if 0%{?el8}
-Requires: vespa-openssl >= 3.1.7
+Requires: vespa-openssl >= 3.5.4
```
toregge, line 174: Consider adding a `_vespa_openssl_version` macro to avoid having to repeat the version number later on.

toregge (approved):

PR #34936, opened 2025-10-03
Added Vespa CLI count command

rharige:
## Summary
Added vespa-cli `count` command to address a common developer need for quickly checking document counts during development and testing. Currently requires cumbersome YQL queries for simple counting operations. This command provides a convenient, developer-friendly alternative that follows established CLI patterns.

## Key features:

- Quick document counting: `vespa count`
- Document type filtering: `vespa count --document-type mydoc`
- Accuracy warnings for match-phase limiting
- Consistent with existing CLI design
- Support for verbose flag 

This improves the developer experience by eliminating the need to construct YQL queries for routine document counting tasks during indexing iterations and testing.

### DIFF
--- a/client/go/internal/cli/cmd/count.go
+++ b/client/go/internal/cli/cmd/count.go
@@ -1,0 +1,200 @@
+// Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
+// vespa count command
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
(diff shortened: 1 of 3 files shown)


arnej27959:
we already have too many top-level commands in the vespa cli, and this is not useful enough for that.
Use this shell function instead:
```
vespa-count () {
    local where="true"
    if [ "$1" ]; then
        where="sddocname contains '$1'"
    fi
    vespa query "select * from sources * where $where limit 0" | jq '.root.fields.totalCount'
}
```

PR #34943, opened 2025-10-06
Reset CC node state and resource usage metrics on leadership loss edge

vekterli:
@hakonhall please review

Metric gauge values are "sticky" once set, which potentially causes max-aggregation of metrics _across_ cluster controllers to return stale and unexpected values unless we explicitly zero out metrics when leadership is lost.

This PR zeroes out the following metrics:
 * Node state counts (storage+distributor)
 * Node non-convergence counts (storage+distributor)
 * Content node resource usage


vekterli:
PR build failure appears to be due to an unrelated, transient glitch in the build dependency matrix

hakonhall (dismissed):
Would be nice to link this to the settings of the same metrics in updateClusterStateMetrics(), e.g. via metric name constants.

vekterli:
> Would be nice to link this to the settings of the same metrics in updateClusterStateMetrics(), e.g. via metric name constants.

Agreed, have moved metric and dimension names into constants, PTAL. Note: haven't yet abstracted away the set of distinct states.

hakonhall (approved):

PR #34944, opened 2025-10-06
Add `VespaIndexInspectClient` for `vespa-significance`

johansolbakken:
Executes `vespa-index-inspect` and creates a stream of term and document frequencies from the stdout. 

Example of retrieving all terms with a document frequency greater than one, in term sorted order.
```java
var client = new VespaIndexInspectClient();
try (var rows = client.streamDumpWords(indexDir, field)) {
    rows.filter(r -> r.documentFrequency() > 1)
        .sorted(Comparator.comparing(VespaIndexInspectClient.TermDocumentFrequency::term))
        .forEach(r -> System.out.println(r.term() + "\t" + r.documentFrequency()));
}
```

Unit tests for process management and line parsing.


hmusum (approved):
👍

hmusum (approved):

PR #34951, opened 2025-10-07
let the order be 1 if there is a single match

havardpe:
Note that this will change the order score for multi-term queries with a single match. Previously we required the query to be a single term to obtain an order score of 1 when there was no additional matches.

@hmusum please review


hmusum (approved):
Looks good to me

PR #34952, opened 2025-10-07
Adds `IndexLocator` to `vespa-significance`

johansolbakken:
This PR:
- Adds `IndexLocator` class with tests.
- Adds `SelectionException` to provide details about ambiguous directories (e.g., two schema folders).

The `IndexLocator` resolves the path to a flushed index if the path is trivial or the user has provided the correct flags (future PR). Example: `$VESPA_HOME/var/db/vespa/search/CLUSTER/NODE_INDEX/documents/SCHEMA/0.ready/index/index.flush.1/`

If there are, for instance, two document types, then the SCHEMA must be defined by the user. If not, it is ambiguous, and the `IndexLocator` will throw a `SelectionException`. The CLI can then provide a helpful message with the information from the exception. This applies to both CLUSTER and NODE_INDEX.


hmusum (approved):
Nice!

PR #34957, opened 2025-10-07
Arnej/cleanup after to protobuf 4

arnej27959:
@vekterli please review
@havardpe FYI

some overlapping tests in ToProtobufTest is OK, I think.


vekterli (approved):
:+1:

PR #34963, opened 2025-10-07
Temporarily select index with greatest seq number

johansolbakken:

Works and can be used for initial testing. But have to investigate more.


hmusum (dismissed):
Let's try this 👍

hmusum (approved):

PR #34964, opened 2025-10-08
Use the correct number of maximum links when removing from HNSW index

boeker:
On the lowest level, the HNSW graph has twice the number of links compared to the other levels. This commit also respects this when removing nodes to avoid a drop in recall when deleting a large number of documents.

@arnej27959 please review


boeker:
The unit test makes it look like the previous behavior was intentional. Thoughts?

arnej27959 (approved):
this looks fine to me

PR #34966, opened 2025-10-08
Rename JSONL `trace` output to `sessionTrace`

vekterli:
@bjorncs please review. I'm open to a better name, but as we know, naming stuff is NP-hard (Naming Problems-hard).

This name more accurately represents the semantics of the contents of the trace, and also frees up `trace` for later use if we want to add streaming per-message trace output.

### DIFF
--- a/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/StreamingJsonLinesResponseTest.java
+++ b/vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/StreamingJsonLinesResponseTest.java
@@ -252,7 +252,7 @@ public class StreamingJsonLinesResponseTest {
     }
 
     @Test
-    void trace_is_rendered_as_own_line() throws IOException {
+    void session_trace_is_rendered_as_own_line() throws IOException {
         var f = new Fixture();
         var trace = new Trace(9);
         trace.trace(7, "Poirot's deductions", false);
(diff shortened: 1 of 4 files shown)


vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/StreamingJsonLinesResponseTest.java
```diff
@@ -252,36 +252,34 @@ void provided_tensor_options_are_used_when_rendering_put() throws IOException {
...
         trace.getRoot().addChild(new TraceNode().setStrict(false)
                 .addChild("The butler did it")
                 .addChild("Weapon was a garden gnome"));
-        f.jsonlResponse.writeTrace(trace);
-        // TODO the nested "trace" objects aren't beautiful; reconsider format for JSONL.
-        //  Since trace nodes may technically output "message" objects at any level, need a wrapper.
+        f.jsonlResponse.writeRequestTrace(trace);
         String expected = """
-                {"trace":{"trace":[{"message":"Poirot's deductions"},{"fork":[{"message":"The butler did it"},{"message":"Weapon was a garden gnome"}]}]}}
+                {"requestTrace":{"trace":[{"message":"Poirot's deductions"},{"fork":[{"message":"The butler did it"},{"message":"Weapon was a garden gnome"}]}]}}
```
bjorncs: My suggestions (in random order): `operationTrace`, `documentApiTrace`, `documentTrace`.
The naming `requestTrace` could indicate that it's specific to HTTP.

bjorncs (dismissed):

vespaclient-container-plugin/src/test/java/com/yahoo/document/restapi/resource/StreamingJsonLinesResponseTest.java
vekterli: Agreed that `requestTrace` can be ambiguous, though this is frustratingly hard to avoid:

* `operationTrace`: the trace may cover 1-_n_ sub-operations 🥺
* `documentApiTrace`: all traces today are document API (but actually also Storage API between nodes) so this becomes sort of the same as saying just `trace` 🥹
* `documentTrace`: the trace may cover 0-_n_ documents 👹

My original thought was `sessionTrace`, so maybe I should give that one another whirl as long as it only ever applies to JSONL visitor output...

bjorncs (approved):

PR #34979, opened 2025-10-09
Add validation for removal of dataplane tokens from services.xml

hmusum:
Now requires validation override to remove a token


hmusum:
Not sure who should review, @olaaun perhaps?

olaaun (approved):

PR #34998, opened 2025-10-10
Skip fetching summary when only matchfeatures are needed

dainiusjocas:

It turns out that the logic introduced in #34029 is not enough to avoid sending a fill request.

For the query
```shell
vespa query 'select matchfeatures from sources * where true' \
  'ranking.profile=my_ranking_profile' \
  'trace.level=6' \
  'presentation.summary=default'
```

In the query trace, I see:
```json
{
    "message": "Sending 1 summary fetch requests with jrt/protobuf"
},
{
    "message": "Not resending query during document summary fetching"
}
```
Which means that the [check](https://github.com/vespa-engine/vespa/blob/6d5fb534b78493c53a8d1aee41c2776efe780e9e/container-search/src/main/java/com/yahoo/search/dispatch/rpc/RpcProtobufFillInvoker.java#L85) `partialSummaryHandler.resultAlreadyFilled()` failed.

In the same trace there are plenty such messages:
```json
{
    "message": "Hits need fill([presentation]); hasFilled=[[f:matchfeatures]] in class com.yahoo.prelude.cluster.ClusterSearcher"
}
```
I read it as the evidence that somehow `default` summary class is turned into the `[presentation]` and there are no checks that properly short circuits the `.fill()`.

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
@@ -1,23 +1,18 @@
 // Copyright Vespa.ai. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 package com.yahoo.prelude.fastsearch;
 
-import com.yahoo.prelude.fastsearch.DocumentDatabase;
-import com.yahoo.prelude.fastsearch.DocsumDefinitionSet;
 import com.yahoo.processing.IllegalInputException;
 
 import com.yahoo.search.Query;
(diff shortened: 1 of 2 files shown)


dainiusjocas:
Adding a notebook that demonstrates that, currently, it is not possible to skip `.fill()`, but doing it is possible
https://github.com/dainiusjocas/notes/blob/main/notebooks/vespa-ignoring-fill.ipynb

arnej27959:
using your other PR instead 👍

dainiusjocas:
> using your other PR instead 👍 

Yeah, another one is simpler

PR #35007, opened 2025-10-10
Cache result from streaming search query node evaluate.

toregge:
@havardpe : please review

### DIFF
--- a/searchlib/src/vespa/searchlib/query/streaming/query.cpp
+++ b/searchlib/src/vespa/searchlib/query/streaming/query.cpp
@@ -40,7 +40,8 @@ QueryConnector::QueryConnector(const char * opName) noexcept
     : QueryNode(),
       _opName(opName),
       _index(),
-      _children()
+      _children(),
+      _cached_evaluate_result()
 {
 }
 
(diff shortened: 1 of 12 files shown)


searchlib/src/vespa/searchlib/query/streaming/query.cpp
```diff
@@ -259,15 +279,12 @@ RankWithQueryNode::~RankWithQueryNode() = default;
...
-    for (const auto & qn : getChildren()) {
-        if (qn->evaluate()) {
-            if (first) firstOk = true;
-        }
-        first = false;
+    if (_cached_evaluate_result.has_value()) {
+        return _cached_evaluate_result.value();
     }
-    return firstOk;
+    bool result = !getChildren().empty() && getChildren().front()->evaluate();
```
havardpe, line 285: are there any secret side-effects from evaluating all the children that we need to keep?

havardpe (approved):

searchlib/src/vespa/searchlib/query/streaming/query.cpp
toregge: No.

PR #35010, opened 2025-10-11
Revert "Arnej/protobuf serialization fixes 1"

toregge:
Reverts vespa-engine/vespa#35008

Multiple system tests fails with this PR applied.

@arnej27959 : please review


arnej27959 (approved):
The changes to the various Item classes may be the cause, will adapt them another way for now.

PR #35011, opened 2025-10-11
Avoid `.fill()` by treating `[presentation]` summary class as the default request class

dainiusjocas:

A different attempt to avoid `.fill()` when only `matchfeatures` are needed.  #34998

The idea is to claim that the`[presentation]` summary class is roughly the same as `default` when it comes to deciding that a synthetic summary generated from summary fields is enough.

I don't know if this is an intention behind the `PartialSummaryHandler.PRESENTATION`, but it helps to avoid `.fill()` when only `matchfeatures` are needed.

### DIFF
--- a/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
+++ b/container-search/src/main/java/com/yahoo/prelude/fastsearch/PartialSummaryHandler.java
@@ -142,7 +142,7 @@ public class PartialSummaryHandler {
     }
 
     private static boolean isDefaultRequest(String summaryClass) {
-        return summaryClass == null || summaryClass.equals(DEFAULT_CLASS);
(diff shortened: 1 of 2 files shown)


dainiusjocas:
Adding a notebook that demonstrates that, currently, it is not possible to skip `.fill()`, but doing it is possible
https://github.com/dainiusjocas/notes/blob/main/notebooks/vespa-ignoring-fill.ipynb

dainiusjocas:
@arnej27959 could you have a look? You've worked on the initial support https://github.com/vespa-engine/vespa/pull/34029

arnej27959:
There are definitely some problems here that makes this not work like expected; I will try your change and see how it impacts logic elsewhere.

arnej27959 (approved):
🥇  looks good so far, more fixes in progress

PR #35015, opened 2025-10-13
Improve text in usage vespa-siginficance

johansolbakken:


vespaclient-java/src/main/java/com/yahoo/vespasignificance/CommandLineOptions.java
```diff
@@ -254,7 +254,7 @@ static Options createMergeOptions() {
                 .longOpt(MIN_KEEP_OPTION)
                 .hasArg()
                 .argName("NUMBER")
-                .desc("Filter on document frequency lower than NUMBER.")
+                .desc("Filter out document with frequency lower than NUMBER.")
```
sebastiannberg:
```suggestion
                .desc("Filter out documents with frequency lower than NUMBER.")
```

sebastiannberg (dismissed):

sebastiannberg (approved):

PR #35018, opened 2025-10-13
Revert "Set rotation size for access logs to 100 MiB in vespa cloud"

bjormel:
Reverts vespa-engine/vespa#34916


bjormel:
this is still broken, will not enable in Vespa Cloud until fixed

onurkaracali (approved):

PR #35020, opened 2025-10-13
Skip unpack for non-matching subtrees in streaming search.

toregge:
This makes unpack for streaming search closer to unpack for indexed search.

@havardpe : please review
@arnej27959 : FYI


arnej27959 (approved):
seems ok to me

